PR #1063, opened 2017-02-03
Bugfix: No more empty 'isa' in generated relations

marco-scoppetta:


marco-scoppetta:
please retest this

grakn-graql/src/main/java/ai/grakn/graql/internal/hal/HALConceptRepresentationBuilder.java
```diff
@@ -151,7 +151,8 @@ private static void attachSingleGeneratedRelation(Representation currentHal, Con
             firstRole = (roleTypes.get(otherVarName).equals(HAS_ROLE_EDGE)) ? "" : roleTypes.get(otherVarName) + ":";
         }
 
-        String isaString = "isa " + relationType.map(StringConverter::typeNameToString).orElse("");
+        String relationTypeString=relationType.map(StringConverter::typeNameToString).orElse("");
+        String isaString = (relationTypeString.length()>0) ? "isa " + relationTypeString : "";
```
fppt: You could just check the value of `relationType` directly. That way you can avoid necessarily calling `StringConverter::typeNameToString` if `relationType` is `null` or empty.

PR #1065, opened 2017-02-03
Grakn Client package 

alexandraorth:
+ Turned Loader into a REST API client
+ Created `grakn-client` package to contain Loader Client (and eventually other client classes)
+ Migration no longer depends on engine, uses `LoaderClient`
+ Migration requires user to specify keyspace (`-k`)
+ Removed `ImportController`
+ Graql shell depends on `grakn-client` and uses `LoaderClient`


alexandraorth:
This is a very large pull request that affects multiple components. Please review carefully!

alexandraorth:
please retest this please

grakn-client/src/main/java/ai/grakn/client/LoaderClient.java
```diff
@@ -0,0 +1,343 @@
...
+
+    public LoaderClient(String keyspace, String uri, Consumer<Json> onCompletionOfTask){
+        this.uri = uri;
+        this.keyspace = keyspace;
+        this.queries = new HashSet<>();
+        this.futures = new ConcurrentHashMap<>();
+        this.onCompletionOfTask = onCompletionOfTask;
+        this.batchNumber = new AtomicInteger(0);
+
+        setBatchSize(25);
```
fppt: These values should probably be configurable no?

grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
```diff
@@ -253,38 +245,20 @@ public static void runShell(String[] args, String version, String historyFilenam
 
     private static String loadQuery(String filePath) throws IOException {
             List<String> lines = Files.readAllLines(Paths.get(filePath), StandardCharsets.UTF_8);
-            return lines.stream().collect(Collectors.joining("\n"));
+            return lines.stream().collect(joining("\n"));
     }
 
     private static void sendBatchRequest(String uriString, String graqlPath, String keyspace) throws IOException {
-        byte[] out = ("{\"path\": \"" + escapeJavaScript(graqlPath) + "\"}").getBytes(StandardCharsets.UTF_8);
+        LoaderClient loaderClient = new LoaderClient(keyspace, uriString);
```
fppt: A common use case for the Loader API will be to push data in without caring that much about immediate consistency. With that in mind the rest API it hits should allow you to use a batch graph or a normal graph.

grakn-migration/base/src/main/java/ai/grakn/migration/base/AbstractMigrator.java
```diff
@@ -69,7 +101,7 @@ public AbstractMigrator registerMacro(Macro macro){
             return Optional.of(queryBuilder.parseTemplate(template, data));
         } catch (GraqlTemplateParsingException e){
             LOG.warn("Query was not sent to loader- " + e.getMessage());
-            LOG.warn("See the Grakn engine logs for more detail about loading status and any resulting stacktraces: " + properties.getLogFilePath());
+            LOG.warn("See the Grakn engine logs for more detail about loading status and any resulting stacktraces");
```
fppt: I am still not happy with this. If migration fails I should not have to log into the server to discover why. I should be told immediately. A usability issue which I think still needs to be addressed.

grakn-migration/base/src/main/java/ai/grakn/migration/base/io/MigrationOptions.java
```diff
@@ -67,7 +64,11 @@ public boolean isNo() {
     }
 
     public String getKeyspace() {
-        return command.hasOption("k") ? command.getOptionValue("k") : keyspace;
+        if(!command.hasOption("k")){
```
fppt: YES!

grakn-migration/base/src/main/java/ai/grakn/migration/base/io/MigrationOptions.java
```diff
@@ -67,7 +64,11 @@ public boolean isNo() {
...
+            die("Keyspace missing (-k)");
```
fppt: YES!

grakn-client/src/main/java/ai/grakn/client/LoaderClient.java
alexandraorth: They are through the Java api (and through migration), but not through the configuration files. The problem I encountered was that, as this is a client, it can't use the `ConfigProperties` object in engine. Do you have an ideas how to go about it?

PR #1071, opened 2017-02-06
Add more tests for the `GraknGraph` class

aelred:


aelred:
please retest this please

grakn-graph/src/test/java/ai/grakn/generator/ConceptIds.java
```diff
@@ -32,6 +32,10 @@ public ConceptIds() {
 
     @Override
     public ConceptId generate() {
-        return ConceptId.of(gen(String.class));
+        if (random.nextBoolean()) {
+            return ConceptId.of(gen(String.class));
+        } else {
+            return ConceptId.of("bar");
```
fppt: Whats the idea behind giving a 50/50 shot of something having the id "bar" ?

grakn-graph/src/test/java/ai/grakn/generator/MetaTypeNames.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.common.collect.ImmutableSet;
+
+import java.util.stream.Stream;
+
+import static ai.grakn.graql.internal.util.CommonUtil.toImmutableSet;
+
+/**
+ * Generator that generates meta type names only
+ */
+public class MetaTypeNames extends AbstractGenerator<TypeName> {
```
fppt: A whole class to randomly pick out meta names? Does it really need to be that verbose?

grakn-graph/src/test/java/ai/grakn/generator/Methods.java
```diff
@@ -0,0 +1,76 @@
...
+
+    public void configure(MethodOf methodOf) {
+        this.clazz = methodOf.value();
+    }
+
+    public static Object[] mockParamsOf(Method method) {
+        return Stream.of(method.getParameters()).map(Parameter::getType).map(Methods::mock).toArray();
+    }
+
+    private static <T> T mock(Class<T> clazz) {
```
fppt: `clazz` ha!

grakn-graph/src/test/java/ai/grakn/generator/ConceptIds.java
aelred: It's two reasons: to produce more "real" data (instead of weird unicode characters), and to more frequently produce edge cases (the generator will often produce duplicate concept IDs). This line was what helped me catch a lot of the stranger bugs I found.

grakn-graph/src/test/java/ai/grakn/generator/MetaTypeNames.java
aelred: It's used in just one test: `whenCallingClear_AllMetaConceptsArePresent`. If I remove that call we could merge this into the `TypeNames` generator. It's still helpful for this to exist (but maybe not in its own class), similar to `bar` above: it catches edge cases when using meta type names where they are not expected.

PR #1073, opened 2017-02-06
Introduce low level caching into the core API - WIP - DO NOT MERGE

fppt:
- Get conceptID to wrap an object in order to get rid of redundant code.
- Cache the conceptID
- Change equality measures
- Cache concept construction in the element factory
- Cache super types


fppt:
Low level API changes are really starting to become expensive. Another failed attempt :(

PR #1074, opened 2017-02-06
Engine Task Failover tests

alexandraorth:
+ `TaskFailover` will only restart tasks that were marked as `RUNNING` in an engine when it dies
+ All other tasks will be left in the state and queues they were in when that engine died
+ Added `TaskFailover` test class

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/backgroundtasks/distributed/TaskFailover.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/backgroundtasks/distributed/TaskFailover.java
@@ -45,6 +45,7 @@ import static ai.grakn.engine.backgroundtasks.config.ZookeeperPaths.RUNNERS_STAT
 import static ai.grakn.engine.backgroundtasks.config.ZookeeperPaths.RUNNERS_WATCH;
 import static ai.grakn.engine.backgroundtasks.config.ZookeeperPaths.TASKS_PATH_PREFIX;
 import static ai.grakn.engine.util.ExceptionWrapper.noThrow;
+import static java.lang.String.format;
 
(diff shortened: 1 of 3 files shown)


grakn-engine/src/main/java/ai/grakn/engine/backgroundtasks/distributed/TaskFailover.java
```diff
@@ -144,8 +145,14 @@ private void reQueue(CuratorFramework client, String engineID) throws Exception
...
+            if(taskState.status() == RUNNING) {
+                LOG.debug(format("Engine [%s] stopped, task [%s] requeued", engineID, taskState.getId()));
+                stateStorage.updateState(taskState.status(SCHEDULED));
+                producer.send(new ProducerRecord<>(WORK_QUEUE_TOPIC, id, taskState.configuration().toString()));
+            } else {
+                LOG.debug(format("Engine [%s] stopped, task [%s] not restarted because state [%s]"
+                        , engineID, taskState.getId(), taskState.status()));
+            }
         }
     }
```
bolerio: That pair of operations in the reQueue method, "set state to schedule, then use producer to put in queue", is essentially the "schedule" operation. It's a how  a task is scheduled to run, correct? So ideally there should be a method that does precisely that - implement all the steps necessary to schedule a task. Not a big deal because those are two lines, but it's good to have all major steps of the workflow represented by distinct methods.

alexandraorth: Yep, agreed. I'll make sure to get this in the next `TaskFailover` pull req

PR #1075, opened 2017-02-06
Introduce low level caching into the core API

fppt:
- Get conceptID to wrap an object in order to get rid of redundant code.
- Cache the conceptID
- Change equality measures
- Cache concept construction in the element factory
- Cache super types

### DIFF
--- a/grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
+++ b/grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
@@ -91,8 +91,8 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
     private final String engine;
     private final boolean batchLoadingEnabled;
     private final G graph;
+    private final ElementFactory elementFactory;
 
-    private final ThreadLocal<ElementFactory> localElementFactory = new ThreadLocal<>();
     private final ThreadLocal<ConceptLog> localConceptLog = new ThreadLocal<>();
     private final ThreadLocal<Boolean> localIsOpen = new ThreadLocal<>();
     private final ThreadLocal<String> localClosedReason = new ThreadLocal<>();
@@ -104,6 +104,8 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
         this.graph = graph;
         this.keyspace = keyspace;
         this.engine = engine;
+        elementFactory = new ElementFactory(this);
+
         localIsOpen.set(true);
 
         if(initialiseMetaConcepts()) {
@@ -233,10 +235,6 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
     }
 
     ElementFactory getElementFactory(){
-        ElementFactory elementFactory = localElementFactory.get();
-        if(elementFactory == null){
-            localElementFactory.set(elementFactory = new ElementFactory(this));
-        }
         return elementFactory;
     }
 
@@ -526,7 +524,7 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
         // Relation To Casting
         EdgeImpl relationToCasting = addEdge(relation, foundCasting, Schema.EdgeLabel.CASTING);
         relationToCasting.setProperty(Schema.EdgeProperty.ROLE_TYPE, role.getId().getValue());
-        getConceptLog().putConcept(relation); //The relation is explicitly tracked so we can look them up without committing
+        getConceptLog().trackConceptForValidation(relation); //The relation is explicitly tracked so we can look them up without committing
 
         putShortcutEdges(relation, relation.type());
 
@@ -719,7 +717,6 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
     }
 
     private void clearLocalVariables(){
-        localElementFactory.remove();
         localConceptLog.remove();
     }
 
--- a/grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
+++ b/grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
@@ -568,7 +568,7 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
         InstanceImpl toRolePlayer = (InstanceImpl) to;
 
         String hash = calculateShortcutHash(relation, relationType, fromRole, fromRolePlayer, toRole, toRolePlayer);
-        boolean exists = getTinkerPopGraph().traversal().V(fromRolePlayer.getBaseIdentifier()).
+        boolean exists = getTinkerPopGraph().traversal().V(fromRolePlayer.getId().getRawValue()).
                     local(outE(Schema.EdgeLabel.SHORTCUT.getLabel()).has(Schema.EdgeProperty.SHORTCUT_HASH.name(), hash)).
                     hasNext();
 
@@ -869,7 +869,7 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
                 }
             }
 
-            getTinkerPopGraph().traversal().V(otherCasting.getBaseIdentifier()).next().remove();
+            getTinkerPopGraph().traversal().V(otherCasting.getId().getRawValue()).next().remove();
         }
 
         return relationsToClean;
--- a/grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
+++ b/grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
@@ -87,12 +87,12 @@ import static org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.__.outE;
  */
 public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph, GraknAdmin, EngineGraknGraph {
     protected final Logger LOG = LoggerFactory.getLogger(AbstractGraknGraph.class);
-    private final ElementFactory elementFactory;
     private final String keyspace;
     private final String engine;
     private final boolean batchLoadingEnabled;
     private final G graph;
 
+    private final ThreadLocal<ElementFactory> localElementFactory = new ThreadLocal<>();
     private final ThreadLocal<ConceptLog> localConceptLog = new ThreadLocal<>();
     private final ThreadLocal<Boolean> localIsOpen = new ThreadLocal<>();
     private final ThreadLocal<String> localClosedReason = new ThreadLocal<>();
@@ -105,7 +105,6 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
         this.keyspace = keyspace;
         this.engine = engine;
         localIsOpen.set(true);
-        elementFactory = new ElementFactory(this);
 
         if(initialiseMetaConcepts()) {
             try {
@@ -162,7 +161,7 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
 
     @Override
     public <T extends Concept> T buildConcept(Vertex vertex) {
-        return elementFactory.buildConcept(vertex);
+        return getElementFactory().buildConcept(vertex);
     }
 
     public boolean hasCommitted(){
@@ -234,6 +233,10 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
     }
 
     ElementFactory getElementFactory(){
+        ElementFactory elementFactory = localElementFactory.get();
+        if(elementFactory == null){
+            localElementFactory.set(elementFactory = new ElementFactory(this));
+        }
         return elementFactory;
     }
 
@@ -254,7 +257,7 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
             if(!byPassDuplicates && vertices.hasNext()) {
                 throw new MoreThanOneConceptException(ErrorMessage.TOO_MANY_CONCEPTS.getMessage(key.name(), value));
             }
-            return elementFactory.buildConcept(vertex);
+            return getElementFactory().buildConcept(vertex);
         } else {
             return null;
         }
@@ -264,7 +267,7 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
         Set<ConceptImpl> concepts = new HashSet<>();
         getTinkerTraversal().has(key.name(), value).
             forEachRemaining(v -> {
-                concepts.add(elementFactory.buildConcept(v));
+                concepts.add(getElementFactory().buildConcept(v));
             });
         return concepts;
     }
@@ -315,7 +318,7 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
     @Override
     public EntityType putEntityType(TypeName name) {
         return putType(name, Schema.BaseType.ENTITY_TYPE,
-                v -> elementFactory.buildEntityType(v, getMetaEntityType()));
+                v -> getElementFactory().buildEntityType(v, getMetaEntityType()));
     }
 
     private <V extends Type> V putType(TypeName name, Schema.BaseType baseType, Function<Vertex, V> factory){
@@ -331,12 +334,12 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
     @Override
     public RelationType putRelationType(TypeName name) {
         return putType(name, Schema.BaseType.RELATION_TYPE,
-                v -> elementFactory.buildRelationType(v, getMetaRelationType(), Boolean.FALSE)).asRelationType();
+                v -> getElementFactory().buildRelationType(v, getMetaRelationType(), Boolean.FALSE)).asRelationType();
     }
 
     RelationType putRelationTypeImplicit(TypeName name) {
         return putType(name, Schema.BaseType.RELATION_TYPE,
-                v -> elementFactory.buildRelationType(v, getMetaRelationType(), Boolean.TRUE)).asRelationType();
+                v -> getElementFactory().buildRelationType(v, getMetaRelationType(), Boolean.TRUE)).asRelationType();
     }
 
     @Override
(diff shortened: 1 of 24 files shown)


fppt:
please retest this please

fppt:
I am sorry for the horrors this may bring to us all :(

grakn-graph/src/main/java/ai/grakn/graph/internal/Cache.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Retrieves the object in the cache. If nothing is cached the database is read.
+     *
+     * @return The cached object.
+     */
+    public V get(){
+        if(!cachedValue.isPresent()){
+            V newValue = databaseReader.get();
+            if(newValue == null) return null;
```
aelred: Suspicious `null` alert! If the result is `null`, it won't be cached and will be constantly recalculated. I guess you did this because `Optional` can't contain `null`. It might be best to not use an optional in this case and instead have a boolean flag for whether the value is present.

grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptLog.java
```diff
@@ -40,9 +43,20 @@
  *
  */
 public class ConceptLog {
-    private Set<ConceptImpl> modifiedConcepts;
+
+    //Caches any concept which has been touched before
+    private final Map<ConceptId, ConceptImpl> conceptCache = new HashMap<>();
```
aelred: This might be a good place to use that guava cache: https://github.com/google/guava/wiki/CachesExplained

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationTypeImpl.java
```diff
@@ -66,9 +64,7 @@ public Relation addRelation() {
      */
     @Override
     public Collection<RoleType> hasRoles() {
-        Set<RoleType> roleTypes = new HashSet<>();
-        getOutgoingNeighbours(Schema.EdgeLabel.HAS_ROLE).forEach(role -> roleTypes.add(role.asRoleType()));
-        return roleTypes;
+        return Collections.unmodifiableCollection(cachedHasRoles.get());
```
aelred: Just so you're clear on this behaviour: even though the resulting collection cannot be modified, it may still change internally. For example:
```
roles = thing.hasRoles();
roles.contains(rt); // False

thing.hasRole(rt);

roles.contains(rt); // True, by magic
```

I don't know if this is _bad_, but it's worth thinking about.

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationTypeImpl.java
```diff
@@ -80,6 +76,13 @@ public Relation addRelation() {
     public RelationType hasRole(RoleType roleType) {
         checkTypeMutation();
         putEdge(roleType, Schema.EdgeLabel.HAS_ROLE);
+
+        //Cache the Role internally
+        cachedHasRoles.get().add(roleType);
```
aelred: If you call `hasRole` while the roles are not cached, this will cause them to be looked up from the graph. Obviously not a big deal, but still an unnecessary lookup.

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationTypeImpl.java
```diff
@@ -95,14 +98,34 @@ public RelationType deleteHasRole(RoleType roleType) {
...
         //Add the Role Type itself
-        getGraknGraph().getConceptLog().putConcept(roleTypeImpl);
+        getGraknGraph().getConceptLog().trackConceptForValidation(roleTypeImpl);
 
         //Add the Relation Type
-        getGraknGraph().getConceptLog().putConcept(roleTypeImpl);
+        getGraknGraph().getConceptLog().trackConceptForValidation(roleTypeImpl);
+
+        //Remove from internal cache
+        if(cachedHasRoles.isPresent()) cachedHasRoles.get().remove(roleType);
```
aelred: Perhaps we should have a helper method for this on the cache. It will also make sure you don't forget the `ifPresent` check.

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationTypeImpl.java
```diff
@@ -95,14 +98,34 @@ public RelationType deleteHasRole(RoleType roleType) {
...
+        //Remove from roleTypeCache
+        ((RoleTypeImpl) roleType).deleteCachedRelationType(this);
 
         return this;
     }
+
+    @Override
+    public void innerDelete(){
+        //Force load the cache
+        cachedHasRoles.get();
```
aelred: Do we have to do this? We call `cachedHasRoles.get()` again below.

grakn-graph/src/main/java/ai/grakn/graph/internal/RoleTypeImpl.java
```diff
@@ -124,6 +154,10 @@ public void innerDelete(){
             throw new ConceptException(ErrorMessage.CANNOT_DELETE.getMessage(getName()));
         } else {
             super.innerDelete();
+
+            //Clear all internal caching
+            cachedRelationTypes.clear();
+            cachedDirectPlayedByTypes.clear();
```
aelred: You also need to alert all connected relation types and played-by types that the role has been deleted (so they can clear their cache).

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
```diff
@@ -95,8 +104,16 @@ protected V addInstance(Schema.BaseType instanceBaseType, BiFunction<Vertex, T,
...
-        getSuperSet().forEach(type -> allRoleTypes.addAll(((TypeImpl <?, ?>) type).getOutgoingNeighbours(Schema.EdgeLabel.PLAYS_ROLE)));
-        return filterImplicitStructures(allRoleTypes);
+
+        //Get the immediate plays roles which may be cached
+        allRoleTypes.addAll(cachedPlaysRoles.get());
+
+        //Now get the super type plays roles (Which may also be cached locally within their own context
+        Set<T> superSet = getSuperSet();
+        superSet.remove(this); //We already have the plays roles from ourselves
+        superSet.forEach(superParent -> allRoleTypes.addAll(superParent.playsRoles()));
```
aelred: I think you have a combinatorial explosion here. For example, if you have `a -subs-> b -subs-> c -subs-> d`, then `a.playsRoles()` will call `b.playsRoles()`, `c.playsRoles()` and `d.playsRoles()`.
`b.playsRoles()` will then call both `c.playsRoles()` and `d.playsRoles()` etc.

You can deal with this by either only calling something like `b.directPlaysRoles()`, `c.directPlaysRoles()`. Or, you just call on the direct super-type only: `b.playsRoles()`

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
```diff
@@ -118,7 +135,23 @@ public void innerDelete(){
...
+            //Force load of linked concepts whose caches need to be updated
+            cachedSuperType.get();
+            cachedPlaysRoles.get();
+
             deleteNode();
+
+            //Update neighbouring caches
+            //noinspection unchecked
+            ((TypeImpl<T, V>) cachedSuperType.get()).deleteCachedDirectedSubType(getThis());
+            cachedPlaysRoles.get().forEach(roleType -> ((RoleTypeImpl) roleType).deleteCachedDirectPlaysByType(getThis()));
```
aelred: You also need to alert your subtypes.

grakn-graph/src/main/java/ai/grakn/graph/internal/RoleTypeImpl.java
```diff
@@ -77,10 +100,17 @@
...
-        getIncomingNeighbours(Schema.EdgeLabel.PLAYS_ROLE).
-                forEach(concept -> playedBy.addAll(concept.asType().subTypes()));
-        return playedBy;
+        Set<Type> playedByTypes = new HashSet<>();
+        cachedDirectPlayedByTypes.get().forEach(type -> playedByTypes.addAll(type.subTypes()));
+        return Collections.unmodifiableCollection(playedByTypes);
+    }
+
+    void addCachedDirectPlaysByType(Type newType){
+        cachedDirectPlayedByTypes.get().add(newType);
```
aelred: Possibly these methods should only be adding to the cache if it already has been cached? Otherwise they can sneakily do nothing.

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
```diff
@@ -312,21 +372,29 @@ public T subType(T type){
...
             }
         }
+        return false;
     }
 
     T playsRole(RoleType roleType, boolean required) {
         checkTypeMutation();
+
+        //Update the internal cache of role types played
+        cachedPlaysRoles.get().add(roleType);
```
aelred: Another case where the cache is perhaps unnecessarily loaded.

grakn-graph/src/test/java/ai/grakn/graph/internal/EntityTypeTest.java
```diff
@@ -579,4 +581,38 @@ public void testChangingSuperTypeBackToMetaType(){
...
+        EntityType e3 = graknGraph.putEntityType("entityType3").superType(e1);
+        EntityType e4 = graknGraph.putEntityType("entityType4").superType(e1);
+        EntityType e5 = graknGraph.putEntityType("entityType5");
+        EntityType e6 = graknGraph.putEntityType("entityType6").superType(e5);
+
+        assertEquals(4, e1.subTypes().size());
+        assertTrue("e1 subtypes does not contain e1", e1.subTypes().contains(e1));
+        assertTrue("e1 subtypes does not contain e2", e1.subTypes().contains(e2));
+        assertTrue("e1 subtypes does not contain e3", e1.subTypes().contains(e3));
+        assertTrue("e1 subtypes does not contain e4", e1.subTypes().contains(e4));
```
aelred: `assertThat(e1.subTypes(), containsInAnyOrder(e1, e2, e3, e4));`
will cover all of these (including the size check).

grakn-test/src/test/java/ai/grakn/test/graql/query/MatchQueryTest.java
```diff
@@ -564,17 +566,17 @@ public void testGraqlPlaysRoleSemanticsMatchGraphAPI() {
         GraknGraph graph = movieGraph.graph();
 
         Stream.of(a, b, c, d, e, f).forEach(type -> {
-            Set<Concept> graqlPlaysRoles = qb.match(name(type).playsRole(var("x"))).get("x").collect(toSet());
-            Collection<RoleType> graphAPIPlaysRoles = graph.getType(type).playsRoles();
+            Set<Concept> graqlPlaysRoles = qb.match(name(type).playsRole(var("x"))).get("x").collect(Collectors.toSet());
+            Collection<RoleType> graphAPIPlaysRoles = new HashSet<>(graph.getType(type).playsRoles());
```
aelred: I don't think you need the `new HashSet` call here.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -162,7 +163,7 @@ public GraknAdmin admin(){
 
     @Override
     public <T extends Concept> T buildConcept(Vertex vertex) {
-        return elementFactory.buildConcept(vertex);
+        return getElementFactory().buildConcept(vertex);
```
fppt: Initially the concept cache was in element factory and element factory was made thread local. This is just a left over of those few attempts. Is it that much of an issue that it needs to be changed ? I typically prefer using the getter rather than the field directly. `getTinkerTraversals()` shows why it is usually a good idea.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -656,7 +657,6 @@ protected void clearGraph(){
      */
     @Override
     public void close() {
-        getConceptLog().clearTransaction();
```
fppt: It's been moved to `clearLocalVariables` method

grakn-graph/src/main/java/ai/grakn/graph/internal/Cache.java
fppt: > and will be constantly recalculated 

Not so. Not with the current use cases at least. This is really to account for when you initially try to set the super type. First time around the database reader `getOutgoingNeighbour(Schema.EdgeLabel.SUB)` is null no matter what.

grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptLog.java
fppt: Indeed. Low priority improvement though. These caches are transaction specific and will never get too large.

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationTypeImpl.java
fppt: That is okay in my opinion. If it is too magic then we clone the set.

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationTypeImpl.java
fppt: It kind of has to be done that way I think. If you don't update the cache fully with all the roles then you simply adding the new role to the cache and that will result in only that role ever being returned when calling `hasRoles()`. This could be improved by tracking internally in `Cache` if the db has been read before but because cache is monitoring Sets and single elements the implementation will not be so simple.

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationTypeImpl.java
fppt: We may need seperate Cache implementations then. This is because the Cache alss just accounts for `X` at the moment and we using it for both ` single element `X` and `Set<X>`

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationTypeImpl.java
fppt: yes because if the cache is not loaded then you will try to read from the database and traverse across the vertex AFTER it has been deleted. This will result in a good old null pointer.

grakn-graph/src/main/java/ai/grakn/graph/internal/RoleTypeImpl.java
fppt: Nope. This check: `boolean hasHasRoles = getVertex().edges(Direction.IN, Schema.EdgeLabel.HAS_ROLE.getLabel()).hasNext();` before hand means that you won't have any relation types to notify.

grakn-graph/src/main/java/ai/grakn/graph/internal/RoleTypeImpl.java
fppt: That may indeed be better.

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
fppt: Nope again. You won't have any subtypes if you get to this stage. This check: `boolean hasSubs = getVertex().edges(Direction.IN, Schema.EdgeLabel.SUB.getLabel()).hasNext();` will make sure of that.

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
fppt: I think you correct on this one.

grakn-graph/src/test/java/ai/grakn/graph/internal/EntityTypeTest.java
fppt: Will make that change.

grakn-test/src/test/java/ai/grakn/test/graql/query/MatchQueryTest.java
fppt: You do. `graphAPIPlaysRoles ` gets mutated directly later on.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
aelred: Nah, this wasn't a request for it to change. Just wondering if there was a reason for it.

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationTypeImpl.java
aelred: In some cases we're using `isPresent` before doing this, though. Is there a rule for when we do that check and when we don't?

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationTypeImpl.java
aelred: Functional programming may be able to save us, with a `Cache<X>#modify(Consumer<X> method)` that only calls the method if the cache is present.

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationTypeImpl.java
fppt: We only call `isPresent` when deleting. That is because when deleting the cache only needs to be updated if if has been loaded. But when we adding something either the cache should not be updated at all (which might be a better approach) or it has to be fully updated so that we can be sure the cache contains everything.

grakn-graph/src/test/java/ai/grakn/graph/internal/EntityTypeTest.java
```diff
@@ -592,27 +594,21 @@ public void checkSubTypeCachingUpdatedCorrectlyWhenChangingSuperTypes(){
         EntityType e6 = graknGraph.putEntityType("entityType6").superType(e5);
 
         assertEquals(4, e1.subTypes().size());
-        assertTrue("e1 subtypes does not contain e1", e1.subTypes().contains(e1));
-        assertTrue("e1 subtypes does not contain e2", e1.subTypes().contains(e2));
-        assertTrue("e1 subtypes does not contain e3", e1.subTypes().contains(e3));
-        assertTrue("e1 subtypes does not contain e4", e1.subTypes().contains(e4));
+        assertThat(e1.subTypes(), containsInAnyOrder(e1, e2, e3, e4));
```
aelred: Yay! I don't care if you do this or not, but I thought you should know `containsInAnyOrder` will also confirm the size is correct.

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
fppt: I think you are correct here. In general the cache should only be updated if it has been loaded.

PR #1076, opened 2017-02-06
New '_name' field in HAL representation

marco-scoppetta:
- HAL representation of Types now have '_name' field with type name
  - '_id' HAL field now always contains the concept ID


grakn-graql/src/main/java/ai/grakn/graql/internal/hal/HALConceptRepresentationBuilder.java
```diff
@@ -238,12 +240,15 @@ static void generateConceptState(Representation resource, Concept concept){
                     .withProperty(BASETYPE_PROPERTY, getBaseType(instance).name());
         } else {
             Type type = concept.asType();
-            resource.withProperty(ID_PROPERTY, type.getName().getValue())
+            resource.withProperty(ID_PROPERTY, type.getId().getValue())
```
fppt: You set the `ID_PROPERTY ` in the if(...) and the else(....). Could clean that up.

grakn-graql/src/main/java/ai/grakn/graql/internal/hal/HALConceptRepresentationBuilder.java
```diff
@@ -233,17 +233,16 @@ private static void attachSingleGeneratedRelation(Representation currentHal, Con
...
-            resource.withProperty(ID_PROPERTY, instance.getId().getValue())
-                    .withProperty(TYPE_PROPERTY, instance.type().getName().getValue())
+            resource.withProperty(TYPE_PROPERTY, instance.type().getName().getValue())
                     .withProperty(BASETYPE_PROPERTY, getBaseType(instance).name());
         } else {
-            Type type = concept.asType();
-            resource.withProperty(ID_PROPERTY, type.getId().getValue())
-                    .withProperty(BASETYPE_PROPERTY, getBaseType(type).name());
-
+            resource.withProperty(BASETYPE_PROPERTY, getBaseType(concept.asType()).name());
```
fppt: Sorry to do this but I just noticed you setting the base type in the if and else as well. I.e. the else block seems completely unneeded.

marco-scoppetta: it's the same method but it's overloaded, so I guess we need the else block? we are performing different checks in the two `getBaseType` methods

PR #1077, opened 2017-02-06
Scheduler initially persists tasks

alexandraorth:
+ The `Scheduler` stores and marks tasks as `CREATED` 
+ This removes any blocking on REST endpoint
+ Kafka passing around Task ID as key and Serialized `TaskState` as value
+ Cleanup in Engine shutdown methods

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/backgroundtasks/distributed/DistributedTaskManager.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/backgroundtasks/distributed/DistributedTaskManager.java
@@ -95,9 +95,7 @@ public final class DistributedTaskManager implements TaskManager {
                 .interval(period)
                 .configuration(configuration);
 
-        stateStorage.newState(taskState);
-
-        producer.send(new ProducerRecord<>(NEW_TASKS_TOPIC, taskState.getId(), configuration.toString()));
+        producer.send(new ProducerRecord<>(NEW_TASKS_TOPIC, taskState.getId(), TaskState.serialize(taskState)));
         producer.flush();
 
         return taskState.getId();
--- a/grakn-engine/src/main/java/ai/grakn/engine/backgroundtasks/distributed/Scheduler.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/backgroundtasks/distributed/Scheduler.java
@@ -20,7 +20,6 @@ package ai.grakn.engine.backgroundtasks.distributed;
 
 import ai.grakn.engine.backgroundtasks.TaskStateStorage;
 import ai.grakn.engine.backgroundtasks.TaskState;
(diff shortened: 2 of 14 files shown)


grakn-engine/src/main/java/ai/grakn/engine/backgroundtasks/distributed/DistributedTaskManager.java
```diff
@@ -95,9 +95,7 @@ public String scheduleTask(BackgroundTask task, String createdBy, Instant runAt,
                 .interval(period)
                 .configuration(configuration);
 
-        stateStorage.newState(taskState);
-
-        producer.send(new ProducerRecord<>(NEW_TASKS_TOPIC, taskState.getId(), configuration.toString()));
+        producer.send(new ProducerRecord<>(NEW_TASKS_TOPIC, taskState.getId(), TaskState.serialize(taskState)));
```
bolerio: If this method is about inserting new tasks, it should be renamed. In our context "schedule task" has a specific so this name leads to confusion.

grakn-engine/src/main/java/ai/grakn/engine/backgroundtasks/distributed/Scheduler.java
```diff
@@ -157,33 +161,21 @@ public void close() {
...
-    private void scheduleTask(String id,  String configuration, TaskState state) {
+    private void scheduleTask(TaskState state) {
         long delay = Duration.between(Instant.now(), state.runAt()).toMillis();
 
         markAsScheduled(state);
         if(state.isRecurring()) {
             Runnable submit = () -> {
                 markAsScheduled(state);
-                sendToWorkQueue(id, configuration);
+                sendToWorkQueue(state);
```
bolerio: Why do we have another markAsScheduled here? In addition to the one at line 169

bolerio: Also, shouldn't "markAsScheduled" after we actually schedule it? I.e. after it is sent to the work queue?

alexandraorth: Because this task is recurring, it should be marked as scheduled every time it is run.

alexandraorth: And re: second question. 
If we first mark it as scheduled first, its possible we could crash before sending it to the work queue. This means it will remain in the new task queue and be picked up on the next scheduler restart.
If we mark it as scheduled after submitting, its possible the task would be picked up by the `TaskRunner` before being marked as `SCHEDULED`. At the moment, the `TaskRunner` will not run any tasks it receives that are not marked as `SCHEDULED`, so that task would never be run.

PR #1079, opened 2017-02-06
Loader task using a `EngineGraknGraph`

alexandraorth:
+ The LoaderTask should use a `EngineGraknGraph` because it will always be executed from within Engine

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/loader/LoaderTask.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/loader/LoaderTask.java
@@ -18,11 +18,12 @@
 
 package ai.grakn.engine.loader;
 
-import ai.grakn.Grakn;
-import ai.grakn.GraknGraph;
 import ai.grakn.engine.backgroundtasks.BackgroundTask;
+import ai.grakn.engine.postprocessing.EngineCacheImpl;
 import ai.grakn.engine.util.ConfigProperties;
 import ai.grakn.exception.GraknValidationException;
+import ai.grakn.factory.EngineGraknGraphFactory;
+import ai.grakn.graph.EngineGraknGraph;
 import ai.grakn.graql.Graql;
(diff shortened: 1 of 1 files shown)


grakn-engine/src/main/java/ai/grakn/engine/loader/LoaderTask.java
```diff
@@ -98,16 +100,27 @@ private void attemptInsertions(String uri, String keyspace, Collection<InsertQue
...
-    private boolean insertQueriesInOneTransaction(GraknGraph graph, Collection<InsertQuery> inserts) {
+    private boolean insertQueriesInOneTransaction(EngineGraknGraph graph, Collection<InsertQuery> inserts) {
 
         try {
             graph.showImplicitConcepts(true);
 
             // execute each of the insert queries
-            inserts.forEach(q -> q.withGraph(graph).execute());
+            // TODO This is a very round-about way of inserting. We cannot execute InsertQuery.withGraph() because
+            // TODO that function does not accept an EngineGraknGraph object. We need the IF to execute a
```
fppt: This is not ideal but I would prefer an ugly casting to `GraknGraph` rather than this.

alexandraorth: A casting might not work because `InsertQuery` accepts a `GraknGraph` object, not a `BaseGraknGraph` object

fppt: The underlaying object in all three cases is still a `AbstractGraknGraph` so it may work.

alexandraorth: So much simpler! I'll push the change.

fppt: It's still not ideal but is a bit better :/

PR #1089, opened 2017-02-07
Add standard deviation aggregate

aelred:


grakn-graql/src/main/java/ai/grakn/graql/internal/query/aggregate/StdAggregate.java
```diff
@@ -0,0 +1,88 @@
...
+
+        OptionalDouble variance = numbers.stream().mapToDouble(x -> pow(x - average, 2)).average();
+
+        if (!variance.isPresent()) {
+            return Optional.empty();
+        } else {
+            return Optional.of(sqrt(variance.getAsDouble()));
+        }
+    }
+
```
bolerio: I think it's worthwhile optimizing this algorithm. There is a lot of overhead to iterate over a bunch of numbers.We should accumulating the numbers in a list and do it in a single pass. Take a look here:

https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance

There is an online version of computing variance and there is a single pass "naive" version. 

Also, shouldn't pow(x, 2) can be written as x*x?

aelred: Thanks, I've implemented the online version instead.

The results came out differently because that algorithm calculates the "unbiased sample" standard deviation. I'm unclear on the difference, so I've just clarified in the JavaDoc that this is the thing we're calculating.

aelred:
please retest this please

PR #1094, opened 2017-02-08
Get rid of empty job sets in Post Processing

fppt:


grakn-engine/src/main/java/ai/grakn/engine/postprocessing/ConceptFixer.java
```diff
@@ -95,7 +93,8 @@ private static boolean runPostProcessingJob(BiFunction<EngineGraknGraph, Set<Con
                 //Finally clear the cache
                 jobFinaliser.accept(EngineCacheImpl.getInstance(), conceptIndex);
 
-                return true; //If it can get here. All post processing has succeeded.
+                return; //If it can get here. All post processing has succeeded.
```
aelred: You don't strictly need the `return` now.

PR #1100, opened 2017-02-09
Unique '_id' for HAL generated relations

marco-scoppetta:
- We used to use "temp-assertion", now we append role players IDs to make it unique


grakn-graql/src/main/java/ai/grakn/graql/internal/hal/HALConceptRepresentationBuilder.java
```diff
@@ -136,26 +136,26 @@ static void attachGeneratedRelations(Representation currentHal, Map.Entry<VarNam
     private static void attachSingleGeneratedRelation(Representation currentHal, Concept currentVar, Concept otherVar, Map<VarName, String> roleTypes, VarName currentVarName, VarName otherVarName, Optional<TypeName> relationType, String keyspace) {
         ConceptId currentID = currentVar.getId();
 
-        ConceptId firstID;
-        ConceptId secondID;
+        String firstID;
```
fppt: What was the reason for changing to the `String` ? 

This change has result in `getValue` being called more often than it needs to be.

grakn-graql/src/main/java/ai/grakn/graql/internal/hal/HALGeneratedRelation.java
```diff
@@ -43,9 +43,9 @@
         this.factory = new StandardRepresentationFactory();
     }
 
-    Representation getNewGeneratedRelation(String assertionID, Optional<TypeName> relationType) {
+    Representation getNewGeneratedRelation(String firstID, String secondID, String assertionID, Optional<TypeName> relationType) {
```
fppt: In my opinion these IDs should be `ConceptID`s and you should only be calling `getValue()` in here because this is really the only place you *need* the `String` representation.

grakn-graql/src/main/java/ai/grakn/graql/internal/hal/HALConceptRepresentationBuilder.java
```diff
@@ -136,26 +136,26 @@ static void attachGeneratedRelations(Representation currentHal, Map.Entry<VarNam
...
             secondRole = (roleTypes.get(currentVarName).equals(HAS_ROLE_EDGE)) ? "" : roleTypes.get(currentVarName) + ":";
             firstRole = (roleTypes.get(otherVarName).equals(HAS_ROLE_EDGE)) ? "" : roleTypes.get(otherVarName) + ":";
         }
 
-        String isaString = (relationType.isPresent()) ? "isa " + relationType.map(StringConverter::typeNameToString) : "";
-        String assertionID = String.format(ASSERTION_URL, keyspace, firstID, secondID, firstRole, secondRole, isaString);
-        currentHal.withRepresentation(roleTypes.get(currentVarName), new HALGeneratedRelation().getNewGeneratedRelation(firstID, secondID, assertionID, relationType));
+        String isaString = (relationType.isPresent()) ? "isa " + StringConverter.typeNameToString(relationType.get()) : "";
+        String assertionID = String.format(ASSERTION_URL, keyspace, firstID, secondID, firstRole, secondRole,isaString);
+        currentHal.withRepresentation(roleTypes.get(currentVarName), new HALGeneratedRelation().getNewGeneratedRelation(firstID.getValue(),secondID.getValue(),assertionID, relationType));
```
fppt: Is there any reason `getNewGeneratedRelation ` cannot take in the `ConceptID` as well?

marco-scoppetta:
I think I have made the changes you asked for

PR #1114, opened 2017-02-10
Quote bash test variables

burukuru:


burukuru:
This is actually passing, for some reason one of the test states on Github can't be set properly.

aelred:
@burukuru sorry, I never saw that this was passing! Can you fix the conflicts? Then I'll merge it.

burukuru:
I'll fix and re-submit with a different name or something so it stops doing that.

PR #1119, opened 2017-02-13
Support `toString` of deleted concepts

aelred:


grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptImpl.java
```diff
@@ -617,7 +617,16 @@ public boolean equals(Object object) {
     }
 
     @Override
-    public String toString(){
+    public final String toString(){
+        if (vertex == null) {
```
fppt: Instead of `vertex == null` you should use `!getGraknGraph().validVertex(vertex)` that method checks (using Titan) if the vertex has been removed and is more reliable.

fppt:
please retest this please

PR #1120, opened 2017-02-13
Fix and new behaviour dashboard

marco-scoppetta:
- fixed bug that fires double request when hitting Enter
  - now also Analytics is usable again thanks to the above fix
  - Alt+doubleclick fires ontology requests like Alt+single click


marco-scoppetta:
please retest this please

PR #1121, opened 2017-02-13
Fill in rule sets on commit

fppt:


fppt:
@bolerio This is not needed for the next release. It's a nice to have feature more than anything else. Safe to go on though.

PR #1124, opened 2017-02-13
Add error messages when trying to attach LHS and RHS to non-rules

aelred:


aelred:
@fppt agreed. PMD was (rightly) telling me off because the classes were basically identical. I wanted to get rid of them entirely. Unfortunately, there are a few places where we explicit ask for "the LHS property" and "the RHS property" by class name, so they couldn't be completely eliminated.

PR #1128, opened 2017-02-14
Get the resource types which are linked to entity types.

fppt:


fppt:
@haikalpribadi this is the feature you asked me for when I was on the plane.

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
```diff
@@ -142,6 +142,25 @@ V addInstance(Schema.BaseType instanceBaseType, BiFunction<Vertex, T, V> produce
...
+        playsRoles().forEach(roleType -> roleType.relationTypes().forEach(relationType -> {
+            if(relationType.isImplicit()){
+                //This is faster than doing the traversal
+                TypeName prefix = Schema.Resource.HAS_RESOURCE.getName(TypeName.of(""));
+                TypeName resourceTypeName = TypeName.of(relationType.getName().getValue().replace(prefix.getValue(), ""));
+                resourceTypes.add(getGraknGraph().getType(resourceTypeName));
+            }
+        }));
+
+        getGraknGraph().showImplicitConcepts(false);
```
aelred: There's a bug here: if the flag was on when entering this method, then it will be off when leaving.
You can add a mark to your "implicit flag" tally!

fppt: I have fixed the bug. Damn implicit flag!

aelred:
please retest this please

PR #1129, opened 2017-02-14
Update cache time whenever cache itself is updated. Fixes bug where PP is always running.

fppt:


fppt:
@alexandraorth this should stop PP from always running

PR #1130, opened 2017-02-14
Possible to clear graph without hitting engine rest API. This is mostly for engine.

fppt:


fppt:
@alexandraorth You can use this method to avoid the REST API when clearing the graph

fppt:
please retest this please

PR #1131, opened 2017-02-14
Expand Data Types to include Integer and float

fppt:


fppt:
@aelred One of your tests is failing on this because it is generating a large number and trying to use the new INTEGER data type. Could you take a look at that please?

aelred:
@fppt I think I've fixed the issue now.

fppt:
Thanks for fixing the generator on this one @aelred

PR #1132, opened 2017-02-14
Fix some of Felix's tests

fppt:


fppt:
@aelred Look at the first commit message please.

PR #1133, opened 2017-02-14
HAL ontology builder fix and resources on rules

marco-scoppetta:
- solved an outstanding TODO in HAL ontology builder
  - added resources to rules instances when building HAL


grakn-graql/src/main/java/ai/grakn/graql/internal/hal/HALConceptData.java
```diff
@@ -195,6 +194,10 @@ private void generateStateAndLinks(Representation resource, Concept concept) {
             generateResources(resource, concept.asEntity().resources());
         }
 
+        if (concept.isRule()) {
+            generateResources(resource, concept.asRule().resources());
+        }
+
         if (concept.isRelation()) {
             generateResources(resource, concept.asRelation().resources());
         }
```
aelred: You can get rid of these three `if`s:

```
if (concept.isInstance()) {
    generateResources(resource, concept.asInstance().resources());
}
```

PR #1134, opened 2017-02-14
optimisations for lazy resolution

kasper-piskorski:


grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/cache/LazyQueryCache.java
```diff
@@ -88,6 +96,16 @@
...
     }
 
+    public void reload(){
+        Map<T, Pair<T, LazyAnswerIterator>> newCache = new HashMap<>();
+        cache.entrySet().forEach(entry ->
+                newCache.put(entry.getKey(), new Pair<>(
+                        entry.getKey(),
+                        new LazyAnswerIterator(entry.getValue().getValue().accumulator.stream()))));
+        cache.clear();
+        cache.putAll(newCache);
```
aelred: There's a neat Guava method `Maps.transformValues` that will do this.

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/iterator/LazyAnswerIterator.java
```diff
@@ -0,0 +1,54 @@
...
+        Iterator<Map<VarName, Concept>> transform = Iterators.transform(iterator(), input -> QueryAnswers.unify(input, unifiers));
+        return new LazyAnswerIterator(transform);
+    }
+
+    public LazyAnswerIterator merge (Stream<Map<VarName, Concept>> stream){
+        return new LazyAnswerIterator(Stream.concat(this.stream(), stream));
+    }
+    public LazyAnswerIterator merge (LazyAnswerIterator iter) {
+        return new LazyAnswerIterator(Stream.concat(this.stream(), iter.stream()));
+    }
```
aelred: For all these methods, it might be best to not create new lazy iterators: otherwise you are caching the results twice: once in this lazy iterator, and once again in the new lazy iterator you create.

PR #1137, opened 2017-02-14
Add logback as a test dependency

aelred:
This guarantees that tests always have access to a logger, without mandating the logger implementation that users of our library must use.


aelred:
@fppt you can review this as a post-mortem, but I'm going to merge now because it is causing peoples' tests to fail.

PR #1139, opened 2017-02-14
Improved selection function.

kasper-piskorski:


grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerQueryImpl.java
```diff
@@ -365,6 +365,26 @@ public void addAtomConstraints(Set<? extends Atomic> cstrs){
...
+        }
+        return atom != null? atom : atoms.iterator().next();
+    }
+
+    private Atom findNextJoinable(Set<Atom> atoms, Set<VarName> vars){
+        Iterator<Atom> ait = atoms.iterator();
+        Atom atom = null;
+        while(ait.hasNext() && atom == null) {
+            Atom next = ait.next();
+            atom = Sets.intersection(vars, next.getVarNames()).isEmpty()? null : next;
```
aelred:
```
while(ait.hasNext()) {
     Atom next = ait.next();
     if (Sets.intersection(vars, next.getVarNames()).isEmpty()) return next;
}
```
maybe?

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerQueryImpl.java
```diff
@@ -365,6 +365,26 @@ public void addAtomConstraints(Set<? extends Atomic> cstrs){
...
 
+    private Atom findFirstJoinable(Set<Atom> atoms){
+        Iterator<Atom> ait = atoms.iterator();
+        Atom atom = null;
+        while(ait.hasNext() && atom == null) {
+            Atom next = ait.next();
+            atom = findNextJoinable(Sets.difference(atoms, Sets.newHashSet(next)), next.getVarNames());
+        }
+        return atom != null? atom : atoms.iterator().next();
+    }
```
aelred: Same as below

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerQueryImpl.java
```diff
@@ -376,15 +396,26 @@ public void addAtomConstraints(Set<? extends Atomic> cstrs){
...
 
-        //order by variables
         Set<Atom> orderedSelection = new LinkedHashSet<>();
-        getVarNames().forEach(var -> orderedSelection.addAll(selectedAtoms.stream()
-                .filter(atom -> atom.containsVar(var))
-                .collect(Collectors.toSet())));
+
+        Atom atom = findFirstJoinable(atomsToSelect);
+        orderedSelection.add(atom);
+        atomsToSelect.remove(atom);
```
aelred: You can remove these two lines (they are repeated in the loop), if you move the line:
```
atom = findNextJoinable(atomsToSelect, joinedVars);
```
to after the conditional.

PR #1140, opened 2017-02-14
'Graceful' shutdown of Engine

alexandraorth:
+We should call the Engine stop method when receiving a kill signal

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
@@ -45,6 +45,7 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


alexandraorth:
Just to note: travis failure is integration test excessive logging bug

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -171,9 +177,13 @@ public static void stopHTTP() {
...
     }
 
-    private static void stopTaskManager() throws Exception {
+    private static void stopTaskManager() {
         PostProcessing.getInstance().stop();
-        taskManager.close();
+        try {
+            taskManager.close();
+        } catch (Exception e){
+            LOG.error(getFullStackTrace(e));
```
aelred: Should this just log if it goes wrong, or re-throw a runtime exception?

PR #1143, opened 2017-02-15
Killing more felix tests

fppt:


grakn-graph/src/test/java/ai/grakn/graph/internal/GraknGraphPropertyIT.java
```diff
@@ -617,33 +625,10 @@ public void whenCallingClear_TheGraphCloses(@Open GraknGraph graph) {
...
-    }
-
-    @Ignore // TODO: Fix this
-    @Property
-    public void whenCallingClear_AllMetaConceptsArePresent(
-            @Open GraknGraph graph, @From(MetaTypeNames.class) TypeName typeName) {
-        graph.clear();
-        graph.open();
-        assertNotNull(graph.getType(typeName));
-    }
```
aelred: Instead of deleting these, can we leave them as ignored?

Although they will have to change with the new transaction behaviour, the behaviour they test for is still expected.

grakn-graph/src/test/java/ai/grakn/graph/internal/GraknGraphPropertyIT.java
```diff
@@ -617,33 +625,10 @@ public void whenCallingClear_TheGraphCloses(@Open GraknGraph graph) {
...
-
-    @Ignore // TODO: Fix this, or remove the test
     @Property
     public void whenCallingGetKeySpace_ReturnTheKeyspaceOfTheGraph(String keyspace) {
```
aelred: Maybe we should change the test name to reflect this: `ReturnTheLowercaseKeyspaceOfTheGraph`

grakn-graph/src/test/java/ai/grakn/graph/internal/GraknGraphPropertyIT.java
```diff
@@ -581,11 +578,22 @@ public void whenCallingGetRuleType_TheResultIsTheSameAsGetType(@Open GraknGraph
...
+        // comparison is more defined.
+
+        Relation foundRelation = graph.getRelation(relation.type(), relation.rolePlayers());
+        if(foundRelation.getId().equals(relation.getId())){
+            assertEquals(relation, foundRelation);
+        } else { //This is possible when we have created duplicate empty relations. So we check everything we can.
+            assertThat(relation.rolePlayers().keySet(), containsInAnyOrder(foundRelation.rolePlayers().keySet()));
+            assertThat(relation.rolePlayers().values(), containsInAnyOrder(foundRelation.rolePlayers().values()));
+            assertEquals(relation.type(), foundRelation.type());
+        }
```
aelred: I might look into a way to split this into two tests, since it feels like two cases.

Alternative more radical idea: this odd behaviour might be a hint that we shouldn't have this method at all, or that it should return all matching relations.

fppt: I think we can remove this method from the API. Internally it is still needed for PP but we can kill it from the API. No one is using it. Literally.

fppt:
please retest this please

PR #1152, opened 2017-02-15
Deep clone cache to avoid transactions from affecting central cache.

fppt:
The need for this feature is best exhibited by the test attached in this PR.
It's to avoid a transaction from affecting the centrally cached ontology. The only way to do this is to deep clone the cached ontology at the start of a transaction.

Please review this PR very carefully. As with other caching PRs this has the potential to cause a lot of harm.

### DIFF
--- a/grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
+++ b/grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
@@ -105,7 +105,7 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
     private final ThreadLocal<Boolean> localIsOpen = new ThreadLocal<>();
     private final ThreadLocal<String> localClosedReason = new ThreadLocal<>();
     private final ThreadLocal<Boolean> localShowImplicitStructures = new ThreadLocal<>();
-    private final ThreadLocal<Map<TypeName, Type>> localCloneCache = new ThreadLocal<>();
+    private final ThreadLocal<Map<TypeName, TypeImpl>> localCloneCache = new ThreadLocal<>();
 
     private boolean committed; //Shared between multiple threads so we know if a refresh must be performed
 
@@ -297,8 +297,8 @@ public abstract class AbstractGraknGraph<G extends Graph> implements GraknGraph,
         return conceptLog;
     }
 
-    private Map<TypeName, Type> getCloneCache(){
-        Map<TypeName, Type> cloneCache = localCloneCache.get();
+    private Map<TypeName, TypeImpl> getCloneCache(){
+        Map<TypeName, TypeImpl> cloneCache = localCloneCache.get();
         if(cloneCache == null){
             localCloneCache.set(cloneCache = new HashMap<>());
         }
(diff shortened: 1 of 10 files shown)


fppt:
please retest this please

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -285,15 +288,105 @@ private EdgeImpl addEdge(Concept from, Concept to, Schema.EdgeLabel type){
...
+        for (TypeImpl<?, ?> type : getCloneCache().values()) {
+            TypeImpl centralType = cachedOntologySnapshot.get(type.getName());
+
+            if(type.isRelationType()){
+                ((RelationTypeImpl) type).copyCachedConcepts((RelationTypeImpl) centralType);
+            } else if(type.isRoleType()){
+                ((RoleTypeImpl) type).copyCachedConcepts((RoleTypeImpl) centralType);
+            } else {
+                type.copyCachedConcepts(centralType);
+            }
```
aelred: Why can't we just call `type.copyCachedConcepts`?

fppt: We need to pass in the type to be copy the sets from and the reason we need the casting is because it will default to the `TypeImpl` copier even if it is a `RelationTypeImpl`. I hate this chunk of code as well but by not casting the wrong copier is called.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -285,15 +288,105 @@ private EdgeImpl addEdge(Concept from, Concept to, Schema.EdgeLabel type){
...
+            clonedType = new RelationTypeImpl((RelationTypeImpl) type);
+        } else if(type.isEntityType()) {
+            clonedType = new EntityTypeImpl((EntityTypeImpl) type);
+        } else if(type.isRuleType()) {
+            clonedType = new RuleTypeImpl((RuleTypeImpl) type);
+        } else if(type.isResourceType()) {
+            clonedType = new ResourceTypeImpl((ResourceTypeImpl) type);
+        } else if(type.isType()) {
+            clonedType = new TypeImpl((TypeImpl) type);
+        }
```
fppt: Ahhh I agree. I can make that change.

fppt: Since I am travelling would you mind letting me do this in another PR since its more of a code cleanup than a bug or issue? 

I want to get this is quickly so we can find issues with it faster.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
aelred: For completeness' sake, here's why we can't use the `TypeImpl` generic `T` as the parameter: "because the cache stores everything as a *Impl you still need to do some odd casting which results in methods being bypassed again."

I would suggest that we just have a method that accepts `TypeImpl`, overridden by `RelationTypeImpl` and `RoleTypeImpl`, which will handle their logic (as well as the dirty casts necessary).

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
aelred: Yeah, that's fine.

PR #1154, opened 2017-02-16
New transaction handling

fppt:


fppt:
@aelred There are some graql shell tests I have not been able to fix. Could you take a look at those please?

fppt:
They failing locally so you can just run the tests and see the issue.

fppt:
please retest this please

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -136,6 +137,13 @@ public AbstractGraknGraph(G graph, String keyspace, String engine, boolean batch
     }
 
     /**
+     * Opens the thread bound transaction
+     */
+    public void openTransaction(){
+        localIsOpen.set(true);
+    }
```
aelred: Does this do the same as what `open` used to do? I still find this behaviour a little confusing, but the name is better.

grakn-graph/src/test/java/ai/grakn/generator/GraknGraphs.java
```diff
@@ -106,9 +107,12 @@ public GraknGraph generate() {
...
 
         if (shouldOpen) {
-            graph.open();
         } else {
-            graph.close();
+            try {
+                graph.close();
+            } catch (GraknValidationException e) {
+                throw new RuntimeException("Generater opened invalid graph somehow", e);
+            }
```
aelred: Should invert the condition instead of an empty `if`. Also the `try...catch` is unnecessary now that `GraknValidationException` is runtime.

grakn-graph/src/test/java/ai/grakn/graph/internal/GraknTinkerGraphTest.java
```diff
@@ -80,8 +79,7 @@ public void testTestThreadLocal(){
 
         for(int i = 0; i < 100; i ++){
             futures.add(pool.submit(() -> {
-                GraknGraph innerTranscation = graknGraph;
-                innerTranscation.open();
+                GraknGraph innerTranscation = Grakn.factory(Grakn.IN_MEMORY, graknGraph.getKeyspace()).getGraph();
```
aelred: So it's no longer necessary to open graphs in a different thread?

grakn-test/src/test/java/ai/grakn/test/graql/query/MatchQueryTest.java
```diff
@@ -161,8 +160,8 @@
 
     private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("EEE MMM dd HH:mm:ss zzz yyyy", Locale.US);
 
-    @ClassRule
```
aelred: We should try and keep this a `@ClassRule` for performance reasons if possible.

grakn-test/src/test/java/ai/grakn/test/graql/query/MatchQueryTest.java
```diff
@@ -175,6 +174,7 @@ public void setUp() {
     @After
     public void tearDown() {
         if (movieGraph.graph() != null) movieGraph.graph().showImplicitConcepts(false);
+//        movieGraph.clearGraph();
```
aelred: Whoops!

grakn-test/src/test/java/ai/grakn/test/graql/shell/GraqlShellIT.java
```diff
@@ -296,6 +296,7 @@ public void testComputeCount() throws Exception {
         assertThat(result, containsString("\n3\n"));
     }
 
+    @Ignore //TODO: Fix rollback function in graql because the graph no longer has a rollback function
```
aelred: Sorry, I missed these when I fixed the shell tests earlier!

grakn-graph/src/main/java/ai/grakn/factory/AbstractInternalFactory.java
```diff
@@ -91,18 +91,21 @@ public synchronized M getGraph(boolean batchLoading){
...
             graknGraph = getGraph(graknGraph, false);
             lastGraphBuiltBatchLoading = false;
+            graknGraph.openTransaction();
             return graknGraph;
         }
     }
     protected M getGraph(M graknGraph, boolean batchLoading){
+        boolean hasCommitted = false;
         //This checks if the previous graph built with this factory is the same as the one we trying to build now.
+        //TODO: Check is this features is still needed. It's very confusing at the end of the day
```
fppt: Basically this weird thing is to force a refresh when someone commits with the batch graph. If we don't force a refresh then the normal graph is out of date. We originally added this because analytics was constantly switching between the batch and normal graph. The question is, do we still need it? @duckofyork in your opinion?

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
fppt: Pretty much yes, but it is not exposed. The purpose is to just maintain a thread bound flag so that we know when to throw an exception is the graph is passed to a different thread.

grakn-graph/src/test/java/ai/grakn/generator/GraknGraphs.java
fppt: True

grakn-graph/src/test/java/ai/grakn/graph/internal/GraknTinkerGraphTest.java
fppt: You have to open graphs in different threads but you do so via the factory. So now you don't pass a graph into a different thread. You pass the factory.

grakn-test/src/test/java/ai/grakn/test/graql/query/MatchQueryTest.java
fppt: Agreed. Can't recall why but I had to change it. Should revisit that actually

grakn-test/src/test/java/ai/grakn/test/graql/query/MatchQueryTest.java
fppt: oops indeed

grakn-graph/src/main/java/ai/grakn/factory/AbstractInternalFactory.java
duckofyork: Analytics no longer has persistence. So we don't use the batch graph any more.

fppt: Great. Then lets try get rid of this magic code. Thanks man.

PR #1157, opened 2017-02-16
Cleanup deep cloning mechanisim

fppt:
- Introduce copy method rather than use constructors directly
- Cleanup the process of finalising the cloning


grakn-graph/src/main/java/ai/grakn/graph/internal/ResourceTypeImpl.java
```diff
@@ -59,10 +59,17 @@
         setImmutableProperty(Schema.ConceptProperty.IS_UNIQUE, isUnique, getProperty(Schema.ConceptProperty.IS_UNIQUE), Function.identity());
     }
 
-    ResourceTypeImpl(ResourceTypeImpl resourceType){
+    private ResourceTypeImpl(ResourceTypeImpl resourceType){
+        //noinspection unchecked
```
aelred: Can you put generics here to avoid suppressing the unchecked warnings?
```
private ResourceTypeImpl<D>(ResourceTypeImpl<D> resourceType) {
```

PR #1161, opened 2017-02-16
Optimised conceptid hashing

kasper-piskorski:


grakn-core/src/main/java/ai/grakn/concept/ConceptId.java
```diff
@@ -59,15 +60,22 @@ public Object getRawValue(){
     public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
-
-        ConceptId conceptId1 = (ConceptId) o;
-
-        return getValue().equals(conceptId1.getValue());
+        ConceptId cast = (ConceptId) o;
+        if (conceptId instanceof String) {
```
fppt: I don't understand why you need this `if else`. For a given graph (for example a titan graph) the concept Ids will always be strings. Whats the need for this ? Why not just toString both and compare? Is the toString that expensive?

kasper-piskorski: If `conceptId` is graph specific, then indeed the conditional should be changed. And yes, the idea was to avoid doing the toString conversion on non-string types. The comparison on concept ids is called in my inner join and these can easily go in thousands, so added:
a) caching the hashcode
b ) avoiding non-string to string conversion and instead doing a direct comparison

bolerio: The concept ID is passed around as a string to other layers which leads to the concept ID sometimes being a string and sometimes being whatever the native ID type is (long for tinkerpop for example) so we end up comparing strings to longs with .equals and it fails. And instanceof check + long comparison is quicker than creating a string object, copying a long into it and then comparing the longs. A UUID is made up of two longs, so ideally it should also be compared as such rather than a string. We got 50% performance boost out of that optimization!

PR #1164, opened 2017-02-21
Remove auto closing of graphs when a commit occurs on a batch graph

fppt:


fppt:
please retest this please

PR #1169, opened 2017-02-23
Close the main connection to the graph by closing on the factory.

fppt:
- Monitor transactions counts in factory
- Add explicit close to factory which must be used at the end of using the graph
- Get rid of magic auto closing on TItan

@alexandraorth @aelred @duckofyork @mikonapoli @sheldonkhall When this goes in guys you will have to call `factory.close()` or your app will NEVER terminate.

### DIFF
--- a/grakn-core/src/main/java/ai/grakn/GraknGraphFactory.java
+++ b/grakn-core/src/main/java/ai/grakn/GraknGraphFactory.java
@@ -19,6 +19,8 @@
 package ai.grakn;
 
 
+import ai.grakn.exception.GraphRuntimeException;
+
 /**
  * <p>
  *     Builds a Grakn Graph factory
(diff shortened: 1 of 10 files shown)


grakn-core/src/main/java/ai/grakn/GraknGraphFactory.java
```diff
@@ -58,4 +60,23 @@
      * @see GraknComputer
      */
     GraknComputer getGraphComputer();
+
+    /**
+     * Closes the main connection to the graph. This should be done at the end of using the graph.
+     *
+     * @throws GraphRuntimeException when more than 1 transaction is open on the graph
```
aelred: is this really "more than one" or "one or more"?

PR #1170, opened 2017-02-23
Add "clean" command to Graql shell to clean and commit the keyspace

aelred:
Because this is a very dangerous command, a confirmation message appears and the user must type exactly "confirm". Any other string will cancel the clean operation.


aelred:
please retest this please

grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
```diff
@@ -504,6 +509,19 @@ private void rollback() throws IOException {
         session.sendJson(Json.object(ACTION, ACTION_ROLLBACK));
     }
 
+    private void clean() throws IOException {
+        // Get user confirmation to clean graph
+        console.println("Are you sure? This will clean ALL data in the current keyspace and immediately commit.");
```
fppt: I like the need to type commit

PR #1175, opened 2017-02-24
Stop printing warning when initialising the ontology on tinker graph

fppt:


grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -121,12 +122,8 @@ public AbstractGraknGraph(G graph, String keyspace, String engine, boolean batch
...
         localIsOpen.set(true);
 
-        if(initialiseMetaConcepts()) {
-            try {
-                commit();
-            } catch (GraknValidationException e) {
-                throw new RuntimeException(ErrorMessage.CREATING_ONTOLOGY_ERROR.getMessage(e.getMessage()));
-            }
+        if(initialiseMetaConcepts() && !(graph instanceof TinkerGraph)) {
+            commitTransaction();
```
aelred: In the long run, we might want a method to ask what features the graph supports (I think Tinkerpop does a similar thing).

aelred:
sorry @fppt im in ur pr breaking ur tests

please retest this please

aelred:
@fppt I'm not totally sure this PR worked. I'm still seeing a lot of logs: https://s3.amazonaws.com/archive.travis-ci.org/jobs/204988927/log.txt

fppt:
@aelred just checked. The message is never printed on initialising the graph BUT it is printed when performing an actual commit which `GraknGraphPropertyIT` does a lot. Maybe it's worth suppressing WARNS for tests?

aelred:
`GraknGraphPropertyIT` shouldn't be committing... I tried running it - it looks like the system keyspace commits hundreds of times for some reason? I'll investigate more and send you stuff later on slack.

PR #1180, opened 2017-02-24
Add GraknSparkComputer

duckofyork:
Persist rdd after compute query
Add vertex property cleaning
Add singleton class containing graph rdd
Add close method for clearing rdd
Remove property dropping from vertex program and map reduce
Optimse grakn spark computer


conf/main/grakn-analytics.properties
```diff
@@ -20,7 +20,8 @@
 factory.internal=ai.grakn.factory.TitanHadoopInternalFactory
 
 # Graph Computer
-graph.computer=org.apache.tinkerpop.gremlin.spark.process.computer.SparkGraphComputer
+#graph.computer=org.apache.tinkerpop.gremlin.spark.process.computer.SparkGraphComputer
```
aelred: We should get rid of these comments

grakn-test/src/test/java/ai/grakn/test/graql/analytics/ShortestPathTest.java
```diff
@@ -371,5 +379,6 @@ private void addOntologyAndEntities2() throws GraknValidationException {
 
             graph.commitOnClose();
         }
+        GraknSparkComputer.close();
```
aelred: Since this is in `@After`, you don't need it here. (Same in a few other places)

grakn-graph/src/main/java/ai/grakn/graph/internal/computer/GraknSparkComputer.java
```diff
@@ -0,0 +1,408 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Future;
+import java.util.stream.Stream;
+
+/**
+ * @author Jason Liu
+ */
```
aelred: Is this one of the classes we had to copy+paste to modify properly? If it is, the JavaDoc should explain that.

Imagine I have this same comment on all the classes below :P

grakn-test/src/test/java/ai/grakn/test/graql/analytics/ShortestPathTest.java
duckofyork: This method clears the cached rdd. So I also need it whenever I wanna reload the graph.

PR #1183, opened 2017-02-25
Another New Post Processing Test

fppt:
This includes a scaling integration test which replicates the long standing issues we have been having. This should hopefully resolve PP once and for all.


fppt:
please retest this please

PR #1190, opened 2017-02-27
Multiple role player tests

aelred:


aelred:
@fppt tell me if you want to implement on this PR, or merge this and fix the tests later

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
```diff
@@ -116,6 +119,16 @@ public static String generateNewHash(RelationType relationType, Map<RoleType, In
         return roleMap;
     }
 
+    @Override
+    public Collection<Instance> newRolePlayers(RoleType... roleTypes) {
+        // TODO: Implement this in a way that is not pure shit
```
fppt: HA!

fppt:
@aelred I will put a base definition in for the moment. Not good enough for the tests to be passing though, but when we have the base definition in place we can merge even if it is imperfect.

grakn-core/src/main/java/ai/grakn/concept/Relation.java
```diff
@@ -66,9 +67,19 @@
...
     Map<RoleType, Instance> rolePlayers();
 
     /**
+     * Retrieves a list of every {@link Instance} involved in the {@link Relation}, filtered by {@link RoleType} played.
+     * @param roleTypes used to filter the returned instances only to ones that play any of the role types.
+     *                  If blank, returns all role players.
+     * @return a list of every {@link Instance} involved in the {@link Relation}.
+     */
+    // TODO: Rename this method to `rolePlayers`
+    Collection<Instance> newRolePlayers(RoleType... roleTypes);
```
fppt: I don't think this is that useful. Without the context of the role it would be weird. Imagine you have ten roles all played by the same instance? 

I would like to remove this.

fppt: My main concern is that calling `newRolePlayers()` with no parameters provides very little information about the relation. Similarly for passing in more than one role type. How do you know which instance plays which role?

fppt:
Closing this because github magic is not working for me. Already created a new PR.

grakn-core/src/main/java/ai/grakn/concept/Relation.java
aelred: That's true... I added this method to parallel `Instance#relations`, which has the same issues you describe. I'd still keep it, because the alternative `Map<RoleType, Collection<Instance>>` is worse, in my opinion.

PR #1199, opened 2017-02-28
Allowing multiple role players for a single role

fppt:


fppt:
I summon the mighty @aelred to fix the broken graql tests while I slowly get rid of all the uses of old `rolePlayers()`. Come mighty graql master and fix the brokeness! Pizza shall (maybe) be your reward!

grakn-graql/src/main/java/ai/grakn/graql/internal/printer/GraqlPrinter.java
```diff
@@ -90,15 +91,17 @@ public String build(Function<StringBuilder, StringBuilder> builder) {
...
-                        String s = colorType(roleType) + ": id " + idToString(rolePlayer.getId());
-                        return Optional.of(s);
-                    } else {
+                    if(instances.isEmpty()){
                         return Optional.<String>empty();
+                    } else {
+                        StringBuilder innerSb = new StringBuilder();
+                        instances.forEach(rolePlayer ->
+                                innerSb.append(colorType(roleType)).append(": id ").append(idToString(rolePlayer.getId())).append(","));
+                        return Optional.of(innerSb.toString());
```
aelred: This one is interesting. What would help here (and elsewhere) is a `Collection<Pair<RoleType, Instance>>`, which the API doesn't provide. I'm not advocating adding it now, but it's something to think about!

grakn-test/src/test/java/ai/grakn/test/migration/MigratorTestUtils.java
```diff
@@ -116,7 +116,8 @@ public static Resource getResource(GraknGraph graph, Instance instance, TypeName
         RoleType roleOther = graph.getType(Schema.Resource.HAS_RESOURCE_VALUE.getName(name));
 
         Collection<Relation> relations = instance.relations(roleOwner);
-        return relations.stream().map(r -> r.rolePlayers().get(roleOther).asResource());
+        //TODO: Cleanup use of iterator here
+        return relations.stream().map(r -> r.rolePlayers(roleOther).iterator().next().asResource());
```
aelred: This might be a bit neater:
```
return relations.stream().flatMap(r -> r.rolePlayers(roleOther).stream()).map(Concept::asResource);
```

aelred:
I think it's best that we address these changes even though it's `unstable`. This is because we may want to backport this to `master`, where it would be useful for it to be a single commit.

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
```diff
@@ -83,39 +84,56 @@ public void setHash(){
...
         for(RoleType role: sortedRoleIds){
             hash.append("_").append(role.getId().getValue().replace("_", "\\_"));
-            Instance instance = roleMap.get(role);
-            if(instance != null){
-                hash.append("_").append(instance.getId().getValue().replace("_", "\\_"));
-            }
+
+            roleMap.get(role).forEach(instance -> {
+                if(instance != null){
+                    hash.append("_").append(instance.getId().getValue().replace("_", "\\_"));
```
fppt: This is really the edgiest case of all. I think real hashes have more likelihood of overlap.

grakn-graph/src/test/java/ai/grakn/graph/internal/EntityTest.java
```diff
@@ -348,4 +330,18 @@ public void testNoKey() throws GraknValidationException {
...
+            RoleType roleType = entry.getKey();
+            entry.getValue().forEach(instance -> {
+                if(instance.equals(entity)){
+                    assertEquals(Schema.Resource.HAS_RESOURCE_OWNER.getName(resourceType.getName()), roleType.getName());
+                } else {
+                    assertEquals(Schema.Resource.HAS_RESOURCE_VALUE.getName(resourceType.getName()), roleType.getName());
+                }
+            });
+        });
+    }
```
fppt: Can you clarify what you mean here?

grakn-graql/src/main/java/ai/grakn/graql/internal/hal/HALConceptData.java
```diff
@@ -243,13 +234,15 @@ private void attachRelation(Representation halResource, Concept rel, TypeName ro
...
-            }
+        rel.allRolePlayers().forEach((roleType, instanceSet) -> {
+            instanceSet.forEach(instance -> {
+                if (instance != null) {
+                    Representation roleResource = factory.newRepresentation(resourceLinkPrefix + instance.getId() + this.keyspace)
+                            .withProperty(DIRECTION_PROPERTY, OUTBOUND_EDGE);
+                    handleConcept(roleResource, instance, separationDegree - 1);
+                    halResource.withRepresentation(roleType.getName().getValue(), roleResource);
+                }
+            });
```
fppt: Dashboard will be broken. I say we merge and then ask Marco to take a look at it. This PR is blocking on a few things.

grakn-graph/src/test/java/ai/grakn/graph/internal/EntityTest.java
aelred: We should check that there are two role types, and that each has exactly one role-player.
This isn't an important issue though, I'm happy to merge without it.

PR #1212, opened 2017-03-01
Single Queue task manager

alexandraorth:


grakn-client/src/main/java/ai/grakn/client/LoaderClient.java
```diff
@@ -240,7 +242,7 @@ private void unblock(CompletableFuture<Json> status){
      *
      * @return A Completable future that terminates when the task is finished
      */
-    private CompletableFuture<Json> executePost(String body) throws HttpRetryException {
+    private String executePost(String body) throws HttpRetryException {
```
aelred: Should these return `TaskId`?

bolerio:
This is again one of those huge PRs. We need to stop doing those. It's one of the reasons I wanted to keep to current implementation in place and do the new in the same code base, so we could make the changes more incremental. It's failing due to zookeeper not connecting. Could be again an issue with Travis....

PR #1214, opened 2017-03-02
Refactor `EngineID` to be passed instead of a singleton

aelred:


aelred:
please retest this please

PR #1215, opened 2017-03-02
Engine Configure type of task storage

alexandraorth:


conf/test/titan/grakn-engine.properties
```diff
@@ -18,6 +18,7 @@
 
 #In-Memory or Distributed task manager
 taskmanager.implementation=ai.grakn.engine.tasks.manager.StandaloneTaskManager
+taskmanager.storage.zk=true
```
fppt: All of these are `true` do we have a test in place which operates under the assumption of this being `false`. Just want to know if we tracking the support of the opposite of this config option.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/singlequeue/SingleQueueTaskManager.java
```diff
@@ -77,7 +80,12 @@
      */
     public SingleQueueTaskManager(EngineID engineId){
         this.zookeeper = new ZookeeperConnection(client());
-        this.storage = new TaskStateZookeeperStore(zookeeper);
+
+        if(properties.getPropertyAsBool(USE_ZOOKEEPER_STORAGE)){
```
fppt: Duplicate Code [1]: No chance of putting this somewhere common? Default method? Abstract class?

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/multiqueue/MultiQueueTaskManager.java
```diff
@@ -61,14 +64,19 @@
 
     public MultiQueueTaskManager(EngineID engineId) {
         this.engineId = engineId;
-        connection = new ZookeeperConnection(client());
-        stateStorage = new TaskStateZookeeperStore(connection);
+        this.zookeeper = new ZookeeperConnection(client());
+
+        if(properties.getPropertyAsBool(USE_ZOOKEEPER_STORAGE)){
```
fppt: Duplicate Code [1]: No chance of putting this somewhere common? Default method? Abstract class?

grakn-test/src/test/java/ai/grakn/generator/TaskStates.java
```diff
@@ -79,7 +80,12 @@ public TaskState generate(SourceOfRandomness random, GenerationStatus status) {
         Json configuration = Json.object();
         TaskState taskState = TaskState.of(taskClass, creator, TaskSchedule.now(), configuration, taskId);
         configuration.set("id", taskState.getId().getValue());
-        return taskState.status(taskStatus);
+        if(taskStatus == TaskStatus.RUNNING){
+            taskState.setRunning(EngineID.me());
+        } else {
+            taskState.status(taskStatus);
```
fppt: When a task changes status from no longer running does the EngineID on the task get wiped? Should it or shouldn't it/

grakn-test/src/test/java/ai/grakn/test/engine/GraknEngineServerTest.java
```diff
@@ -22,51 +22,90 @@
...
+    public final EngineContext kafka = EngineContext.startKafkaServer();
+
+    @After
+    public void stopEngine(){
+        GraknEngineServer.stop();
+    }
+
     @Test
-    public void testInMemoryMain() throws Exception {
+    public void whenEnginePropertiesIndicatesStandaloneTM_StandaloneTmIsStarted() {
```
fppt: Much better :)

grakn-test/src/test/java/ai/grakn/generator/TaskStates.java
alexandraorth: Only `RUNNING` tasks should have an EngineID because this is used during failover. When an engine fails, we ask the storage for any `RUNNING` tasks that may have been on that machine so that we can resubmit them to another machine. 

We could potentially keep the EngineID and then filter by `RUNNING` state, but it would make querying during failover much heavier.

conf/test/titan/grakn-engine.properties
alexandraorth: Yep! Cant figure out how to link: It is test `whenEnginePropertiesDoesNotIndicateZookeeperStorage_GraphStorageIsUsed` in `GraknEngineServerTest`

fppt: Okay cool. Just wanted to make sure we covering out bases. [DONE]

fppt:
please retest this please

PR #1218, opened 2017-03-02
Support multiple engines in the same JVM

aelred:


aelred:
please retest this please

alexandraorth:
At least it passes on travis :P

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -49,125 +49,132 @@
...
+
+    public static GraknEngineServer mainWithServer() {
+        // Start Engine
+        int port = prop.getPropertyAsInt(ConfigProperties.SERVER_PORT_NUMBER);
+        String taskManagerClass = prop.getProperty(TASK_MANAGER_IMPLEMENTATION);
+        GraknEngineServer server = start(taskManagerClass, port);
+
         // close GraknEngineServer on SIGTERM
-        Thread closeThread = new Thread(GraknEngineServer::stop, "GraknEngineServer-shutdown");
+        Thread closeThread = new Thread(server::close, "GraknEngineServer-shutdown");
```
alexandraorth: Should we be adding the shutdown hook here? In our tests we will be adding the hook many times, which may not be ideal.

alexandraorth: Should we be adding the shutdown hook here? In our tests we will be adding the hook many times, which may not be ideal.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/ZookeeperConnection.java
```diff
@@ -45,11 +53,15 @@
...
+
     /**
      * Start the connection to zookeeper. This method is blocking.
      */
-    public ZookeeperConnection(CuratorFramework zookeeperConnection) {
-        this.zookeeperConnection = zookeeperConnection;
+    public ZookeeperConnection() {
+        zookeeperConnection = client();
+
+        CONNECTION_COUNTER.incrementAndGet();
```
alexandraorth: It seems like here we are always starting the Zookeeper connection, but in the `close` method below, it only gets closed once. 
Do we need this counter? Why dont we try and use a singleton instead?

PR #1221, opened 2017-03-03
Refactor client to use unirest for HTTP requests

aelred:
Also add `TaskClient` for stopping tasks


aelred:
Closing this because it is entirely broken for some reason! I will work on introducing these changes more incrementally.

PR #1224, opened 2017-03-03
Fix the bug that causes statistics to ignore the type of resource owner

duckofyork:
Fix the bug that causes statistics to ignore the type of resource owner
-Add a new degree vertex program for statistics

### DIFF
--- a/grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/DegreeStatisticsVertexProgram.java
+++ b/grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/DegreeStatisticsVertexProgram.java
@@ -21,18 +21,13 @@ package ai.grakn.graql.internal.analytics;
 import ai.grakn.concept.TypeName;
 import ai.grakn.util.Schema;
 import com.google.common.collect.Sets;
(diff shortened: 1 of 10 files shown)


grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/DegreeStatisticsVertexProgram.java
```diff
@@ -0,0 +1,137 @@
...
+    static void degreeStatisticsStepRelation(Vertex vertex, Messenger<Long> messenger) {
+        if (vertex.label().equals(Schema.BaseType.RELATION.name()) && messenger.receiveMessages().hasNext()) {
+            messenger.sendMessage(messageScopeOutCasting, 1L);
+        }
+    }
+
+    static void degreeStatisticsStepCastingOut(Vertex vertex, Messenger<Long> messenger) {
+        if (vertex.label().equals(Schema.BaseType.CASTING.name()) && !vertex.property(VISITED).isPresent()
+                && messenger.receiveMessages().hasNext()) {
+            vertex.property(VISITED, true);
```
aelred: No need to mark as visited this time.

PR #1226, opened 2017-03-06
TaskState refactor

alexandraorth:


grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/StandaloneTaskManager.java
```diff
@@ -169,22 +168,20 @@ public TaskStateStorage storage() {
...
-                stateStorage.updateState(state
-                        .status(FAILED)
-                        .statusChangedBy(EXCEPTION_CATCHER_NAME)
-                        .exception(getFullStackTrace(t)));
+            catch (Throwable throwable) {
+                LOG.error(getFullStackTrace(throwable));
+                state.markFailed(throwable);
             }
+            stateStorage.updateState(state);
+            stateUpdateLock.unlock();
```
aelred: Might need to go in a `finally`.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/ZookeeperConnection.java
```diff
@@ -84,7 +83,7 @@ public CuratorFramework connection(){
         return zookeeperConnection;
     }
 
-    public InterProcessMutex mutex(TaskId id){
+    public InterProcessMutex mutex(String id){
```
aelred: :(

grakn-graql/src/main/java/ai/grakn/graql/internal/template/TemplateVisitor.java
```diff
@@ -346,7 +346,7 @@ public String visitReplaceStatement(GraqlTemplateParser.ReplaceStatementContext
         if(value == ObjectUtils.NULL){
             throw new GraqlTemplateParsingException("Key [" + ctx.getText() + "] not present in data: " + originalContext);
         }
-
+        ;
```
aelred: good stuff

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/ZookeeperConnection.java
alexandraorth: You can now get a mutex on the engine ID or the task ID. This is why I was asking about that ID interface earlier ;) Related: in `TaskId` we use `getValue` and in `EngineID` we use `value()`

PR #1229, opened 2017-03-06
Pipeline cleanup and more speed

kasper-piskorski:


grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerAtomicQuery.java
```diff
@@ -299,45 +297,39 @@ public void unify(Map<VarName, VarName> unifiers) {
...
+        //if query not exactly equal to the rule head, do some conversion
+        if (!this.equals(ruleHead)) {
+            Set<VarName> vars = this.getVarNames();
+            Set<Map<VarName, VarName>> permutationUnifiers = getPermutationUnifiers(ruleHead.getAtom());
+            Set<IdPredicate> unmappedIdPredicates = atom.getUnmappedIdPredicates();
             Set<TypeAtom> mappedTypeConstraints = atom.getMappedTypeConstraints();
-            answers = dCache.record(ruleHead, newAnswers)
-                    .filter(a -> entityTypeFilter(a, mappedTypeConstraints));
+            Set<TypeAtom> unmappedTypeConstraints = atom.getUnmappedTypeConstraints();
+            answers = this.getIdPredicateAnswerStream(answers)
```
fppt: I find this lambda incredibly difficult to follow. I don't have a clue what any of these functions do. It may be due to my lack of understanding of the component but maybe we can come up with better and more clear names for all the functions in this lambda.

kasper-piskorski:
please retest this please

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerAtomicQuery.java
kasper-piskorski: filter answers by ensuring each concept fits into type hierarchy specified by the query :)

fppt: With big chunks of code like this a one, a comment like your reply can clarify a lot. Not every line needs a comment but in cases like this I think it would be warranted.

PR #1231, opened 2017-03-06
Loader Client periodic post processing

alexandraorth:
Every 100 batches the Loader Client will:
1. Wait for all submitted tasks to finish
2. Submit a post processing task
3. Wait for this post processing task to finish

### DIFF
--- a/grakn-client/src/main/java/ai/grakn/client/LoaderClient.java
+++ b/grakn-client/src/main/java/ai/grakn/client/LoaderClient.java
@@ -40,6 +40,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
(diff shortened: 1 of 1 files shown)


alexandraorth:
This still needs tests. 

When using the MQTM, Postprocessing is a recurring task and this is not necessary, but still helpful to avoid memory overflows server-side when using the client. 

I'm not 100% sure this should go in central. What do you think @bolerio ?

bolerio:
@alexandraorth It's the only available option that we have right now to actually run post-processing during long running loads. An alternative is the not forcefully trigger the PP task, but just wait for its timeout and for it to trigger on its own. Just as bad. Anyway, we have to put it in central so we have a working loading mechanism, and put a task in TP to remove after PP is completed.

fppt:
By the way guys there is one way we can cut our PP caching down substantially. We just get rid of merging castings. It will have no impact on results what so ever. The only reason we do it is for the sake of perfection :)

bolerio:
@fppt that will just delay the problem... but didn't castings have the potential to alleviate the supernode problem actually?

fppt:
@bolerio potentially they do but only a specific super node problem. Castings wont help with trillions of instances for example.

Anyway, if we do seriously want to use castings to alleviate the super node problem then getting rid of the PP job which merges them is a good start.

alexandraorth:
I've also realised this design has a serious flaw. Every 1000 batches we wait for all tasks to complete on all of the machines yet send only 1 postprocessing task. Our cluster has 6 machines and only one of them will pick up and execute this task. Because PP is machine-local we won't actually be postprocessing everything each time :P

PR #1233, opened 2017-03-07
Add support for stopping tasks

aelred:
Currently this will only work correctly if you send the request to stop a task to the engine it is running on.

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/loader/LoaderTask.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/loader/LoaderTask.java
@@ -59,15 +59,16 @@ public class LoaderTask implements BackgroundTask {
     private final QueryBuilder builder = Graql.withoutGraph().infer(false);
 
     @Override
-    public void start(Consumer<String> saveCheckpoint, Json configuration) {
+    public boolean start(Consumer<String> saveCheckpoint, Json configuration) {
         attemptInsertions(
                 getKeyspace(configuration),
                 getInserts(configuration));
+        return true;
     }
 
     @Override
-    public void stop() {
-        throw new UnsupportedOperationException("Loader task cannot be stopped");
+    public boolean stop() {
+        return false;
     }
 
     @Override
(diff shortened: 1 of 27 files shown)


grakn-engine/src/main/java/ai/grakn/engine/loader/LoaderTask.java
```diff
@@ -59,15 +59,16 @@
     private final QueryBuilder builder = Graql.withoutGraph().infer(false);
 
     @Override
-    public void start(Consumer<String> saveCheckpoint, Json configuration) {
+    public boolean start(Consumer<String> saveCheckpoint, Json configuration) {
         attemptInsertions(
                 getKeyspace(configuration),
                 getInserts(configuration));
+        return true;
```
fppt: A method that always returns true. I am surprised nothing complained.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/singlequeue/SingleQueueTaskManager.java
```diff
@@ -142,8 +142,15 @@ public void addTask(TaskState taskState){
      * Stop a task from running.
      */
     @Override
-    public TaskManager stopTask(TaskId id, String requesterName) {
-        throw new UnsupportedOperationException("SingleQueueTaskManager does not support stopping tasks.");
+    public void stopTask(TaskId id, String requesterName) {
+        // TODO: Make only one call to storage if possible
+        if (!storage.containsTask(id)) {
+            // TODO: Figure out a nicer way than all these nulls...
+            TaskState task = TaskState.of(null, null, null, null, id).markStopped();
```
fppt: Why not just make alternative `TaskState.of()` with a single parameter?

grakn-engine/src/main/java/ai/grakn/engine/loader/LoaderTask.java
aelred: Yup! There isn't an obvious way to stop loader tasks just yet.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/singlequeue/SingleQueueTaskManager.java
aelred: Honestly, because I'm hoping there's a better way than using all these nulls. It doesn't seem right that all properties of a `TaskState` except ID can be null. For example, a `TaskState` until this point is meant to _always_ contain a class indicating the task to run, plus a schedule saying when to run it - now both of these things can be missing.

grakn-engine/src/main/java/ai/grakn/engine/loader/LoaderTask.java
fppt: Ahh that is my confusion then. So it's on the todo of making loader tasks stoppable ?

aelred: Yes, although not in the immediate future.

PR #1234, opened 2017-03-07
`@Ignore` `TaskFailoverTest` because it frequently fails randomly

aelred:


fppt:
Just going to push this in because I am guessing the random failures are annoying a few people.

PR #1235, opened 2017-03-08
Experimental Scoping Of Types

fppt:
- Move methods to Type Interface
- Simplify Tests
- Change HasScopeProperty

### DIFF
--- a/grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
+++ b/grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
@@ -23,7 +23,6 @@ import ai.grakn.concept.Relation;
 import ai.grakn.concept.RelationType;
 import ai.grakn.concept.Resource;
 import ai.grakn.concept.RoleType;
-import ai.grakn.exception.ConceptException;
 import ai.grakn.exception.ConceptNotUniqueException;
 import ai.grakn.util.ErrorMessage;
 import ai.grakn.util.Schema;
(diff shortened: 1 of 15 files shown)


grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
```diff
@@ -313,6 +313,7 @@ private void deleteCachedDirectedSubType(T oldSubType){
             ConceptImpl<Concept> concept = getGraknGraph().getElementFactory().buildConcept(vertex);
             if(!concept.isCasting()){
                 instances.add((V) concept);
+                instances.add((V) concept);
```
aelred: Looks like a mistake?

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
```diff
@@ -234,7 +201,6 @@ void cleanUp() {
      */
     @Override
     public void delete() {
-        scopes().forEach(this::deleteScope);
```
aelred: I don't see the equivalent for this in `TypeImpl`

fppt: Nope. Relations no longer have scopes so why would you delete them?

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
fppt: Yep. Will fix

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
aelred: Sorry, dumb comment: I meant to say, do types delete their scopes now?

fppt: Yes they do. You can delete the scope from a type. If you asking if the scope dies when the type dies then the answer to that is no.

aelred: Thanks, that's fine! Am I right in thinking that deleting the relation also used to delete the scope? Not that it matters, since we weren't using the old scope behaviour.

fppt: Onnncccceeee upon a time yes.

PR #1236, opened 2017-03-08
Engine remove distributed ZK lock & Fix Atomicity issue 

alexandraorth:


grakn-engine/src/main/java/ai/grakn/engine/tasks/storage/TaskStateZookeeperStore.java
```diff
@@ -100,8 +104,6 @@ public Boolean updateState(TaskState task){
             // Get the previously stored task
             TaskState previousTask = (TaskState) deserialize(zookeeper.connection().getData().forPath(taskPath(task)));
 
-            return executeWithMutex(() -> {
-
```
aelred: Your lack of consistent indentation disturbs me.

PR #1242, opened 2017-03-09
Ctrl drag nodes

marco-scoppetta:
When drag a node/multiple nodes with ctrl key pressed, also drag all the connected nodes to the selected ones.


fppt:
There are only JS changes so I am going to just accept. This needs to be pushed into unstable.

PR #1243, opened 2017-03-09
Engine SQTR takes in a SQTM as parameter

alexandraorth:


grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/singlequeue/SingleQueueTaskRunner.java
```diff
@@ -222,4 +220,13 @@ private void debugConsumerStatus(ConsumerRecords<TaskId, TaskState> records ){
...
     }
+
+    /**
+     * Get a new kafka consumer listening on the new tasks topic
+     */
+    private static Consumer<TaskId, TaskState> consumer(SingleQueueTaskManager manager){
+        Consumer<TaskId, TaskState> consumer = kafkaConsumer(TASK_RUNNER_GROUP);
+        consumer.subscribe(ImmutableList.of(NEW_TASKS_TOPIC), rebalanceListener(consumer, manager.zookeeper()));
+        return consumer;
+    }
```
aelred: We can put this in the manager class instead, to avoid exposing zookeeper.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/singlequeue/SingleQueueTaskRunner.java
```diff
@@ -193,7 +189,9 @@ private void handleRecord(ConsumerRecord<TaskId, TaskState> record) {
             } finally {
                 runningTask = null;
                 runningTaskId = null;
-                storage.updateState(task);
+
+                //TODO Using "manager" here is only so that PMD will be happy
+                manager.storage().updateState(task);
```
aelred: Does this still need to be here?

PR #1246, opened 2017-03-09
Engine SQTM delayed tasks

alexandraorth:


alexandraorth:
please retest this please

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/singlequeue/SingleQueueTaskRunner.java
```diff
@@ -134,49 +135,60 @@ public boolean stopTask(TaskId taskId) {
     /**
      * Returns false if cannot handle record because the executor is full
      */
-    private void handleRecord(ConsumerRecord<TaskId, TaskState> record) {
-        TaskState task = record.value();
-
+    private void handleRecord(TaskState task) {
```
aelred: Might best be called `handleTask` or `handleTaskState` now

PR #1250, opened 2017-03-09
Make findbugs more strict

aelred:
I've fixed the "dumb" bugs, the more complex ones have been left and the bugs in question ignored in the findbugs config.

### DIFF
--- a/grakn-core/src/main/java/ai/grakn/Grakn.java
+++ b/grakn-core/src/main/java/ai/grakn/Grakn.java
@@ -18,7 +18,9 @@
 
 package ai.grakn;
 
+import java.lang.reflect.InvocationTargetException;
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 
 /**
@@ -59,9 +61,9 @@ public class Grakn {
             @SuppressWarnings("unchecked")
             Class<F> cl = (Class<F>)Class.forName(className);
             return cl.getConstructor(String.class, String.class).newInstance(keyspace, location);
-        }
-        catch (Exception ex) {
-            throw new RuntimeException(ex);
+        } catch (InstantiationException | InvocationTargetException | IllegalAccessException | NoSuchMethodException
+                | ClassNotFoundException e) {
+            throw new RuntimeException(e);
         }
     }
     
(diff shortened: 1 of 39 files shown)


grakn-core/src/main/java/ai/grakn/Grakn.java
```diff
@@ -59,9 +61,9 @@
             @SuppressWarnings("unchecked")
             Class<F> cl = (Class<F>)Class.forName(className);
             return cl.getConstructor(String.class, String.class).newInstance(keyspace, location);
-        }
-        catch (Exception ex) {
-            throw new RuntimeException(ex);
+        } catch (InstantiationException | InvocationTargetException | IllegalAccessException | NoSuchMethodException
```
fppt: `ex` offended you?

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
```diff
@@ -202,7 +202,7 @@ private String scheduleTask(Request request, Response response) {
 
             response.type("application/json");
             return Json.object("id", taskState.getId().getValue()).toString();
-        } catch (Exception e) {
+        } catch (ClassNotFoundException | RuntimeException e) {
```
fppt: If findbugs forced more specific error catching bit +1 from me!

grakn-engine/src/main/java/ai/grakn/engine/loader/LoaderTask.java
```diff
@@ -143,7 +143,7 @@ private void throwException(String message, Collection<InsertQuery> inserts){
      * @param e exception to log
      * @param i amount of time to sleep
      */
-    private void handleError(Throwable e, int i) {
+    private void handleError(Throwable e, long i) {
```
fppt: Can we really fail sooo many time in a so as to need a long here?

grakn-core/src/main/java/ai/grakn/Grakn.java
aelred: I took exception to it.

grakn-engine/src/main/java/ai/grakn/engine/loader/LoaderTask.java
aelred: We've caused `OutOfMemory` on Maven due to error messages before, so... :P

fppt: Okay but realistically if you fail more then 2147483647 times then you probably not going to succeed on the 2147483648th time. Surely at some point we say enough enough and just log the permanent failure?

aelred: In this case, the change is because `Thread.sleep` accepts a long, rather than because of our lack of faith in our code quality :P

I don't necessary know what this code does, all I know is it works with `long`, not `int`.

fppt: Fair enough.

PR #1252, opened 2017-03-09
materialisation fixes

kasper-piskorski:
- global cache to ensure no duplicates produced
- inverse cache lookup for checking for answer existence


kasper-piskorski:
please retest this please

PR #1253, opened 2017-03-09
Engine SQTM Recurring Tasks

alexandraorth:


grakn-engine/src/main/java/ai/grakn/engine/tasks/TaskState.java
```diff
@@ -191,6 +191,11 @@ public TaskSchedule schedule() {
         return schedule;
     }
 
+    public TaskState schedule(TaskSchedule schedule){
+        this.schedule = schedule;
+        return this;
+    }
```
aelred: We could drop schedule from the `TaskState.of` methods now if we wanted, and make the default `now()`.

PR #1256, opened 2017-03-10
Client-side logging during graql -b

alexandraorth:


grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
```diff
@@ -252,7 +255,18 @@ private static String loadQuery(String filePath) throws IOException {
     }
 
     private static void sendBatchRequest(String uriString, String graqlPath, String keyspace) throws IOException {
-        LoaderClient loaderClient = new LoaderClient(keyspace, uriString);
+        AtomicInteger numberBatchesCompleted = new AtomicInteger(0);
+
+        LoaderClient loaderClient = new LoaderClient(keyspace, uriString).setRetryPolicy(true);
+        loaderClient.setTaskCompletionConsumer((json) -> {
```
aelred: Should this handle `json` being `null` (when there are connection issues)?

PR #1257, opened 2017-03-10
Add `TaskClient` for stopping tasks

aelred:


grakn-engine/src/main/java/ai/grakn/engine/tasks/config/ConfigHelper.java
```diff
@@ -18,7 +18,7 @@
 
 package ai.grakn.engine.tasks.config;
 
-import ai.grakn.engine.tasks.TaskId;
+import ai.grakn.engine.TaskId;
```
alexandraorth: makes me sad this is in different package from `TaskState` now :(

aelred: But the same package as `TaskId`! Something probably needs to change, but I don't know which.

alexandraorth: Yeah. There is the whole question of "should internal-engine classes be in `core`?" to also consider. Although maybe there are no longer internal engine classes.

grakn-core/src/main/java/ai/grakn/util/REST.java
```diff
@@ -47,6 +47,7 @@
         public static final String ALL_TASKS_URI = "/tasks/all";
         public static final String TASKS_URI = "/tasks";
         public static final String TASKS_SCHEDULE_URI ="/tasks/schedule";
+        public static final String TASKS_STOP_URI = "/tasks/{id}/stop";
```
alexandraorth: This is so nice! We should do this to all our URIs

grakn-test/src/test/java/ai/grakn/test/client/TaskClientTest.java
```diff
@@ -0,0 +1,70 @@
...
+    private Service spark;
+
+    @Before
+    public void setUp() {
+        client = TaskClient.of(Grakn.DEFAULT_URI);
+
+        spark = Service.ignite();
+        spark.port(4567);
+
+        manager = mock(TaskManager.class);
```
alexandraorth: Mockito is THE BEST THING EVER

PR #1260, opened 2017-03-10
Nested Transactions (Kind of)

fppt:
- Implement internal Transaction Count
- Cleanup factory implementation
- Take nested transaction into account when trying to close the factory


fppt:
Closing this as I don't think we have thought hard enough about using Fake Nested Transactions. The current behaviour is very confusing.

PR #1261, opened 2017-03-10
Consolidating properties files (Step 1)

alexandraorth:
This pull req consolidates all three of the properties file into one.
Tasks that remain:
+ Sections in the properties files with headers for different services (`kafka`, `graql`)
+ Detailed commenting of each property
+ Refactor of the `GraknEngineConfig` class - potentially extend `Properties` class. 
+ `GraknEngineServer` should accept a generic java `Properties` object
+ Remove `StatusController` - this now does essentially the same thing as `GraphFactoryController`
+ Make copies of properties objects in Kafka ConfigHelper


alexandraorth:
please retest this please

conf/test/orientdb/grakn.properties
```diff
@@ -48,12 +47,12 @@ blockingLoader.batch-size=1
...
-tasks.kafka.producer.retries=0
-tasks.kafka.producer.batch-size=16384
-tasks.kafka.producer.linger-ms=1
-tasks.kafka.producer.buffer-mem=33554432
+bootstrap.servers=localhost:9092
+consumer.session-timeout=30000
+producer.retries=0
+producer.batch-size=16384
+producer.linger-ms=1
+producer.buffer-mem=33554432
```
aelred: Do we have to repeat this here, or can we somehow inherit them from the properties file above?

grakn-engine/src/main/java/ai/grakn/engine/tasks/config/ConfigHelper.java
```diff
@@ -47,17 +44,11 @@
 public class ConfigHelper {
 
     public static Consumer<TaskId, TaskState> kafkaConsumer(String groupId) {
-        Properties properties = new Properties();
-        properties.put("bootstrap.servers", ConfigProperties.getInstance().getProperty(KAFKA_BOOTSTRAP_SERVERS));
+        Properties properties = GraknEngineConfig.getInstance().getProperties();
```
aelred: Does this mean this method now writes to the properties singleton?

grakn-engine/src/main/java/ai/grakn/engine/tasks/config/ConfigHelper.java
```diff
@@ -66,23 +57,11 @@
...
-        properties.put("linger.ms", ConfigProperties.getInstance().getPropertyAsInt(KAFKA_LINGER_MS));
-        properties.put("buffer.memory", ConfigProperties.getInstance().getPropertyAsInt(KAFKA_BUFFER_MEM));
-        properties.put("key.serializer", "ai.grakn.engine.tasks.TaskIdSerializer");
-        properties.put("key.deserializer", "ai.grakn.engine.tasks.TaskIdDeserializer");
-        properties.put("value.serializer", "ai.grakn.engine.tasks.TaskStateSerializer");
-        properties.put("value.deserializer", "ai.grakn.engine.tasks.TaskStateDeserializer");
-
-        return new KafkaProducer<>(properties);
+        Properties properties = GraknEngineConfig.getInstance().getProperties();
+        return new KafkaProducer<>(properties, new TaskIdSerializer(), new TaskStateSerializer());
```
aelred: Same question here

grakn-engine/src/main/java/ai/grakn/engine/controller/GraphFactoryController.java
```diff
@@ -75,27 +76,29 @@ public GraphFactoryController(Service spark) {
...
+        Properties properties = new Properties();
+        properties.putAll(GraknEngineConfig.getInstance().getProperties());
+        switch (graphConfig) {
+            case REST.GraphConfig.DEFAULT:
+                break; // Factory is already correctly set
+            case REST.GraphConfig.COMPUTER:
+                properties.setProperty(FACTORY_INTERNAL, properties.get(FACTORY_ANALYTICS).toString());
+                break;
+            default:
+                throw new RuntimeException("Unrecognised graph config: " + graphConfig);
```
aelred: It seems weird we need to switch and modify our returned config based on a parameter.

fppt: In this case we need to. This is because analytics requires one type of graph computer and titan requires another. 

There might be a more elegant way to do this though.

grakn-graph/src/main/java/ai/grakn/factory/GraknGraphFactoryImpl.java
```diff
@@ -157,21 +156,17 @@ static ConfiguredFactory configureGraphFactory(String keyspace, String location,
...
-                computer = properties.get(COMPUTER).toString();
-            }
+        Properties properties = new Properties();
+        properties.putAll(read(contactEngine(restFactoryUri, REST.HttpConn.GET_METHOD)).asMap());
 
-            return new ConfiguredFactory(computer, FactoryBuilder.getFactory(keyspace, engineUrl, properties));
-        } catch (IOException e) {
-            throw new IllegalArgumentException(ErrorMessage.CONFIG_NOT_FOUND.getMessage(engineUrl, e.getMessage()));
+        String computer = null;
+        if(properties.containsKey(COMPUTER)){
```
fppt: I think we can actually remove this check now. There should always be a computer I believe.

alexandraorth: Yes, I agree. Will do this now.

conf/test/orientdb/grakn.properties
alexandraorth: These are the testing config files- I think it would be a good idea to change the few properties we modify for testing programmatically instead.

burukuru:
retest this please

alexandraorth:
please retest this please

PR #1264, opened 2017-03-10
Add kill and recovery method for analytics

duckofyork:


duckofyork:
please retest this please

duckofyork:
please retest this please

duckofyork:
please retest this please

grakn-graph/src/main/java/ai/grakn/graph/internal/GraknComputerImpl.java
```diff
@@ -50,17 +51,18 @@
  */
 public class GraknComputerImpl implements GraknComputer {
     private final Graph graph;
-    private final Class<? extends GraphComputer> graphComputer;
+    private final Class<? extends GraphComputer> graphComputerClass;
```
fppt: Can't you just get this information from the graph computer you have in the class?

grakn-graph/src/main/java/ai/grakn/graph/internal/GraknComputerImpl.java
```diff
@@ -91,16 +101,16 @@ private RuntimeException asRuntimeException(Throwable throwable) {
      * @return A graph compute supported by this grakn graph
      */
     @SuppressWarnings("unchecked")
-    protected Class<? extends GraphComputer> getGraphComputer(String graphComputerType) {
+    protected Class<? extends GraphComputer> getGraphComputerClass(String graphComputerType) {
         try {
             return (Class<? extends GraphComputer>) Class.forName(graphComputerType);
```
fppt: Instead of doing this why not just say `graphComputer.getClass()` ?

grakn-graph/src/main/java/ai/grakn/graph/internal/GraknComputerImpl.java
bolerio: Looks like the class name is passed at construction time, but the actual computer is lazily created only when needed. Not sure if there is a penalty to create the computer right away.

fppt: Is there a penalty @duckofyork ?

duckofyork: Graph computer contains hadoopConfiguration and reference to the preloaded spark rdd. So it is better to create it lazily.

grakn-graph/src/main/java/ai/grakn/graph/internal/GraknComputerImpl.java
duckofyork: I think Factory gets this information from conf file, so grakn computer may have a spark computer inside or, e.g. a fulgora computer.

PR #1268, opened 2017-03-13
Ignore randomly failing `GraqlShellIT`

aelred:


fppt:
Pushing in since it's a simple ignore.

PR #1270, opened 2017-03-13
Change the way we validate vertices and build unknown concepts. This should be more robust.

fppt:


grakn-graph/src/main/java/ai/grakn/graph/internal/ElementFactory.java
```diff
@@ -181,15 +184,35 @@ RoleTypeImpl buildRoleType(Vertex vertex, RoleType type, Boolean isImplicit){
...
         return graknGraph.getConceptLog().getCachedConcept(conceptId);
     }
 
-    EdgeImpl buildEdge(org.apache.tinkerpop.gremlin.structure.Edge edge, AbstractGraknGraph graknGraph){
+    //TODO: Simplify this if it does not make a difference to large loading
+    private Schema.BaseType getBaseType(Vertex vertex){
+        try {
+            return Schema.BaseType.valueOf(vertex.label());
+        } catch (IllegalArgumentException e){
+            //Base type appears to be invalid. Let's try getting the type via the isa edge
```
aelred: So this will work for instances, but not types? I guess the main issue this is solving is ghosts when loading lots of instances, so it's fine.

PR #1272, opened 2017-03-14
Remove unused abort query behaviour

aelred:
This used to be used on the client-side with Ctrl-C, but it never worked reliably. If we still need it, it should be re-worked from the ground up.


aelred:
please retest this please

fppt:
please retest this please

PR #1279, opened 2017-03-14
Change logs to never hide `ai.grakn`

aelred:
Now we print all logs from `ai.grakn` as `INFO` and everything else only if it is `ERROR`.

This means we don't see loads of cassandra/titan logs, but we still see real problems from them. Additionally, we see a reasonable amount of output from `ai.grakn`, without overwhelming the longs.

### DIFF
--- a/grakn-test/src/test/java/ai/grakn/test/GraknTestEnv.java
+++ b/grakn-test/src/test/java/ai/grakn/test/GraknTestEnv.java
@@ -147,6 +147,10 @@ public abstract class GraknTestEnv {
 
     public static void hideLogs() {
         ((Logger) org.slf4j.LoggerFactory.getLogger(ROOT_LOGGER_NAME)).setLevel(Level.ERROR);
+
+        // Some dependencies are using log4j, so must be managed separately
+        org.apache.log4j.Logger.getRootLogger().setLevel(org.apache.log4j.Level.ERROR);
+
         ((Logger) org.slf4j.LoggerFactory.getLogger("ai.grakn")).setLevel(Level.INFO);
     }
 
--- a/grakn-test/src/test/java/ai/grakn/test/GraknTestEnv.java
+++ b/grakn-test/src/test/java/ai/grakn/test/GraknTestEnv.java
@@ -6,7 +6,6 @@ import ai.grakn.engine.postprocessing.EngineCache;
 import ai.grakn.engine.util.ConfigProperties;
 import ai.grakn.factory.EngineGraknGraphFactory;
 import ai.grakn.factory.SystemKeyspace;
-import ai.grakn.test.engine.tasks.BackgroundTaskTestUtils;
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
 import com.jayway.restassured.RestAssured;
@@ -19,6 +18,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import static ai.grakn.engine.tasks.config.KafkaTerms.NEW_TASKS_TOPIC;
 import static ai.grakn.graql.Graql.var;
+import static org.slf4j.Logger.ROOT_LOGGER_NAME;
 
 /**
  * <p>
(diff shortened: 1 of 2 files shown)


aelred:
Do not merge - there are still too many logs.

grakn-test/src/test/java/ai/grakn/test/GraknTestEnv.java
```diff
@@ -146,10 +146,12 @@ static String randomKeyspace(){
...
     public static void hideLogs() {
-        ((Logger) org.slf4j.LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME)).setLevel(Level.OFF);
-        ((Logger) org.slf4j.LoggerFactory.getLogger(GraknTestEnv.class)).setLevel(Level.DEBUG);
-        ((Logger) org.slf4j.LoggerFactory.getLogger(BackgroundTaskTestUtils.class)).setLevel(Level.DEBUG);
+        ((Logger) org.slf4j.LoggerFactory.getLogger(ROOT_LOGGER_NAME)).setLevel(Level.ERROR);
+
+        // Some dependencies are using log4j, so must be managed separately
         org.apache.log4j.Logger.getRootLogger().setLevel(org.apache.log4j.Level.ERROR);
+
+        ((Logger) org.slf4j.LoggerFactory.getLogger("ai.grakn")).setLevel(Level.INFO);
```
fppt: I would say we would _not_ want to hide debug and error logs relating to `ai.grakn`

aelred: We don't hide `ERROR` logs - it will show every level `INFO` and above (so `INFO`, `WARN` and `ERROR`).

It's true we won't see `DEBUG`. In general, the `DEBUG` level seems to high - I wouldn't be surprised if travis exceeds the log limit. I can try it and see if travis can handle it.

fppt: I stand corrected about `Error` logs but maybe we would want `Debug` logs

grakn-test/src/test/java/ai/grakn/test/GraknTestEnv.java
```diff
@@ -146,10 +146,12 @@ static String randomKeyspace(){
     }
 
     public static void hideLogs() {
-        ((Logger) org.slf4j.LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME)).setLevel(Level.OFF);
-        ((Logger) org.slf4j.LoggerFactory.getLogger(GraknTestEnv.class)).setLevel(Level.DEBUG);
-        ((Logger) org.slf4j.LoggerFactory.getLogger(BackgroundTaskTestUtils.class)).setLevel(Level.DEBUG);
+        ((Logger) org.slf4j.LoggerFactory.getLogger(ROOT_LOGGER_NAME)).setLevel(Level.ERROR);
```
alexandraorth: I think we shouldn't be changing the log level programmatically at all - we should be using main and test `logback.xml` files to set these levels.

grakn-test/src/test/java/ai/grakn/test/engine/tasks/manager/StandaloneTaskManagerTest.java
```diff
@@ -142,6 +145,9 @@ public void testStopSingle() {
 
     @Test
     public void consecutiveStopStart() {
+        // Disable excessive logs for this test
+        ((Logger) LoggerFactory.getLogger(StandaloneTaskManager.class)).setLevel(Level.WARN);
```
alexandraorth: If we are going to keep doing this programmatically, can we add a handy static method so that we don't need to keep remembering the correct syntax?

grakn-test/src/test/java/ai/grakn/test/GraknTestEnv.java
fppt: This is only for the sake of testing so I am personally okay with it.

aelred: Agreed on not doing things programmatically!

grakn-test/src/test/java/ai/grakn/test/GraknTestEnv.java
aelred: I'll make another PR after this with `DEBUG` enabled.

PR #1282, opened 2017-03-15
REST server exception handling

alexandraorth:


fppt:
rebase please

alexandraorth:
please retest this please

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -193,7 +199,12 @@ public TaskManager getTaskManager(){
         return taskManager;
     }
 
-    private void checkAuthorization(Request request) {
+    /**
+     * If authorization is enabled, check the client has correct JWT Token before allowing
+     * access to specific endpoints.
+     * @param request request information from the client
+     */
+    private static void checkAuthorization(Service spark, Request request) {
```
fppt: This is *very* nitpicky but you passing in `Service spark` and you only use it when the authentication fails. It might be better for this to return a boolean and then halt the service outside of this method. 

It might make things more readable and also creates a separation of concerns between methods. Like I said VERY nitpicky.

PR #1284, opened 2017-03-15
Optimisation and un-ginore some property tests

aelred:


grakn-graph/src/test/java/ai/grakn/generator/TypeNames.java
```diff
@@ -50,14 +50,21 @@ public TypeName generateFromGraph() {
             return withImplicitConceptsVisible(graph(), graph -> {
                 TypeName name;
 
+                int attempts = 0;
                 do {
-                    name = randomName();
+                    // After a certain number of attempts, generate truly random strings instead
```
fppt: This scares me. We either dumb luck our way into something readable or we smash against a wall give up and go with something unreadable?

aelred: This is a bit weird, I agree! The issue is, there are two conflicting things happening:
1. Something wants to generate a totally unique type name that doesn't conflict with something in the graph
2. The type name should be readable

Given we only have a list of 10 "readable" names, sometimes it may not be possible to produce a readable one, so we have to fall back to a totally random string.

PR #1285, opened 2017-03-15
Hal fixes

marco-scoppetta:
- don't attach relations that connect a relation to its resources
  - limit number of resource owners, usual offset limit system

### DIFF
--- a/grakn-dist/src/bin/grakn-engine.sh
+++ b/grakn-dist/src/bin/grakn-engine.sh
@@ -48,9 +48,9 @@ start)
         # engine has not already started
         echo -n "Starting engine"
         if [[ "$FOREGROUND" = true ]]; then
-            java -cp "${CLASSPATH}" -Dgrakn.dir="${GRAKN_HOME}/bin" -Dgrakn.conf="${GRAKN_ENGINE_CONFIG}" ai.grakn.engine.GraknEngineServer
+            java -cp "${CLASSPATH}" -Dgrakn.dir="${GRAKN_HOME}/bin" -Dgrakn.conf="${GRAKN_CONFIG}" ai.grakn.engine.GraknEngineServer
         else
-            java -cp "${CLASSPATH}" -Dgrakn.dir="${GRAKN_HOME}/bin" -Dgrakn.conf="${GRAKN_ENGINE_CONFIG}" ai.grakn.engine.GraknEngineServer &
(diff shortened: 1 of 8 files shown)


grakn-engine/src/main/java/ai/grakn/engine/controller/GraphFactoryController.java
```diff
@@ -116,14 +116,14 @@ private String getKeySpaces(Request request, Response response) {
                 Collection<Resource<?>> names = keyspace.resources(keyspaceName);
                 if (names.size() != 1) {
                     throw new GraknEngineServerException(500,
-                            ErrorMessage.INVALID_SYSTEM_KEYSPACE.getMessage(" keyspace " + keyspace.getId() + " hos no unique name."));
```
fppt: Spelling is hard.

grakn-graql/src/main/java/ai/grakn/graql/internal/hal/HALConceptData.java
```diff
@@ -176,7 +176,9 @@ private void generateRuleLHS(Representation halResource, Rule rule) {
 
     private void generateOwnerInstances(Representation halResource, Resource<?> conceptResource, int separationDegree) {
         final TypeName roleType = conceptResource.type().getName();
-        conceptResource.ownerInstances().forEach(instance -> {
+        Stream<Instance> ownersStream = conceptResource.ownerInstances().stream().skip(offset);
```
fppt: So here is something that sucks. `ownerInstances` serialises everything into a set which means we still reading all the instances from disk despite skipping over a few. If we wanted to make this faster we would need to get `ownerInstances` to lazy load and if we doing that I cote for making it return a `Stream`

fppt:
please retest this please

fppt:
please retest this please

grakn-dist/src/bin/grakn-engine.sh
```diff
@@ -48,9 +48,9 @@ start)
         # engine has not already started
         echo -n "Starting engine"
         if [[ "$FOREGROUND" = true ]]; then
-            java -cp "${CLASSPATH}" -Dgrakn.dir="${GRAKN_HOME}/bin" -Dgrakn.conf="${GRAKN_ENGINE_CONFIG}" ai.grakn.engine.GraknEngineServer
+            java -cp "${CLASSPATH}" -Dgrakn.dir="${GRAKN_HOME}/bin" -Dgrakn.conf="${GRAKN_CONFIG}" ai.grakn.engine.GraknEngineServer
```
alexandraorth: We've got a separate pull request in for these startup script changes - could we remove them from this one? Just to try and keep things delineated.

PR #1288, opened 2017-03-15
Change some internal graph api methods to use streams so we can lazy load in order to speed up some simple behaviour

fppt:


grakn-graph/src/main/java/ai/grakn/graph/internal/CastingImpl.java
```diff
@@ -65,9 +65,9 @@ public RoleType getRole() {
...
      */
     public InstanceImpl getRolePlayer() {
-        Concept concept = getOutgoingNeighbour(Schema.EdgeLabel.ROLE_PLAYER);
-        if(concept != null) {
-            return (InstanceImpl) concept;
+        Optional<Concept> concept = getOutgoingNeighbours(Schema.EdgeLabel.ROLE_PLAYER).findFirst();
+        if(concept.isPresent()) {
+            return (InstanceImpl) concept.get();
         } else {
             return null;
```
aelred: This will be a fun PR for me!
You can probably simplify this with `orElse`:
```
return concept.orElse(null);
```
Also you can use the generic parameter on `getOutgoingNeighbours` to avoid the cast.

Although obviously I would personally suggest returning an `Optional` :P

fppt: Ya I realised later I would use `return concept.orElse(null);` don't know why I did not go back and change this. Will do so now. Thanks!

PR #1289, opened 2017-03-15
Optional explanation and some fixes

kasper-piskorski:


grakn-core/src/main/java/ai/grakn/graql/admin/ReasonerQuery.java
```diff
@@ -91,9 +91,10 @@
     /**
      * resolves the query
      * @param materialise materialisation flag
+     * @param explanation whether to provide explanation
      * @return stream of answers
      */
-    Stream<Answer> resolve(boolean materialise);
+    Stream<Answer> resolve(boolean materialise, boolean explanation);
```
fppt: This is really just for my own understanding: 

1. I am assuming you cannot retroactively get the explanation for something that was resolved? I.e. if something is materialised then you can't get the explanation?

2. Does computing the explanation affect performance?

grakn-test/src/test/java/ai/grakn/test/graql/reasoner/ExplanationTest.java
```diff
@@ -0,0 +1,121 @@
+package ai.grakn.test.graql.reasoner;
```
fppt: I think this is missing the copy right info. @aelred I though checkstyle looked for this? Does it skip the check for tests?

grakn-core/src/main/java/ai/grakn/graql/admin/ReasonerQuery.java
kasper-piskorski: 1. Your understanding is correct. Once materialised when asked for the explanation it will be explained on the basis of lookup retrieval.
2. Yes it does.

grakn-test/src/test/java/ai/grakn/test/graql/reasoner/ExplanationTest.java
aelred: Yeah, the checkers don't read the tests at all. I'm not totally sure it's possible to make them do that...

fppt: In that case @kasper-piskorski could you add the copyright info please?

PR #1293, opened 2017-03-15
Add test `logback.xml`

aelred:
This avoids the need to set logging levels within the code


fppt:
Rebase please

aelred:
@duckofyork assigned you because I've removed a lot of log-level setting calls in analytics tests. I'm going to work on making all grakn test logs at `DEBUG` across the board today, so you'll see the same logs as before.

PR #1300, opened 2017-03-16
Bumping Cassandra to 2.1.17

burukuru:
Re-opening this to make sure we test against latest 2.1 version.

Countless Cassandra fixes but more practically it fixes the cqlsh bug (https://issues.apache.org/jira/browse/CASSANDRA-11850).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,6 +75,7 @@
         <csv.version>1.4</csv.version>
         <jool.version>3.8.5</jool.version>
         <cassandra-unit.version>2.1.9.2</cassandra-unit.version>
+        <cassandra-all.version>2.1.17</cassandra-all.version>
         <h2-database.version>1.4.192</h2-database.version>
         <hermit.version>1.3.8.500</hermit.version>
         <owl-api.version>5.0.2</owl-api.version>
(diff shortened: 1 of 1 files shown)


fppt:
@burukuru Is this to be merged or just for testing purposes?

burukuru:
This is for merging purposes.
I want to deploy this to servers so need this tested against.

bolerio:
Shouldn't we first test somewhere (cluster, unstable etc.) before merging in master. If we want to push a release soon, I'd be uncomfortable bumping Cassandra version without properly testing it.

burukuru:
I can update the cluster on Monday and test there for a couple of weeks.
I think testing on the unstable branch would be good as well, it'll be much more diverse in terms of use cases.

fppt:
Is this still a valid PR @burukuru ? Have you tested on the cluster? If so it needs a rebase.

burukuru:
It is still valid. I will update the cluster today but it might take a week or so to put it through its paces. Shall I close and re-open the PR then?

bolerio:
Yes, probably best to test with latest code as well. And if we can switch easily between versions of Cassandra, and make it possible for customers to do so, that's even better!

burukuru:
I upgraded the cluster earlier today, ran some match/compute queries and things seemed to be working fine. I'll leave it until next week so it gets tested more and will do a rebase.

Compiling the code with another version of cassandra should be a matter of changing the dependency version. For bundling multiple versions, I would recommend creating an archive per cassandra version.

bolerio:
Thanks, Thanh.

I would actually expect  that we should be able to switch Cassandra versions at deployment time!

burukuru:
retest this please

PR #1302, opened 2017-03-16
Fix the bug that may cause failure when running concurrent analytics jobs

duckofyork:
Add several tests to make sure analytics jobs can be carried out concurrently

### DIFF
--- a/grakn-test/src/test/java/ai/grakn/test/graql/analytics/AnalyticsTest.java
+++ b/grakn-test/src/test/java/ai/grakn/test/graql/analytics/AnalyticsTest.java
@@ -38,6 +38,8 @@ import org.junit.Before;
 import org.junit.ClassRule;
 import org.junit.Test;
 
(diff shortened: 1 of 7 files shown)


grakn-test/src/test/java/ai/grakn/test/graql/analytics/AnalyticsTest.java
```diff
@@ -127,4 +140,52 @@ public void testNullResourceDoesntBreakAnalytics() throws GraknValidationExcepti
             fail();
         }
     }
+
+    @Test
+    public void testConcurrentJobs() {
```
fppt: Can we get a more descriptive name? Imagine I broke this test and I need to fix it without bothering you? 
Most of us are following the paradigm of describing the main action and the expected result. For example `whenCallingGetConceptWithAnExistingConceptId_ItReturnsThatConcept`

PR #1303, opened 2017-03-16
Kill compute queries when exiting Graql shell

aelred:


grakn-engine/src/main/java/ai/grakn/engine/session/GraqlSession.java
```diff
@@ -82,6 +84,7 @@
     // All requests are run within a single thread, so they always happen in a single thread-bound transaction
     private final ExecutorService queryExecutor =
             Executors.newSingleThreadExecutor(new ThreadFactoryBuilder().setNameFormat("graql-session-%s").build());
+    private List<Query<?>> queries = null;
```
alexandraorth: Is this variable used?

PR #1307, opened 2017-03-17
Node labels

marco-scoppetta:
- store node label preferences in localstorage
- default label on nodes type:id


fppt:
Only JS changes. Needed urgently. Merging now without waiting for tests.

fppt:
Nevermind

PR #1308, opened 2017-03-17
Default labels

marco-scoppetta:
-default label to type:id
-fix nodes after align when all nodes unblocked


fppt:
Only JS changes. Needed urgently. Merging now without waiting for tests. (For real this time)

PR #1311, opened 2017-03-17
Node labels changes

marco-scoppetta:
- now the node labels preferences are stored in localStorage
      - 'type' can be selected as a node label


fppt:
Again only JS changes. Needed out of urgency. Merging regardless of tests not being complete yet.

PR #1323, opened 2017-03-20
Make `Var` class totally immutable

aelred:
This is for a variety of reasons:
- Easier to reason about behaviour when sharing variables or queries
- `MatchQuery` and `Pattern` classes are already immutable
- We don't have to clone queries anymore
- Prevents bug when performing clones/sharing Graql objects

### DIFF
--- a/grakn-core/src/main/java/ai/grakn/graql/Var.java
+++ b/grakn-core/src/main/java/ai/grakn/graql/Var.java
@@ -23,6 +23,8 @@ import ai.grakn.concept.ResourceType;
 import ai.grakn.concept.TypeName;
 import ai.grakn.graql.admin.VarAdmin;
 
+import javax.annotation.CheckReturnValue;
+
 /**
  * A wildcard variable to refers to a concept in a query.
  * <p>
@@ -43,36 +45,42 @@ public interface Var extends Pattern {
      * @param id a ConceptId that this variable's ID must match
      * @return this
      */
+    @CheckReturnValue
     Var id(ConceptId id);
 
     /**
      * @param name a string that this variable's name must match
      * @return this
      */
+    @CheckReturnValue
     Var name(String name);
 
     /**
      * @param name a type name that this variable's name must match
      * @return this
      */
+    @CheckReturnValue
     Var name(TypeName name);
 
     /**
      * @param value a value that this variable's value must exactly match
      * @return this
      */
+    @CheckReturnValue
     Var value(Object value);
 
     /**
      * @param predicate a atom this variable's value must match
      * @return this
      */
+    @CheckReturnValue
     Var value(ValuePredicate predicate);
 
     /**
      * @param var a variable representing a resource
      * @return this
      */
+    @CheckReturnValue
     Var has(Var var);
 
     /**
@@ -82,6 +90,7 @@ public interface Var extends Pattern {
      * @param value a value of a resource
      * @return this
      */
+    @CheckReturnValue
     Var has(String type, Object value);
 
     /**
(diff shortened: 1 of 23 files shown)


grakn-core/src/main/java/ai/grakn/graql/Var.java
```diff
@@ -109,84 +120,98 @@
...
      * @return this
      */
+    @CheckReturnValue
     Var hasResource(Var type);
 
     /**
      * @param type a resource type that this type variable can be one-to-one related to
      * @return this
      */
+    @CheckReturnValue
```
fppt: I am not familiar with this annotation. What does it do?

grakn-engine/src/main/java/ai/grakn/engine/user/SystemKeyspaceUsers.java
```diff
@@ -55,22 +55,22 @@
      */
     @Override
     public boolean addUser(Json userJson) {
-        final Var user = var().isa(USER_ENTITY);
+        final Var[] user = {var().isa(USER_ENTITY)};
```
fppt: Uhm why change to an array if you only ever going to use the first element of the array?

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/Utility.java
```diff
@@ -346,15 +346,15 @@ public static Rule createSubPropertyRule(RelationType parent, RelationType child
...
-        Var parentVar = var().isa(name(parent.getName()));
-        Var childVar = var().isa(name(child.getName()));
+        final Var[] parentVar = {var().isa(name(parent.getName()))};
+        final Var[] childVar = {var().isa(name(child.getName()))};
 
         roleMappings.forEach( (parentRoleName, childRoleName) -> {
             VarName varName = VarName.anon();
-            parentVar.rel(name(parentRoleName), var(varName));
-            childVar.rel(name(childRoleName), var(varName));
+            parentVar[0] = parentVar[0].rel(name(parentRoleName), var(varName));
```
fppt: Again, why all the arrays?

grakn-core/src/main/java/ai/grakn/graql/Var.java
aelred: It's a suggestion that findbugs recognises, stating that a caller to this function should use the return value for something.

Because `Var` is immutable, calling these functions will never have any side-effects, so if a caller does not use the return value, then they have made a mistake. This helped findbugs catch several places in our code where we assumed `Var` was mutable!

fppt: Nice

grakn-engine/src/main/java/ai/grakn/engine/user/SystemKeyspaceUsers.java
aelred: Because Java lambdas are crap! You can't assign to variables in them, but you _can_ assign to the elements of an array, so you can make a single-element array instead.

In a few cases, I've instead replaced the lambda with a `for` loop. We could also replace this with a `for` loop, but I chose not to because we're iterating over a map, which is less readable using a `for` loop (because you have to go `Map.Entry<Blah, Blah>...key...value`).

Ditto for the below comment as well.

fppt: I thought it was because of that. I actually find this even more unreadable. I would prefer to just use the traditional `for` and `entry` rather than to start creating these things everywhere. I think this will set a bad example for the future.

That is my opinion though. I would ask you to change that but if you feel strongly about it then feel free to merge.

aelred: No, I don't feel strongly about it. I'll change them to regular `for` loops.

PR #1327, opened 2017-03-20
Engine external offset storage

alexandraorth:


alexandraorth:
please retest this please

alexandraorth:
please retest this please

grakn-engine/src/main/java/ai/grakn/engine/tasks/ExternalOffsetStorage.java
```diff
@@ -0,0 +1,102 @@
...
+    public long getOffset(TopicPartition partition) {
+        try {
+            String partitionPath = getPartitionPath(partition);
+            long offset = (long) deserialize(zookeeper.connection().getData().forPath(partitionPath));
+
+            LOG.debug("Offset {} read for partition %{}", partitionPath, partitionPath);
+
+            return offset;
+        } catch (RuntimeException e) {
+            throw e;
```
fppt: What's the point of catching the runtime exception just to rethrow it? Maybe you meant to wrap it?

grakn-engine/src/main/java/ai/grakn/engine/tasks/ExternalOffsetStorage.java
```diff
@@ -0,0 +1,102 @@
...
+            try {
+                zookeeper.connection().setData()
+                        .forPath(partitionPath, serialize(currentOffset));
+            } catch(KeeperException.NoNodeException e) {
+                zookeeper.connection().create()
+                        .creatingParentContainersIfNeeded()
+                        .forPath(partitionPath, serialize(currentOffset));
+            }
+        } catch (RuntimeException e) {
+            throw e;
```
fppt: DItto

grakn-test/src/test/java/ai/grakn/test/engine/tasks/ExternalOffsetStorageTest.java
```diff
@@ -0,0 +1,103 @@
...
+        offsetStorage = new ExternalOffsetStorage(zookeeper);
+    }
+
+    @AfterClass
+    public static void shutdownZKConnection(){
+        zookeeper.close();
+    }
+
+    @Test
+    public void whenSavingNewConsumerOffset_CanGetOffset() throws Exception{
```
fppt: +1 on this style of naming tests `Action_ExpectedResult` :)

grakn-engine/src/main/java/ai/grakn/engine/tasks/ExternalOffsetStorage.java
bolerio: @fppt, probably @alexandraorth knows best, but from what I see there is an EXception catch after that, so the idea is probably *not* to wrap an unchecked exception, but wrapped the checked ones. I don't know what exceptions are expected here, but I agree it's probably best to wrap all of them in EngineStorageException

alexandraorth: I'm going to have to refer this question to @aelred . This was added in to correct something for either PMD or Findbugs

alexandraorth: In any case, I've just added another commit to wrap it.

aelred: Yes, findbugs found it for exactly the reason @bolerio said. The reason is to avoid accidentally wrapping, say, a `NullPointerException`.

PR #1331, opened 2017-03-21
Fix failing test due to self shortcut edges

fppt:


fppt:
Failure due to: StatisticsTest.testStd - please retest this please.

PR #1334, opened 2017-03-21
make sure duplicate or equivalent rules not processed twice

kasper-piskorski:


grakn-test/src/test/java/ai/grakn/test/graql/reasoner/ReasonerTest.java
```diff
@@ -160,6 +162,30 @@ public void testPropertyChainRule() {
     }
 
     @Test
+    public void testRuleEquivalence() {
```
fppt: Can we try to get a better description for this test? Again, imagine I break it and I need to figure out how to fix it. For example `public void whenAddingRulesWithTheSamePatternsEnsureTheyAreEqual` tells me much more than just `testRuleEquivalence `

PR #1337, opened 2017-03-21
Require rule heads to specify all role types

kasper-piskorski:
- require rule heads to specify roles
- simplified and fixed finding mappings between relation players


grakn-test/src/test/java/ai/grakn/test/graql/reasoner/ReasonerTest.java
```diff
@@ -171,6 +186,15 @@ public void testPropertyChainRuleCreation() {
     }
 
     @Test
+    public void testAddingRuleWithHeadWithoutRolePlayersNotAllowed() {
```
aelred: I think this is `WithoutRoleTypes` right?

grakn-test/src/test/java/ai/grakn/test/graql/reasoner/AtomicQueryTest.java
```diff
@@ -102,6 +102,23 @@ public void testCopyConstructor(){
     }
 
     @Test
+    public void testCopyConstructor2(){
```
aelred: Maybe this could be two tests? `whenModifyingAQuery_TheCopyDoesNotChange` and `whenCopyingAQuery_TheyHaveTheSameRoleVarTypeMaps`

grakn-test/src/test/java/ai/grakn/test/graql/reasoner/ReasonerTest.java
kasper-piskorski: true

PR #1340, opened 2017-03-22
Engine failover relies on current ZK state

alexandraorth:
+ Engine Failover was previously missing tasks if it was down while other engines died. Now, when an Engine dies we get all running tasks from storage, and if they are running on a machine that is no longer up, resubmit them. 
+ SQTM will register itself for failover

### DIFF
--- a/grakn-test/src/test/java/ai/grakn/test/engine/tasks/manager/singlequeue/FailoverElectorTest.java
+++ b/grakn-test/src/test/java/ai/grakn/test/engine/tasks/manager/singlequeue/FailoverElectorTest.java
@@ -55,20 +55,26 @@ public class FailoverElectorTest {
 
     @Test
     public void whenFailoverElectorIsInstantiated_ThisEngineBecomesLeader(){
-        FailoverElector elector = new FailoverElector(EngineID.of("Engine1"), zookeeper, storage);
-        assertEquals("Engine1", elector.awaitLeader());
+        EngineID engineId = EngineID.of("Engine1");
+        FailoverElector elector = new FailoverElector(engineId, zookeeper, storage);
+        assertEquals(engineId, elector.awaitLeader());
         elector.renounce();
     }
 
     @Test
     public void whenFailoverElectorIsKilled_AnotherFailoverElectorBecomesLeader() throws Exception {
-        FailoverElector elector1 = new FailoverElector(EngineID.of("Engine1"), zookeeper, storage);
-        FailoverElector elector2 = new FailoverElector(EngineID.of("Engine2"), zookeeper, storage);
+        EngineID engine1 = EngineID.of("Engine1");
+        EngineID engine2 = EngineID.of("Engine2");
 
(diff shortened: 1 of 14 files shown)


grakn-test/src/test/java/ai/grakn/test/engine/tasks/manager/singlequeue/FailoverElectorTest.java
```diff
@@ -55,31 +55,49 @@ public void shutdownZkConnection() {
...
             elector1.renounce();
         } else {
             elector2.renounce();
         }
 
         Thread.sleep(1000);
 
-        assertEquals(elector1.awaitLeader(), elector2.awaitLeader());
-        assertNotEquals(currentLeader, elector1.awaitLeader());
+        System.out.println("whale");
```
aelred: 🐳

grakn-test/src/test/java/ai/grakn/test/engine/tasks/manager/singlequeue/FailoverElectorTest.java
```diff
@@ -55,31 +55,49 @@ public void shutdownZkConnection() {
...
+        EngineID currentLeader;
+        if(Objects.equals(leader1, engine1)) {
+            currentLeader = elector2.awaitLeader();
+        } else {
+            currentLeader = elector1.awaitLeader();
+        }
+
+        System.out.println("starfish");
```
aelred: ⭐️

grakn-test/src/test/resources/logback.xml
```diff
@@ -0,0 +1,36 @@
...
+    </root>
+
+    <logger name="ai.grakn" level="DEBUG" additivity="false">
+        <appender-ref ref="STDOUT"/>
+    </logger>
+
+    <logger name="ai.grakn.engine.GraknEngineServer" level="WARN">
+        <appender-ref ref="STDOUT"/>
+    </logger>
+</configuration>
```
aelred: Eh, this is fine I guess. I don't know what the real way to have this file is. Can we have symbolic links in git repos?

grakn-test/src/test/java/ai/grakn/test/engine/tasks/manager/singlequeue/FailoverElectorTest.java
alexandraorth: Oops :P

grakn-test/src/test/java/ai/grakn/test/engine/tasks/manager/singlequeue/FailoverElectorTest.java
alexandraorth: Yeah.... I'll remove those in the next PR

grakn-test/src/test/resources/logback.xml
alexandraorth: This was actually accidental :P Let me look into the symlink thing

PR #1347, opened 2017-03-22
Remove soft references in `TaskStateInMemoryStore`

alexandraorth:


alexandraorth:
please retest this please

grakn-engine/src/main/java/ai/grakn/engine/tasks/storage/TaskStateInMemoryStore.java
```diff
@@ -42,27 +41,27 @@
  * @author Denis Lobanov, alexandraorth
  */
 public class TaskStateInMemoryStore implements TaskStateStorage {
-    private final Map<TaskId, SoftReference<TaskState>> storage;
+    private final Map<TaskId, TaskState> storage;
```
fppt: Huge +1. Soft references have always led to the worst debugging challenges. Hate those things.

PR #1351, opened 2017-03-23
Fix bug with Graql output of relations with multiple role-payers

aelred:


aelred:
please retest this please - failed due to a known analytics failure

PR #1352, opened 2017-03-23
Daily Merge From Master

fppt:


fppt:
As always do NOT squash when merging this.

PR #1354, opened 2017-03-23
Support for the DataType Date in our graph. 

fppt:
- Note: The date is persisted as a string, so `order by` and `sort` operations on traversals will work on a string.


grakn-core/src/main/java/ai/grakn/concept/ResourceType.java
```diff
@@ -169,12 +175,53 @@
...
+                LocalDateTime.class.getName(),
+                Schema.ConceptProperty.VALUE_DATE,
+                (d) -> d.atZone(ZoneId.systemDefault()).toInstant().toEpochMilli(),
+                (o) -> {
+                    if (o == null) return null;
+                    if (!(o instanceof Long)) {
+                        throw new InvalidConceptValueException(ErrorMessage.INVALID_DATATYPE.getMessage(o, Long.class.getName()));
+                    }
+                    return LocalDateTime.ofInstant(Instant.ofEpochMilli((long) o), ZoneId.systemDefault());
+                });
```
aelred: At this point, I would make these classes implementing an interface:

```
private static DateType implements DataType<LocalDateTime> {
    @Override
    public D getValue(Object o) {
        if (o == null) return null;
        if (!(o instanceof Long)) {
            throw new InvalidConceptValueException(ErrorMessage.INVALID_DATATYPE.getMessage(o, Long.class.getName()));
        }
        return LocalDateTime.ofInstant(Instant.ofEpochMilli((long) o), ZoneId.systemDefault());
    }
    ... etc ...
}
```
You can also give the interface a default implementation for the converter functions, to save restating it.

fppt: Can we make that it another PR? I just want to get the functionality in.

aelred: Yeah, I have no problem with that.

PR #1361, opened 2017-03-24
Add explicit path function to answer class

kasper-piskorski:


kasper-piskorski:
please retest this please

PR #1366, opened 2017-03-27
Fix concurrency problems of analytics, add separate concurrency tests

duckofyork:


grakn-graph/src/main/java/ai/grakn/graph/internal/computer/GraknSparkComputer.java
```diff
@@ -156,131 +159,126 @@ public void cancelJobs() {
...
+                    apacheConfiguration, this.resultGraph, this.persist), finalMemory.asImmutable());
         }, exec);
     }
 
     /////////////////
 
     private static void loadJars(final JavaSparkContext sparkContext,
                                  final Configuration hadoopConfiguration,
-                                 Logger logger) {
+                                 Logger LOGGER) {
```
aelred: WHY ARE WE SHOUTING

PR #1367, opened 2017-03-27
Throw exception when opening an already opened graph on the same thread.

fppt:


grakn-graph/src/main/java/ai/grakn/factory/GraknGraphFactoryImpl.java
```diff
@@ -124,14 +124,14 @@ private void checkClosure(int numOpenTransactions, Supplier<GraknGraph> graphSup
 
     @Override
     public int openGraphTxs() {
-        if(!graphOpen) return 0;
-        return ((AbstractGraknGraph)getGraph()).numOpenTx();
+        if(graph == null) return 0;
+        return ((AbstractGraknGraph) graph).numOpenTx();
```
aelred: Will this always be 1 now?

grakn-test/src/test/java/ai/grakn/test/migration/export/GraphWriterMainTest.java
```diff
@@ -33,20 +32,27 @@
...
     @BeforeClass
     public static void loadMovieGraph() {
         graph =  engineContext.factoryWithNewKeyspace().getGraph();
     }
 
+    @Before
+    public void setup(){
+        keyspace = graph.getKeyspace();
+        graph.close();
+    }
```
aelred: This looks weird to me: we open the graph only once (with `@BeforeClass`), but then we close that same graph before every test?

fppt: This is to get a hold of a generated keyspace.

grakn-graph/src/main/java/ai/grakn/factory/GraknGraphFactoryImpl.java
fppt: No

fppt:
please retest this please - ai.grakn.test.engine.GraknEngineFailoverIT.whenSubmittingTasksToOneEngineAndRandomlyKillingTheOthers_TheyComplete

fppt:
please retest this please - ai.grakn.test.engine.GraknEngineFailoverIT.whenSubmittingTasksToOneEngineAndRandomlyKillingTheOthers_TheyComplete

fppt:
@aelred could I get another review on this please?

PR #1370, opened 2017-03-27
Graql-native explanation

kasper-piskorski:
- removed Reasoner endpoint for explanations
- explanations provided via MatchQueryInfer with all query modifiers applicable


kasper-piskorski:
please retest this please

PR #1373, opened 2017-03-28
Add property tests for `Type`

aelred:
Also included a refactor to the generators, mainly:
- The graph generator now produces a string describing the graph, to help with debugging failures
- Change `@NotMeta` to `@Meta`

The tests are not quite complete, we still need to:
- Create graphs with rules linked to types
- Write tests for `resource` and `key` :(


aelred:
@fppt I think there's only one test here you need to worry about: `whenATypeWithInstancesIsSetToAbstract_Throw`.

grakn-core/src/main/java/ai/grakn/concept/Type.java
```diff
@@ -41,49 +43,58 @@
  */
 public interface Type extends Concept {
     //------------------------------------- Modifiers ----------------------------------
+    // TODO: Describe behaviour when setting a type with direct instances as abstract
```
fppt: This throws an exception on commit no?

PR #1374, opened 2017-03-28
Ignore a frequently failing test

aelred:


fppt:
pushing in as these tests are slowing down PRs

PR #1375, opened 2017-03-28
Revert "switch reasoner to on by default"

fppt:
Reverts graknlabs/grakn#1371


fppt:
Ahhh nevermind. This PR pulls out too much!

PR #1378, opened 2017-03-28
Fix bug in post processing where relations were not being copies over correctly.

fppt:


fppt:
please retest this please - no test failures

fppt:
please retest this please - no test failures

PR #1379, opened 2017-03-29
More Changes to Transactions and Factories

fppt:
- Refactor GraknFactory to GraknSession
- Get all graphs via open()
- Finalise API on Session and GraknGraph

TODO: 

The read only graph behaviour and the transaction count cleanup on session. I will do this on a different PR. If we want to release with these changes in they should go in soon.


grakn-core/src/main/java/ai/grakn/GraknTransaction.java
```diff
@@ -0,0 +1,40 @@
...
+ *     When producing a graph using a {@link GraknSession} one of the following enums must be provided:
+ *         READ - A read only transaction. If you attempt to mutate the graph with such a transaction an exception will be thrown.
+ *         WRITE - A transaction which allows you to mutate the graph.
+ *         BATCH - A transaction which allows mutations to be performed more quickly but disables some consitency checks.
+ * </p>
+ *
+  @author fppt
+ */
+public enum GraknTransaction {
+    READ,  //Read only transaction where mutations to the graph are prohibited
```
bolerio: I would rename this to GraknTransactionType - there is such a strong expectation for the name GraknTransaction to actually represent a transaction and not a configuration options :)

grakn-core/src/main/java/ai/grakn/Grakn.java
```diff
@@ -124,9 +124,9 @@ WHY you would want a toy graph (e.g. what would you be doing or not doing to use
      * If one doesn't exist, it will be created for you.
      * @return A factory instance that can produce concurrent connection to the specified knowledge graph.
      */
-    public static GraknGraphFactory factory(String location, String keyspace) {
+    public static GraknSession factory(String location, String keyspace) {
```
bolerio: The name 'factory' for that method is kind of obsolete, no?

grakn-core/src/main/java/ai/grakn/GraknTransaction.java
fppt: I think I called it that initially for the same reasons but then went for this shorter option because the length annoyed me when you say `session.open(GraknTransactionType.WRITE)`

I will change it back. I can get over my hatred of the length.

grakn-core/src/main/java/ai/grakn/Grakn.java
fppt: Yes correct. Javadoc needs to change. Good spot.

grakn-core/src/main/java/ai/grakn/GraknTransaction.java
bolerio: Don't if it would make to scope it somewhere inside another class, so that it's just TxType for example. I like short names for very common things - yes they take time to remember, but they are common...

grakn-core/src/main/java/ai/grakn/Grakn.java
fppt: Ooo the method name as well. I am so blind.

grakn-core/src/main/java/ai/grakn/GraknTransaction.java
fppt: Maybe we can compromise with `GraknTxType` ?

grakn-engine/src/main/java/ai/grakn/engine/session/GraqlSession.java
```diff
@@ -87,7 +88,7 @@
     private List<Query<?>> queries = null;
 
     GraqlSession(
-            Session session, GraknGraphFactory factory, String outputFormat,
+            Session session, GraknSession factory, String outputFormat,
```
bolerio: Again, I don't think you should do it right now, better to get the PR in and find out the actual problems. But eventually that 'factory' nomenclature will have to be changed...

fppt: Yes I think we can reserve that specific change for later. Even the package names should change and I think that should be reserved for after the release

grakn-core/src/main/java/ai/grakn/util/ErrorMessage.java
```diff
@@ -54,7 +54,7 @@
     ROLE_TYPE_ERROR("The role type [%s] cannot play itself"),
     BACKEND_EXCEPTION("Unknown Backend Exception."),
     GRAPH_CLOSED("The Graph for keyspace [%s] is closed"),
-    GRAPH_PERMANENTLY_CLOSED("The Graph for keyspace [%s] is closed. Use the factory to get a new graph."),
+    GRAPH_CLOSED_ON_ACTION("The transaction was %s and closed graph [%s]. Use the session to get a new trsnaction for the graph."),
```
fppt: oops

grakn-core/src/main/java/ai/grakn/GraknGraph.java
```diff
@@ -369,17 +368,21 @@
...
+    void abort();
 
-    // TODO: what does this do when the graph is closed?
     /**
-     * Will cause the current transaction to be committed when closing the transaction.
+     * Commits any changes to the graph and closes the transaction. You must use the {@link GraknSession} to
+     * get a new open transaction.
      */
-    void commitOnClose();
+    void commit();
```
fppt: correct. Will fix in next PR.

PR #1382, opened 2017-03-29
Refactor Graql EquivalentFragmentSets

aelred:
Add subclasses for every kind of equivalent fragment set in Graql. This is necessary so that we can recognise certain patterns of fragment sets in the queries and optimise them.


aelred:
please retest this please - due to `SingleQueueTaskManagerTest`

grakn-core/src/main/java/ai/grakn/util/Schema.java
```diff
@@ -263,4 +263,14 @@ public TypeName getName() {
...
         }
     }
+
+    /**
+     *
+     * @param typeName The resource type name
+     * @param value The value of the resource
+     * @return A unique id for the resource
+     */
+    public static String generateResourceIndex(TypeName typeName, String value){
```
fppt: I am starting to think we dumping too much in `Schema`. Maybe we should create a `CoreUtils` ? 

In my view `Schema` should only contain our hardcoded enums and configs.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/sets/CastingFragmentSet.java
```diff
@@ -0,0 +1,36 @@
...
+package ai.grakn.graql.internal.gremlin.sets;
+
+import ai.grakn.graql.VarName;
+import ai.grakn.graql.internal.gremlin.EquivalentFragmentSet;
+
+import static ai.grakn.graql.internal.gremlin.fragment.Fragments.inCasting;
+import static ai.grakn.graql.internal.gremlin.fragment.Fragments.outCasting;
+
+/**
+ * @author Felix Chapman
```
fppt: -1 for bad javadocs

aelred: Yeah, I was torn on this, too... my reasoning was that these are package-private. The `public` factory methods to create these are in `EquivalentFragmentSets` - and that _is_ JavaDoc'd.

grakn-core/src/main/java/ai/grakn/util/Schema.java
aelred: I agree. I don't much like `CoreUtils` though - I'll think about it!

fppt: Bad name I admit but some place where we can dump these methods used throughout the modules would be handy.

PR #1386, opened 2017-03-30
Make `GraknGraphs` generator produce more complex type hierarchies

aelred:
This makes the tests more thorough and also prevents issues where some tests fail because they cannot generate graphs that satisfy their assumptions (such as a test that requires a type that is the sub-type of a non-meta type).

### DIFF
--- a/grakn-graph/src/test/java/ai/grakn/generator/GraknGraphs.java
+++ b/grakn-graph/src/test/java/ai/grakn/generator/GraknGraphs.java
@@ -147,30 +147,40 @@ public class GraknGraphs extends AbstractGenerator<GraknGraph> implements Minima
     private final ImmutableList<Runnable> mutators = ImmutableList.of(
             () -> {
                 TypeName typeName = typeName();
-                EntityType entityType = graph.putEntityType(typeName);
+                EntityType superType = entityType();
+                EntityType entityType = graph.putEntityType(typeName).superType(superType);
                 summaryAssign(entityType, "graph", "putEntityType", typeName);
+                summary(entityType, "superType", superType);
             },
             () -> {
                 TypeName typeName = typeName();
                 ResourceType.DataType dataType = gen(ResourceType.DataType.class);
-                ResourceType resourceType = graph.putResourceType(typeName, dataType);
+                ResourceType superType = resourceType();
+                ResourceType resourceType = graph.putResourceType(typeName, dataType).superType(superType);
                 summaryAssign(resourceType, "graph", "putResourceType", typeName, dataType);
+                summary(resourceType, "superType", superType);
             },
             () -> {
                 TypeName typeName = typeName();
                 ResourceType.DataType dataType = gen(ResourceType.DataType.class);
-                ResourceType resourceType = graph.putResourceTypeUnique(typeName, dataType);
+                ResourceType superType = resourceType();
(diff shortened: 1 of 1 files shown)


grakn-graph/src/test/java/ai/grakn/generator/GraknGraphs.java
```diff
@@ -147,30 +147,40 @@ public GraknGraphs setOpen(boolean open) {
     private final ImmutableList<Runnable> mutators = ImmutableList.of(
             () -> {
                 TypeName typeName = typeName();
-                EntityType entityType = graph.putEntityType(typeName);
+                EntityType superType = entityType();
+                EntityType entityType = graph.putEntityType(typeName).superType(superType);
```
fppt: This can fail if by some miracle `typeName == superType.getName()`. Also fails if meta type are involved. Actually what happens if this does fail?

grakn-graph/src/test/java/ai/grakn/generator/GraknGraphs.java
```diff
@@ -147,30 +147,40 @@ public GraknGraphs setOpen(boolean open) {
...
                 summaryAssign(entityType, "graph", "putEntityType", typeName);
+                summary(entityType, "superType", superType);
             },
             () -> {
                 TypeName typeName = typeName();
                 ResourceType.DataType dataType = gen(ResourceType.DataType.class);
-                ResourceType resourceType = graph.putResourceType(typeName, dataType);
+                ResourceType superType = resourceType();
+                ResourceType resourceType = graph.putResourceType(typeName, dataType).superType(superType);
```
fppt: Can fail for the same reason as above.

grakn-graph/src/test/java/ai/grakn/generator/GraknGraphs.java
```diff
@@ -147,30 +147,40 @@ public GraknGraphs setOpen(boolean open) {
...
                 summaryAssign(resourceType, "graph", "putResourceType", typeName, dataType);
+                summary(resourceType, "superType", superType);
             },
             () -> {
                 TypeName typeName = typeName();
                 ResourceType.DataType dataType = gen(ResourceType.DataType.class);
-                ResourceType resourceType = graph.putResourceTypeUnique(typeName, dataType);
+                ResourceType superType = resourceType();
+                ResourceType resourceType = graph.putResourceTypeUnique(typeName, dataType).superType(superType);
```
fppt: `putResourceTypeUnique ` still exists???? I think we can kill that especially since graql does not use it.

grakn-graph/src/test/java/ai/grakn/generator/GraknGraphs.java
aelred: You're totally right - all of these methods can fail. I have a `try...catch` wrapping these at line 92 that catches expected exception types. If there's a failure, it will try another random mutation.

I considered checking the arguments ahead of time, but it felt like replicating the logic and didn't seem worth it - given the result is the same.

grakn-graph/src/test/java/ai/grakn/generator/GraknGraphs.java
aelred: Awesome! I can delete a lot of tests too.

PR #1391, opened 2017-03-30
Add cache of query plans

aelred:


aelred:
Closing this because it looks like it might not be what we want :(

PR #1392, opened 2017-03-30
Read only graph support

fppt:


fppt:
please retest this please - ai.grakn.test.engine.GraknEngineServerTest.whenEnginePropertiesDoesNotIndicateZookeeperStorage_GraphStorageIsUsed

fppt:
please retest this please - ai.grakn.test.engine.GraknEngineServerTest.whenEnginePropertiesDoesNotIndicateZookeeperStorage_GraphStorageIsUsed

grakn-engine/src/main/java/ai/grakn/engine/controller/VisualiserController.java
```diff
@@ -109,7 +110,7 @@ public VisualiserController(Service spark) {
     private String conceptById(Request req, Response res) {
         String keyspace = getKeyspace(req);
 
-        try (GraknGraph graph = getInstance().getGraph(keyspace)) {
+        try (GraknGraph graph = getInstance().getGraph(keyspace, GraknTxType.WRITE)) {
```
aelred: This one (and others) can probably be read-only.  This may be something for another PR though.

grakn-core/src/main/java/ai/grakn/GraknGraph.java
```diff
@@ -382,7 +390,10 @@
     /**
      * Commits any changes to the graph and closes the transaction. You must use the {@link GraknSession} to
      * get a new open transaction.
+     *
+     * @throws GraknValidationException when the transaction contains graph mutations which does not form to the Grakn
```
fppt: correct. Typo

grakn-engine/src/main/java/ai/grakn/engine/controller/VisualiserController.java
fppt: Another PR. Right now there is not performance benefit of read only graphs

PR #1393, opened 2017-03-30
Add property tests for `EntityType`

aelred:


grakn-core/src/main/java/ai/grakn/concept/EntityType.java
```diff
@@ -77,10 +85,12 @@
     EntityType deletePlaysRole(RoleType roleType);
 
     /**
-     * Creates and returns a new Entity instance.
+     * Creates and returns a new Entity instance, whose direct type will be this instance.
```
aelred: "... will be this *type*" - silly me.

grakn-graph/src/test/java/ai/grakn/graph/property/EntityTypePropertyTest.java
```diff
@@ -0,0 +1,163 @@
...
+        assumeThat(type.getRulesOfHypothesis(), empty());
+        assumeThat(type.getRulesOfConclusion(), empty());
+
+        type.delete();
+
+        assertNull(graph.getType(type.getName()));
+    }
+
+    @Property
+    public void whenAddingAnEntityOfAMetaType_Throw(@Meta EntityType type) {
```
fppt: Does this take into account the rule meta types `MetaSchema.INFERENCE_RULE` and `MetaSchema.CONSTRAINT_RULE` ? You allowed to add instances to those.

aelred: Actually this is just a badly-named test. It only tests for the meta `EntityType` (the only one with an `addEntity()` method).

fppt: New request then. Rename test please :)

PR #1394, opened 2017-03-30
Get rid of putResoureceTypeUnique

fppt:
This method is completely unused outside of tests and since we have changed our definition of key/uniques/resources etc . . . 

This method no longer fits the requirement.

### DIFF
--- a/grakn-graph/src/test/java/ai/grakn/graph/property/GraknGraphPutPropertyTest.java
+++ b/grakn-graph/src/test/java/ai/grakn/graph/property/GraknGraphPutPropertyTest.java
@@ -29,11 +29,9 @@ import ai.grakn.concept.Type;
 import ai.grakn.concept.TypeName;
 import ai.grakn.exception.ConceptException;
 import ai.grakn.exception.ConceptNotUniqueException;
-import ai.grakn.exception.InvalidConceptValueException;
 import ai.grakn.generator.FromGraphGenerator.FromGraph;
 import ai.grakn.generator.GraknGraphs.Open;
 import ai.grakn.generator.PutTypeFunctions;
-import ai.grakn.generator.ResourceTypes.Unique;
 import ai.grakn.generator.TypeNames.Unused;
 import ai.grakn.util.ErrorMessage;
 import ai.grakn.util.Schema;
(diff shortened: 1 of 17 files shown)


grakn-graph/src/test/java/ai/grakn/graph/property/GraknGraphPutPropertyTest.java
```diff
@@ -125,13 +122,12 @@ public void whenCallingPutResourceType_CreateATypeWithDefaultProperties(
...
 
         assertEquals("The data-type should be as specified", dataType, resourceType.getDataType());
-        assertFalse("The resource type should not be unique", resourceType.isUnique());
         assertNull("The resource type should have no regex constraint", resourceType.getRegex());
     }
 
     @Property
     public void whenCallingPutResourceTypeWithThePropertiesOfAnExistingNonUniqueResourceType_ItReturnsThatType(
-            @Open GraknGraph graph, @FromGraph @Unique(false) ResourceType<?> resourceType) {
+            @Open GraknGraph graph, @FromGraph  ResourceType<?> resourceType) {
```
aelred: The words `NonUnique` or `Unique` can be removed from all these test names.

grakn-graph/src/test/java/ai/grakn/graph/property/RelationPropertyTest.java
```diff
@@ -48,41 +44,6 @@
...
-        assumeFalse(rolePlayer.isResource() && rolePlayer.asResource().type().isUnique());
-
-        Instance[] rolePlayers = relation.rolePlayers(roleType).toArray(new Instance[0]);
-
-        relation.addRolePlayer(roleType, rolePlayer);
-
-        assertThat(relation.rolePlayers(roleType), hasItems(rolePlayers));
-    }
-
-    @Property
```
aelred: These tests are for multiple role-players, they don't need to be deleted. They only checked the resource was not unique so they could be sure they could put that resource in new relationships.

grakn-graph/src/test/java/ai/grakn/graph/property/RelationPropertyTest.java
```diff
@@ -44,6 +48,41 @@
     public final ExpectedException exception = ExpectedException.none();
 
     @Property
+    public void whenAddingARolePlayer_ItIsAddedToTheCollectionOfRolePlayers(
+        Relation relation, @FromGraph RoleType roleType, @FromGraph Instance rolePlayer) {
+
+        assumeFalse(rolePlayer.isResource());
```
aelred: Actually this check can also be removed in all these tests, I suspect. Sorry, wasn't clear!

PR #1397, opened 2017-03-31
Simpler fix for PP breaking indices

fppt:


grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -983,6 +984,9 @@ public boolean fixDuplicateResources(String index, Set<ConceptId> resourceVertex
                 otherResource.deleteNode();
             }
 
+            String newIndex = mainResource.getIndex();
+            mainResource.getVertex().property(Schema.ConceptProperty.INDEX.name(), newIndex);
```
aelred: Should this still have a comment indicating it's temporary? It's a lot nicer than the other one anyway.

fppt: Am I blind I don't see the work temporary anywhere here?

PR #1399, opened 2017-03-31
Stop refreshing after every query in shell

aelred:
This might be unnecessary and is causing issues with the new transactions.


aelred:
@duckofyork can you confirm this change is alright? I'm not totally convinced myself.

fppt:
Going with @duckofyork's review so will just merge.

PR #1401, opened 2017-03-31
Explanation colours

marco-scoppetta:
- New Explore classes and 'explore' endpoint
- Removed the 'ontology' endpoint
- Updated dashboard accordingly to ontology removal and new explore
- New HALUtils class to allow small refactoring
- Small change in NodeLabelSetting panel's title
- Added Abstract class extended to builder explore Instance or Type concepts
 - ctrl+dblck clik on relation return explanation from reasoner
 - implemented HALExplanationBuilder class to build hal repr of explanation
 - Roll down graql editor on focus, if it is rolled up
 - colour picker when holding on node (together with label settings)
 - Fixed bug with resources on graph not attached to the result


marco-scoppetta:
please retest this please

PR #1402, opened 2017-04-03
Various Internal Factory Fixes

fppt:
- Cleanup internal factory implementation
- Log warning instead of throwing exception when closing factory with open transactions

### DIFF
--- a/grakn-factory/titan-factory/src/main/java/ai/grakn/factory/TitanHadoopInternalFactory.java
+++ b/grakn-factory/titan-factory/src/main/java/ai/grakn/factory/TitanHadoopInternalFactory.java
@@ -62,11 +62,6 @@ public class TitanHadoopInternalFactory extends AbstractInternalFactory<Abstract
     }
 
     @Override
-    boolean isClosed(HadoopGraph innerGraph) {
-        return false;
-    }
-
-    @Override
     AbstractGraknGraph<HadoopGraph> buildGraknGraphFromTinker(HadoopGraph graph, boolean batchLoading) {
         throw new UnsupportedOperationException(ErrorMessage.CANNOT_PRODUCE_GRAPH.getMessage(HadoopGraph.class.getName()));
     }
@@ -78,7 +73,7 @@ public class TitanHadoopInternalFactory extends AbstractInternalFactory<Abstract
     }
 
     @Override
-    protected HadoopGraph getGraphWithNewTransaction(HadoopGraph graph) {
+    protected HadoopGraph getGraphWithNewTransaction(HadoopGraph graph, boolean batchloading) {
         return graph;
     }
 }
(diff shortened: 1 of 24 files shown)


grakn-factory/titan-factory/src/main/java/ai/grakn/factory/TitanHadoopInternalFactory.java
```diff
@@ -78,7 +73,7 @@ HadoopGraph buildTinkerPopGraph(boolean batchLoading) {
     }
 
     @Override
-    protected HadoopGraph getGraphWithNewTransaction(HadoopGraph graph) {
+    protected HadoopGraph getGraphWithNewTransaction(HadoopGraph graph, boolean batchloading) {
```
aelred: Do we need a `TODO` here to acknowledge the `batchloading` param?

grakn-graph/src/main/java/ai/grakn/factory/AbstractInternalFactory.java
```diff
@@ -74,61 +75,58 @@
...
+
+    private void checkOtherGraphOpen(GraknGraph otherGraph){
+        if(otherGraph != null && !otherGraph.isClosed()){
+            throw new GraphRuntimeException(ErrorMessage.TRANSACTION_ALREADY_OPEN.getMessage(otherGraph.getKeyspace()));
+        }
+    }
+
+    protected M getGraph(M graknGraph, GraknTxType txType){
+        boolean batchLoading = false;
+        if(GraknTxType.BATCH.equals(txType)) batchLoading = true;
```
aelred: `boolean batchLoading = GraknTxType.BATCH.equals(txType);`

grakn-graph/src/main/java/ai/grakn/factory/GraknSessionImpl.java
```diff
@@ -99,8 +102,10 @@ public GraknComputer getGraphComputer() {
 
     @Override
     public void close() throws GraphRuntimeException {
-        checkClosure(openGraphTxs(), graph);
-        checkClosure(openGraphBatchTxs(), graphBatch);
+        int openTransactions = openTransactions(graph) + openTransactions(graphBatch);
+        if(openTransactions > 1){
```
aelred: Why is this `1` instead of `0`?

grakn-factory/titan-factory/src/main/java/ai/grakn/factory/TitanHadoopInternalFactory.java
fppt: I suppose we do

grakn-graph/src/main/java/ai/grakn/factory/GraknSessionImpl.java
fppt: That should be 0. You are correct. I was confused by TinkerGraph because that will always have an open transaction.

fppt:
@aelred could I get another review of this PR?

PR #1407, opened 2017-04-03
Make grakn-test-profiles parent of grakn-dist

burukuru:
The bash tests only run in the titan profile so need this dependency.

### DIFF
--- a/grakn-dist/pom.xml
+++ b/grakn-dist/pom.xml
@@ -21,7 +21,8 @@
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <parent>
-        <artifactId>grakn</artifactId>
+        <artifactId>grakn-test-profiles</artifactId>
+        <relativePath>../grakn-test-profiles</relativePath>
         <groupId>ai.grakn</groupId>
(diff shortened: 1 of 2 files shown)


burukuru:
retest this please

burukuru:
I'm not sure why the PR is failing on these two tests:
```
ai.grakn.test.engine.GraknEngineFailoverIT.whenSubmittingTasksToTwoEngines_TheyComplete
ai.grakn.test.engine.GraknEngineFailoverIT.whenSubmittingTasksToOneEngine_TheyComplete
```
Does anyone have any input on how I can make the profile options available in grakn-dist tests?
I want the titan profile to set grakn.skip-bash-tests to false.

fppt:
@alexandraorth @aelred I think those tests fall under your area of expertise?

I remember those used to fail randomly. Is that still the case?

aelred:
I'll look into it. I have no idea why this change should cause that failure...

alexandraorth:
The test that failed (`GraknEngineFailoverIT`) was one that unzips and runs the packaged distribution. This error was actually because the package was broken. I'm going to create a specific test that only runs the package so that we don't keep finding these issues in the failover tests and getting confused.

burukuru:
Thanks for spotting that Alex. Changing the parent changes project.basedir which we redefine; that's used in the assembly file so config includes fail.

burukuru:
Don't merge this just yet, the tests I added don't seem to be triggered still.

PR #1410, opened 2017-04-04
Make EngineCache Distributed

fppt:


fppt:
This is a very scrappy implementation of making the engine cache distributed. There is a lot to cleanup here when we have the time.

Current limitations:
1. The cache is not thread safe. So only one PP job should be running at ANY given time. This means that when @alexandraorth completes the locking framework. PP needs to use it. 
2. We are constantly serialising and deserialising HUGE sets and maps. This is very slow and needs to be optimised.

Other Issues:
1. I think that given more time we can merge the two caches into a unified class which depends on something which will always be running. For example Kafka. The reason we have two caches now is because our distributed cache depends on ZK running which is not true in Stand Alone Mode.

grakn-core/src/main/java/ai/grakn/graph/admin/ConceptCache.java
```diff
@@ -20,12 +20,52 @@
...
 
+import java.util.Map;
+import java.util.Set;
+
 /**
  * Admin interface for Caching concepts which can be used later.
  *
  * @author Filipe Teixeira
  */
+//TODO: Remove bloat from this interface
```
aelred: I suppose the main refactor is to make a keyspace-specific cache, seeing as nearly all these methods take `keyspace` as the first argument.

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
```diff
@@ -0,0 +1,185 @@
...
+/**
+ * <p>
+ *     Engine's internal Concept ID cache
+ * </p>
+ *
+ * <p>
+ *    This is a cache which houses {@link org.apache.tinkerpop.gremlin.structure.Vertex} ids. It keeps these ids
+ *    so that we can lookup the vertices in need of postprocessing directly.
+ *
+ *    We cannot relay on {@link ai.grakn.concept.ConceptId}s because the indexed lookup maybe faulty with
```
aelred: relay/rely

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
```diff
@@ -0,0 +1,185 @@
...
+    private final ZookeeperConnection zookeeper;
+
+
+    private EngineCacheDistributed(ZookeeperConnection zookeeper){
+        this.zookeeper = zookeeper;
+    }
+
+    public static synchronized EngineCacheDistributed init(ZookeeperConnection connection){
+        if(instance == null) instance = new EngineCacheDistributed(connection);
+        return instance;
```
aelred: This doesn't need to be a singleton, we can keep the singleton just to `EngineCacheProvider`.

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
```diff
@@ -0,0 +1,185 @@
...
+ *    We cannot relay on {@link ai.grakn.concept.ConceptId}s because the indexed lookup maybe faulty with
+ *    vertices in need of post processing.
+ *
+ *    This distributed version persists everything using {@link ai.grakn.engine.tasks.manager.ZookeeperConnection}
+ * </p>
+ *
+ * @author fppt
+ */
+//TODO: Maybe we can merge this with Stand alone engine cache using Kafka for example?
+//TODO: This class may be in need of sever optimisation. Constantly serialising and deserialising maps and sets is likely to be slow.
```
aelred: I think @alexandraorth has said this, but we can probably improve things by using the zookeeper paths like sets, using concept IDs or similar.

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheStandAlone.java
```diff
@@ -39,59 +36,40 @@
...
  *    We cannot relay on {@link ai.grakn.concept.ConceptId}s because the indexed lookup maybe faulty with
  *    vertices in need of post processing.
+ *
+ *    This stand alone version keeps everything in memory when running a single engine.
  * </p>
  *
  * @author fppt
  */
-public class EngineCache implements ConceptCache{
+//TODO: Maybe we can merge this with distributed engine cache using Kafka for example?
```
aelred: Kafka isn't used in Standalone - so I don't think this is possible.

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheProvider.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public class EngineCacheProvider {
+    private static ConceptCache conceptCache = null;
+
+    public static void init(ConceptCache cache){
+        if(conceptCache != null && !conceptCache.equals(cache)) {
+            throw new RuntimeException("The engine cache has already been initialised with another cache");
+        }
+        conceptCache = cache;
+    }
```
aelred: I'm not necessarily suggesting it for this PR, but we could eliminate this class if we:
- Put the method to access the `ConceptCache` in the `TaskManager` instead of here
- Pass the `TaskManager` to the tasks when they are started

Then no singletons 👍

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheStandAlone.java
```diff
@@ -39,59 +36,40 @@
...
 
-    private final AtomicBoolean saveInProgress;
-    private static EngineCache instance=null;
-    private final AtomicLong lastTimeModified;
+    private static EngineCacheStandAlone instance=null;
 
-    public static synchronized EngineCache getInstance(){
-        if(instance==null) instance=new EngineCache();
+    public static synchronized EngineCacheStandAlone getCache(){
+        if(instance == null) instance = new EngineCacheStandAlone();
```
aelred: We can probably eliminate this singleton as well, like `EngineCacheDistributed`.

grakn-engine/src/main/java/ai/grakn/engine/tasks/config/ZookeeperPaths.java
```diff
@@ -41,4 +41,8 @@
     String SINGLE_ENGINE_PATH = ALL_ENGINE_PATH + "/%s";
     String ZK_TASK_PATH =  TASKS_PATH_PREFIX + "/%s";
     String ZK_ENGINE_TASK_PATH = ALL_ENGINE_PATH + "/%s/%s";
+
+    String ENGINE_CACHE = "/engine/cache/";
+    String ENGINE_CACHE_KEYSPACES = ENGINE_CACHE + "keyspaces";
+    String ENGINE_CACHE_JOBS = ENGINE_CACHE + "%s/%s"; //Format is /keyspace/job_type
```
aelred: Will it be safe to have a keyspace called `keyspaces`?

grakn-test/src/test/java/ai/grakn/test/EngineContext.java
```diff
@@ -102,6 +102,8 @@ public void before() throws Throwable {
         }
 
         if (startMultiQueueEngine){
+            //TODO: This patch should die when
```
aelred: when what? D:

grakn-core/src/main/java/ai/grakn/graph/admin/ConceptCache.java
fppt: Agreed. Mostly a code cleanup optimisation though.

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
fppt: typo police!

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
fppt: Yes agreed. I think that should be in a separate PR though. To quote @alexandraorth a "A post-get-it-working optimisation"

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
fppt: I will try to change that but I recall getting some epic failures on that change

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheStandAlone.java
fppt: bugger

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheProvider.java
fppt: +1 to making that change in another PR

grakn-engine/src/main/java/ai/grakn/engine/tasks/config/ZookeeperPaths.java
fppt: it should

grakn-test/src/test/java/ai/grakn/test/EngineContext.java
fppt: oops

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
fppt: Yep just tried to make this not a singleton. Lots of test failures. We will need to fix our tests a bit in order to make this simple change

fppt:
Okay, managed to optimise it to make better use of Zookeeper paths. Just need to make it thread safe but I would like to leave that for another PR for the moment.

fppt:
please retest this please - Not sure why it failed:

```
ERROR: Failed to parse POMs
java.io.IOException: Backing channel is disconnected.
	at hudson.remoting.RemoteInvocationHandler.channelOrFail(RemoteInvocationHandler.java:187)
	at hudson.remoting.RemoteInvocationHandler.invoke(RemoteInvocationHandler.java:252)
	at com.sun.proxy.$Proxy64.isAlive(Unknown Source)
	at hudson.Launcher$RemoteLauncher$ProcImpl.isAlive(Launcher.java:996)
	at hudson.maven.ProcessCache$MavenProcess.call(ProcessCache.java:166)
	at hudson.maven.MavenModuleSetBuild$MavenModuleSetBuildExecution.doRun(MavenModuleSetBuild.java:873)
	at hudson.model.AbstractBuild$AbstractBuildExecution.run(AbstractBuild.java:534)
	at hudson.model.Run.execute(Run.java:1728)
	at hudson.maven.MavenModuleSetBuild.run(MavenModuleSetBuild.java:544)
	at hudson.model.ResourceController.execute(ResourceController.java:98)
	at hudson.model.Executor.run(Executor.java:404)
ERROR: Step ‘Set build status on GitHub commit’ failed: no workspace for unit-test-pull-request-grakn-titan #2823
ERROR: aws-jenkins-slave (sir-t9vr6zxq) is offline; cannot locate maven
Setting status of 386735e1ee8e8569fa1fa7304a69e39bcb21860c to FAILURE with url http://10.0.1.19:9998/job/unit-test-pull-request-grakn-titan/2823/ and message: 'Build finished. 529 tests run, 7 skipped, 0 failed.'
ERROR: aws-jenkins-slave (sir-t9vr6zxq) is offline; cannot locate maven
Finished: FAILURE
```

aelred:
Are you happy to merge it now @fppt ?

PR #1416, opened 2017-04-04
Refactor StandaloneTaskManager 

alexandraorth:


grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/StandaloneTaskManager.java
```diff
@@ -92,118 +92,108 @@ public TaskManager open() {
...
         }
     }
 
     public TaskStateStorage storage() {
-        return stateStorage;
+        return storage;
     }
 
-    private Runnable exceptionCatcher(TaskState state, BackgroundTask task) {
+    private Runnable exceptionCatcher(TaskState task) {
```
aelred: Rename this

PR #1421, opened 2017-04-05
Remove Graql `plays` property

aelred:
This property is unused and is necessary to rename `plays-role` to `plays`.


aelred:
please retest this please - `testGrandParentship`

PR #1422, opened 2017-04-05
Engine Distributed Locking 

alexandraorth:


grakn-engine/src/main/java/ai/grakn/engine/lock/ZookeeperLock.java
```diff
@@ -0,0 +1,125 @@
...
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+
+/**
+ * <p>
+ *     A fully distributed lock that is globally synchronous. No two engines or tasks will think they hold the same lock.
+ * </p>
+ *
+ * @author alexandraorth
+ */
+public class ZookeeperLock implements Lock {
```
fppt: You have a few unsupported operation exceptions in here. Are they going to be there forever or are you planning to implement something for them?

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/multiqueue/MultiQueueTaskManager.java
```diff
@@ -74,6 +77,7 @@ public MultiQueueTaskManager(EngineID engineId) {
 
         this.producer = ConfigHelper.kafkaProducer();
         EngineCacheProvider.init(EngineCacheDistributed.init(zookeeper));
+        LockProvider.init(new ZookeeperLock(zookeeper, LOCK));
```
fppt: Since your lock provider can only house 1 lock can you allow for locking different tasks? For example updating counts needs a lock and post processing needs another lock. They should not be sharing locks.

grakn-test/src/test/java/ai/grakn/test/engine/lock/ZookeeperLockTest.java
```diff
@@ -0,0 +1,156 @@
...
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class ZookeeperLockTest {
```
fppt: I would change this to `LockTest` and test against the interface. Then by using Junit `Theories` you can make sure that your different locks behave in the same manner. This could be reserved for a different PR though but I think is a good change to make.

fppt: Basically this will let you be 100% sure that the `ReentrantLock` and `ZKLock` behave the same way.

grakn-engine/src/main/java/ai/grakn/engine/lock/ZookeeperLock.java
alexandraorth: I don't think I'm planning on implementing the other lock functionalities because we wont be using those. We could, of course, but way in the future when we have time.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/multiqueue/MultiQueueTaskManager.java
alexandraorth: I hadn't thought of this. At the moment the LockProvider can only house one lock- so while you can have multiple locks you will only be able to access one. We will definitely need to change this.

grakn-engine/src/main/java/ai/grakn/engine/lock/ZookeeperLock.java
fppt: Okay fair enough. Then I am not 100% sure it's worth implementing this specific interface but it's no big deal for the moment as I am also not a fan of defining other interfaces. I would say `TODO: Evaluate the need of implementing this exact interface`

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/multiqueue/MultiQueueTaskManager.java
fppt: Yep that needs to change then. I would prefer for that change to happen in this PR but since we in a bit of a hurry we could put this in and get the updating counts task using it. This will make PP unsafe to run on the cluster but we could disable PP for the quick load. It means duplicates will be seen though.

grakn-test/src/test/java/ai/grakn/test/engine/lock/ZookeeperLockTest.java
alexandraorth: The two don't entirely behave the same way, though. As one is designed for use in a distributed environment. Can you use theories even if the constructors are different? For example, two ZK locks with the same path will have contention with each other, but two Reentrant locks are not given a path and will never contend with one another.

alexandraorth:
please retest this please

burukuru:
dear jenkins retest this please

grakn-test/src/test/java/ai/grakn/test/engine/lock/ZookeeperLockTest.java
fppt: You can bypass the different constructors easily. For example, the code below is used to get then cache to be tested. In this case using the cache provider.

```
private ConceptCache getCache(Caches caches){
        switch(caches){
            case DISTRIBUTED:
                return EngineCacheProvider.getCache();
            case STAND_ALONE:
                return EngineCacheStandAlone.getCache();
        }
        throw new RuntimeException("Invalid cache [" + caches + "]");
    }
```

You could do something similar. Startup the distributed engine context so when you can to use the ZK lock just use `LockProvider.get(...)`. When you want to use the re-entrant lock just return the re-erntrant lock instead.

> The two don't entirely behave the same way, though.

They should from the interface perspective. `getLock()` from the user perspective should do the same thing regardless of the implementation. So even though they behave differently under the hood, they implementing the same interface and therefore the functional result should be the same. If the functional result is not the same then we may have an issue.

alexandraorth: I'll give it a try now!

burukuru:
retest this please

PR #1427, opened 2017-04-06
Rename `plays-role` to `plays`

aelred:


grakn-graql/src/main/antlr4/ai/grakn/graql/internal/antlr/Graql.g4
```diff
@@ -71,7 +71,7 @@ varPattern     : VARIABLE | variable? property (','? property)* ;
 property       : 'isa' variable                 # isa
                | 'sub' variable                 # sub
                | 'relates' variable             # relates
-               | 'plays-role' variable          # playsRole
+               | 'plays' variable          # plays
```
alexandraorth: This is off by one tab

aelred: jesus christ @fppt look what you've done

PR #1428, opened 2017-04-06
New Background Task Which Counts The Number of Instances per Type

fppt:
- Track instance change in graph API
- Submit expanded commit logs to engine
- Cache the details needed for the new job
- Background task to persist count's to types

Note: The lack of thread safety here is going to make the counts inaccurate. Might have to synchronise on the ZK mutations.

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
@@ -233,7 +233,7 @@ public class EngineCacheDistributed extends EngineCacheAbstract{
     }
 
     @Override
-    public void deleteJobCasting(String keyspace, TypeName name) {
+    public void deleteJobInstanceCount(String keyspace, TypeName name) {
         throw new UnsupportedOperationException("Not Yet Implemented");
     }
 }
--- a/grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
@@ -19,6 +19,7 @@
 package ai.grakn.engine.cache;
 
 import ai.grakn.concept.ConceptId;
+import ai.grakn.concept.TypeName;
 import ai.grakn.engine.tasks.manager.ZookeeperConnection;
 import ai.grakn.exception.EngineStorageException;
 
@@ -217,5 +218,22 @@ public class EngineCacheDistributed extends EngineCacheAbstract{
         if(indices.isPresent()){
             indices.get().forEach(index -> cleanJobSet(jobType, keyspace, index));
         }
+    }
+
+    //-------------------- Instance Count Jobs
+
+    @Override
+    public Map<TypeName, Long> getInstanceCountJobs(String keyspace) {
+        throw new UnsupportedOperationException("Not Yet Implemented");
+    }
+
+    @Override
+    public void addJobInstanceCount(String keyspace, TypeName name, long instances) {
+        throw new UnsupportedOperationException("Not Yet Implemented");
+    }
+
+    @Override
+    public void deleteJobCasting(String keyspace, TypeName name) {
+        throw new UnsupportedOperationException("Not Yet Implemented");
     }
 }
--- a/grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
@@ -32,8 +32,9 @@ import java.util.Set;
 
 import static ai.grakn.engine.tasks.config.ZookeeperPaths.ENGINE_CACHE_CONCEPT_IDS;
 import static ai.grakn.engine.tasks.config.ZookeeperPaths.ENGINE_CACHE_EXACT_JOB;
-import static ai.grakn.engine.tasks.config.ZookeeperPaths.ENGINE_CACHE_INDICES;
+import static ai.grakn.engine.tasks.config.ZookeeperPaths.ENGINE_CACHE_JOB_TYPE;
 import static ai.grakn.engine.tasks.config.ZookeeperPaths.ENGINE_CACHE_KEYSPACES;
+import static ai.grakn.engine.tasks.config.ZookeeperPaths.ENGINE_CACHE_TYPE_INSTANCE_COUNT;
 import static java.lang.String.format;
 import static org.apache.commons.lang.SerializationUtils.deserialize;
 import static org.apache.spark.util.Utils.serialize;
@@ -56,6 +57,7 @@ import static org.apache.spark.util.Utils.serialize;
  * @author fppt
  */
 public class EngineCacheDistributed extends EngineCacheAbstract{
+    private static final String COUNTING_JOB = "counting";
     private static final String RESOURCE_JOB = "resources";
     private static final String CASTING_JOB = "castings";
     private static EngineCacheDistributed instance = null;
@@ -70,14 +72,17 @@ public class EngineCacheDistributed extends EngineCacheAbstract{
         return instance;
     }
 
-    private String getPathIndices(String jobType, String keyspace){
-        return format(ENGINE_CACHE_INDICES, keyspace, jobType);
+    private String getPathJobRoot(String jobType, String keyspace){
+        return format(ENGINE_CACHE_JOB_TYPE, keyspace, jobType);
     }
     private String getPathConceptIds(String jobType, String keyspace, String index){
         return format(ENGINE_CACHE_CONCEPT_IDS, keyspace, jobType, index);
     }
     private String getPathExactJob(String jobType, String keyspace, String index, ConceptId conceptId){
         return format(ENGINE_CACHE_EXACT_JOB, keyspace, jobType, index, conceptId.getValue());
+    }
+    private String getPathTypeInstanceCount(String keyspace, TypeName name){
+        return format(ENGINE_CACHE_TYPE_INSTANCE_COUNT, keyspace, COUNTING_JOB, name);
     }
 
     @Override
@@ -157,7 +162,7 @@ public class EngineCacheDistributed extends EngineCacheAbstract{
     }
     private Map<String, Set<ConceptId>> getJobs(String jopType, String keyspace){
         Map<String, Set<ConceptId>> jobs = new HashMap<>();
-        Optional<List<String>> indices = getChildrenFromZookeeper(getPathIndices(jopType, keyspace));
+        Optional<List<String>> indices = getChildrenFromZookeeper(getPathJobRoot(jopType, keyspace));
 
         if(indices.isPresent()){
             for (String index : indices.get()) {
@@ -216,7 +221,7 @@ public class EngineCacheDistributed extends EngineCacheAbstract{
         clearAllJobs(CASTING_JOB, keyspace);
     }
     private void clearAllJobs(String jobType, String keyspace){
-        Optional<List<String>> indices = getChildrenFromZookeeper(getPathIndices(jobType, keyspace));
+        Optional<List<String>> indices = getChildrenFromZookeeper(getPathJobRoot(jobType, keyspace));
         if(indices.isPresent()){
             indices.get().forEach(index -> cleanJobSet(jobType, keyspace, index, true));
         }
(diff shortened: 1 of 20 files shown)


fppt:
Had to restart on travis because of failure:

```
Failed tests: 
  TasksControllerTest.afterSendingTaskWithRunAt_ItIsDelayedInStorage:129 
Argument(s) are different! Wanted:
taskManager.addTask(
    <Tasks controller test$$ lambda$ 1 6 3 4/ 7 5 3 9 5 5 8 0 2>
);
-> at ai.grakn.test.engine.controller.TasksControllerTest.afterSendingTaskWithRunAt_ItIsDelayedInStorage(TasksControllerTest.java:129)
Actual invocation has different arguments:
taskManager.addTask(
    TaskState(ShortExecutionMockTask, "fd2d6a22-7ece-4535-b5e4-4f40d8b8f856").status(CREATED)
);
-> at ai.grakn.engine.controller.TasksController.createTask(TasksController.java:192)

```

fppt:
@alexandraorth this PR can go in as is but will need to be integrated with your locking mechanism.

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
```diff
@@ -189,21 +195,23 @@ private void addJob(String jobType, String keyspace, String index, ConceptId id)
...
 
-        if(ids.isPresent()){
+        if(force && ids.isPresent()){
             ids.get().forEach(id -> deleteJob(jobType, keyspace, index, ConceptId.of(id)));
         }
 
-        deleteObjectFromZookeeper(idPath);
+        if(force || ids.isPresent() && ids.get().isEmpty()){
+            deleteObjectFromZookeeper(idPath);
+        }
```
aelred: What does this method do and what's the `force` parameter for?

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheStandAlone.java
```diff
@@ -42,11 +43,11 @@
  *
  * @author fppt
  */
-//TODO: Maybe we can merge this with distributed engine cache using Kafka for example?
```
aelred: Don't give up on the dream!

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheStandAlone.java
```diff
@@ -140,5 +143,23 @@ public void clearAllJobs(String keyspace){
...
         if(castings.containsKey(keyspace)) castings.remove(keyspace);
         if(resources.containsKey(keyspace)) resources.remove(keyspace);
+        if(instanceCounts.containsKey(keyspace)) instanceCounts.remove(keyspace);
+    }
+
+    //-------------------- Instance Count Jobs
+
+    @Override
+    public Map<TypeLabel, Long> getInstanceCountJobs(String keyspace) {
+        return instanceCounts.computeIfAbsent(keyspace, (key) -> new ConcurrentHashMap<>());
```
aelred: Might be better to use `getOrDefault`, to avoid putting unnecessary empty stuff in the `Map`.

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheStandAlone.java
```diff
@@ -140,5 +143,23 @@ public void clearAllJobs(String keyspace){
...
+    }
+
+    @Override
+    public void addJobInstanceCount(String keyspace, TypeLabel name, long instanceCount) {
+        getInstanceCountJobs(keyspace).compute(name, (key, value) -> value == null ? instanceCount : value + instanceCount);
```
aelred: You can use `merge` here. I think like this:
```
getInstanceCountJobs(keyspace).merge(name, instanceCount, (old, new) -> old + new);
```

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/UpdatingInstanceCountTask.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @Override
+    public boolean start(Consumer<TaskCheckpoint> saveCheckpoint, Json configuration) {
+        cache.getKeyspaces().parallelStream().forEach(this::updateCountsOnKeySpace);
+        return true;
+    }
+
+    private void updateCountsOnKeySpace(String keyspace){
+        Map<TypeLabel, Long> jobs = new HashMap<>(cache.getInstanceCountJobs(keyspace));
+        //Clear the cache optimistically because we think we going to update successfully
```
aelred: So bold!

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/UpdatingInstanceCountTask.java
```diff
@@ -0,0 +1,117 @@
...
+        } catch (InterruptedException e1) {
+            LOG.error("Exception",e1);
+        }
+
+        return retry;
+    }
+
+    @Override
+    public boolean stop() {
+        throw new UnsupportedOperationException("Not yet implemented");
```
aelred: Maybe we should make these `default` implementations in the interface @alexandraorth

grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptLog.java
```diff
@@ -216,4 +227,16 @@ boolean isTypeCached(TypeLabel label){
         //noinspection unchecked
         return (X) typeCache.get(label);
     }
+
+    void addedInstance(TypeLabel name){
+        instanceCount.compute(name, (key, value) -> value == null ? 1 : value + 1);
```
aelred: These can also use `merge`, the same as above:
```
instanceCount.merge(name, 1, (old, new) -> old + new);
```

grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptLog.java
```diff
@@ -216,4 +227,16 @@ boolean isTypeCached(TypeLabel label){
...
+        cleanupInstanceCount(name);
```
aelred: I think you don't need to run this when adding an instance, only removing.

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
```diff
@@ -150,6 +150,7 @@ V addInstance(Schema.BaseType instanceBaseType, BiFunction<Vertex, T, V> produce
         }
 
         Vertex instanceVertex = getGraknGraph().addVertex(instanceBaseType);
+        getGraknGraph().getConceptLog().addedInstance(getLabel());
```
aelred: This is really petty: but if the behaviour for decrementing the count is in `InstanceImpl`, then that's also where the increment behaviour should be, not `TypeImpl`.

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
fppt: So `cleanJobSet` is used to clear the key of a job. For example the index to be merged. However, it only actually clears this *if* there are no pending jobs under that key. 

*force* lets you clear the jobs regardless if you have any pending or not. And this is needed whenever you clear the graph.

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheStandAlone.java
fppt: :(

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheStandAlone.java
fppt: How does `merge` work if the value does not exist?

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/UpdatingInstanceCountTask.java
fppt: So desperate! Honestly this is a hack because the distributed cache is not threadsafe whatsoever.

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/UpdatingInstanceCountTask.java
fppt: +1 I have seen that in many cases there are no implemntations

grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptLog.java
fppt: Same question. What if the value does not exist? Does it just put the new value in?

grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptLog.java
fppt: You do, if you do something dumb like:

```
entityType.instances().iterator.next().delete();
entityType.addEntity();
```

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
fppt: Ideally yes but where would you put it? In the constructor? The constructor has no way of knowing if a vertex is new or old. Same for the factory. You could maybe put it in the `type` setter (because that is only called for new concepts) but that sounds odd.

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/UpdatingInstanceCountTask.java
```diff
@@ -0,0 +1,117 @@
...
+        //TODO: All this boiler plate retry should be moved into a common graph mutating background task
+
+        boolean notDone = true;
+        int retry = 0;
+
+        while(notDone) {
+            notDone = false;
+            try (GraknGraph graknGraph = EngineGraknGraphFactory.getInstance().getGraph(keyspace, GraknTxType.WRITE)) {
+                graknGraph.admin().updateTypeCounts(jobs);
+                graknGraph.admin().commitNoLogs();
```
alexandraorth: What an awesome function!

grakn-test/src/test/java/ai/grakn/test/engine/cache/EngineCacheTest.java
```diff
@@ -42,25 +45,33 @@
 
 //NOTE: This test is only in grakn-test because it needs a running ZK
 //Ideally this should be moved to the grakn-engine module
-public class EngineCacheDistributedTest {
-    private ConceptCache cache;
+@RunWith(Theories.class)
```
alexandraorth: Fancy

grakn-test/src/test/java/ai/grakn/test/engine/cache/EngineCacheTest.java
```diff
@@ -42,25 +45,33 @@
...
+public class EngineCacheTest {
 
     //We do this just so we can start ZK
     @ClassRule
     public static final EngineContext engine = EngineContext.startSingleQueueServer();
```
alexandraorth: If you only need kafka, and not engine, you can do:
```
      public static final EngineContext kafka = EngineContext.startKafkaServer(); 
```

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheStandAlone.java
aelred: It just adds it, like `put` would do.

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
aelred: Oh yeah, I forget that the constructor doesn't always mean the instance is new.

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheStandAlone.java
fppt: will update that later then

grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptLog.java
aelred: Ohh, because the count can go negative? Damn...

grakn-test/src/test/java/ai/grakn/test/engine/cache/EngineCacheTest.java
fppt: It only needs ZK not kafka. Basically my point here is that all this engine test setup code should probably be moved to engine (at some point) so we can start breaking down tests more modularly. Right now grakn-test is becoming a monster. I wanted to move this test into grakn-engine but then I realised I can't because all the setup is in `grakn-test`

grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptLog.java
fppt: Yep. If you do a large deletion then that may mean you will not need to do the super node split. So it's important to track these numbers.

PR #1431, opened 2017-04-07
Get rid of redundant parse

fppt:


fppt:
@alexandraorth I agree.

PR #1432, opened 2017-04-07
Fixing some mistakes on the new post processing jobs + time stamp on cache

fppt:


grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
```diff
@@ -220,6 +221,20 @@ public void clearAllJobs(String keyspace) {
...
         clearAllJobs(CASTING_JOB, keyspace);
     }
+
+    @Override
+    public long getLastTimeJobAdded() {
+        Optional<Object> value = getObjectFromZookeeper(ENGINE_CACHE_UPDATE_TIME);
+        if(value.isPresent()){
+            return (long) value.get();
+        }
+        return 0L;
```
aelred: Maybe:
`return getObjectFromZookeeper(ENGINE_CACHE_UPDATE_TIME).orElse(0L);`
but you might need a nasty cast...

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
```diff
@@ -150,7 +150,9 @@ V addInstance(Schema.BaseType instanceBaseType, BiFunction<Vertex, T, V> produce
         }
 
         Vertex instanceVertex = getGraknGraph().addVertex(instanceBaseType);
-        getGraknGraph().getConceptLog().addedInstance(getLabel());
+        if(!Schema.MetaSchema.isMetaLabel(getLabel())) {
+            getGraknGraph().getConceptLog().addedInstance(getLabel());
+        }
```
aelred: When are we making instances of meta-types?

fppt: `graknGraph.admin().getMetaRuleInference().putRule(lhs, rhs);` :)

grakn-engine/src/main/java/ai/grakn/engine/cache/EngineCacheDistributed.java
fppt: better yes

PR #1435, opened 2017-04-07
Grakn Sharding - VERY BIG WIP

fppt:


fppt:
@aelred This is "stable enough" for the graql changes. If you have time now I could use your help if not I will try to handle the graql part later.

burukuru:
moon of my life retest this please

aelred:
@fppt I think you also need to update `ElementFactory` line 201 (the ghostbuster)

grakn-graph/src/main/java/ai/grakn/graph/internal/InstanceImpl.java
```diff
@@ -59,7 +59,17 @@
  */
 abstract class InstanceImpl<T extends Instance, V extends Type> extends ConceptImpl<T> implements Instance {
     private ComponentCache<TypeLabel> cachedInternalType = new ComponentCache<>(() -> TypeLabel.of(getProperty(Schema.ConceptProperty.TYPE)));
-    private ComponentCache<V> cachedType = new ComponentCache<>(() -> this.<V>getOutgoingNeighbours(Schema.EdgeLabel.ISA).findFirst().orElse(null));
+
+    private ComponentCache<V> cachedType = new ComponentCache<>(() -> {
+        ConceptImpl<?> currentType = (ConceptImpl) this.getOutgoingNeighbours(Schema.EdgeLabel.ISA).findFirst().orElse(null);
+
+        while (currentType.isShard()){
```
aelred: Does this have to be a loop? Given we only have one depth of shards it seems unnecessary. If this is just a matter of preparing for when we do have a tree of shards, then that's alright - although I'd face that when it comes.

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
```diff
@@ -229,7 +243,7 @@ V addInstance(Schema.BaseType instanceBaseType, BiFunction<Vertex, T, V> produce
     public void delete(){
         checkTypeMutation();
         boolean hasSubs = getVertex().edges(Direction.IN, Schema.EdgeLabel.SUB.getLabel()).hasNext();
-        boolean hasInstances = getVertex().edges(Direction.IN, Schema.EdgeLabel.ISA.getLabel()).hasNext();
+        boolean hasInstances = ((TypeImpl) currentShard()).getIncomingNeighbours(Schema.EdgeLabel.ISA).findAny().isPresent();
```
aelred: This will allow the type to be deleted even if it has another shard which still has instances.

grakn-graph/src/main/java/ai/grakn/graph/internal/ValidateGlobalRules.java
```diff
@@ -186,7 +185,7 @@ private ValidateGlobalRules() {
      * @return An error message if the conceptType  abstract and has incoming isa edges
      */
     static Optional<String> validateIsAbstractHasNoIncomingIsaEdges(TypeImpl conceptType){
-        if(conceptType.isAbstract() && conceptType.getVertex().edges(Direction.IN, Schema.EdgeLabel.ISA.getLabel()).hasNext()){
+        if(conceptType.isAbstract() && ((TypeImpl) conceptType.currentShard()).getIncomingNeighbours(Schema.EdgeLabel.ISA).findAny().isPresent())  {
```
aelred: This will accept an abstract type with multiple shards, even if one of those shards has instances.

grakn-graph/src/main/java/ai/grakn/graph/internal/InstanceImpl.java
fppt: It's prepping for that very scenario

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
fppt: Good catch. Will fix

grakn-graph/src/main/java/ai/grakn/graph/internal/ValidateGlobalRules.java
fppt: Good catch again. Will fix

fppt:
Waiting on @kasper-piskorski to do his magic with reasoner then we put this in and watch the stack explode.

PR #1445, opened 2017-04-10
Fix query `$x sub $y` to only return types

aelred:


aelred:
please retest this please jenkins why did you go funny

PR #1448, opened 2017-04-11
Get updating count task to lock

fppt:


grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/singlequeue/SingleQueueTaskManager.java
```diff
@@ -135,7 +136,8 @@ public SingleQueueTaskManager(EngineID engineId) throws Exception {
 
         EngineCacheProvider.init(EngineCacheDistributed.init(zookeeper));
 
-        LockProvider.add(POST_PROCESSING_LOCK, new ZookeeperLock(zookeeper, LOCK));
+        LockProvider.add(PostProcessingTask.LOCK_KEY, new ZookeeperLock(zookeeper, ZookeeperPaths.LOCK));
```
alexandraorth: Because these two locks are on the same path (second argument of creating `ZookeeperLock`), they will lock with one another. They should be on two paths.

fppt: Oh I misunderstood what that did. Okay this needs fixing. Will do so now.

grakn-engine/src/main/java/ai/grakn/engine/tasks/storage/LockingBackgroundTask.java
```diff
@@ -0,0 +1,76 @@
...
+ * @author alexandraorth, fppt
+ */
+public abstract class LockingBackgroundTask implements BackgroundTask{
+
+    @Override
+    public boolean start(Consumer<TaskCheckpoint> saveCheckpoint, Json configuration) {
+        Lock engineLock = LockProvider.getLock(getLockingKey());
+        try {
+            // Try to get lock for one second. If task cannot acquire lock, it should return successfully.
+            boolean hasLock = engineLock.tryLock(1, TimeUnit.SECONDS);
```
alexandraorth: To frame the below: If a task returns `true` from the start method, it is marked as completed. If it returns `false`, it is marked as stopped. `STOPPED` tasks are never re-submitted if they are recurring, but `COMPLETED` tasks are. 

Using `tryLock` here might be an issue. This task will wait for the lock for 1 second, if it cannot acquire it, it returns `true` and gets marked as `COMPLETED`. 

This is the use case that we have designed for in the `PostProcessingTask`, which is recurring. If the PP task cannot acquire the lock it returns `true`, gets marked as `COMPLETED` and is resubmitted. If it can acquire the lock, it runs the post processing, returns `false` and gets marked as `STOPPED`, never to run again. 

Im not sure if we designed for this same use case in the counting task, which I *think* is non-recurring. This means that most instances of counting task will never run if they cannot acquire the lock the first time.

grakn-engine/src/main/java/ai/grakn/engine/tasks/storage/LockingBackgroundTask.java
```diff
@@ -0,0 +1,76 @@
...
+                    return runLockingBackgroundTask(saveCheckpoint, configuration);
+                } finally {
+                    engineLock.unlock();
+                }
+            }
+        } catch (InterruptedException e){
+            throw new RuntimeException(e);
+        }
+
+        return true;
```
alexandraorth: See above long-comment about the return value.

grakn-engine/src/main/java/ai/grakn/engine/tasks/storage/LockingBackgroundTask.java
fppt: Oh that is a big bug. What's the simplest solution for the moment? Make counting recurring?

fppt:
@alexandraorth another review please

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/UpdatingInstanceCountTask.java
```diff
@@ -50,11 +50,17 @@
  *
```
alexandraorth: Lovely javadoc ;)

grakn-engine/src/main/java/ai/grakn/engine/tasks/storage/LockingBackgroundTask.java
alexandraorth: Yep, the easiest solution at the moment is to make counting behave like Post Processing (this is, of course, the way PP behaves in the PR I have yet to make). We should make it recurring and return `false` when it actually acquires the lock and successfully runs.

fppt:
@alexandraorth one more review for the lols please

PR #1449, opened 2017-04-11
Making Sharding Threshold configurable via config file

fppt:
TODO: Pass `properties` around by itself to try minimise the size of the constructor


grakn-graph/src/test/java/ai/grakn/factory/GraknTinkerGraphFactoryTest.java
```diff
@@ -38,13 +45,25 @@
 import static org.junit.Assert.assertThat;
 
 public class GraknTinkerGraphFactoryTest {
+    private final static String TEST_CONFIG = "../conf/test/tinker/grakn.properties";
```
alexandraorth: Hardcoding?

PR #1472, opened 2017-04-21
Update strategy and add better string output

kasper-piskorski:


kasper-piskorski:
please retest this please - ai.grakn.test.engine.controller.TasksControllerTest.afterSendingTaskWithRunAt_ItIsDelayedInStorage

PR #1473, opened 2017-04-21
Fixed two Engine bugs

marco-scoppetta:
- GraqlController now uses WRITE graph to let Reasoner materialise
  - response body is returned outside the try block to allow the open transactions to autocommit


pouledodue:
exactly what I needed

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
```diff
@@ -102,16 +102,19 @@ private Json executeGraql(Request request, Response response){
...
+        try(GraknGraph graph = factory.getGraph(keyspace, WRITE)){
             Query<?> query = graph.graql().materialise(materialise).infer(infer).parse(queryString);
 
             if(!readOnly(query)){
                 throw new GraknEngineServerException(405, "Only \"read-only\" queries are allowed.");
             }
-
-            return respond(request, query, response);
+            //Do not return directly inside the try block otherwise open transactions will not autocommit and close properly.
+            responseBody = respond(request, query, response);
```
bolerio: How is that possible that transaction don't auto-close in the try block? What's happening?

marco-scoppetta: The inference from Reasoner doesn't get materialised in the graph when we have the return statement inside the try block. Moving it outside solves the problem. I will try to debug to see if the close method on the graph gets called correctly even with the return statement, but the observed behaviour suggests otherwise

bolerio: Couldn't that be because the tx was declared as "READ" and you changed it to "WRITE"?

marco-scoppetta: With READ tx, the reasoner inference throws an exception when trying to materialise because the graph is read-only. We changed it to WRITE and it stopped throwing exceptions, but then the IDs returned to the client, computed by Reasoner, were not materialised in the graph, so we figured the tx was not committing properly. Moving the return outside the try-with-resource block solved it

bolerio: Ok, I see. This seems symptomatic of a deeper problem that needs Filipe's attention, @fppt. Do you mind entering a bug for it in TP so a test case is written as well, @marco-scoppetta ?

marco-scoppetta: Yes, I also thought the resource should be auto-closed even with the return statement being inside the block. Maybe something else is happening, I will add this to TP for further investigation.

bolerio: Cool, I might have to do with the bug I observed where the JVM doesn't exit after a seemingly trivial, simple try block. Something not getting closed properly.

fppt: Hi guys is it that the transactions is not closing or a commit is not happening? Because the auto close no longer auto commits regardless of the transaction type. To commit you HAVE to say `graph.commit()` this will commit and close.

alexandraorth: In this situation it was that the commit was not happening. Why does the auto close no longer auto commit?

fppt: That was part of the transaction changes that happened a few weeks ago. The auto close JUST closes now.

bolerio: Ah, that explains it, I had forgotten about that part. Since I didn't expect such a prompt response from Filipe, I already merged the PR, so we should make (and test) a fix with the commit call.

bolerio: @alexandraorth It's not an obvious whether to commit or abort on an auto-close, absent any hint from the user. I'd go with a commit whenever there is no exception, but it was decided otherwise.

alexandraorth: Interestingly, it works both with the explicit `commit` and the `return` from inside the try, as well as above, with only the `return` outside the try. Either we were mistaken or we have some strange hidden behavior happening here. 

@marco-scoppetta we should try an figure out why/debug this tomorrow!

PR #1489, opened 2017-04-25
Add unification tests with types

kasper-piskorski:


kasper-piskorski:
please retest this please - cassandra failure

PR #1507, opened 2017-05-02
Make Graql use `Answer` class instead of `Map`

aelred:
This allows the removal of several methods on the Graql interfaces.


aelred:
please retest this please

kasper-piskorski:
shouldn't we move the `QueryAnswer` class outside of the reasoner package?

aelred:
@kasper-piskorski that's true. I'll move it out.

PR #1508, opened 2017-05-02
Wrap titan exception in grakn exception

fppt:


grakn-factory/titan-factory/src/main/java/ai/grakn/graph/internal/GraknTitanGraph.java
```diff
@@ -90,8 +91,25 @@ protected void clearGraph() {
...
     @Override
     public void commitTransactionInternal(){
+        executeLockingMethod(super::commitTransactionInternal);
+    }
+
+    @Override
+    Vertex addVertex(Schema.BaseType baseType){
+        final Vertex[] v = new Vertex[1];
+        executeLockingMethod(() -> v[0] = super.addVertex(baseType));
+        return v[0];
```
aelred: Dammit java. You could maybe neaten this a bit by making `executeLockingMethod` return the result of whatever function its passed, then it's just:
```
return executeLockingMethod(() -> super.addVertex(baseType));
```

fppt: Thought of that. The problem is that sometimes the function passed into `executeLockingMethod` does return something, sometimes it doesn't. 

I thought of using an optional in that case but it still looks ugly. Either way you end up either returning a null which gets ignored (not ideal I admit) or this ugly thing.

Any preference on the ugliness ?

aelred: Both approaches seem pretty bad, so I'll approve this. If we end up using this method a lot more, we can reconsider.

PR #1509, opened 2017-05-02
remove hard coded analytics results from basic genealogy example

sheldonkhall:


sheldonkhall:
please retest this please

PR #1510, opened 2017-05-02
Engine Cleanup BackgroundTasks

alexandraorth:


grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessingTask.java
```diff
@@ -81,95 +74,49 @@ public boolean start(Consumer<TaskCheckpoint> saveCheckpoint, TaskConfiguration
...
 
         return true;
     }
 
     @Override
-    public boolean stop() {
-        throw new UnsupportedOperationException("Post processing cannot be stopped while in progress");
-    }
+    public boolean runGraphMutatingTask(GraknGraph graph, Consumer<TaskCheckpoint> saveCheckpoint, TaskConfiguration configuration) {
+        applyPPToMapEntry(graph, configuration, Schema.BaseType.CASTING.name(), PostProcessingTask::runCastingFix);
```
fppt: Nitpick `applyPPToMapEntry` is a horrible name for a method. How about `runPostProcessingTask`

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessingTask.java
```diff
@@ -81,95 +74,49 @@ public boolean start(Consumer<TaskCheckpoint> saveCheckpoint, TaskConfiguration
...
+    private void applyPPToMapEntry(GraknGraph graph, TaskConfiguration configuration, String type,
+                                   TriFunction<GraknGraph, String, Set<ConceptId>, Boolean> postProcessingMethod){
+        Json innerConfig = configuration.json().at(COMMIT_LOG_FIXING);
 
-                return;
-            } catch (Throwable t) { //These exceptions need to become more specialised
-                LOG.error(ErrorMessage.POSTPROCESSING_ERROR.getMessage(t.getMessage()), t);
-            }
+        Map<String, Json> conceptsByIndex = innerConfig.at(type).asJsonMap();
+        for(Map.Entry<String, Json> castingIndex:conceptsByIndex.entrySet()){
```
fppt: `castingIndex` is a misleading name. Should just be `conceptIndex`

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessingTask.java
```diff
@@ -81,95 +74,49 @@ public boolean start(Consumer<TaskCheckpoint> saveCheckpoint, TaskConfiguration
...
-    }
+    public boolean runGraphMutatingTask(GraknGraph graph, Consumer<TaskCheckpoint> saveCheckpoint, TaskConfiguration configuration) {
+        applyPPToMapEntry(graph, configuration, Schema.BaseType.CASTING.name(), PostProcessingTask::runCastingFix);
+        applyPPToMapEntry(graph, configuration, Schema.BaseType.RESOURCE.name(), PostProcessingTask::runResourceFix);
 
-    @Override
-    public void pause() {
-        throw new UnsupportedOperationException("Post processing cannot be paused");
-    }
+        graph.admin().commitNoLogs();
```
fppt: The current flow of this method `runGraphMutatingTask` risks creating large commits by committing at the *end* of all the PP. `graph.admin().commitNoLogs();` should rather be done after each fix for each entry is validated to be done correctly. Maybe after:

```
if(postProcessingMethod.apply(graph, castingIndex.getKey(), conceptIds)) {
             if (retry > MAX_RETRY) {		 +                validateMerged(graph, castingIndex.getKey(), conceptIds).
                 notDone = false;		 +                        ifPresent(message -> {
                             throw new RuntimeException(message);
                         });
              }		              
}
```

This will at least let you fix some things rather than ALL or nothhing.

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessingTask.java
alexandraorth: 100% agree. This is what I get for copy-and-pasting old code.

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessingTask.java
alexandraorth: `graph.admin().commitNoLogs();` closes the graph, so to do this will we need to refactor the way that `AbstractGraphMutationTask` works? I was working under the design that each task would only be one graph mutation.

fppt: That was an incorrect assumption. If you look more carefully at the old `ConceptFixer#runPostProcessingJob` you will see that each fix and commit was per `index` not per job type. This new format has changed that and made pp commits substantially larger and more importantly longer living, which is very dangerous as we scale.

alexandraorth: Okay, this will require a somewhat large refactor. Im not 100% sure how to go about it but will get to it tomorrow morning.

alexandraorth:
@fppt Ive made the changes- can you re-review?

PR #1513, opened 2017-05-02
Bug #13531: JVM does not close down at the end of application

fppt:


fppt:
I am trying to figure out that test now. For the moment I am just trying to make sure that all the sessions are closed by the end of a test.

grakn-factory/titan-factory/src/main/java/ai/grakn/graph/internal/GraknTitanGraph.java
```diff
@@ -74,31 +74,45 @@ public void openTransaction(GraknTxType txType){
...
+            return null;
+        });
     }
 
     @Override
     Vertex addVertex(Schema.BaseType baseType){
-        final Vertex[] v = new Vertex[1];
-        executeLockingMethod(() -> v[0] = super.addVertex(baseType));
-        return v[0];
+        return executeLockingMethod(() -> super.addVertex(baseType));
```
aelred: yayyy

fppt: boooooo: 

```
@Override
    public void commitTransactionInternal(){
        executeLockingMethod(() -> {
            super.commitTransactionInternal();
            return null;
        });
    }
```

PR #1517, opened 2017-05-03
Fix Graql bug where resource index optimisation was not applied

aelred:


aelred:
please retest this please - due to PP failure

aelred:
please retest this please - due to PP failure

PR #1519, opened 2017-05-03
Fix problem with repeating answers when materialising resources

kasper-piskorski:


grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/Resource.java
```diff
@@ -149,10 +149,7 @@ public boolean isAllowedToFormRuleHead(){
     }
 
     @Override
-    public boolean requiresMaterialisation(){
-        //requires materialisation if value variable is user defined
-        return getMultiPredicate().isEmpty();
-    }
+    public boolean requiresMaterialisation(){ return true;}
```
fppt: That's the fix?? Really?? Wow. . .

PR #1520, opened 2017-05-03
Enable loader client arguments in graql shell

sheldonkhall:


grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
```diff
@@ -169,6 +169,8 @@ public static void runShell(String[] args, String version, String historyFilenam
         options.addOption("f", "file", true, "graql file path to execute");
         options.addOption("r", "uri", true, "uri to factory to engine");
         options.addOption("b", "batch", true, "graql file path to batch load");
+        options.addOption("s", "size", true, "the size of the batches");
+        options.addOption("a", "active", true, "the number of active tasks");
```
aelred: It would probably good to include in the help text that this is only for batch loads.

grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
```diff
@@ -262,10 +272,12 @@ private static String loadQuery(String filePath) throws IOException {
             return lines.stream().collect(joining("\n"));
     }
 
-    private static void sendBatchRequest(String uriString, String graqlPath, String keyspace) throws IOException {
+    private static void sendBatchRequest(LoaderClient loaderClient, String graqlPath, String activeTasks, String batchSize) throws IOException {
         AtomicInteger numberBatchesCompleted = new AtomicInteger(0);
 
-        LoaderClient loaderClient = new LoaderClient(keyspace, uriString).setRetryPolicy(true);
+        if (Integer.parseInt(activeTasks)!=0) loaderClient.setNumberActiveTasks(Integer.parseInt(activeTasks));
+        if (Integer.parseInt(activeTasks)!=0) loaderClient.setBatchSize(Integer.parseInt(batchSize));
```
aelred: We should consider the case when a user passes something that is not an integer:
`graql.sh -b file.gql -a "users are the worst"`

I would parse and handle the edge cases outside of `sendBatchRequest`, then just passing the integers into this method.

sheldonkhall:
@aelred I made the changes - not sure how to write a test for the case when the args are not an integer. Don't know whether it is even worth it...

PR #1526, opened 2017-05-03
Dashboard fixes and tests:

marco-scoppetta:
- move Canvas mouse events handler in new module CanvasEvents
  - refactored some methods in Visualiser class to use visjs more APIs
  - added canvas-prebuilt dependecy in order to add support to jsdom and mock canvas elements
  - new visualiser tests
  - refactored addEdge method so that edges become smooth when multiple edges between same 2 nodes


grakn-dashboard/test/visualiser.test.js
```diff
@@ -62,3 +66,55 @@ test('Visualiser test \'alreadyConnected\' method', () => {
   expect(v.alreadyConnected('id-1', 'id-2', 'e')).toBeTruthy();
   expect(v.alreadyConnected('id-1', 'id-3', 'e')).toBeFalsy();
 });
+
+test('Visualiser test when deleting a node also delete all its edges', () => {
```
fppt: much happieness

PR #1528, opened 2017-05-03
Bug #13405: Ensure that all regexes are applied.

fppt:
When having multiple resource types with different regexes ensure they are all applied to the instances of those resource types.

### DIFF
--- a/grakn-graph/src/main/java/ai/grakn/graph/internal/ResourceTypeImpl.java
+++ b/grakn-graph/src/main/java/ai/grakn/graph/internal/ResourceTypeImpl.java
@@ -20,6 +20,7 @@ package ai.grakn.graph.internal;
 
 import ai.grakn.concept.Resource;
 import ai.grakn.concept.ResourceType;
+import ai.grakn.concept.TypeLabel;
 import ai.grakn.exception.InvalidConceptValueException;
 import ai.grakn.util.ErrorMessage;
 import ai.grakn.util.Schema;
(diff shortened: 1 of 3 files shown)


grakn-graph/src/main/java/ai/grakn/graph/internal/ResourceTypeImpl.java
```diff
@@ -68,6 +69,16 @@ private ResourceTypeImpl(ResourceTypeImpl resourceType){
...
 
     /**
+     * This method is overridden so that we can check that the regex of the new super type (if it has a regex)
+     * can be applied to all the existing instances.
+     */
+    @Override
+    public ResourceType<D> superType(ResourceType<D> superType){
+        checkInstancesMatchRegex(superType.getLabel(), superType.getRegex());
+        return super.superType(superType);
+    }
```
aelred: I suspect we might have a subtle bug (which existed before, to be fair):
Given we have three types `foo`, `bar`, `baz` with regexes and instances, plus `foo sub bar;`.
If we state `baz sub foo`, it will check that instances of `baz` satisfy the `foo` regex, but not `bar`.

grakn-graph/src/test/java/ai/grakn/graph/internal/ResourceTypeTest.java
```diff
@@ -80,23 +78,14 @@ public void testRegexInstance(){
...
     }
 
     @Test
-    public void checkSuper() throws Exception{
-        ResourceType<String> superConcept = graknGraph.putResourceType("super", ResourceType.DataType.STRING);
-        ResourceType<String> resourceType = graknGraph.putResourceType("resourceType", ResourceType.DataType.STRING);
-        resourceType.superType(superConcept);
-        assertThat(resourceType.superType(), instanceOf(ResourceType.class));
-        assertEquals(superConcept, resourceType.superType());
-    }
```
aelred: Was this deleted because it's redundant?

fppt: Yep

grakn-graph/src/main/java/ai/grakn/graph/internal/ResourceTypeImpl.java
fppt: Ah yes good catch.

PR #1535, opened 2017-05-04
Refactor Graql shortcut edge optimisation

aelred:
We now use a design previously used for resource indices, where we perform a "pass" over the fragments, searching for fragments that can be combined into shortcut edges.

This new approach makes the shortcut edge optimisation much more flexible - it should now be applied in nearly all cases (e.g. ternary relations, relations with resources etc.)

This will also resolve some bugs related to when relations or roles have sub-types.


aelred:
@fppt be picky

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/OutShortcutFragment.java
```diff
@@ -0,0 +1,93 @@
...
+                VarName relation, VarName edge, VarName rolePlayer, Optional<TypeLabel> roleType,
+                Optional<TypeLabel> relationType) {
+            super(relation, rolePlayer, edge);
+            this.edge = edge;
+            this.roleType = roleType;
+            this.relationType = relationType;
+    }
+
+    @Override
+    public void applyTraversal(GraphTraversal<Vertex, Vertex> traversal) {
```
fppt: This is *very* similar to `InShortcutFragment#applyTraversal`. Specifically applying the type labels. No way to make that code common amongst the two traversals?

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/OutShortcutFragment.java
```diff
@@ -0,0 +1,93 @@
...
+        GraphTraversal<Vertex, Edge> edgeTraversal = traversal.outE(SHORTCUT.getLabel()).as(edge.getValue());
+        applyTypeLabelToTraversal(edgeTraversal, ROLE_TYPE_LABEL, roleType);
+        applyTypeLabelToTraversal(edgeTraversal, RELATION_TYPE_LABEL, relationType);
+        edgeTraversal.otherV();
```
fppt: This is *super* pedantic but when possible `otherV` should be avoided. This is merely because in Titan's case it results in doing some extra equality checks and those checks have the *very* remote possibility of causing a read on the database if your cache has been cleared.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/sets/EquivalentFragmentSets.java
```diff
@@ -178,11 +176,30 @@ public static EquivalentFragmentSet regex(VarName resourceType, String regex) {
     public static void optimiseFragmentSets(
             Collection<EquivalentFragmentSet> fragmentSets, GraknGraph graph) {
 
+        // TODO: Create a real interface for these when there are more of them
+        ImmutableList<Supplier<Boolean>> optimisations = ImmutableList.of(
```
fppt: Oooo I like where you going with this.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/sets/ShortcutFragmentSet.java
```diff
@@ -19,24 +19,126 @@
...
+import ai.grakn.util.Schema;
 
+import java.util.Collection;
 import java.util.Optional;
 
+import static ai.grakn.graql.internal.gremlin.sets.EquivalentFragmentSets.hasDirectSubTypes;
+import static ai.grakn.graql.internal.gremlin.sets.EquivalentFragmentSets.fragmentSetOfType;
+
 /**
+ * A query can use a shortcut edge traversal when the following criteria are met:
```
fppt: Best javadocs ever.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/sets/ShortcutFragmentSet.java
```diff
@@ -19,24 +19,126 @@
...
+        VarName casting = castingFragmentSet.casting();
+
+        RolePlayerFragmentSet rolePlayerFragmentSet = findRolePlayerFragmentSet(fragmentSets, casting);
+
+        // Try and get type of relation
+        Optional<IsaFragmentSet> relIsaFragment = fragmentSetOfType(IsaFragmentSet.class, fragmentSets)
+                .filter(isaFragmentSet -> isaFragmentSet.instance().equals(relation))
+                .findAny();
+
+        Optional<TypeLabel> relType = relIsaFragment
```
fppt: I had to stare at this for a while before I finally understood it and even now I am still not 100% sure.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/sets/ShortcutFragmentSet.java
```diff
@@ -19,24 +19,126 @@
...
+        if (castingIsaFragment.isPresent() && !roleType.isPresent()) {
+            return false;
+        }
+
+        // When the meta role-type is specified, it's the same as not specifying the role at all
+        if (roleType.isPresent() && roleType.get().equals(Schema.MetaSchema.ROLE.getLabel())) {
+            roleType = Optional.empty();
+        }
+
+        // We can't use the shortcut edge in the presence of role-type hierarchies, because we don't know precisely
```
fppt: bummer and I am not sure how to do indexing on sets. If we could do that then the shortcut edges could contain all of the role/relation types

grakn-graql/src/test/java/ai/grakn/graql/internal/gremlin/GraqlMatchers.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.hamcrest.Description;
+import org.hamcrest.FeatureMatcher;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeDiagnosingMatcher;
+
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+/**
```
fppt: Nitpick: Bad javadocs are bad.

grakn-graql/src/test/java/ai/grakn/graql/internal/gremlin/GraqlMatchers.java
```diff
@@ -0,0 +1,58 @@
...
+ * @author Felix Chapman
+ */
+public class GraqlMatchers {
```
fppt: This whole class is a bit difficult to follow. Can you provide some context?

grakn-graql/src/test/java/ai/grakn/graql/internal/gremlin/GraqlTraversalTest.java
```diff
@@ -236,58 +223,55 @@ public void makeSureTypeIsCheckedBeforeFollowingAShortcut() {
...
 
+        // I know this is horrible, unfortunately I can't think of a better way...
+        // The issue is that some things we want to inspect are not public, mainly:
+        // 1. The variable name assigned to the casting
+        // 2. The shortcut fragment classes
+        // Both of these things should not be made public if possible, so I see this regex as the lesser evil
         assertThat(graqlTraversal, anyOf(
-                is(traversal(xShortcutY)),
-                is(traversal(yShortcutX))
+                matches("\\{\\$x-\\[shortcut:\\$.*]->\\$y}"),
```
fppt: Ewwww

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/AbstractFragment.java
```diff
@@ -30,14 +30,15 @@
     // TODO: Find a better way to represent these values (either abstractly, or better estimates)
     static final long NUM_INSTANCES_PER_TYPE = 100;
     static final long NUM_INSTANCES_PER_SCOPE = 100;
-    static final long NUM_RELATION_PER_CASTING = 10;
-    static final long NUM_SHORTCUT_EDGES_PER_INSTANCE = 10;
+    static final long NUM_RELATIONS_PER_CASTING = 10;
```
aelred: Definitely needs to be a better method! @duckofyork is thinking about it I think.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/OutShortcutFragment.java
aelred: It was similar enough that PMD complained. That was part of the reason I extracted `applyTypeLabelToTraversal`. I don't think extracting any more makes sense. It would have to be a higher-order function accepting `GraphTraversal#inE`/`GraphTraversal#outE`, which I feel is overkill for when something is only duplicated once.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/sets/ShortcutFragmentSet.java
aelred: Is it important that these properties are indexed? I don't use these as starting points for queries like most other indexed properties.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/Fragments.java
```diff
@@ -168,4 +176,13 @@ public static Fragment resourceIndex(VarName start, TypeLabel typeLabel, Object
...
         return traversal.union(__.not(__.has(TYPE.name())), __.repeat(__.in(SUB.getLabel())).emit()).unfold();
     }
+
+    static String displayOptionalTypeLabel(Optional<TypeLabel> typeLabel) {
+        return typeLabel.map(label -> " " + typeLabelToString(label)).orElse("");
+    }
+
+    static void applyTypeLabelToTraversal(
+            GraphTraversal<Vertex, Edge> traversal, Schema.EdgeProperty property, Optional<TypeLabel> typeLabel) {
+        typeLabel.ifPresent(label -> traversal.has(property.name(), label.getValue()));
```
aelred: Sounds fun 👍

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/sets/ShortcutFragmentSet.java
aelred: I've cleared it up a little now.

grakn-graql/src/test/java/ai/grakn/graql/internal/gremlin/GraqlMatchers.java
aelred: All better now!

aelred:
@fppt maybe a bit less picky

PR #1536, opened 2017-05-05
More meaningful log messages

kasper-piskorski:


grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/Relation.java
```diff
@@ -106,7 +106,10 @@ private Relation(Relation a) {
 
     @Override
     public String toString(){
-        return (getType() != null? getType().getLabel() : "") + getRelationPlayers().toString();
+        String relationString = (isUserDefinedName()? "$" + getVarName().getValue() + " ": " ") +
```
aelred: `VarName#toString` will already prepend the `$`, so this can just be `getVarName() + " "`.

PR #1539, opened 2017-05-08
Change meta data from strings to ints - BIG WIP

fppt:


kasper-piskorski:
Should I expect to be summoned to propagate some changes?

PR #1540, opened 2017-05-08
Fix problems when unifying castings with metaroles

kasper-piskorski:
Addresses the discuss issue:
https://discuss.grakn.ai/t/inference-kicks-in-later/284/3


grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/Relation.java
```diff
@@ -640,13 +640,14 @@ public void unify(Unifier mappings) {
         Set<Pair<RelationPlayer, RelationPlayer>> rolePlayerMappings = new HashSet<>();
 
         //establish compatible castings for each parent casting
-        Map<RelationPlayer, Collection<RelationPlayer>> compatibleMappings = new HashMap<>();
+        Multimap<RelationPlayer, RelationPlayer> compatibleMappings = HashMultimap.create();
```
aelred: Wonderful guava!

PR #1544, opened 2017-05-09
GraqlController INSERT support

alexandraorth:


grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
```diff
@@ -95,39 +103,68 @@ public GraqlController(EngineGraknGraphFactory factory, Service spark) {
...
+    @ApiImplicitParams({
+            @ApiImplicitParam(name = KEYSPACE,    value = "Name of graph to use", required = true, dataType = "string", paramType = "query"),
+            @ApiImplicitParam(name = QUERY,       value = "Insert query to execute", required = true, dataType = "string", paramType = "body"),
+    })
+    private Json executeGraqlPOST(Request request, Response response){
+        String queryString = mandatoryBody(request);
+        String keyspace = mandatoryQueryParameter(request, KEYSPACE);
+
+        try(GraknGraph graph = factory.getGraph(keyspace, WRITE)){
+            Query<?> query = graph.graql().materialise(false).infer(false).parse(queryString);
```
aelred: What happens if the query has a syntax error or similar?

grakn-test/src/test/java/ai/grakn/test/engine/controller/GraqlControllerTest.java
```diff
@@ -70,11 +73,15 @@
...
 import static org.mockito.ArgumentMatchers.booleanThat;
+import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+//TODO Run in name order until TP Bug #13730 Fixed
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
aelred: JUnit magic :o

grakn-test/src/test/java/ai/grakn/test/engine/controller/GraqlControllerTest.java
```diff
@@ -126,401 +133,539 @@ public static void shutdown(){
...
+    public void GETGraqlComputeWithHALType_ResponseStatusIs406(){
         String query = "compute count in person;";
-        Response response = sendMatch(query, APPLICATION_HAL);
+        Response response = sendGET(query, APPLICATION_HAL);
 
         assertThat(response.statusCode(), equalTo(406));
         assertThat(exception(response), containsString(APPLICATION_HAL));
     }
 
+    //TODO Prefix with Z to run last until TP Bug #13730 Fixed
```
aelred: ok

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
alexandraorth: There is a registered exception handler in line 94 - it will return a 400 to the client with the syntax error

grakn-engine/src/main/java/ai/grakn/engine/controller/ConceptController.java
```diff
@@ -167,7 +167,7 @@ static void validateRequest(Request request){
      */
     static String mandatoryRequestParameter(Request request, String parameter){
         return Optional.ofNullable(request.params(parameter)).orElseThrow(() ->
-                new GraknEngineServerException(400, MISSING_MANDATORY_PARAMETERS, parameter));
+                new GraknEngineServerException(400, MISSING_MANDATORY_QUERY_PARAMETERS, parameter));
```
alexandraorth: You're correct, that's a bit confusing. Should make it "request" parameters

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
```diff
@@ -95,39 +103,68 @@ public GraqlController(EngineGraknGraphFactory factory, Service spark) {
             @ApiImplicitParam(name = KEYSPACE,    value = "Name of graph to use", required = true, dataType = "string", paramType = "query"),
             @ApiImplicitParam(name = QUERY,       value = "Match query to execute", required = true, dataType = "string", paramType = "query"),
             @ApiImplicitParam(name = INFER,       value = "Should reasoner with the current query.", required = true, dataType = "boolean", paramType = "query"),
-            @ApiImplicitParam(name = MATERIALISE, value = "Should reasoner materialise results with the current query.", required = true, dataType = "boolean", paramType = "query")
```
alexandraorth: Need to add this back in.

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
marco-scoppetta: Will it also catch validation exceptions upon commit?

alexandraorth: Validation exceptions will be caught by an error handlers, yes, but they won't return their own code. Ill make that change now.

PR #1547, opened 2017-05-09
Add support for dates and date-times to Graql

aelred:


grakn-graql/src/main/java/ai/grakn/graql/internal/parser/QueryVisitor.java
```diff
@@ -726,6 +739,8 @@ private Order getOrder(TerminalNode order) {
                 return ResourceType.DataType.STRING;
             case "boolean":
                 return ResourceType.DataType.BOOLEAN;
+            case "date":
```
fppt: Just realised, instead of an ugly switch you might be able to iterate over `ResourceType.DataType.SUPPORTED_TYPES` and compare the `dataType.getText()` with the lower case of the keyset. Would also mean you would never need to worry about amending this part of code potentially.

grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/DataTypeProperty.java
```diff
@@ -70,6 +70,8 @@ public String getProperty() {
             return "long";
         } else if (datatype == ResourceType.DataType.STRING) {
             return "string";
+        } else if (datatype == ResourceType.DataType.DATE) {
```
fppt: Same here. Might be able to iterate over the supported types.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/predicate/ComparatorPredicate.java
```diff
@@ -62,6 +65,13 @@
                 value = ((Integer) value).longValue();
             }
 
+            this.originalValue = Optional.of(value);
+
+            // Convert dates to epoch time to reflect how they are stored
+            if (value instanceof LocalDateTime) {
+                value = ResourceType.DataType.DATE.getPersistenceValue((LocalDateTime) value);
```
fppt: nice you realised what `getPersistenceValue ` was for :)

aelred: Yeah! Oh, just realised: I should do this for _all_ values, not just `LocalDateTime` right?

grakn-graql/src/main/java/ai/grakn/graql/internal/parser/QueryVisitor.java
```diff
@@ -717,17 +730,6 @@ private Order getOrder(TerminalNode order) {
...
-            case "double":
-                return ResourceType.DataType.DOUBLE;
-            case "string":
-                return ResourceType.DataType.STRING;
-            case "boolean":
-                return ResourceType.DataType.BOOLEAN;
-            default:
-                throw new RuntimeException("Unrecognized datatype " + datatype.getText());
-        }
+        return QueryParser.DATA_TYPES.get(datatype.getText());
```
alexandraorth: What will happen now when the datatype is not in the map? Is this prevented by ANTLR?

aelred: It will return `null`! As long as the `DATA_TYPES` list is complete this will never happen.

I added the `throws` in the old method because there had to be a default case. I guess we could add `assert(result != null)` for the same effect here.

PR #1548, opened 2017-05-09
Optimise query plan by removing fragment with fixed cost

duckofyork:


grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -95,8 +95,16 @@ public static GraqlTraversal createTraversal(
...
+        Plan initialPlan = Plan.base();
+        Set<EquivalentFragmentSet> fragmentSets = query.getEquivalentFragmentSets().stream().filter(fragmentSet -> {
+            if (fragmentSet.fragments().size() == 1) {
+                Fragment fragment = fragmentSet.fragments().iterator().next();
+                if (fragment.hasFixedFragmentCost()) {
+                    return !initialPlan.tryPush(fragment);
+                }
+            }
+            return true;
+        }).collect(Collectors.toSet());
```
aelred: We should have a comment to explain what this does, or extract it to a method with a good name.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/Fragment.java
```diff
@@ -99,4 +99,8 @@ default boolean isStartingFragment() {
     }
 
     double fragmentCost(double previousCost);
+
+    default boolean hasFixedFragmentCost() {
+        return false;
+    }
```
aelred: I know I'm being a hypocrite, since the method above doesn't have it - but we should put some Javadoc for this method :P

duckofyork:
please retest this please

PR #1549, opened 2017-05-09
Add a jUnit RunListener to measure time for each test

bolerio:
The listener creates a CSV file under grakn-test/test-timings.csv which contains for each package, class, method name, the time it took to run it. It only records successful tests. 

To make it work without falling into a circular dependency issue, the surefire maven plugin configuration was moved from the main pom.xml to the grakn-test pom.xml.


grakn-test/pom.xml
```diff
@@ -198,6 +197,23 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>2.19.1</version>
+                <configuration>
+                    <systemProperties>
+                        <logback.configurationFile>${main.basedir}/conf/test/logback-test.xml</logback.configurationFile>
```
alexandraorth: This is how we specify the logback file in the other poms: 
`<logback.configurationFile>../conf/test/logback-test.xml</logback.configurationFile>`

bolerio: I have this in the surefire plugin properties: 
`<logback.configurationFile>${main.basedir}/conf/test/logback-test.xml</logback.configurationFile>`

alexandraorth: Could it be using the wrong `${main.basedir}`? Maybe the base dir of the test package instead of the room folder?

bolerio: It's define as `<main.basedir>${project.parent.basedir}</main.basedir>`

PR #1551, opened 2017-05-10
Convert String Metadata to Integers

fppt:


fppt:
@duckofyork your wisdom is needed on the analytics tests.

fppt:
@aelred be picky.

fppt:
please retest this please

grakn-graph/src/test/java/ai/grakn/graph/internal/GraknGraphTest.java
```diff
@@ -128,19 +131,26 @@ public void whenGettingSubTypesFromRootMeta_IncludeAllTypes(){
 
     @Test
     public void whenClosingReadOnlyGraph_EnsureTypesAreCached(){
-        assertThat(graknGraph.getCachedOntology().asMap().keySet(), is(empty()));
+        //assertThat(graknGraph.getCachedOntology().asMap().keySet(), is(empty()));
```
aelred: 🐐

grakn-core/src/main/java/ai/grakn/graph/admin/GraknAdmin.java
```diff
@@ -126,6 +126,15 @@
     //------------------------------------- Admin Specific Operations ----------------------------------
 
     /**
+     * Converts a Type Label into a type Id for this specific graph. Mapping labels to ids will differ between graphs
+     * so be sure to use the correct graph when performing the mapping.
+     *
+     * @param label The label to be converted to the id
+     * @return The matching type id or -1 if no such type exists
```
fppt: `null` breaks your traversals in graql and an `Optional` which is what it used to be led to inserting magic numbers later in your traversals anyway. Not a fan of magic number but since this is internal I think it's okay.

grakn-core/src/main/java/ai/grakn/util/Schema.java
```diff
@@ -145,11 +151,11 @@ public Class getClassType(){
...
     public enum ConceptProperty {
         //Unique Properties
-        TYPE_LABEL(String.class), INDEX(String.class), ID(String.class),
+        TYPE_LABEL(String.class), INDEX(String.class), ID(String.class), TYPE_ID(Integer.class),
 
         //Other Properties
-        TYPE(String.class), IS_ABSTRACT(Boolean.class), IS_IMPLICIT(Boolean.class), IS_SHARD(Boolean.class),
-        REGEX(String.class), DATA_TYPE(String.class), INSTANCE_COUNT(Long.class),
+        INSTANCE_TYPE_ID(Integer.class), IS_ABSTRACT(Boolean.class), IS_IMPLICIT(Boolean.class), IS_SHARD(Boolean.class),
+        REGEX(String.class), DATA_TYPE(String.class), INSTANCE_COUNT(Long.class), CURRENT_TYPE_ID(Integer.class),
```
fppt: `CURRENT_TYPE_ID ` ? You mean `TYPE_ID`? 

`TYPE_ID` - Is only stored in types and contains the integer id we assign to types. It is indexed to be unique.

`INSTANCE_TYPE_ID ` - Is only stored on instances and refers to the `TYPE_ID` of the instance's type. This is NOT indexed in any way. Hence why we have two properties.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -592,22 +555,34 @@ public RuleType putRuleType(TypeLabel label) {
...
+    private <T extends Type> T getType(TypeLabel label, Schema.BaseType baseType){
+        operateOnOpenGraph(() -> null); //Makes sure the graph is open
+
+        Type type = buildType(label, ()-> {
+            Integer id = convertToId(label);
+            if (id != -1) {
+                return getType(id);
+            } else {
+                return null;
+            }
```
fppt: See my comment above about turning it into an Optional. Happy to do it if I dont end up passing in magic number into the graql traversal.

grakn-core/src/main/java/ai/grakn/util/Schema.java
aelred: `CURRENT_TYPE_ID(Integer.class)` <- it's real, I'm not crazy :P

grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptCache.java
```diff
@@ -38,11 +38,11 @@
  * @author fppt
  *
  */
-class ComponentCache<V> {
+class ConceptCache<V> {
```
fppt: It can cache anything but at this stage it is ONLY used for caching concept data hence the name change. Just to provide a bit more context between `TxCache`, `GraphCache`, and `ConceptCache`

grakn-graph/src/main/java/ai/grakn/graph/internal/GraphCache.java
```diff
@@ -0,0 +1,120 @@
...
+        cachedTypes.put(label, type);
+    }
+
+    /**
+     * Caches a label so we can map type labels to type ids. This is necesssary so we can make fast indexed lookups.
+     *
+     * @param label The label of the type to cache
+     * @param id The id of the type to cache
+     */
+    void cacheLabel(TypeLabel label, Integer id){
```
fppt: Ya I know :(  I think I have a TODO marked somewhere to do just that.

grakn-graph/src/main/java/ai/grakn/graph/internal/TxCache.java
```diff
@@ -18,41 +18,50 @@
...
- *      concepts that have had edges added/deleted,
- *      edge cases, for example, relationship where a new role player is added.
+ *     Caches Transaction specific data this includes:
+ *     <ol>
+ *         <li>Validation Concepts - Concepts which need to undergo validation.</li>
+ *         <li>Built Concepts -  Prevents rebuilding when the same vertex is encountered</li>
+ *         <li>The Ontology - Optimises validation checks by preventing db read. </li>
+ *         <li>Type Labels - Allows mapping type labels to type Ids</li>
+ *         <li>Transaction meta Data - Allows transactions to function in different ways</li>
+ *     <ol/>
```
fppt: That was my thinking as well.

grakn-graph/src/test/java/ai/grakn/graph/internal/GraknGraphTest.java
fppt: Ooops

grakn-core/src/main/java/ai/grakn/util/Schema.java
fppt: Ahhhh my shame :( . That stores the current number of types we have in the graph and it is persisted ONLY on the meta type. Thats how we know what to use as the next id.

aelred: Thanks, I understand! I can't see it being used anywhere, but that might be GitHub's fault.

fppt:
please retest this please - ai.grakn.test.graql.analytics.StatisticsTest.ai.grakn.test.graql.analytics.StatisticsTest

grakn-core/src/main/java/ai/grakn/util/Schema.java
fppt: It's used in `AbstractGraknGraph#getNextTypeId`: 

```
private int getNextTypeId(){
        TypeImpl<?, ?> metaConcept = (TypeImpl<?, ?>) getMetaConcept();
        Integer currentValue = metaConcept.getProperty(Schema.ConceptProperty.CURRENT_TYPE_ID);
        if(currentValue == null){
            currentValue = Schema.MetaSchema.values().length + 1;
        } else {
            currentValue = currentValue + 1;
        }
        //Vertex is used directly here to bypass meta type mutation check
        metaConcept.getVertex().property(Schema.ConceptProperty.CURRENT_TYPE_ID.name(), currentValue);
        return currentValue;
    }
```

PR #1553, opened 2017-05-10
BUG #13743 - optional arguments -a and -s were not optional in graql shell

sheldonkhall:


sheldonkhall:
@aelred can you review this one

grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
```diff
@@ -216,12 +216,20 @@ public static void runShell(String[] args, String version, String historyFilenam
...
-            } catch (IOException e) {
-                throw new RuntimeException(e);
+                Optional<Integer> activeTasks = Optional.empty();
+                Optional<Integer> batchSize = Optional.empty();
+                if (cmd.hasOption("a")) {
+                    activeTasks = Optional.of(Integer.parseInt(cmd.getOptionValue("a")));
+                }
+                if (cmd.hasOption("s")) {
+                    batchSize = Optional.of(Integer.parseInt(cmd.getOptionValue("s")));
+                }
```
aelred: I think this is up to personal preference, but I'd do it this way:
```
Optional<Integer> activeTasks = Optional.ofNullable(cmd.getOptionValue("a")).map(Integer::parseInt);
Optional<Integer> batchSize = Optional.ofNullable(cmd.getOptionValue("s")).map(Integer::parseInt);
```

PR #1557, opened 2017-05-10
Fix bug with output for batch loading queries

aelred:


alexandraorth:
@aelred @fppt I just want to reiterate once more if you can explicitly check all of the requested reviewers have approved before merging. This one is obviously fine, because it is so small, but it can be a bit soul-crushing.

fppt:
@alexandraorth as you know I usually do but for such small PRs I think waiting on both reviewers is overkill.

PR #1558, opened 2017-05-10
REST Delete query support

alexandraorth:


grakn-core/src/main/java/ai/grakn/util/ErrorMessage.java
```diff
@@ -168,7 +168,7 @@
     //Server Errors
     UNAVAILABLE_TASK_CLASS("Could not find task class [%s]"),
     MISSING_MANDATORY_REQUEST_PARAMETERS("Missing mandatory query parameter [%s]"),
-    MISSING_POST_REQUEST_BODY("Empty POST body- it should contain the Graql query to be executed."),
+    MISSING_REQUEST_BODY("Empty POST body- it should contain the Graql query to be executed."),
```
aelred: I see the word `POST` here still!

alexandraorth: Very good catch :1st_place_medal:

alexandraorth:
@aelred please re-review

PR #1561, opened 2017-05-11
Reorganise query iterators

kasper-piskorski:


grakn-graql/src/main/java/ai/grakn/graql/internal/query/QueryAnswer.java
```diff
@@ -59,7 +59,9 @@ public QueryAnswer(Map<VarName, Concept> m){
     }
 
     @Override
-    public String toString(){ return map.toString();}
+    public String toString(){
+        return map.entrySet().stream().map(e -> "[" + e.getKey() + "/" + e.getValue().getId() + "]").collect(Collectors.joining());
```
fppt: I love `Collectors.joining()` 👍

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/QueryAnswerIterator.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * check whether answers available, if answers not fully computed compute more answers
+     * @return true if answers available
+     */
+    @Override
+    public boolean hasNext() {
+        if (answerIterator.hasNext()) return true;
+            //iter finished
+        else {
```
fppt: The above `if` results in a `return` making this `else` block unneeded. Getting rid of the `else` will make the code more readable.

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerAtomicQueryIterator.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        //mark as visited and hence not admissible
+        if (ruleIterator.hasNext()) subGoals.add(query);
+    }
+
+    @Override
+    public boolean hasNext() {
+        if (queryIterator.hasNext()) return true;
+        else{
```
fppt: Same here, no need for the else block. 

As a future nitpick if you end up having lots of implementations of `hasNext` and if all the implementations have a similar structure then it maybe possible to get rid of all the boiler plate `ifs` by adding a common method in the super class to handle that boiler plate code. Something like:

```
public boolean hasNextHelper(Iterator it,  Supplier<Boolean> checker, Supplier<Boolean> finalCheck){
    if(it.hasNext()) return true;
    return checker.get() && finalCheck.get();
}
```

But that's a nitpick only if we see more of this.

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerQueryImpl.java
```diff
@@ -178,7 +175,7 @@ private boolean isTransitive() {
     }
 
     boolean isAtomic() {
-        return selectAtoms().size() == 1;
+        return getAtoms().stream().filter(Atomic::isSelectable).count() == 1;
```
fppt: Question: Does `getAtoms()` get very large? I ask because I often see `isAtomic()` and if calling this means iterating over a large set over and over again then performance may get affected.

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerQueryImplIterator.java
```diff
@@ -41,43 +43,43 @@
...
-        atomicQueryIterator = isAtomic? Collections.emptyIterator() : q.iterator(subGoals, cache);
-        queryIterator = isAtomic? q.iterator(subGoals, cache) : Collections.emptyIterator();
+        this.atomicQueryIterator = new ReasonerAtomicQuery(topAtom).iterator(new QueryAnswer(), subGoals, cache);
+        this.queryPrime = ReasonerQueries.prime(query, topAtom);
     }
 
     @Override
     public boolean hasNext() {
         if (queryIterator.hasNext()) return true;
         else {
```
fppt: Once again the `else` is not needed. Check all the implementations of `ReasonerQueryIterator#hasNext` you may have the possibility of getting rid of a lot of boilerplate stuff.

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerQueryImpl.java
kasper-piskorski: For big queries yes. However `getAtoms()` just returns a member variable and the `isSelectable` check is cheap. I think that we should have the atomicity recognition at the pattern, not atom level at some point.

PR #1563, opened 2017-05-11
Rule applicability fix for ontological type atoms

kasper-piskorski:


grakn-test/src/test/java/ai/grakn/test/graql/reasoner/AtomicTest.java
```diff
@@ -388,6 +388,23 @@ public void testRuleApplicability_TypeRelation(){
...
+        GraknGraph graph = ruleApplicabilitySet.graph();
+        String typeString = "{$x sub relation;}";
+        String typeString2 = "{$x relates role1;}";
+        String typeString3 = "{$x plays role1;}";
+        String typeString4 = "{$x has res1;}";
+        TypeAtom type = (TypeAtom) new ReasonerAtomicQuery(conjunction(typeString, graph), graph).getAtom();
+        TypeAtom type2 = (TypeAtom) new ReasonerAtomicQuery(conjunction(typeString2, graph), graph).getAtom();
+        TypeAtom type3 = (TypeAtom) new ReasonerAtomicQuery(conjunction(typeString3, graph), graph).getAtom();
+        TypeAtom type4 = (TypeAtom) new ReasonerAtomicQuery(conjunction(typeString4, graph), graph).getAtom();
+        assertEquals(0, type.getApplicableRules().size());
```
fppt: A better way for checking that sets are empty is to do `assertThat(type.getApplicableRules(), empty())` where `empty` is from `org.hamcrest.Matchers.empty()`.

This means that when the test fails just by looking at the logs you have more contextual information rather than just 2 number not being equal.

PR #1567, opened 2017-05-11
Print exceptions in REST controllers

alexandraorth:


grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -237,6 +238,7 @@ private static void handleGraknServerError(Exception exception, Response respons
      * @param response response to the client
      */
     private static void handleInternalError(Exception exception, Response response){
```
fppt: As far as I can tell the only difference between `handleGraknServerError` and `handleInternalError` is setting the status code. If we end up manually setting more status codes then it may be worth creating a generic method. This is only a future possible nitpick though.

PR #1572, opened 2017-05-12
Engine Graph Mutation don't re-throw exceptions

alexandraorth:


alexandraorth:
please retest this please

PR #1579, opened 2017-05-15
Fix Graql output to use `val` instead of `value`

aelred:


aelred:
please retest this please

PR #1582, opened 2017-05-15
Do not load test graphs on titan with tinker requirement 

kasper-piskorski:


kasper-piskorski:
please retest this please

PR #1594, opened 2017-05-16
Better Centralised Sharding

fppt:
- Use Redis To Keep Track of counts
- Engine now triggers sharding explicitly


fppt:
please retest this please - VM failure

conf/main/grakn.properties
```diff
@@ -39,6 +39,10 @@ server.static-file-dir=../assets/
 #RESTLoader Config
 loader.repeat-commits=5
 
+#Redis Config
+redis.url=localhost
+redis.port=7000
```
aelred: How do things work when there's a cluster?

grakn-core/src/main/java/ai/grakn/graph/admin/GraknAdmin.java
```diff
@@ -160,9 +160,15 @@
     /**
      * Updates the counts of all the types
      *
-     * @param typeCounts The types and the changes to put on their counts
+     * @param conceptCounts The concepts and the changes to put on their counts
      */
-    void updateTypeShards(Map<TypeLabel, Long> typeCounts);
+    void updateTypeCounts(Map<ConceptId, Long> conceptCounts);
```
aelred: We should rename to `updateConceptCounts`

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/util/TaskConfigReader.java
```diff
@@ -0,0 +1,81 @@
...
+ * <p>
+ *     This class is used to extract information from Task configs which is needed by
+ *     {@link ai.grakn.engine.postprocessing.PostProcessingTask},
+ *     {@link ai.grakn.engine.postprocessing.UpdatingInstanceCountTask} and
+ *     {@link ai.grakn.engine.loader.MutatorTask}
+ * </p>
+ *
+ * @author fppt
+ */
+public class TaskConfigReader {
```
alexandraorth: I don't think these functions should be in a separate class. Other than the "keyspace" one, they are only used in one place. And Its not necessary that the user places the keyspace in the root of the configuration object. 

Even if we do keep the class, we should remove `getCountUpdatingJobs` and `getPostProcessingJobs`, which are only used in their respective tasks.

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/UpdatingInstanceCountTask.java
```diff
@@ -41,31 +45,92 @@
  *
  * @author fppt
  */
-public class UpdatingInstanceCountTask extends AbstractLockingTask {
-
-    public static final String LOCK_KEY = "/updating-instance-count-lock";
+public class UpdatingInstanceCountTask implements BackgroundTask {
+    public static final RedisConnection redis = RedisConnection.getConnection();
```
alexandraorth: What happens if Redis is not running?

grakn-engine/src/main/java/ai/grakn/engine/tasks/connection/RedisConnection.java
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Returns and possibly initliases a connection pool to the redis server.
+     *
+     * @return a connction to the redis server.
+     */
+    public static synchronized RedisConnection getConnection(){
+        if(redis == null) redis = new RedisConnection();
+        return redis;
+    }
```
aelred: This is `synchronized`, but does it definitely do the right thing? This will mean only one thing can establish a connection at once, but it is still possible for two things to call `adjustCount` simultaneously.

grakn-test/src/test/java/ai/grakn/test/engine/tasks/connection/RedisConnectionTest.java
```diff
@@ -0,0 +1,94 @@
...
+ * <p>
+ *    Tests that the connection to redis is working and that data can be persisted and retrieved into/from redis.
+ * </p>
+ *
+ * @author fppt
+ */
+public class RedisConnectionTest {
+
+    @ClassRule
+    public static final EngineContext engine = EngineContext.startInMemoryServer();
```
alexandraorth: I would recommend creating an EngineContext.startRedisServer(); method. This test does not need all of engine running.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -1034,20 +1032,22 @@ private void copyRelation(ResourceImpl main, ResourceImpl<?> other, RelationImpl
...
-               long newValue = type.getInstanceCount() + entry.getValue();
-               if(newValue < shardingFactor) {
-                   type.setInstanceCount(type.getInstanceCount() + entry.getValue());
-               } else {
-                   //TODO: Maintain the count properly. We reset so we can split with simpler logic
-                   type.setInstanceCount(0L);
-                   type.createShard();
-               }
+               ConceptImpl type = getConcept(entry.getKey());
+               type.setShardCount(type.getShardCount() + entry.getValue());
```
aelred: Nitpick alert! We should call this `concept` rather than `type`.

conf/main/grakn.properties
fppt: Redis would only be on one machine I would imagine. Similar to kafka?

grakn-core/src/main/java/ai/grakn/graph/admin/GraknAdmin.java
fppt: agreed

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessingTask.java
```diff
@@ -46,7 +43,7 @@
  *     This task begins only if enough time has passed (configurable) since the last time a job was added.
  * </p>
  *
- * @author alexandraorth
+ * @author alexandraorth, fppt
```
alexandraorth: yay, shared responsibility!

grakn-engine/src/main/java/ai/grakn/engine/tasks/connection/RedisConnection.java
fppt: Ahh good spot. I made this change because of PMD

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
fppt: nitpick accepted

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/UpdatingInstanceCountTask.java
fppt: Crash. Redis is needed by default

grakn-test/src/test/java/ai/grakn/test/engine/tasks/connection/RedisConnectionTest.java
fppt: If we don't start redis and an updating count task starts we will see a lot of exceptions which makes debugging harder I think. 

Redis is cheap to start so I think it's okay to have it starting by default.

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/UpdatingInstanceCountTask.java
alexandraorth: We should try and handle it a bit better. If it throws a runtime exception within a task all of the `UpdatingInstanceCountTask`s will be marked as `FAILED`, which is fine, but we need to make sure the error is descriptive enough that people know there is an issue with Redis

alexandraorth:
We shouldn't accept this until we have Redis starting from the dist. I think once this goes running the dist will crash.

fppt:
Yes you right Alex. I missed that part. Will add redis to the distribution next.

grakn-test/src/test/java/ai/grakn/test/engine/tasks/connection/RedisConnectionTest.java
alexandraorth: Yeah, when engine starts we definitely need reds running. I actually meant that if we have a separate Redis context, you wouldn't need to start the entirety of engine for this test that does not need engine.

fppt: Ahhh that's a good idea. Will note that as a todo. Low priority though because i think only one test uses redis but not engine.

fppt:
please retest this please- ai.grakn.test.engine.postprocessing.UpdatingInstanceCountTaskTest.whenShardingThresholdIsBreached_ShardTypes

fppt:
@alexandraorth @aelred can I get another review please. The changes are all in and I think this is "good enough" to go in.

On a related note once this goes in @alexandraorth and @burukuru we going to need to put Redis on the cluster.

fppt:
I have performed a manual load and everything seems to be working

alexandraorth:
@fppt you checked the shard sizes as well?

fppt:
@alexandraorth nope only made sure nothing broke. Did not check the shard sizes.

fppt:
please retest this please - timeout

fppt:
Tested and checked that shards are being created. Seems to be better than the previous approach but still not prefect. I suspect the missing piece is to make the sharding lock index specific.

PR #1599, opened 2017-05-17
Allow cluster query params in any order

aelred:


grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractComputeQuery.java
```diff
@@ -187,4 +187,22 @@ public String toString() {
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        AbstractComputeQuery<?> that = (AbstractComputeQuery<?>) o;
+
+        if (!graph.equals(that.graph)) return false;
+        return subTypeLabels.equals(that.subTypeLabels);
+    }
```
duckofyork: This method would tell you any two path queries (or any other type of query except cluster) are the same.

aelred: That's true. Does that mean we should implement `equals` for every query type?

duckofyork: Do you need all of them?

PR #1608, opened 2017-05-17
Split post processing into two methods

alexandraorth:
Stepping stone on the way to fixing post processing race conditions.

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessingTask.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessingTask.java
@@ -63,8 +63,8 @@ public class PostProcessingTask implements BackgroundTask {
      */
     @Override
     public boolean start(Consumer<TaskCheckpoint> saveCheckpoint, TaskConfiguration configuration) {
-        runPostProcessingMethod(configuration, Schema.BaseType.CASTING, this::runCastingFix);
-        runPostProcessingMethod(configuration, Schema.BaseType.RESOURCE, this::runResourceFix);
+        runPostProcessingMethod(configuration, Schema.BaseType.CASTING, this::duplicateCastingsExist, this::runCastingFix);
+        runPostProcessingMethod(configuration, Schema.BaseType.RESOURCE, this::duplicateResourcesExist, this::runResourceFix);
 
         return true;
     }
@@ -101,13 +101,15 @@ public class PostProcessingTask implements BackgroundTask {
     /**
      * Main method which attempts to run all post processing jobs.
      *
-     * @param postProcessingMethod The post processing job.
+     * @param duplicatesExistMethod Method to determine if there are duplicates to be post processed.
+     *                      Either {@link ai.grakn.engine.postprocessing.PostProcessingTask#duplicateCastingsExist(GraknGraph, String, Set)} or
+     *                      {@link ai.grakn.engine.postprocessing.PostProcessingTask#duplicateResourcesExist(GraknGraph, String, Set)}
+     * @param postProcessingMethod The post processing job to be executed if duplicates exist
      *                      Either {@link ai.grakn.engine.postprocessing.PostProcessingTask#runResourceFix(GraknGraph, String, Set)} or
      *                      {@link ai.grakn.engine.postprocessing.PostProcessingTask#runCastingFix(GraknGraph, String, Set)}.
-     *                      This then returns a function which will complete the job after going through validation
-
      */
     private void runPostProcessingMethod(TaskConfiguration configuration, Schema.BaseType baseType,
+                                         TriFunction<GraknGraph, String, Set<ConceptId>, Boolean> duplicatesExistMethod,
                                          TriFunction<GraknGraph, String, Set<ConceptId>, Boolean> postProcessingMethod){
 
         Map<String, Set<ConceptId>> allToPostProcess = conceptFromConfig(baseType, configuration);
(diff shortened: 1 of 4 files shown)


grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessingTask.java
```diff
@@ -197,4 +205,26 @@ private boolean runResourceFix(GraknGraph graph, String index, Set<ConceptId> co
...
         return graph.admin().fixDuplicateCastings(index, conceptIds);
     }
+
+    /**
+     * Check if there are duplicate castings for the given index
+     * @param graph Graph on which to apply the fixes
+     * @param index The unique index of the concept that may have duplicates
+     * @param conceptIds The conceptIds which may be the duplicates
+     */
+    private boolean duplicateCastingsExist(GraknGraph graph, String index, Set<ConceptId> conceptIds) {
```
fppt: Nitpick, considering this is used in exactly one place and it only has one line I would inline it.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -858,39 +858,62 @@ public void validVertex(Vertex vertex){
...
+
     /**
-     * Merges the provided duplicate castings.
-     *
-     * @param castingVertexIds The vertex Ids of the duplicate castings
-     * @return if castings were merged, a commit is required and the casting index exists
+     * Returns the duplicates of the given concept
+     * @param mainConcept primary concept - this one is returned by the index and not considered a duplicate
+     * @param conceptVertexIds Set of Ids containing potential duplicates of the main concept
+     * @return true if the given set of Ids contains concepts that are duplicates of the main concept
```
fppt: Java docs are wrong. This method does not return true.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -858,39 +858,62 @@ public void validVertex(Vertex vertex){
...
+    /**
+     * Given an index, get the concept associated with it
+     * @param index retrieve the concept associated with this index
+     * @return Concept representing the vertex at the given index
+     */
+    private ConceptImpl getMainConcept(String index){
         //This is done to ensure we merge into the indexed casting. Needs to be cleaned up though
-        CastingImpl mainCasting = getConcept(Schema.ConceptProperty.INDEX, index);
-        duplicated.remove(mainCasting);
+        return getConcept(Schema.ConceptProperty.INDEX, index);
```
fppt: Since this is one line I would once again inline it. It does not have many places using it.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -858,39 +858,62 @@ public void validVertex(Vertex vertex){
...
+        duplicateRelations.forEach(relation -> ((ConceptImpl) relation).deleteNode());
 
-            return true;
-        }
+        //Restore the index
+        String newIndex = mainCasting.getIndex();
+        mainCasting.getVertex().property(Schema.ConceptProperty.INDEX.name(), newIndex);
 
-        return false;
+        return true;
```
fppt: This method looks to always returns `true` now which means it should either return false at some stage or just be void. I would say you should check if `duplicated` contains anything. If it doesn't then return false.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -958,45 +981,46 @@ private String getRelationIndex(RelationImpl relation){
...
+            otherResource.castings().forEach(ConceptImpl::deleteNode);
+            otherResource.deleteNode();
         }
 
-        return false;
+        //Restore the index
+        String newIndex = mainResource.getIndex();
+        mainResource.getVertex().property(Schema.ConceptProperty.INDEX.name(), newIndex);
+
+        return true;
```
fppt: Again this only returns true now. Same thing check if there are duplicates to actually merge or make it return void.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
alexandraorth: Yeah, 100% agreed. Especially if the return value means that something has been modified and you should commit.

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessingTask.java
alexandraorth: I think that I disagree on this one. That would mean that we are externalizing logic behind checking for duplicates exist to the PPTask itself, which is the only class that calls `duplicateCastingsExist`.

fppt: Not overly concerned about this one. Just a preference. Feel free to leave as is.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
alexandraorth: I also sort of disagree on this one. It's used in four places and if we inline it, we have to include that explanatory comment (of what it means to be the "main" concept) in each one.

PR #1610, opened 2017-05-18
Added super user to user management

bolerio:
This commit is a bit weird in that it's just some leftover changes that were done while implemented user authorization (access right management) which was moved out in a separate, enterprise edition module. The `admin` super user still needs unit tests.


bolerio:
!rtg

bolerio:
!rtg

bolerio:
please retest this please

bolerio:
I am going to merge this because the work on the enterprise module depends on it and it's small changes that should impact anyone else.

conf/main/grakn.properties
```diff
@@ -109,6 +109,7 @@ loader.threads=0
 #Authentication
 password.protected=false
 JWT.secret=Anzdz9l4TG8N5y18oaTDueqJQca1aX7loaz0U8Hz
+admin.password=grakn
```
alexandraorth: We should not be storing passwords in plain text. Im not sure what an alternative would be in this case. Perhaps we store it in the system graph and remove access to that graph?

conf/test/titan/grakn.properties
```diff
@@ -108,6 +108,7 @@ loader.threads=8
 #Authentication
 password.protected=false
 JWT.secret=Anzdz9l4TG8N5y18oaTDueqJQca1aX7loaz0U8Hz
+admin.password=grakn
```
alexandraorth: There is one configuration file that is missing this new property: `conf/test/orientdb/grakn.properties`.

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -203,12 +203,17 @@ private static void checkAuthorization(Service spark, Request request) {
             boolean authenticated;
             try {
                 if (request.headers("Authorization") == null || !request.headers("Authorization").startsWith("Bearer ")) {
-                    throw new GraknEngineServerException(400, "Authorization field in header corrupted or absent.");
+                    throw new GraknEngineServerException(401, "Authorization field in header corrupted or absent.");
```
alexandraorth: We should move this error message into `ErrorMessage` class in core.

grakn-engine/src/main/java/ai/grakn/engine/user/UsersHandler.java
```diff
@@ -31,6 +33,7 @@
  * @author Marco Scoppetta
  */
 public class UsersHandler {
```
alexandraorth: Should we merge `UsersHandler` and `SystemKeyspaceUsers`. Is there a case in which we will use `UsersHandler` on its own?

grakn-test/src/test/java/ai/grakn/test/engine/user/SuperUserTest.java
```diff
@@ -0,0 +1,70 @@
...
+    @ClassRule
+    public static final EngineContext engine = EngineContext.startInMemoryServer();
+    
+    @Test
+    public void testSuperuserPresent() {
+        Assert.assertNotNull(users.getUser(users.superUsername()));
+    }
+    
+    @Test
+    public void testSuperuserLogin() {
```
alexandraorth: Nitpick: I think we should all be moving towards the test naming scheme where we write `whatTestDoes_expectedResult`.

conf/main/grakn.properties
bolerio: Yes, that's an idea. And same would apply to the JWT.secret above, which is also in plain text. Passwords are not easy to manage - there are enterprise products just for that, especially since they proliferate for all sorts of components and it's hard to keep track of them. 

The problem with storing in system graph is that if you forget/lose the password, you can't login anymore at all and you have no means of recovery. If that one and only that one password is externalized, at least you can reset it somewhere and recover. The proper method would be probably to have some "security provider" that gives the admin password so a user can pull it from a secure external place.

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
bolerio: I'm not sure...if it's not a core error message, it shouldn't be in core. But it's a good idea them all in  one place, preferably in a ResourceBundle actually for i18n.

grakn-test/src/test/java/ai/grakn/test/engine/user/SuperUserTest.java
bolerio: :)

bolerio:
Thanks for the review @alexandraorth !

PR #1611, opened 2017-05-18
Change `Graql#var` to return `Var` instead of `VarPattern`

aelred:
These changes might be a bit hard to parse, so:
- `var("x")` will now return a `Var` (variable name), not a `VarPattern`
- `var("x").isa("movie");` will still return a `VarPattern`, by using the new interface `VarPatternBuilder`. `Var` and `VarPattern` both implement this interface. It provides a method that returns a `VarPattern`, as well as all the "property" methods such as `isa`, `rel` etc.
- The `Var` class now includes information about whether the variable is user-defined or not


aelred:
@fppt I'd appreciate better names for `PatternBuilder` and `VarPatternBuilder` if you have any ideas!

grakn-core/src/main/java/ai/grakn/graql/PatternBuilder.java
```diff
@@ -0,0 +1,38 @@
...
+ * along with Grakn. If not, see <http://www.gnu.org/licenses/gpl.txt>.
+ *
+ */
+
+package ai.grakn.graql;
+
+import javax.annotation.CheckReturnValue;
+
+/**
+ * An interface representing something that is or can be transformed into a {@link Pattern}.
```
fppt: What is `something` ?

grakn-engine/src/main/java/ai/grakn/engine/user/SystemKeyspaceUsers.java
```diff
@@ -126,9 +127,9 @@ public Json getUser(String username) {
         if (superUsername().equals(username)) {
             return Json.object(USER_NAME, username);
         }
-        try (GraknGraph graph = factory.getGraph(SystemKeyspace.SYSTEM_GRAPH_NAME, GraknTxType.READ)) {
-            VarPattern lookup = var("entity").isa(USER_ENTITY).has(USER_NAME, username);
-            VarPattern resource = var("property");
+        try (GraknGraph graph = factory.getGraph(SystemKeyspace.SYSTEM_GRAPH_NAME, GraknTxType.READ)) {VarPattern lookup = var("entity").isa(USER_ENTITY).has(USER_NAME, username);
+        Var resource = var("property");
```
fppt: Why is one a `Var` and the other a `VarPattern` ? Was this intentional?

aelred: Yes. `resource` is just a variable name - it is only used to identify concepts in a query. Whereas, `lookup` is a pattern applied to the variable `var("entity")`.

Making these kind of distinctions is the main reason I made this change. I'll send you some more details on the motivation if you like.

fppt:
Will merge but the following changes should follow:
1. Calling `pattern()` everywhere should not be needed this means sorting out the inheritance structure.
2. Get rid of `PatternBuilder` interface

PR #1612, opened 2017-05-18
Fix post processing race conditions

alexandraorth:
+ Delay post processing task start - This is needed becuase we must ensure that post processing runs only after the index has been set. The index is modifiable so long as duplicates are being created.
+ Make PP task aquire an index-based lock - Eliminate race conditions cause by post processing the same index at the same time.

### DIFF
--- a/conf/main/grakn.properties
+++ b/conf/main/grakn.properties
@@ -45,6 +45,7 @@ redis.port=6379
 
 #Background tasks Config
 backgroundTasks.time-lapse=300000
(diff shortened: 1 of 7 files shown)


alexandraorth:
please retest this please - `Unable to query GitHub for status of PullRequestorg.kohsuke.github.HttpException: Server returned HTTP response code: -1, message: 'null' for URL: https://api.github.com/repos/graknlabs/grakn/pulls/1612`

PR #1614, opened 2017-05-18
Unit test templating macros

alexandraorth:


grakn-graql/src/test/java/ai/grakn/graql/internal/template/macro/ConcatMacroTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+        concatMacro.apply(ImmutableList.of("true"));
+    }
+
+    @Test
+    public void applyConcatMacroToTwoOrMoreArguments_ResultIsNonNull(){
+        exception.expect(IllegalArgumentException.class);
+
+        concatMacro.apply(ImmutableList.of("true"));
+    }
```
aelred: This test doesn't test what it says it tests.

alexandraorth: Keen eyes.

alexandraorth:
@aelred Have made the changes

PR #1619, opened 2017-05-18
Fix bug where queries like `match $x isa title, val = $y;` didn't work

aelred:


fppt:
@sheldonkhall ping me once you have reviewed so I can merge.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/sets/ResourceIndexFragmentSet.java
```diff
@@ -96,7 +97,10 @@ private static void optimise(
 
     private static Stream<ValueFragmentSet> equalsValueFragments(Collection<EquivalentFragmentSet> fragmentSets) {
         return fragmentSetOfType(ValueFragmentSet.class, fragmentSets)
-                .filter(valueFragmentSet -> valueFragmentSet.predicate().equalsValue().isPresent());
```
sheldonkhall: I won't pretend that I understand what is happening here, but the tests cover the issue I had so I am happy based on that.

PR #1620, opened 2017-05-18
Make sure lock is released even if error is thrown

fppt:


grakn-engine/src/main/java/ai/grakn/engine/postprocessing/UpdatingInstanceCountTask.java
```diff
@@ -110,22 +110,25 @@ private static boolean updateShardCounts(String keyspace, ConceptId conceptId, l
      */
     private static void shardConcept(String keyspace, ConceptId conceptId){
         Lock engineLock = LockProvider.getLock(getLockingKey());
-        engineLock.lock(); //Try to get the lock
 
-        //Check if sharding is still needed. Another engine could have sharded whilst waiting for lock
-        if(updateShardCounts(keyspace, conceptId, 0)) {
+        try {
+            engineLock.lock(); //Try to get the lock
```
alexandraorth: you should acquire the lock outside of the try - if it throws an exception, you don't need to release

PR #1628, opened 2017-05-19
Fix several concurrency issues of analytics

duckofyork:
- Fix the bug where statistic method may fail when runing concurrently
- Fix the bug where degreeVertexProgram may fail when runing concurrently

### DIFF
--- a/grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/DegreeDistributionMapReduce.java
+++ b/grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/DegreeDistributionMapReduce.java
@@ -47,14 +47,16 @@ public class DegreeDistributionMapReduce extends GraknMapReduce<Set<String>> {
     public DegreeDistributionMapReduce() {
     }
 
-    public DegreeDistributionMapReduce(Set<TypeId> selectedTypeIds) {
+    public DegreeDistributionMapReduce(Set<TypeId> selectedTypeIds, String degreeLabel) {
         super(selectedTypeIds);
+        this.persistentProperties.put(DegreeVertexProgram.DEGREE, degreeLabel);
     }
 
     @Override
     public void safeMap(final Vertex vertex, final MapEmitter<Serializable, Set<String>> emitter) {
         if (selectedTypes.contains(Utility.getVertexTypeId(vertex))) {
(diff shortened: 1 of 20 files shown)


grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/DegreeDistributionMapReduce.java
```diff
@@ -47,14 +47,16 @@
     public DegreeDistributionMapReduce() {
     }
 
-    public DegreeDistributionMapReduce(Set<TypeId> selectedTypeIds) {
+    public DegreeDistributionMapReduce(Set<TypeId> selectedTypeIds, String degreeLabel) {
```
fppt: For the sake of clarity `degreeLabel` should be a `TypeLabel`. Makes the method more readable and easier to work with form the outsider's perspective.

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/MinMapReduce.java
```diff
@@ -34,14 +34,14 @@
...
-public class MinMapReduce extends GraknMapReduce<Number> {
+public class MinMapReduce extends StatisticsMapReduce<Number> {
 
     // Needed internally for OLAP tasks
     public MinMapReduce() {
     }
 
-    public MinMapReduce(Set<TypeId> selectedTypeIds, String resourceDataType) {
-        super(selectedTypeIds, resourceDataType);
+    public MinMapReduce(Set<TypeId> selectedTypeIds, String resourceDataType, String degreeKey) {
```
fppt: if `resourceDataType` does refer to the datatype of a resource type then this maybe should be `ResourceType.DataType dataType`. Again it just makes things more explicit.

duckofyork: Done. 
ResourceType.DataType is not serialisable so I still need to use string inside the implementation method.

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/DegreeDistributionMapReduce.java
duckofyork: Degree is a property on the vertex. So I guess I could change the variable name to sth like degreePropertyKey?

PR #1634, opened 2017-05-22
Use shortcut edges in Graql even in the presence of sub-types

aelred:
e.g.
`match (spouse: $x, spouse: $y);` will now use shortcut edges, even if `spouse` has sub-types.

`match ($x, $y) isa parenthood;` will use shortcut edges, even if `parenthood` has sub-types.

### DIFF
--- a/grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/Fragments.java
+++ b/grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/Fragments.java
@@ -51,14 +51,14 @@ public class Fragments {
 
     public static Fragment inShortcut(
             Var rolePlayer, Var edge, Var relation,
-            Optional<Set<TypeLabel>> roleTypes, Optional<TypeLabel> relationType) {
-        return new InShortcutFragment(rolePlayer, edge, relation, roleTypes, relationType);
+            Optional<Set<TypeLabel>> roleTypes, Optional<Set<TypeLabel>> relationTypes) {
+        return new InShortcutFragment(rolePlayer, edge, relation, roleTypes, relationTypes);
     }
 
     public static Fragment outShortcut(
             Var relation, Var edge, Var rolePlayer,
-            Optional<Set<TypeLabel>> roleTypes, Optional<TypeLabel> relationType) {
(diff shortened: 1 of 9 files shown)


grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/Fragments.java
```diff
@@ -174,12 +178,17 @@ public static Fragment resourceIndex(Var start, TypeLabel typeLabel, Object reso
...
     }
 
-    static void applyTypeLabelToTraversal(
-            GraphTraversal<Vertex, Edge> traversal, Schema.EdgeProperty property, Optional<TypeLabel> typeLabel, GraknGraph graph) {
-        typeLabel.ifPresent(label -> traversal.has(property.name(), graph.admin().convertToId(label).getValue()));
+    static void applyTypeLabelsToTraversal(
+            GraphTraversal<Vertex, Edge> traversal, Schema.EdgeProperty property, Optional<Set<TypeLabel>> typeLabels, GraknGraph graph) {
+        typeLabels.ifPresent(labels -> {
+            Set<Integer> typeIds = labels.stream().map(label -> graph.admin().convertToId(label).getValue()).collect(toSet());
+            traversal.has(property.name(), P.within(typeIds));
```
fppt: Why does this need to take in multiple type labels now? I ask because I am not sure of the efficiency of using `P.within` with an indexed lookup. It should be fast but . . . you know.

aelred: This is used for `ROLE_TYPE_ID` and `RELATION_TYPE_ID`. Those are edge properties, so I don't think we ever need the index. Correct me if I'm wrong!

aelred: Oh, and the reason it needs to take in multiple is because it accepts the entire sub-type heirarchy. For example, if you query `(spouse: $x)`, it will follow a shortcut edge labelled with either `spouse`, `husband` or `wife`.

fppt: So those do need to be indexed so we can traverse them efficiently. Specifically they are vertex centric indices.

PR #1635, opened 2017-05-22
TasksController priorities

alexandraorth:


grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
```diff
@@ -100,21 +95,21 @@ public TasksController(Service spark, TaskManager manager) {
     })
     private Json getTasks(Request request, Response response) {
         TaskStatus status = null;
-        String className = request.queryParams(TASK_CLASS_NAME_PARAMETER);
-        String creator = request.queryParams(TASK_CREATOR_PARAMETER);
+        String className = request.queryParams(REST.Request.TASK_CLASS_NAME_PARAMETER);
```
fppt: YESSSSS!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! I hate static imports of strings and enums.

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
```diff
@@ -179,6 +177,9 @@ private Json createTask(Request request, Response response) {
                     .map(interval -> recurring(time, interval))
                     .orElse(TaskSchedule.at(time));
 
+            // Get the priority of a task (default is low)
+            priority = Optional.ofNullable(priorityParam).map(TaskState.Priority::valueOf).orElse(DEFAULT_TASK_PRIORITY);
```
fppt: Nitpick: What happens if I send the priority "asdfkhjvd". I expect a nasty exception?

alexandraorth: I have a test for this! You get back an IllegalArgumentException and a server 400.

alexandraorth: Which might not be the desired behavior :P

fppt: Ooooo nice.

PR #1637, opened 2017-05-22
Get rid of graph computer config since it is not actually mutable

fppt:


fppt:
!rtg https://jenkins.grakn.ai/job/unit-test-pull-request-grakn-titan/ai.grakn$grakn-test/3687/testReport/junit/ai.grakn.test.engine.controller/CommitLogControllerTest/whenCommittingGraph_TaskManagerReceivesCountTask/

fppt:
!rtg

fppt:
!rtg ai.grakn.test.engine.controller.TasksControllerTest.afterSendingTaskWithRunAt_ItIsDelayedInStorage

grakn-graph/src/main/java/ai/grakn/graph/internal/GraknComputerImpl.java
```diff
@@ -54,9 +56,13 @@
...
 
-    public GraknComputerImpl(Graph graph, String graphComputerType) {
+    public GraknComputerImpl(Graph graph) {
         this.graph = graph;
-        this.graphComputerClass = getGraphComputerClass(graphComputerType);
+        if(graph instanceof TinkerGraph){
+            graphComputerClass = TinkerGraphComputer.class;
+        } else {
+            graphComputerClass = GraknSparkComputer.class;
+        }
```
aelred: At this point, would it be possible to move this into `AbstractGraknGraph` and `GraknTinkerGraph` directly?

fppt: no because sometimes you want to initialise a computer without initialising a grakn graph

aelred: Got it! I see we only do that in one case: `ResourceDeduplicationTask`, so it might still be worth looking into.

PR #1639, opened 2017-05-22
Experimentally replace some `Optional`s with `@Nullable`

aelred:
@fppt I'd be interested if you think this is an improvement or not...


aelred:
I'm liking this less and less

PR #1643, opened 2017-05-23
Fix client race condition in logger test

alexandraorth:


alexandraorth:
@bolerio @aelred Please review this, you've both experienced this bug.

alexandraorth:
I'm going to look for 5 travis passes on this one.

alexandraorth:
travis pass #1

alexandraorth:
please retest this please - `ai.grakn.test.engine.controller.CommitLogControllerTest.whenCommittingGraph_TaskManagerReceivesCountTask`

alexandraorth:
travis pass #2

alexandraorth:
travis pass #3

grakn-test/src/test/java/ai/grakn/client/BatchMutatorClientTest.java
```diff
@@ -71,17 +73,28 @@ public void setupSession(){
...
-    public void whenSingleQueryLoadedAndTaskCompletionFunctionThrowsError_ErrorIsLogged(){
+    public void whenSingleQueryLoadedAndTaskCompletionFunctionThrowsError_ErrorIsLogged() throws InterruptedException {
+        CountDownLatch callbackCompleted = new CountDownLatch(1);
+
         // Create a BatchMutatorClient with a callback that will fail
         BatchMutatorClient loader = loader();
-        loader.setTaskCompletionConsumer((json) -> assertTrue("Testing Log failure",false));
+        loader.setTaskCompletionConsumer((json) -> {
+            try {
+                fail("Testing failures appear in logs");
```
bolerio: Why is that a failure? What happens when a Assert.fail is called asynchronously? Does it actually fail the test?

grakn-test/src/test/java/ai/grakn/client/BatchMutatorClientTest.java
```diff
@@ -71,17 +73,28 @@ public void setupSession(){
...
 
         // Load some queries
         generate(this::query).limit(1).forEach(loader::add);
 
         // Wait for queries to finish
         loader.waitToFinish();
 
+        // Wait for callback function to execute - Callbacks are run asynchronously
+        callbackCompleted.await(10, TimeUnit.SECONDS);
+
```
bolerio: I think you should check if callbackCompleted.await returns 'false' and fail the test if hat happens. You are expecting it to complete in 10s, but if it doesn't that would reveal a bug somewhere.

grakn-test/src/test/java/ai/grakn/client/BatchMutatorClientTest.java
alexandraorth: It does not fail the test. Its the equivalent of throwing an `AssertionError`

grakn-test/src/test/java/ai/grakn/client/BatchMutatorClientTest.java
alexandraorth: Yep, good idea, will do.

alexandraorth:
travis pass #4

alexandraorth:
@aelred Im not sure we want to do that. I don't think I've well explained this test explained in the comments. I only want the callback function to fail so that we can test the failure is logged.

alexandraorth:
travis pass #5

aelred:
@alexandraorth yes, that's clearer! Maybe include a message like "deliberate failure" or something.

alexandraorth:
@aelred done.

PR #1645, opened 2017-05-23
Bug #15092: Fix cache leak between transactions

fppt:


fppt:
@aelred keen eyes Mr.Chapman, keen and supremely picky. This one has the potential to hurt us all.

grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptCache.java
```diff
@@ -39,8 +39,14 @@
  *
  */
 class ConceptCache<V> {
+    //If no cache can produce the data then the database is read
     private final Supplier<V> databaseReader;
-    private Optional<V> cachedValue = Optional.empty();
+
+    //Transaction bound. If this is not set it does not yet exist in the graph
+    private ThreadLocal<V> cachedValue = new ThreadLocal<>();
```
aelred: Not sure what this comment means. Isn't it possible for something to exist in the graph, but not be set in the cache?

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
```diff
@@ -123,10 +123,16 @@
...
+    /**
+     * Flushes the internal transaction caches so that persisted information can be cached and shared between
+     * concepts
+     */
+    public void flushTxCache(){
+        cachedIsImplicit.flush();
+        cachedIsAbstract.flush();
+        cachedSuperType.flush();
+        cachedDirectSubTypes.flush();
+        cachedShards.flush();
```
aelred: `cachedDirectPlays.flush()` needs to be here.

grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptCache.java
fppt: yes you right. The comment is misleading.

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
fppt: Good eyes.

PR #1648, opened 2017-05-23
GraknEngineConfig handles unavailable properties

alexandraorth:


alexandraorth:
please retest this please - weird jenkins thing

alexandraorth:
please retest this please - `Command execution failed.`

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineConfig.java
```diff
@@ -237,20 +237,24 @@ public Properties getProperties() {
     }
 
     public String getProperty(String property) {
-        return prop.getProperty(property);
+         if(prop.containsKey(property)) {
```
fppt: It bothers me so much that this is 3 lines and. . . . [*]

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineConfig.java
```diff
@@ -237,20 +237,24 @@ public Properties getProperties() {
...
     }
 
     public String getProperty(String property, String defaultValue) {
-        String res = prop.getProperty(property);
-        if(res != null) {
-            return res;
-        }
-
-        return defaultValue;
+        return prop.containsKey(property) ? prop.getProperty(property)
```
fppt: [*] . . . this is 2 lines

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineConfig.java
alexandraorth: This bothered me as well :( After the `:` side I tried to put the throw, but java doesn't let you do that.

PR #1651, opened 2017-05-23
Change spark memory-level

duckofyork:


fppt:
Merged into master here: https://github.com/graknlabs/grakn/pull/1663

PR #1652, opened 2017-05-23
HAL update

marco-scoppetta:
- fixed HAL bug adding check when 'select' is used in query
  - added Academic graph to tests
  - wrote first HAL tests, useful to test explanation as well


grakn-test/src/test/java/ai/grakn/test/graql/hal/HALBuilderTest.java
```diff
@@ -0,0 +1,72 @@
...
+import static org.junit.Assert.assertTrue;
+
+
+public class HALBuilderTest {
+
+    @ClassRule
+    public static final GraphContext academyGraph = GraphContext.preLoad(AcademyGraph.get());
+
+    @Test
+    public void testHALStructure() {
```
fppt: Sorry to nitpick. Could you try to follow the same convention as everyone else i.e. something like
```
@test
public void whenDoingThisSpecificThing_ThisHappens
```

It makes the tests more clear to the outsider. So for example `testHALStructure()` might be `whenReceivingHALResponseForMatchQuery_EnsureResponseContainsConceptDetails`

fppt:
Closing due to a duplicate PR going into stable: https://github.com/graknlabs/grakn/pull/1682

PR #1654, opened 2017-05-23
Update dashboard configuration properties

alexandraorth:


fppt:
@alexandraorth rebase please

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineConfig.java
```diff
@@ -103,12 +106,20 @@ private GraknEngineConfig() {
...
         LOG.info("Project directory in use: [" + getProjectPath() + "]");
         LOG.info("Configuration file in use: [" + configFilePath + "]");
         LOG.info("Number of threads set to [" + numOfThreads + "]");
     }
 
+    private void setGraknVersion(){
+        LOG.info("setting version");
+        LOG.info(GraknVersion.VERSION);
+        prop.setProperty(GRAKN_VERSION_KEY, GraknVersion.VERSION);
+        System.out.println("Version set");
```
marco-scoppetta: Is this System.out necessary?

alexandraorth: Not at all. Neither are those logs, left over from debugging. I will remove.

PR #1659, opened 2017-05-24
Clean redis when cleaning grakn

fppt:


fppt:
Merged into master here: https://github.com/graknlabs/grakn/pull/1663

PR #1666, opened 2017-05-24
Depend on titan-library without nop exclusion

alexandraorth:


fppt:
rebase please

alexandraorth:
@fppt rebased. @sheldonkhall would you be able to test using this PR as a fix for your analytics logging problem?

sheldonkhall:
Tested with a standalone jar that depends on titan-factory only and has no exclusions or logger dependencies. Analytics jobs execute successfully without problems. This is OK to go in from my perspective.

PR #1668, opened 2017-05-24
Clear system graph when graph is deleted

fppt:


grakn-graph/src/main/java/ai/grakn/factory/SystemKeyspace.java
```diff
@@ -106,6 +107,25 @@ public SystemKeyspace(InternalFactory<T> factory){
     }
 
     /**
+     * This is called when a graph is deleted via {@link GraknAdmin#delete()}.
+     * This removes the keyspace of the deleted graph from
```
aelred: from where??? D:

Also if this removes the keyspace, shouldn't it be `deleteKeyspace`, as opposed to `keyspaceDeleted`?

grakn-graph/src/test/java/ai/grakn/graph/property/GraknGraphPropertyTest.java
```diff
@@ -92,7 +92,7 @@ public void whenCallingMostMethodOnAClosedGraph_Throw(
         // TODO: Should `admin`, `close`, `implicitConceptsVisible`, `showImplicitConcepts`, `getKeyspace` and `graql` be here?
         assumeThat(method.getName(), not(isOneOf(
                 "isClosed", "admin", "close", "commit", "abort", "isReadOnly", "implicitConceptsVisible", "showImplicitConcepts",
-                "getKeyspace", "graql", "getId"
+                "getKeyspace", "graql", "getId", "clear"
```
aelred: Since "clear" is not on the API, you don't need it here.

grakn-graph/src/main/java/ai/grakn/factory/SystemKeyspace.java
```diff
@@ -108,12 +108,12 @@ public SystemKeyspace(InternalFactory<T> factory){
 
     /**
      * This is called when a graph is deleted via {@link GraknAdmin#delete()}.
-     * This removes the keyspace of the deleted graph from
+     * This removes the keyspace of the deleted graph from thr system graph
```
aelred: ...good enough!

fppt: typing hard.

PR #1671, opened 2017-05-25
Reuse role type maps

kasper-piskorski:


grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/Atom.java
```diff
@@ -65,7 +61,8 @@
     protected Atom(Atom a) {
         super(a);
         this.type = a.type;
-        this.typeId = a.getTypeId() != null? ConceptId.of(a.getTypeId().getValue()) : null;
+        this.typeId = a.typeId != null? ConceptId.of(a.getTypeId().getValue()) : null;
```
fppt: We might be confusing our terminology between components. In this case is `a.getTypeId()` returning the concept ID or the type ID because those are two different things. Also maybe it's worth making `a.getTypeId` return the actual ID object it would clear up these confusions for example.

kasper-piskorski: ok makes sense

kasper-piskorski:
please retest this please - ai.grakn.test.engine.controller.CommitLogControllerTest.whenCommittingGraph_TaskManagerReceivesCountTask

PR #1675, opened 2017-05-25
Some overdue cleanups in graph API

fppt:


grakn-graph/src/main/java/ai/grakn/graph/internal/EdgeImpl.java
```diff
@@ -57,7 +57,7 @@ public int hashCode() {
 
     @Override
     public boolean equals(Object object) {
-        return object instanceof EdgeImpl && ((EdgeImpl) object).edgeEquals(edge);
+        return object instanceof EdgeImpl && ((EdgeImpl) object).edge.equals(edge);
```
aelred: I can imagine you reading this and going "wtf past @fppt?"

fppt: Thankfully this was not my piece of wizardry :)

PR #1677, opened 2017-05-25
Simplify system keyspace + stop tasks from running on graphs which may have been deleted

fppt:


grakn-graph/src/main/java/ai/grakn/factory/SystemKeyspace.java
```diff
@@ -77,21 +78,61 @@
...
+     * Initialises the system keyspace for a specific running instance of engine.
+     * This initializer is used when the first graph created is the system graph.
+     * @param internalFactory the factory to use when initialising the system graph.
+     */
+    static void initialise(InternalFactory internalFactory){
+        if(factory == null) {
+            factory = internalFactory;
+            loadSystemOntology();
+        }
+    }
```
aelred: I haven't totally wrapped my head around this - but why does this second `initialise` method not use `isFactorySet`?

fppt: The answer is simple my dear Felix, it is because I am an idiot and missed that one :)

aelred: In that case - you can probably make one `initialise` method call the other 👍

aelred: Wow, you're too fast for me...

fppt:
!rtg - https://jenkins.grakn.ai/job/unit-test-pull-request-grakn-titan/3811/console

fppt:
@aelred re-review please

fppt:
!rtg - https://jenkins.grakn.ai/job/unit-test-pull-request-grakn-titan/3832/

PR #1679, opened 2017-05-25
[Stable fix] Set logging type inference to trace

kasper-piskorski:


fppt:
@kasper-piskorski please create a merge from stable into master and reference it here.

fppt:
https://github.com/graknlabs/grakn/pull/1684

PR #1680, opened 2017-05-26
Remove some uses of singletons in codebase, where feasible

aelred:


aelred:
!rtg some npm issue?

grakn-dist/src/bin/grakn-engine.sh
```diff
@@ -38,6 +38,9 @@ for jar in "${GRAKN_HOME}"/lib/*.jar; do
     fi
 done
 
+# Add path containing logback.xml
+CLASSPATH="$CLASSPATH":"${GRAKN_HOME}"/conf/main/
```
alexandraorth: This probably should not be hard coded. What if someone specifies their properties and logback conf somewhere else

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -57,46 +63,48 @@
...
+    private final EngineGraknGraphFactory factory;
+
+    private GraknEngineServer(GraknEngineConfig prop) {
+        this.prop = prop;
+
+        factory = EngineGraknGraphFactory.create(prop.getProperties());
+        taskManager = startTaskManager();
+        usersHandler = UsersHandler.create(prop.getProperty(ADMIN_PASSWORD_PROPERTY), factory);
+
+        Optional<String> secret = prop.tryProperty(JWT_SECRET_PROPERTY);
```
alexandraorth: We probably should not be doing these static imports.

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -57,46 +63,48 @@
...
+    private final UsersHandler usersHandler;
+    private final @Nullable JWTHandler jwtHandler;  // TODO: Make sure controllers handle the null case
+    private final EngineGraknGraphFactory factory;
+
+    private GraknEngineServer(GraknEngineConfig prop) {
+        this.prop = prop;
+
+        factory = EngineGraknGraphFactory.create(prop.getProperties());
+        taskManager = startTaskManager();
+        usersHandler = UsersHandler.create(prop.getProperty(ADMIN_PASSWORD_PROPERTY), factory);
```
alexandraorth: I would move this down to the startHTTP() method because its only used there.

alexandraorth: Ditto for JWTHandler

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -120,40 +130,47 @@ private TaskManager startTaskManager(String taskManagerClassName) {
...
-        EngineGraknGraphFactory factory = EngineGraknGraphFactory.getInstance();
         new GraqlController(factory, spark);
         new ConceptController(factory, spark);
         new DashboardController(factory, spark);
-        new SystemController(spark);
-        new AuthController(spark);
-        new UserController(spark);
-        new CommitLogController(spark, taskManager);
+        new SystemController(spark, prop.getProperties());  // TODO: Only pass the necessary properties
+        new AuthController(spark, passwordProtected, jwtHandler, usersHandler);
```
alexandraorth: Should we only start the `AuthController` if `isPasswordProtected` is true? This would mean that we can instantiate the `JWTHandler` only when it is needed for authentication.

grakn-engine/src/main/java/ai/grakn/engine/controller/AuthController.java
```diff
@@ -92,6 +96,6 @@ private String newSession(Request req, Response res) {
     @ApiOperation(
             value = "Returns true if Engine endpoints are password protected. False otherwise.")
     private String isPasswordProtected(Request req, Response res) {
```
alexandraorth: This should return a boolean.

grakn-engine/src/main/java/ai/grakn/engine/controller/SystemController.java
```diff
@@ -69,8 +69,12 @@
  */
 public class SystemController {
     private final Logger LOG = LoggerFactory.getLogger(SystemController.class);
+    private final Properties graphProperties;
+    private final EngineGraknGraphFactory factory;
 
-    public SystemController(Service spark) {
+    public SystemController(Service spark, Properties graphProperties) {
+        this.graphProperties = graphProperties;
+        this.factory = EngineGraknGraphFactory.create(graphProperties);
```
alexandraorth: Can we create a method on the `EngineGraknGraphFactory` that will return the properties? That way we will no have to instantiate another one here.

grakn-engine/src/main/java/ai/grakn/engine/session/RemoteSession.java
```diff
@@ -46,44 +43,46 @@
...
-    @SuppressWarnings("unused")
-    public RemoteSession() {
+    private RemoteSession(@Nullable UsersHandler usersHandler) {
+        this.usersHandler = usersHandler;
     }
 
-    @OnWebSocketConnect
-    public void onConnect(Session session) {
+    public static RemoteSession create() {
+        return new RemoteSession(null);
```
alexandraorth: :(

grakn-engine/src/main/java/ai/grakn/engine/user/SystemKeyspaceUsers.java
```diff
@@ -49,6 +49,12 @@
  */
 public class SystemKeyspaceUsers extends UsersHandler {
     private final Logger LOG = LoggerFactory.getLogger(SystemKeyspaceUsers.class);
+    private final EngineGraknGraphFactory factory;
+
+    SystemKeyspaceUsers(String adminPassword, EngineGraknGraphFactory factory) {
```
alexandraorth: I don't like that sometimes we pass in this `EngineGraknGraphFactory` and other times we don't.

aelred:
!rtg - looks like there was a timeout? I'm assuming it's a random failure.

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
aelred: Yes, probably. I'm going to make a TP bug for this and leave it for a different PR - since this bug already existed.

grakn-dist/src/bin/grakn-engine.sh
aelred: I agree, but for the future! Ideally all the hard-coded paths in the shell scripts should be defined just once somewhere.

aelred:
!rtg `TitanPreviousPropertyStepTest.whenFilteringAPropertyToBeEqualToAPreviousProperty_UseTitanGraphStep`

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -120,40 +118,53 @@ private TaskManager startTaskManager(String taskManagerClassName) {
...
         // Start all the controllers
-        EngineGraknGraphFactory factory = EngineGraknGraphFactory.getInstance();
         new GraqlController(factory, spark);
         new ConceptController(factory, spark);
         new DashboardController(factory, spark);
-        new SystemController(spark);
-        new AuthController(spark);
-        new UserController(spark);
-        new CommitLogController(spark, taskManager);
+        new SystemController(factory, spark);
```
fppt: This actually makes a lot more sense now. It reduces the cohesion of the controller which will help with making the service orientated architecture we have been talking about. In other words. .  . nice.

grakn-engine/src/main/java/ai/grakn/engine/controller/CommitLogController.java
```diff
@@ -49,9 +47,11 @@
  */
 //TODO Implement delete
 public class CommitLogController {
+    private final String defaultKeyspace;
```
fppt: I suspect that this can actually be removed. Future PR though as not relevant to the objectives of this PR.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/singlequeue/SingleQueueTaskManager.java
```diff
@@ -92,23 +93,28 @@
...
-        this.zookeeper = new ZookeeperConnection();
+    public SingleQueueTaskManager(EngineID engineId, GraknEngineConfig config) {
+        this.config = config;
+        this.zookeeper = new ZookeeperConnection(config);
         this.storage = new TaskStateZookeeperStore(zookeeper);
         this.offsetStorage = new ExternalOffsetStorage(zookeeper);
 
         //TODO check that the number of partitions is at least the capacity
-        this.producer = kafkaProducer();
+        //TODO only pass necessary Kafka properties
```
fppt: This TODO should also be whacked on for the engine graph factory properties

PR #1682, opened 2017-05-26
HAL fix when Select is used in queries

marco-scoppetta:
- Fixed HAL builder to support "select" in user queries
- Updated error message in Dashboard when aggregate queries are attempted from Graph page.
- Regression test and new academic graph used for tests


fppt:
@marco-scoppetta your stable fix has been merged. Please create merge into master and reference it here.

marco-scoppetta:
Merging into master with this PR: [#1684](https://github.com/graknlabs/grakn/pull/1684)

PR #1683, opened 2017-05-26
Fix some incorrect tests

aelred:


grakn-test/src/test/java/ai/grakn/test/engine/controller/UserControllerTest.java
```diff
@@ -47,23 +46,8 @@ public void testAddNewUser(){
...
-        //Add User
-        Json user = Json.object(UsersHandler.USER_NAME, userName, UsersHandler.USER_PASSWORD, password);
-        users.addUser(user);
-
-        //Get Him Back
-        Response dataResponse =
-                get("/user/one?" + UsersHandler.USER_NAME + "=" + userName).then().
-                        statusCode(200).extract().response().andReturn();
-
-        assertTrue(dataResponse.getBody().asString().contains(userName));
```
aelred: To clarify what happened here: I removed the dependency on `UsersHandler`, using the REST API exclusively to add/check for users.

This meant the tests to `GET` and `POST` users became exactly the same! So I deleted one.

PR #1691, opened 2017-05-30
Bug #15229: Instances could be added to abstract types

fppt:


fppt:
!rtg

grakn-graph/src/main/java/ai/grakn/graph/internal/ValidateGlobalRules.java
```diff
@@ -189,7 +189,7 @@ private ValidateGlobalRules() {
                 conceptType.<TypeImpl>getIncomingNeighbours(Schema.EdgeLabel.SHARD).anyMatch(thing ->
                 thing.getIncomingNeighbours(Schema.EdgeLabel.ISA).findAny().isPresent())){
 
-            return Optional.of(VALIDATION_IS_ABSTRACT.getMessage(conceptType.getLabel()));
+            return Optional.of(IS_ABSTRACT.getMessage(conceptType.getLabel()));
         }
         return Optional.empty();
     }
```
aelred: Can we get rid of this validation rule?

grakn-graph/src/test/java/ai/grakn/generator/AbstractTypeGenerator.java
```diff
@@ -92,11 +98,20 @@ private final boolean includeNonMeta() {
         return !meta.orElse(false);
     }
 
+    private final boolean includeAbstract(){
+        return isAbstract.orElse(false);
+    }
```
aelred: This should be `.orElse(true)`. Essentially, if you don't specifically ask to include/exclude abstract types, then they should be included.

I'm wondering if there's a way to remove the `Optional` from `isAbstract` and `meta`, but I'm not sure...

grakn-graph/src/test/java/ai/grakn/graph/internal/ValidatorTest.java
```diff
@@ -129,20 +130,12 @@ public void whenCommittingNonAbstractRelationTypeNotLinkedToAnyRoleType_Throw(){
...
-
-        x4.superType(x3);
-
-        expectedException.expect(GraknValidationException.class);
-        expectedException.expectMessage(containsString(ErrorMessage.VALIDATION_IS_ABSTRACT.getMessage(x1.getLabel())));
+        expectedException.expect(GraphRuntimeException.class);
+        expectedException.expectMessage(containsString(ErrorMessage.IS_ABSTRACT.getMessage(x1.getLabel())));
 
-        graknGraph.commit();
+        x1.setAbstract(true);
```
aelred: I'd delete this test, it seems the same as the above. The test name is also now incorrect.

grakn-test/src/test/java/ai/grakn/test/graql/query/InsertQueryTest.java
```diff
@@ -417,7 +416,6 @@ public void testInsertReferenceByName() {
         //noinspection OptionalGetWithoutIsPresent
         EntityType newType = typeQuery.get("n").findFirst().get().asEntityType();
 
-        assertTrue(newType.asEntityType().isAbstract());
```
aelred: Was this failing for some reason?

grakn-graph/src/main/java/ai/grakn/graph/internal/ValidateGlobalRules.java
fppt: yes we can.

grakn-test/src/test/java/ai/grakn/test/graql/query/InsertQueryTest.java
fppt: Yes the test was attemping to add instances to an abstract type. So I stopped the type from being abstract which meant this line had to die.

grakn-graph/src/test/java/ai/grakn/graph/internal/ValidatorTest.java
fppt: Which test above?

grakn-test/src/test/java/ai/grakn/test/graql/query/InsertQueryTest.java
aelred: Got it! Missed the `isa`.

PR #1692, opened 2017-05-30
HAL fix for NPE when using select in query with no inference.

marco-scoppetta:


marco-scoppetta:
!rtg

fppt:
@marco-scoppetta please create a merge of stable into master and reference it here.

fppt:
https://github.com/graknlabs/grakn/pull/1695

PR #1693, opened 2017-05-30
Added endpoint for bulk creation of tasks

pluraliseseverythings:


pluraliseseverythings:
!rtg - Not sure what's going on

pluraliseseverythings:
!rtg - Flaky test methinks

pom.xml
```diff
@@ -330,6 +330,7 @@
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <version>2.9.1</version>
                 <configuration>
+                    <javadocExecutable>${java.home}/../bin/javadoc</javadocExecutable>
```
alexandraorth: What is this for?

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
```diff
@@ -65,32 +79,41 @@
 @Api(value = "/tasks", description = "Endpoints used to query and control queued background tasks.", produces = "application/json")
 public class TasksController {
 
+    private static final Logger LOG = LoggerFactory.getLogger(GraknVertexProgram.class);
```
alexandraorth: This logger should be for the current class. i.e. `private static final Logger LOG = LoggerFactory.getLogger(TasksController.class);`

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
```diff
@@ -154,9 +177,9 @@ private Json stopTask(Request request, Response response) {
...
             @ApiImplicitParam(name = "runAt", value = "Time to run at as milliseconds since the UNIX epoch", required = true, dataType = "long", paramType = "query"),
-            @ApiImplicitParam(name = "interval",value = "If set the task will be marked as recurring and the value will be the time in milliseconds between repeated executions of this task. Value should be as Long.",
+            @ApiImplicitParam(name = "interval", value = "If set the task will be marked as recurring and the value will be the time in milliseconds between repeated executions of this task. Value should be as Long.",
                     dataType = "long", paramType = "query"),
-            @ApiImplicitParam(name = "configuration", value = "JSON Object that will be given to the task as configuration.", dataType = "String", paramType = "body")
+            @ApiImplicitParam(name = REST.Request.CONFIGURATION_PARAM, value = "JSON Object that will be given to the task as configuration.", dataType = "String", paramType = "body")
```
alexandraorth: This is pedantic- if we use the variable for one of these comments, we should for all of them. I like the idea! Keeps us consistent.

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
```diff
@@ -65,32 +79,41 @@
...
 
-        spark.get(TASKS,       this::getTasks);
-        spark.get(GET,         this::getTask);
-        spark.put(STOP,        this::stopTask);
-        spark.post(TASKS,      this::createTask);
+        spark.get(TASKS, this::getTasks);
+        spark.get(GET, this::getTask);
+        spark.put(STOP, this::stopTask);
+        spark.post(TASKS, this::createTask);
+        spark.post(TASKS_BULK, this::createTaskBulk);
```
alexandraorth: Do you guys think we need to have both the bulk endpoint and the single task endpoint? @marco-scoppetta We could merge the two...

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
```diff
@@ -165,54 +188,190 @@ private Json createTask(Request request, Response response) {
...
+        response.status(200);
+
+        return Json.object("id", taskState.getId().getValue());
+    }
+
+    @POST
+    @Path("/bulk")
+    @ApiOperation(value = "Schedule a set of tasks.")
+    @ApiImplicitParams({
+            @ApiImplicitParam(name = REST.Request.TASKS_PARAM, value = "JSON Array containing an ordered list of tasks (see single task endpoint). The response references the order the task appears in the list as an index.", required = true, dataType = "List", paramType = "body"),
```
alexandraorth: Let me know if I have this wrong: While we can send multiple tasks to this endpoint, they must all have the same configuration–– we would not be able to send 10 loader tasks where each one is loading a different set of queries?

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
```diff
@@ -165,54 +188,190 @@ private Json createTask(Request request, Response response) {
...
+                response.status(HttpStatus.SC_OK);
+            } else if (responseJson.asJsonList().size() > 0) {
+                response.status(HttpStatus.SC_ACCEPTED);
+            } else {
+                response.status(HttpStatus.SC_INTERNAL_SERVER_ERROR);
+            }
+            return responseJson;
+        } catch (TimeoutException | InterruptedException e) {
+            LOG.error("Task interrupted", e);
+            response.status(HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
alexandraorth: Im not sure if this is standard, but it might be nice if we included a bit of the exception when we send the response back to the client so they know what went wrong.

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
fppt: In my opinion we should not have both. Merge them.

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
fppt: That is only usual if the system is in debug mode. Returning full stack traces **all* the time is not usual.

grakn-core/src/main/java/ai/grakn/util/REST.java
```diff
@@ -46,6 +46,7 @@
          */
         public static class Tasks {
             public static final String TASKS = "/tasks";
+            public static final String TASKS_BULK = "/tasks/bulk";
```
fppt: I don't think there should be a separate end point for bulk loading. Unless there is a clear advantage I am missing. We only have one pipeline for processing tasks so there should only be one endpoint for uploading those tasks.

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
fppt: +1

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
fppt: +1

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
pluraliseseverythings: Shouldn't we deprecate it for now? Or the alternative is to support both requests types in the same endpoint/REST verb but that would be messy

fppt: At this stage I think  we can still afford rapid changes. We don't have any external reliance on the task controller so there should be no harm in killing it.

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
pluraliseseverythings: That's right, I think it's because when I started working on this I thought the relevant info to instantiate the task was in the task parameters. What's the distinction between config and parameters? If we are refactoring the whole endpoint it might make sense to rethink that.

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
pluraliseseverythings: Now I regret doing this 😭

pom.xml
pluraliseseverythings: Not sure whether to leave this there or not. Currently if you just clone the code and set the JDK path the build fails. This fixes it although it doesn't look pretty, I'd prefer adding a note on how to set the javadoc path in the docs.

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
alexandraorth: Well, we only really have the `TaskState` and `TaskConfiguration` separate because we do not want to store the configuration in our external storage - it is too large. 

I think its pretty important to be able to send tasks with different configurations to this endpoint. Take our primary use case (which i now realise we should have talked about before.): we want to be able to send a `PostProcessingTask` and `UpdatingInstanceCountTask` in the same request, and each of these need different configurations (is this right, @fppt?).

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
pluraliseseverythings: Ok!

grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
```diff
@@ -259,55 +299,14 @@ private void unblock(CompletableFuture<Json> status){
     }
 
     /**
-     * Set POST request to host containing information
-     * to execute a Loading Tasks with the insert queries as the body of the request
-     *
-     * @return A Completable future that terminates when the task is finished
-     */
-    private String executePost(String body) throws HttpRetryException {
```
pluraliseseverythings: I deleted this and using the TaskClient instead to hit the tasks endpoint

grakn-client/src/main/java/ai/grakn/client/TaskClient.java
```diff
@@ -85,31 +89,49 @@ public static TaskClient of(String host, int port) {
...
+        return sendTask(taskClass.getName(), creator, runAt, interval, configuration, -1);
+    }
+
+    TaskId sendTask(String taskClass, String creator, Instant runAt, Duration interval, Json configuration, long limit){
         try {
             URIBuilder uri = new URIBuilder(TASKS)
                     .setScheme(DEFAULT_SCHEME_NAME)
                     .setHost(host)
-                    .setPort(port)
-                    .setParameter(TASK_CLASS_NAME_PARAMETER, taskClass.getName())
```
pluraliseseverythings: These are now part of the request body, not parameters

grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
fppt: Nice

grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
```diff
@@ -338,7 +337,7 @@ private Json getStatus(String id) throws HttpRetryException {
      * @param id ID of the task to wait on completion
      * @return Completable future that will await completion of the given task
      */
-    private CompletableFuture<Json> makeTaskCompletionFuture(String id){
+    private CompletableFuture<Json> makeTaskCompletionFuture(TaskId id){
```
alexandraorth: As we are using the `Retryer` library above, we should also be using it here.

grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
```diff
@@ -219,38 +239,58 @@ void sendQueriesToLoader(Collection<Query> queries){
...
+                .set(BATCH_NUMBER, batchNumber)
+                .set(TASK_LOADER_MUTATIONS,
+                        queries.stream().map(Query::toString).collect(toList()));
+
+        Callable<TaskId> callable = () -> taskClient
+                .sendTask("ai.grakn.engine.loader.MutatorTask",
+                        BatchMutatorClient.class.getName(),
+                        Instant.ofEpochMilli(new Date().getTime()), null, configuration, 10000);
+
+        Retryer<TaskId> retryer = RetryerBuilder.<TaskId>newBuilder()
```
alexandraorth: Previously we would only re-attempt to send the task if the `retry` boolean was true, but now it looks like we are always doing it?

grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
```diff
@@ -259,55 +299,14 @@ private void unblock(CompletableFuture<Json> status){
...
      * @return Json object containing status of the task
      */
-    private Json getStatus(String id) throws HttpRetryException {
+    private Json getStatus(TaskId id) throws HttpRetryException {
         HttpURLConnection connection = null;
         try {
-            URL url = new URL(format(GET, uri, id));
+            URL url = new URL(format(GET, uri, id.getValue()));
 
             connection = (HttpURLConnection) url.openConnection();
```
alexandraorth: Here we can also use the `TasksClient` now, as you are doing above.

grakn-test/src/test/java/ai/grakn/client/BatchMutatorClientTest.java
```diff
@@ -73,6 +75,7 @@ public void setupSession(){
     }
 
     @Test
+    @Ignore("Testing log output, it's pretty flimsy. It's also prone to race conditions (the whole test task halts forever)")
```
alexandraorth: Yes, this is 100% true. I think I've identified the race condition that causes it to randomly fail, but could not figure out how to prevent it :P

alexandraorth: Also, I love the comment in the ignore! Did not realise you could do that.

grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
pluraliseseverythings: Good point here, I'll reintroduce the flag

pluraliseseverythings:
!rtg - Titan test didn't run

pluraliseseverythings:
!rtg - Titan test didn't run - trying again before I spend time debugging it

PR #1695, opened 2017-05-31
HAL fix for select with no reasoner. Merging with stable.

marco-scoppetta:


marco-scoppetta:
Skipping the "update to latest npm version" step for now. We will need to wait for npm 5.0.1. Re:
[#16809](https://github.com/npm/npm/issues/16809)

PR #1701, opened 2017-06-01
Cleaning up internal exceptions

fppt:


fppt:
!rtg - ai.grakn.test.engine.controller.CommitLogControllerTest.whenCommittingGraph_TaskManagerReceivesCountTask

fppt:
@alexandraorth @aelred it looks big but most of it is javadocs

grakn-client/src/main/java/ai/grakn/client/TaskClient.java
```diff
@@ -111,7 +111,7 @@ public TaskId sendTask(Class<?> taskClass, String creator, Instant runAt, Durati
 
             return TaskId.of(jsonResponse.at("id").asString());
         } catch (IOException e){
-            throw new EngineUnavailableException(e);
+            throw GraknBackendException.engineUnavailable(host, port, e);
```
aelred: This PR might be a good opportunity to add the `throws` declarations to the functions when appropriate.

grakn-core/src/main/java/ai/grakn/GraknGraph.java
```diff
@@ -64,8 +63,8 @@
      * @param label A unique label for the {@link EntityType}
      * @return A new or existing {@link EntityType} with the provided label
      *
-     * @throws GraphRuntimeException if the graph is closed
-     * @throws ConceptNotUniqueException if the {@param label} is already in use by an existing non-{@link EntityType}.
+     * @throws GraphOperationException if the graph is closed
+     * @throws PropertyNotUniqueException if the {@param label} is already in use by an existing non-{@link EntityType}.
```
aelred: Maybe this is for a future PR, but this should say `throws` on the method signature as well.
Ditto for a lot of methods in this PR.

grakn-core/src/main/java/ai/grakn/exception/GraknBackendException.java
```diff
@@ -18,19 +18,170 @@
...
+ * </p>
  *
- * @author Filipe Teixeira
+ * @author fppt
  */
-public class GraknBackendException extends GraphRuntimeException {
-    public GraknBackendException(Exception e) {
-        super(ErrorMessage.BACKEND_EXCEPTION.getMessage(), e);
+public class GraknBackendException extends GraknException {
+    private Optional<Integer> status = Optional.empty();
```
aelred: I'm not convinced this is the right way to handle this. In this case we could probably use a dedicated sub-class.

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
```diff
@@ -137,9 +138,8 @@ static String generateNewHash(RelationType relationType, Map<RoleType, Set<Insta
      */
     @Override
     public Relation addRolePlayer(RoleType roleType, Instance instance) {
-        if(roleType == null){
-            throw new IllegalArgumentException(ErrorMessage.ROLE_IS_NULL.getMessage(instance));
-        }
+        if(roleType == null) throw new IllegalArgumentException(ErrorMessage.ROLE_IS_NULL.getMessage(instance));
```
aelred: There are some useful helper methods `Objects.requireNonNull` and Guava's `Preconditions.checkNotNull`. 

I'm not totally convinced you need a custom message for this either... after all, we would end up with a lot of messages if we did this for every API method!

grakn-core/src/main/java/ai/grakn/exception/GraknBackendException.java
fppt: I am not happy with this approach either and was trying to avoid declaring a sub-class but I suppose there is a good enough cause for it in this case.

grakn-client/src/main/java/ai/grakn/client/TaskClient.java
alexandraorth: I agree. Can we enable a check style that enforces this?

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
fppt: Agh. Missed this one. Silly.

grakn-core/src/main/java/ai/grakn/concept/EntityType.java
```diff
@@ -53,8 +53,8 @@
      * @param type The supertype of this EntityType
      * @return The EntityType itself
      *
-     * @throws ConceptException if this is a meta type
-     * @throws ConceptException if the given supertype is already an indirect subtype of this type
+     * @throws GraphOperationException if this is a meta type
```
alexandraorth: Do we need multiple throw for the same exception here?

grakn-core/src/main/java/ai/grakn/exception/GraknBackendException.java
```diff
@@ -18,19 +18,169 @@
...
+     */
+    public static GraknBackendException invalidQueryExplaination(String query){
+        return new GraknBackendException(ErrorMessage.EXPLAIN_ONLY_MATCH.getMessage(query), 405);
+    }
+
+    /**
+     * Thrown when an incorrect query is used with a REST endpoint
+     */
+    public static GraknBackendException authenticationFailure(){
+        return new GraknBackendException("Authentication parameters are incorrect or invalid", 401);
```
alexandraorth: We should put this in the `ErrorMessage` class

grakn-core/src/main/java/ai/grakn/exception/GraknBackendException.java
```diff
@@ -18,19 +18,169 @@
...
+     */
+    public static GraknBackendException stateStorage(String error){
+        return new GraknBackendException(STATE_STORAGE_ERROR.getMessage() + error);
+    }
+
+    /**
+     * Thrown when a task id is missing from the task state storage
+     */
+    public static GraknBackendException stateStorageMissingId(TaskId id){
+        return new GraknBackendException("Could not retrieve id " + id);
```
alexandraorth: We should put this in the `ErrorMessage` class. Along with the next 4.

grakn-core/src/main/java/ai/grakn/exception/GraphOperationException.java
```diff
@@ -0,0 +1,204 @@
...
+        } else {
+            return new GraphOperationException(reason);
+        }
+    }
+
+    /**
+     * Thrown when the graph can not be closed due to an unknown reason.
+     */
+    public static GraphOperationException closingGraphFailed(GraknGraph graph, Exception e){
+        return new GraphOperationException("Unable to close graph [" + graph.getKeyspace() + "]", e);
```
alexandraorth: We should put this in the `ErrorMessage` class

grakn-core/src/main/java/ai/grakn/exception/GraqlQueryException.java
```diff
@@ -19,12 +19,20 @@
...
+ *     For example limiting the results of a query -1
+ * </p>
+ *
+ * @author fppt
  */
-public class GraknValidationException extends GraphRuntimeException{
-    public GraknValidationException(String error) {
+//TODO: Felix to integrate with Graql
+public class GraqlQueryException extends GraknException{
+    public GraqlQueryException(String error) {
```
alexandraorth: Should this Exception take in the query which had the exception so that it can be used in the message?

grakn-core/src/main/java/ai/grakn/exception/GraqlSyntaxException.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public static GraqlSyntaxException parsingError(String error){
+        return new GraqlSyntaxException(error);
+    }
+
+    /**
+     * Thrown when there is a syntactic error in a Graql template
+     */
+    public static GraqlSyntaxException parsingTemplateError(String statmentType, String invalidText, Map<String, Object> data){
+        return new GraqlSyntaxException("Invalid " + statmentType + " statement: " + invalidText + " for data " + data);
```
alexandraorth: We should put this in the `ErrorMessage` class

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
```diff
@@ -71,7 +69,7 @@
 
     public TasksController(Service spark, TaskManager manager) {
         if (manager==null) {
-            throw new GraknEngineServerException(500,"Task manager has not been instantiated.");
+            throw GraknBackendException.internalError("Task manager has not been instantiated.");
```
alexandraorth: We should put this in the `ErrorMessage` class?

grakn-core/src/main/java/ai/grakn/concept/EntityType.java
fppt: Its a good way to list the cuases of the exception I think

grakn-core/src/main/java/ai/grakn/exception/GraknBackendException.java
fppt: Ugh . . . fine . . . you right but ugh . . . :)

grakn-core/src/main/java/ai/grakn/exception/GraqlQueryException.java
fppt: I will let Felix make that decision but given the context that maybe a good idea.

fppt:
@alexandraorth @aelred re-review please

grakn-client/src/main/java/ai/grakn/client/TaskClient.java
aelred: Apparently yes!
http://checkstyle.sourceforge.net/config_javadoc.html#JavadocMethod

grakn-core/src/main/java/ai/grakn/exception/GraqlQueryException.java
aelred: Sounds like a good idea. I'll do it when I update the graql errors properly.

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
```diff
@@ -210,23 +209,24 @@ private Json createTask(Request request, Response response) {
...
 
     /**
      * Error accessing or retrieving a task from storage. This throws a 404 Task Not Found to the user.
-     * @param exception {@link EngineStorageException} thrown by the server
+     * @param exception {@link GraknServerException} thrown by the server
      * @param response The response object providing functionality for modifying the response
      */
     private void handleNotFoundInStorage(Exception exception, Response response){
-        response.status(404);
+        if(exception instanceof GraknServerException){
```
alexandraorth: You don't need to handle `GraknServerException` here. You have registered the method using 
```        spark.exception(GraknBackendException.class, (e, req, res) -> handleNotFoundInStorage(e, res));
``` 
this means that only exceptions of the type `GraknBackendException` will be handled by this method. Any `GraknServerException` that is thrown is handled by `GraknEngineServer.handleInternalError`

fppt:
!rtg

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
fppt: Turns out this check is needed. I have to put it back because of the mix of exceptions being thrown + the new exception tree.

fppt: Basically the problem is that `GraknServerException extends GraknBackendException `

fppt:
!rtg

PR #1707, opened 2017-06-01
Download redis using maven plugin instead of wget.

burukuru:
I forked your repo to create a release with a unique name (and keep files executable).
Otherwise maven caches the files by name  which would clash with shunit2.

Tested on Linux, packages/starts/connects/stops fine.


burukuru:
!rtg

fppt:
!rtg

PR #1710, opened 2017-06-02
Single resolution plan for a single conjunctive query and simplified construction

kasper-piskorski:


kasper-piskorski:
please retest this please - grakn-dist failure

PR #1712, opened 2017-06-02
Reduce memory usage when parsing very large lists of queries

aelred:


aelred:
!rtg - failure on ai.grakn.test.engine.controller.CommitLogControllerTest.whenCommittingGraph_TaskManagerReceivesCountTask

grakn-engine/src/main/java/ai/grakn/engine/session/GraqlSession.java
```diff
@@ -218,7 +218,7 @@ void receiveQuery(Json json) {
                 String queryString = queryStringBuilder.toString();
                 queryStringBuilder = new StringBuilder();
 
-                queries = graph.graql().infer(infer).materialise(materialise).parseList(queryString);
+                queries = graph.graql().infer(infer).materialise(materialise).parseList(queryString).collect(toList());
```
fppt: Since you serialise the stream into a list I am not entirely sure you saving much memory (or any) in this one case.

grakn-graql/src/main/java/ai/grakn/graql/internal/parser/QueryParser.java
```diff
@@ -199,6 +246,34 @@ public Pattern parsePattern(String patternString){
...
+     * @return a new {@link TokenSource} containing the tokens comprising the query
+     */
+    private TokenSource consumeOneQuery(TokenStream tokenStream) {
+        List<Token> tokens = new ArrayList<>();
+
+        boolean startedQuery = false;
+
+        while (true) {
+            Token token = tokenStream.LT(1);
+            boolean isNewQuery = NEW_QUERY_TOKENS.contains(token.getType());
```
fppt: Could you clarify the logic here. This many booleans makes it a bit hard to read.

grakn-engine/src/main/java/ai/grakn/engine/session/GraqlSession.java
aelred: Good point! I check the list later to see if the queries are read-only, so I'll have to do something a bit nasty, but it's doable.

aelred: Actually I take it back - I'm not sure if this is possible: we do other stuff with the list of queries, such as killing them if they're compute queries, etc.

aelred:
@fppt re-review time!

PR #1713, opened 2017-06-02
Fix pom erros

fppt:


fppt:
https://github.com/graknlabs/grakn/pull/1716

PR #1722, opened 2017-06-06
Get rid of casting vertices. Make them an implementation detail

fppt:
- Define `Element` class to be the root of all graph constructs. This lets us interface between things which are not `Concepts`
- No longer use Casting vertices when validating
- No longer use Casting vertices in any traversal
- Don't actually make any casting vertices
- Make the `Casting` object and java wrap the shortcut edge and use that for everything.


fppt:
This PR is blocked until https://github.com/graknlabs/grakn/pull/1724 goes in. However most of the changes on this side are done and can be reviewed.

grakn-graph/src/main/java/ai/grakn/graph/internal/EdgeElement.java
```diff
@@ -33,11 +33,18 @@
...
     private final AbstractGraknGraph graknGraph;
 
-    EdgeImpl(Edge e, AbstractGraknGraph graknGraph){
+    EdgeElement(Edge e, AbstractGraknGraph graknGraph){
+        super(graknGraph, e.id());
+        edge = e;
+        this.graknGraph = graknGraph;
+    }
+
+    EdgeElement(AbstractGraknGraph graknGraph, Edge e){
```
aelred: So much choice of your order of arguments! (We should remove one)

grakn-graph/src/main/java/ai/grakn/graph/internal/InstanceImpl.java
```diff
@@ -88,26 +92,20 @@
      */
     @Override
     public void delete() {
-        InstanceImpl<?, ?> parent = this;
-        Set<CastingImpl> castings = parent.castings();
+        Set<Relation> relations = castingsInstance().map(Casting::getRelation).collect(Collectors.toSet());
```
aelred: This is picky, but there's no need to collect these. You can just stream over them.

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
```diff
@@ -58,20 +62,32 @@
...
-    void setHash(){
-        setUniqueProperty(Schema.ConceptProperty.INDEX, generateNewHash(type(), allRolePlayers()));
+    Stream<Casting> castingsRelation(RoleType... roleTypes){
+        if(roleTypes.length == 0){
+            return getEdgesOfType(Direction.OUT, Schema.EdgeLabel.SHORTCUT).
+                    map(edge -> getGraknGraph().getElementFactory().buildRolePlayer(edge));
+        }
+
+        //Traversal is used so we can potentially optimise on the index
+        Set<Integer> roleTypesIds = Arrays.stream(roleTypes).map(r -> r.getTypeId().getValue()).collect(Collectors.toSet());
```
aelred: Do we also want to retrieve role-players playing sub-types in this method? For example, `castingsRelation(graph.getMetaRoleType())`.

grakn-graph/src/main/java/ai/grakn/graph/internal/TxCache.java
```diff
@@ -128,57 +132,38 @@ void refreshOntologyCache(){
...
+            modifiedResources.add((ResourceImpl) element);
+        } else if (element.isRolePlayer()){
+            modifiedCastings.add(element.asRolePlayer());
         }
 
         //Caching of relations in memory so they can be retrieved without needing a commit
-        if (concept.isRelation()) {
-            RelationImpl relation = (RelationImpl) concept;
-            modifiedRelations.put(RelationImpl.generateNewHash(relation.type(), relation.allRolePlayers()), relation);
+        if (element.isRelation()) {
```
aelred: I think this `if` can be joined with the `if (element.isRelation())` above.

grakn-core/src/main/java/ai/grakn/util/Schema.java
```diff
@@ -53,6 +53,7 @@ private Schema() {
         RELATES("relates"),
         PLAYS("plays"),
         HAS_SCOPE("has-scope"),
+        //TODO: Remove Casting and Role Player edges
```
aelred: Will this be eventually done in this PR? I guess you're just waiting for analytics.

grakn-graph/src/main/java/ai/grakn/graph/internal/EdgeElement.java
fppt: Ooops. I meant to remove one. Will do so :)

grakn-graph/src/main/java/ai/grakn/graph/internal/InstanceImpl.java
fppt: Nope, wrong. There is a line below this and before actually iterating over the set that deletes the instance and therefore eliminates the ability to get the relations. 

It needs to be done in this wonky order otherwise we get null pointers when we empty relations and auto delete them.

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
fppt: No, this is used for internal testing + validation. Getting the roleplayers of the sub types will lead to extra processing for no reason. This method is not even on the public API.

grakn-graph/src/main/java/ai/grakn/graph/internal/TxCache.java
fppt: yes you are correct.

grakn-core/src/main/java/ai/grakn/util/Schema.java
fppt: yes I will do it in this PR. Once https://github.com/graknlabs/grakn/pull/1724 goes in.

grakn-graph/src/main/java/ai/grakn/graph/internal/InstanceImpl.java
aelred: Oh, I see. So some other method that `delete` calls is iterating over the relations?

fppt: Yep which in turn iterates over the instances of those relations. So if one thing is deleted in the wrong order you get a null pointer.

fppt:
@aelred re-review please

grakn-test/src/test/java/ai/grakn/test/engine/controller/TasksControllerTest.java
```diff
@@ -111,6 +112,7 @@ public void afterSendingTask_TheResponseTypeIsJson(){
     }
 
     @Test
+    @Ignore("Fails randomly when running on travis")
```
fppt: https://work.grakn.ai/entity/15482

fppt:
!rtg - weird failure

fppt:
!rtg

fppt:
!rtg - ai.grakn.test.engine.tasks.manager.TaskManagerTest.whenATaskIsStoppedDuringExecution_ItSavesItsLastCheckpoint


First time I have seen this one

PR #1723, opened 2017-06-06
Remove `GraknEngineConfig` and `RedisConnection` singletons

aelred:


grakn-engine/src/main/java/ai/grakn/engine/GraknEngineConfig.java
```diff
@@ -84,14 +84,8 @@
 
     private final int MAX_NUMBER_OF_THREADS = 120;
     private final Properties prop;
-    private static GraknEngineConfig instance = null;
     private int numOfThreads = -1;
 
-    public synchronized static GraknEngineConfig getInstance() {
```
fppt: Mission accomplished!

alexandraorth:
@fppt Just reviewed! Bye bye, singletons!

PR #1725, opened 2017-06-06
Ignore randomly failing CommitLogController test

alexandraorth:


fppt:
Merging despite the failures because this is ignoring tests only

PR #1727, opened 2017-06-07
Updated edges labels for plays, relates and has in HAL builder.

marco-scoppetta:


marco-scoppetta:
!rtg

grakn-graql/src/main/java/ai/grakn/graql/internal/hal/HALUtils.java
```diff
@@ -58,9 +58,9 @@
     final static String SUB_EDGE = "sub";
     final static String OUTBOUND_EDGE = "OUT";
     final static String INBOUND_EDGE = "IN";
-    final static String HAS_ROLE_EDGE = "has-role";
-    final static String HAS_RESOURCE_EDGE = "has-resource";
-    final static String PLAYS_ROLE_EDGE = "plays-role";
+    final static String RELATES_EDGE = "relates";
```
fppt: This information is redundant I think. Why not just use: `Schema.EdgeLabel.getLabel()`. The same for all the other edge labels as well.

PR #1728, opened 2017-06-07
Fix several bugs with matching implicit relations

aelred:


aelred:
!rtg - not sure what happened here: https://jenkins.grakn.ai/job/unit-test-pull-request-grakn-titan/ai.grakn$titan-factory/4085/console

aelred:
!rtg - another cassandra timeout

grakn-graql/src/main/java/ai/grakn/graql/internal/util/CommonUtil.java
```diff
@@ -48,6 +49,18 @@
 
     private CommonUtil() {}
 
+    public static <T> T withImplicitConceptsVisible(GraknGraph graph, Supplier<T> function) {
```
fppt: Might be worth using these methods across the whole stack. I.e. put it in grakn-core and use it everywhere.

aelred: That's true. Do you know a good place to put them?

I could also just move the whole of `CommonUtil` into `grakn-core`.

fppt: You mean like I suggested in my comment?

aelred: Oh, you're saying just move the whole class? I like that!

fppt: Sorry. Should have been more clear. My bad writing never ceases to confuse :)

PR #1730, opened 2017-06-07
Do not print Log directory when starting Grakn [BUG]

alexandraorth:


fppt:
@alexandraorth please create a merge of stable into master and reference it here.

fppt:
https://github.com/graknlabs/grakn/pull/1734

alexandraorth:
Thanks for doing that, @fppt. I'll make sure to, next time

PR #1731, opened 2017-06-07
Make test server use a random port every time

aelred:
Spark will always operate on a random, unused port.

However, ZooKeeper, Redis etc. are still using the default ports.


aelred:
!rtg - maven timeout, added to TP

aelred:
!rtg - titan strategy failures (fixed on stable)

burukuru:
!rtg

aelred:
!rtg - cassandra time out

grakn-test/src/test/java/ai/grakn/test/migration/json/JsonMigratorMainTest.java
```diff
@@ -65,12 +65,12 @@ public void setup() {
 
     @Test
     public void jsonMigratorMainTest(){
-        runAndAssertDataCorrect("-input", dataFile, "-template", templateFile, "-keyspace", graph.getKeyspace());
+        runAndAssertDataCorrect("-u", engine.uri(), "-input", dataFile, "-template", templateFile, "-keyspace", graph.getKeyspace());
     }
 
     @Test
     public void jsonMainDistributedLoaderTest(){
-        runAndAssertDataCorrect("-input", dataFile, "-template", templateFile, "-keyspace", graph.getKeyspace(), "-uri", "localhost:4567");
```
aelred: This test has changed to the extent that it doesn't really test anything. Any idea on how to resolve this?

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/ResourceDeduplicationTask.java
```diff
@@ -90,7 +90,16 @@ static void transact(GraknSession factory, Consumer<GraknGraph> work, String des
         
         private boolean deleteUnattached = false;
         private String keyspace;
-        
+        private String uri;
+
+        /**
+         * Specify the uri to use for the deduplication job.
+         */
+        public Job uri(String uri) {
```
alexandraorth: `Job` isn't such a great class name...

alexandraorth: But I now see that this was preexisting.

grakn-test/src/main/java/ai/grakn/test/GraknTestEnv.java
```diff
@@ -68,6 +74,19 @@ public static void ensureCassandraRunning() throws Exception {
     }
 
     /**
+     * Create a configuration for use in tests, using random ports.
+     */
+    static GraknEngineConfig createTestConfig() {
+        GraknEngineConfig config = GraknEngineConfig.create();
+
+        Integer serverPort = getEphemeralPort();
```
alexandraorth: What would happen if the port generated here conflicted with one of our in-use ports?

grakn-test/src/main/java/ai/grakn/test/GraknTestEnv.java
```diff
@@ -141,6 +160,15 @@ static void stopEngine(GraknEngineServer server) throws Exception {
         // There is no way to stop the embedded Casssandra, no such API offered.
     }
 
+    static Service startSpark(GraknEngineConfig config) {
+        LOG.info("starting spark on port " + config.uri());
+
+        Service spark = Service.ignite();
+        configureSpark(spark, config, JWTHandler.create(config.getProperty(JWT_SECRET_PROPERTY)));
```
alexandraorth: Should we account for the case where there is no `JWT_SECRET_PROPERTY` here?

grakn-test/src/main/java/ai/grakn/test/SparkContext.java
```diff
@@ -20,32 +20,26 @@
...
+
 /**
  * Context that starts spark
  * @author Felix Chapman
  */
 public class SparkContext extends ExternalResource {
 
     private final BiConsumer<Service, GraknEngineConfig> createControllers;
-    private final GraknEngineConfig config = GraknEngineConfig.create();
+    private GraknEngineConfig config = GraknTestEnv.createTestConfig();
```
alexandraorth: Why is this no longer final?

grakn-test/src/main/java/ai/grakn/test/GraknTestEnv.java
aelred: This calls `new ServerSocket(0)` which requests a new "ephemeral" port from the OS, which basically means that it isn't in use.

grakn-test/src/test/java/ai/grakn/test/migration/json/JsonMigratorMainTest.java
aelred: @alexandraorth can I poke you on this one? I want to remove the test, but not 100% sure it's a good idea.

alexandraorth: Go ahead and remove it. As far as I can tell, it should be identical to `jsonMigratorMainTest` now?

grakn-test/src/main/java/ai/grakn/test/SparkContext.java
aelred: ¯\_(ツ)_/¯

grakn-test/src/main/java/ai/grakn/test/GraknTestEnv.java
aelred: Yes. This is [an existing bug](https://work.grakn.ai/entity/15230), so I'm happy to just leave it for now.

aelred:
!rtg

Failure on `TaskManagerTest.whenRecurringTaskSubmitted_ThereIsAnIntervalBetweenExecutions` 

https://jenkins.grakn.ai/job/unit-test-pull-request-grakn-titan/ai.grakn$grakn-test/4242/testReport/junit/ai.grakn.test.engine.tasks.manager/TaskManagerTest/whenRecurringTaskSubmitted_ThereIsAnIntervalBetweenExecutions/

Logged to TP

PR #1732, opened 2017-06-07
Throw exception when there is a version mismatch

fppt:


grakn-graph/src/test/java/ai/grakn/factory/SystemKeyspaceTest.java
```diff
@@ -7,29 +7,64 @@
...
                 graph.getEntityType("keyspace").instances().forEach(Concept::delete);
+                graph.commit();
+            }
+        }
+    }
+
+    @Test
+    public void whenOpeningGraphBuiltUsingDifferentVersionOfGrakn_Throw(){
+        String versionResourceType = "system-version";
+        String rubbishVersion = "Hippo Version";
```
aelred: hmm

fppt: Hint is hinty :)

PR #1738, opened 2017-06-08
[Stable] Fix variable loss bug

kasper-piskorski:


fppt:
@kasper-piskorski please create a merge into master and reference it here.

kasper-piskorski:
Merged here: https://github.com/graknlabs/grakn/pull/1744

PR #1739, opened 2017-06-08
Update to latest Guava version

aelred:


aelred:
Unfortunately, it looks like Guava 19.0 is the latest version we can use, because we are using a very old version of Cassandra that depends on a method that exists (but deprecated) in 19.0, and is removed in 20.0.

aelred:
I'm closing this, since we can't currently support a newer Guava version due to Cassandra.

PR #1743, opened 2017-06-09
[Stable] Fix problems when unifying atoms with metaroles and ids

kasper-piskorski:


fppt:
@kasper-piskorski merge into master and reference here please.

kasper-piskorski:
Merged here: https://github.com/graknlabs/grakn/pull/1744

PR #1745, opened 2017-06-12
Fixed HAL shortest path response.

marco-scoppetta:
The array containing the shortest path response now contains a list of JSON objects instead of Strings, so we do not get parsing error on Dashboard


fppt:
@marco-scoppetta Please merge this into master and reference the merge here.

fppt:
https://github.com/graknlabs/grakn/pull/1751

PR #1747, opened 2017-06-12
Temporary patch to fix Null Pointers In merging castings

fppt:


aelred:
We don't intend to merge this to `master` right?

fppt:
we do but only for sorting out the git history

fppt:
https://github.com/graknlabs/grakn/pull/1751

PR #1748, opened 2017-06-13
refactored graph module to remove compile/test time dependency on graql

bolerio:
In grakn-graph module:

- Graql dependency is declared with <scope>runtime</scope> now
- A new QueryBuilderFactory interface is introduced to handle the response from a graph.graql() call, where the default construct a new instance of QueryBuilderImpl associated with the graph. No property configuration yet for this, probably will never be needed, but the implementation is dynamically instantiated through reflection
- A bunch of other small moves where needed, such as creating a StringUtil class in core with some of Graql's string utilities moved there
- There was one main "real" dependency remaining: rules do need actual patterns to be instantiated. The way I got around this (and it only came up in tests), was to use the query builder parsing methods.


fppt:
If https://github.com/graknlabs/grakn/pull/1749 goes in we can close this PR

fppt:
I have been through this and I am wondering why make `QueryBuilderFactory` part of the `InternalFactory` structure? Why not just leave it as a helper which can be called from within `AbstractGraknGraph#graql()` ?

As it stands `AbstractGraknGraph#graql()` creates a new instance of `QueryBuilder` each time anyway.

bolerio:
@fppt Maybe I don't understand well that notion of "internal factory", but it wasn't so much how many instanstances `AbstractGraknGraph#graql()` creates, more about the idea that the query builder is coupled to the graph and therefore the process of creating is in principle dependent on type of graph db implementation used, and we had mentioned having a factory during the call, so I went with a factory.  Once I decided that we'd go the "factory route", the question was who creates that factory and when. I thought the factory implementation, should it ever vary, would be coupled with the graph implementation, hence the addition to InternalFactory and voice the need of another "factory builder". That was my thought process :) Of course, feel free to change it - a simpler helper class would also do the job for now. One thing to make sure of is lookup that `Constructor` once only because that's what's most time consuming during reflection. After that calling the constructor `newInstance` is fast enough (not as fast as 'new' operator because of some extra security checks, but close).

fppt:
It's certainly more flexible but I think it's a flexibility we don't need. 

Personally I don't see the need for such a coupling. I doubt the `QueryBuilder` will ever need anything outside the context of what `AbstractGraknGraph` contains. 

What do you think @aelred ?

aelred:
Agreed that it's not needed right now. I'd say go for the simpler solution - if we need the extra indirection we can always re-introduce the factory!

bolerio:
@fppt @aelred Ok, sounds good, I'm fine either way. So @fppt I noticed you merged this into some other PR or something? Should I make the changes here and push them or have you already done so somewhere?

fppt:
@bolerio ignore that. That was a failed attempt and me getting this in faster because I thought I needed it for some other work. I was mistaken. You can continue working on this PR.

fppt:
Merge conflicts were resolved and merged in https://github.com/graknlabs/grakn/pull/1763

PR #1752, opened 2017-06-13
[Bug fix] Logs always relative to starting directory

alexandraorth:
This is a bug fix for #15307, where the logs were always being placed relative to the directory where grakn was starting. 

Now, if a relative log dir is specified in the properties file, it will be relative to the system variable grakn.dir. Otherwise the absolute value is used.

### DIFF
--- a/conf/main/logback.xml
+++ b/conf/main/logback.xml
@@ -46,7 +46,6 @@
     <!--Configure the "main" appender used for all non-post processing Grakn logs. Logs are kept for
         60 days and each file is capped at 50MB, but at most 1GB -->
     <appender name="MAIN" class="ch.qos.logback.core.rolling.RollingFileAppender">
-        <!--grakn.dir is a system variable, log.dirs is taken from the grakn.properties on the classpath-->
         <file>${configuredLogDir}/${graknLogFileName}</file>
         <encoder>
             <pattern>%d{dd-MM-yyyy HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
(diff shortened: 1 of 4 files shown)


conf/main/logback.xml
```diff
@@ -21,32 +21,46 @@
     <property name="graknLogFileName" value="grakn.log"/>
     <property name="graknPPLogFileName" value="grakn-postprocessing.log"/>
 
+    <!--Configure a properties file (located on the classpath) from which values can be referenced-->
     <property resource="grakn.properties"/>
 
+    <!--Extract the directory of the grakn logs from the properties file. If the path is relative it
+    will be appended to the directory containing grakn (system property grakn.dir). Otherwise the absolute
+    path is used. -->
+    <if condition='property("log.dirs").startsWith("/")'>
```
fppt: Ewwwwwwwwwww . . . Oh god. Conditionals in xml . . . ugh.

fppt:
@alexandraorth merge this into master and reference here please

conf/main/logback.xml
aelred: wow

aelred: there is no god

alexandraorth: I know :( Can you guys think of another way??

fppt: No, but that does not mean I won't judge you for putting conditional XML statements in

alexandraorth: </3

alexandraorth:
#1758

kasper-piskorski:
merge here: https://github.com/graknlabs/grakn/pull/1760

PR #1755, opened 2017-06-13
[Bug fix] System Keyspace race condition

alexandraorth:


grakn-graph/src/main/java/ai/grakn/factory/SystemKeyspace.java
```diff
@@ -186,8 +198,8 @@ public static void deleteKeyspace(String keyspace){
      * only consists of types, the inserts are idempotent and it is safe to load it
      * multiple times.
      */
-    private static void loadSystemOntology() {
-        try (GraknGraph graph = factory().open(GraknTxType.WRITE)) {
+    private static void loadSystemOntology(InternalFactory factory) {
```
fppt: Why pass in the factory now? This would essentially allow you to have multiple factories (graph connections) servicing the system keyspace. This is not ideal. The factory should be set internally to ensure only 1 graph connection can be made to the system keyspace.

grakn-graph/src/main/java/ai/grakn/factory/SystemKeyspace.java
```diff
@@ -82,10 +83,11 @@
     protected static final Logger LOG = LoggerFactory.getLogger(SystemKeyspace.class);
 
     private static final ConcurrentHashMap<String, Boolean> openSpaces = new ConcurrentHashMap<>();
-    private static final AtomicBoolean isFactorySet = new AtomicBoolean(false);
-    private static InternalFactory factory;
+    static final AtomicBoolean factoryBeingInstantiated = new AtomicBoolean(false);
```
fppt: Not sure if you can work around this but it would be better to maintain these as `private`. I know you did it so your mock class can access these fields but is there no other way to achieve the same result without having to change this implementation? 

Maybe override the `initialise` methods somehow? If there is no way so be it, but exposing fields for the sake of testing is a little bit bad.

grakn-graph/src/main/java/ai/grakn/factory/SystemKeyspace.java
alexandraorth: If you use the `factory()` method there is a circular dependency - you will be waiting for the countdown latch to release forever, because you are now blocking the thread that should release it.

fppt: Then the order of loading is incorrect. The factory should always be initialized before attempting to load the ontology.

grakn-graph/src/main/java/ai/grakn/factory/SystemKeyspace.java
alexandraorth: Unfortunately this is the best way that I could come up with and I hoped that package-private might not be so nasty. 

@aelred Do you have any ideas how we can better test this?

fppt:
@alexandraorth  Please merge this into master

alexandraorth:
#1764

alexandraorth:
Messed up #1767. Now merging with  #1768.

PR #1761, opened 2017-06-14
Test Refactor - Allowing test Profiles To Be executed on Different modules

fppt:


fppt:
@aelred This is far from done, but as we both know if we waited for this PR to be 100% complete it would take days and there would be over 2000 lines of code to review so I think this is "good enough" to go in.

What's basically done:

- Embedded Third party stuff, like cass, kafka, and redis is moved into a test tool library independent of grakn. It does not need anything form the grakn stack which means these test tools can be used anywhere.

- Graphs can be created without the need of a running engine. All that is needed is a dependency on `grakn-graph`

- We can now run "profiled tests" in any module. For example `GraphLoaderTest` is in engine (to be moved to graql later on) and it can create titan or tinker graphs without needing engine to be running. 

What's not done:

Everything else :)

grakn-test-tools/src/main/java/ai/grakn/util/GraknTestSetup.java
```diff
@@ -0,0 +1,70 @@
...
+public class GraknTestSetup {
+    private static String CONFIG = System.getProperty("grakn.test-profile");
+
+    /**
+     * Starts cassandra if needed.
+     */
+    public static void ensureCassandraRunning() {
+        if (GraknTestSetup.usingTitan()) {
+            EmbeddedCassandra.start();
+        }
```
aelred: I'm not 100% sure if we should call this `ensureCassandraRunning`, since it only sometimes does that. Maybe something to indicate it initialises the graph backend? Like `setupGraphDependencies` or similar.

grakn-test/src/test/java/ai/grakn/test/GraknTestEngineSetup.java
```diff
@@ -112,41 +98,20 @@ static GraknEngineServer startEngine(GraknEngineConfig config) throws Exception
...
     }
 
-    static void stopRedis(){
-        if (REDIS_COUNTER.decrementAndGet() == 0) {
-            LOG.info("Stopping Redis...");
-            redisServer.stop();
-            LOG.info("Redis stopped.");
-        }
+    static void stopKafka() throws Exception {
+        EmbeddedKafka.stop();
```
aelred: We should probably just get rid of this (but later :P )

grakn-graph/src/main/java/ai/grakn/factory/FactoryBuilder.java
```diff
@@ -53,13 +52,8 @@ private FactoryBuilder(){
 
     public static InternalFactory getFactory(String keyspace, String engineUrl, Properties properties){
         try{
-            String factoryType;
-            if (!Grakn.IN_MEMORY.equals(engineUrl)) {
-                factoryType = properties.get(FACTORY).toString();
-            } else {
-                factoryType = TinkerInternalFactory.class.getName();
-            }
-            return getGraknGraphFactory(factoryType, keyspace, engineUrl, properties);
```
fppt: Yep. This was a very redundant check

grakn-test-tools/src/main/java/ai/grakn/util/GraknTestSetup.java
fppt: Agreed. Will make that change in the next PR

grakn-test/src/test/java/ai/grakn/test/GraknTestEngineSetup.java
fppt: Yep again.

PR #1763, opened 2017-06-14
Split Graql/Grakn dependencies

fppt:
Just resolved the merge conflicts in https://github.com/graknlabs/grakn/pull/1748


grakn-graph/src/test/java/ai/grakn/generator/GraknGraphs.java
```diff
@@ -259,7 +262,7 @@ public GraknGraphs setOpen(boolean open) {
             },
             () -> {
                 RuleType ruleType = ruleType();
-                Rule rule = ruleType.putRule(var("x").pattern(), var("x").pattern());// TODO: generate more complicated rules
+                Rule rule = ruleType.putRule(graql.parsePattern("$x"), graql.parsePattern("$x"));// TODO: generate more complicated rules
```
aelred: This is pretty nasty, but I can't think of a better way to handle it right now.

grakn-graql/src/main/java/ai/grakn/graql/internal/util/StringConverter.java
```diff
@@ -48,48 +46,6 @@
...
-            df.setMinimumFractionDigits(1);
-            df.setMaximumFractionDigits(12);
-            df.setMinimumIntegerDigits(1);
-            return df.format(value);
-        } else {
-            return value.toString();
-        }
-    }
-
-    /**
```
aelred: Ideally we should see if all of this can be moved to `StringUtil`, instead of having two `String` helper classes.

PR #1768, opened 2017-06-14
Merge from stable

alexandraorth:


alexandraorth:
The stall we see here is fixed by  #1778, so we need to wait for that to be merged to stable

alexandraorth:
As seen in #1780

PR #1773, opened 2017-06-14
Make system keyspace a singleton

alexandraorth:


alexandraorth:
!rtg - `ai.grakn.test.engine.controller.CommitLogControllerTest.whenCommittingGraph_TaskManagerReceivesCountTask`

alexandraorth:
@fppt When you have time, please go over this with very keen eyes. It fixes the problem we were having about the system keyspace not being initialised, but I get very afraid of breaking things whenever I change anything in graph :P

PR #1781, opened 2017-06-15
Change log format to ISO8601

burukuru:
eg. 2017-06-01 19:00:09,701

### DIFF
--- a/conf/main/logback.xml
+++ b/conf/main/logback.xml
@@ -48,7 +48,7 @@
     <appender name="MAIN" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${configuredLogDir}/${graknLogFileName}</file>
         <encoder>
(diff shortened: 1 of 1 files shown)


conf/main/logback.xml
```diff
@@ -48,7 +48,7 @@
     <appender name="MAIN" class="ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${configuredLogDir}/${graknLogFileName}</file>
         <encoder>
-            <pattern>%d{dd-MM-yyyy HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+            <pattern>%date{ISO8601} HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
```
alexandraorth: I think `%date{ISO8601}` includes the time? we might not need the `HH:mm:ss.SSS` bit. (https://logback.qos.ch/manual/layouts.html) Unless you've seen something different when running?

burukuru:
!rtg

PR #1783, opened 2017-06-15
Un-ignore randomly failing TasksController test

alexandraorth:


alexandraorth:
Reproduced by running 1000 times locally (it failed 4 times). Passed 1000 times locally after the fix.

PR #1785, opened 2017-06-16
Split engine tests

bolerio:
. This is a first pass putting a few controller tests back to engine module. Take a look at how UserControllerTest work for example. I think the base class for tests is a good idea if used properly as it avoid repeating setup code in many tests classes and thus makes it a bit easier to split the tests into more classes if necessary. The idea is, as I explained in my email, to start the complete engine server once and keep it running. 

I added an example of state cleanup, delete a user that was created during the test. The cleanup itself checks an assertion and note that that assertion actually caught a bug in the controller which didn't really have a functioning user delete operation.  The cleanup is again simple: tests add closures that get collected and called at the end of the class.

I think this setup gives enough isolation, is simple enough and lends itself to writing sort of functional/realistic tests


grakn-engine/src/test/java/ai/grakn/engine/EngineTestHelper.java
```diff
@@ -0,0 +1,68 @@
...
+ * <p>
+ * This class instantiates an engine server-side component once, statically until the JVM exists.
+ * Any state cleanup needs to be done by tests themselves - more code, more work, but more performant
+ * test suite. So all unit tests will share the same components and will be written with that in mind.  
+ * </p>
+ * 
+ * @author borislav
+ *
+ */
+public class EngineTestHelper {
```
fppt: I have just noticed that this class is inside the engine module. We said that we would be keeping the integration tests in the test module. I imagine the integration tests would need a class very similar if not exactly this. Or am I mistaken? 

I am just wondering how we can minimise the code duplication.

grakn-engine/src/test/java/ai/grakn/engine/controller/ControllerFixture.java
```diff
@@ -0,0 +1,75 @@
...
+    private void restAssuredSetup() {
+        RestAssured.baseURI = "http://localhost:" + EngineTestHelper.config().getProperty(GraknEngineConfig.SERVER_PORT_NUMBER);
+        RestAssured.requestSpecification = new RequestSpecBuilder()
+                .addHeader("Accept", "application/json")
+                .addHeader("Content-Type", "application/json").build();        
+    }
+    
+    private void doCleanup() {
+        final ArrayList<Throwable> cleanupFailures = new ArrayList<Throwable>();
+        cleanupOperations.forEach(r -> {
```
fppt: Is the clean up run after every individual test or after every test class?

grakn-engine/src/test/java/ai/grakn/engine/EngineTestHelper.java
bolerio: Yes, maybe. We can move it then. Or pull it out of the tests altogether and have it in the main artifact as a utility class to start an embedded engine. But to share code b/w module tests and the integration tests, the idea is to put that code in the test-tools module, right?  The question then is what are dependencies of this module? Is it ok for it to depend on any of the other grakn modules?

grakn-engine/src/test/java/ai/grakn/engine/controller/ControllerFixture.java
bolerio: This mechanism is set up so that it can be run at the end of the test class. To run them after a test itself, it can be just a try/catch within the test method. Or we could add another rule/utility to handle it with lambdas in the same way which would be more concise but perhaps a bit weird.

grakn-engine/src/test/java/ai/grakn/engine/EngineTestHelper.java
bolerio: Anyway, @fppt let me know if you'd rather put this class in the test tools module. Everything I did with this PR initially was with the goal to communicate the approach (better with code than lots of explanations). We can move things around now.

grakn-test/src/test/java/ai/grakn/test/engine/controller/UserControllerTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.mockito.stubbing.Answer;
+
+import static com.jayway.restassured.RestAssured.get;
+import static com.jayway.restassured.RestAssured.given;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@Ignore
+public class UserControllerTest {
```
aelred: It looks like the merge commit has accidentally created a duplicate `UserControllerTest`.

bolerio:
@aelred right...the whole "controller" testing package from grakn-test will disappear, and the duplicate one with it :)

PR #1789, opened 2017-06-19
New graql planner with min spanning tree

duckofyork:
New graql query planner based on Chu–Liu/Edmonds' algorithm and greedy arborescence traversal


grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/DataTypeFragment.java
```diff
@@ -48,7 +48,7 @@ public String getName() {
 
     @Override
     public double fragmentCost(double previousCost) {
```
aelred: We can remove the `previousCost` argument now.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/spanningtree/Arborescence.java
```diff
@@ -0,0 +1,102 @@
...
+import com.google.common.collect.Sets;
+
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * @param <V> the type of the nodes
+ * @author Jason Liu
+ */
```
aelred: Given how abstract this class is, it could probably do with a quick description (or maybe a link to a definition)

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/spanningtree/Arborescence.java
```diff
@@ -0,0 +1,102 @@
...
+    public static <T> Arborescence<T> of(ImmutableMap<T, T> parents) {
+        Arborescence<T> arborescence = new Arborescence<>(parents);
+        if (parents != null && !parents.isEmpty()) {
+            HashSet<T> allParents = Sets.newHashSet(parents.values());
+            allParents.removeAll(parents.keySet());
+            if (!allParents.isEmpty()) {
+                arborescence.root = allParents.iterator().next();
+            }
+        }
+        return arborescence;
```
aelred: Probably simpler to set the root once in the constructor: `return new Arborsecence<>(parents, root);`

Then it can be `final` too 👍

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/spanningtree/EdgeQueueMap.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @param <V> the type of the nodes stored
+ * @author Sam Thomson
+ * @author Jason Liu
+ */
```
aelred: Another class that could do with a short description.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/spanningtree/ExclusiveEdge.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public static <T> ExclusiveEdge<T> of(DirectedEdge<T> edge, double weight) {
+        return ExclusiveEdge.of(edge, ImmutableList.of(), weight);
+    }
+
+    @Override
+    public int compareTo(ExclusiveEdge<V> exclusiveEdge) {
+        return Doubles.compare(weight, exclusiveEdge.weight);
+    }
```
aelred: If we make two edges `e1` and `e2` with the same weight, but different properties then we get `e1.compareTo(e2) == 0`, which is not consistent with `equals`.

I think this means that e.g. putting an `ExclusiveEdge` in a tree set will mess up.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/spanningtree/KBestArborescences.java
```diff
@@ -0,0 +1,228 @@
...
+ * along with Grakn. If not, see <http://www.gnu.org/licenses/gpl.txt>.
+ */
+
+package ai.grakn.graql.internal.gremlin.spanningtree;
+
+import ai.grakn.graql.internal.gremlin.spanningtree.graph.DirectedEdge;
+import ai.grakn.graql.internal.gremlin.spanningtree.graph.WeightedGraph;
+import ai.grakn.graql.internal.gremlin.spanningtree.util.Pair;
+import ai.grakn.graql.internal.gremlin.spanningtree.util.Weighted;
+import com.google.common.base.Optional;
```
aelred: We're using Guava's `Optional` in a few places, but we should stick to the built-in Java one.

grakn-graql/src/test/java/ai/grakn/graql/internal/gremlin/spanningtree/ChuLiuEdmondsTest.java
```diff
@@ -0,0 +1,176 @@
...
+		(0) -------> (1) \
+		 |  \       /  ^  \
+		 |   \30   |   |20 \
+		 |10  \    |10 |    \10
+		 |     \   |  /      \
+		 V  15  V  V /   20   V
+		(3)<----- (2) -----> (4)
+		  \-------^
+		     40
+		 */
```
aelred: wowwww

duckofyork: Beautiful, right?

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/spanningtree/Arborescence.java
duckofyork: I think we need some basic validation here just to make sure the edge map represents a valid arborescence, though I have made root final.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/spanningtree/ExclusiveEdge.java
duckofyork: Is it necessary to make them consistent? compareTo is only used to sort exclusiveEdge by weight.

PR #1790, opened 2017-06-19
Moving more graql tests back into graql

fppt:
This sweep includes:

- Moving all the test graphs into `grakn-test-tools`
- Moving all the reasoner test except `BenchmarkTest` into `grakn-graql`


fppt:
!rtg

PR #1791, opened 2017-06-19
Remove `PatternBuilder` and `VarPatternBuilder`

aelred:


grakn-core/src/main/java/ai/grakn/graql/PatternBuilder.java
```diff
@@ -1,38 +0,0 @@
...
-package ai.grakn.graql;
-
-import javax.annotation.CheckReturnValue;
-
-/**
- * An interface representing something that is or can be transformed into a {@link Pattern}.
- *
- * @author Felix Chapman
- */
-public interface PatternBuilder {
```
fppt: yay! I never liked having an interface with one method on it :)

PR #1793, opened 2017-06-19
BUG - tinker graph computer was not working with analytics

sheldonkhall:
- The main part of this PR is the change of dependency to our version of Tinkerpop
- This has a small patch to the tinkergraph computer
- more details here: https://issues.apache.org/jira/browse/TINKERPOP-1519


Jenkinsfile
```diff
@@ -0,0 +1,72 @@
+#!groovy
+//properties([pipelineTriggers([cron('H H/6 * * *')])])
+node('slave1-dev-jenkins') {
```
fppt: I am not going to even pretend I understand this file. What's this for? Is it related to this PR?

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/ShortestPathVertexProgram.java
```diff
@@ -249,6 +249,8 @@ public boolean terminate(final Memory memory) {
                 if (predecessors.length > 2) {
                     memory.set(MIDDLE, predecessors[2]);
                 }
+                //todo: be careful in tinkergraph as things set here cannot be got
```
fppt: Nitpick `//TODO:`

grakn-test/src/test/java/ai/grakn/test/graql/analytics/AnalyticsTest.java
```diff
@@ -74,9 +74,6 @@ public void setUp() {
 
     @Test
     public void testInferredResourceRelation() throws InvalidGraphException {
-        // TODO: Fix on TinkerGraphComputer
```
fppt: These removals must have felt great! :)

pom.xml
```diff
@@ -48,7 +48,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <java.version>1.8</java.version>
-        <tinkerpop.version>3.1.1-incubating</tinkerpop.version>
```
fppt: Cool. But a nitpick for a future PR is to change 

```
<titan.version>1.1.0.grakn-SNAPSHOT</titan.version>
```

to 

```
<titan.version>1.1.0-grakn-SNAPSHOT</titan.version>
```

Or maybe just make that change when we migrate to Janus. Anyway, just noting it here.

Jenkinsfile
sheldonkhall: sorry this was not supposed to be in the PR I will remove it

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/ShortestPathVertexProgram.java
sheldonkhall: will change seeing as I have to push a change anyway

pom.xml
sheldonkhall: ok, this requires redeploying so will put it off

PR #1795, opened 2017-06-19
Move System Keyspace

alexandraorth:


alexandraorth:
!rtg

grakn-engine/src/main/java/ai/grakn/engine/controller/SystemController.java
```diff
@@ -71,10 +76,45 @@
     private final Logger LOG = LoggerFactory.getLogger(SystemController.class);
     private final EngineGraknGraphFactory factory;
 
-    public SystemController(EngineGraknGraphFactory factory, Service spark) {
+    public SystemController(EngineGraknGraphFactory factory , Service spark) {
         this.factory = factory;
-        spark.get(KEYSPACES,     this::getKeyspaces);
-        spark.get(CONFIGURATION, this::getConfiguration);
+
+        spark.get(INITIALISE,         this::initialiseSession);
```
fppt: Isn't `spark.get(CONFIGURATION,      this::getConfiguration);` now redundant because of this? Why maintain both?

grakn-engine/src/main/java/ai/grakn/engine/factory/EngineGraknGraphFactory.java
```diff
@@ -50,19 +54,29 @@ public static EngineGraknGraphFactory create(Properties properties) {
...
+        this.systemKeyspace = new SystemKeyspace(this);
     }
 
     public synchronized void refreshConnections(){
         FactoryBuilder.refresh();
     }
 
     public GraknGraph getGraph(String keyspace, GraknTxType type){
-        return FactoryBuilder.getFactory(keyspace, Grakn.DEFAULT_URI, properties).open(type);
+        if(!keyspace.equals(SystemKeyspace.SYSTEM_GRAPH_NAME)) {
```
fppt: Just to point something out. This is what is going to guarantee the load order. So in the end wether we use a real singleton or a static class or whatever the load order has to be ensured somewhere.

grakn-engine/src/main/java/ai/grakn/engine/util/SystemOntologyElements.java
```diff
@@ -1,45 +0,0 @@
...
-import ai.grakn.concept.TypeLabel;
-
-/**
- * <p>
- *     Describes the system ontology elements related to task loading
- * </p>
- *
- * @author Denis Lobanov
- */
-public interface SystemOntologyElements {
```
fppt: I did not even know this existed but I am glad to see it die

grakn-factory/titan-factory/src/main/java/ai/grakn/graph/internal/GraknTitanGraph.java
```diff
@@ -73,16 +72,6 @@ public void openTransaction(GraknTxType txType){
     }
 
     @Override
-    public void closeSession(){
-        super.closeSession();
-
-        //Close the system graph if possible
-        if(!getKeyspace().equalsIgnoreCase(SystemKeyspace.SYSTEM_GRAPH_NAME)) {
```
fppt: YES! I did not even think about this but this is another wonderful benefit of making the system keyspace an engine concern

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -661,8 +660,10 @@ public void delete() {
         clearGraph();
         txCache().closeTx(ErrorMessage.CLOSED_CLEAR.getMessage());
 
+        //TODO We should not hit the REST endpoint when deleting keyspaces through a graph
+        //TODO retrieved from and EngineGraknGraphFactory
```
fppt: Stylistic issue:

Not 

```
//TODO line 1
//TODO line 2
```

rather

```
//TODO line 1
              line 2
```

Otherwise multiple todos are registered on the IDE when there is really only 1 todo

grakn-engine/src/main/java/ai/grakn/engine/controller/SystemController.java
alexandraorth: Sometimes I query for the configuration using the URL just to check its contents. Figured it was worth maintaining for that reason?

fppt: When exactly do you do that outside the context of creating a graph connection?

alexandraorth: Just as a user of the system. Also, dashboard calls the plain `/configuration` endpoint.

fppt: I don't see those as valid enough use cases to maintain two controllers which essentially do the same thing. Just more code to maintain in my eyes. I would make the keyspace parameter optional and just have one endpoint.

grakn-engine/src/main/java/ai/grakn/engine/factory/EngineGraknGraphFactory.java
alexandraorth: Yeah, I get what you are saying. Right now it is neither a singleton nor a static class but because we instantiate it in the server constructor and pass the object around, its alright.

grakn-engine/src/main/java/ai/grakn/engine/controller/SystemController.java
alexandraorth: I'm not sure. I feel that we should be separating the functionality between endpoints that modify the backend and endpoints that only get data. It feels a bit odd to be writing to the graph based on the presence of an optional parameter.

Writing this, I realised that the `INITIALISE` endpoint should be a `POST` no matter what, because it does modify the graph.

Ideally I would like `INITIALISE` to only add the keyspace to the system graph and not return the properties. But I realise that would mean that graph needs to make two API calls, which is also not something we want.

fppt: @alexandraorth can you please make sure there is a task to authenticate the config controller and also minimise the config options returned by it.

PR #1796, opened 2017-06-20
general graql execute query endpoint

bolerio:


grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
```diff
@@ -108,6 +109,35 @@ public GraqlController(EngineGraknGraphFactory factory, Service spark) {
...
+            if(!validContentType(acceptType, query)) {
+                throw GraknServerException.contentTypeQueryMismatch(acceptType, query);
+            }            
+            if (query instanceof DeleteQuery) {
+                query.execute();
+                graph.commit();
+                return respond(response, query, APPLICATION_JSON, Json.object());
+            }
+            else if (query instanceof InsertQuery) {
+                return respond(response, query, APPLICATION_JSON, executeInsertQuery((InsertQuery) query));
```
alexandraorth: We should be committing when executing an insert query.

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
```diff
@@ -108,6 +109,35 @@ public GraqlController(EngineGraknGraphFactory factory, Service spark) {
         spark.exception(InvalidGraphException.class, (e, req, res) -> handleError(422, e, res));
     }
 
+    @POST
+    @Path("/execute")
+    @ApiOperation(value = "Execute an arbitrary Gralql queryEndpoints used to query the graph by ID or Graql match query and build HAL objects.")
+    @Produces({"application/json", "text/plain"})
```
alexandraorth: It looks like this method can also produce hal, graql and graql+json?

grakn-test/src/test/java/ai/grakn/test/engine/controller/GraknExecuteQueryControllerTest.java
```diff
@@ -0,0 +1,93 @@
...
+    @ClassRule
+    public static GraphContext graphContext = GraphContext.preLoad(MovieGraph.get());
+
+    @ClassRule
+    public static SparkContext sparkContext = SparkContext.withControllers(spark -> {
+        new GraqlController(engineGraknGraphFactory, spark);
+    });
+        
+    @Test
+    public void testMatchQuery() {
```
alexandraorth: We need a bunch more tests for these endpoints - what happens when you send invalid queries, no accept type, compute queries, etc.

alexandraorth: In fact, we should probably make sure that all of the tests from the other three endpoints work against this one before removing the other endpoints.

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
bolerio: good catch!

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
bolerio: Yes, true, but I will remove this line, I think it's a leftover while I was troubleshooting why the method wasn't called at some point. The same annotation is specified at the class level and it's absent in all other methods. I doubt Spark is using it at all.

alexandraorth: Yeah, I think because these annotations are only used for the documentation we've sort of slacked on using them/keeping them up to date.

grakn-test/src/test/java/ai/grakn/test/engine/controller/GraknExecuteQueryControllerTest.java
bolerio: I had a compute query but it has problems starting spark (nothing to do with engine or the controller or the test) so I took it out for the moment because I know Sheldon is getting rid of that dependency. Otherwise, yes, good idea to add bad requests (it returns 400). However, I'm not sure when the other controllers are going away. We discussed replacing them with GraphQL, which I spent some time investigating so we should meet and chat about it....

bolerio:
@alexandraorth this is failing on the BatchMutatorClient ...didn't we take that test out?

alexandraorth:
@bolerio We took out one of the tests, the logging one, but not the one that is failing.

bolerio:
@alexandraorth do you have a suggestion? if you can't tell why it's failing from the logs, I'd ignore and enter a TP issue for you to look at it later...

alexandraorth:
I took a look earlier and wasn't able to tell. Its possible that it was caused by the cassandra timeouts? I would just retest this.

bolerio:
!rpg

bolerio:
!rtg

PR #1797, opened 2017-06-20
Migration more gracefully handles failure client side

alexandraorth:


grakn-migration/json/src/main/java/ai/grakn/migration/json/JsonMigrator.java
```diff
@@ -46,28 +46,30 @@
...
-                .filter(Optional::isPresent)
-                .map(Optional::get)
-                .forEach(JsonMigrator::runJson);
+        try {
+            MigrationCLI.init(args, JsonMigrationOptions::new).stream()
+                    .filter(Optional::isPresent)
+                    .map(Optional::get)
+                    .forEach(JsonMigrator::runJson);
+        } catch (Throwable throwable){
+            MigrationCLI.die(throwable.getMessage());
```
fppt: Is `die` still needed? We managed to get rid of it everywhere else.

alexandraorth: I think the method was never renamed - it doesn't actually kill anything anymore. Just prints a message to system.err

fppt: But it throws another exception correct? I think it's a bit unintuitive to do it this way especially since all it does it rewrap the exception into a runtime exception.

alexandraorth: Nope, doesn't even throw an exception anymore.

fppt: Ah okay. So new question, why not just inline it and do `System.err.println(errorMsg);` or are we planning for more handling to go into that method? 

One line methods which only print stuff should usually be inlined.

alexandraorth: Yeah, we can probably do that. I did wonder, however, if we should log the whole error to the logs and print the message to system.err. Do you think that would be helpful? (In which case, we should keep the method?)

fppt: Hmmm that is a good point. When you migrating data is it all or nothing? I.e. can migration fail midway and only some of the data makes it in? 

If that is the case then we may want to do both, bring the error to the user and log it, so in that case keep the method but rename it because it's certainly not killing anything.

alexandraorth: Migration can fail midway and continue migrating the data, yes. It will only fail the queries that caused the exception/had incorrect data. That is handled by a different method, though: 

when this method is called it is an all-or-nothing situation. It is always called right before the program will end.

fppt: If it's all or nothing then in my opinion in-line it. Only the user should be aware of a failure in an all or nothing situation. Logs are reserved fro tracking issues which can affect us in the long run. In this case nothing went in so nothing to worry about.

alexandraorth: Alright, I'll make the change to inline it, then

alexandraorth:
!rtg - not sure why

alexandraorth:
!rtg - cassandra timeout

PR #1799, opened 2017-06-20
Fixed relation ad roleplayer bug in HAL builder

marco-scoppetta:
- Removed unnecessary check that would not render relation nodes when they are roleplayers


marco-scoppetta:
This fix is needed for tomorrow's demo, and the test would require to rewrite some ontology to re-create this use case, so I have added a [TP task](https://work.grakn.ai/entity/15744) to write a proper task in the coming days

fppt:
@marco-scoppetta please create a merge into master and reference it here.

fppt:
https://github.com/graknlabs/grakn/pull/1802

PR #1801, opened 2017-06-20
Check engine is actually running when shell starts

aelred:


aelred:
!rtg - cassandra timeout stuff

PR #1803, opened 2017-06-21
XML Migration Attribute support

alexandraorth:


grakn-graql/src/main/java/ai/grakn/graql/internal/template/Scope.java
```diff
@@ -119,7 +118,7 @@ else if(!isGlobalScope()) {
         }
         else {
             // Unknown variable
-            return ObjectUtils.NULL;
+            return null;
```
fppt: SO why the change to ObjectUtils.NULL? According to the java docs it is used when:

> used as a null placeholder where null has another meaning. 

Is this the case? Does null have another meaning in these contexts?

PR #1805, opened 2017-06-21
Fixed bug in HALBuilder for explore requests

marco-scoppetta:
- HALExploreConcept were injecting the wrong URI in "explore" links


fppt:
@marco-scoppetta please merge this into master and reference it here.

marco-scoppetta:
Mergin in master with [#1806](https://github.com/graknlabs/grakn/pull/1806)

PR #1807, opened 2017-06-22
Refactor Instance and Meta Concept

fppt:
- Meta Concept is now Meta Thing
- Instance Class is not Thing

### DIFF
--- a/conf/test/cluster/grakn.properties
+++ b/conf/test/cluster/grakn.properties
@@ -54,7 +54,7 @@ tasks.postprocessing.delay=60000
 #Keyspace to be used when none is provided
 graphdatabase.default-keyspace=grakntest
 
-#A Type will be sharded when this instance threshold is hit
+#A Type will be sharded when this thing threshold is hit
 graph.sharding-threshold=100000
 
 # Graph ontology caching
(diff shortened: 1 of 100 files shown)


conf/test/cluster/grakn.properties
```diff
@@ -54,7 +54,7 @@ tasks.postprocessing.delay=60000
 #Keyspace to be used when none is provided
 graphdatabase.default-keyspace=grakntest
 
-#A Type will be sharded when this instance threshold is hit
+#A Type will be sharded when this thing threshold is hit
```
aelred: hmm

fppt:
@aelred I am glad I can entertain you :)

fppt:
!rtg

grakn-core/src/main/java/ai/grakn/concept/Concept.java
```diff
@@ -61,12 +61,12 @@
     Type asType();
 
     /**
-     * Return as an Instance if the Concept is an Instance.
+     * Return as an Thing if the Concept is an Thing.
```
aelred: There are a lot of cases of "an Thing" in the javadocs now.

PR #1808, opened 2017-06-22
Add test for multiple variable deletion

aelred:


aelred:
!rtg - a timeout bug fixed on master

PR #1810, opened 2017-06-23
Split engine tests

bolerio:
Moved lots of engine tests from grakn-test to grakn-engine module. Preserved the mocking approach in most places. Haven't gone through the tests in detail to ensure cleanup, need to do that next. But the restructuring itself is quite time consuming. And there is one problem: I couldn't find a way to avoid duplicating the SparkContext rule, it's now both in grakn-test and a variation of it is in grakn-engine. The two versions are slightly different and maybe it's ok to have it like this. That small duplications avoid test-tools package to depend on engine which would drag a huge pile of things with it!


bolerio:
!rtg

PR #1811, opened 2017-06-23
Ignore visualiser test until we figure out why it fails.

marco-scoppetta:


marco-scoppetta:
!rtg - Gral tests on travis and grakn-dist `testMarriageCount` on jenkins

PR #1814, opened 2017-06-23
Make `grakn.sh` wait for Grakn to completely initialise before returning

aelred:


grakn-dist/src/bin/grakn-engine.sh
```diff
@@ -41,6 +41,26 @@ done
 # Add path containing logback.xml
 CLASSPATH="$CLASSPATH":"${GRAKN_HOME}"/conf/main/
 
+wait_for_engine() {
+    local now_s=`date '+%s'`
+    local stop_s=$(( $now_s + $ENGINE_STARTUP_TIMEOUT_S ))
+    local status_thrift=
```
burukuru: Don't need this.

aelred:
!rtg - cassandra timeout

grakn-client/src/main/java/ai/grakn/client/Client.java
```diff
@@ -55,7 +57,20 @@
             return reader.lines().collect(joining("\n"));
         }
     };
-    
+
+    public static void main(String[] args) throws IOException {
+        String confPath = System.getProperty("grakn.conf");
```
alexandraorth: Should we extract this property into core?

grakn-client/src/main/java/ai/grakn/client/Client.java
```diff
@@ -55,7 +57,20 @@
...
+        Properties properties = new Properties();
+        try (FileInputStream stream = new FileInputStream(confPath)) {
```
alexandraorth: I think here we should check if the file exists and print a nice, helpful message if it does not.

aelred:
@alexandraorth I tried to find some to remove, but only found one in the bash tests. The others (such as `DistributionContext`) do not call `grakn.sh`, so can't use our new fancy method.

grakn-client/src/main/java/ai/grakn/client/Client.java
aelred: I _guess_ so...

grakn-client/src/main/java/ai/grakn/client/Client.java
aelred: Good point!

grakn-client/src/main/java/ai/grakn/client/Client.java
alexandraorth: What do you think @fppt ? Should we have configuration constants in core?

fppt: Put it somewhere yes. The less duplicated hardcoded strings we have the better. It's always a pain to change one and have to hunt down the 500 others you missed.

aelred: I've done it in a separate PR #1815. I'll wait for that to merge, then use it here (so don't merge this one yet!)

aelred:
@alexandraorth @burukuru please re-review!

PR #1815, opened 2017-06-26
Remove use of hard-coded system property strings

aelred:


grakn-core/src/main/java/ai/grakn/GraknSystemProperty.java
```diff
@@ -0,0 +1,48 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * @author Felix Chapman
+ */
+public enum GraknSystemProperty {
+
+    CURRENT_DIRECTORY("grakn.dir"),
+    CONFIGURATION_FILE("grakn.conf"),
+    GRAPH_CONFIG_PROPERTY("grakn.conf"),
```
alexandraorth: Why are there two properties that are the same?

aelred: all better

aelred:
!rtg - spooky cassandra problem. filipe on high alert

grakn-core/src/main/java/ai/grakn/GraknSystemProperty.java
```diff
@@ -0,0 +1,47 @@
...
+ * along with Grakn. If not, see <http://www.gnu.org/licenses/gpl.txt>.
+ *
+ */
+
+package ai.grakn;
+
+import javax.annotation.Nullable;
+
+/**
+ * @author Felix Chapman
```
fppt: Bad javadocs are bad. We should elaborate the purpose of this class.

aelred:
@fppt I added javadocs, but Github doesn't seem to care? :/

Undeniable proof: https://github.com/aelred/grakn/commits/system-properties

PR #1821, opened 2017-06-28
Better error for reserved words

fppt:


fppt:
!rtg- cassandra timeout

fppt:
!rtg

PR #1822, opened 2017-06-28
Refactor Due To Roles no longer being types

fppt:


CHANGELOG
```diff
@@ -68,7 +68,7 @@ Version v0.8.0
 - Re-add unintentionally deleted test (#772)
 - Fixing quotes in migration.sh (#771)
 - Admin API for Grakn Graphs (#770)
-- roleTypes map bug fixed (#769)
+- roles map bug fixed (#769)
```
aelred: We shouldn't change the changelog!

grakn-core/src/main/java/ai/grakn/util/Schema.java
```diff
@@ -119,7 +119,7 @@ public static boolean isMetaLabel(TypeLabel label) {
         //Ontology Elements
         ONTOLOGY_ELEMENT(OntologyConcept.class),
         TYPE(Type.class),
-        ROLE_TYPE(RoleType.class),
+        ROLE_TYPE(Role.class),
```
aelred: `ROLE_TYPE` -> `ROLE`

CHANGELOG
fppt: oops

grakn-core/src/main/java/ai/grakn/util/Schema.java
fppt: good catch

fppt:
@aelred re-review please

PR #1824, opened 2017-06-29
[DO NOT MERGE] Engine API consistency refactor

alexandraorth:
+ Graql controller does not always return a JSON object - depends on content type
+ Graql insert endpoint can return either JSON or text, if requested.


alexandraorth:
@marco-scoppetta needs to make dashboard changes before merging this one

PR #1826, opened 2017-07-03
Change Graql endpoint to return results formatted like the shell

aelred:


grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
```diff
@@ -112,25 +107,14 @@ private Object executeGraql(Request request, Response response) {
...
-            else if (query instanceof InsertQuery) {
-                Object resp = respond(response, APPLICATION_JSON, executeQuery(request, query, acceptType));
-                graph.commit();
-                return resp;
-            }
-            else {
-                return respond(response, acceptType, executeQuery(request, query, acceptType));
-            }
+            Object resp = respond(response, acceptType, executeQuery(keyspace, limitEmbedded, query, acceptType));
+            graph.commit();
```
fppt: Need to double check that this does not add overhead when just doing match queries

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
```diff
@@ -112,25 +107,14 @@ private Object executeGraql(Request request, Response response) {
         String keyspace = mandatoryQueryParameter(request, KEYSPACE);
         boolean infer = parseBoolean(mandatoryQueryParameter(request, INFER));
         boolean materialise = parseBoolean(mandatoryQueryParameter(request, MATERIALISE));
+        int limitEmbedded = queryParameter(request, LIMIT_EMBEDDED).map(Integer::parseInt).orElse(-1);
         String acceptType = getAcceptType(request);
+
         try(GraknGraph graph = factory.getGraph(keyspace, WRITE)){
```
fppt: Worth noting that making this a `READ` transaction when you are indeed only reading has some tiny performance benefits. Basically it allows some more aggressive central caching to be done.

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
```diff
@@ -268,66 +254,34 @@ private Object respond(Response response, String contentType, Object responseBod
...
+     * @param keyspace the keyspace the query is running on
      * @param query read query to be executed
      * @param acceptType response format that the client will accept
      */
-    private Object executeQuery(Request request, Query<?> query, String acceptType){
-        switch (acceptType){
+    private Json executeQuery(String keyspace, int limitEmbedded, Query<?> query, String acceptType){
+        Printer<?> printer;
+
+        switch (acceptType) {
```
fppt: The printer API finally makes sense to me!

PR #1828, opened 2017-07-03
Use graql planner when determining resolution order

kasper-piskorski:


kasper-piskorski:
please retest this please - grakn-dist failure

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/sets/ResourceIndexFragmentSet.java
```diff
@@ -91,7 +92,11 @@ private static void optimise(
         // Add a new fragment set to replace the old ones
         Var resource = valueSet.resource();
         Object value = valueSet.predicate().equalsValue().get();
-        ResourceIndexFragmentSet indexFragmentSet = new ResourceIndexFragmentSet(resource, typeLabel, value);
+
+        // Use isa property as the property of the fragment
+        VarProperty varProperty = isaSet.fragments().iterator().next().getVarProperty();
+        ResourceIndexFragmentSet indexFragmentSet = new ResourceIndexFragmentSet(varProperty, resource, typeLabel, value);
```
aelred: @duckofyork is it important that we do this? Can we provide `null` instead?

duckofyork: kasper said yes

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ResolutionPlan.java
```diff
@@ -0,0 +1,214 @@
...
+                .filter(Objects::nonNull)
+                .filter(properties::contains)
+                .distinct()
+                .map(propertyMap::get)
+                .collect(Collectors.toCollection(LinkedList::new));
+
+        Set<Atom> nonResolvableAtoms = new HashSet<>();
+        while (!atoms.isEmpty()) {
+            Atom top = atoms.getFirst();
+            atoms.remove(top);
```
aelred: `atoms.remove()` will actually return the first element, so this can be:
```
Atom top = atoms.remove();
```

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ResolutionPlan.java
```diff
@@ -0,0 +1,214 @@
...
+            top = top.getNeighbours().filter(atoms::contains)
+                    .flatMap(at -> Stream.concat(Stream.of(at), at.getNeighbours().filter(atoms::contains)))
+                    .sorted(Comparator.comparing(at -> -at.computePriority(subbedVars)))
+                    .findFirst().orElse(null);
+
+            //top is disconnected atom
+            if (top == null) {
+                top = atoms.stream()
+                        .sorted(Comparator.comparing(at -> -at.computePriority(subbedVars)))
+                        .findFirst().orElse(null);
```
aelred: I'd call `get()` instead of `orElse(null)` in this case. If there's nothing present, we actually want to crash immediately (since we try and access `top` in the next iteration).

IntelliJ gives a warning about using `get()` but that's just because it has a bad name :P

PR #1837, opened 2017-07-04
Refactor lhs and rhs to when and then

fppt:


grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/ThenProperty.java
```diff
@@ -22,24 +22,24 @@
...
 
 /**
- * Represents the {@code rhs} (right-hand side) property on a {@link ai.grakn.concept.Rule}.
+ * Represents the {@code then} (right-hand side) property on a {@link ai.grakn.concept.Rule}.
  *
  * This property can be inserted and not queried.
  *
- * The right-hand side describes the right-hand of an implication, stating that when the left-hand side of a rule is
- * true the right-hand side must hold.
+ * The then side describes the right-hand of an implication, stating that when the when side of a rule is
```
aelred: > the then ... when the when ... the then must hold

it's "thingy" all over again!

fppt: hahahaha

PR #1843, opened 2017-07-05
Fix genealogy example

aelred:


fppt:
@aelred merge this into master please and reference it here.

aelred:
Merged here: #1846

fppt:
⭐️

PR #1844, opened 2017-07-05
Introducing Reified Relations as part of the work to allowing resource edges and automatic reification

fppt:


fppt:
@aelred before this gets any bigger. I think it's "good enough" to go in.

grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptImpl.java
```diff
@@ -146,10 +142,18 @@ public ConceptId getId(){
...
-     * @return The hash code of the underlying vertex
+     * @param concept The interface to be casted
+     * @param <X> The type of the implementation to be returned
+     * @return The concept casted to it's implementation
      */
-    public int hashCode() {
+    @SuppressWarnings("unchecked")
+    public static <X extends ConceptImpl> X from(Concept concept){
+        return (X) concept;
+    }
```
aelred: I'm gonna dump my thoughts about this thing here:

Ideally I'd suggest making sure these cast methods are always safe. If you want to perform a potentially-unsafe cast then you should do it with a manual cast.

We can make them safe if we remove the generic and we only put them in places where we know there is just one valid implementation. For example `ResourceImpl.from(Resource)`, `ConceptVertex.from(Concept)` etc.

grakn-graph/src/main/java/ai/grakn/graph/internal/GraphCache.java
```diff
@@ -97,7 +97,8 @@ void readTxCache(TxCache txCache){
         cachedTypes.putAll(txCache.getOntologyConceptCache());
 
         //Flush All The Internal Transaction Caches
-        txCache.getOntologyConceptCache().values().forEach(OntologyConceptImpl::txCacheFlush);
+        txCache.getOntologyConceptCache().values().forEach(concept
+                -> ConceptImpl.<OntologyConceptImpl>from(concept).txCacheFlush());
```
aelred: This would be clearer with `OntologyConceptImpl.from(ontologyConcept)`

grakn-graph/src/main/java/ai/grakn/graph/internal/OntologyConceptImpl.java
```diff
@@ -280,7 +280,7 @@ public T sup(T newSuperType) {
 
             //Modify the graph once we have checked no loop occurs
             deleteEdge(Direction.OUT, Schema.EdgeLabel.SUB);
-            putEdge(newSuperType, Schema.EdgeLabel.SUB);
+            putEdge((ConceptVertex) newSuperType, Schema.EdgeLabel.SUB);
```
aelred: Another use case for `ConceptVertex.from`!

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
```diff
@@ -51,90 +46,89 @@
...
-            roleMap.get(role).forEach(instance -> {
-                if(instance != null){
-                    hash.append("_").append(instance.getId().getValue().replace("_", "\\_"));
-                }
-            });
-        }
-        return hash.toString();
+    private <X> Collection<X> readFromReified(Function<RelationReified, Collection<X>> producer){
+        if(reified().isPresent()) return producer.apply(reified().get());
+        return Collections.emptyList();
```
aelred: Dull optional shit: `return reified().map(producer).orElseGet(Collections::emptyList);`

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
```diff
@@ -51,90 +46,89 @@
...
+        throw new UnsupportedOperationException("Reification is not yet supported");
+    }
+
+    RelationStructure structure(){
+        return relationStructure;
+    }
+
+    @Override
+    public Relation resource(Resource resource) {
+        reified().orElseGet(this::reify).resource(resource);
```
aelred: I think `reify().resource(resource);` will do the same right?

fppt: good spot. yes you are correct

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
fppt: I like one liners :)

grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptImpl.java
```diff
@@ -193,6 +181,16 @@ public int compareTo(Concept o) {
...
 
+    /**
+     * Helper method to cast interfaces into their implementation.
+     *
+     * @param concept The interface to be casted
+     * @return The concept casted to it's implementation
+     */
+    public static ConceptImpl from(Concept concept){
+        return (ConceptImpl) concept;
+    }
```
aelred: This one isn't always safe. I'd consider removing it and just casting. Unsafe example:
```
RelationImpl rel = thingy();
ConceptImpl.from(rel); // crash
```

grakn-graph/src/main/java/ai/grakn/graph/internal/ThingImpl.java
```diff
@@ -251,19 +251,21 @@ protected T type(TypeImpl type) {
...
      * @return The id of the type of this concept. This is a shortcut used to prevent traversals.
      */
-    public Label getInternalType(){
+    Label getInternalType(){
         return cachedInternalType.get();
     }
+
+    public static ThingImpl<?, ?> from(Thing thing){
+        return (ThingImpl) thing;
+    }
```
aelred: This one isn't always safe. I'd consider removing it and just casting. Unsafe example:
```
RelationImpl rel = thingy();
ThingImpl.from(rel); // crash
```

grakn-graph/src/main/java/ai/grakn/graph/internal/ConceptImpl.java
fppt: That is a good point actually. It's not idealogical. I't practical. I have gotten these crashes.

PR #1849, opened 2017-07-06
Docker profile should not skip tests outside of grakn-test.

burukuru:
We are moving tests back into their own modules.


burukuru:
@fppt @aelred do you know what I can do about those tests I've re-enabled?

burukuru:
!rtg

burukuru:
!rtg

PR #1852, opened 2017-07-06
Fix performance issue when specifying meta role

aelred:


kasper-piskorski:
shouldn't this go into stable?

aelred:
@kasper-piskorski you're right! I'll move it.

aelred:
On stable here: #1858

PR #1857, opened 2017-07-07
Resource Edges and Automatic Reification

fppt:


fppt:
Current Negative: `relationType.instances()` will only return reified relations. Still trying to think of a way around this issue without adding a HUGE performance penalty

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
```diff
@@ -63,15 +63,16 @@
...
 
     /**
      * Reifys and returns the {@link RelationReified}
      */
     RelationReified reify(){
-        if(relationStructure.isReified()) return relationStructure.reified();
-        throw new UnsupportedOperationException("Reification is not yet supported");
+        if(relationStructure.isReified()) return relationStructure.reify();
+        relationStructure = relationStructure.reify();
+        return relationStructure.reify();
```
aelred: Last line should just be `return relationStructure;`

We can also just remove the check in the first line, but I guess it skips an assignment? I don't know if it's worth the cost of the extra function call though.

grakn-graph/src/main/java/ai/grakn/graph/internal/RoleImpl.java
```diff
@@ -130,10 +127,11 @@ void deleteCachedDirectPlaysByType(Type oldType){
      * @return Get all the roleplayers of this role type
      */
     public Stream<Casting> rolePlayers(){
-        //TODO: This should not call reify() it should call reified() but we need a way of getting the roleplayers from relation edges which are not linked to the relation type
         return relationTypes().stream().
                 flatMap(relationType -> relationType.instances().stream()).
-                flatMap(relation -> ((RelationImpl)relation).reify().castingsRelation(this));
+                map(relation -> RelationImpl.from(relation).reified()).
+                filter(Optional::isPresent).
```
aelred: A perfect case for `flatMap(CommonUtil::optionalToStream)`!

grakn-graph/src/main/java/ai/grakn/graph/internal/Shard.java
```diff
@@ -55,10 +55,10 @@ VertexElement vertex(){
 
     /**
      *
-     * @return The id of this shard. String ares used because shards are looked up via the string index.
+     * @return The id of this shard. String are used because shards are looked up via the string index.
```
aelred: lol

also it's "Strings" :P

grakn-graph/src/main/java/ai/grakn/graph/internal/EdgeElement.java
```diff
@@ -35,9 +35,10 @@
  * @author fppt
  */
 class EdgeElement extends AbstractElement<Edge, Schema.EdgeProperty> {
+    static String PREFIX = "E";
```
aelred: We don't need this. We have `Schema.PREFIX_EDGE` 👍

grakn-graql/src/test/java/ai/grakn/graql/internal/query/match/MatchQueryTest.java
```diff
@@ -848,6 +848,9 @@ public void testQueryDoesNotCrash() {
     @Test
     public void testMatchHas() {
         MatchQuery query = qb.match(x.has("name"));
+
+        Role thing = movieGraph.graph().getRole("has-name-owner");
```
aelred: Accidental commit?

grakn-core/src/main/java/ai/grakn/concept/LabelId.java
```diff
@@ -73,7 +73,7 @@ public String toString(){
     }
 
     public boolean isValid(){
-        return typeId != -1;
+        return typeId != null && typeId != -1;
```
fppt: Just going to make the invalid be -1

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -363,10 +363,15 @@ void checkMutationAllowed(){
 
     //----------------------------------------------Concept Functionality-----------------------------------------------
     //------------------------------------ Construction
-    @Nullable
```
fppt: Ugh for some reason i read it as it can accept nulls. Which it can't. Will correct

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -363,10 +363,15 @@ void checkMutationAllowed(){
...
     VertexElement addVertex(Schema.BaseType baseType){
         Vertex vertex = operateOnOpenGraph(() -> getTinkerPopGraph().addVertex(baseType.name()));
-        vertex.property(Schema.VertexProperty.ID.name(), vertex.id().toString());
+        vertex.property(Schema.VertexProperty.ID.name(), Schema.PREFIX_VERTEX + vertex.id().toString());
+        return factory().buildVertexElement(vertex);
+    }
```
fppt: I tried but the problem is that the one case required the vertex to exist. You basically end up making a lot of really ugly changes just to avoid 2 similar lines.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -649,16 +654,17 @@ public RuleType getMetaRuleConstraint() {
     }
 
     void putShortcutEdge(Thing toThing, Relation fromRelation, Role roleType){
-        boolean exists  = getTinkerPopGraph().traversal().V(fromRelation.getId().getRawValue()).
+        //TODO: hasId() may be faster than has(ID, ...)
```
fppt: later. need to bechmark it

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractElement.java
```diff
@@ -125,6 +132,47 @@ public boolean equals(Object object) {
...
+
+    /**
+     * Sets a property which cannot be mutated
+     *
+     * @param property The key of the immutable property to mutate
+     * @param newValue The new value to put on the property (if the property is not set)
+     * @param foundValue The current value of the property
+     * @param converter Helper method to ensure data is persisted in the correct format
+     */
+    <X> void propertyImmutable(P property, X newValue, X foundValue, Function<X, Object> converter){
```
fppt: It makes it easier to handle comparing the properties using one type and persisting them using another. For example with resources of type date they are persisted as longs but compared as time stamps.

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationImpl.java
fppt: Can't just `return relationStructure` because it's not the correct return type. I think I can clean that method up though.

grakn-graph/src/main/java/ai/grakn/graph/internal/RoleImpl.java
fppt: I was initally excited about this but it's not as nice as I thought it would be :( just avoids the get call which is something I suppose

grakn-graph/src/main/java/ai/grakn/graph/internal/Shard.java
fppt: Spelling hard

grakn-graph/src/main/java/ai/grakn/graph/internal/EdgeElement.java
fppt: Left over crap which needs to die. Good eye

grakn-graql/src/test/java/ai/grakn/graql/internal/query/match/MatchQueryTest.java
fppt: yep. killing it now.

fppt:
@aelred maybe we can review the code today and try to get this in by end of day? Performance wise we not gaining what I expected BUT space wise we saving 30% so I think this is "good enough" to go in.

grakn-core/src/main/java/ai/grakn/util/CommonUtil.java
```diff
@@ -70,6 +73,20 @@ public static void withImplicitConceptsVisible(GraknGraph graph, Runnable functi
...
+     * @param implicitType the implicit type label
+     * @return The original label which was used to build this type
+     */
+    public static Label explicitLabel(Label implicitType){
+        if(!implicitType.getValue().startsWith("key") && implicitType.getValue().startsWith("has")){
+            throw new IllegalArgumentException(INVALID_IMPLICIT_TYPE.getMessage(implicitType));
+        }
+        int endIndex = implicitType.getValue().lastIndexOf("-");
+        return Label.of(implicitType.getValue().substring(4, endIndex));
+    }
```
aelred: We should move this method to `Schema.ImplicitType` (ideally one for each kind, which will make it simpler).

grakn-core/src/main/java/ai/grakn/util/CommonUtil.java
```diff
@@ -70,6 +73,20 @@ public static void withImplicitConceptsVisible(GraknGraph graph, Runnable functi
...
+     * Helper method which converts the implicit type label back into the original label from which is was built.
+     *
+     * @param implicitType the implicit type label
+     * @return The original label which was used to build this type
+     */
+    public static Label explicitLabel(Label implicitType){
+        if(!implicitType.getValue().startsWith("key") && implicitType.getValue().startsWith("has")){
+            throw new IllegalArgumentException(INVALID_IMPLICIT_TYPE.getMessage(implicitType));
+        }
+        int endIndex = implicitType.getValue().lastIndexOf("-");
```
aelred: This won't work in the case of implicit relations of resources with dashes like `has-first-name`. Result will be `first`.

grakn-factory/titan-factory/src/main/resources/indices-edges.properties
```diff
@@ -16,4 +16,5 @@
 # along with Grakn. If not, see <http://www.gnu.org/licenses/gpl.txt>.
 #
 
-shortcut=RELATION_TYPE_ID,ROLE_TYPE_ID
+shortcut=RELATION_TYPE_LABEL_ID,ROLE_LABEL_ID
+resource=RELATION_TYPE_LABEL_ID
```
aelred: Should we also put the `ROLE` properties for `resource` here?

grakn-graph/src/main/java/ai/grakn/graph/internal/ResourceImpl.java
```diff
@@ -71,9 +73,15 @@
      */
     @Override
     public Collection<Thing> ownerInstances() {
-        return getShortcutNeighbours().stream().
+        //Get Owner via implicit structure
+        Set<Thing> owners = getShortcutNeighbours().stream().
```
aelred: If `getShortcutNeighbours` also followed relation edges, you might be able to simplify this and some other cases.

grakn-graph/src/test/java/ai/grakn/graph/internal/RelationTest.java
```diff
@@ -272,7 +274,20 @@ public void whenDeletingFinalInstanceOfRelation_RelationIsDeleted(){
...
         relationType.addRelation().addRolePlayer(null, rolePlayer1);
     }
+
+    @Test
+    public void whenAttemptingToLinkTheInstanceOfAResourceRelationToTheResourceWhichCreatedIt_ThrowIfTheRelationTypeDoesNotHavePermissionToPlayTheNecessaryRole(){
+        ResourceType<String> resourceType = graknGraph.putResourceType("what a pain", ResourceType.DataType.STRING);
+        Resource<String> resource = resourceType.putResource("a real pain");
+
+        EntityType entityType = graknGraph.putEntityType("yay").resource(resourceType);
+        Relation implicitRelation = Iterators.getOnlyElement(entityType.addEntity().resource(resource).relations().iterator());
```
aelred: Boring but: this could use `Iterables#getOnlyElement` instead.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/InPlaysFragment.java
```diff
@@ -44,14 +45,14 @@
...
-    public void applyTraversal(GraphTraversal<Vertex, Vertex> traversal, GraknGraph graph) {
+    public void applyTraversal(GraphTraversal<? extends Element, ? extends Element> traversal, GraknGraph graph) {
         if (required) {
             traversal.inE(PLAYS.getLabel()).has(Schema.EdgeProperty.REQUIRED.name()).otherV();
         } else {
             traversal.in(PLAYS.getLabel());
         }
 
-        Fragments.inSubs(traversal);
+        Fragments.inSubs((GraphTraversal<Vertex, Vertex>) traversal);
```
aelred: It's possible this is bad. Let me check.

grakn-test-tools/src/main/java/ai/grakn/generator/RelationsFromRolePlayers.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * A generator that produces {@link Relation}s from existing role-players.
+ *
+ * This means the relation is navigated to from another {@link Thing} attached to it. This will find relations even
+ * if they are not returned by {@link RelationType#instances}.
+ *
+ * @author Felix Chapman
+ */
+public class RelationsFromRolePlayers extends FromGraphGenerator<Relation> {
```
aelred: We can delete this generator now. It only existed because `RelationType#instances` did not return all types.

On the other hand, keeping it makes our tests a bit more robust if `instances` breaks again.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/InPlaysFragment.java
aelred: I can now confirm this is bad. Will fix.

grakn-factory/titan-factory/src/main/resources/indices-edges.properties
fppt: No. We don't need those to be indexed

grakn-core/src/main/java/ai/grakn/util/CommonUtil.java
fppt: Yes you right. Will fix

grakn-core/src/main/java/ai/grakn/util/CommonUtil.java
fppt: I actually did that initially but was not a fan. Will change it back though because I am not too fussed.

grakn-graph/src/main/java/ai/grakn/graph/internal/ResourceImpl.java
fppt: Good suggestion. Will think about doing that in a future cleanup PR. May break stuff all over again.

grakn-graph/src/test/java/ai/grakn/graph/internal/RelationTest.java
fppt: `Iterables` thats the one!

grakn-test-tools/src/main/java/ai/grakn/generator/RelationsFromRolePlayers.java
fppt: Keep it for the moment

grakn-core/src/main/java/ai/grakn/util/CommonUtil.java
fppt: Just realised why this didn't break. It was designed for converting the implicit role labels only.

grakn-graph/src/main/java/ai/grakn/graph/internal/ElementFactory.java
```diff
@@ -185,7 +203,46 @@ RoleImpl buildRole(VertexElement vertex, Role type, Boolean isImplicit){
...
+     * @param edge A {@link Edge} of an unknown type
+     * @return A concept built to the correct type
+     */
+    @Nullable
+    <X extends Concept> X buildConcept(Edge edge){
+        return buildConcept(buildEdgeElement(edge));
+    }
+
+    @Nullable
+    <X extends Concept> X buildConcept(EdgeElement edgeElement){
```
fppt: This has a lot of duplicate code with it's brother `buildConcept(VertexElement)` need to clean that up in another PR

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationEdge.java
```diff
@@ -0,0 +1,162 @@
...
+    private final Cache<Thing> value = new Cache<>(() -> edge().graph().factory().buildConcept(edge().target()));
+
+    RelationEdge(EdgeElement edgeElement) {
+        this.edgeElement = edgeElement;
+    }
+
+    RelationEdge(RelationType relationType, Role ownerRole, Role valueRole, EdgeElement edgeElement) {
+        this(edgeElement);
+
+        edgeElement.propertyImmutable(Schema.EdgeProperty.RELATION_ROLE_OWNER_LABEL_ID, ownerRole, null, o -> o.getLabelId().getValue());
```
fppt: Possible to get rid of `RELATION_ROLE_OWNER_LABEL_ID` and `RELATION_ROLE_VALUE_LABEL_ID` as this information is already stored in it's type. For another PR though.

grakn-graph/src/main/java/ai/grakn/graph/internal/RelationTypeImpl.java
```diff
@@ -155,8 +155,38 @@ public void delete(){
...
+
+        if(isImplicit()){
+            //If the relation type is implicit then we need to get any relation edges it may have.
+            //Unfortunately this is a slow process
+
+            relates().stream().
+                    flatMap(role -> role.playedByTypes().stream()).
+                    forEach(type ->{
+                        //Traversal is used here to take advantage of vertex centric index
+                        vertex().graph().getTinkerTraversal().V().
```
fppt: I am really not happy with this one. We need elastic search or a more clever way to handle these situations.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/IdFragment.java
```diff
@@ -38,8 +40,24 @@
...
+                    edgeTraversal(),
+                    vertexTraversal(__.identity())
+            );
+        } else {
+            vertexTraversal(traversal);
+        }
+    }
+
+    private <S, E> GraphTraversal<S, E> vertexTraversal(GraphTraversal<S, E> traversal) {
+        return traversal.has(Schema.VertexProperty.ID.name(), id.getValue());
```
fppt: Maybe worth noting here why the separation is needed. Why can't vertices and edges be looked up by `hasId(xxxxx)` ?

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/IdFragment.java
```diff
@@ -75,4 +93,9 @@ public double fragmentCost() {
     public boolean hasFixedFragmentCost() {
         return true;
     }
+
+    @Override
+    public boolean operatesOnEdge() {
+        return id.getValue().startsWith(Schema.PREFIX_EDGE);
```
fppt: I don't think this is a good check. `RelationReified` id's start with `Schema.PREFIX_EDGE`. This will be true in most cases but not all.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/InIsaFragment.java
```diff
@@ -41,8 +53,46 @@
...
+                ),
+                toVertexInstances(__.identity())
+        );
+    }
+
+    private <T> GraphTraversal<T, Vertex> toVertexInstances(GraphTraversal<T, Vertex> traversal) {
+        return traversal.in(SHARD.getLabel()).in(ISA.getLabel());
+    }
+
+    private GraphTraversal<?, Edge> toEdgeInstances() {
```
fppt: This looks oddly familiar :)

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/OutIsaFragment.java
```diff
@@ -60,4 +71,9 @@ public double fragmentCost() {
                                                               Map<Node, Map<Node, Fragment>> edges) {
         return getDirectedEdges(NodeId.NodeType.ISA, nodes, edges);
     }
+
+    @Override
+    public boolean operatesOnEdge() {
+        return true;
```
fppt: Why always `true`?

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/OutShortcutFragment.java
```diff
@@ -123,4 +157,9 @@ public int hashCode() {
         result = 31 * result + relationTypeLabels.hashCode();
         return result;
     }
+
+    @Override
+    public boolean operatesOnEdge() {
+        return true;
```
fppt: I sense having this be defaulted to true may come back to bite us. I would rather have `throw UnsupportedOperation` or something like that.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/IdFragment.java
aelred: In this case, we're doing the right thing. `operatesOnEdge` just indicates "this fragment may operate on edges", not _only_ edges. Tell me if you can think of a way to make that clearer though!

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/OutIsaFragment.java
aelred: It's always possible for `OutIsaFragment` to start from an edge. We can't tell in advance, so when we build the traversal we must assume we could be starting from an edge.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/OutShortcutFragment.java
aelred: Sorry, don't understand. I think I just named this badly!

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/IdFragment.java
fppt: Maybe `canOperateOnEdge()` ?

fppt: Or just javadoc it

aelred: It is javadoc'd in the interface 👍 
Will rename, though. I think the `can` makes it clearer.

PR #1858, opened 2017-07-10
Fix performance issue when specifying meta role

aelred:


aelred:
!rtg - cassandra timeouts

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/OutShortcutFragment.java
```diff
@@ -98,6 +98,7 @@ public boolean equals(Object o) {
         OutShortcutFragment that = (OutShortcutFragment) o;
 
         if (!edge.equals(that.edge)) return false;
+        if (!roleType.equals(that.roleType)) return false;
```
fppt: I only point this out because of the above refactors but `role` not `roleType`

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/OutShortcutFragment.java
```diff
@@ -106,6 +107,7 @@ public boolean equals(Object o) {
     public int hashCode() {
         int result = super.hashCode();
         result = 31 * result + edge.hashCode();
+        result = 31 * result + roleType.hashCode();
```
fppt: `role`

fppt:
https://github.com/graknlabs/grakn/pull/1872

PR #1862, opened 2017-07-11
Quickcheck refactors

aelred:


grakn-core/src/main/java/ai/grakn/graql/Pattern.java
```diff
@@ -42,4 +42,16 @@
      */
     @CheckReturnValue
     PatternAdmin admin();
+
+    /**
+     * Join patterns in a conjunction
+     */
+    @CheckReturnValue
+    Pattern and(Pattern pattern);
```
fppt: Why did quickcheck refactors need these methods to be added to the `Pattern` interface?

grakn-core/src/main/java/ai/grakn/graql/admin/RelationPlayer.java
```diff
@@ -26,23 +26,71 @@
  *
  * @author Felix Chapman
  */
-public interface RelationPlayer {
+public class RelationPlayer {
+    private final int hashCode;
+    private final Optional<VarPatternAdmin> roleType;
```
fppt: `role`

grakn-core/src/main/java/ai/grakn/graql/admin/RelationPlayer.java
```diff
@@ -26,23 +26,71 @@
...
+    public static RelationPlayer of(VarPatternAdmin roleType, VarPatternAdmin rolePlayer) {
+        return new RelationPlayer(Optional.of(roleType), rolePlayer);
+    }
+
+    /**
+     * @return the role type, if specified
      */
     @CheckReturnValue
-    VarPatternAdmin getRolePlayer();
+    public Optional<VarPatternAdmin> getRoleType() {
```
fppt: `getRole`

grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/AbstractPattern.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package ai.grakn.graql.internal.pattern;
+
+import ai.grakn.graql.Graql;
+import ai.grakn.graql.Pattern;
+import ai.grakn.graql.admin.PatternAdmin;
+
+/**
+ * @author Felix Chapman
```
fppt: Potential to improve java docs

grakn-test/src/test/java/ai/grakn/generator/ConceptIds.java
```diff
@@ -22,7 +22,7 @@
 import ai.grakn.concept.ConceptId;
 
 /**
- * Generator that generates totally concept IDs
+ * Generator that generates totally random concept IDs
  */
 public class ConceptIds extends AbstractGenerator<ConceptId> {
```
fppt: Can we not move all these generator classes to `grakn-test-tools` instead? By moving them to `grakn-test` it means that they can only be used in that module (very bad in my opinion), but by moving them to `grakn-test-tools` they can be used in any module (except `grakn-graph`).

I just feel that moving the generators to `grakn-test` is a step towards making that a monolithic test module again.

grakn-core/src/main/java/ai/grakn/graql/Pattern.java
aelred: This is a helpful sugar method that proved useful for writing the Quickcheck properties (and useful for normal users too). It allows you to write `and` and `or` like actual binary operators 

e.g. we have tests stating `match(patternA.and(patternB)) == match(patternB.and(patternA))`

grakn-core/src/main/java/ai/grakn/graql/admin/RelationPlayer.java
aelred: Thanks. This change will plague us for a while I think :P

grakn-test/src/test/java/ai/grakn/generator/ConceptIds.java
aelred: I agree on the monolith thing and I'll move them back to `grakn-test-tools`.

Here's the layout:
- Generators for `grakn-graph` will go in `grakn-test-tools`
- Property tests for `grakn-graph` will go in `grakn-test`
- Generators and property tests for `grakn-graql` will go in `grakn-graql`

I'm not super-happy with this - the layout is scattered and inconsistent. We'll also have issues if we ever need to write say, tests for `grakn-engine` that generate queries. On the other hand I have no better ideas... except Gradle of course 👍

fppt: I agree that it's not ideal but at least with this approach we not forcing *all* the generator tests into the same module. The problem is the Graph API generator tests being in `grakn-test`. We need to resolve that.

aelred: I don't think we can 😞 
The`grakn-graph` property tests must be in a module that depends on `grakn-test-tools` (because that's where the generators are).

aelred:
@fppt time for re-review!

grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/AbstractPattern.java
```diff
@@ -24,6 +24,8 @@
 import ai.grakn.graql.admin.PatternAdmin;
 
 /**
+ * Abstract implementation of {@link PatternAdmin}.
```
fppt: I think that is given away in the class header. Maybe state what this abstract class is used for. What common functionality it houses. Etc. . .

PR #1865, opened 2017-07-11
Squashed Redis store changes

pluraliseseverythings:
Deleted Kafka/Zookeeper references
Using Redisson over Redis for distributed locks
Using Jesque for Task queues
Added Jackson dependency for serialising task in queue


pluraliseseverythings:
!rtg

PR #1868, opened 2017-07-12
Check whether rules are valid Horn clauses on commit

kasper-piskorski:


grakn-graph/src/main/java/ai/grakn/graph/internal/ValidateGlobalRules.java
```diff
@@ -275,6 +279,35 @@ private ValidateGlobalRules() {
         return Optional.empty();
     }
 
+    static Set<String> validateRuleIsHornClause(GraknGraph graph, Rule rule){
```
fppt: Javadoc what this rule actually checks. I.e. state what a horn clause it and how you check the rule is one. Validtor gets confusing fast without very clear javadocs.

grakn-graph/src/main/java/ai/grakn/graph/internal/ValidateGlobalRules.java
```diff
@@ -275,6 +279,35 @@ private ValidateGlobalRules() {
...
+        Set<String> errors = new HashSet<>();
+        if (rule.getWhen().admin().isDisjunction()){
+            errors.add(ErrorMessage.VALIDATION_RULE_DISJUNCTION_IN_BODY.getMessage(rule.getId(), rule.type().getLabel()));
+        }
+        errors.addAll(checkRuleHeadInvalid(graph, rule, rule.getThen()));
+        return errors;
+    }
+    
+    private static Set<String> checkRuleHeadInvalid(GraknGraph graph, Rule rule, Pattern head) {
```
fppt: Ditto for here. Please be very descriptive in your javadocs.

PR #1869, opened 2017-07-12
Change Labels On Ontology Concepts

fppt:


fppt:
!rtg

grakn-core/src/main/java/ai/grakn/concept/OntologyConcept.java
```diff
@@ -38,6 +38,14 @@
  * @author fppt
  */
 public interface OntologyConcept extends Concept {
+    //------------------------------------- Modifiers ----------------------------------
+    /**
+     * Changes the {@link Label} of this {@link Concept} to a new one.
+     * @param label The new {@link Label}.
+     * @return The {@link Concept} itself
+     */
+    OntologyConcept setLabel(Label label);
```
aelred: We need a description and `@throw` clause of what happens if the label in question is already in use, plus a test for it.

fppt: I have added the test. But I thought we agreed that trying to document runtime exceptions was just not maintainable in the long run?

aelred: In that case, a test is fine!

PR #1873, opened 2017-07-13
Added CAS to inflight task processing. 

pluraliseseverythings:
Now it only restore a task from a dead letter queue if nothing modified the queue between read and write. Also sped up failover test

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisInflightTaskConsumer.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisInflightTaskConsumer.java
@@ -34,6 +34,7 @@ import static net.greghaines.jesque.utils.ResqueConstants.QUEUE;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import redis.clients.jedis.Jedis;
+import redis.clients.jedis.Transaction;
 import redis.clients.util.Pool;
 
 /**
@@ -42,10 +43,11 @@ import redis.clients.util.Pool;
  * @author Domenico Corapi
  */
 public class RedisInflightTaskConsumer extends TimerTask {
-    private final static Logger LOG = LoggerFactory.getLogger(RedisInflightTaskConsumer.class);
-    private final static ObjectMapper objectMapper = new ObjectMapper();
+    private static final Logger LOG = LoggerFactory.getLogger(RedisInflightTaskConsumer.class);
+    private static final ObjectMapper objectMapper = new ObjectMapper();
 
-    public static final int END = 10;
+    static final int ITERATIONS = 10;
+
     private Pool<Jedis> jedisPool;
     private Duration processInterval;
     private Config config;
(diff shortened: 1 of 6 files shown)


pluraliseseverythings:
- There is a process executed periodically that checks if there are some dead jobs. This PR fixes a race condition in that
- It also makes the retry delay smaller for tests. In particular there's one test that kills engines while running. In some cases they might leave dead tasks. Previously they would be picked after 10 minutes, now it's 5 seconds

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisInflightTaskConsumer.java
```diff
@@ -42,10 +43,11 @@
  * @author Domenico Corapi
  */
 public class RedisInflightTaskConsumer extends TimerTask {
-    private final static Logger LOG = LoggerFactory.getLogger(RedisInflightTaskConsumer.class);
-    private final static ObjectMapper objectMapper = new ObjectMapper();
+    private static final Logger LOG = LoggerFactory.getLogger(RedisInflightTaskConsumer.class);
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    static final int ITERATIONS = 10;
```
fppt: Should this be configurable?

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisInflightTaskConsumer.java
```diff
@@ -65,29 +67,51 @@ public void run() {
...
         }
     }
+
+    private void processElement(String key, Jedis resource, String head) {
+        // TODO Use Jackson for this
+        long runAt = Json.read(head).at("args").at(0).at("taskState")
+                .at("schedule").at("runAt").asLong();
+        Instant runAtDate = Instant.ofEpochMilli(runAt);
+        Duration gap = Duration.between(runAtDate, Instant.now());
+        if (gap.getSeconds() > processInterval.getSeconds()) {
```
fppt: How is `processInterval` ever set?

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisTaskQueue.java
```diff
@@ -105,25 +113,36 @@ void putJob(Task job) {
...
-            GraknEngineConfig config,
-            EngineGraknGraphFactory factory) {
+            GraknEngineConfig engineConfig,
+            EngineGraknGraphFactory factory,
+            int poolSize) {
         LOG.info("Subscribing worker to jobs in queue {}", QUEUE_NAME);
-        Worker worker = getJobSubscriber();
+        // sync to avoid close while starting
+        synchronized(this) {
+            this.workerPool = new WorkerPool(() -> getWorker(redisTaskManager, engineId, engineConfig, factory), poolSize);
```
fppt: Maybe in the future it's worth creating a central handles for all the thread related pools we make? I imagine us creating quite a lot.

pluraliseseverythings: Yes ideally we should make the thread pools in the main class and pass them over (even better, use circuit breakers for these things).

This thing is not even a thread pool, it's a special set of threads handled by Jesque. It doesn't spin up threads, it starts all the workers at start time and it provides some management methods.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisInflightTaskConsumer.java
pluraliseseverythings: It's set in config under 
`tasks.retry.delay`

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisInflightTaskConsumer.java
pluraliseseverythings: Not 100% sure, it's quite low level. I'll leave it there for now but if I end up changing it once I'll move it to conf

PR #1874, opened 2017-07-13
[Stable] To Miko with love (aka various small performance fixes)

kasper-piskorski:


kasper-piskorski:
@mikonapoli  please review, it should fix the onshore problem :)

kasper-piskorski:
Yes it is arbitrary, in fact there's a number of arbitrary parameters. Until we use some graph statistics I think we have to stick to using them.

kasper-piskorski:
Regarding tests, we don't have any explicit tests testing resolution order. They are all tested on a query basis.

fppt:
@kasper-piskorski merge this into master please

PR #1875, opened 2017-07-13
Janus Graph Support

fppt:
Merging this will result in fire and brimstone falling from the sky, dogs and cats getting along, all dental floss becoming razor wire, all movies to be increasing versions of Sharknado, and many other horrible unspeakable things will happen.

DO NOT MERGE!!!!


fppt:
I think I have done all I could on this one. Pausing until @duckofyork gets a chance to bring analytics back from the dead

mikonapoli:
> all movies to be increasing versions of Sharknado

This is possibly the best thing I have read in a while

burukuru:
!rtg - just testing Jenkins

fppt:
@duckofyork I have had to fix a few PMD errors. The only really controversial one was the suppression I had to do. Just noting it here. Maybe it's not necessary but it looks like it was.

fppt:
@aelred @duckofyork this should be "reviewable" now if you guys can get the time to look at it. I have a secret hope of getting this in before end of day.

duckofyork:
This one is huge :-)

fppt:
I did not say it would be easy to review :D

grakn-factory/janus-factory/pom.xml
```diff
@@ -276,6 +276,19 @@
...
+            <groupId>com.google.cloud.bigtable</groupId>
+            <artifactId>bigtable-hbase-1.0</artifactId>
+            <version>${google-bigtable.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-tcnative-boringssl-static</artifactId>
+            <version>${netty.version}</version>
+        </dependency>
```
aelred: What are these two dependencies for?

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
```diff
@@ -774,9 +770,9 @@ public void commit() throws InvalidGraphException{
         return logs;
     }
 
-    private void closeTransaction(String closedReason){
+    private void closeTransaction(String closedReason) {
         try {
-            // TODO: We check `isOpen` because of a Titan bug which decrements the transaction counter even if the transaction is closed
+            // TODO: We check `isOpen` because of a Janus bug which decrements the transaction counter even if the transaction is closed
```
aelred: Does this bug still exist? I suppose we should assume it does...

grakn-graph/src/main/java/ai/grakn/graph/internal/computer/GraknComputerImpl.java
```diff
@@ -57,33 +67,41 @@
...
         }
     }
 
     @Override
-    public ComputerResult compute(VertexProgram program, MapReduce... mapReduces) {
+    public ComputerResult compute(Boolean includesShortcut, Set<LabelId> types,
+                                  VertexProgram program, MapReduce... mapReduces) {
         try {
-            graphComputer = getGraphComputer().program(program);
+            if (program != null) graphComputer = getGraphComputer().program(program);
```
aelred: I vote for annotating `program` param as `@Nullable`! Same for methods below as well.

grakn-core/src/main/java/ai/grakn/GraknComputer.java
```diff
@@ -50,19 +52,45 @@
...
      *
+     * @param includesShortcut whether the graph computer should include shortcut edges
+     * @param types            instance types in the subgraph
+     * @param program          the vertex program
+     * @param mapReduce        a list of mapReduce job
+     * @return the result of the computation
+     * @see ComputerResult
+     */
+    @CheckReturnValue
+    ComputerResult compute(Boolean includesShortcut, Set<LabelId> types, VertexProgram program, MapReduce... mapReduce);
```
aelred: All these overloaded methods are confusing, especially since the optional arguments are at the beginning, not the end. We can't move them to the end because of the varargs param though...

I've got two bad ideas: one is to make the `MapReduce` argument a `List`, so we can move the optional args to the end. Or we can make this a builder? I think autovalue can help here too.

Then you can write:
```
computer.builder(program).withShortcuts(true).types(types).mapReduce(reduce).build();
```

But this might be overkill. Any other ideas?

grakn-graph/src/main/java/ai/grakn/graph/internal/computer/GraknComputerImpl.java
```diff
@@ -118,4 +136,22 @@ protected GraphComputer getGraphComputer() {
         return graph.compute(this.graphComputerClass);
     }
 
+    private void applyFilters(Set<LabelId> types, boolean includesShortcut) {
+        if (types == null || types.isEmpty()) return;
```
aelred: In this case, we ignore `includesShortcut`. Is that a bug?
Also tag `types` as `@Nullable`! This is my new vendetta :D

grakn-graph/src/main/java/ai/grakn/graph/internal/computer/GraknSparkComputer.java
```diff
@@ -86,37 +96,24 @@
...
+    private final ExecutorService computerService = Executors.newSingleThreadExecutor(threadFactoryBoss);
 
-    private org.apache.commons.configuration.Configuration apacheConfiguration = null;
-    private Configuration hadoopConfiguration = null;
+    static {
+        TraversalStrategies.GlobalCache.registerStrategies(GraknSparkComputer.class,
+                TraversalStrategies.GlobalCache.getStrategies(GraphComputer.class).clone().addStrategies(
+                        SparkSingleIterationStrategy.instance(),
+                        SparkInterceptorStrategy.instance()));
+    }
```
aelred: @duckofyork is this class still different from the `SparkComputer` we copied?

grakn-graph/src/main/java/ai/grakn/graph/internal/computer/GraknSparkVertexProgramInterceptor.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.apache.spark.api.java.JavaPairRDD;
+import org.apache.tinkerpop.gremlin.hadoop.structure.io.VertexWritable;
+import org.apache.tinkerpop.gremlin.process.computer.VertexProgram;
+import org.apache.tinkerpop.gremlin.process.computer.traversal.strategy.VertexProgramInterceptor;
+
+/**
+ * @param <V> Vertex Program
+ * @author Jason Liu
+ */
```
aelred: We could do with some javadoc to explain why this class exists.

grakn-graph/src/main/java/ai/grakn/graph/internal/computer/interceptor/GraknSparkStarBarrierInterceptor.java
```diff
@@ -0,0 +1,186 @@
...
+import org.apache.tinkerpop.gremlin.util.iterator.IteratorUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.BinaryOperator;
+
+/**
+ * @author Jason Liu
+ * @author Marko A. Rodriguez (http://markorodriguez.com)
+ */
```
aelred: This also needs javadoc to explain why we need it. I guess it was copied from somewhere?

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/ConnectedComponentVertexProgram.java
```diff
@@ -113,11 +104,10 @@ private void update(Vertex vertex, Messenger<String> messenger, Memory memory) {
...
             vertex.property(clusterLabel, max);
-            messenger.sendMessage(messageScopeShortcutIn, max);
-            messenger.sendMessage(messageScopeResourceIn, max);
-            messenger.sendMessage(messageScopeShortcutOut, max);
-            messenger.sendMessage(messageScopeResourceOut, max);
-            memory.and(VOTE_TO_HALT, false);
+            messenger.sendMessage(messageScopeIn, max);
+            messenger.sendMessage(messageScopeOut, max);
+            memory.add(VOTE_TO_HALT, false);
+            memory.getIteration();
```
aelred: Does calling `getIteration` do anything?

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/GraknMapReduce.java
```diff
@@ -46,6 +46,8 @@
 
     private static final String RESOURCE_DATA_TYPE_KEY = "RESOURCE_DATA_TYPE_KEY";
 
+    public static final int RESERVED_TYPE_LABEL_KEY = -10;
```
aelred: we should turn this up to -11

(but seriously what is this? :P )

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractComputeQuery.java
```diff
@@ -117,30 +125,35 @@ void initSubGraph() {
...
+        entityTypes.addAll(metaEntityType.subs());
+        entityTypes.remove(metaEntityType);
+
+        ResourceType<?> metaResourceType = theGraph.admin().getMetaResourceType();
+        resourceTypes.addAll(metaResourceType.subs());
+        resourceTypes.remove(metaResourceType);
+
+        RelationType metaRelationType = theGraph.admin().getMetaRelationType();
+        relationTypes.addAll(metaRelationType.subs());
+        relationTypes.remove(metaRelationType);
```
aelred: If we do this here, doesn't it mean we end up doing a lot of extra work if `subLabels` is not empty?

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/CountQueryImpl.java
```diff
@@ -49,13 +54,38 @@ public Long execute() {
...
-                subLabels.stream().map(graph.get().admin()::convertToId).collect(Collectors.toSet())));
-        Map<Serializable, Long> count = result.memory().get(CountMapReduce.class.getName());
+        Set<LabelId> rolePlayerLabelIds = subTypes.stream()
+                .filter(Concept::isRelationType)
+                .map(relationType -> ((RelationType) relationType).relates())
+                .filter(roles -> roles.size() == 2)
+                .flatMap(roles -> roles.stream().flatMap(role -> role.playedByTypes().stream()))
+                .map(type -> graph.get().admin().convertToId(type.getLabel()))
+                .filter(LabelId::isValid)
+                .collect(Collectors.toSet());
```
aelred: Now we have this same thing in 3 places. @fppt maybe we need some common way to access it? Especially since it's easy to get wrong.

grakn-test/src/test/java/ai/grakn/test/graql/analytics/DegreeTest.java
```diff
@@ -114,7 +113,7 @@ public void testDegrees() throws Exception {
         for (long i = 0L; i < workerNumber; i++) {
             list.add(i);
         }
-        GraknSparkComputer.clear();
+//        GraknSparkComputer.clear();
```
aelred: 🐐

grakn-test/src/test/java/ai/grakn/test/graql/analytics/ShortestPathTest.java
```diff
@@ -140,7 +139,7 @@ public void testShortestPath() throws Exception {
     public void testShortestPathConcurrency() {
         List<String> correctPath;
         addOntologyAndEntities();
-        GraknSparkComputer.clear();
+//        GraknSparkComputer.clear();
```
aelred: 🐐

grakn-test/src/test/java/ai/grakn/test/graql/reasoner/BenchmarkTests.java
```diff
@@ -140,7 +140,7 @@ public void testTransitiveChain()  {
         //results @N = 450 101475   ?     ?       ?     14s
         //results @N = 500 125250         ?       ?     21-26s
 
-        //with materialisation            IC    IC titan
+        //with materialisation            IC    IC janus
```
aelred: Given these are old benchmark results, maybe we should keep it saying `titan`? I'm not totally sure.

grakn-factory/janus-factory/pom.xml
fppt: If you read above  Google bit table. Should be more clear though.

grakn-graph/src/main/java/ai/grakn/graph/internal/AbstractGraknGraph.java
fppt: Yes because it is a feature I patched into Janus/Titan which I have not revisited yet.

grakn-test/src/test/java/ai/grakn/test/graql/reasoner/BenchmarkTests.java
fppt: I think we can remove them actually.

grakn-graph/src/main/java/ai/grakn/graph/internal/computer/GraknSparkComputer.java
duckofyork: Yes. We still wanna make sure multiple compute queries can be executed at the same time.

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/GraknMapReduce.java
duckofyork: Vertex emits type label id, but has-resource edge can not. Instead, a message is sent via the edge, and a vertex property is added. So the resource vertex can emit an extra key value pair, key being this constant. Here, I just wanna choose a number that we would never use as type id. So yes, -11 is also good!

grakn-graph/src/main/java/ai/grakn/graph/internal/computer/GraknComputerImpl.java
duckofyork: For now, if subgraph is not specified, shortcut edges must be included. Other wisethere is no useful edges in the graph for message passing.

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/GraknMapReduce.java
aelred: Thanks, I get it now! We should put a comment in the code.

fppt:
@aelred this is pretty cool. The comment you made on Jason's PR into my branch has appeared here ^

fppt:
!rtg

duckofyork:
yay!

fppt:
yay! Now lets run Jason before people start finding the bugs ;)

PR #1879, opened 2017-07-14
Fix degree query not throwing exceptions when ofType is invalid

duckofyork:


aelred:
Please merge into master 👍

PR #1882, opened 2017-07-14
Re-enable tests for docs

aelred:
These tests compile and execute any Java and Graql code blocks.

I've mae an effort to make the tests permissive, to reduce the number of changes to documentation. However, some docs have been changed because the tests couldn't handle them (e.g. they had a few invalid symbols like `...` to indicate ommitted content).


aelred:
!rtg - odd failure, logged on TP

PR #1886, opened 2017-07-17
Fixed concurrent start problem

pluraliseseverythings:
Previously when starting multiple engines, one of them would start creating the system ontology and the others would fail.
Now they try to acquire a distributed lock if running on Redis or local lock if using the Standalone Queue.

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
@@ -50,6 +50,8 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
 import javax.annotation.Nullable;
 import mjson.Json;
 import static org.apache.commons.lang.exception.ExceptionUtils.getFullStackTrace;
@@ -76,6 +78,7 @@ import spark.Service;
 
 public class GraknEngineServer implements AutoCloseable {
 
+    private static final String LOAD_SYSTEM_ONTOLOGY_LOCK_NAME = "load-system-ontology";
     private static final Logger LOG = LoggerFactory.getLogger(GraknEngineServer.class);
     private static final Set<String> unauthenticatedEndPoints = new HashSet<>(Arrays.asList(
             REST.WebPath.NEW_SESSION_URI,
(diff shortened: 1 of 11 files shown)


docs/pages/documentation/11-examples/working-with-tweets.md
```diff
@@ -114,7 +114,8 @@ Let's kick things off by defining a `Main` class inside the `ai.grakn.twitterexa
 
 First, we have decided to use an **in-memory graph** for simplicity's sake — working with an in-memory graph frees us from having to set up a Grakn distribution in the local machine. The in-memory graph is not for storing data and will be lost once the program finishes execution. Second, the graph will be stored in a **keyspace** named `twitter-example`.
 
-```java
+<!-- A lot of these examples are not valid Groovy, so they've been ignored in tests -->
```
pluraliseseverythings: These are merged from Felix's PR

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -121,11 +124,30 @@ public static void main(String[] args) {
     }
 
     public void start() {
+        lockAndInitializeSystemOntology();
         startHTTP();
         printStartMessage(prop.getProperty(GraknEngineConfig.SERVER_HOST_NAME),
                 prop.getProperty(GraknEngineConfig.SERVER_PORT_NUMBER));
     }
 
+    private void lockAndInitializeSystemOntology() {
```
pluraliseseverythings: This is the meat of the change

grakn-test/src/test/java/ai/grakn/test/engine/GraknEngineStartIT.java
```diff
@@ -0,0 +1,90 @@
...
+        GraknSystemProperty.CURRENT_DIRECTORY.set(StandardSystemProperty.USER_DIR.value());
+    }
+
+    @AfterClass
+    public static void tearDownClass() {
+        EmbeddedRedis.stop();
+    }
+
+    @Test
+    public void whenStartingMultipleEngines_InitializationSucceeds() throws InterruptedException {
```
pluraliseseverythings: This failed before the change with the error Thanh reported

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -121,11 +124,30 @@ public static void main(String[] args) {
...
+        try {
```
fppt: So many indentations!!  The horror!

PR #1890, opened 2017-07-18
Fix issues with tests against new twitter docs

aelred:


aelred:
@lolski I added some tests for the docs at just the same time as you added this page. Together they broke :P

The tests attempt to evaluate any Java examples as Groovy (so we can do it at runtime). Some of your examples are valid Java, but not valid Groovy. This isn't an issue, so I've just skipped them in the tests.

aelred:
!rtg - failure in property tests! new one. logging to TP

fppt:
Merging this in because it is blocking other PRs

PR #1895, opened 2017-07-19
Clean up `grakn-graph` property tests and generators

aelred:


grakn-test-tools/src/main/java/ai/grakn/generator/AbstractOntologyConceptGenerator.java
```diff
@@ -112,26 +100,21 @@ private final boolean includeNonMeta() {
...
 
-    final AbstractOntologyConceptGenerator<T> excludeAbstract() {
-        includeAbstract = Optional.of(false);
-        return this;
-    }
-
     public final void configure(Meta meta) {
-        this.meta = Optional.of(meta.value());
+        Preconditions.checkArgument(
+                !this.meta.isPresent() || this.meta.get(), "Cannot specify parameter is both meta and non-meta");
```
fppt: > Cannot specify parameter is both meta and non-meta

What?? How can this happen?

grakn-test-tools/src/main/java/ai/grakn/generator/AbstractTypeGenerator.java
```diff
@@ -0,0 +1,77 @@
...
+
+    AbstractTypeGenerator(Class<T> type) {
+        super(type);
+    }
+
+    private boolean includeAbstract(){
+        return includeAbstract.orElse(true);
+    }
+
+    final AbstractOntologyConceptGenerator<T> excludeAbstract() {
```
fppt: `includeAbstract()` returns a boolean while this one returns the generator itself. Sad :(

grakn-test-tools/src/main/java/ai/grakn/generator/AbstractOntologyConceptGenerator.java
aelred: Haha, just if you write a property test with parameter: `@Meta @NonMeta Concept concept` :P

grakn-test-tools/src/main/java/ai/grakn/generator/AbstractTypeGenerator.java
aelred: I should name them better... one is a getter, the other a setter.

PR #1898, opened 2017-07-19
More engine fixes. Redisson wasn't instantiated properly

pluraliseseverythings:


grakn-engine/src/main/java/ai/grakn/engine/GraknEngineConfig.java
```diff
@@ -59,9 +61,9 @@
 
     public static final String LOADER_REPEAT_COMMITS = "loader.repeat-commits";
 
-    public static final String REDIS_SERVER_URL = "redis.host";
-    public static final String REDIS_SERVER_PORT = "redis.port";
-    public static final String REDIS_SENTINEL_MASTER = "redis.sentinel-master";
+    public static final String REDIS_HOST = "redis.host";
```
pluraliseseverythings: Left just the host in the form host:port so we can add lists of those

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -160,18 +158,20 @@ private void loadAndUnlock(Lock lock) {
...
-            RedisCountStorage redisCountStorage,
-            Pool<Jedis> jedisPool, LockProvider lockProvider) {
+            Pool<Jedis> jedisPool,
+            LockProvider lockProvider) {
         TaskManager taskManager;
         if (!inMemoryQueue) {
             taskManager = new RedisTaskManager(engineId, prop, jedisPool, factory, lockProvider, metricRegistry);
         } else  {
+            // Redis storage for counts, in the RedisTaskManager it's created in consumers
+            RedisCountStorage redisCountStorage = RedisCountStorage.create(jedisPool);
```
fppt: Aren't the count's always stored in redis? Why the need for this wrapper only for `StandaloneTaskManager `?

grakn-engine/src/main/java/ai/grakn/engine/SystemKeyspace.java
```diff
@@ -157,18 +157,20 @@ public boolean deleteKeyspace(String keyspace){
...
             }
+            LOG.info("No other version found, loading ontology for version {}", GraknVersion.VERSION);
             loadSystemOntology(graph);
             graph.getResourceType(SYSTEM_VERSION).putResource(GraknVersion.VERSION);
             graph.admin().commitNoLogs();
-            LOG.info("Loaded system ontology to system keyspace.");
-        } catch (InvalidGraphException | NullPointerException e) {
-            e.printStackTrace(System.err);
-            LOG.error("Could not load system ontology. The error was: " + e);
+            LOG.info("Loaded system ontology to system keyspace. Took: {}", timer.stop());
```
fppt: I bet you most of the time is spent in initializing the keyspace. This is not related to this PR but I thought we should note it here.

grakn-engine/src/main/java/ai/grakn/engine/lock/JedisLock.java
```diff
@@ -0,0 +1,181 @@
...
+     */
+    private String lockName;
+
+    /**
+     * Lock expiration in miliseconds.
+     */
+    private int expireMsecs = 60 * 1000;
+
+    /**
+     * Acquire timeout in miliseconds.
```
fppt: I think these java docs are misplaced. The line below this is a boolean

grakn-engine/src/main/java/ai/grakn/engine/util/SimpleURI.java
```diff
@@ -0,0 +1,65 @@
...
+ * along with Grakn. If not, see <http://www.gnu.org/licenses/gpl.txt>.
+ */
+
+package ai.grakn.engine.util;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * This Util class just takes care of going from host and port to string and viceversa
+ * The URI class would require a schema
```
fppt: I was about to ask why did you not use the java URI class

grakn-engine/src/test/java/ai/grakn/engine/data/RedisWrapperTest.java
```diff
@@ -0,0 +1,58 @@
...
+    @Test(expected = JedisConnectionException.class)
+    public void whenBuildingSentinelWellFormed_JedisCantConnect() {
+        RedisServer server = mockRedisRule.getServer();
+        RedisWrapper.builder().setUseSentinel(true)
+                .addURI(new SimpleURI(server.getHost(), server.getBindPort()).toString())
+                .setMasterName("masterName")
+                .build();
+    }
+
+    @Test(expected = IllegalStateException.class)
```
fppt: I don't know if you can do it here but try to expect both the Exception class and the exception message. I just find it makes it easier to debug from the outsiders perspective.

grakn-test-tools/pom.xml
```diff
@@ -69,7 +69,12 @@
             <artifactId>embedded-redis</artifactId>
             <version>${redis-test.version}</version>
         </dependency>
-        
+        <dependency>
+            <groupId>com.github.zxl0714</groupId>
+            <artifactId>redis-mock</artifactId>
```
fppt: Maybe it's worth switching to a mocked redis completely for our tests rather than the fake embedded redis we use for some of the tests? Do you think this mock can handle those cases? If so we should make that change.

pluraliseseverythings: I added this instead of replacing the other one because I'm not sure if it has the full behaviour. Sometimes we just need a mock and this works fine for that but I'm not sure if it store things for example.

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
pluraliseseverythings: Good question! 
I preferred moving it to the consumers in this case. It seemed a bit like breaking the responsibilities of a generic task manager to pass it there. Instead it's created inside the consumers. It doesn't have any particular performance hit because the heavy object, the jedisPool, is passed as an argument and RedisCountStorage is just a class with convenience methods for the counts

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -338,28 +338,20 @@ private static void handleInternalError(Exception exception, Response response){
...
+                .setUseSentinel(useSentinel)
+                .setURI((useSentinel ? sentinelUrl : redisUrl));
+        if (useSentinel) {
+            builder.setMasterName(prop.tryProperty(REDIS_SENTINEL_MASTER).orElse("graknmaster"));
+        }
+        return builder.build();
     }
 
-    private static void printStartMessage(String host, String port) {
+    private void logStartMessage(String host, String port) {
```
pluraliseseverythings: Very minor thing, I don't mind keeping static. But I prefer avoiding static in general. It's also a conceptual choice here: GraknEngineServer.logStartMessage doesn't make much sense when you have more than one engine (e.g. in tests). We log the message for a certain engine not of the class engine.

grakn-engine/src/main/java/ai/grakn/engine/SystemKeyspace.java
pluraliseseverythings: Correct, it's 15-30 seconds to load the ontology. When that's not needed it starts in ~100ms

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
fppt: Just seems a bit weird. I am wondering if we can just collapse everything into one task manager and just get rid of this two mode thing all together. Before hand it was not possible because the "distributed mode" required ZK and Kafka, but now both modes just need Redis and nothing else. Do you think it will be possible to just collapse everything into one manager?

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
fppt: Fair enough. From a conceptual standpoint I agreed. I was just wondering if there was a "pragmatic" choice for it.

grakn-test-tools/pom.xml
fppt: I am just looking for a way of getting rid of the fake "embedded" redis we are using. I still sometimes get redis running in the background even with the shutdown hook. It happens much more rarely though.

grakn-engine/src/test/java/ai/grakn/engine/data/RedisWrapperTest.java
fppt: Just checked your additional commit and you did not address this one. Is it not possible?

pluraliseseverythings:
!rtg

grakn-test-tools/pom.xml
pluraliseseverythings: I just had a deeper look at this library. I'm a bit conflicted. On one side it does what a minimal in memory redis mock should do. Opens a connection on the port, reads commands and mocks redis as a Map. So you can read and write from it and it also supports TTL of data.
On the other side it's full of typos and the code is not amazing. 

In the next set of changes maybe I can scan through where we use Redis in tests and see if I can use this instead for stricter unit tests where we are not relying too much on testing Redis behaviour.

grakn-engine/src/test/java/ai/grakn/engine/data/RedisWrapperTest.java
pluraliseseverythings: I didn't find a way to do it unfortunately :(

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
pluraliseseverythings: Yes, I'm trying to converge towards that, after all we always need a Redis to run with Grakn. If that's ok with you I can make a separated PR for that.

grakn-engine/src/test/java/ai/grakn/engine/data/RedisWrapperTest.java
fppt: Awwww :( . We do it in many other tests so I thought it mat be doable here as well.

pluraliseseverythings:
!rtg

PR #1900, opened 2017-07-20
Role no longer subs thing in meta ontology

fppt:


grakn-test/src/test/java/ai/grakn/property/TypePropertyTest.java
```diff
@@ -189,6 +190,7 @@ public void whenSettingATypeAbstractFlag_TheTypesAbstractFlagIsSet(@Meta(false)
     @Property
     public void whenAnOntologyElementHasADirectSuper_ItIsADirectSubOfThatSuper(
             @Open GraknGraph graph, @FromGraph OntologyConcept ontologyConcept) {
+        assumeFalse(Schema.MetaSchema.ROLE.getLabel().equals(ontologyConcept.getLabel()));
```
aelred: Does the test fail without this assumption?

fppt: yep. `role` no longer has a super type.

fppt:
@aelred can we merge this one? or do you want to look into the generator?

aelred:
@fppt I fixed it on another branch where I'm cleaning the generators.

PR #1903, opened 2017-07-20
Add deprecated support for keywords `lhs` and `rhs`

aelred:
These have been replaced by `when` and `then`.

Keeping these keywords for just one more version means that we can test both `master` and `stable` using the same Graql queries.


aelred:
@fppt Done!

fppt:
rtg

aelred:
!rtg - all failing tests ignored and logged to TP

fppt:
!rtg - ran ignored test fro some odd reason

PR #1904, opened 2017-07-20
Bug #16183: Resources with null value would still create the resource

fppt:


fppt:
!rtg - Jenkins failed before it even started . . . weird

fppt:
!rtg

grakn-graph/src/main/java/ai/grakn/graph/internal/ResourceImpl.java
```diff
@@ -99,59 +94,11 @@ public Thing owner() {
...
-            }
-            return ((Number) value).longValue();
-        } else {
-            try {
-                return Class.forName(parentDataType.getName()).cast(value);
-            } catch (ClassNotFoundException e) {
-                throw new RuntimeException(ErrorMessage.INVALID_RESOURCE_CAST.getMessage(value, parentDataType.getName()));
-            }
-        }
+    private Resource<D> setValue(Object value) {
```
aelred: Why remove the generic here?

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
```diff
@@ -119,7 +119,7 @@ public void txCacheClear(){
      * @return A new or already existing instance
      */
     V putInstance(Schema.BaseType instanceBaseType, Supplier<V> finder, BiFunction<VertexElement, T, V> producer) {
-        vertex().graph().checkMutationAllowed();
+        preCheckForInstanceCreation();
```
aelred: Given that `putInstance` calls `addInstance`, you're making this check twice.

grakn-graph/src/main/java/ai/grakn/graph/internal/ResourceImpl.java
fppt: Because `setValue` now takes in the "persisted value" i.e the value in the format in which it's going to be saved, and this can anything really.

grakn-graph/src/main/java/ai/grakn/graph/internal/TypeImpl.java
fppt: Yes and I always was because `addInstance` gets called exclusively as well. I might be able to fix that though. Hold on. . . .

fppt:
@aelred re-review please

PR #1909, opened 2017-07-21
Use autovalue for some classes

aelred:
This is an experimental PR to see how we feel about it!


grakn-core/src/main/java/ai/grakn/concept/LabelId.java
```diff
@@ -34,49 +35,24 @@
  *
  * @author fppt
  */
-public class LabelId implements Comparable<LabelId>, Serializable {
-    private static final long serialVersionUID = 3181633335040468179L;
-
-    private Integer typeId;
-
-    private LabelId(Integer typeId){
-        Preconditions.checkNotNull(typeId);
```
fppt: How does autovalue handle nulls? This particular null check is to prevent a NPE later in life.

grakn-core/src/main/java/ai/grakn/concept/LabelId.java
```diff
@@ -34,49 +35,24 @@
...
     /**
      *
      * @return Used for indexing purposes and for graql traversals
      */
-    public Integer getValue(){
-        return typeId;
-    }
-
-    @Override
-    public boolean equals(Object o) {
```
fppt: On that note, How does autovalue handle equality and the hashcode? Is it based on the fields somehow? because if so I imagine a little room for error there.

grakn-core/src/main/java/ai/grakn/concept/LabelId.java
```diff
@@ -85,13 +61,13 @@ public boolean isValid(){
      * @return The matching type ID
      */
     public static LabelId of(Integer value){
-        return new LabelId(value);
+        return new AutoValue_LabelId(value);
```
fppt: Not a fan of this syntax :/

pom.xml
```diff
@@ -202,6 +202,12 @@
             <version>${guava.version}</version>
         </dependency>
         <dependency>
+            <groupId>com.google.auto.value</groupId>
+            <artifactId>auto-value</artifactId>
+            <version>${auto-value.version}</version>
+            <scope>provided</scope>
```
fppt: provided? Man I actually had to google this one. Why is maven so bloody complicated.

grakn-core/src/main/java/ai/grakn/concept/LabelId.java
aelred: It disallows them, unless you mark the method as `@Nullable`, so we're exactly as safe as before 👍

grakn-core/src/main/java/ai/grakn/concept/LabelId.java
aelred: It's based on all the fields, which is what you want most of time. If you want different behaviour, you can implement them yourself.

grakn-core/src/main/java/ai/grakn/concept/LabelId.java
aelred: That's true! It only occurs in one place though.

aelred:
@fppt I've pushed some stuff that you might want to re-review.

grakn-core/src/main/java/ai/grakn/concept/ConceptId.java
```diff
@@ -60,4 +60,10 @@ public int compareTo(ConceptId o) {
     public static ConceptId of(String value){
         return new AutoValue_ConceptId(value);
     }
+
+    @Override
+    public final String toString() {
+        // TODO: Consider using @AutoValue toString
```
fppt: Decent compromise I suppose

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/MedianVertexProgram.java
```diff
@@ -129,7 +129,7 @@ public void storeState(final Configuration configuration) {
     public void loadState(final Graph graph, final Configuration configuration) {
         super.loadState(graph, configuration);
         configuration.subset(RESOURCE_TYPE).getKeys().forEachRemaining(key ->
-                statisticsResourceLabelIds.add(LabelId.of(configuration.getInt(RESOURCE_TYPE + "." + key))));
+                statisticsResourceLabelIds.add((LabelId) configuration.getProperty(RESOURCE_TYPE + "." + key)));
```
fppt: More explicit now, which I like.

fppt: Also if we create our own `configuration ` wrapper we can return the object as they should be. No need for dirty castings everywhere. But that's a future concern I think.

grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/VarImpl.java
```diff
@@ -76,4 +76,21 @@ public Var getVarName() {
     public String toString() {
         return "$" + getValue();
     }
+
+    @Override
+    public final boolean equals(Object o) {
```
fppt: Why is this ugly thing here now when before it wasn't?

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/MedianVertexProgram.java
aelred: I like this more too. We put the actual `LabelId` in the config. I'm slightly concerned this won't work on a "real" cluster or something, so I'm going to wait for @duckofyork to confirm it's OK.

grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/VarImpl.java
aelred: It was here before - you're just looking at a sub-set of changes! I reverted it back to how it was.

I don't think this custom `equals` implementation is a good idea, but fixing it is too much for this PR.

fppt:
Question answered. Everything is good again. Assigning @duckofyork for him to checkout and approve the analytics related changes.

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/MedianVertexProgram.java
duckofyork: Theoretically, this should be fine, but I remember something always went wrong when we put non-primitive type in the config.

fppt: Wouldn't our titan tests catch any failures if it wasn't okay?

duckofyork: It can catch the failure of spark computer in standalone mode, but on a real cluster I guess it's more difficult to test this

PR #1912, opened 2017-07-21
[WIP - do not merge] Redis + cassandra supervision

lolski:
- Functionality: Starts and stops Redis and Cassandra together with Engine
- Tests
- Binary path and PID file path are currently hardcoded. Not sure if we need to make it configurable from a config file?
- This functionality currently depends on these binaries: `bin/nodetool`, `bin/redis-*`, and `bin-cassandra`


lolski:
It broke some other tests. Whoops

grakn-engine/src/main/java/ai/grakn/engine/externalcomponents/CassandraSupervisor.java
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Responsible for supervising cassandra.
+ *
+ * @author Ganeshwara Herawan Hananda
+ */
+
+public class CassandraSupervisor {
+    private OperatingSystemCalls osCalls;
+    private final Logger LOG = Logger.getLogger(CassandraSupervisor.class.getName());
+    private final long WAIT_TIME_BETWEEN_ATTEMPT_MS = 2000;
```
fppt: This should probably be a configurable value to aid with debugging. 

I am just thinking, when users encounter timeout issues with grakn, one of the first things they will likely do is increase the timeout as a sanity check. I think we have all done that at some point.

grakn-engine/src/main/java/ai/grakn/engine/externalcomponents/RedisSupervisor.java
```diff
@@ -0,0 +1,100 @@
...
+
+        if (osCalls.isMac()) {
+            this.startRedisCmd = baseWorkDir + "bin/redis-server-osx " + baseWorkDir + "conf/redis/redis.conf";
+            this.stopRedisCmd = baseWorkDir + "bin/redis-cli-osx shutdown";
+
+        }
+        else {
+            this.startRedisCmd = baseWorkDir + "bin/redis-server-linux " + baseWorkDir + "conf/redis/redis.conf";
+            this.stopRedisCmd = baseWorkDir + "bin/redis-cli-linux shutdown";
+        }
```
fppt: //TODO: Windows?

grakn-engine/src/main/java/ai/grakn/engine/externalcomponents/RedisSupervisor.java
```diff
@@ -0,0 +1,100 @@
...
+        } else {
+            LOG.info("no redis process is found.");
+        }
+    }
+
+
+    public void start() throws IOException, InterruptedException {
+        int status = osCalls.execAndReturn(new String[] { "sh", "-c", startRedisCmd } );
+        System.out.println("== " + startRedisCmd + " / " + status + " ==");
+        if (status != 0) throw new ExternalComponentException("unable to start redis - " + startRedisCmd);
```
fppt: For your exception handling see how the other exception classes are implemented. For example look at `GraphOperationException`. If possible try to change your exceptions to be the same style. This makes it easier to reuse exception messages and it makes it easier to keep all our exceptions in the same place.

grakn-engine/src/main/java/ai/grakn/engine/externalcomponents/RedisSupervisor.java
```diff
@@ -0,0 +1,100 @@
...
+            LOG.info("redis has been stopped.");
+        } else {
+            LOG.info("no redis process is found.");
+        }
+    }
+
+
+    public void start() throws IOException, InterruptedException {
+        int status = osCalls.execAndReturn(new String[] { "sh", "-c", startRedisCmd } );
+        System.out.println("== " + startRedisCmd + " / " + status + " ==");
```
fppt: Are these `printlns` needed?

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -112,15 +117,29 @@ public GraknEngineServer(GraknEngineConfig prop) {
         this.taskManager = startTaskManager(inMemoryQueue, redisCountStorage, jedisPool, lockProvider);
     }
 
-
     public static void main(String[] args) {
         GraknEngineConfig prop = GraknEngineConfig.create();
-        // Start Engine
+
+        // Start external components (Cassandra and Redis)
```
fppt: With this approach how can I connect grakn to my own running instance of cass and redis? Can I even do that anymore? 

We don't formally support windows but I *think* this will kill off the "hacks" that people can do to get grakn running on windows. I.e. starting up cass and redis and configuring grakn to connect to them.

grakn-engine/src/main/java/ai/grakn/engine/externalcomponents/ExternalComponentException.java
```diff
@@ -0,0 +1,31 @@
...
+
+package ai.grakn.engine.externalcomponents;
+
+/**
+ * Should be thrown upon encountering any error when unable to start / stop an external component
+ *
+ * @author Ganeshwara Herawan Hananda
+ */
+
+public class ExternalComponentException extends RuntimeException {
```
fppt: This should probably extend `GraknException` or `GraknBackendException` . . . the latter seems like a good one. Check the javadocs of `GraknBackendException` and tell me what you think.

lolski:
@fppt @pluraliseseverythings In my opinion, this work lacks proper testing. We have a couple unit tests that test the process supervision and that's it. For a change of this scale I want to feel more confident.

This week I'm gonna do a real world testing scenario and see how well this solution holds. If that works Then we can think of merging this codebase.

fppt:
@lolski we have a basic CI framework which @sheldonkhall has designed for us. Maybe there is a way we could leverage that to do the testing you planning? That way any tests you write/perform can be reused.

lolski:
@fppt @sheldonkhall as the work is mainly about supervision an OS process, the test will need to be able to start, stop, and do health check on process. I think it can be done however not sure how easy it is.

If I can get a CI account to play around that would be ideal

grakn-engine/src/main/java/ai/grakn/engine/externalcomponents/RedisSupervisor.java
lolski: Good idea. Comment added

grakn-engine/src/main/java/ai/grakn/engine/externalcomponents/RedisSupervisor.java
lolski: Removed

grakn-engine/src/main/java/ai/grakn/engine/externalcomponents/RedisSupervisor.java
lolski: I didn't realise we have `GraknBackendException` already. Therefore the code now uses that.

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
lolski: @fppt do we want this? at one side it is good that people can still have the flexibility of doing it. But on the other side, this would expose the internal storage and queue which is what we're aiming to simplify.

Technically this is quite easy to do as I just need to add an option such as `cassandra_host=internal | <external cassandra IP address>`

grakn-engine/src/main/java/ai/grakn/engine/externalcomponents/ExternalComponentException.java
lolski: Yes, now it uses `GraknBackendException`.

lolski:
Closing this PR for now. Will reopen after we've got good results from the load testing

PR #1918, opened 2017-07-24
Merge from master

fppt:

PR #1920, opened 2017-07-24
[Stable] Bug #16211: Correctly infer types in conjunction if type present

kasper-piskorski:
Bug #16211


fppt:
This should not be merged until https://github.com/graknlabs/grakn/pull/1918 is merged. If this is merged before https://github.com/graknlabs/grakn/pull/1918 then we would need to remerge back into master . . . again

fppt:
@kasper-piskorski you prefixed this with [Stable] but this PR is not actually targeting stable.

PR #1921, opened 2017-07-24
[Stable] Bug #16211: Correctly infer types in conjunction if type present

kasper-piskorski:


fppt:
This should not be merged until https://github.com/graknlabs/grakn/pull/1918 is merged. If this is merged before https://github.com/graknlabs/grakn/pull/1918 then we would need to remerge back into master . . . again

fppt:
!rtg - because a merge went in and I just want to be super sure that this won't break stable

kasper-piskorski:
merged here: https://github.com/graknlabs/grakn/pull/1924

PR #1922, opened 2017-07-25
Bug #16214: Odd validation errors caused by nasty cache leak

fppt:
This one was painful. Lost a few years of life after this one.


grakn-graph/src/main/java/ai/grakn/graph/internal/GraphCache.java
```diff
@@ -60,7 +59,7 @@
         int cacheTimeout = Integer.parseInt(properties.get(AbstractGraknGraph.NORMAL_CACHE_TIMEOUT_MS).toString());
         cachedTypes = CacheBuilder.newBuilder()
                 .maximumSize(1000)
-                .expireAfterAccess(cacheTimeout, TimeUnit.MILLISECONDS)
+                //.expireAfterAccess(cacheTimeout, TimeUnit.MILLISECONDS)
```
aelred: Is this meant to be committed?

fppt: oops no. Just something accidentally committed during the madness of it all.

fppt:
!rtg - failed to contact github or something

PR #1927, opened 2017-07-26
Add check to property test

fppt:


fppt:
This is the worst PR that ever PR'd

PR #1928, opened 2017-07-27
Bug #16036: Even nastier caching bug

fppt:
This one made the previous one feel like a warm hug.


fppt:
Yes I am aware this is going into master but given the depths of the changes and how hidden the bug is I would rather this one go into master than stable. It is a potentially breaking change.

fppt:
@aelred Any better ideas that having `CachedSet` I ask because I just realised I will need a `CachedMap` as well and all this is starting to feel very in elegant.

grakn-graph/src/main/java/ai/grakn/graph/internal/Cache.java
```diff
@@ -74,6 +74,10 @@ public V get(){
         return cachedValue.get();
     }
 
+    V copyShared(){
+        return sharedValue.get();
+    }
```
aelred: We should add a comment/javadoc indicating this method is intended to be overridden.

grakn-test/src/test/java/ai/grakn/test/graql/shell/GraqlShellIT.java
```diff
@@ -68,7 +69,7 @@
 public class GraqlShellIT {
 
     @ClassRule
-    public static final DistributionContext dist = DistributionContext.startInMemoryEngineProcess().inheritIO(false);
+    public static final DistributionContext dist = DistributionContext.startInMemoryEngineProcess().inheritIO(true);
```
aelred: I accidentally committed this! Will remove.

grakn-graph/src/main/java/ai/grakn/graph/internal/TxCache.java
```diff
@@ -386,6 +386,11 @@ Json getFormattedLog(){
     void closeTx(String closedReason){
         isTxOpen = false;
         this.closedReason = closedReason;
+
+        //Clear Concept Caches
+        conceptCache.values().forEach(concept -> ((ContainsTxCache) concept).txCacheClear());
```
aelred: Since we know this cast is safe, we can add one of your `ContainsTxCache.from(Concept)` methods.

grakn-graph/src/main/java/ai/grakn/graph/internal/Cache.java
```diff
@@ -114,6 +118,9 @@ void ifPresent(Consumer<V> modifier){
      * that it can be accessed via all transactions.
      */
     void flush(){
-        if(isPresent()) sharedValue = Optional.of(get());
+        if(isPresent() && !sharedValue.equals(get())) {
+            this.sharedValue = Optional.of(get());
+        }
```
aelred: Are you sure this works? `sharedValue` is `Optional<V>` so I don't think you can compare it to `get()` which is just `V`.

grakn-graph/src/main/java/ai/grakn/graph/internal/GraphCache.java
```diff
@@ -59,7 +59,7 @@
 
         int cacheTimeout = Integer.parseInt(properties.get(AbstractGraknGraph.NORMAL_CACHE_TIMEOUT_MS).toString());
         cachedTypes = CacheBuilder.newBuilder()
-                .maximumSize(1000)
+                //.maximumSize(1000)
```
fppt: 🐄

grakn-graph/src/main/java/ai/grakn/graph/internal/Cache.java
fppt: Yar. Will do.

grakn-graph/src/main/java/ai/grakn/graph/internal/Cache.java
fppt: Oops. Fixing required.

grakn-graph/src/main/java/ai/grakn/graph/internal/TxCache.java
fppt: Good point!

grakn-test/src/test/java/ai/grakn/test/graql/shell/GraqlShellIT.java
fppt: 🐐

fppt:
@aelred re-review please

grakn-graph/src/main/java/ai/grakn/graph/internal/Cache.java
```diff
@@ -110,10 +114,14 @@ void ifPresent(Consumer<V> modifier){
...
     /**
-     * Takes the current value in the transaction cache if it is present and puts it in the sharedValue reference so
+     * Takes the current value in the transaction cache if it is present and puts it in the valueGlobal reference so
      * that it can be accessed via all transactions.
      */
     void flush(){
-        if(isPresent()) sharedValue = Optional.of(get());
+        if(isPresent()) {
+            V newValue = get();
+            if(!valueGlobal.isPresent() || !valueGlobal.get().equals(newValue)) valueGlobal = Optional.of(get());
```
aelred: Should we also copy when flushing something into the global cache?

fppt: No need really. Immediately after flushing the cached value is removed. Might even put the remove method in here at some point

grakn-graph/src/main/java/ai/grakn/graph/internal/Cacheable.java
```diff
@@ -0,0 +1,118 @@
...
+        return new Cacheable<>((o) -> o);
+    }
+
+    static Cacheable<Boolean> bool(){
+        return new Cacheable<>((o) -> o);
+    }
+
+    static <T extends OntologyConcept> Cacheable<T> ontologyConcept(){
+        return new Cacheable<>((o) -> o);
+    }
```
aelred: Can we do `<T extends Concept> Cacheable<T> concept()` and remove all the others?

grakn-graph/src/main/java/ai/grakn/graph/internal/Cacheable.java
```diff
@@ -0,0 +1,118 @@
...
+ * </p>
+ *
+ * @author fppt
+ *
+ * @param <V>
+ */
+public class Cacheable<V> {
+    private final Function<V, V> copier;
+
+    private Cacheable(Function<V, V> copier){
```
aelred: There's a more specific functional type called `UnaryOperator<V>` you could use if you want.

grakn-graph/src/main/java/ai/grakn/graph/internal/ThingImpl.java
```diff
@@ -78,7 +78,7 @@
         }
 
         return vertex().graph().factory().buildConcept(typeEdge.get().target());
-    });
+    }, Cacheable.type());
```
aelred: This is so picky, but you should make Cacheable the first parameter! For prettiness sake :P

aelred:
This looks good! I like.

grakn-graph/src/main/java/ai/grakn/graph/internal/Cacheable.java
fppt: Good point. Will change

grakn-graph/src/main/java/ai/grakn/graph/internal/Cacheable.java
fppt: Nope. Too broad as in come cases it would require down casting. In some cases we explicitly cache `Thing` for example

grakn-graph/src/main/java/ai/grakn/graph/internal/Cache.java
aelred: Got it!

But now I confused myself again. What's the check for? I'm starting to think it doesn't really do anything.

grakn-graph/src/main/java/ai/grakn/graph/internal/ThingImpl.java
fppt: Ugh . . . no later. I agree it needs to be prettier but I am too bored/lazy . . . blazy? to make that change now :P

grakn-graph/src/main/java/ai/grakn/graph/internal/Cacheable.java
aelred: I'm not sure I understand. You could do `concept<Thing>()` to get a `Cacheable<Thing>`.

fppt: Yes you right I misunderstood things.

PR #1934, opened 2017-07-27
The engine was bouincing things in the queues. 

pluraliseseverythings:
Now they are destroyed after a long time as a fallback


grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -123,15 +125,18 @@ public void start() {
         logStartMessage(
                 prop.getProperty(GraknEngineConfig.SERVER_HOST_NAME),
                 prop.getProperty(GraknEngineConfig.SERVER_PORT_NUMBER));
-        lockAndInitializeSystemOntology();
-        startHTTP();
+        synchronized (this){
```
pluraliseseverythings: Avoid weird issues with close happening before startup is completed

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -123,15 +125,18 @@ public void start() {
...
+            CompletableFuture.allOf(CompletableFuture.runAsync(this::lockAndInitializeSystemOntology), CompletableFuture.runAsync(this::startHTTP)).join();
```
pluraliseseverythings: Run in parallel, no point waiting for one to complete before we start the other

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisInflightTaskConsumer.java
```diff
@@ -92,26 +110,43 @@ public void run() {
     }
 
     private void processElement(String key, Jedis resource, String head) {
+        processed.mark();
         // TODO Use Jackson for this
         long runAt = Json.read(head).at("args").at(0).at("taskState")
                 .at("schedule").at("runAt").asLong();
         Instant runAtDate = Instant.ofEpochMilli(runAt);
         Duration gap = Duration.between(runAtDate, Instant.now());
+        if (gap.getSeconds() > destroyInterval.getSeconds()) {
```
pluraliseseverythings: New logic. After a while we stop retrying a task and we log it

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisTaskStorage.java
```diff
@@ -102,17 +117,24 @@ public TaskState getState(TaskId id) throws GraknBackendException {
...
     }
 
     @Override
-    public Set<TaskState> getTasks(TaskStatus taskStatus, String taskClassName, String createdBy,
-            EngineID runningOnEngine, int limit, int offset) {
+    public Set<TaskState> getTasks(@Nullable TaskStatus taskStatus, @Nullable String taskClassName,
+            @Nullable String createdBy, @Nullable EngineID runningOnEngine, int limit, int offset) {
         try (Jedis jedis = redis.getResource()) {
-            // TODO change structure of task keys with a prefix. Can we do better than filtering?
-            Stream<TaskState> stream = jedis.keys("*-*-*-*").stream().map(value -> (TaskState) deserialize(Base64.getDecoder().decode(value)));
```
pluraliseseverythings: Now we store keys properly, with a "state" prefix

pluraliseseverythings:
!rtg Checking the failure is not due to some weird DB state

pluraliseseverythings:
!rtg

grakn-engine/src/main/java/ai/grakn/engine/SystemKeyspace.java
```diff
@@ -98,6 +98,9 @@ public boolean ensureKeyspaceInitialised(String keyspace) {
 
         try (GraknGraph graph = factory.getGraph(SYSTEM_GRAPH_NAME, GraknTxType.WRITE)) {
             ResourceType<String> keyspaceName = graph.getOntologyConcept(KEYSPACE_RESOURCE);
+            if (keyspaceName == null) {
+                throw new RuntimeException("Graph for keyspace [" + keyspace + "] not properly initialized. Missing keyspace name resource");
```
fppt: It concerns be that this can happen. Why is that? 

Also, tiny poke for the sake of being consistent should this not be a `GraknException` ?

grakn-engine/src/main/java/ai/grakn/engine/lock/JedisLock.java
```diff
@@ -45,11 +46,12 @@
...
     private Lock lock = new ReentrantLock();
 
-    public JedisLock(Pool<Jedis> jedis, String lockName) {
+    public JedisLock(Pool<Jedis> jedis, String internalLockName) {
         Preconditions.checkNotNull(jedis,"JedisPool used in lock cannot be null");
-        Preconditions.checkArgument(lockName != null && !lockName.isEmpty(),"Lock name not valid");
+        Preconditions.checkArgument(internalLockName != null && !internalLockName.isEmpty(),"Lock name not valid");
         this.jedis = jedis;
-        this.lockName = lockName;
+        this.lockName = internalLockName;
```
fppt: Is there really any reason to keep the unprefixed `lockName` ? I just find it slightly confusing because I was trying to find it's purpose.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisInflightTaskConsumer.java
```diff
@@ -47,18 +50,33 @@
...
     public RedisInflightTaskConsumer(Pool<Jedis> jedisPool, Duration processInterval,
-            Config config, String queueName) {
+            Config config, String queueName, MetricRegistry metricRegistry) {
         this.jedisPool = jedisPool;
         this.processInterval = processInterval;
+        // After a certain time we get rid of it
+        this.destroyInterval = processInterval.multipliedBy(10);
         this.config = config;
         this.queueName = queueName;
+        this.exceptions = metricRegistry.meter(name(RedisInflightTaskConsumer.class, "exceptions"));
```
fppt: How much overhead does the `metricRegistry` add? I ask because to my eye it looks like it is tracking a lot.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisInflightTaskConsumer.java
```diff
@@ -92,26 +110,43 @@ public void run() {
...
+            destroyable.mark();
+            LOG.info("Found really old dead task in inflight, destroying it: {}", head);
+            attemptDestroy(key, resource);
+        }
         if (gap.getSeconds() > processInterval.getSeconds()) {
-            LOG.info("Found dead task in inflight: {}", head);
-            String keyDest = JesqueUtils
-                    .createKey(config.getNamespace(), QUEUE, queueName);
+            dead.mark();
+            LOG.info("Found dead task in inflight, moving it: {}", head);
```
fppt: So what is a "dead" task anyway? How do we end up making them? 

Shouldn't we rather not make "dead" tasks rather then working around them?

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisTaskStorage.java
```diff
@@ -68,18 +72,29 @@ public static RedisTaskStorage create(Pool<Jedis> jedisPool, MetricRegistry metr
...
+            String key = encodeKey.apply(state.getId().getValue());
+            LOG.debug("New state {}", key);
+            String value = new String(Base64.getEncoder().encode(SerializationUtils.serialize(state)),
+                    Charsets.UTF_8);
+            String status = jedis.set(key, value, "nx", "ex", 60*60/*expire time in seconds*/);
+            if (status != null && status.equalsIgnoreCase("OK")) {
+                return state.getId();
+            } else {
+                LOG.error("Could not write state {} to redis. Returned: {}", key, status);
+                throw new GraknBackendException("Could not save task state " + key);
```
fppt: Tiny poke again. We usually try to use a factory approach for creating and throwing errors . We find this makes it easier to prevent making duplicate error messages as well as keeping error messages of the same format.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisTaskStorage.java
```diff
@@ -102,17 +117,24 @@ public TaskState getState(TaskId id) throws GraknBackendException {
...
 
     @Override
-    public Set<TaskState> getTasks(TaskStatus taskStatus, String taskClassName, String createdBy,
-            EngineID runningOnEngine, int limit, int offset) {
+    public Set<TaskState> getTasks(@Nullable TaskStatus taskStatus, @Nullable String taskClassName,
+            @Nullable String createdBy, @Nullable EngineID runningOnEngine, int limit, int offset) {
         try (Jedis jedis = redis.getResource()) {
-            // TODO change structure of task keys with a prefix. Can we do better than filtering?
-            Stream<TaskState> stream = jedis.keys("*-*-*-*").stream().map(value -> (TaskState) deserialize(Base64.getDecoder().decode(value)));
+            Stream<TaskState> stream = jedis.keys(PREFIX + "*").stream().map(key -> {
```
fppt: So correct me is I am wrong. Doesn't using a regex in a redis query constitute a complex or slow query? If so do we know how this will scale? Could this become a bottleneck?

grakn-test-tools/src/main/java/ai/grakn/util/EmbeddedRedis.java
```diff
@@ -64,8 +65,14 @@ public static void start(int port){
...
                     .build();
             if (!redisServer.isActive()) {
-                redisServer.start();
-                LOG.info("Redis started.");
+                try {
+                    redisServer.start();
+                } catch (EmbeddedRedisException e) {
+                    LOG.warn("Unexpected Redis instance already running on port {}", port, e);
+                } catch (Exception e) {
+                    LOG.warn("Exception while trying to start Redis on port {}. Will attempt to continue.", port, e);
```
fppt: I hope this makes things more robust

grakn-test/src/test/java/ai/grakn/test/engine/GraknEngineFailoverIT.java
```diff
@@ -94,9 +101,9 @@ public void whenSubmittingTasksToOneEngine_TheyComplete(List<TaskState> tasks1)
     }
 
 
-    @Property(trials=10)
+    @Property
     public void whenSubmittingTasksToTwoEngines_TheyComplete(
-            List<TaskState> tasks1, List<TaskState> tasks2) throws Exception {
+            @When(seed = 3819756137707870400L)  List<TaskState> tasks1, @When(seed = -2649665058549711049L)  List<TaskState> tasks2) throws Exception {
```
fppt: I don't think you want to commit the seeds.

grakn-test/src/test/java/ai/grakn/test/engine/tasks/BackgroundTaskTestUtils.java
```diff
@@ -90,15 +95,15 @@ private static void waitForStatus(TaskStateStorage storage, TaskState task, Set<
     public static synchronized void waitForStatus(TaskStateStorage storage, TaskId task, Set<TaskStatus> status) {
         Instant initial = Instant.now();
         long duration = Duration.between(initial, Instant.now()).toMillis();
-        while(duration < 30000) {
+        while(duration < 120000) {
```
fppt: We are so patient now!

grakn-engine/src/main/java/ai/grakn/engine/SystemKeyspace.java
pluraliseseverythings: This is the race condition I was dealing with at some point. I'll change it to GraknException but it's nicer to have a proper message there rather than a nullpointer exception

grakn-engine/src/main/java/ai/grakn/engine/lock/JedisLock.java
pluraliseseverythings: The get takes the unprefixed version, it breaks things if we return the prefixed version

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisInflightTaskConsumer.java
pluraliseseverythings: Virtually no overhead. A mark has as much overhead as a lookup on a map and a +1.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisInflightTaskConsumer.java
pluraliseseverythings: Yes, next step is to filter dead tasks in the logs and replay them and see what's wrong.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisTaskStorage.java
pluraliseseverythings: Ok!

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisTaskStorage.java
pluraliseseverythings: Should be fine™ 

"While the time complexity for this operation is O(N), the constant times are fairly low. For example, Redis running on an entry level laptop can scan a 1 million key database in 40 milliseconds."

Considering it's just to scan the tasks, we should rarely get to 1 million tasks in the queue

grakn-test-tools/src/main/java/ai/grakn/util/EmbeddedRedis.java
pluraliseseverythings: I'm going to refactor the EmbeddedRedis next

grakn-test/src/test/java/ai/grakn/test/engine/GraknEngineFailoverIT.java
pluraliseseverythings: Correct

grakn-test/src/test/java/ai/grakn/test/engine/tasks/BackgroundTaskTestUtils.java
pluraliseseverythings: I should probably log the final duration so we can see what the distribution looks like. This is just to avoid the occasional annoying test failure

PR #1946, opened 2017-08-01
Refactor insert query execution

aelred:
A few objectives:
- Prepare for the new syntax to refer to resource relations in `has`
- Simplify and correct errors when creating concepts with unexpected/missing parameters (such as `val` on an entity)
- Make a `VarProperty` always responsible for inserting itself, not `InsertQueryExecutor`
- Simplify and better document intention of `InsertQueryExecutor`


grakn-core/src/main/java/ai/grakn/exception/GraqlQueryException.java
```diff
@@ -128,48 +118,44 @@ public static GraqlQueryException insertRelationWithoutType() {
...
         return new GraqlQueryException(var + " cannot be an instance of meta-type " + type.getLabel());
     }
 
     public static GraqlQueryException insertMetaType(Label label, OntologyConcept ontologyConcept) {
         return new GraqlQueryException(ErrorMessage.INSERT_METATYPE.getMessage(label, ontologyConcept.getLabel()));
     }
 
-    public static GraqlQueryException insertMultipleValues(ValuePredicateAdmin predicate, Object value) {
-        return new GraqlQueryException(INSERT_MULTIPLE_VALUES.getMessage(value, predicate));
+    public static GraqlQueryException insertMultipleProperties(String property, Object value1, Object value2) {
```
fppt: Take this more as a suggestion than a "must change" but for each exception I like to javadoc what conditions result in the exception being thrown. In many cases the title of the method or the contents give away the condition but I still think it's nice to have it in clear text. For example:

```
    /**
     * Thrown when a {@link Thing} is not allowed to have {@link Resource} of that {@link ResourceType}
     */
    public static GraphOperationException hasNotAllowed(Thing thing, Resource resource)
```

grakn-core/src/main/java/ai/grakn/util/CommonUtil.java
```diff
@@ -160,4 +162,47 @@ public static RuntimeException unreachableStatement(@Nullable String message, Th
...
         };
     }
+
+    public static <T, K, V> Collector<T, ?, HashMultimap<K, V>> flatteningToMultimap(
+            Function<? super T, ? extends K> keyFunction,
+            Function<? super T, ? extends Stream<? extends V>> valueFunction) {
+        return flatteningToMultimap(keyFunction, valueFunction, HashMultimap::create);
+    }
+
+    public static <T, K, V, M extends Multimap<K, V>> Collector<T, ?, M> flatteningToMultimap(
```
fppt: What???? All these generics make my eyes bleed. I have been staring at this for 10 mins and I am still not sure I fully get it. The end result is clear but could you maybe make the process a bit clearer?

Some ways you could clarify this:
1. Define the purpose of each function
2. Provide an example usage in the comments

grakn-core/src/main/java/ai/grakn/util/ErrorMessage.java
```diff
@@ -146,24 +146,13 @@
     NO_LABEL_SPECIFIED_FOR_HAS("no label was specified for a resource type in a 'has' property"),
     MULTIPLE_GRAPH("a graph has been specified twice for this query"),
 
-    INSERT_RULE_WITHOUT_WHEN("the rule '%s' doesn't have a when specified"),
```
fppt: I noticed you moved all of these messages into the exceptions themselves. I am now wondering if we should do this across the board.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/ConceptBuilder.java
```diff
@@ -0,0 +1,334 @@
...
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Function;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Class for building a {@link Concept}, by providing properties.
```
fppt: Great javadocs!

grakn-graql/src/main/java/ai/grakn/graql/internal/query/ConceptBuilder.java
```diff
@@ -0,0 +1,334 @@
...
+
+    /**
+     * A set of expected parameters, used when executing {@link #build()}.
+     * <p>
+     * Every time {@link #expect(BuilderParam)} or {@link #expectOrDefault(BuilderParam, Object)} is called, the argument is
+     * added to this set. After the concept is built, any parameter not in this set is considered "unexpected". If it
+     * is present in the field {@link #params}, then an error is thrown.
+     * </p>
+     */
+    private final Set<BuilderParam<?>> expectedParams = new HashSet<>();
```
fppt: So this is like a list of "authorisedParams" ? So if something is added to `params` which is not in `expectedParams` you get an exception?

grakn-graql/src/main/java/ai/grakn/graql/internal/query/InsertQueryExecutor.java
```diff
@@ -67,285 +63,321 @@
...
+    // A mutable map of concepts "under construction" that require more information before they can be built
+    private final Map<Var, ConceptBuilder> conceptBuilders = new HashMap<>();
+
+    // An immutable set of all properties
+    private final ImmutableSet<VarAndProperty> properties;
+
+    // A partition (disjoint set) indicating which `Var`s should refer to the same concept
+    private final Partition<Var> equivalentVars;
+
+    // A map, where `dependencies.containsEntry(x, y)` implies that `y` must be inserted before `x` is inserted.
```
fppt: 👍

grakn-graql/src/main/java/ai/grakn/graql/internal/query/InsertQueryExecutor.java
```diff
@@ -67,285 +63,321 @@
...
+            the dependencies between properties and variables.
+
+            `propDependencies.containsEntry(prop, var)` indicates that the property `prop` cannot be inserted until
+            the concept represented by the variable `var` is created.
+
+            For example, the property `$x isa $y` depends on the existence of the concept represented by `$y`.
+         */
+
+        Multimap<VarAndProperty, Var> propDependencies =
+                properties.stream().collect(flatteningToMultimap(identity(), VarAndProperty::requiredVars));
```
fppt: Ya these uses of `flatteningToMultimap ` don't make it any clearer how it works

grakn-graql/src/main/java/ai/grakn/graql/internal/query/InsertQueryExecutor.java
```diff
@@ -67,285 +63,321 @@
...
+        Multimaps.invertFrom(producedVars, varDependencies);
+
+        /*
+            Equivalent vars are variables that must represent the same concept as another var.
+
+                 $x label movie, sub entity;
+                 $y label movie;
+                 $z isa $y;
+
+            In this example, `$z isa $y` must not be inserted before `$y` is. However, `$y` does not have enough
```
fppt: I am not sure I follow here. You say:

> `$z isa $y` must not be inserted before `$y`

Why is that? I don't see anything depending on `$z` if anything it looks like $y need to be inserted before $z and due to:

> $y label movie;

It looks like it can be.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/InsertQueryExecutor.java
```diff
@@ -67,285 +63,321 @@
...
+
+            equivalentVars.merge(props.first.var(), props.second.var());
+        });
+
+        /*
+            Together, `propDependencies` and `varDependencies` can be composed into a single many-to-many relation:
+
+                dependencies = propDependencies ∘ varDependencies
+
+            By doing so, we map _directly_ between properties, skipping the vars. For example, if we previously had:
```
fppt: This following portion really makes clear the purpose of computing equivalentVars. Nice.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/InsertQueryExecutor.java
```diff
@@ -67,285 +63,321 @@
...
-        Optional<Label> typeLabel = var.getTypeLabel();
-        Optional<ConceptId> id = var.getId();
+        VarAndProperty property;
 
-        typeLabel.ifPresent(label -> {
-            if (type.isPresent()) {
-                throw GraqlQueryException.insertInstanceWithLabel(label);
-            }
-        });
+        while ((property = propertiesWithoutDependencies.poll()) != null) {
```
fppt: Just checking my understanding here for this specific loop:

You basically iteratively filtering out dependencies by starting with with things which have no dependencies. I.e. things which can be inserted. When you know what these things are you then eliminate them as dependencies. This in turn lets you find new things with no dependencies and then the process repeats itself?

If my above description is correct than stating this in a comment may have value.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/InsertQueryExecutor.java
```diff
@@ -67,285 +63,321 @@
...
-                        .orElseThrow(() -> GraqlQueryException.insertRuleWithoutLhs(var));
-                ThenProperty then = var.getProperty(ThenProperty.class)
-                        .orElseThrow(() -> GraqlQueryException.insertRuleWithoutRhs(var));
-                return type.asRuleType().putRule(when.getPattern(), then.getPattern());
-            });
-        } else if (type.getLabel().equals(Schema.MetaSchema.THING.getLabel())) {
-            throw GraqlQueryException.createInstanceOfMetaConcept(var, type);
-        } else {
-            throw CommonUtil.unreachableStatement("Can't recognize type " + type);
+    public ConceptBuilder builder(Var var) {
```
fppt: I really like `ConceptBuilder` it adds a nice isolation of concerns.

grakn-graql/src/test/java/ai/grakn/graql/internal/query/InsertQueryTest.java
```diff
@@ -381,7 +382,11 @@ public void testErrorWhenInsertWithMultipleIds() {
     @Test
     public void testErrorWhenInsertWithMultipleValues() {
```
fppt: I know it's not related to this PR but you have a bunch of crappy test names which could use your love and attention.

grakn-core/src/main/java/ai/grakn/util/CommonUtil.java
aelred: Yeah, this one is terrifying. The method already exists in the latest Guava. I may copy their Javadoc (and indicate that's where it's from).

grakn-core/src/main/java/ai/grakn/util/ErrorMessage.java
aelred: I agree!

These particular messages haven't actually been moved though, they've all been deleted and replaces with more general messages.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/ConceptBuilder.java
aelred: Yes, that's right. It gets modified as the concept is built, like so:

```
if (has(LABEL)) { // Checks for LABEL in params
    ...
    // Retrieve LABEL and adds LABEL to expectedParams
    Label label = expect(LABEL);
    ...
    // Retrieve SUPER_CONCEPT and adds it to expectedParams
    OntologyConcept superConcept = expect(SUPER_CONCEPT);
    ...
    return graph.putEntityType(label).sup(superConcept.asEntityType());
    ...
}
```

Then at the end we check that `params` only contains things that are in `expectedParams`.

Maybe I should add a similar example to the code?

grakn-graql/src/main/java/ai/grakn/graql/internal/query/InsertQueryExecutor.java
aelred: Agreed! I'll add some comments to the first usage describing how it helps.

grakn-core/src/main/java/ai/grakn/util/CommonUtil.java
aelred: I've re-discovered the incredible `for` loop, so I'll get rid of this scary method.

grakn-core/src/main/java/ai/grakn/util/ErrorMessage.java
fppt: >  they've all been deleted and replaced with more general messages

Moved and improved then :)

grakn-graql/src/main/java/ai/grakn/graql/internal/query/ConceptBuilder.java
fppt: Yes. I think there is value in that.

aelred:
@kasper-piskorski this PR has caused some reasoner failures. I tried to investigate, but couldn't figure it out. This PR should be only a refactor so it could be a bug in the new insert query implementation.

```
Failed tests: 
  RecursiveInferenceTest.testReachability:275->assertQueriesEqual:556 expected:<ai.grakn.graql.internal.reasoner.query.QueryAnswers@20fb75d> but was:<ai.grakn.graql.internal.reasoner.query.QueryAnswers@2384c2b>
  RecursiveInferenceTest.testSameGeneration:243->assertQueriesEqual:556 expected:<ai.grakn.graql.internal.reasoner.query.QueryAnswers@2894e2> but was:<ai.grakn.graql.internal.reasoner.query.QueryAnswers@513066>
Tests in error: 
  RecursiveInferenceTest.testNguyen:356->queryAnswers:550 Â» NullPointer
  RecursiveInferenceTest.testReachabilitySymmetric:289->assertQueriesEqual:554->queryAnswers:550 Â» NullPointer
  RecursiveInferenceTest.testSameGenerationCao:371->assertQueriesEqual:554->queryAnswers:550 Â» GraqlQuery
```

kasper-piskorski:
@aelred  it's materialisation that is failing :)

aelred:
@kasper-piskorski that makes sense at least! So it's definitely because something about the insert behaviour has changed. My guess is either there's a bug, or we're being more strict about something than we were before.

kasper-piskorski:
@aelred  it seems like it is losing some variable mappings and then things go south, I will investigate

aelred:
@fppt I think I've made all the changes.

It would be good if you could re-review the last commit especially. I'm not particularly happy with it. It's to handle the special case that allows you to set the super on an existing concept.

grakn-core/src/main/java/ai/grakn/util/CommonUtil.java
```diff
@@ -162,47 +160,4 @@ public static RuntimeException unreachableStatement(@Nullable String message, Th
...
         };
     }
-
-    public static <T, K, V> Collector<T, ?, HashMultimap<K, V>> flatteningToMultimap(
-            Function<? super T, ? extends K> keyFunction,
-            Function<? super T, ? extends Stream<? extends V>> valueFunction) {
-        return flatteningToMultimap(keyFunction, valueFunction, HashMultimap::create);
-    }
-
-    public static <T, K, V, M extends Multimap<K, V>> Collector<T, ?, M> flatteningToMultimap(
```
fppt: yay!

grakn-core/src/main/java/ai/grakn/exception/GraqlQueryException.java
```diff
@@ -138,22 +138,47 @@ public static GraqlQueryException insertMetaType(Label label, OntologyConcept on
         return new GraqlQueryException(ErrorMessage.INSERT_METATYPE.getMessage(label, ontologyConcept.getLabel()));
     }
 
+    /**
+     * Thrown when a concept is inserted with multiple properties when it can only have one.
```
fppt: yay you took my suggestion! 👍

grakn-graql/src/main/java/ai/grakn/graql/internal/query/ConceptBuilder.java
```diff
@@ -90,14 +90,38 @@
     private final Map<BuilderParam<?>, Object> params = new HashMap<>();
```
fppt: I think I realised why this change it confusing to follow. Maybe this variable can be renamed to `expectedParams`. My justification for this is the following:

```
params.forEach((param, value) -> {
    if (!usedParams.contains(param)) {
        throw GraqlQueryException.insertUnexpectedProperty(param.name(), value, concept);
    }
});
```

This will make more sense as:

```
expectedParams.forEach((expectedParam, value) -> {
    if (!usedParams.contains(expectedParam)) {
        throw GraqlQueryException.insertUnexpectedProperty(param.name(), value, concept);
    }
});
```

Then it's clear that you used something that was not expected so BOOM. What do you think?

fppt: This would also mean `expect(BuilderParam)` would need to turn into `used(BuilderParam)` which i think is a good change anyway.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/ConceptBuilder.java
```diff
@@ -314,21 +349,47 @@ private OntologyConcept putOntologyConcept() {
...
+
+        return concept;
     }
 
+    /**
+     * Make the second argument the super of the first argument
+     *
+     * @throws GraqlQueryException if the types are different, or setting the super to be a meta-type
+     */
+    public static void setSuper(OntologyConcept subConcept, OntologyConcept superConcept) {
```
fppt: Oh bugger. This is done because the setters are only on the leaf interfaces . . . I am wondering if maybe graql can start getting access to some internal methods? At the end of the day this setter is actually implemented in `OntologyConceptImpl`

grakn-graql/src/main/java/ai/grakn/graql/internal/query/InsertQueryExecutor.java
```diff
@@ -116,9 +116,13 @@ private static InsertQueryExecutor create(Collection<VarPatternAdmin> patterns,
 
             For example, the property `$x isa $y` depends on the existence of the concept represented by `$y`.
          */
+        Multimap<VarAndProperty, Var> propDependencies = HashMultimap.create();
 
-        Multimap<VarAndProperty, Var> propDependencies =
-                properties.stream().collect(flatteningToMultimap(identity(), VarAndProperty::requiredVars));
+        for (VarAndProperty property : properties) {
```
fppt: 👍

grakn-graql/src/main/java/ai/grakn/graql/internal/query/InsertQueryExecutor.java
```diff
@@ -276,25 +301,48 @@ private Answer insertAll(Answer results) {
...
+     * <p>
+     *     If the concept has already been created, this will return empty.
+     * </p>
+     */
+    public Optional<ConceptBuilder> tryBuilder(Var var) {
         var = equivalentVars.componentOf(var);
 
         if (concepts.containsKey(var)) {
-            throw GraqlQueryException.insertExistingConcept(printableRepresentation(var), concepts.get(var));
+            return Optional.empty();
```
fppt: Does this mean that you just postponing throwing the exception now? Or is there a use case I missed where in one case you need to throw but in another you don't?

aelred: Yeah, the latter. In `tryBuilder` we return an `Optional<ConceptBuilder>`, so properties like `SubProperty` can change behaviour based on if the concept exists or not. But we still have `builder` that will throw.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/ConceptBuilder.java
aelred: I'm not 100% sure that's a good idea. I get that it means they are "expected to be used", but to me they mean something more general than that, like:
```
builder.isa(name).value("Bob");
```

This will make two parameters for `name` and `"Bob"`. It isn't immediately clear to me that these are "expected". It seems more like I have provided them. They are "expected to be used" is more of an afterthought to me. After all, if I provide them of course I expect them to be used :P

I kind of like `providedParams`, but not totally behind that idea either. Something needs to change if its confusing though!

grakn-graql/src/main/java/ai/grakn/graql/internal/query/ConceptBuilder.java
aelred: That sounds like a good idea! Such a method would need to handle being given invalid concepts though, like `name.sup(movie);` or whatever.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/ConceptBuilder.java
fppt: can we compromise to `preProvidedParams` then? At least then when there is a mismatch between the `usedParams` and `preProvidedParams` resulting in a throw it's a bit more clear why ? 

`params` says nothing. It's kind of one of the most generic names possible. Kind of like `fileds` :P

aelred:
!rtg - redis server error

aelred:
!rtg - redis server error

aelred:
Finally!

@fppt one more review s'il vous plait

fppt:
@aelred  your PR may have slowed things down quite a lot. If you look at the logs that went in just before this PR note the test times: https://travis-ci.org/graknlabs/grakn/builds/260953016?utm_source=github_status&utm_medium=notification.

It looks like this PR may have taken the test times from 20 to 40 mins.

PR #1951, opened 2017-08-03
Bug #16317: Fix randomly failing property test

fppt:
This test reproduces the edge case more simply:

```
@Test
    public void whenChangingTheSuperTypeOfAnEntityTypeWhichHasAResource_EnsureTheResourceIsStillAccessibleViaTheRelationTypeInstancesMethod(){
        ResourceType<String> name = graknGraph.putResourceType("name", ResourceType.DataType.STRING);

        //Create a person and allow person to have a name
        EntityType person = graknGraph.putEntityType("person").resource(name);

        //Create a man which is a person and is therefore allowed to have a name
        EntityType man = graknGraph.putEntityType("man").sup(person);
        RelationType has_name = graknGraph.putRelationType("has-name");

        //Create a Man and name him Bob
        Resource<String> nameBob = name.putResource("Bob");
        man.addEntity().resource(nameBob);

        //Get The Relation which says that our man is name bob
        Relation expectedEdge = Iterables.getOnlyElement(has_name.instances());

        assertThat(expectedEdge.type().instances(), hasItem(expectedEdge));

        //Man is no longer a person and therefore is not allowed to have a name
        man.sup(graknGraph.admin().getMetaEntityType());

        assertThat(expectedEdge.type().instances(), hasItem(expectedEdge));
    }
```

So the problem is that we traversing to the `Relation Edges` via the `plays` edge between `person` and `has-name-owner`. When you change the super type of `man` to no longer be `person` then you disconnecting the instances accidentally.

This is not the ideal solution but it's the "minimum" impact one. We need to start discussing traversing to edge Relations either via clever listing in the shards or more clever third party indexing.

### DIFF
--- a/grakn-graph/src/main/java/ai/grakn/graph/internal/concept/TypeImpl.java
+++ b/grakn-graph/src/main/java/ai/grakn/graph/internal/concept/TypeImpl.java
@@ -367,7 +367,7 @@ public class TypeImpl<T extends Type, V extends Thing> extends OntologyConceptIm
 
             //It is possible to be disconnecting from a role which is no longer in use but checking this will take too long
             //So we assume the role is in sure and throw if that is the case
-            if(!superPlays.isEmpty()){
+            if(!superPlays.isEmpty() && instancesDirect().findAny().isPresent()){
                 throw GraphOperationException.changingSuperWillDisconnectRole(oldSuperType, newSuperType, superPlays.iterator().next());
             }
 
--- a/grakn-graph/src/main/java/ai/grakn/graph/internal/concept/TypeImpl.java
+++ b/grakn-graph/src/main/java/ai/grakn/graph/internal/concept/TypeImpl.java
@@ -364,7 +364,7 @@ public class TypeImpl<T extends Type, V extends Thing> extends OntologyConceptIm
             Set<Role> superPlays = new HashSet<>(oldSuperType.plays());
 
             //Get everything that this can play bot including the supers
-            Set<Role> plays = directPlays().keySet();
+            Set<Role> plays = new HashSet<>(directPlays().keySet());
             subs().stream().flatMap(sub -> TypeImpl.from(sub).directPlays().keySet().stream()).
                     forEach(play -> plays.add((Role) play));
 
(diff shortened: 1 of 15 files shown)


grakn-graph/src/main/java/ai/grakn/graph/internal/concept/TypeImpl.java
```diff
@@ -348,6 +348,35 @@ public T plays(Role role) {
...
+            //It is possible to be disconnecting from a role which is no longer in use but checking this will take too long
+            //So we assume the role is in sure and throw if that is the case
+            if(!superPlays.isEmpty() && instancesDirect().findAny().isPresent()){
+                throw GraphOperationException.changingSuperWillDisconnectRole(oldSuperType, newSuperType, superPlays.iterator().next());
+            }
+
+            return true;
+        }
+        return changingSuperAllowed;
+    }
```
aelred: I think this is using methods like `directInstances` and `directPlays` when it shouldn't...

For example:
```
boy sub man;
man sub person;
person plays hungry;

$x isa boy;
(hungry: $x);
```

If we then go `boy sub rubiks-cube`, then the change should not be allowed because a `rubiks-cube` should not be hungry. We won't catch it because we only check `directPlays`.

A similar problem occurs if we go `man sub rubiks-cube`, because we only check `directInstances`.

grakn-test/src/test/java/ai/grakn/test/property/OntologyConceptPropertyTest.java
```diff
@@ -163,6 +162,8 @@ public void whenSettingTheDirectSuper_TheDirectSuperIsSet(
         assumeTrue(sameOntologyConcept(subConcept, superConcept));
         assumeThat((Collection<OntologyConcept>) subConcept.subs(), not(hasItem(superConcept)));
 
+        if(subConcept.isType()) assumeThat(subConcept.asType().sup().instances(), is(empty()));
```
aelred: I feel we should add the same TODO to do these, since this assumption is temporary.

fppt:
!rtg - ai.grakn.engine.controller.UserControllerTest.ai.grakn.engine.controller.UserControllerTest

fppt:
!rtg - More redis failures

fppt:
@aelred one more review please :)

PR #1958, opened 2017-08-04
Convert Graph API Collection Returns To Streams

fppt:


grakn-graph/src/main/java/ai/grakn/graph/internal/concept/OntologyConceptImpl.java
```diff
@@ -337,21 +329,17 @@ private boolean superLoops(){
...
     /**
      *
      * @return A collection of {@link Rule} for which this {@link OntologyConcept} serves as a conclusion
      */
     @Override
-    public Collection<Rule> getRulesOfConclusion() {
-        Set<Rule> rules = new HashSet<>();
-        neighbours(Direction.IN, Schema.EdgeLabel.CONCLUSION).forEach(concept -> rules.add(concept.asRule()));
-        return Collections.unmodifiableCollection(rules);
+    public Stream<Rule> getRulesOfConclusion() {
```
pluraliseseverythings: The logic looks a bit different, maybe double check this

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationAtom.java
```diff
@@ -225,7 +225,7 @@ public boolean isAllowedToFormRuleHead(){
         }
 
         //check roles are ok
-        Collection<Role> possibleRoles = type != null? type.asRelationType().relates() : Collections.EMPTY_SET;
+        Set<Role> possibleRoles = type != null? type.asRelationType().relates().collect(toSet()) : Collections.EMPTY_SET;
```
pluraliseseverythings: You can keep this a s a stream too, it's only used with a contains

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationAtom.java
```diff
@@ -550,7 +550,7 @@ private RelationAtom inferRoleTypes(){
         //remaining roles
         //role types can repeat so no matter what has been allocated still the full spectrum of possibilities is present
         //TODO make restrictions based on cardinality constraints
-        Set<Role> possibleRoles = Sets.newHashSet(relType.relates());
+        Set<Role> possibleRoles = relType.relates().collect(toSet());
```
pluraliseseverythings: This too, the getOntologyConcept method can take a stream

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationAtom.java
```diff
@@ -662,7 +662,7 @@ private RelationAtom inferRoleTypes(){
         Map<Var, OntologyConcept> parentVarOntologyConceptMap = parentAtom.getParentQuery().getVarOntologyConceptMap();
         Map<Var, OntologyConcept> childVarOntologyConceptMap = this.getParentQuery().getVarOntologyConceptMap();
 
-        Set<Role> relationRoles = new HashSet<>(getOntologyConcept().asRelationType().relates());
+        Set<Role> relationRoles = getOntologyConcept().asRelationType().relates().collect(toSet());
```
pluraliseseverythings: And this

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationAtom.java
fppt: I think I tried to but I was not able to do so. Will try again. Maybe It was something silly I was doing during the original attempt to change everything.

grakn-graph/src/main/java/ai/grakn/graph/internal/concept/OntologyConceptImpl.java
fppt: This is correct. Just much less verbose. This does the exact same thing as before (minus the sets). Basically it's shorter because:

1. `neighbours` already uses a generic to determine the return type. So `concept.asRule()` is not needed.
2. This means that "manually" iterating over the stream and adding it to the new set is not needed either. 

If you look above the exact same change happened for `getRulesOfHypothesis ()`

PR #1959, opened 2017-08-04
Make `VarProperty` implementations use autovalue

aelred:


grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/HasResourceTypeProperty.java
```diff
@@ -63,93 +64,87 @@
...
-        this.relationOwner = relationType.relates(this.ownerRole).admin();
-        this.relationValue = relationType.admin().getVarName().relates(this.valueRole).admin();
-
-    }
+        VarPatternAdmin relationOwner = relationType.relates(ownerRole).admin();
+        VarPatternAdmin relationValue = relationType.admin().getVarName().relates(valueRole).admin();
 
-    public VarPatternAdmin getResourceType() {
-        return resourceType;
+        return new AutoValue_HasResourceTypeProperty(
```
fppt: This makes me wonder if we have gone too far

grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/PlaysProperty.java
```diff
@@ -71,63 +69,45 @@ public String getName() {
...
 
     @Override
     public void delete(GraknGraph graph, Concept concept) {
-        Label roleLabel = role.getTypeLabel().orElseThrow(() -> GraqlQueryException.failDelete(this));
+        Label roleLabel = role().getTypeLabel().orElseThrow(() -> GraqlQueryException.failDelete(this));
         concept.asType().deletePlays(graph.getOntologyConcept(roleLabel));
     }
 
     @Override
-    public boolean equals(Object o) {
```
fppt: This really is one of the nicest side effects of Autovalue

grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/HasResourceTypeProperty.java
aelred: True! I think the issue is mostly just that this class has way more fields than necessary...

PR #1960, opened 2017-08-04
Miscellaneous fixes

aelred:
Some fixes for edge cases that were found while writing property tests for Graql.


aelred:
@duckofyork this PR is failing for query-planning reasons. It's probably my fault, but if you like you can try and fix it next week :P

Otherwise I will fix it when I get back.

PR #1964, opened 2017-08-07
Added more metrics and cleaned up closure

pluraliseseverythings:


grakn-engine/src/test/java/ai/grakn/engine/data/RedisWrapperTest.java
```diff
@@ -39,6 +39,22 @@ public void whenBuildingNoSentinelWellFormed_Succeeds() {
         assertTrue(redisWrapper.getJedisPool() != null);
     }
 
+    @Test
+    public void whenBuildingFromStringWithPort_Succeeds() {
+        RedisWrapper redisWrapper = RedisWrapper.builder().setUseSentinel(false)
+                .addURI("localhost:2345")
+                .build();
+        assertTrue(redisWrapper.getJedisPool() != null);
```
fppt: This is a pathetic poke I know but why not `assertNotNull(redisWrapper.getJedisPool())` the only reason for this is that it provides slightly more context if the test fails.

grakn-engine/src/test/java/ai/grakn/engine/data/RedisWrapperTest.java
```diff
@@ -39,6 +39,22 @@ public void whenBuildingNoSentinelWellFormed_Succeeds() {
...
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void whenBuildingFromStringWithDoublePort_Fails() {
+        RedisWrapper redisWrapper = RedisWrapper.builder().setUseSentinel(false)
+                .addURI("localhost:2345:5678")
+                .build();
+        assertTrue(redisWrapper.getJedisPool() != null);
```
fppt: Same for here.

PR #1966, opened 2017-08-08
Distribution simplification

lolski:
Changes:
- simplified grakn distribution folder
- simplified grakn executable. now we don't have `grakn-redis.sh`, `grakn-engine.sh`, `migration.sh` et cetera. we only have `grakn` and `graql`.
- simplified console output so it's more elegant. 
- changed the shutdown order meaning we don't have those Jedis exception anymore

Compile and extract it as such
- `cd <grakn base directory>`
- `mvn package -T 2.5C -DskipTests=true`
- ` cd grakn-dist/target`
- `tar -xf grakn-dist-0.17.0-SNAPSHOT.tar.gz`
- `cd grakn-dist-0.17.0-SNAPSHOT`

And try the various commands:
```
./grakn
./grakn server start
./grakn server start [cassandra|redis|grakn]
./grakn help
./grakn server help
./grakn version

./graql console [options]
./graql migrate [options]
```


burukuru:
Package looks quite neat now. I noticed we are not including the license, readme or changelog anywhere. Do we want to keep those?

Also I think grakn/graql files should be symlinks instead of bash scripts wrapping the other scripts.

lolski:
@burukuru 
1. we are not including them anymore. the idea is that user can just visit our page if they need them
2. if you try to run them you will see that `./grakn` does not merely call `bin/grakn.sh`. instead, it has its own responsibility which is to unify various "sub" scripts plus providing helpful messages. 

would this model work with your idea of symlinking? if so, perhaps you can describe your approach and maybe give a few points why it is better (i.e., if it leads to a more testable code, perhaps?)

lolski:
!rtg

burukuru:
!rtg

burukuru:
The old grakn script is already a wrapper so the new script is just an unnecessary layer of abstraction. This should be consolidated into one.

It would also help avoid cases where the parent script runs the child script without arguments and you get confusing help messages:
```
14:49 $ ./grakn server help
Usage: /home/thanh/grakn/lolski-grakn/grakn-dist/target/grakn-dist-1.0.0-SNAPSHOT/services/grakn.sh {start|stop|clean|status}
```

lolski:
@burukuru agreed, there are some cleanup i can do on this department. give me a couple of hours

lolski:
@burukuru done. now we only have `grakn` and `graql` script. and they're not that long either.

lolski:
@aelred true!

do we plan to keep both pre 1.0 and 1.0 docs (this change is meant for 1.0)? or can we update it?

lolski:
@burukuru also, do we support distribution package such as brew or apt-get?

fppt:
We only plan on keeping 1.0 docs. For the moment you can just update the docs on this PR to reflect the changes. 

> do we support distribution package such as brew or apt-get?

Only brew right now.

aelred:
@lolski we can update it right in this PR, no need to keep the old docs! The docs are only meant to replace the ones on the site once we actually make a release (although that wasn't true this week :P ).

lolski:
@fppt @burukuru On another note, I would opt for removing the `clean` command before production. 

Having such `rm -rf` esque command right in the front door, even when asking for a confirmation prior, invites people to shoot themselves.

Imagine a mistake where an unsuspecting user tried to invoke some harmless command from the bash history (Ctrl+R), but the `clean` command was accidentally pulled out due to a typo

lolski:
Also, the docs has been updated! @aelred

docs/_jekyll/.bundle/config
```diff
@@ -0,0 +1,3 @@
+---
+BUNDLE_PATH: "/Users/lolski/.bundle"
```
burukuru: These files should not be here.

grakn-dist/src/dist.xml
```diff
@@ -28,82 +28,46 @@
...
-            <directory>${basedir}/src/conf/</directory>
+            <directory>${main.basedir}/conf/main</directory>
             <includes>
-                <include>**</include>
+                <include>grakn.properties</include>
             </includes>
         </fileSet>
+        <!-- and logback.xml (main config) -->
         <fileSet>
-            <outputDirectory>examples</outputDirectory>
```
burukuru: I think we want to still keep the examples in the dist for the quickstart. @mikonapoli @fppt

grakn-dist/src/dist.xml
```diff
@@ -28,82 +28,46 @@
...
+                <include>README.md</include>
             </includes>
-            <excludes>
-                <exclude>**/target/</exclude>
-                <exclude>grakn-dashboard/node_modules/</exclude>
-                <exclude>grakn-dashboard/static/dashboard.js</exclude>
-            </excludes>
         </fileSet>
         <fileSet>
-            <outputDirectory>docs/api</outputDirectory>
```
burukuru: do we want to stop including docs with the dist? @fppt @aelred

grakn-dist/src/dist.xml
lolski: IMHO, people would normally just go to the website or google for a tutorial.

docs/_jekyll/.bundle/config
lolski: good find.

lolski: additionally, i also removed a .ruby-version which i have carelessly added as well

grakn-dist/src/dist.xml
lolski: again I think, people would normally just go to the website or google for a tutorial. Let's wait for the others to chime in as well

grakn-dist/src/dist.xml
aelred: I find these really helpful for quickly checking stuff works, like `graql.sh -f examples/pokemon.gql`. I don't know how useful they are to users though.

burukuru: I use them personally that way too and would say that it removes the additional step of downloading the examples for new users.

grakn-dist/src/dist.xml
burukuru: It will be useful for people running older versions of the software post-1.0.
Including docs mitigates against things like javadoc.io being unavailable, especially as we're not including the full source code with the dist anymore.

Also the REST API documentation is only available on our website and I pull it out of the dist.

Forgot to tag @pluraliseseverythings.

grakn-dist/src/graql
```diff
@@ -0,0 +1,96 @@
...
+
+# =============================================
+# main routine
+# =============================================
+
+update_classpath
+
+case "$1" in
+  console)
+    graql_console
```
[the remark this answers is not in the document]
lolski: yes we have that for graql and for migration, they can be invoked via `graql console help` and `graql migrate help`, respectively.

Different than `grakn`, the help function for graql and migration are already built-in in the Java program since way back (not very consistent, but I guess it'll do for now)

grakn-dist/src/services/cassandra/cassandra-env.sh
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
lolski: done

grakn-dist/src/test/bash/bash-startup-test.sh
```diff
@@ -1,24 +1,30 @@
 #!/bin/bash
 
+# Requirements
+# shunit2
+# distribution tar under target
+
 BASEDIR=$(dirname "$0")
 GRAKN_DIST_TARGET=$BASEDIR/../../../target/
 GRAKN_DIST_TMP=$GRAKN_DIST_TARGET/grakn-bash-test/
 GRAKN_DIST_BIN=${GRAKN_DIST_TMP}/bin/
```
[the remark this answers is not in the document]
lolski: done

grakn-dist/src/test/bash/bash-startup-test.sh
```diff
@@ -27,20 +33,21 @@ oneTimeSetUp() {
 }
 
 oneTimeTearDown() {
-  echo "y" | "${GRAKN_DIST_BIN}"/grakn.sh clean
-  "${GRAKN_DIST_BIN}"/grakn.sh stop
+
```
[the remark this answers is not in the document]
lolski: done

grakn-dist/src/dist.xml
lolski: Cool. so, to include, or not to include?

Current standing:
- don't include: Ganesh
- include: Felix, Thanh

grakn-dist/src/dist.xml
lolski: Need more opinion

Current standing:
1. Don't include: Ganesh, Felix
2. Include: Thanh

fppt: I personally have never looked at the java docs which come with the package so I say remove it. If you using the java API your IDE will likely show you the docs before you even think of checking the distribution.

pluraliseseverythings: It's definitely useful if we only ever keep the last docs online. If we delete this we need to be able to access docs as grakn.ai/docs/1.0.0/ or grakn.ai/docs/0.15.0

grakn-dist/src/dist.xml
fppt: I fall on the include side here. I find this useful for testing and with a little more jiggering it can be made useful for users as well. Kind of how the first thing most people learning Tinkerpop do is load their embedded example. +1 Include.

pluraliseseverythings: I'm for including them, we can reference them in the tutorials/docs and we know people already have them

grakn-dist/src/dist.xml
aelred: @pluraliseseverythings this is only removing the javadocs, which actually _are_ versioned (to my surprise :P )

see: https://static.javadoc.io/ai.grakn/grakn/0.12.0/overview-summary.html

fppt: yes our javadocs are already online and versioned.

pluraliseseverythings: Let's kill them then

grakn-dist/src/dist.xml
lolski: OK, I'll put it back in if it's useful

grakn-dist/src/dist.xml
lolski: Got it

lolski:
@fppt @burukuru @aelred @pluraliseseverythings done! including examples back, and excluding docs. any other reviews / changes?

lolski:
!rtg

lolski:
!rtg

PR #1969, opened 2017-08-08
Use single traversal for getting resources of entity

fppt:
This should be slightly faster.

### DIFF
--- a/grakn-graph/src/main/java/ai/grakn/graph/internal/concept/ThingImpl.java
+++ b/grakn-graph/src/main/java/ai/grakn/graph/internal/concept/ThingImpl.java
@@ -38,6 +38,7 @@ import ai.grakn.graph.internal.structure.VertexElement;
 import ai.grakn.util.Schema;
 import org.apache.tinkerpop.gremlin.process.traversal.P;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.__;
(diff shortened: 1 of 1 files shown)


fppt:
!rtg

fppt:
!rtg

fppt:
!rtg - Incorrect version of Janus Used

grakn-graph/src/main/java/ai/grakn/graph/internal/concept/ThingImpl.java
```diff
@@ -38,6 +38,7 @@
 import ai.grakn.util.Schema;
 import org.apache.tinkerpop.gremlin.process.traversal.P;
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal;
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.__;
```
pluraliseseverythings: wut? __?

fppt: :) 

So in the `TinkerPop` world you have the ability to creates `Traversals`. Usually you start these things from a graph like so:

```
graph.traversal().V()........
```

Now `Tinkerpop` also enables you to preform more complex traversals like:

```
graph.union(traversal1, traversal2).out()........
```

The `union` step is made up of two separate traversals. So how do you create those traversals? Well you use the weird shorthand `__.V().in()......`. This also lets you create traversals which can be passed around and used on different graphs. It's quite handy actually. I just hate that namespace `__`

PR #1978, opened 2017-08-09
Duplicate relations via Keys

fppt:
This is a bit of a naive implementation probably not covering some edge cases but its a start to see how usable it is.

### DIFF
--- a/grakn-graph/src/main/java/ai/grakn/graph/internal/concept/RelationReified.java
+++ b/grakn-graph/src/main/java/ai/grakn/graph/internal/concept/RelationReified.java
@@ -27,6 +27,7 @@ import ai.grakn.exception.GraphOperationException;
 import ai.grakn.graph.internal.structure.Casting;
 import ai.grakn.graph.internal.structure.VertexElement;
 import ai.grakn.util.Schema;
+import ai.grakn.util.StringUtil;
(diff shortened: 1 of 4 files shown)


fppt:
!rtg - ai.grakn.client.BatchMutatorClientTest.whenSending50InsertQueries_50EntitiesAreLoadedIntoGraph

grakn-graph/src/main/java/ai/grakn/graph/internal/concept/RelationReified.java
```diff
@@ -89,34 +92,60 @@ public void addRolePlayer(Role role, Thing thing) {
...
     }
 
     /**
+     * Creates a hash for a relation based on it's {@link RelationType} and the {@link Resource} which serves as it's key
+     *
+     * @param relationType the {@link RelationType} of the {@link Relation}
+     * @param resourceMap a sorted map of {@link ai.grakn.concept.ResourceType} Ids to {@link Resource} Ids
+     * @return A unique hash identifying this {@link Relation}
+     */
+    public static String generateNewHash(RelationType relationType, TreeMap<String, String> resourceMap){
```
fppt: Good questions. It's actually an id composed of several other ids. I think I landed on using the word "hash" because "id" and "Index" were already taken in different contexts. Maybe "InternalID" would be more clear?

PR #1981, opened 2017-08-10
[Stable] Get rid of javafx pair

kasper-piskorski:


fppt:
@kasper-piskorski please merge this into master

PR #1994, opened 2017-08-14
Fix performance issue with insert queries

aelred:
This was because my previous method to find equivalent properties was O(n^2), now it should be O(n).


aelred:
!rtg - sad time out :(

PR #1997, opened 2017-08-15
Refactors part 2: OntologyConcept and Relation Refactored

fppt:


fppt:
!rtg - timeout

fppt:
!rtg - redis failure

PR #1999, opened 2017-08-16
Make migrator client print any errors

aelred:


grakn-core/src/main/java/ai/grakn/util/REST.java
```diff
@@ -198,6 +198,13 @@
             public static final String RELATIONS_JSON_FIELD = "relations";
             public static final String RESOURCES_JSON_FIELD = "resources";
         }
+
+        /**
+         * Json fields used to describe tasks
+         */
+        public static class Task {
```
fppt: I have a sneaking suspicion that either this is missing keys or this should not exist at all. Surely we have many other keys on task related json representations?

grakn-engine/src/main/java/ai/grakn/engine/controller/TasksController.java
```diff
@@ -430,7 +431,7 @@ private Json serialiseStateFull(TaskState state) {
                 .set("interval", state.schedule().interval().map(Duration::toMillis).orElse(null))
                 .set("recurring", state.schedule().isRecurring())
                 .set("exception", state.exception())
```
fppt: Oooo here they are hiding. As hard coded strings???? FLOGGINGS!!!!!! _starts foaming at the mouth_

aelred: Damn it, thought I'd got away with this one...

PR #2000, opened 2017-08-16
Refactor old terminology mixing `Var` and `VarPattern`

aelred:


aelred:
Nah, it's not back and forth. This is just stuff I didn't find when I first made this terminology change. We've always been going in one direction for this one :P

PR #2009, opened 2017-08-21
Get rid of scope

fppt:


grakn-graql/src/test/java/ai/grakn/graql/internal/query/AutocompleteTest.java
```diff
@@ -71,9 +71,9 @@ public void whenAutocompletingWithinQuery_AppropriateCandidatesAreFound() {
     public void whenAutocompletingWithinWord_AppropriateCandidatesAreFound() {
         String queryString = "match $x has title";
         Autocomplete autocomplete = Autocomplete.create(types, queryString, 11);
-        assertThat(autocomplete.getCandidates(), hasItem("has"));
+        assertThat(autocomplete.getCandidates(), hasItem(" "));//Nothing returned because only "has" was found
         assertThat(autocomplete.getCandidates(), not(hasItem("delete")));
-        assertEquals(9, autocomplete.getCursorPosition());
+        assertEquals(11, autocomplete.getCursorPosition());
```
aelred: I feel like this test should have passed unchanged. I'm going to pull down and check.

fppt: It may be a long standing logic error. Checkout: `Autocomplete#findCandidates`, you have this check in place:

```
 if (candidates.size() == 1 && candidates.iterator().next().equals(token.getText())) {
                        return ImmutableSet.of(" ");
}
```

that's why I put my comment

PR #2012, opened 2017-08-21
Simplify `delete` query to only accept sets of variables

aelred:


aelred:
!rtg - something went wrong?

fppt:
@aelred conflicts need resolving please

PR #2013, opened 2017-08-21
Refactor `VarPropertyInternal`

aelred:
This is to allow us to introduce `undefine` queries, using the same `InsertQueryExecutor` (now `QueryOperationExecutor`).

### DIFF
--- a/grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/IdProperty.java
+++ b/grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/IdProperty.java
@@ -83,16 +83,6 @@ public abstract class IdProperty extends AbstractVarProperty implements NamedPro
     }
 
     @Override
-    public Set<Var> requiredVars(Var var) {
-        return ImmutableSet.of();
-    }
-
-    @Override
-    public Set<Var> producedVars(Var var) {
-        return ImmutableSet.of(var);
-    }
-
-    @Override
     public boolean uniquelyIdentifiesConcept() {
         return true;
     }
--- a/grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/IdProperty.java
+++ b/grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/IdProperty.java
@@ -71,9 +71,11 @@ public abstract class IdProperty extends AbstractVarProperty implements NamedPro
 
     @Override
     public PropertyExecutor insert(Var var) throws GraqlQueryException {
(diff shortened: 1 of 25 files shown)


aelred:
@fppt this is pretty WIP. I'd appreciate any suggestions

grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/IdProperty.java
```diff
@@ -71,24 +70,16 @@ public String getProperty() {
     }
 
     @Override
-    public void insert(Var var, InsertQueryExecutor executor) throws GraqlQueryException {
-        executor.builder(var).id(id());
+    public PropertyExecutor insert(Var var) throws GraqlQueryException {
```
fppt: You know after seeing this over and over again I am wondering if this abstraction, while handy, will come to bite us during debugging. This maybe insanely verbose but maybe it will be better for `PropertyExecutor.builder(....)` to accept an specific interface rather than a generic consumer? That way we can have explicit implementations of that interface which might be easier to debug.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/QueryOperationExecutor.java
```diff
@@ -424,35 +428,35 @@ GraknTx tx() {
...
-        }
-
-        Set<Var> requiredVars() {
-            return property().requiredVars(var());
-        }
-
-        Set<Var> producedVars() {
-            return property().producedVars(var());
+        private PropertyExecutor executor(ExecutionType executionType) {
+            switch (executionType) {
```
fppt: Ughhhh . . . . I hate switch statements with the obligatory "Cannot reach this line because java". 

maybe it would be better to turn `ExecutionType` into a more complex enum, one which provides the relevant property rather than relying on a switch.

I am thinking you make execution type be:

```
private enum ExecutionType {
    INSERT((var) -> property().insert(var)), 
    DEFINE((var) -> property(). define(var))

    private Function<Var, PropertyExecutor > function;

    public ExecutionType(Function<Var, PropertyExecutor > function){
        this.function = function;
    }

    public property(Var var){
        return function.apply(var);
    }
}
```
This is A LOT MORE VEROSE but it kills the ugly switch and it makes sure that if you add another `ExecutionType` you are forced to define its `PropertyExecutor`

aelred: Agreed. I've now learned you can add abstract methods to enums so I'll do that 👍

PR #2016, opened 2017-08-22
Remove attributes from types and things

fppt:


grakn-core/src/main/java/ai/grakn/concept/Attribute.java
```diff
@@ -91,6 +91,14 @@
     @Override
     Attribute attribute(Attribute attribute);
 
+    /**
+     * Removes the provided {@link Attribute} from this {@link Attribute}
+     * @param attribute the {@link Attribute} to be removed
+     * @return The {@link Attribute} itself
+     */
+    @Override
+    Attribute deleteAttribute(Attribute attribute);
```
aelred: I'm not 100% sure on this, but I think if you skip the javadoc here, it will get inherited from the super-type (same for `Entity` etc.

grakn-kb/src/main/java/ai/grakn/kb/internal/concept/ThingImpl.java
```diff
@@ -258,6 +260,35 @@ public T attribute(Attribute attribute){
...
+     * Returns a stream computed from a <code>streamProvider</code> which takes in as input an array or {@link Role}s
+     * which cannot contain any nulls.
+     *
+     * This helper exists to filter away the nulls
+     */
+    private <X> Stream<X> validStreamOf(Function<Role[], Stream<X>> streamProvider, Role ... roles){
+        if(roles.length == 0) return Stream.empty();
+        Role[] filteredRoles = Arrays.stream(roles).filter(Objects::nonNull).toArray(Role[]::new);
+        return streamProvider.apply(filteredRoles);
+    }
```
aelred: I wouldn't bother passing the lambda in here, just make a method `Role[] filterNullls(Role... roles)`:
```
Stream<Relationship> relationships = relationships(filterNulls(roleHasOwner, roleKeyOwner));
```

grakn-core/src/main/java/ai/grakn/concept/Attribute.java
fppt: Let's leave them there for the moment. Can delete them more easily than add them back in.

grakn-kb/src/main/java/ai/grakn/kb/internal/concept/ThingImpl.java
fppt: Alrighty.

fppt:
@aelred re-review please

aelred:
!rtg - something went wrong?

PR #2018, opened 2017-08-22
Make shell return a non-zero exit code on error

aelred:


fppt:
@aelred please resolve conflicts

aelred:
@fppt I re-merged but it looks like the continuous integration step has broken :(

PR #2029, opened 2017-08-25
Fix bad link in docs

aelred:


aelred:
!rtg - something bad happened

PR #2030, opened 2017-08-25
Added FB pixel to header of Documentation

tomassabat:
I added the Facebook tracking pixel to the header of the Documentation. This is necessary to enable tracking 'events' in the FB Ads manager.


haikalpribadi:
Can one of the admins verify this patch?

astofsel:
This is Tomas Sabat is by the way :)

PR #2032, opened 2017-08-25
[Stable] Use rule graph for rule prioritisation (Bug #16749) - DO NOT MERGE YET 

kasper-piskorski:


fppt:
@kasper-piskorski I suck at following instructions and merged this in accidentally and then pulled it back out. Could you reopen this PR please?

PR #2035, opened 2017-08-29
[Stable] Use rule graph for rule prioritisation (Bug #16749) 

kasper-piskorski:


kasper-piskorski:
merge here: https://github.com/graknlabs/grakn/pull/2041

PR #2036, opened 2017-08-29
[Stable] Add neq predicate state to accommodate recursive queries with neq predicates

kasper-piskorski:


fppt:
@kasper-piskorski please merge this into master

PR #2042, opened 2017-08-30
Only trigger validation dataset on master/stable.

burukuru:
Skip validation on PRs.


fppt:
@burukuru I think this PR should go into stable branch?

burukuru:
I agree.

PR #2048, opened 2017-08-31
Introduce `get` query

aelred:
`match` alone is no longer a valid query!


grakn-graql/src/test/java/ai/grakn/graql/internal/query/match/MatchModifierTest.java
```diff
@@ -70,55 +70,38 @@ public void setUp() {
...
                         y.isa("genre").has("name", "crime")
                 )
-        ).orderBy("v", desc);
+        ).orderBy("v", desc).get();
 
         assertThat(query, variable(x, contains(MovieMatchers.godfather, MovieMatchers.godfather, MovieMatchers.apocalypseNow)));
     }
 
     @Test
-    public void testOrPatternOrderByUnselected() {
```
fppt: Why is this not a relevant test anymore?

grakn-graql/src/test/java/ai/grakn/graql/internal/query/match/MatchTest.java
```diff
@@ -169,10 +170,9 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertThat;
-import static org.junit.Assert.assertTrue;
 
 @SuppressWarnings({"OptionalGetWithoutIsPresent", "unchecked"})
-public class MatchQueryTest {
+public class MatchTest {
```
fppt: Maybe this should be `GetQueryTest` ? Based on all the actual test below?

grakn-graql/src/test/java/ai/grakn/graql/internal/query/match/MatchModifierTest.java
aelred: Good point. I think I got confused about what it did. I'll try and add it back.

grakn-graql/src/test/java/ai/grakn/graql/internal/query/match/MatchTest.java
aelred: Maybe. The thing is, 99% of the behaviour we are testing is within the `match`. We only call `get` because stuff like the matchers and helper methods work only with a `GetQuery`. I'll try and make them more generic.

I should also try and split it into two tests: one small `GetQueryTest` and a big `MatchTest`.

aelred:
@fppt think I fixed all the stuff!

PR #2053, opened 2017-09-01
Add debug mode to migration.

fppt:
This mode makes it fail faster and log more direct info to the user.

### DIFF
--- a/grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
+++ b/grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
@@ -18,21 +18,6 @@
 
 package ai.grakn.client;
 
-import static ai.grakn.engine.TaskStatus.COMPLETED;
-import static ai.grakn.engine.TaskStatus.FAILED;
-import static ai.grakn.engine.TaskStatus.STOPPED;
-import static ai.grakn.util.ErrorMessage.READ_ONLY_QUERY;
-import static ai.grakn.util.REST.Request.BATCH_NUMBER;
-import static ai.grakn.util.REST.Request.KEYSPACE_PARAM;
-import static ai.grakn.util.REST.Request.TASK_LOADER_MUTATIONS;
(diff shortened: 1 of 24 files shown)


grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
```diff
@@ -223,15 +222,27 @@ private void sendQueriesWhenBatchLargerThanValue(int value) {
...
                 Thread.sleep(500);
             } catch (InterruptedException e) {
-                LOG.error(e.getMessage());
+                printError(e.getMessage());
+            }
+        }
+
+        for (CompletableFuture completableFuture : futures.values()) {
+            try {
+                //TODO: Figure out why a null sometimes gets put in here
```
aelred: haha what

grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
```diff
@@ -275,41 +286,36 @@ void sendQueriesToLoader(Collection<Query> queries){
...
+                .thenAcceptAsync(onCompletionOfTask);
 
+        // TODO: use an async client
+        futures.put(newStatus.hashCode(), newStatus);
     }
 
-    private void unblock(CompletableFuture<Json> status){
+    private void unblock(){
         blocker.release();
-        futures.remove(status.hashCode());
```
aelred: We should remove the futures at _some_ point, or we'll have a leak. I guess I'd do it in `waitToFinish` after `get`?

grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
fppt: Ooooo wait I did figure it out! Need to remove that comment and check

grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
fppt: Ah yes good point.

PR #2057, opened 2017-09-04
Fix syntax errors on SNB

fppt:


fppt:
!rtg - ai.grakn.client.BatchMutatorClientTest.whenSending50InsertQueries_50EntitiesAreLoadedIntoGraph

PR #2060, opened 2017-09-04
Update Guava version

aelred:


aelred:
This is still being held back by our old Cassandra dependency.

PR #2061, opened 2017-09-04
Recommend `define` or `insert` in appropriate cases

aelred:


aelred:
!rtg - a good old failure for no reason

PR #2063, opened 2017-09-04
Query planner optimisation: combining node and edge fragment cost, improving greedy traversal

duckofyork:


duckofyork:
!rtg

duckofyork:
!rtg

duckofyork:
!rtg

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -257,6 +258,27 @@ private static void greedyTraversal(List<Fragment> plan, Arborescence<Node> arbo
...
+        }
+        return weight[0];
+    }
+
+    private static double nodeFragmentWeight(Node node) {
+        return node.getFragmentsWithoutDependency().stream()
+                .map(Fragment::fragmentCost).reduce(0D, (x, y) -> x + y) +
+                node.getFragmentsWithDependencyVisited().stream()
+                        .map(Fragment::fragmentCost).reduce(0D, (x, y) -> x + y) +
+                node.getFixedFragmentCost();
```
aelred: Streams have a method `count()` that does what you want here. Add some variables as well to make it a bit clearer :P

duckofyork: I guess you mean method `sum()`, as fragment cost is not 1

aelred: Oops! You're right. There is still a way to do it even neater:
`.mapToDouble(Fragment::fragmentCost).sum();`

duckofyork: I think `sum()` only supports integer stream.

duckofyork:
@aelred merge merge merge!

PR #2064, opened 2017-09-05
Config Cleanup and misdirection

fppt:


fppt:
Yes I have assigned lot of reviewers. This is because I ended up touching (and hiding) configs which nearly everyone uses.

grakn-dist/src/bin/grakn.in.sh
```diff
@@ -4,8 +4,8 @@ if [ -z "$FOREGROUND" ]; then
   export FOREGROUND=false
 fi
 
-# Set USE_CASSANDRA if factory.internal is Janus
-GRAKN_ENGINE_FACTORY=$(grep ^factory.internal= "${GRAKN_CONFIG}"| cut -d '=' -f 2)
-if [[ "$GRAKN_ENGINE_FACTORY" == "ai.grakn.factory.TxFactoryJanus" ]]; then
+# Set USE_CASSANDRA if knowledge-base.mode is Janus
```
aelred: if it's "production", not Janus.

grakn-factory/janus-factory/src/main/java/ai/grakn/factory/TxFactoryJanus.java
```diff
@@ -66,10 +68,28 @@
...
+            throw new RuntimeException(ErrorMessage.INVALID_PATH_TO_CONFIG.getMessage(DEFAULT_CONFIG), e);
+        }
+    }
+
+    //This maps the storage backend to the needed value
+    private static final Map<String, String> storageBackendMapper = new HashMap<>();
+    static{
+        //TODO: BIG TABLE WILL HAVE A VALUE HERE
+        storageBackendMapper.put("grakn-production", "cassandra");
+    }
```
aelred: Can we use an `ImmutableMap` instead?

grakn-factory/janus-factory/src/main/java/ai/grakn/factory/TxFactoryJanus.java
```diff
@@ -110,24 +130,27 @@ private synchronized JanusGraph newJanusGraph(String name, String address, Prope
...
+        //Load Passed in properties
+        properties.forEach((key, value) -> {
+
+            //Overwrite storage
+            if(key.equals(storageBackend)){
+                value = storageBackendMapper.get(value);
+            }
+
+            builder.set(key.toString(), value);
+        });
```
aelred: This kind of indirection is very confusing for both users and developers. However, if we really have to hide cassandra from the config file I suppose it's unavoidable.

grakn-factory/janus-factory/src/main/resources/default-configs.properties
```diff
@@ -0,0 +1,75 @@
...
+
+# TODO should be moved away from this file because it cannot be changed
+gremlin.hadoop.inputLocation=none
+
+###############################################################################
+############################# Spark Configuration #############################
+###############################################################################
+spark.master=local[*]
+spark.serializer=org.apache.spark.serializer.KryoSerializer
+spark.ui.showConsoleProgress=false
```
aelred: I like this! Removes some duplication.

grakn-kb/src/main/java/ai/grakn/factory/FactoryBuilder.java
```diff
@@ -43,16 +44,24 @@
...
+    public static final String KB_MODE = "knowledge-base.mode";
     private static final Map<String, TxFactory<?>> openFactories = new ConcurrentHashMap<>();
 
+    //This is used to map grakn value properties into the underlaying properties
+    private static final Map<String, String> factoryMapper = new HashMap<>();
+    static{
+        factoryMapper.put("in-memory", "ai.grakn.factory.TxFactoryTinker");
+        factoryMapper.put("production", "ai.grakn.factory.TxFactoryJanus");
+        factoryMapper.put("distributed", "ai.grakn.factory.TxFactoryJanusHadoop");
+    }
```
aelred: Another case for `ImmutableMap.of(..)`

fppt:
!rtg - did not start the first time

grakn-dist/src/bin/grakn.in.sh
fppt: Good spot

grakn-factory/janus-factory/src/main/java/ai/grakn/factory/TxFactoryJanus.java
fppt: YAR!!

grakn-factory/janus-factory/src/main/java/ai/grakn/factory/TxFactoryJanus.java
fppt: :(

grakn-factory/janus-factory/src/main/resources/default-configs.properties
fppt: :)

grakn-kb/src/main/java/ai/grakn/factory/FactoryBuilder.java
fppt: Yar!

fppt:
@aelred re-review please

fppt:
!rtg - cass timeout

fppt:
!rtg - ai.grakn.factory.JanusPreviousPropertyStepTest.ai.grakn.factory.JanusPreviousPropertyStepTest

grakn-factory/janus-factory/src/main/resources/default-configs.properties
```diff
@@ -0,0 +1,75 @@
...
+##############################################################################################
+############################# Configuration Needed For Analytics #############################
+##############################################################################################
+
+# Janus InputFormat configuration for using hadoop
+# Not clear why these need to be set. See http://stackoverflow.com/questions/38524151/counting-vertices-on-a-titan-graph-using-sparkgraphcomputer-throws-org-apache-sp/38529076
+# for more information.
+# These two properties should be equal to the value of storage.backend and storage.hostname, respectively.
+janusmr.ioformat.conf.storage.backend=cassandra
+janusmr.ioformat.conf.storage.hostname=localhost
```
burukuru: This needs to be the same as `storage.hostname` in grakn.properties.
Is there a way to capture `storage.hostname` and pass it to the layer below?
In any other scenario aside from running Grakn on your local PC, you will need to change this setting as well which makes the whole exercise moot.

fppt: Ahhh good spot you are correct. We can map and inject the values from the following:

```
storage.backend=grakn-production
storage.hostname=127.0.0.1
```

It will be weird but it will work.

aelred:
@fppt why do you make me sad

fppt:
@burukuru I have put in the injection. So the values of `storage.backend` and `storage.hostname` get injected into `janusmr.ioformat.conf.storage.backend` and `janusmr.ioformat.conf.storage.hostname` respectively. 

Can you re-review please?

PR #2069, opened 2017-09-06
Fix Metrics Dependency Conflicts 

fppt:


grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
```diff
@@ -138,7 +134,6 @@ public BatchMutatorClient(String keyspace, String uri, Consumer<TaskResult> onCo
             final ConsoleReporter reporter = ConsoleReporter.forRegistry(metricRegistry)
                     .convertRatesTo(TimeUnit.SECONDS)
                     .convertDurationsTo(TimeUnit.MILLISECONDS)
-                    .disabledMetricAttributes(ImmutableSet.of(P999, STDDEV, P98, P95, M5_RATE))
```
fppt: I did manage to sort out some of the dependency conflicts (we had about 3 different versions of `dropbox-metrics` running around) but after struggling for a long while I was unable to get rid of the yammer version `metrics-core` because Cass needs it.

In the end I had to surrender and delete the problematic line. So as the commit says . . . I am sorry :(

PR #2075, opened 2017-09-07
Dashboard fixes and updates for 1.0 release

marco-scoppetta:
- renamed strings `relation` to `relationship` and `resource` to `attribute`
- HAL builder now also returns the attribute dataType + test
- fixed low priority outstanding bugs in Dashboard
- Moved from Browserify to Webpack
- Production code in dashboard.js is now minified


fppt:
@marco-scoppetta maybe this should go into `stable` ?

fppt:
Also you committing `package-lock.json` here. We might want to gitignore that.

marco-scoppetta:
@fppt This contains some changes and small bug fixes, I can replicate the bug fixes to `stable`, but with all the other changes included I thought it was better to have it on `master` rather than `stable`. And also `package-lock.json` is better to have it committed so we lock the JS dependencies versions, it is usually committed in JS projects.

fppt:
The only reason I thought it might be worth moving to stable is because of:

> renamed strings relation to relationship and resource to attribute

Since stable already has these refactors it would make sense for this PR to be going against stable. Unless these are not public facing refactors. Then I am not that bothered.

With regards to `package-lock` thanks for clearing that up. I did not know it was the norm to commit that.

marco-scoppetta:
@fppt Ok I will make a new PR against `stable` and then merge it here.

PR #2082, opened 2017-09-08
Cleanup cache calls

kasper-piskorski:


kasper-piskorski:
!rtg

kasper-piskorski:
!rtg - timeout

kasper-piskorski:
!rtg - com.datastax.driver.core.exceptions.NoHostAvailableException: All host(s) tried for query failed (tried: localhost/127.0.0.1:9042 (com.datastax.driver.core.exceptions.OperationTimedOutException:

kasper-piskorski:
!rtg

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
```diff
@@ -93,8 +93,10 @@
 
     private int hashCode = 0;
     private Multimap<Role, Var> roleVarMap = null;
+    private Multimap<Role, SchemaConcept> roleTypeMap = null;
```
fppt: `roleMap` but meh you can fix it in another PR

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
```diff
@@ -324,14 +325,16 @@ public int computePriority(Set<Var> subbedVars) {
     }
 
     private Multimap<Role, SchemaConcept> getRoleTypeMap() {
```
fppt: `getRoleMap`. . .  but again, meh

PR #2086, opened 2017-09-11
Migration: Clear futures when failing make retry number configurable. 

fppt:


grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
```diff
@@ -139,10 +137,10 @@ public BatchMutatorClient(String keyspace, String uri, Consumer<TaskResult> onCo
...
      *
-     * @param retry boolean representing if engine should retry
+     * @param maxRetries The number of times to retry if a failure occurs
      */
-    public BatchMutatorClient setRetryPolicy(boolean retry){
-        this.retry = retry;
+    public BatchMutatorClient setRetryPolicy(int maxRetries){
+        this.maxRetries = maxRetries;
         return this;
     }
```
aelred: If this is intended to always be called, put it in the constructor 👍

PR #2089, opened 2017-09-12
Update to new syntax

marco-scoppetta:
- updated to new Graql syntax in dashboard and HALBuilder related classes


marco-scoppetta:
!rtg

PR #2093, opened 2017-09-12
Use Optionals To Handle and Filter Out Ghosts

fppt:


grakn-kb/src/main/java/ai/grakn/kb/internal/concept/RelationshipEdge.java
```diff
@@ -62,8 +63,15 @@
     private final Cache<Role> valueRole = new Cache<>(Cacheable.concept(), () -> edge().tx().getSchemaConcept(LabelId.of(
             edge().property(Schema.EdgeProperty.RELATIONSHIP_ROLE_VALUE_LABEL_ID))));
 
-    private final Cache<Thing> owner = new Cache<>(Cacheable.concept(), () -> edge().tx().factory().buildConcept(edge().source()));
-    private final Cache<Thing> value = new Cache<>(Cacheable.concept(), () -> edge().tx().factory().buildConcept(edge().target()));
+    private final Cache<Thing> owner = new Cache<>(Cacheable.concept(), () -> {
+        VertexElement vertexElement = edge().source().orElseThrow(() -> GraknTxOperationException.missingOwner(getId()));
+        return edge().tx().factory().<Thing>buildConcept(vertexElement).orElseThrow(() -> GraknTxOperationException.missingOwner(getId()));
```
aelred: Desperate attempt to improve the caches everywhere:
```
return edge().source().flatMap(edge().tx().factory()::buildConcept).orElseThrow(() ->
    GraknTxOperationException.missingOwner(getId())
);
```

(feel free to extract more variables to make it clearer or w/e)

grakn-kb/src/main/java/ai/grakn/kb/internal/concept/TypeImpl.java
```diff
@@ -241,7 +248,9 @@ boolean deletionAllowed(){
 
     Stream<V> instancesDirect(){
         return vertex().getEdgesOfType(Direction.IN, Schema.EdgeLabel.SHARD).
-                map(edge -> vertex().tx().factory().buildShard(edge.source())).
+                map(EdgeElement::source).
+                flatMap(CommonUtil::optionalToStream).
+                map(source -> vertex().tx().factory().buildShard(source)).
```
aelred: I _think_ these kinds of lambdas can be made into method references like this:
```
map(vertex().tx().factory()::buildShard)
```
(although remember `vertex().tx().factory()` would now be called outside the lambda)

fppt: It's can't in this case because you passing in `edge.source()` not `edge`

fppt:
!rtg

fppt:
!rtg - ai.grakn.test.engine.postprocessing.PostProcessingIT.checkThatDuplicateResourcesAtLargerScaleAreMerged

PR #2097, opened 2017-09-13
Added initial code for insert  micro benchmarks

pluraliseseverythings:


pluraliseseverythings:
!rtg

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -141,6 +139,9 @@ public static void main(String[] args) {
     }
 
     public void start() {
+        redisWrapper.testConnection();
+        LOG.info("Starting task manager {}", taskManager.getClass().getCanonicalName());
+        taskManager.start();
```
pluraliseseverythings: Moved start of task manager to start from constructor

pluraliseseverythings:
!rtg

Jenkinsfile
```diff
@@ -2,97 +2,100 @@
...
+                            sh '../grakn-test/test-snb/src/generate-SNB/load-SNB.sh arch validate'
+                        }
+                    }
+                    stage('Measure Size') {
+                        sh 'nodetool flush'
+                        sh 'du -hd 0 grakn-package/db/cassandra/data'
+                    }
+                }
+                timeout(360) {
+                    stage('Run the benchmarks') {
```
pluraliseseverythings: This is the only bit I added, the rest is formatting the indentation

grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
```diff
@@ -263,11 +256,12 @@ void sendQueriesToLoader(Collection<Query> queries){
         Retryer<TaskResult> sendQueryRetry = RetryerBuilder.<TaskResult>newBuilder()
                 .retryIfExceptionOfType(IOException.class)
                 .retryIfRuntimeException()
+                .retryIfResult(r -> r != null && r.getCode().startsWith("5"))
```
pluraliseseverythings: I'm fixing this awfulness next. I'll try to put a benchmark on REST and change the http client to something that doesn't require us to do the retry explicitly

grakn-test/test-integration/src/test/java/ai/grakn/test/engine/postprocessing/PostProcessingTaskTest.java
```diff
@@ -52,7 +52,7 @@
```
pluraliseseverythings: From here it's just renaming (that method doesn't actually start anything)

grakn-client/src/main/java/ai/grakn/client/BatchMutatorClient.java
aelred: thanks for apologising

PR #2098, opened 2017-09-13
Merge from stable

kasper-piskorski:


kasper-piskorski:
!rtg

grakn-core/src/main/java/ai/grakn/util/Schema.java
```diff
@@ -59,8 +59,8 @@ private Schema() {
         PLAYS("plays"),
         HYPOTHESIS("hypothesis"),
         CONCLUSION("conclusion"),
-        ROLE_PLAYER("role-player"),
-        RESOURCE("resource"),
+        SHORTCUT("shortcut"),
```
fppt: This should stay `ROLE_PLAYER`

grakn-factory/janus-factory/src/main/resources/indices-edges.properties
```diff
@@ -16,5 +16,5 @@
 # along with Grakn. If not, see <http://www.gnu.org/licenses/gpl.txt>.
 #
 
-role-player=RELATIONSHIP_TYPE_LABEL_ID,ROLE_LABEL_ID
-resource=RELATIONSHIP_TYPE_LABEL_ID
\ No newline at end of file
+shortcut=RELATIONSHIP_TYPE_LABEL_ID,ROLE_LABEL_ID
```
fppt: This should also stay role-player

fppt:
@kasper-piskorski you have some merge errors

PR #2101, opened 2017-09-14
Periodic commit log submission

fppt:
This improves the performance of insertions when committing


grakn-kb/src/main/java/ai/grakn/kb/internal/CommitLog.java
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Submits the commit logs to the provided server address and under the provided {@link Keyspace}
+     */
+    public String submit(String engineUri, Keyspace keyspace){
+        if(newInstanceCount.isEmpty() && newAttributes.isEmpty()){
+            return "Commit logs empty. No submission made";
+        }
+        String endPoint = getCommitLogEndPoint(engineUri, keyspace);
+        String response = EngineCommunicator.contactEngine(endPoint, REST.HttpConn.POST_METHOD, getFormattedLog().toString());
+        clear();
```
aelred: I think there might be a race condition here. We execute `submit` in one thread, but actually add to the log in another. Therefore, we can have:
- Thread 1 formats the log and starts sending it to server
- Thread 2 adds a new thing to the log
- Thread 1 clears the log (including the uncommitted thing)

grakn-kb/src/main/java/ai/grakn/kb/internal/CommitLog.java
```diff
@@ -47,16 +50,29 @@
...
+    /**
+     * Read locks are used when acquiring the data.
+     * This is to ensure we are not busy clearing the data during a commit log submission.
+     *
+     * @param dataAdder the data addition
+     */
+    private void lockDataAddition(Runnable dataAdder) {
+        lock.readLock().lock();
+        dataAdder.run();
+        lock.readLock().unlock();
```
aelred: Needs to be in a `finally` block!

grakn-kb/src/main/java/ai/grakn/kb/internal/CommitLog.java
```diff
@@ -71,14 +87,16 @@ public Json getFormattedLog(){
...
         if(newInstanceCount.isEmpty() && newAttributes.isEmpty()){
-            return "Commit logs empty. No submission made";
+            return Optional.empty();
         }
         String endPoint = getCommitLogEndPoint(engineUri, keyspace);
+        lock.writeLock().lock();
         String response = EngineCommunicator.contactEngine(endPoint, REST.HttpConn.POST_METHOD, getFormattedLog().toString());
         clear();
-        return "Response from engine [" + response + "]";
+        lock.writeLock().unlock();
```
aelred: Another `finally` for this.

fppt:
!rtg

PR #2105, opened 2017-09-15
Fixed tiny bug in Migration Macros

fppt:


grakn-core/src/main/java/ai/grakn/graql/macro/Macro.java
```diff
@@ -38,10 +38,6 @@
     @CheckReturnValue
     T apply(List<Object> values);
 
-    /**
-     * The name of the macro
-     * @return the name of the macro
-     */
     @CheckReturnValue
     String name();
```
aelred: how will we ever know this is the "name" now???

fppt:
!rtg

fppt:
!rtg

fppt:
!rtg

fppt:
!rtg - ai.grakn.graql.internal.query.match.MatchModifierTest.ai.grakn.graql.internal.query.match.MatchModifierTest

Due to no host exception.

PR #2112, opened 2017-09-19
Optimisation: When linking new concepts to types don't read for existing edges

fppt:


fppt:
screen shot 2017-09-19 at 16 22 55

fppt:
screen shot 2017-09-19 at 16 29 08

fppt:
Top is before the optimisation and bottom is after. Another 20 ops per second!

PR #2117, opened 2017-09-20
Optimisation: Cache properties to speed up transaction constructions

fppt:


fppt:
screen shot 2017-09-20 at 17 04 45

As always, top is before, bottom is after

grakn-kb/src/main/java/ai/grakn/factory/GraknSessionImpl.java
```diff
@@ -81,11 +81,44 @@ public GraknSessionImpl(Keyspace keyspace, String engineUri){
...
+        return properties;
+    }
+
+    /**
+     * Gets properties which let you build a toy in-mempoty {@link GraknTx}.
+     * This does nto contact engine in anyway and can be run in an isolated manner
+     *
+     * @return the properties needed to build an in-memory {@link GraknTx}
+     */
+    private static Properties getTxRemoteInMemory(){
```
aelred: the word "remote" is misleading here

fppt:
@aelred re-review please

PR #2121, opened 2017-09-21
Introduce structural query cache (aka caching graql traversals)

kasper-piskorski:


grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GraqlTraversal.java
```diff
@@ -89,6 +91,17 @@ static GraqlTraversal create(Set<? extends List<Fragment>> fragments) {
...
     /**
+     * @param conceptMap map defining mappings between concept ids
+     * @return graql traversal with concept id transformed according to the provided concept map
+     */
+    public GraqlTraversal transform(Map<ConceptId, ConceptId> conceptMap){
+        ImmutableList<Fragment> fragments = ImmutableList.copyOf(
+                Iterables.getOnlyElement(fragments()).stream().map(f -> f.transform(conceptMap)).collect(Collectors.toList())
+        );
+        return new AutoValue_GraqlTraversal(ImmutableSet.of(fragments));
+    }
```
aelred: This won't work if the query has a disjunction. Is that intended?

grakn-graql/src/main/java/ai/grakn/graql/internal/query/match/MatchBase.java
```diff
@@ -85,6 +85,16 @@ public MatchBase(Conjunction<PatternAdmin> pattern) {
...
+        return streamWithTraversal(this.getPattern(), graph, graqlTraversal);
+    }
+
+    /**
+     * @param pattern a pattern to match in the graph
+     * @param graph the graph to get results from
+     * @param graqlTraversal gral traversal corresponding to the provided pattern
+     * @return resulting answer stream
+     */
+    public static Stream<Answer> streamWithTraversal(Conjunction<PatternAdmin> pattern, GraknTx graph, GraqlTraversal graqlTraversal) {
```
aelred: Instead of passing the `pattern` through here, we can instead just pass `pattern.commonVars()`.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GraqlTraversal.java
kasper-piskorski: all disjunctions are DNFed to sets of conjunctions on entry in `MatchInfer`

kasper-piskorski:
gonna try using the guava Equivalence thingie so bare with me

kasper-piskorski:
@aelred removed the query wrapper class and used guava instead

kasper-piskorski:
@aelred merge :)?

PR #2124, opened 2017-09-21
Simplify Graql traversals without disjunctions

aelred:
When a traversal contains a single conjunction, avoid the step of union'ing the disjunctions together.


kasper-piskorski:
does it make any difference or is it just a cleanup thing?

kasper-piskorski:
hmmm rather not a cleanup ;p

aelred:
@kasper-piskorski _should_ have performance implications only :P

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GraqlTraversal.java
```diff
@@ -66,10 +67,18 @@ static GraqlTraversal create(Set<? extends List<Fragment>> fragments) {
...
+        if (fragments().size() == 1) {
+            // If there are no disjunctions, we don't need to union them and get a performance boost
+            return getConjunctionTraversal(graph, graph.admin().getTinkerTraversal().V(), Iterables.getOnlyElement(fragments()));
+        } else {
+            Traversal[] traversals =
+                    fragments().stream().map(list -> getConjunctionTraversal(graph, __.V(), list)).toArray(Traversal[]::new);
+
+            // This is a sneaky trick - we want to do a union but tinkerpop requires all traversals to start from
+            // somewhere, so we start from a single arbitrary vertex.
+            return graph.admin().getTinkerTraversal().V().limit(1).union(traversals);
```
fppt: tricky

PR #2126, opened 2017-09-21
Optimisation: Remove redundant relationship validation

fppt:
screen shot 2017-09-21 at 18 00 22

Top = Before
Bottom = After


grakn-kb/src/main/java/ai/grakn/kb/internal/Validator.java
```diff
@@ -117,7 +117,7 @@ private void validateRelation(GraknTxAbstract<?> graph, Relationship relationshi
         Optional<RelationshipReified> relationReified = ((RelationshipImpl) relationship).reified();
         //TODO: We need new validation mechanisms for non-reified relations
         relationReified.ifPresent(relationReified1 -> {
-            ValidateGlobalRules.validateRelationshipStructure(relationReified1).ifPresent(errorsFound::add);
+            //ValidateGlobalRules.validateRelationshipStructure(relationReified1).ifPresent(errorsFound::add);
```
aelred: 🐐

PR #2128, opened 2017-09-22
Fix Jenkinsfile to run benchmarks

aelred:


aelred:
@fppt I've merged this to get the benchmarks going, but I'd still like a review when you have time!

PR #2130, opened 2017-09-22
Delete artifacts older than 7 days

burukuru:


burukuru:
This will stop of running out of disk space on Jenkins. Nice.

PR #2131, opened 2017-09-22
Optimise traversal to avoid checking for edge instances

aelred:
If the query can already tell that a traversal does not go to edge instances, then it doesn't have to be checked mid-traversal.

### DIFF
--- a/grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/InIsaFragment.java
+++ b/grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/InIsaFragment.java
@@ -60,6 +60,8 @@ abstract class InIsaFragment extends Fragment {
     public GraphTraversal<Vertex, ? extends Element> applyTraversalInner(
             GraphTraversal<Vertex, ? extends Element> traversal, GraknTx graph, Collection<Var> vars) {
 
+        GraphTraversal<Element, Vertex> vertexTraversal = Fragments.isVertex(traversal);
+
         if (mayHaveEdgeInstances()) {
             GraphTraversal<Vertex, Vertex> isImplicitRelationType =
                     __.<Vertex>hasLabel(RELATIONSHIP_TYPE.name()).has(IS_IMPLICIT.name(), true);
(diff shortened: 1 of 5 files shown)


grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/InIsaFragment.java
```diff
@@ -54,22 +54,30 @@
...
     @Override
     public GraphTraversal<Vertex, ? extends Element> applyTraversalInner(
             GraphTraversal<Vertex, ? extends Element> traversal, GraknTx graph, Collection<Var> vars) {
 
-        GraphTraversal<Vertex, Vertex> isImplicitRelationType =
-                __.<Vertex>hasLabel(RELATIONSHIP_TYPE.name()).has(IS_IMPLICIT.name(), true);
+        GraphTraversal<Vertex, Vertex> vertexTraversal = Fragments.isVertex(traversal);
+
+        if (mayHaveEdgeInstances()) {
+            GraphTraversal<Vertex, Vertex> isImplicitRelationType =
```
fppt: DO you have the RelationType object when executing this traversal? Because if so we have another DB read which can be avoided. This is all cached in the core API.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/sets/IsaFragmentSet.java
```diff
@@ -45,4 +56,41 @@ Var instance() {
...
         return type;
     }
+
+    /**
+     * We can skip the mid-traversal check for edge instances in the following case:
+     *
+     * <ol>
+     *     <li>There is an {@link IsaFragmentSet} {@code $x-[isa:with-edges]->$X}
+     *     <li>There is a {@link LabelFragmentSet} {@code $X[label:foo,bar]}
+     *     <li>The labels {@code foo} and {@code bar} are all not types that may have edge instances</li>
```
fppt: 👍

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/InIsaFragment.java
aelred: No, I don't. I would have to build it mid-traversal - which is possible but it would involve the lambda step at Marko would be sad :(

PR #2132, opened 2017-09-22
Minimize selected labels in Graql traversal

aelred:


grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GraqlTraversal.java
```diff
@@ -66,18 +69,22 @@ static GraqlTraversal create(Set<? extends List<Fragment>> fragments) {
      */
     // Because 'union' accepts an array, we can't use generics
     @SuppressWarnings("unchecked")
-    public GraphTraversal<Vertex, Map<String, Element>> getGraphTraversal(GraknTx graph) {
+    public GraphTraversal<Vertex, Map<String, Element>> getGraphTraversal(GraknTx graph, Set<Var> vars) {
```
fppt: KNOWLEDGE BASE!!!! . . .  oh sorry, wrong time for it.

PR #2139, opened 2017-09-25
Revert change that might have broken travis

aelred:
Prior to merging, confirm that the benchmark tests are still running


aelred:
This will fix Travis, so I'm merging without a review.

PR #2141, opened 2017-09-25
Separate user-defined, generated and reserved vars

aelred:
Additionally includes the performance improvement to generate var names using an incrementing counter instead of UUIDs.

Merge after #2132 - this PR includes those changes.

### DIFF
--- a/grakn-core/src/main/java/ai/grakn/graql/Var.java
+++ b/grakn-core/src/main/java/ai/grakn/graql/Var.java
@@ -76,25 +76,25 @@ public interface Var extends VarPattern {
 
         UserDefined {
             @Override
-            public String prefix() {
-                return "$";
+            public char prefix() {
+                return '§';
             }
         },
 
         Generated {
             @Override
-            public String prefix() {
-                return "£";
+            public char prefix() {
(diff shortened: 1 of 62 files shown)


aelred:
@kasper-piskorski this test fails, but I can't see why :(
```
ReasonerTest.testReasoningWithQueryContainingTautology:474 expected:<ai.grakn.graql.internal.reasoner.query.QueryAnswers@0> but was:<ai.grakn.graql.internal.reasoner.query.QueryAnswers@205a61b2>
```

grakn-core/src/main/java/ai/grakn/graql/Var.java
```diff
@@ -58,4 +68,33 @@
 
     @Override
     int hashCode();
+
+    /**
+     * The {@link Kind} of a {@link Var}, such as whether it is user-defined.
+     */
+    enum Kind {
```
fppt: hahahahahaha I am guessing this is because "Type" is so overloaded in our world ? :D

aelred: 🏆

PR #2142, opened 2017-09-25
More benchmarks

pluraliseseverythings:


grakn-test/test-integration/src/test/java/ai/grakn/test/benchmark/AddBenchmark.java
```diff
@@ -39,7 +49,15 @@ public void tearDown() {
...
+    public void addEntity() {
         entityType.addEntity();
     }
+
+    @Benchmark
+    public void addRelation() {
+            Entity entity1 = entityType.addEntity();
+            Entity entity2 = entityType.addEntity();
+            relationshipType.addRelationship().addRolePlayer(role1, entity1).addRolePlayer(role2, entity2);
+            graph.commit();
```
fppt: Since we have `AddWithCommitBenchmark ` which already contains the above I am guessing having this commit line here is a booboo

PR #2147, opened 2017-09-26
Avoid memory leak in `TaskStateInMemoryStore`

aelred:


grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/TaskStateInMemoryStore.java
```diff
@@ -39,21 +42,35 @@
  * @author Denis Lobanov, alexandraorth
  */
 public class TaskStateInMemoryStore implements TaskStateStorage {
```
fppt: I actually want to kill off all the in memory task management of engine. It never gets deployed and it is hiding bugs :(

PR #2148, opened 2017-09-26
Bug #17579: Migration validation bug. Null stack traces causing migration to crash

fppt:
This was a nasty one!!


grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/redisqueue/RedisTaskStorage.java
```diff
@@ -92,7 +94,16 @@ public TaskState getState(TaskId id) throws GraknBackendException {
             // TODO return optional
             throw GraknBackendException.stateStorage();
         }
-        return TaskState.of(id, TaskStatus.fromState(state.get().getState()));
```
aelred: ohhhh

PR #2149, opened 2017-09-26
Hal printer

marco-scoppetta:
- fixed HALExplore with Roles that are not types anymore
- fixed parameter used in HALPrinter to call HALConceptData method


grakn-graql/src/main/java/ai/grakn/graql/internal/hal/HALExploreSchemaConcept.java
```diff
@@ -41,29 +42,31 @@
...
-        attachTypeResources(halResource, type);
-        // Subtypes
-        attachSubTypes(halResource, type);
+        if (schemaConcept.isType()) {
+            // Role types played by current type
+            attachRolesPlayed(halResource, schemaConcept.asType().plays());
+            // Resources types owned by the current type
+            attachTypeResources(halResource, schemaConcept.asType());
+            // Subtypes
+            attachSubTypes(halResource, schemaConcept.asType());
```
aelred: A schema concept does still have sub-concepts, so this can go outside of the `if` (unless `roleTypeSchema` handles this below)

PR #2150, opened 2017-09-26
Add `GraqlParser` with parameter to define all vars

aelred:
This includes:
- a refactor to introduce a `GraqlParser` interface housing most of the `parseSomething` methods.
- a flag on `GraqlParser` to "define all vars", meaning give names for things like relations so they will be returned in the results


aelred:
!rtg - build timed out for no reason!

PR #2152, opened 2017-09-26
Analytics without attribute

duckofyork:
Hide implicit relationships and attributes from users. Add method includeAttribute() so that advanced users can still include all the implicit relationships and attributes when using analytics queries.


grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractComputeQuery.java
```diff
@@ -113,33 +121,50 @@ public void kill() {
...
+        // only include attributes and implicit "has-xxx" relationships when user specifically asked for them
+        GraknTx graknTx = tx.get();
         if (subLabels.isEmpty()) {
-            tx.admin().getMetaConcept().subs().forEach(subTypes::add);
+            if (includeAttribute) {
+                graknTx.admin().getMetaConcept().subs().forEach(subTypes::add);
+            } else {
+                graknTx.admin().getMetaEntityType().subs().forEach(subTypes::add);
+                graknTx.admin().getMetaRelationType().subs()
+                        .filter(entityType -> !entityType.isImplicit()).forEach(subTypes::add);
```
aelred: `relationshipType` :P

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractComputeQuery.java
```diff
@@ -113,33 +121,50 @@ public void kill() {
...
+                }
+                return type.asType();
             }).collect(Collectors.toSet());
 
-            subTypes = subTypes.stream().flatMap(Type::subs).collect(Collectors.toSet());
+            if (includeAttribute) {
+                subTypes = subTypes.stream().flatMap(Type::subs).collect(Collectors.toSet());
+            } else {
+                subTypes = subTypes.stream().flatMap(Type::subs)
+                        .filter(entityType -> !entityType.isImplicit()).collect(Collectors.toSet());
```
aelred: This is just `type`, not `entityType`

PR #2169, opened 2017-10-03
[Stable] Add more role expansion tests and fix issues

kasper-piskorski:


grakn-graql/pom.xml
```diff
@@ -57,6 +57,12 @@
             <artifactId>halbuilder-standard</artifactId>
             <version>${halbuilder-standard.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-math3</artifactId>
+            <version>3.4.1</version>
```
fppt: This version number should be a property

kasper-piskorski:
!rtg - timeout

PR #2173, opened 2017-10-04
Changes tests to using new mock redis - Part 1

fppt:


fppt:
!rtg - ai.grakn.test.engine.tasks.connection.RedisCountStorageTest.whenIncreasingCountOnRedisConcurrently_EnsureAllThreadCountsArePersisted

fppt:
!rtg - failed before it even started

fppt:
!rtg - damn cache

grakn-engine/src/test/java/ai/grakn/engine/EngineTestHelper.java
```diff
@@ -55,7 +54,7 @@ public static synchronized void engine() {
         if (server != null) {
             return;
         }
-        GraknTestSetup.startRedisIfNeeded(new SimpleURI(config().getProperty(GraknEngineConfig.REDIS_HOST)).getPort());
+        //GraknTestSetup.startRedisIfNeeded(new SimpleURI(config().getProperty(GraknEngineConfig.REDIS_HOST)).getPort());
```
aelred: 🐐

grakn-engine/src/test/java/ai/grakn/engine/config/GraknEngineConfigTest.java
```diff
@@ -35,6 +39,9 @@
 
     private GraknEngineConfig configuration = GraknEngineConfig.create();
 
+    @ClassRule
+    public static MockRedisRule mockRedisRule = MockRedisRule.create(new SimpleURI(EngineTestHelper.config().getProperty(GraknEngineConfig.REDIS_HOST)).getPort());
```
aelred: Maybe we should have a helper method to do this, since we do the same thing several times.

Also, I don't understand why this test needs `MockRedis`, but then `GraknEngineConfig` is a scary place.

grakn-test/test-integration/src/test/java/ai/grakn/test/engine/GraknEngineFailoverIT.java
```diff
@@ -73,10 +75,10 @@
     public static final DistributionContext engine1 = DistributionContext.startSingleQueueEngineProcess().port(7890).redisPort(redisURI.getPort());
 
     @ClassRule
-    public static final DistributionContext engine2 = DistributionContext.startSingleQueueEngineProcess().port(5678).redisPort(redisURI.getPort());
+    public static final DistributionContext engine2 = DistributionContext.startSingleQueueEngineProcess().port(5678).redisPort(5512);
 
     @ClassRule
-    public static final DistributionContext engine3 = DistributionContext.startSingleQueueEngineProcess().port(6789).redisPort(redisURI.getPort());
+    public static final DistributionContext engine3 = DistributionContext.startSingleQueueEngineProcess().port(6789).redisPort(5513);
```
aelred: How does this work?

grakn-test/test-integration/src/test/java/ai/grakn/test/graql/shell/GraqlShellIT.java
```diff
@@ -71,8 +71,7 @@
 public class GraqlShellIT {
 
     @ClassRule
-    public static final DistributionContext dist = DistributionContext.startInMemoryEngineProcess().inheritIO(true);
-
+    public static final DistributionContext dist = DistributionContext.startInMemoryEngineProcess().inheritIO(false);
```
aelred: I don't think we need this change anymore, since we suppressed the logs above.

grakn-engine/src/test/java/ai/grakn/engine/config/GraknEngineConfigTest.java
fppt: I tried to make a helper method. The problem is that any helper method will have go into `MockRedisRule` which is in `grakn-test-tools`. This helper method will need to use the config `GraknEngineConfig` which is in `grakn-engine`. 

So that means making the helper method re-read the config (ewww) and that is a bad idea because often tests override options in the config. This is the least evil approach for the moment.

grakn-test/test-integration/src/test/java/ai/grakn/test/engine/GraknEngineFailoverIT.java
fppt: Right so before hand we were stating a REAL redis which every engine shared. It did not matter which engine started it. 

Now we giving each engine their own FAKE redis. We give each one their own redis because without doing that you will have to control the startups of engines, which is a lot of bloat work for this test. I.e. you will have to make sure at least one starts fully before starting the others. This is just simpler.

grakn-test/test-integration/src/test/java/ai/grakn/test/graql/shell/GraqlShellIT.java
fppt: yes you are correct

grakn-engine/src/test/java/ai/grakn/engine/EngineTestHelper.java
fppt: Will kill the goat

grakn-engine/src/test/java/ai/grakn/engine/config/GraknEngineConfigTest.java
aelred: OK, I see! This seems reasonable then.

grakn-test/test-integration/src/test/java/ai/grakn/test/engine/GraknEngineFailoverIT.java
aelred: Ahh, I see. And `engine1` is different because that's the one we read the tasks from to check the test passed.

fppt:
@aelred re-review please

fppt:
!rtg - ai.grakn.factory.GraknTxJanusTest.ai.grakn.factory.GraknTxJanusTest

PR #2174, opened 2017-10-04
Fix documentation

micheleorsi:
Following the documentation I found some small problems in running graql queries, so here is the fix!


aelred:
!rtg - can't see what went wrong

aelred:
!rtg - tests timed out

aelred:
!rtg - tests never stopped

PR #2183, opened 2017-10-05
Add @ to the front of implicit labels and make it a reserved character

fppt:


fppt:
@aelred could I get another review please?

PR #2188, opened 2017-10-06
Make sure long-running instance clears old dists

aelred:


aelred:
Merging because this only impacts the long-running instance

PR #2189, opened 2017-10-06
Re-enable SNB update queries

aelred:


aelred:
!rtg - testing timed out :(

PR #2191, opened 2017-10-06
Run Jenkinsfile in parallel

aelred:
Build once, then run everything in parallel


burukuru:
!rtg

aelred:
!rtg - random failure, logged [here](https://work.grakn.ai/entity/18086)

aelred:
yay! @fppt have a look whenever

Jenkinsfile
```diff
@@ -97,45 +93,72 @@ def buildGrakn() {
...
+        }
     }
-}
 
-// Deploy long-running instance on stable branch
-if (env.BRANCH_NAME == 'stable') {
-    node {
-        String workspace = pwd()
-        checkout scm
+    jobs = [
```
burukuru: Could you add some comments here to say what you're doing?

Jenkinsfile
```diff
@@ -97,45 +93,72 @@ def buildGrakn() {
...
+                    }
+                }
+            }
+        }
+    ];
 
-        stage('Build Grakn') {
-            withScripts(workspace) {
-                buildGrakn()
+    for (String moduleName : integrationTests) {
```
burukuru: and here

Jenkinsfile
```diff
@@ -97,45 +93,72 @@ def buildGrakn() {
...
             }
         }
+    }
 
-        stage('Deploy Grakn') {
-            sshagent(credentials: ['jenkins-aws-ssh']) {
-                sh "scp -o StrictHostKeyChecking=no grakn-dist/target/grakn-dist*.tar.gz ubuntu@${LONG_RUNNING_INSTANCE_ADDRESS}:~/"
-                sh "scp -o StrictHostKeyChecking=no scripts/repeat-query ubuntu@${LONG_RUNNING_INSTANCE_ADDRESS}:~/"
-                ssh "'bash -s' < scripts/start-long-running-instance.sh"
+    parallel(jobs);
```
burukuru: and here

Jenkinsfile
```diff
@@ -97,45 +93,72 @@ def buildGrakn() {
...
+
+    node {
```
burukuru: You can remove it as there's a nested 'node' clause here.

aelred: This node is actually not inside another. I think the `parallel` part has to be executed outside of a node, so I leave the first node to run then - then I have to make a new one to deploy at the end.

burukuru: There's another node clause on 153 though.

aelred: Ohhh! I see it, thanks.

PR #2193, opened 2017-10-06
Re-enable a subset of the SNB queries

aelred:


aelred:
!rtg - cassandra timeout

aelred:
!rtg - failed to set github status

PR #2199, opened 2017-10-10
Re-enable last short SNB query

aelred:


aelred:
We haven't been able to figure out why SQ3 is so slow. We'll revisit it later.

PR #2201, opened 2017-10-10
Load example into long-running instance and periodically query

aelred:


aelred:
!rtg - random failure in `BatchMutatorClientTest`

aelred:
!rtg - timed out after 2 hours

scripts/periodic
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -e
+
+PATH="/home/ubuntu/grakn:$PATH"
+
+while true; do
+        graql console -e 'match $x isa pokemon; get;';
```
fppt: This is not really periodic right? It's more a continuous and repeated query with no breaks?

aelred: Yes, I guess so! It's always querying. I should rename it.

PR #2203, opened 2017-10-11
Bring back EmbeddedRedis for benchmarks

fppt:
:(


grakn-test/test-integration/src/test/java/ai/grakn/test/benchmark/MutatorTaskManagerBenchmark.java
```diff
@@ -0,0 +1,10 @@
+package ai.grakn.test.benchmark;
+
+import ai.grakn.test.EngineContext;
+
+
+public class MutatorTaskManagerBenchmark extends MutatorTaskBenchmark {
+    protected EngineContext makeEngine() {
+        return EngineContext.createWithInMemoryRedis();
+    }
+}
```
aelred: Who is using this class?

fppt: It's a test class so this is just used internally. I think that method is protected because we are planning on adding more Task related benchmarks which need to make an Engine

fppt: And it's currently used in the `@Setup` of this Benchmark

aelred: Are you sure it's in use? I pulled down the branch, but I can't find where it's used.
There's no other mention of `MutatorTaskManagerBenchmark` in the code and this class itself contains no benchmarks.

aelred: I just want an excuse to delete shit :P

fppt: Oh just realised you referring to the OTHER one. I thought you were talking about `MutatorTaskBenchmark`. . . Okay so it looks like `MutatorTaskMANAGERBenchmark` is not in use. Lets try deleting it then.

PR #2205, opened 2017-10-11
[Stable] Disallow implicit concepts in rule heads

kasper-piskorski:


kasper-piskorski:
!rtg - VM terminated without properly saying goodbye

PR #2206, opened 2017-10-11
[Stable] Check instances when matching rules

kasper-piskorski:


grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
```diff
@@ -261,7 +261,7 @@ public boolean isAllowedToFormRuleHead(){
         }
 
         //check roles are ok
```
fppt: Roles are no longer Types and this map seems to be collecting Types. Maybe this comment is now misleading?

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerQueryImpl.java
```diff
@@ -123,6 +129,17 @@
                 .build();
     }
 
+    private Set<IsaAtom> inferEntityTypes() {
```
fppt: Maybe this can return the `Stream` to begin with? I ask because this returns a `Set` but in `getVarTypeMap` it gets converted to a stream again.

kasper-piskorski:
!rtg

kasper-piskorski:
!rtg - ai.grakn.client.BatchMutatorClientTest.whenEngineRESTFailsWhileLoadingWithRetryTrue_LoaderRetriesAndWaits

kasper-piskorski:
@fppt re-review?

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
```diff
@@ -260,9 +260,8 @@ public boolean isAllowedToFormRuleHead(){
             return errors;
         }
 
-        //check roles are ok
-        Map<Var, SchemaConcept> varSchemaConceptMap = getParentQuery().getVarSchemaConceptMap();
-
+        //check role-tyoe compatibility
```
fppt: type-o `tyoe` can be fixed in another PR though

PR #2210, opened 2017-10-12
Try optimise relation writing - Part 1

fppt:
Remove redundant uniqueness check

### DIFF
--- a/grakn-kb/src/main/java/ai/grakn/kb/internal/ValidateGlobalRules.java
+++ b/grakn-kb/src/main/java/ai/grakn/kb/internal/ValidateGlobalRules.java
@@ -276,16 +276,15 @@ class ValidateGlobalRules {
     }
 
     /**
-     * @param graph graph used to ensure the {@link Relationship} is unique
      * @param relationReified The {@link Relationship} whose hash needs to be set.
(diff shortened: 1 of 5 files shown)


grakn-kb/src/main/java/ai/grakn/kb/internal/ValidateGlobalRules.java
```diff
@@ -276,29 +276,27 @@ private ValidateGlobalRules() {
     }
 
     /**
-     * @param graph graph used to ensure the {@link Relationship} is unique
      * @param relationReified The {@link Relationship} whose hash needs to be set.
      * @return An error message if the {@link Relationship} is not unique.
      */
-    static Optional<String> validateRelationIsUnique(GraknTxAbstract<?> graph, RelationshipReified relationReified){
+    static Optional<String> validateRelationIsUnique(RelationshipReified relationReified){
```
aelred: Since we're doing the renaming thing - can we make these relation-ship now? 👍

fppt: Yep. . .also. . . I hate you.

fppt:
@aelred re-review please

PR #2211, opened 2017-10-12
Fix incorrect filename in jenkinsfile

aelred:


aelred:
@fppt with your permission, can we merge this before tests?

fppt:
Merge what? Also wink wink: https://github.com/graknlabs/grakn/pull/2210

PR #2216, opened 2017-10-12
Try optimise relation writing - Part 3

fppt:
Pre emptive caching on Casting to prevent DB reads

### DIFF
--- a/grakn-kb/src/main/java/ai/grakn/kb/internal/concept/RelationshipReified.java
+++ b/grakn-kb/src/main/java/ai/grakn/kb/internal/concept/RelationshipReified.java
@@ -38,6 +38,7 @@ import org.apache.tinkerpop.gremlin.structure.Direction;
 import org.apache.tinkerpop.gremlin.structure.Edge;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 
+import javax.annotation.Nullable;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
@@ -71,7 +72,7 @@ public class RelationshipReified extends ThingImpl<Relationship, RelationshipTyp
      * This is because validation requires iterating over all castings anyway.
      */
     private Cache<Set<Casting>> allCastings = new Cache(Cacheable.set(), () -> lazyCastings(Collections.EMPTY_SET).collect(Collectors.toSet()));
-    private RelationshipImpl owner;
+    @Nullable private RelationshipImpl owner;
 
     private RelationshipReified(VertexElement vertexElement) {
         super(vertexElement);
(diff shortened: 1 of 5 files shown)


grakn-kb/src/main/java/ai/grakn/kb/internal/structure/Casting.java
```diff
@@ -43,21 +45,42 @@
...
+            return (RelationshipType) edge().tx().getSchemaConcept(LabelId.of(edge().property(Schema.EdgeProperty.RELATIONSHIP_TYPE_LABEL_ID)));
+        }
+    });
+
+    private Casting(EdgeElement edgeElement, @Nullable Relationship relationship, @Nullable Role role, @Nullable Thing thing){
         this.edgeElement = edgeElement;
+        if(relationship != null) this.cachedRelationship.set(relationship);
+        if(role != null) this.cachedRole.set(role);
+        if(thing != null) this.cachedInstance.set(thing);
+    }
```
aelred: Alternative suggestion: given these caches are mutable anyway, maybe expose some setters for them? Then you only need one constructor and a setter for each cache. I'm not 100% sold on either way though.

grakn-kb/src/main/java/ai/grakn/kb/internal/concept/RelationshipReified.java
```diff
@@ -71,6 +71,7 @@
      * This is because validation requires iterating over all castings anyway.
      */
     private Cache<Set<Casting>> allCastings = new Cache(Cacheable.set(), () -> lazyCastings(Collections.EMPTY_SET).collect(Collectors.toSet()));
+    private RelationshipImpl owner;
```
aelred: I'd really like to avoid this field if possible, but can't figure out how...
It should definitely be marked `@Nullable` either way!

grakn-kb/src/main/java/ai/grakn/kb/internal/structure/Casting.java
fppt: Nothing else sets these caches explicitly that's why I don't think it's worth making setters.

grakn-kb/src/main/java/ai/grakn/kb/internal/concept/RelationshipReified.java
fppt: There is a way. Instead of storing a reference to this field. When the actual relationship is needed I can just do:

```
relationship = vertex().tx().getConcept(getId());
```

reckon that would be better? I was never sure.

burukuru:
!rtg

burukuru:
!rtg

PR #2219, opened 2017-10-13
Fix performance when traversing to concept subs

aelred:
we actually don't know if this works just yet


fppt:
@aelred it looks good to me but let's confirm the fix first.

aelred:
@fppt good call! It doesn't work :P

aelred:
@fppt please re-review

@mikonapoli if you have time, can you confirm the Markit query is now fast? I ran it locally and it seems to be, but this has taken so long that we should make sure we get it right.

fppt:
Still looks good to me. Not merging just because of the title of the PR. @aelred merge at will.

PR #2222, opened 2017-10-16
Remove casting cache [DON'T MERGE]

aelred:
This is an experiment to see if it's causing SNB to not load


fppt:
The cache really did break SNB. Closing this and continuing dev here: https://github.com/graknlabs/grakn/pull/2223 .

For some weird reason I cannot push onto your PRs :/

PR #2223, opened 2017-10-17
Remove casting cache

fppt:


fppt:
!rtg - ai.grakn.factory.GraknTxJanusTest.ai.grakn.factory.GraknTxJanusTest

PR #2224, opened 2017-10-17
Always add the role player edge when using a BATCH transaction. Also PP task to clean up duplicate Role Player Edges 

fppt:


aelred:
We shouldn't merge this until I (or someone) adds some tests to Graql to check that it disregards any duplicate edges.

aelred:
I've fixed Graql to handle this situation, but I'm not convinced the trade-off is worth it. If you look at the code, you'll see the role-player traversal is much more complicated - which is a large chunk of our querying.

fppt:
@aelred re-review please

PR #2226, opened 2017-10-17
Remove text output from `graql` script

aelred:
The `graql console` command should not produce unnecessary output to allow it to be used by other programs, such as piping the output to another program to parse.


aelred:
!rtg - timeout

PR #2230, opened 2017-10-18
Merge from stable

kasper-piskorski:


fppt:
!rtg - ai.grakn.graql.internal.query.UndefineQueryTest.ai.grakn.graql.internal.query.UndefineQueryTest
 - due to cass timeout

fppt:
@aelred I am reverting this because the commits were squashed which should not when merging from stable.

fppt:
Sorry I am an idiot. Don't know why I thought this was squashed.

PR #2233, opened 2017-10-18
refactored grakn engine

micheleorsi:
I cleaned up a bit the grakn engine module in order to prepare the modularisation of new components. I focused on the main, because it's the heart of a dependency injection structure. 

The responsibility of 'GraknCreator' is to create singletons to use all across the application. 

I duplicated a bit the creation code also in 'EngineTestHelper' because integration tests need fresh instances (not the already-used ones, learnt in the hard-way after so many test failures ..). 
I could have put those methods in 'GraknCreator', but then it was not clear anymore for a user of 'GraknCreator' which one to use to get an instance of a specific class. This decision forced me to add a new test dependency between 'grakn-engine' module and 'grakn-test', that has only 'test' as a scope. Final thought: I think that duplication is better than risk of misuse!

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/GraknApplication.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/GraknApplication.java
@@ -1,0 +1,45 @@
+/*
+ * Grakn - A Distributed Semantic Database
+ * Copyright (C) 2016  Grakn Labs Limited
+ *
+ * Grakn is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Grakn is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Grakn. If not, see <http://www.gnu.org/licenses/gpl.txt>.
+ */
+
+package ai.grakn.engine;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ *
+ * Main class invoked by bash scripting
+ *
+ * @author Michele Orsi
+ *
+ */
+public class GraknApplication {
+
+    private static final Logger LOG = LoggerFactory.getLogger(GraknEngineServer.class);
+
+    public static void main(String[] args) {
+        try {
+            // Start Engine
+            GraknEngineServer graknEngineServer = GraknCreator.instantiateGraknEngineServer(Runtime.getRuntime());
+            graknEngineServer.start();
+        } catch (Exception e) {
+            LOG.error("An exception has occurred", e);
+        }
+    }
+
+}
--- a/grakn-engine/src/main/java/ai/grakn/engine/GraknCreator.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/GraknCreator.java
@@ -54,18 +54,35 @@ import static com.codahale.metrics.MetricRegistry.name;
  */
 public class GraknCreator {
 
-    private final static EngineID ENGINE_ID = EngineID.me();
-    private final static Service SPARK_SERVICE = Service.ignite();
-    private final static GraknEngineStatus GRAKN_ENGINE_STATUS = new GraknEngineStatus();
-    private final static MetricRegistry METRIC_REGISTRY = new MetricRegistry();
-    private final static ExecutorService EXECUTOR_SERVICE = TasksController.taskExecutor();
-    private final static GraknEngineConfig GRAKN_ENGINE_CONFIG = GraknEngineConfig.create();
+    private final static EngineID ENGINE_ID = engineId();
+    private final static Service SPARK_SERVICE = sparkService();
+    private final static GraknEngineStatus GRAKN_ENGINE_STATUS = graknEngineStatus();
+    private final static MetricRegistry METRIC_REGISTRY = metricRegistry();
+    private final static ExecutorService EXECUTOR_SERVICE = executorService();
+    private final static GraknEngineConfig GRAKN_ENGINE_CONFIG = graknEngineConfig();
 
     private static GraknEngineServer graknEngineServer;
     private static RedisWrapper redisWrapper;
     private static LockProvider lockProvider;
     private static EngineGraknTxFactory engineGraknTxFactory;
     private static TaskManager taskManager;
+
+    protected static EngineID engineId() {
+        return EngineID.me();
+    }
+    protected static Service sparkService() {
+        return Service.ignite();
+    }
+    protected static GraknEngineStatus graknEngineStatus() {
+        return new GraknEngineStatus();
+    }
+    protected static MetricRegistry metricRegistry() { return new MetricRegistry(); }
+    protected static ExecutorService executorService() {
+        return TasksController.taskExecutor();
+    }
+    protected static GraknEngineConfig graknEngineConfig() {
+        return GraknEngineConfig.create();
+    }
 
     static synchronized GraknEngineServer instantiateGraknEngineServer(Runtime runtime) {
         if (graknEngineServer == null) {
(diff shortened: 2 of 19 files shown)


micheleorsi:
!rtg no test failures present, but Jenkins failures are there!

micheleorsi:
!rtg no test failures present, but Jenkins failed!

micheleorsi:
!rtg timed out

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -192,245 +117,42 @@ private void lockAndInitializeSystemSchema() {
             if (lock.tryLock(60, TimeUnit.SECONDS)) {
                 loadAndUnlock(lock);
             } else {
-                LOG.info("{} found system schema lock already acquired by other engine", this.engineId);
+                LOG.info("{} found system schema lock already acquired by other engine");
```
fppt: Maintaining the engine id here and elsewhere might be useful for debugging on the cluster.

grakn-engine/src/main/java/ai/grakn/engine/GraknApplication.java
```diff
@@ -0,0 +1,45 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ *
+ * Main class invoked by bash scripting
+ *
+ * @author Michele Orsi
+ *
+ */
+public class GraknApplication {
```
lolski: If we can call this class `Grakn` it'll display better to the user when they do `ps` or `jps`.

Do you think it is easy enough to do?

grakn-engine/src/test/java/ai/grakn/engine/EngineTestHelper.java
```diff
@@ -81,4 +108,60 @@ public static synchronized void noEngine() {
             server = null;
         }
     }
+    
+    public static synchronized GraknEngineServer graknEngineServer(GraknEngineConfig config) {
+        RedisWrapper redisWrapper = redisWrapper(config);
+        return EngineTestHelper.graknEngineServer(config,redisWrapper);
+    }
+
+    private static RedisWrapper redisWrapper(GraknEngineConfig config) {
```
fppt: To eliminate all of this duplicated code might it not be best to expose the relevant methods in `GraknCreator` or provide some common factory methods? 

I imagine this being problematic to maintain long term.

grakn-engine/src/main/java/ai/grakn/engine/GraknApplication.java
fppt: We already have a class called `Grakn` but it's in a different package so it should be okay. Just need to take care with those auto imports :)

grakn-engine/src/test/java/ai/grakn/engine/EngineTestHelper.java
micheleorsi: I put the reason of this duplication in the first comment, let me know you if it convinces you .. ;)
Otherwise I will just create private method and let EngineTestHelper.java extend GraknCreator

> I duplicated a bit the creation code also in 'EngineTestHelper' because integration tests need fresh instances (not the already-used ones, learnt in the hard-way after so many test failures ..).
I could have put those methods in 'GraknCreator', but then it was not clear anymore for a user of 'GraknCreator' which one to use to get an instance of a specific class. This decision forced me to add a new test dependency between 'grakn-engine' module and 'grakn-test', that has only 'test' as a scope. Final thought: I think that duplication is better than risk of misuse!

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
micheleorsi: you're right! I thought it was worth to remove, because basically this class depends already on so many other classes ..

```java
GraknEngineServer(GraknEngineConfig prop, TaskManager taskManager, EngineGraknTxFactory factory, LockProvider lockProvider, GraknEngineStatus graknEngineStatus, RedisWrapper redisWrapper, ExecutorService taskExecutor, HttpHandler httpHandler)
```

Just let me know if you think it's worth, otherwise I will revert back!

grakn-engine/src/test/java/ai/grakn/engine/EngineTestHelper.java
fppt: I read that and because of how we work I am not entirely. Each component is very much "owned" by someone so the risk of misuse is less (in my opinion) than the risk of forgetting to maintain the duplicated code. Especially because of the rate of flux on how we change these components.

micheleorsi: ok got it! I will change!
Thanks for the overview!

grakn-engine/src/main/java/ai/grakn/engine/GraknApplication.java
micheleorsi: Ok, I am going to rename this class as "Grakn"

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
fppt: Hmmm not sure. @pluraliseseverythings mostly debugs engine. How helpful is knowing the engine ID?

grakn-engine/src/main/java/ai/grakn/engine/HttpHandler.java
```diff
@@ -0,0 +1,255 @@
...
+
+import static ai.grakn.engine.GraknEngineConfig.WEBSOCKET_TIMEOUT;
+
+/**
+ *
+ * @author Michele Orsi
+ */
+public class HttpHandler {
+
+    private static final Logger LOG = LoggerFactory.getLogger(GraknEngineServer.class);
```
lolski: Should we change the logger name to be `getLogger(HttpHandler.class)` for consistency?

In our codebase, every logger is named after the class it belongs to

grakn-engine/src/main/java/ai/grakn/engine/GraknApplication.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(GraknEngineServer.class);
```
lolski: LoggerFactory.getLogger(GraknApplication.class)

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
lolski: I imagine when we're working with many nodes in a cluster it will useful, especially when logs also displays engine ID.

grakn-engine/src/main/java/ai/grakn/engine/HttpHandler.java
micheleorsi: Sure! I maintained the GraknEngineServer name so that my refactor was invisible from a user point of view.

But it makes sense to be consistent with the codebase! I will do fix it!

micheleorsi:
Just fixed everything you asked for!

lolski:
LGTM!

grakn-engine/src/main/java/ai/grakn/engine/GraknCreator.java
```diff
@@ -0,0 +1,151 @@
...
+    private final static ExecutorService EXECUTOR_SERVICE = TasksController.taskExecutor();
+    private final static GraknEngineConfig GRAKN_ENGINE_CONFIG = GraknEngineConfig.create();
+
+    private static GraknEngineServer graknEngineServer;
+    private static RedisWrapper redisWrapper;
+    private static LockProvider lockProvider;
+    private static EngineGraknTxFactory engineGraknTxFactory;
+    private static TaskManager taskManager;
+
+    static synchronized GraknEngineServer instantiateGraknEngineServer(Runtime runtime) {
```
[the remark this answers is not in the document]
micheleorsi: Well these are just the factories, you can still manage lifecycle and mocking through real constructor. I agree with you that we shouldn't use these factories in the tests, but we should properly mock the dependencies. On the other hand at the moment the refactoring was so painful that I just stop here the clean-up for now!

Moreover if you make them static you cannot access anymore from the main. If you want to compare this class is like the @Configuration with @Bean in Spring. But here we don't have a proper DI framework to handle singleton instances ..

grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServer.java
```diff
@@ -96,57 +45,29 @@
...
     private final LockProvider lockProvider;
-    private final GraknEngineStatus graknEngineStatus = new GraknEngineStatus();
+    private final GraknEngineStatus graknEngineStatus;
     private final RedisWrapper redisWrapper;
     private final ExecutorService taskExecutor;
+    private final HttpHandler httpHandler;
 
-    private GraknEngineServer(GraknEngineConfig prop, RedisWrapper redisWrapper) {
+
+    public GraknEngineServer(GraknEngineConfig prop, TaskManager taskManager, EngineGraknTxFactory factory, LockProvider lockProvider, GraknEngineStatus graknEngineStatus, RedisWrapper redisWrapper, ExecutorService taskExecutor, HttpHandler httpHandler) {
```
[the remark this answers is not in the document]
micheleorsi: I know that it looks weird, but it highlights the dependencies. Again (as before) we should work on improving encapsulation and reducing coupling, but probably that would trigger a big re-factoring!

PR #2234, opened 2017-10-18
Make long-running query script executable

aelred:


aelred:
Merging this without tests because it only changes the long-running instance behaviour.

PR #2235, opened 2017-10-18
[Stable] Fix issues with resources playing roles

kasper-piskorski:


grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/Atom.java
```diff
@@ -226,7 +226,11 @@ private boolean isRuleApplicable(InferenceRule child){
     }
 
     public IdPredicate getIdPredicate(Var var){
-        return getPredicates(IdPredicate.class).filter(p -> p.getVarName().equals(var)).findFirst().orElse(null);
+        return getPredicate(var, IdPredicate.class);
+    }
+
```
fppt: Can you add `@Nullable` here? That way we at least know this can NPE. That or just use the `Optional` to force handling the Nulls

kasper-piskorski: done

PR #2240, opened 2017-10-19
Bug #17981: Fix crash when undefining types

fppt:


grakn-kb/src/main/java/ai/grakn/kb/internal/cache/GlobalCache.java
```diff
@@ -100,9 +120,22 @@ public void cacheLabel(Label label, LabelId id) {
      * @param txCache The transaction cache
      */
     void readTxCache(TxCache txCache) {
-        //TODO: The difference between the caches need to be taken into account. For example if a type is delete then it should be removed from the cachedLabels
-        cachedLabels.putAll(txCache.getLabelCache());
-        cachedTypes.putAll(txCache.getSchemaConceptCache());
+        //Check if the ontology has been changed and should be flushed into this cache
+        if(!cachedLabels.equals(txCache.getLabelCache())) {
```
aelred: I'm not 100% sure if the lock should be acquired before this point or not.

fppt: I think it should be okay (don't hold me to that). This will only really become and issue if we concurrently mutating the schema and if you doing that then you just a loon.

PR #2246, opened 2017-10-20
Merge from stable

aelred:


aelred:
@fppt this one probably doesn't have to wait for the tests

fppt:
Will merge I think the periodic build is failing right now anyway.

PR #2248, opened 2017-10-20
Change logs on jenkins back to INFO [DON'T MERGE]

aelred:


aelred:
This demonstrates that something else is causing SNB not to load :(

PR #2249, opened 2017-10-20
Make match benchmark test a bit more similar to real use case

duckofyork:
so I don't have to change this every time.


aelred:
@pluraliseseverythings can you have a look at this when you have time?

pluraliseseverythings:
Definitely an improvement LGTM

PR #2251, opened 2017-10-20
Check SNB load back in time [DON'T MERGE]

aelred:
This is just before the role-player edge change


fppt:
Continued here: https://github.com/graknlabs/grakn/pull/2254

PR #2256, opened 2017-10-23
Remove JWT handler

micheleorsi:


grakn-engine/src/main/java/ai/grakn/engine/session/RemoteSession.java
```diff
@@ -56,10 +56,6 @@ public static RemoteSession create() {
         return new RemoteSession(null);
     }
 
-    public static RemoteSession passwordProtected(UsersHandler usersHandler) {
-        return new RemoteSession(usersHandler);
-    }
```
aelred: This means the `usersHandler` is always `null`. We can remove even more code!

PR #2258, opened 2017-10-24
Cache shard counts

fppt:
This is what @duckofyork needs in order to be able to optimise some queries.


grakn-kb/src/main/java/ai/grakn/kb/internal/concept/ConceptImpl.java
```diff
@@ -57,6 +57,7 @@
         Optional<Shard> shard = vertex().tx().factory().buildShard(shardVertex);
         return shard.orElseThrow(() -> GraknTxOperationException.missingShard(getId()));
     });
+    private final Cache<Long> shardCount = new Cache(Cacheable.number(), () -> shards().count());
```
aelred: waiiiit... why isn't this in `TypeImpl`?

fppt: Because `technically` all concept can have shards and eventually will. Relationship sharding for example.

duckofyork: 👍👍👍  looking forwards to instance sharding

PR #2259, opened 2017-10-24
Improve internal caching

fppt:


grakn-kb/src/main/java/ai/grakn/kb/internal/concept/ConceptImpl.java
```diff
@@ -84,6 +89,13 @@ public void delete() throws GraknTxOperationException {
         deleteNode();
     }
 
+    /**
+     * Registers a {@link Cache} so that later it can be cleaned up
+     */
+    public void registerCahce(Cache cache){
```
aelred: *Cache

grakn-kb/src/main/java/ai/grakn/kb/internal/concept/RelationshipEdge.java
```diff
@@ -57,21 +57,21 @@
     private final Logger LOG = LoggerFactory.getLogger(RelationshipEdge.class);
     private final EdgeElement edgeElement;
 
-    private final Cache<RelationshipType> relationType = new Cache<>(Cacheable.concept(), () ->
+    private final Cache<RelationshipType> relationType = Cache.createTxCache(Cacheable.concept(), () ->
```
aelred: Maybe we can make this work like `Concept` by adding a `CacheOwner` interface? `Concept` and `RelationshipStructure` can implement it - then we don't need the factory method that doesn't take an owner.

grakn-kb/src/main/java/ai/grakn/kb/internal/concept/ThingImpl.java
```diff
@@ -70,13 +70,13 @@
...
 public abstract class ThingImpl<T extends Thing, V extends Type> extends ConceptImpl implements Thing {
-    private final Cache<Label> cachedInternalType = new Cache<>(Cacheable.label(), () -> {
+    private final Cache<Label> cachedInternalType = Cache.createTxCache(this, Cacheable.label(), () -> {
         int typeId = vertex().property(Schema.VertexProperty.THING_TYPE_LABEL_ID);
         Optional<Type> type = vertex().tx().getConcept(Schema.VertexProperty.LABEL_ID, typeId);
         return type.orElseThrow(() -> GraknTxOperationException.missingType(getId())).getLabel();
     });
 
-    private final Cache<V> cachedType = new Cache<>(Cacheable.concept(), () -> {
+    private final Cache<V> cachedType = Cache.createTxCache(Cacheable.concept(), () -> {
```
aelred: I think this one needs to be passed `this` so it is cleared.

It might be a good idea to rename the `Cache#createTxCache` method that doesn't take an owner, otherwise we will definitely make this mistake again :P

grakn-kb/src/main/java/ai/grakn/kb/internal/structure/Casting.java
```diff
@@ -45,18 +45,18 @@
  */
 public class Casting {
     private final EdgeElement edgeElement;
-    private final Cache<Role> cachedRole = new Cache<>(Cacheable.concept(), () -> (Role) edge().tx().getSchemaConcept(LabelId.of(edge().property(Schema.EdgeProperty.ROLE_LABEL_ID))));
-    private final Cache<Thing> cachedInstance = new Cache<>(Cacheable.concept(), () -> edge().target().
+    private final Cache<Role> cachedRole = Cache.createTxCache(Cacheable.concept(), () -> (Role) edge().tx().getSchemaConcept(LabelId.of(edge().property(Schema.EdgeProperty.ROLE_LABEL_ID))));
+    private final Cache<Thing> cachedInstance = Cache.createTxCache(Cacheable.concept(), () -> edge().target().
```
aelred: These caches aren't cleared either - don't know if it's deliberate.

fppt:
@aelred re-review please. I tried to incorporate the new interface which basically led to a lot of redundant code about adding a cache to a set and then clearing the caches in the set later. That sucked. 

So now I did this. . . which also sucks. Opinions?

fppt:

PR #2260, opened 2017-10-24
Flush tx cache to central cache even when not committing and not usin…

fppt:
…g a read transaction


fppt:
!rtg - i.grakn.client.BatchExecutorClientTest.whenEngineRESTFailsWhileLoadingWithRetryTrue_LoaderRetriesAndWaits

PR #2261, opened 2017-10-24
Trying to fix validation errors

pluraliseseverythings:


pluraliseseverythings:
1) Cleans up the migrator class. Now it has a builder, you set the options there and that's it. No more bloated load(...)
2) Added a --lines command to be used when debugging if you need only a number of lines to be migrated
3) Now we don't get thousands of failures in the validation but around 5-10. @aelred fixed some queries
4) Now the CI validation actually fails in Jenkins. Before it succeeded but there were thousands of failures

grakn-test/test-integration/src/test/bash/build-grakn.sh
```diff
@@ -18,4 +18,4 @@ echo "Setting version using branch name"
 mvn versions:set -DnewVersion=${BRANCH_NAME} -DgenerateBackupPoms=false
 
 echo "Installing grakn"
-mvn clean install --batch-mode -Dmaven.repo.local=${WORKSPACE}/maven -DskipTests -U -Djetty.log.level=WARNING -Djetty.log.appender=STDOUT
+mvn clean install -T 14 --batch-mode -Dmaven.repo.local=${WORKSPACE}/maven -DskipTests -U -Djetty.log.level=WARNING -Djetty.log.appender=STDOUT
```
aelred: Can we turn this up to 15 please?

Also uh what does this number do

pluraliseseverythings: ahem, 14 is a lot maybe but it's the number of thread per core used to build. If we are not testing that doesn't cause problems

pluraliseseverythings: I'll reduce it on the next one and merge it then!

PR #2264, opened 2017-10-24
Add docs explaining how doc tests work

aelred:
I put some docs in your docs etc. I'm tired


aelred:
!rtg - added to [TP](https://work.grakn.ai/entity/18085)

PR #2267, opened 2017-10-25
missed commit statement

lolski:


fppt:
Uhm . . . I don't think this will work. Or rather it shouldn't. I don't think we allow creating relationships with no role players.

fppt:
I stand corrected :/

lolski:
@fppt Yes, the endpoint takes the following, which means it should work:
```
{
  "relationshipType": {
    "label": "<relationship-type-name>",
    "roles": ["role1", "role2", ...]
  }
}
```

If this endpoint satisfies all of the possible use cases of relationship type creation though, is a different matter of discussions. Though I hope it does.

I also think the lack of documentation for what's supposed to be a public API must be addressed

fppt:
@lolski I am talking about the `Relationship` not the `RelationshipType` definition. The `RelationshipType` has roles but the relationship you creating does not have any _role players_.

You create the relationship but never add a role-player to it.

I thought we prohibited this via an internal validation rule. I don't know if we should be allowing relationships with no role players.

lolski:
@fppt I see what you mean now. Yes, when I did the Java API PR, I managed to create a relationship without any role-players and I thought it's how it should work.

If it isn't how it should be, the relationship endpoint should be updated accordingly.

fppt:
Yep, I think you may have found a bug in the internal Java API. Although I am not sure if it's worth fixing yet . . . hmmm

lolski:
I think we should merge this PR as it fixes another bug.

For the relationship issue, I have created a ticket here which we can prioritise accordingly: https://work.grakn.ai/entity/18096

fppt:
Agreed

PR #2269, opened 2017-10-25
More robust 'grakn server status' check

lolski:
There were two occasions where `ps ... | grep ...` used for engine and redis uptime check would give wrong result. For example, if the working directory is also named `Grakn`, or `redis-server`. 

This PR adds a fix to protect against these edge cases.

In addition to that, the logic for the uptime checks and getting the proper redis binary name are now extracted into their own dedicated functions, in order to reduce cognitive overload to human readers.

Blergh, so much engineering effort spent just for uptime check. I am open to suggestion if there are better approach to ps | grep

### DIFF
--- a/grakn-dist/src/grakn
+++ b/grakn-dist/src/grakn
@@ -76,7 +76,7 @@ ps_ef_grep_processname() {
 
 
 ps_ef_grep_processname_by_java_classname() {
-  echo -n `ps -ef | grep -v grep | awk '{print $2" "$(NF)}' | grep "$1"`
+  ps -ef | grep -v grep | awk '{print $2" "$(NF)}' | grep "$1"
 }
 
 # ================================================
--- a/grakn-dist/src/grakn
+++ b/grakn-dist/src/grakn
@@ -199,7 +199,8 @@ queue_start_process() {
 
 queue_check_if_running() {
   local queue_fullpath="${GRAKN_HOME}/services/redis/`get_redis_binary_name`"
-  local count_running_queue_process=`ps -ef | grep 'redis-server' | grep -v grep | awk '{ print $2}' | wc -l`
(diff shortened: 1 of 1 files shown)


grakn-dist/src/grakn
```diff
@@ -68,6 +70,15 @@ update_classpath_global_var() {
   CLASSPATH="$CLASSPATH":"${GRAKN_HOME}"/services/grakn
 }
 
+ps_ef_grep_processname() {
+  ps -ef | grep -v grep | awk '{print $2" "$(NF-1)}' | grep "$1"
```
burukuru: I'd suggest pgrep if that works on mac

grakn-dist/src/grakn
```diff
@@ -68,6 +70,15 @@ update_classpath_global_var() {
...
+}
+
+
+ps_ef_grep_processname_by_java_classname() {
+  ps -ef | grep -v grep | awk '{print $2" "$(NF)}' | grep "$1"
```
burukuru: `jps` should give you what you want

lolski: @burukuru as far as I know, `jps` isn't part of OpenJDK which we claim to support

burukuru: oh that sucks

grakn-dist/src/grakn
burukuru: apparently mac doesn't have pgrep so nevermind..

grakn-dist/src/grakn
lolski: yup. the joy of maintaining platform compatibility...

aelred:
Is this ready for review/merging? I don't know if @burukuru wants to request changes.

PR #2273, opened 2017-10-26
Rework REST endpoints

aelred:
In particular, we move the keyspace into the path.

This does not include a few things that still need doing:
- Removing the `Concept`, `Auth` and `UserController`
- Removing the HAL representation
- Reworking the KB `/api` endpoints


aelred:
!rtg - timeouts

grakn-core/pom.xml
```diff
@@ -63,10 +63,20 @@
...
         </dependency>
         <dependency>
+            <artifactId>httpclient</artifactId>
+            <groupId>org.apache.httpcomponents</groupId>
+            <version>${httpclient.version}</version>
+        </dependency>
+        <dependency>
             <groupId>io.reactivex</groupId>
             <artifactId>rxjava</artifactId>
             <version>1.3.2</version>
```
fppt: Not related to this PR but this should be a property in the main pom.

grakn-core/src/main/java/ai/grakn/util/REST.java
```diff
@@ -42,44 +48,45 @@
...
             @Deprecated
             public static final String GRAQL = "/kb/graql";
-            public static final String ANY_GRAQL = "/kb/graql/execute";
+            public static final String ANY_GRAQL = "/kb/:keyspace/graql";
         }
 
         /**
          * URIs to Tasks Controller endpoints
          */
         public static class Tasks {
```
fppt: Aren't tasks always linked with a specific keyspace? If that is true shouldn't that keyspace be on the path?

grakn-engine/src/main/java/ai/grakn/engine/SystemKeyspaceImpl.java
```diff
@@ -150,12 +110,18 @@ public boolean deleteKeyspace(Keyspace keyspace){
         return true;
     }
 
-    /**
-     * Load the system schema into a newly created system keyspace. Because the schema
-     * only consists of types, the inserts are idempotent and it is safe to load it
-     * multiple times.
-     */
-    public void loadSystemSchema() {
+    @Override public Stream<Keyspace> keyspaces() {
```
fppt: Does this actually work? We returning a stream which depends on the transaction being open and if the stream is executing lazily it might do so with a closed transaction?

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
```diff
@@ -137,9 +156,9 @@ private Object executeGraql(Request request, Response response) {
             return resp;
         }
     }
-    
+
     @GET
-    @Path("/")
+    @Path("/graph/graql")
```
fppt: I don't get why we need this endpoint? What is it actually used for?

grakn-engine/src/test/java/ai/grakn/engine/controller/CommitLogControllerTest.java
```diff
@@ -0,0 +1,87 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.isA;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+/**
+ * @author Felix Chapman
+ */
+public class CommitLogControllerTest {
```
fppt: We didn't have one of these??

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
aelred: This is a deprecated endpoint that returns things in a very particular format that the HAL representation needs. @marco-scoppetta is working on removing it.

grakn-core/src/main/java/ai/grakn/util/REST.java
aelred: There's nothing _requiring_ tasks to have a keyspace param, but I think you might be right that they all do now. We can work on that in another PR, I'm worried about how long this one is becoming!

grakn-engine/src/main/java/ai/grakn/engine/SystemKeyspaceImpl.java
aelred: Good point! I think I always immediately collect the result. I'll return a `Set` instead.

grakn-engine/src/test/java/ai/grakn/engine/controller/CommitLogControllerTest.java
aelred: 👍

fppt:
@aelred merge conflicts :)

aelred:
!rtg -  🎵 time outs 🎵

PR #2274, opened 2017-10-26
Tackling some old Todos

fppt:


grakn-core/src/main/java/ai/grakn/GraknConfigKey.java
```diff
@@ -72,6 +72,8 @@
 
     public static final GraknConfigKey<Path> STATIC_FILES_PATH = key("server.static-file-dir", required(Paths::get));
 
+    public static final GraknConfigKey<Integer> SESSION_CACHE_TIMEOUT = key("knowledge-base.schema-cache-timeout-ms", INT);
```
aelred: I would keep `_MS` in the name here (or literally make it return a `Duration`, but that's more complicated)

grakn-kb/src/main/java/ai/grakn/factory/GraknSessionImpl.java
```diff
@@ -114,7 +116,7 @@ private static Properties getTxRemoteProperties(String engineUrl, Keyspace keysp
     private static Properties getTxInMemoryProperties(){
         Properties inMemoryProperties = new Properties();
         inMemoryProperties.put(GraknConfigKey.SHARDING_THRESHOLD, 100_000);
-        inMemoryProperties.put(GraknTxAbstract.NORMAL_CACHE_TIMEOUT_MS, 30_000);
+        inMemoryProperties.put(GraknConfigKey.SESSION_CACHE_TIMEOUT, 30_000);
         inMemoryProperties.put(FactoryBuilder.KB_MODE, TxFactoryTinker.class.getName());
```
aelred: I'm wondering if this code technically works... I suspect you actually want to call
```
inMemoryProperties.setProperty(GraknConfigKey.SESSION_CACHE_TIMEOUT.name(), 30_000);
```
otherwise, the key is going to be a `GraknConfigKey`, when it should be a string.

fppt: That's likely true for the sharding threshold as well. I suppose this never got caught because the In memory graph is only ever used as a toy

fppt:
!rtg - ramdom timeout

aelred:
!rtg - known issue that is being fixed

aelred:
!rtg - same again...

aelred:
!rtg - this is a logged flaky test

aelred:
!rtg - same old same old

fppt:
!rtg - this is getting ridiculous

fppt:

aelred:
@fppt are we sure we didn't break something? This is pretty silly now :/

fppt:

aelred:
@fppt if it's the one I think it is, @pluraliseseverythings is fixing it.

fppt:
Yep it is. I have seen it fail on other PRs before. If @pluraliseseverythings can't fix it we should ignore this for now. I am starting to pile up PRs on this one test.

fppt:
!rtg - The test is ignored. It should FINALLY pass

aelred:
FINALLY

PR #2279, opened 2017-10-27
Incorporate shard count for better fragment cost - DO NOT MERGE

duckofyork:


grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -86,57 +89,76 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, GraknTx graph
...
+
+                    edges.get(weightedDirectedEdge.val.destination).get(weightedDirectedEdge.val.source)
+                            .setAccurateFragmentCost(nodesWithFixedCost.get(weightedDirectedEdge.val.source));
+                    weightedDirectedEdge.weight = -nodesWithFixedCost.get(weightedDirectedEdge.val.source);
+                    startingNodeSet.add(weightedDirectedEdge.val.source);
+                }
+                weightedGraph.add(weightedDirectedEdge);
+                connectedNodes.add(weightedDirectedEdge.val.destination);
+                connectedNodes.add(weightedDirectedEdge.val.source);
+            });
```
aelred: It might be a good idea to split this method up plus add some clarifying variables and comments.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/Fragment.java
```diff
@@ -99,16 +99,26 @@
     static final double COST_NODE_NOT_INTERNAL = -Math.log(1.1D);
     static final double COST_NODE_IS_ABSTRACT = -Math.log(1.1D);
 
+    public static final double SHARD_LOAD_FACTOR = 0.25 - 1.0; // there is at least one shard
```
aelred: It's not totally clear what these numbers mean. What does 0.25 mean?

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/Fragment.java
```diff
@@ -99,16 +99,26 @@
...
+
+    private static final long DEFAULT_SHARDING_THRESHOLD = 10_000L;
```
aelred: I guess we can't use the one from the config because that needs an engine dependency :(

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/spanningtree/util/Weighted.java
```diff
@@ -27,12 +27,11 @@
  * Add a weight to anything!
  *
  * @param <T> the type of object
- * @author Sam Thomson
  * @author Jason Liu
  */
 public class Weighted<T> implements Comparable<Weighted<T>> {
     public final T val;
-    public final double weight;
+    public double weight;
```
aelred: Why not final? :o

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -228,6 +226,88 @@ private static void addUnvisitedNodeFragments(List<Fragment> plan,
...
+                connectedNodes.add(superType);
+                nodesWithFixedCost.put(
+                        Node.addIfAbsent(NodeId.NodeType.VAR, fragment.end(), allNodes),
+                        nodesWithFixedCost.get(superType));
+                subFragmentSet.add(fragment);
+            });
+            subFragmentSet.forEach(inSubFragment -> {
+                allFragments.remove(inSubFragment);
+                allFragments.remove(Fragments.outSub(
+                        inSubFragment.varProperty(), inSubFragment.end(), inSubFragment.start()));
```
aelred: You can replace this with `inSubFragment.getInverse()`

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/OutSubFragment.java
```diff
@@ -53,7 +58,7 @@ public String name() {
 
     @Override
     public double fragmentCost() {
-        return COST_SAME_AS_PREVIOUS;
+        return accurateFragmentCost.orElse(COST_SAME_AS_PREVIOUS);
```
aelred: This "default" behaviour using `accurateFragmentCost` needs to move down to the `Fragment` interface.

I'd suggest having something like:

```
default double fragmentCost() {
    return accurateFragmentCost.orElse(internalFragmentCost());
}

double internalFragmentCost();
```

Then everything implements `internalFragmentCost()` instead of `fragmentCost()`.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -228,6 +226,88 @@ private static void addUnvisitedNodeFragments(List<Fragment> plan,
...
+
+        // check whether it's value fragment
+        if (fragment.varProperty() instanceof ValueProperty) {
+            // as value fragment is not symmetric, we need to add it again
+            other.getFragmentsWithDependency().add(fragment);
+            start.getDependants().add(fragment);
+        }
+    }
+
+    private static void processSubFragment(List<Fragment> plan,
```
aelred: This should have a comment/javadoc explaining why we need it! 👍

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/Fragment.java
fppt: Do you have access to a transaction/graph/ 🍌  at this stage? If so then i can give you the correct number.

duckofyork: Yes please!

fppt: I will give you a method to get this number then.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/Fragment.java
duckofyork: By default I assume a shard is 25% full. So one shard mean 2500 instances, two shards = 12500, so on so forth

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
duckofyork: Let me check if I have implemented this method for sub :-)

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/fragment/Fragment.java
duckofyork: cool. Thank Filipe!

duckofyork: @fppt should I add a commit in this pr or it's something for the future?

fppt: I would say merge this PR and we can patch it in later.

duckofyork: Cool

PR #2282, opened 2017-10-27
Get updating counts PP Task to be immediate and not use tasks

fppt:
This makes the Updating Counts task run immediately. The queue is skipped entirely. The reasoning for this is that the queue part just adds overhead.

In addition to this I have created a `PostProcessor` class to encapsulate the logic for post processing. This is the first step towards making Post Processing run without needing the queue at all.

This change prompted a lot of cleanup so I also ended up killing off the stop resume functionality. It was not used in anyway.


fppt:
!rtg - ???

fppt:
!rtg - ai.grakn.factory.TxFactoryJanusTest.ai.grakn.factory.TxFactoryJanusTest
ai.grakn.test.property.GetQueryPropertyTest.anyPropertyCanBeExecutedOnAMatchQuery

fppt:
ai.grakn.test.property.GetQueryPropertyTest.anyPropertyCanBeExecutedOnAMatchQuery

fppt:
!rtg -

grakn-engine/src/main/java/ai/grakn/engine/GraknCreator.java
```diff
@@ -150,14 +152,18 @@ protected static EngineGraknTxFactory engineGraknTxFactory(GraknEngineConfig con
      */
     private static synchronized TaskManager instantiateTaskManager(MetricRegistry metricRegistry, GraknEngineConfig config, EngineID engineId, EngineGraknTxFactory factory,
```
pluraliseseverythings: Indentation

grakn-engine/src/main/java/ai/grakn/engine/loader/MutatorTask.java
```diff
@@ -96,8 +97,8 @@ private boolean insertQueriesInOneTransaction(GraknTx graph, Collection<Query> i
                     addTask(PostProcessingTask.createTask(this.getClass(), engineConfiguration()
```
pluraliseseverythings: Actually, I just realised we can delete the MutatorTask. Maybe for another PR

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessor.java
```diff
@@ -0,0 +1,274 @@
...
+     */
+    public void updateCounts(Keyspace keyspace, Json commitLog){
+        final long shardingThreshold = engineConfig.getProperty(GraknConfigKey.SHARDING_THRESHOLD);
+        final int maxRetry = engineConfig.getProperty(GraknConfigKey.LOADER_REPEAT_COMMITS);
+        try (Timer.Context context = metricRegistry.timer(name(PostProcessor.class, "execution")).time()) {
+            Map<ConceptId, Long> jobs = getCountUpdatingJobs(commitLog);
+            metricRegistry.histogram(name(PostProcessor.class, "jobs"))
+                    .update(jobs.size());
+
+            //We Use redis to keep track of counts in order to ensure sharding happens in a centralised manner.
```
pluraliseseverythings: If the problem here is only the fact that each engine has a different snapshot, this is something we can solve with a G-Counter 
https://github.com/aphyr/meangirls#g-counter

pluraliseseverythings: (without redis)

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessor.java
```diff
@@ -0,0 +1,274 @@
...
+     * - Actually sharding
+     * - Incrementing the number of shards on each type
+     *
+     * @param keyspace The database containing the {@link ai.grakn.concept.Type} to shard
+     * @param conceptId The id of the concept to shard
+     */
+    private void shardConcept(RedisCountStorage redis, EngineGraknTxFactory factory,
+                              Keyspace keyspace, ConceptId conceptId, int maxRetry, long shardingThreshold){
+        Lock engineLock = lockProvider.getLock(getLockingKey(keyspace, conceptId));
+        engineLock.lock(); //Try to get the lock
```
pluraliseseverythings: Is this a local lock? Should we get a distributed lock like in the duplicate method?

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessor.java
pluraliseseverythings: Also, just a comment, not for this PR

grakn-engine/src/main/java/ai/grakn/engine/loader/MutatorTask.java
fppt: I will give it a try. I like deleting code 🙂

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessor.java
fppt: Oh that is cool, but sadly we need to be able to decrement as well. For example when you delete instances.

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessor.java
fppt: The implementation is `JedisLockProvider` which is distributed. You know we may be able to get rid of the interface. We only have one implementation of `LockProvider` what do you think?

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessor.java
pluraliseseverythings: There's the PN-counter for that. The downside of course is that instead of a number we are saving (number of engines * 2) numbers in that case. It might still be good enough to avoid synchronisation and Redis

grakn-engine/src/main/java/ai/grakn/engine/postprocessing/PostProcessor.java
pluraliseseverythings: Ok! We can leave the interface

grakn-engine/src/main/java/ai/grakn/engine/loader/MutatorTask.java
fppt: On snd thought I will leave it for another PR. Ripping it out will affect a lot it seems.

fppt:
!rtg - dist failure

fppt:
!ci

PR #2283, opened 2017-10-28
Run tests with Janus profile in Jenkinsfile.

burukuru:


burukuru:
I seem to be missing around 20 tests from the results, need to find out where they're hidden.

Also the test results are show as a flat structure, not per module; I assume that's not an issue?

Jenkinsfile
```diff
@@ -13,6 +13,7 @@ class Constants {
 
 //This sets properties in the Jenkins server. In this case run every 8 hours
 properties([pipelineTriggers([cron('H H/8 * * *')])])
+    properties([buildDiscarder(logRotator(numToKeepStr: '30', artifactNumToKeepStr: '7'))])
```
aelred: INDENTATION DISTRESS

Jenkinsfile
```diff
@@ -94,9 +95,25 @@ def buildGrakn() {
...
+node {
+    String workspace = pwd()
+    checkout scm
+
+    slackGithub "Janus tests started"
+
+    timeout(90) {
+	stage('Run Janus test profile') {
+	    sh "mvn clean verify -P janus -U -Djetty.log.level=WARNING -Djetty.log.appender=STDOUT || true"
+	    junit "**/target/surefire-reports/*.xml"
```
aelred: We think this means if the tests fail, the build won't. I guess we need a `try...finally` instead.

PR #2286, opened 2017-10-30
[Stable] Fix visualisation bugs for attributes attached to explanation nodes

marco-scoppetta:
- if an attribute is a roleplayer of explicitly asked relationship, it needs to be added to the graph
  - after a new HAL response is rendered, check if new attributes nodes need to be linked to existing nodes


marco-scoppetta:
!rtg - on Integration tests Build timed out (after 120 minutes)

PR #2287, opened 2017-10-30
Fix bug where timeout when starting Grakn printed "SUCCESS"

aelred:


aelred:
@burukuru done! With a fancy function

aelred:
!rtg - I can't see what went wrong here

aelred:
!rtg - looks like the build stopped?

PR #2289, opened 2017-10-31
[Stable] Enable implicit types during type inference

kasper-piskorski:
NB: In general this currently will decrease performance of queries requiring type inference as these will result in more possible types.

### DIFF
--- a/grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
+++ b/grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
@@ -464,6 +464,7 @@ public class RelationshipAtom extends IsaAtom {
 
         return compatibleTypes.asMap().entrySet().stream()
                 .sorted(Comparator.comparing(e -> -e.getValue().size()))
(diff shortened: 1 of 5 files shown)


grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
```diff
@@ -442,6 +442,7 @@ else if(compatibleTypesFromRoles.isEmpty() || types.isEmpty()){
 
         return compatibleTypes.asMap().entrySet().stream()
                 .sorted(Comparator.comparing(e -> -e.getValue().size()))
+                .sorted(Comparator.comparing(e -> e.getKey().isImplicit()))
```
aelred: Can we add a comment explaining why the ordering is necessary? I assume it's either some technical reason or just to make it easier to read.

kasper-piskorski: it's the order in which the relations are looked at, so trying to prioritise the ones that most probably will yield answers

fppt:
@kasper-piskorski merge conflicts :(

PR #2290, opened 2017-10-31
redis process check bugfix by not using queue full path

lolski:
`grep`-ing a full path (i.e. 

```
ps x | grep /path/to/grakn-1.0.0/services/redis/redis-server-{linux|osx}
```

seems to cause problem in the jenkins machine. therefore i opted to change it:
```
ps -ef | grep services/redis/redis-server-{linux|osx}
```

### DIFF
--- a/grakn-dist/src/grakn
+++ b/grakn-dist/src/grakn
@@ -186,7 +186,7 @@ get_redis_binary_name() {
 
 queue_start_process() {
   # run queue
-  # queue needs to be ran with $GRAKN_HOME as the working directory
+  # the pushd "$GRAKN_HOME" line is needed because queue needs to be ran with $GRAKN_HOME as the working directory
   # otherwise it won't be able to find its data directory located at $GRAKN_HOME/db/redis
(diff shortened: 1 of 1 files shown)


fppt:
@lolski this should be a PR against stable no?

aelred:
@fppt I think this is just a fix for the bug introduced earlier that is causing `master` to fail. @lolski correct me if I'm wrong.

lolski:
@fppt @aelred yes, it should be against master as the likely culprint ( https://github.com/graknlabs/grakn/commit/87747b4c3590763b108f5e73173deb4d88e6f0e5 ) hasn't been merged to stable.

lolski:
!rtg - "java.net.BindException: Address already in use" https://jenkins.grakn.ai/job/unit-test-pull-request-grakn-titan/ai.grakn$grakn-engine/6713/

lolski:
@fppt @aelred closing this pull request as the CI issue was resolved by reverting the commit

PR #2294, opened 2017-10-31
Upload perf logs s3

burukuru:


burukuru:
OK so it's uploading the JSON files to the S3 bucket and you can see thefile  list in XML format... https://s3-eu-west-1.amazonaws.com/performance-logs.grakn.ai/.
You can find the files by Key tag and filter by branch. Does that work at all? @fppt @pluraliseseverythings

burukuru:
@aelred i think this can be merged despite the red cross! LMK if there are any issues with the script.

PR #2296, opened 2017-10-31
More graql property tests

fppt:


grakn-test/test-integration/src/test/java/ai/grakn/test/property/DefineQueryPropertyTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+@RunWith(JUnitQuickcheck.class)
+public class DefineQueryPropertyTest {
+
+    @Rule
+    public ExpectedException exception = ExpectedException.none();
+
+    @Ignore("Currently no error message is returned when trying to insert an empty set of propoerties. I am not entirely sure this is correct")
+    @Property
+    public void anInsertQueryWithoutAnIsaProperty_CannotBeInserted(@Open GraknTx tx, @Size(max=5) Set<VarProperty> properties){
```
aelred: This test is named wrong, I think. Should be about `define` and `sub`.

Also, I think it may not be true... for example, `define person plays actor;` is valid but has no `sub`.

fppt: Ahhhhh yes you are right. I will come up with better ones.

fppt:
@aelred re-review please

PR #2297, opened 2017-10-31
Run tests with Janus profile in Jenkinsfile

burukuru:


burukuru:
Number of tests match (1512) and time is roughly the same! Good to merge!

PR #2299, opened 2017-11-01
Fix bug with `CommitLog` building endpoint

aelred:


aelred:
!rtg - another cassandra timeout

PR #2303, opened 2017-11-01
Centralise keyspace creation

fppt:


fppt:
!rtg - ai.grakn.graql.internal.query.AutocompleteTest.ai.grakn.graql.internal.query.AutocompleteTest 

Due to `Caused by: com.datastax.driver.core.exceptions.NoHostAvailableException:`

PR #2313, opened 2017-11-02
Reduced logs and clarified where the null events were coming from (it…

pluraliseseverythings:
… was an artifact of one of the observable methods)


pluraliseseverythings:
!rtg

burukuru:
!rtg

PR #2314, opened 2017-11-02
Fix Academy homepage links

mikonapoli:


fppt:
Merging this because's it's only small doc changes

PR #2317, opened 2017-11-02
Link fixes

mikonapoli:


fppt:
Same deal these are only doc changes. Merging so I can push up the fixes.

PR #2318, opened 2017-11-03
Add acadmey link to overview side bar

fppt:


fppt:
Merging because this is just doc changes

PR #2319, opened 2017-11-03
Run tests in random order

fppt:
This is so I can help find the flakier flaky tests and get rid of them.


aelred:
Great idea! Is the idea to make this the default?

fppt:
Yep. I can't get PIT mutation testing going until I get rid of our flaky tests. So that's my next goal

fppt:
@aelred should we talk about merging this? I kind of have reservations but in the long run it may be a good thing :/

aelred:
@fppt I'm OK with merging it. Like you say, we'll find lots of issues very fast.

fppt:
Alright merging it. But if we find that PRs never pass we can pull it out.

PR #2325, opened 2017-11-06
Always define some Grakn env vars.

burukuru:


grakn-dist/src/grakn
```diff
@@ -40,11 +40,12 @@ WAIT_INTERVAL_S=2
 if [ -z "${GRAKN_HOME}" ]; then
     [[ $(readlink $0) ]] && path=$(readlink $0) || path=$0
     GRAKN_HOME=$(cd "$(dirname "${path}")" && pwd -P)
-    GRAKN_CONFIG="${GRAKN_HOME}/conf/grakn.properties"
-    GRAKN_PID=/tmp/grakn.pid
-    GRAKN_STARTUP_TIMEOUT_S=120
 fi
 
+GRAKN_CONFIG="${GRAKN_HOME}/conf/grakn.properties"
```
aelred: This one should still stay in the `if` because it refers to `${GRAKN_HOME}`

burukuru: The if clause sets GRAKN_HOME when it's empty.
If it's pre-defined, we still want to set GRAKN_CONFIG.

aelred: oh yeah! thanks for the wisdom

burukuru: Thanks for the merge!

PR #2327, opened 2017-11-06
Get rid of test-jar dependency

fppt:


grakn-engine/src/main/java/ai/grakn/engine/GraknCreator.java
```diff
@@ -172,6 +180,22 @@ protected static TaskManager taskManager(GraknEngineConfig config, EngineGraknTx
         return new RedisTaskManager(engineId, config, jedisPool, consumers, factory, lockProvider, metricRegistry);
     }
 
+    @VisibleForTesting
```
aelred: what is this witchcraft

fppt: Nothing special. Just makes it obvious that this is exclusively exposed for testing only.

PR #2330, opened 2017-11-06
Adding a more informative failure diagnostic output if the script somehow fails to start

lolski:


fppt:
Maybe should be on stable considering it looks like we have to hotfix 0.17 release?

lolski:
@fppt Hmm should we? this is intended to print informative logs in order to help identifying CI failures.

fppt:
Ah I did not realise it was not meant for external use. In that case leave as is.

fppt:
Maybe we can make a verbose flag in this case? That's way we don't have to remember to take this our later.

lolski:
@fppt Actually, this should prove to be useful for users in case they're having problems with the start up / shut down.

I am just of the opinion that we should not include it on the very next hotfix - as we might want to tweak the stuff which is printed, still

what do you think?

fppt:
I agree lets not release it with the hotfix but we should add a flag so that when we debug people issues we can have more info to play with.

lolski:
@fppt `grakn server status --verbose`?

lolski:
!rtg - ai.grakn.factory.JanusPreviousPropertyStepTest  Time elapsed: 17.485 sec  <<< ERROR!
com.datastax.driver.core.exceptions.NoHostAvailableException: All host(s) tried for query failed (tried: localhost/127.0.0.1:9042 (com.datastax.driver.core.exceptions.OperationTimedOutException: [localhost/127.0.0.1:9042] Operation timed out))

lolski:
!ci - ai.grakn.factory.JanusPreviousPropertyStepTest  Time elapsed: 17.485 sec  <<< ERROR!
com.datastax.driver.core.exceptions.NoHostAvailableException: All host(s) tried for query failed (tried: localhost/127.0.0.1:9042 (com.datastax.driver.core.exceptions.OperationTimedOutException: [localhost/127.0.0.1:9042] Operation timed out))

lolski:
!rtg - All host(s) tried for query failed (tried: localhost/127.0.0.1:9042 (com.datastax.driver.core.exceptions.OperationTimedOutException: [localhost/127.0.0.1:9042] Operation timed out))

PR #2331, opened 2017-11-06
Introduce more `TestRule` classes

aelred:


fppt:
Continued on: https://github.com/graknlabs/grakn/pull/2337

PR #2333, opened 2017-11-07
Dynamically merge resources which can be created in highly concurrent…

fppt:
… environments.

This should not be merged until we check if it has helped the Markit load. This is another high impact change from me :).


fppt:
@mikonapoli when you have time could you try the markit load using this PR?

mikonapoli:
@fppt should I wait for the tests to pass first?

fppt:
@aelred I think it's "safe" to put this in.

aelred:
@fppt you're right. I avoided this one because it had a scary complicated name :P

PR #2338, opened 2017-11-07
homebrew workaround.

lolski:
Added a workaround for the issue where grakn starts slower when the executable is invoked from a "fat" directory.

The workaround is done with `pushd` / `popd`. At the start of the script we `pushd`-ed into `$GRAKN_HOME`, so everything that is executed, is always executed from the `$GRAKN_HOME`

At the end, we `popd`-ed to get back to whathever directory we were before.

This PR includes other fixes that I actually want to exclude, but are too error prone to extract out:
1. better error message
2. ascii art + grammar fix
3. faster `grakn server status`


lolski:
!rtg - Caused by: com.datastax.driver.core.exceptions.NoHostAvailableException: All host(s) tried for query failed (tried: localhost/127.0.0.1:9042 (com.datastax.driver.core.exceptions.OperationTimedOutException: [localhost/127.0.0.1:9042] Operation timed out))
	at com.datastax.driver.core.ControlConnection.reconnectInternal(ControlConnection.java:233)

PR #2345, opened 2017-11-08
moved dependencies out from parent pom, into specific modules

micheleorsi:


micheleorsi:
I did this split so that I can use the parent pom of the open source version for the enterprise one!

grakn-client/pom.xml
```diff
@@ -31,6 +31,10 @@
 
     <dependencies>
         <dependency>
+            <groupId>com.google.auto.value</groupId>
+            <artifactId>auto-value</artifactId>
+        </dependency>
```
aelred: I think these dependencies need a version number (and the same in a bunch of places below).

micheleorsi: Actually no, it's the new _cool_ thing I inserted in the grakn root pom: <dependencyManagement>.

It's cool because you can define there all the dependency (with versions) in one place and then just mention group and artefact when you need it!
If you agree I can bring all the dependencies in the grakn root pom, so that everything behaves in the same way!

aelred: ohhhh I totally missed that! Yeah, that sounds like a really good idea. We can merge this as-is for now though, if you need it.

PR #2347, opened 2017-11-08
Some of the writes were being lost on the queue and after that,

pluraliseseverythings:
the writing threads were killing the server


pluraliseseverythings:
!rtg

PR #2349, opened 2017-11-08
Fix bug where we ran the same tests 4 times

aelred:


aelred:
@fppt can you review? this one should go in fast so we know stable is still working

PR #2352, opened 2017-11-09
Resolution state cleanup

kasper-piskorski:


grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/state/AnswerState.java
```diff
@@ -38,13 +31,21 @@
...
  */
-class AnswerState extends ResolutionState {
+public class AnswerState extends ResolutionState {
 
+    private final InferenceRule rule;
     private final Unifier unifier;
 
-    AnswerState(Answer sub, Unifier u, QueryState parent) {
+    public AnswerState(Answer sub, Unifier u, QueryStateBase parent) {
+        super(sub, parent);
```
fppt: Some tiny redundant code. This constructor could just call the other with:

```
this(sub, u, parent, null);
```

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/state/AtomicState.java
```diff
@@ -41,85 +40,100 @@
...
  */
-public class AtomicState extends QueryState{
-
-    private final ReasonerAtomicQuery query;
-    private final Iterator<Answer> dbIterator;
-    private final Iterator<Pair<InferenceRule, Unifier>> ruleIterator;
-
-    private InferenceRule currentRule = null;
-    private final MultiUnifier cacheUnifier;
+@SuppressFBWarnings("BC_UNCONFIRMED_CAST_OF_RETURN_VALUE")
```
fppt: Why did you have to suppress this? What's causing it?

kasper-piskorski: call to getQuery()

PR #2353, opened 2017-11-09
Cleanup session handling

fppt:


grakn-factory/janus-factory/src/test/java/ai/grakn/factory/GraknTxJanusTest.java
```diff
@@ -56,15 +57,21 @@
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 public class GraknTxJanusTest extends JanusTestBase {
+    private final static GraknSession session = mock(GraknSession.class);
```
aelred: The super-class also has a (private) mock `session`. I would use that one if you can.

grakn-factory/janus-factory/src/test/java/ai/grakn/factory/TxFactoryJanusTest.java
```diff
@@ -49,13 +50,18 @@
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 public class TxFactoryJanusTest extends JanusTestBase {
+    private final static GraknSession session = mock(GraknSession.class);
```
aelred: same here with the session in the super-class

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractComputeQuery.java
```diff
@@ -171,7 +171,11 @@ void getAllSubTypes() {
 
     GraknComputer getGraphComputer() {
         if (graknComputer == null) {
-            graknComputer = Grakn.session(url, keySpace).getGraphComputer();
+            if(tx.isPresent()){
+                graknComputer = tx.get().session().getGraphComputer();
+            } else {
+                graknComputer = Grakn.session(url, keySpace).getGraphComputer();
+            }
```
aelred: This class can probably be tidied some more. I don't think we need the `keySpace` and `url` fields for instance: they can just be retrieved from the `tx`.

I also suspect that this `else` should just throw. I think in all cases where the `tx` is absent, the `url` and `keyspace` will both be null (another reason to make them no longer fields).

grakn-test-tools/src/main/java/ai/grakn/util/GraknSessionInMemory.java
```diff
@@ -0,0 +1,75 @@
...
+import ai.grakn.GraknSession;
+import ai.grakn.GraknTx;
+import ai.grakn.GraknTxType;
+import ai.grakn.Keyspace;
+import ai.grakn.exception.GraknTxOperationException;
+
+import java.util.Properties;
+
+/**
+ * An in memory {@link GraknSession} which is used by {@link SampleKBLoader} to load test KBs
```
aelred: This is a bit odd, but I can't think of another way so ¯\\\_(ツ)_/¯

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractComputeQuery.java
fppt: Ya I did not want to take that risk without checking with @duckofyork . Is it possible to rely only on the transaction here?

grakn-test-tools/src/main/java/ai/grakn/util/GraknSessionInMemory.java
fppt: ¯\_(ツ)_/¯

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractComputeQuery.java
duckofyork: Yes, we should remove `url` and `keySpace.`

fppt: Then I will always use the provided transaction

fppt:
@aelred re-review please

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractComputeQuery.java
```diff
@@ -174,7 +173,7 @@ GraknComputer getGraphComputer() {
             if(tx.isPresent()){
                 graknComputer = tx.get().session().getGraphComputer();
             } else {
-                graknComputer = Grakn.session(url, keySpace).getGraphComputer();
+                throw new IllegalStateException("Transaction has not been provided. Cannot initialise graph computer");
             }
```
aelred: You can kill the `url` and `keySpace` fields from this class now!

fppt:
@aelred Third's the charm!

PR #2354, opened 2017-11-09
Assets added to Docs repository

excerebrose:
Simple addition of -
geogrotesque font
favicon.ico


excerebrose:
@marco-scoppetta yup, they have been tested using rake serve and manually as well

PR #2359, opened 2017-11-10
Cleanup analytics tests

fppt:


grakn-test-tools/src/main/java/ai/grakn/factory/GraknSessionTest.java
```diff
@@ -0,0 +1,85 @@
...
+ *     A {@link ai.grakn.GraknSession} used for testing which bypasses the need to hit the REST controller responsible
+ *     for feeding in the appropriate config file.
+ *
+ *     This session can be used to provide {@link ai.grakn.GraknTx}'s dependent on either a TinkerGraph or
+ *     JanusGraph
+ * </p>
+ *
+ * @author Filipe Peliz Pinto Teixeira
+ */
+public class GraknSessionTest extends GraknSessionImpl{
```
aelred: I think this needs a different name. I assumed this would contain tests, but it's actually a utility class.

grakn-test/test-integration/src/test/java/ai/grakn/test/rule/EngineContext.java
```diff
@@ -161,7 +161,7 @@ public void before() throws Throwable {
         //
         // When using janus, add -Dgrakn.test-profile=janus
         //
-        // The reason is that the default configuration of Grakn uses the Janus factory while the default
+        // The reason is that the default configuration of Grakn uses the Janus session while the default
```
aelred: I think this change might have been a mistake

fppt: Yep

grakn-test-tools/src/main/java/ai/grakn/factory/GraknSessionTest.java
fppt: Hmmmm good point. . . Will have to think of one.

PR #2362, opened 2017-11-10
Fix SNB dependencies again

aelred:
This time I'm using an 🐘 so we'll see if SNB validation runs before merging

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -103,13 +103,6 @@ def buildGrakn() {
 }
 
 def shouldRunAllTests() {
-    /*
-              _
-       ______/ \-.   _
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -103,6 +103,13 @@ def buildGrakn() {
 }
 
 def shouldRunAllTests() {
+    /*
+              _
+       ______/ \-.   _
+    .*/return/*  o\_*/true/*
```
fppt: What's with the elephants? Why do they keep haunting me.

aelred:
@fppt merging now because yeah we really need to fix SNB already

PR #2365, opened 2017-11-10
Make SNB use test log config

aelred:


aelred:
@fppt this has already passed everything that matters, so good to merge

PR #2369, opened 2017-11-13
Disable very slow query in SNB

aelred:


aelred:
@fppt can I request a merge despite the failure? 👍 
(it's unrelated I promise)

PR #2372, opened 2017-11-13
Clean up Jenkinsfile

aelred:
This one will make more sense commit-by-commit


aelred:
!ci - not sure I understand what happened

aelred:
@fppt all ready to review! This one might need to be read commit-by-commit to understand...

Jenkinsfile
```diff
@@ -52,12 +41,12 @@ def slackGithub(String message, String color = null) {
     slackSend channel: "#github", color: color, message: formattedMessage
 }
 
-def runIntegrationTest(String workspace, String moduleName) {
-    String modulePath = "${workspace}/grakn-test/${moduleName}"
+def runIntegrationTest(String moduleName) {
+    String modulePath = "${pwd()}/grakn-test/${moduleName}"
```
fppt: Kind of find `workspace` easier to work with. I hate "embedded paths" but that's just me.

Jenkinsfile
```diff
@@ -133,28 +115,24 @@ def shouldDeployLongRunningInstance() {
 }
 
 def mvn(String args) {
-    sh "mvn ${args}"
+    sh "mvn --batch-mode ${args}"
```
fppt: Why the addition of `--batch-mode`?

Jenkinsfile
```diff
@@ -266,3 +252,12 @@ if (shouldRunAllTests()) {
         slackGithub "Build Success", "good"
     }
 }
+
+try {
+    runBuild()
+} catch (Exception e) {
+    node {
+        slackGithub "Build Failure", "danger"
```
fppt: Fair point. Danger is afoot.

Jenkinsfile
aelred: It tells maven that we're not in an interactive shell, so it shouldn't ever prompt, or print fancy download bars that refresh or anything.

Jenkinsfile
aelred: Yeah, I'm not sure which is better. I'll try changing it back.

aelred:
@fppt all ready 👍

PR #2373, opened 2017-11-13
Fix some resolution planning issues

kasper-piskorski:
This PR solves an issue of omittance of value predicates when doing planning. Additionally it fixes a problem with unifying subbed atoms with themselves. However, this PR will make some of the queries slower due to the fact that it groups relation with same type together. This is fixed here: 

https://github.com/graknlabs/grakn/pull/2382


Jenkinsfile
```diff
@@ -108,6 +108,41 @@ def buildGrakn() {
...
+      `>>%%%%>>> `:::`%%%%%%%%%%>;.     ;;>>%%%%%                   ;>%%'
+       `>>%%%%>>, `::::`%%%%%%%%%%>,   .;>>%%%%%'                   ;>%'
+        `>>%%%%>>, `:::::`%%%%%%%%%>>. ;;>%%%%%%                    ;>%,
+         `>>%%%%>>, :::::::`>>>%%%%>>> ;;>%%%%%'                     ;>%,
+          `>>%%%%>>,::::::,>>>>>>>>>>' ;;>%%%%%                       ;%%,
+            >>%%%%>>,:::,%%>>>>>>>>'   ;>%%%%%.                        ;%%
+             >>%%%%>>``%%%%%>>>>>'     `>%%%%%%.
+             >>%%%%>> `@@a%%%%%%'     .%%%%%%%%%.
+             `a@@a%@'    `%a@@'       `a@@a%a@@a'
+     */
```
aelred: I need to remove this before we merge

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/AtomicBase.java
```diff
@@ -82,8 +83,13 @@ protected AtomicBase(AtomicBase a) {
     @Override
     public VarPattern getPattern(){ return atomPattern;}
 
+    protected Pattern combinedPattern(){ return atomPattern;}
```
aelred: Maybe this should be `createCombinedPattern()` or something? We need to distinguish it from `getCombinedPattern()` below.

grakn-graql/src/test/java/ai/grakn/graql/internal/hal/HALBuilderTest.java
```diff
@@ -115,7 +115,7 @@ public void whenExplainInferred_returnsLinkedExplanation() {
             String specificQuery = "match " +
                     "$x id " + cousin1 + ";" +
                     "$y id " + cousin2 + ";" +
-                    "(cousin: $x, cousin: $y) isa cousins; get;";
+                    "(cousin: $x, cousin: $y) isa cousins; limit 1;get;";
```
aelred: ahhhh... that's why this was slow :P

kasper-piskorski:
@aelred ready to merge?

PR #2375, opened 2017-11-14
Disable failing SNB query

aelred:


aelred:
@fppt quick review. This will let us occasionally pass SNB

PR #2377, opened 2017-11-14
Switch to Jackson For Json Serialisation + Other REST cleanups

fppt:


fppt:
@pluraliseseverythings @aelred this is only the first pass of Jacksonizing everything. My only concern is the need for classes like `KeyspaceResponse`. This class exists so we don't push the Jackson dependency down the stack.  

But I am not entirely sure yet if this duplication is worth it.

Imagine when we have `ConceptResponse` for example ?

Just need an opinion on this first pass.

grakn-core/src/main/java/ai/grakn/engine/TaskId.java
```diff
@@ -49,6 +50,6 @@ public static TaskId generate() {
      * Get the string value of the task ID
      */
     @CheckReturnValue
-    @JsonProperty("value")
+    @JsonValue
```
aelred: We might need to change `getValue()` into just `value()`, otherwise the JSON representation will have a key `"getValue": "123"`.

grakn-engine/src/main/java/ai/grakn/engine/controller/response/KeyspaceResponse.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * <p>
+ *     Response object representing {@link ai.grakn.Keyspace}
+ * </p>
+ *
+ * @author Filipe Peliz Pinto Teixeira
+ */
+@AutoValue
+public abstract class KeyspaceResponse {
```
aelred: It looks like `grakn-core` already has a Jackson dependency (since `TaskId` seems to be using it). Is this class definitely necessary in that case?

grakn-engine/src/main/java/ai/grakn/engine/controller/response/KeyspaceResponse.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @CheckReturnValue
+    @JsonValue
+    public abstract String getValue();
+
+    @CheckReturnValue
+    public static KeyspaceResponse of(String value){
+        return new AutoValue_KeyspaceResponse(value);
```
aelred: This class should strictly be a wrapper around `Keyspace` itself, not around the enclosed `String`.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/TaskState.java
```diff
@@ -116,22 +116,22 @@ public static TaskState of(TaskId id) {
...
-        return new TaskState(null, null, null, id, null, taskStatus);
+        return new TaskState(null, null, null, id.getValue(), null, taskStatus);
     }
 
     @JsonCreator
     public TaskState(@JsonProperty("taskClass") Class<?> taskClass,
             @JsonProperty("creator") String creator,
             @JsonProperty("schedule") TaskSchedule schedule,
-            @JsonProperty("id") TaskId id,
+            @JsonProperty("id") String id,
```
aelred: This seems a step backwards! Why do we have to do it?

grakn-engine/src/main/java/ai/grakn/engine/util/EngineID.java
```diff
@@ -65,27 +67,8 @@ public static EngineID me() {
...
-    @Override
-    public int hashCode() {
-        return value.hashCode();
-    }
-
     @Override
     public String toString() {
-        return escapeJava(value);
+        return escapeJava(getValue());
     }
```
aelred: I would just delete this `toString` completely. `AutoValue` makes a pretty good one.

grakn-core/src/main/java/ai/grakn/engine/TaskId.java
fppt: Really? I thought that `@JsonValue` would serialise the whole object as a String? I will make the change anyway. Not a fan of getters

grakn-engine/src/main/java/ai/grakn/engine/controller/response/KeyspaceResponse.java
fppt: Well that's the main design decision. Should we embed the annotations or should we create new objects rep represent them. Think of the concepts for example?

I keep swinging back and forth on this one.

grakn-engine/src/main/java/ai/grakn/engine/controller/response/KeyspaceResponse.java
fppt: If we keep this class then I will make the change.

grakn-core/src/main/java/ai/grakn/engine/TaskId.java
aelred: Whoops, you're right!

> ... indicates that results of the annotated "getter" method ... is to be used as the single value to serialize for the instance.

https://fasterxml.github.io/jackson-annotations/javadoc/2.0.0/com/fasterxml/jackson/annotation/JsonValue.html

grakn-engine/src/main/java/ai/grakn/engine/controller/response/KeyspaceResponse.java
aelred: Right now we're doing a hybrid approach, where `TaskId` are using the original approach and `Keyspace` isn't. I don't know which is right yet, but I don't think we should mix it up unless we have to.

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/TaskState.java
fppt: Serialisation issues within RedisQ due to varying representations. In order for Jackson to build the `TaskId` it needs this:

```
"id" : {
    "value" : "blab"
}
```

BUT what we actually use almost everywhere is:

```
"id" : "blab"
```

I.e. we don't use the embedded object just the string literal.

grakn-engine/src/main/java/ai/grakn/engine/util/EngineID.java
fppt: I will try kill it then. I thought we would need the `escapeJava ` part

grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/TaskState.java
aelred: Hmm... is this where the `@JsonValue` annotation saves us? Now that I understand what it does :P

aelred: Update: yeah, I definitely think we can solve this. You just have to update the `TaskId#of` factory to use `@JsonValue` instead of `@JsonProperty`.

fppt: No `JsonValue` only serialises one way. You would have to write a custom serialiser which kind of defeats the purpose of all this.

fppt: Just tried it. No go.

grakn-engine/src/main/java/ai/grakn/engine/controller/response/KeyspaceResponse.java
fppt: > I keep swinging back and forth on this one.

As is evident  by those issues. In `TaskId ` I can "justify" it because it exists within the context of engine where the Jackson dependency is supposed to live.

With the `keyspace` object that is not the case.

So ya not really sure. I am leaning more toward embedding them in the original objects but the affect on concepts may be kind of shitty. . . .

PR #2378, opened 2017-11-14
Add more information when GraknClient fails

aelred:


aelred:
@fppt this passed with the elephant so we don't need to wait for tests again

fppt:
those elephants are getting more and more elaborate @aelred

PR #2379, opened 2017-11-14
Reveal shard count

fppt:
@duckofyork this is the method you need. `graknTx.admin().shardingThreshold()`

### DIFF
--- a/grakn-kb/src/main/java/ai/grakn/kb/internal/GraknTxAbstract.java
+++ b/grakn-kb/src/main/java/ai/grakn/kb/internal/GraknTxAbstract.java
@@ -204,7 +204,7 @@ public abstract class GraknTxAbstract<G extends Graph> implements GraknTx, Grakn
 
     @Override
     public long shardingThreshold(){
-        return Long.parseLong(session().config().getProperty(GraknConfigKey.SHARDING_THRESHOLD.name()));
+        return Long.parseLong(session().config().get(GraknConfigKey.SHARDING_THRESHOLD.name()).toString());
     }
 
     public TxCache txCache() {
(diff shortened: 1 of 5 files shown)


duckofyork:
Thanks! @fppt

grakn-kb/src/main/java/ai/grakn/kb/internal/GraknTxAbstract.java
```diff
@@ -201,6 +202,11 @@ public Keyspace keyspace() {
         return session().keyspace();
     }
 
+    @Override
+    public long shardingThreshold(){
+        return Long.parseLong(session().config().get(GraknConfigKey.SHARDING_THRESHOLD.name()).toString());
```
aelred: Yeah, we really need to move `GraknEngineConfig` out of `grakn-engine`...

fppt:
@aelred what? The method does not exist yet which is why this PR exists.

fppt:
@duckofyork what needs to change to use this method?

duckofyork:
@aelred  @fppt I'll check out this branch and make the change

grakn-kb/src/test/java/ai/grakn/kb/internal/GraknTxTest.java
```diff
@@ -131,6 +131,11 @@ public void whenGettingSubTypesFromRootMeta_IncludeAllTypes(){
     }
 
     @Test
+    public void whenGettingTheShardingThreshold_TheCorrectValueIsReturned(){
+        assertEquals(100000L, tx.admin().shardingThreshold());
```
duckofyork: so we are not using the config in grakn.properties?

fppt: Technically this IS the config from grakn.properties. Just in a very roundabout way

fppt:
@aelred re-review please?

PR #2380, opened 2017-11-14
Get rid of REST doc generation which we are not using anymore

fppt:


grakn-engine/pom.xml
```diff
@@ -221,47 +221,6 @@
                     </execution>
                 </executions>
             </plugin>
-            <plugin>
```
aelred: Also you can remove the plugin above :D

grakn-engine/pom.xml
```diff
@@ -221,47 +221,6 @@
...
-                            <arguments>
-                                <argument>-i</argument>
-                                <argument>${project.build.directory}/jaxrs-raml/engine-apis.raml</argument>
-                                <argument>-o</argument>
-                                <argument>${project.build.directory}/jaxrs-raml/engine-apis.html</argument>
-                            </arguments>
-                        </configuration>
-                    </execution>
-                </executions>
-            </plugin>
```
aelred: AND THE PLUGIN MANAGEMENT SECTION BELOW :D :D

fppt:
@aelred re-review please?

PR #2381, opened 2017-11-15
Investigate SNB ShortQuery7 [DO NOT MERGE]

aelred:


aelred:
Investigation over. PP running during loading creates ghosts, which SQ7 finds while navigating other entities.

PR #2382, opened 2017-11-15
Make atom-var-property mapping explicit

kasper-piskorski:


kasper-piskorski:
@aelred review please :)

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
```diff
@@ -139,6 +140,11 @@ private RelationshipAtom(RelationshipAtom a) {
     }
 
     @Override
+    public Stream<VarProperty> getVarProperties() {
+        return getCombinedPattern().admin().varPatterns().stream().flatMap(vp -> vp.getProperties(RelationshipProperty.class));
+    }
```
aelred: Since the implementations only differ on the class used, maybe we can instead add an abstract method `Class<? extends VarProperty> getPropertyClass()` and have just one implementation of `getVarProperties()`.

PR #2383, opened 2017-11-15
More jacksonising + changing representation to agreed upon docs

fppt:
@aelred @pluraliseseverythings we have an interesting side effect of our object wrappers. I will give you an example.

I changed the `Keyspace` json representation to be:

```
{
      "@id": "/kb/genealogy",
      "name": "genealogy",
      "types": "/kb/genealogy/type",
      "roles": "/kb/genealogy/role",
      "rules": "/kb/genealogy/rule",
      "graql": "/kb/genealogy/graql"
}
```

and that works very well. NOW, get getting a list of all the Keyspaces this is the representation we agreed on: 

```
{
  "@id": "/kb",
  "keyspaces": [
    {
      "@id": "/kb/genealogy",
      "name": "genealogy",
      "types": "/kb/genealogy/type",
      "roles": "/kb/genealogy/role",
      "rules": "/kb/genealogy/rule",
      "graql": "/kb/genealogy/graql"
    }
  ],
  "keyspace": "/kb/{keyspace}"
}
```

To achieve this I created the `Keyspaces` object which wraps the list plus the additional metadata. 

Now we get to an interesting problem. Serialising a json representation into one of our wrappers either `Keyspace` or `Keyspaces` throws the following exception (for example):

```
Exception in thread "main" com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "types" , not marked as ignorable
```

This is because the serialisation does not use the entire Json representation. For example when Serialising:

```
{
  "@id": "/kb",
  "keyspaces": [
    {
      "@id": "/kb/genealogy",
      "name": "genealogy",
      "types": "/kb/genealogy/type",
      "roles": "/kb/genealogy/role",
      "rules": "/kb/genealogy/rule",
      "graql": "/kb/genealogy/graql"
    }
  ],
  "keyspace": "/kb/{keyspace}"
}
```

The only thing we need is the name `"name": "genealogy",` the rest is all wrapped up metadata so we get the above exception. I workaround this by saying that partial usage of the Json representation is okay with `@JsonIgnoreProperties(ignoreUnknown = true)` I don't know if there is a neater way to do this BUT if there is please to tell me.

For the moment I will be continuing down this route.


aelred:
@fppt the "ignore properties" seems totally acceptable to me. If we think it's too permissive, we can selectively list them, like `@JsonIgnoreProperties({"@id", "types", "roles", "rules", "graql"})`.

fppt:
We likely to see them everywhere because most of our wrappers will serialise with only a part of the json representation. Just something to be aware of as I proceed

fppt:
**Note:** At this stage in the PR. List of keyspaces has changed from:

```
{
    keyspaces: ["a", "b", "c"]
}
```

To 

```
{
  "@id": "/kb",
  "keyspaces": [
    {
      "@id": "/kb/genealogy",
      "name": "genealogy",
      "types": "/kb/genealogy/type",
      "roles": "/kb/genealogy/role",
      "rules": "/kb/genealogy/rule",
      "graql": "/kb/genealogy/graql"
    }
  ],
  "keyspace": "/kb/{keyspace}"
}
```

fppt:
@marco-scoppetta Are there any dashboard breaking changes in this PR? 

I have changed the response of ConceptController which i believe you are not using but I also changed the response of the keyspace list. As stated above. I think that will break dashboard?

grakn-core/src/main/java/ai/grakn/util/Schema.java
```diff
@@ -133,7 +134,8 @@ public static boolean isMetaLabel(Label label) {
         ATTRIBUTE(Attribute.class),
 
         //Internal
-        SHARD(Vertex.class);
+        SHARD(Vertex.class),
+        CONCEPT(Concept.class);
```
aelred: New base type? :O

grakn-engine/src/main/java/ai/grakn/engine/controller/api/ConceptController.java
```diff
@@ -0,0 +1,104 @@
...
+
+    public ConceptController(EngineGraknTxFactory factory, Service spark,
+                             MetricRegistry metricRegistry){
+        this.factory = factory;
+        this.conceptIdGetTimer = metricRegistry.timer(name(ConceptController.class, "concept-by-identifier"));
+
+        spark.get(WebPath.CONCEPT_ID,  this::getConceptById);
+        spark.get(WebPath.TYPE_LABEL,  this::getSchemaByLabel);
+        spark.get(WebPath.RULE_LABEL,  this::getSchemaByLabel);
+        spark.get(WebPath.ROLE_LABEL,  this::getSchemaByLabel);
```
aelred: These are all the same right now. I guess we need to make the wrong one give some kind of 400 response.

grakn-core/src/main/java/ai/grakn/engine/Jacksonisable.java
```diff
@@ -0,0 +1,36 @@
...
+ * </p>
+ *
+ * <p>
+ *     Use to pass the necessary annotation to all objects which we use Jackson for serliasation and deserilisation
+ * </p>
+ *
+ * @author Filipe Peliz Pinto Teixeira
+ */
+@JsonIgnoreProperties(value={ "@id", "roles", "types", "graql", "rules" }, allowGetters=true)
+public interface Jacksonisable {
```
aelred: Can we move `@id` into this interface?

grakn-engine/src/main/java/ai/grakn/engine/controller/response/Role.java
```diff
@@ -0,0 +1,57 @@
...
+    @JsonProperty
+    public abstract Set<Link> relationships();
+
+    @JsonProperty
+    public abstract Set<Link> roleplayers();
+
+    @JsonCreator
+    public static Role create(
+            @JsonProperty("id") ConceptId id,
+            @JsonProperty("@id") Link selfLink,
```
aelred: Maybe properties like `@id` can be derived instead of provided in the constructors.

grakn-engine/src/main/java/ai/grakn/engine/controller/response/Thing.java
```diff
@@ -0,0 +1,42 @@
...
+public abstract class Thing extends Concept {
+
+    @JsonProperty
+    public abstract Set<Link> attributes();
+
+    @JsonProperty
+    public abstract Set<Link> keys();
+
+    @JsonProperty
+    public abstract Set<Link> relationships();
```
aelred: These links need to include the role (like `RolePlayer`)

grakn-engine/src/main/java/ai/grakn/engine/controller/response/AttributeType.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * <p>
+ *     Wrapper class for {@link ai.grakn.concept.AttributeType}
+ * </p>
+ *
+ * @author Filipe Peliz Pinto Teixeira
+ */
+@AutoValue
+public abstract class AttributeType extends Type{
```
aelred: This needs a datatype, plus some other stuff (like regex).

grakn-engine/src/main/java/ai/grakn/engine/controller/response/Link.java
```diff
@@ -0,0 +1,74 @@
...
+        return new AutoValue_Link(id);
+    }
+
+    public static Link create(ai.grakn.concept.Thing thing){
+        String id = REST.resolveTemplate(
+                WebPath.CONCEPT_LINK,
+                thing.keyspace().getValue(),
+                Schema.BaseType.CONCEPT.name().toLowerCase(Locale.getDefault()),
+                thing.getId().getValue());
+        return new AutoValue_Link(id);
```
aelred: Boring thing, but these should delegate to `create(String)` above instead of `AutoValue_Link`

grakn-engine/src/main/java/ai/grakn/engine/controller/response/Link.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public static Link create(ai.grakn.concept.SchemaConcept schemaConcept){
+        String type = Schema.BaseType.TYPE.name().toLowerCase(Locale.getDefault());
+        if(schemaConcept.isRule()){
+            type = Schema.BaseType.RULE.name().toLowerCase(Locale.getDefault());
+        } else if(schemaConcept.isRole()){
+            type = Schema.BaseType.ROLE.name().toLowerCase(Locale.getDefault());
```
aelred: I guess we can simplify this to `if (concept.isType()) { TYPE } else { concept.baseType() }`

grakn-engine/src/main/java/ai/grakn/engine/controller/response/Rule.java
```diff
@@ -0,0 +1,57 @@
...
+ *     Wrapper class for {@link ai.grakn.concept.Rule}
+ * </p>
+ *
+ * @author Filipe Peliz Pinto Teixeira
+ */
+@AutoValue
+public abstract class Rule extends SchemaConcept{
+
+    @JsonProperty
+    public abstract String when();
```
aelred: These _could_ be `Pattern`, but there may be issues with that...

grakn-engine/src/test/java/ai/grakn/engine/controller/SystemControllerTest.java
```diff
@@ -98,17 +104,19 @@ public void whenCallingPutKBEndpoint_Return200_AndConfigInBody() throws JsonProc
     }
 
     @Test
-    public void whenCallingPutKBEndpointOnNonExistentKeyspace_KeyspaceAppearsInList() {
+    public void whenCallingPutKBEndpointOnNonExistentKeyspace_KeyspaceAppearsInList() throws IOException {
         RestAssured.put("/kb/myks");
-        when().get("/kb").then().body("", hasItem("myks"));
+        String content = when().get("/kb").thenReturn().body().asString();
+        Keyspaces keyspaces = objectMapper.readValue(content, Keyspaces.class);
+        assertThat(keyspaces.keyspaces(), not(empty()));
```
aelred: Apparently Rest Assured should handle this for us: https://github.com/rest-assured/rest-assured/wiki/Usage#object-mapping

```
Keyspaces keyspaces = get("/kb").as(Keyspaces.class);
```

grakn-engine/src/test/java/ai/grakn/engine/controller/api/ConceptControllerTest.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    //We can't use the class of the wrapper because it will be an AutoValue class
+    private static void assertExists(Concept wrapper, Class clazz) throws IOException {
+        String request = wrapper.selfLink().id();
+        Response response = RestAssured.when().get(request);
+        assertEquals(SC_OK, response.statusCode());
+        String content = response.thenReturn().body().asString();
+
+        assertEquals(wrapper, objectMapper.readValue(content, clazz));
```
aelred: Same as before: `assertEquals(wrapper, RestAssured.get(request).as(clazz));`

grakn-kb/src/main/java/ai/grakn/kb/internal/concept/RelationshipImpl.java
```diff
@@ -118,6 +122,21 @@ public Relationship attributeRelationship(Attribute attribute) {
...
+
+        if(!attributeTypesSet.isEmpty()){
+            keyTypes = Sets.intersection(attributeTypesSet, keyTypes);
+        }
+
+        if(keyTypes.isEmpty()) return Stream.empty();
+
+        AttributeType[] keyTypesArray = keyTypes.toArray(new AttributeType[keyTypes.size()]);
+        return readFromReified((relationReified) -> relationReified.attributes(keyTypesArray));
+    }
```
aelred: This is duplicated from `ThingImpl`. Can you instead delegate to the reified instance? If the relationship is not reified, then it has no keys... I think.

grakn-engine/src/main/java/ai/grakn/engine/controller/response/ConceptBuilder.java
```diff
@@ -0,0 +1,168 @@
...
+public class ConceptBuilder {
+
+    /**
+     * Takes a {@link ai.grakn.concept.Concept} and returns the equivalent response object
+     *
+     * @param concept The {@link ai.grakn.concept.Concept} to be converted into a response object
+     * @return the response object wrapper {@link Concept}
+     */
+    public static <X extends Concept> Optional<X> build(ai.grakn.concept.Concept concept){
+        if(concept == null) return Optional.empty();
```
aelred: This looks like the only case where it returns `Optional.empty()`. I think it's clearer to handle it at the call-site:

```
repr = Optional.ofNullable(concept).map(ConceptBuilder::build);
```

But if you want to keep the null check, remember to mark the argument as `@Nullable` 👍

grakn-engine/src/main/java/ai/grakn/engine/controller/response/ConceptBuilder.java
```diff
@@ -0,0 +1,168 @@
...
+        } else if (concept.isThing()) {
+            response = buildThing(concept.asThing());
+        } else {
+            throw GraknBackendException.convertingUnknownConcept(concept);
+        }
+
+        return Optional.of((X) response);
+    }
+
+    //TODO: This will scale poorly with super nodes. Need to introduce some sort of paging maybe?
```
aelred: Yeah... we might need a paged endpoint for relationships and roleplayers :(

grakn-core/src/main/java/ai/grakn/util/Schema.java
fppt: It's used in the link class so I don't have to hard code the string "concept" there. The other way would be to have more Paths on `REST.WebPath` which would be redundant. This feels like the least evil appraoch.

grakn-engine/src/main/java/ai/grakn/engine/controller/response/AttributeType.java
fppt: Yep you are right.

grakn-core/src/main/java/ai/grakn/util/Schema.java
aelred: That's fair! I'd definitely have a comment to say that no vertex will _really_ have this type.

grakn-engine/src/main/java/ai/grakn/engine/controller/response/Rule.java
fppt: That would be leaking Jackson annotations into more core objects and I am not a fan of that personally. `Keyspace` is okay because it's pretty much a wrapper by itself but `Pattern` contains a bit too much low level logic to allow the leak in my opinion.

grakn-core/src/main/java/ai/grakn/engine/Jacksonisable.java
fppt: This will lead to a fun generic but I will give it a try.

fppt: Actually scratch that. I am not sure how to do it because they all use it in different contexts. For example `ai.grakn.Keyspace` uses it but does not have a `@id` json property. It IS the `@id` property.

On the other hand, `ai.grakn.engine.controller.response.Concept` and `ai.grakn.engine.controller.response.Keyspace` use it and contain `@id` as a json property.

All 3 require the annotation provided by `Jacksonisable`. In other words if you want `@id` to be common place then another intr=erface might be needed and I am not sure we will gain much from that.

grakn-engine/src/main/java/ai/grakn/engine/controller/response/Role.java
fppt: Can you elaborate on this? What do you mean?

fppt:
@aelred re-review please.

grakn-engine/src/main/java/ai/grakn/engine/controller/response/Role.java
aelred: The `@id` `selfLink()` can (usually?) be derived from other properties, so you don't necessarily have to provide it in the constructor - just implement the method.

fppt:
@aelred this was merged (not complaining because I can continue working) without @marco-scoppetta confirming if these changes broke dashboard. So central may have a broken dashboard now.

PR #2385, opened 2017-11-16
Add notifications when build fails

aelred:


aelred:
@burukuru yes, I think so!

PR #2387, opened 2017-11-16
Fix benchmarks

aelred:


aelred:
@fppt this one passed pre-🐘 so good to merge anytime

PR #2390, opened 2017-11-16
Reduce bash

micheleorsi:


micheleorsi:
I moved a lot of functionalities from bash to java. 
For the moment I didn't add any more tests because this step already took a lot of time (problem with class-path, grakn dir, grakn config, etc ..) .. 
So no clean code here .. At all! 😄

grakn-dist/src/main/java/ai/grakn/dist/DistGrakn.java
```diff
@@ -0,0 +1,719 @@
...
+                e.printStackTrace();
+            }
+        }
+        System.out.println("FAILED!");
+        System.out.println("Unable to start Storage");
+    }
+
+    private void startQueue() {
+        queueStart();
+    }
```
aelred: We should probably just inline this method 😛

grakn-dist/src/main/java/ai/grakn/dist/DistGrakn.java
```diff
@@ -0,0 +1,719 @@
...
+
+    private void queueStartProcess() {
+        System.out.print("Starting Queue...");
+        System.out.flush();
+        OutputCommand operatingSystem = processHandler.executeAndWait(new String[]{
+                "/bin/sh",
+                "-c",
+                "uname"
+        },null,null);
+        String queueBin = operatingSystem.output.trim().equals("Darwin") ? "redis-server-osx" : "redis-server-linux";
```
aelred: This is duplicated somewhere else, we should make a method for it.

grakn-dist/src/main/java/ai/grakn/dist/DistGrakn.java
```diff
@@ -0,0 +1,719 @@
...
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+            if (graknPid.trim().isEmpty()) {
+                return;
+            }
+            processHandler.kill(graknPid);
+        }
+
+        while(true) {
```
aelred: We could use the seldom-seen `do .. while` loop in these places (and the other `stop` methods):

```
do {
    ...
} while (outputCommand.exitStatus != 0);

System.out.println("SUCCESS");
```

grakn-dist/src/main/java/ai/grakn/dist/DistGrakn.java
```diff
@@ -0,0 +1,719 @@
...
+                Thread.sleep(WAIT_INTERVAL_S * 1000);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
+        }
+    }
+
+    private void queueStopProcess() {
+        String queuePid="";
+        if(Files.exists(Paths.get(QUEUE_PID))) {
```
aelred: I think this would be clearer flipped:

```
if (!Files.exists(Paths.get(QUEUE_PID))) {
    return;
}

...

if (queuePid.isEmpty()) {
    return;
}
```

grakn-dist/src/main/java/ai/grakn/dist/DistGrakn.java
```diff
@@ -0,0 +1,719 @@
...
+                "stop [grakn|queue|storage]   Stop Grakn (or optionally, only one of the component)\n" +
+                "status                         Check if Grakn is running\n" +
+                "clean                          DANGEROUS: wipe data completely\n" +
+                "\n" +
+                "Tips:\n" +
+                "- Start Grakn with 'grakn server start'\n" +
+                "- Start or stop only one component with, e.g. 'grakn server start storage' or 'grakn server stop storage', respectively\n");
+    }
+
+    private void graknServerStatus(String arg2) {
```
aelred: `graknServerStatus(String verboseFlag)`

grakn-dist/src/main/java/ai/grakn/dist/DistGraql.java
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+    }
+
+    private void migrate(String[] args) {
+        String arg0 = args.length > 0 ? args[0] : "";
+
+        switch (arg0) {
+            case "csv":
+                CSVMigrator.main(Arrays.copyOfRange(args, 1, args.length));
```
aelred: I'd extract the `Arrays#copyOfRange` call out.

grakn-dist/src/main/java/ai/grakn/dist/DistGraql.java
```diff
@@ -0,0 +1,118 @@
...
+                ai.grakn.migration.export.Main.main(Arrays.copyOfRange(args, 1, args.length));
+                break;
+            case "sql":
+                SQLMigrator.main(Arrays.copyOfRange(args, 1, args.length));
+                break;
+            case "xml":
+                XmlMigrator.main(Arrays.copyOfRange(args, 1, args.length));
+                break;
+            default:
+                XmlMigrator.main(new String[]{"-h"});
```
aelred: Shouldn't this just print the regular `help` below?

grakn-dist/src/main/java/ai/grakn/dist/DistGraql.java
```diff
@@ -0,0 +1,118 @@
...
+                "COMMAND:\n" +
+                "console  Start a REPL console for running Graql queries. Defaults to connecting to http://localhost\n" +
+                "migrate  Run migration from a file\n" +
+                "version  Print Grakn version\n" +
+                "help     Print this message");
+
+    }
+
+    private void version() {
+        GraqlShell.main(new String[]{"--v"});
```
aelred: Another case for an extracted "version" method.

grakn-dist/src/main/java/ai/grakn/dist/OutputCommand.java
```diff
@@ -0,0 +1,34 @@
...
+
+package ai.grakn.dist;
+
+/**
+ *
+ * @author Michele Orsi
+ */
+class OutputCommand {
+    final String output;
+    final int exitStatus;
```
aelred: I'd add and use a method like `success()` to confirm if `exitStatus` is zero.

grakn-dist/src/main/java/ai/grakn/dist/ProcessHandler.java
```diff
@@ -0,0 +1,117 @@
...
+        return outputCommand;
+    }
+
+    String getPidFromFile(String fileName) {
+        String pid="";
+        if (Files.exists(Paths.get(fileName))) {
+            try {
+                pid = new String(Files.readAllBytes(Paths.get(fileName)),StandardCharsets.UTF_8).trim();
+            } catch (IOException e) {
+                // DO NOTHING
```
aelred: Return `@Nullable String` or an `Optional<String>` for this. Hiding errors scares me :(

grakn-dist/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1,2 @@
+# this is useful to enable mockito to mock final classes (https://github.com/mockito/mockito/wiki/What%27s-new-in-Mockito-2#unmockable)
+mock-maker-inline
```
aelred: Neat! 😮

grakn-dist/src/main/resources/logback.xml
```diff
@@ -0,0 +1,24 @@
...
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with Grakn. If not, see <http://www.gnu.org/licenses/gpl.txt>.
+  -->
+
+<configuration debug="false">
+
+    <statusListener class="ch.qos.logback.core.status.NopStatusListener" />
+    <root level="OFF"/>
+
+</configuration>
```
aelred: What's this config for? Does it replace the one in `conf/main`?

micheleorsi: Added an explanation in the file

grakn-dist/src/main/java/ai/grakn/dist/DistGraql.java
micheleorsi: I think you're right: it was my mistake of just copying from bash scripting ..

micheleorsi:
@aelred: I think I reviewed all your comments! 
Probably there is still place for improvements in the new Process handlers, but I would say that for the moment could be enough. I need all this things for the enterprise version, so once I did the first iteration also there I can improve even more ..

What do you think?

PR #2393, opened 2017-11-18
Added example files location to setup guide when brew install

Makopo:
I had hard time to find these sample files...


fppt:
Thank you for the contribution @Makopo

PR #2395, opened 2017-11-20
Handle failures whe measuring database in SNB

aelred:


aelred:
@fppt since this doesn't impact the tests I'm merging pre-review.

PR #2396, opened 2017-11-20
Move all dependency versions and exclusions into base pom

aelred:


aelred:
@fppt this adds a lot of shit to the bottom pom, so I'm not 100% on whether it's a good plan.

PR #2400, opened 2017-11-21
Fix and enable ShortQuery7

aelred:


aelred:
Turns out this wasn't the problem 👍

PR #2401, opened 2017-11-21
Concept Collections for REST

fppt:


grakn-engine/src/main/java/ai/grakn/engine/controller/api/ConceptController.java
```diff
@@ -104,4 +113,34 @@ private String getConcept(Response response, Keyspace keyspace, Function<GraknTx
             }
         }
     }
+
+    private String getTypes(Request request, Response response) throws JsonProcessingException {
+        Function<GraknTx, Stream<? extends ai.grakn.concept.Concept>> getter = (tx) -> {
+            Stream<? extends ai.grakn.concept.Concept> stream = tx.admin().getMetaEntityType().subs();
+            stream = Stream.concat(stream, tx.admin().getMetaRelationType().subs());
+            stream = Stream.concat(stream, tx.admin().getMetaAttributeType().subs());
+            return stream;
```
aelred: What about `(tx) -> tx.admin().getMetaConcept().subs()`?

grakn-engine/src/main/java/ai/grakn/engine/controller/response/ConceptBuilder.java
```diff
@@ -142,15 +145,23 @@ private static Role buildRole(ai.grakn.concept.Role role, Link selfLink, Link su
     }
 
     private static Rule buildRule(ai.grakn.concept.Rule rule, Link selfLink, Link sup, Set<Link> subs){
-        return Rule.create(rule.getId(), selfLink, rule.getLabel(), rule.isImplicit(), sup, subs,
-                rule.getWhen().toString(), rule.getThen().toString());
+        String when = null;
+        if(rule.getWhen() != null) when = rule.getWhen().toString();
+
+        String then = null;
+        if(rule.getThen() != null) then = rule.getThen().toString();
```
aelred: Damn meta-types :P

grakn-engine/src/main/java/ai/grakn/engine/controller/response/RelationshipType.java
```diff
@@ -51,7 +53,7 @@ public static RelationshipType create(
             @JsonProperty("plays") Set<Link> plays,
             @JsonProperty("attributes") Set<Link> attributes,
             @JsonProperty("keys") Set<Link> keys,
-            @JsonProperty("relates") Set<Link> relates){
+            @Nullable @JsonProperty("relates") Set<Link> relates){
```
aelred: I'm more in favour of using an empty set instead of `null` (unless they have different meanings).

grakn-engine/src/main/java/ai/grakn/engine/controller/api/ConceptController.java
fppt: Ah yes we can indeed! I forgot that we disconnected the meta role and meta rule

fppt:
@aelred re-review please

PR #2402, opened 2017-11-21
Changes needed for enterprise version

micheleorsi:


micheleorsi:
I realised we can reduce the amount of static method in the main partition, so I just reduced some of them ..

pom.xml
```diff
@@ -293,37 +293,37 @@
...
                 <groupId>ai.grakn</groupId>
                 <artifactId>migration-xml</artifactId>
-                <version>${project.version}</version>
+                <version>1.0.0-SNAPSHOT</version>
             </dependency>
             <dependency>
                 <groupId>ai.grakn</groupId>
                 <artifactId>migration-csv</artifactId>
-                <version>${project.version}</version>
+                <version>1.0.0-SNAPSHOT</version>
```
fppt: Why can't we use the property anymore?

micheleorsi: That's a very good point @fppt! It took me half a day to understand!

The problem is that if you inherit this project into another one (let's call it "enterprise") ..
```
<parent>
<groupId>ai.grakn</groupId>
<artifactId>grakn</artifactId>
<version>1.0.0-SNAPSHOT</version>
</parent>
```

and you want to give a new version to that project (let's say "0.0.1-SNAPSHOT") 
```
<groupId>ai.grakn.enterprise</groupId>
<artifactId>grakn-enterprise</artifactId>
<version>0.0.1-SNAPSHOT</version>
```

.. you end up in a situation where the `${project.version}` of the enterprise is resolved as `0.0.1-SNAPSHOT`, not only in the enterprise version but also in the open source module (that is the parent of the enterprise)!
And maven returns an error, saying that it cannot find `ai.grakn.migration-csv:0.0.1-SNAPSHOT`

fppt: Ahhh that makes sense. Thank you for the explanation @micheleorsi!

PR #2404, opened 2017-11-21
Add initial API spec

aelred:
Go [HERE](https://rebilly.github.io/ReDoc/?url=raw.githubusercontent.com/aelred/grakn/api-docs/docs/api.yaml) to see the docs for this PR LIVE!

Note: we're using Swagger 2.0, which is the older version of OpenAPI 3.0.

OpenAPI 3.0 is a proper web standard and is super fancy and new (from this year), but a lot of tools such as Dredd don't support it yet. I still have a few commented OpenAPI 3.0 features in the spec for when we switch.


docs/api.yaml
```diff
@@ -0,0 +1,798 @@
+swagger: '2.0'
+info:
+  title: Grakn
+  description: 'TODO: Add a description'
```
fppt: Can we do this todo now? Seems simple

aelred:
@fppt I added the dredd config so now it can! But we still need to add before/after hooks for it to work properly in some cases.

PR #2407, opened 2017-11-21
Introduce Type Instances Endpoint

fppt:


grakn-engine/src/main/java/ai/grakn/engine/controller/response/ConceptBuilder.java
```diff
@@ -52,6 +52,38 @@
...
+
+    public static Things buildThings(ai.grakn.concept.Type type, int offset, int limit){
+        Link selfLink = Link.createInstanceLink(type);
+        Link next = Link.createInstanceLink(type, offset + limit, limit);
+
+        int previousIndex = offset - limit;
+        if(previousIndex < 0) previousIndex = 0;
+        Link previous = Link.createInstanceLink(type, previousIndex, limit);
+        Set<Thing> things = type.instances().skip(offset).limit(limit).
+                map(ConceptBuilder::buildThing).collect(Collectors.toSet());
```
aelred: I guess we should put a TODO to remind us that this doesn't actually scale :P

fppt: Yep. . . :(

PR #2420, opened 2017-11-23
Change GraqlControllerResponse to Jackson Objects

fppt:
This is very high impact and will break dashboard for sure.


fppt:
@marco-scoppetta this PR is ready. HAL is dead now so we will need you to propagate the changes into dashboard when you have time.

The response objects are detailed here: https://app.swaggerhub.com/apis/aelred/Grakn/0.1#/

grakn-client/src/main/java/ai/grakn/client/QueryResponse.java
```diff
@@ -48,7 +50,13 @@ public Json getJsonResponse() {
...
         ArrayList<QueryResponse> result = new ArrayList<>();
         for (int i = 0; i < queries.size(); i++) {
-            result.add(new QueryResponse(queries.get(i), json.get(i)));
+            QueryResponse resp;
+            if(EMPTY_RESPONSE.equals(response)){
+                resp = new QueryResponse(queries.get(i), Json.nil());
+            } else {
+                resp = new QueryResponse(queries.get(i), json.get(i));
+            }
+            result.add(resp);
```
aelred: This is in a loop but the `if` condition doesn't seem to depend on the loop at all. Why did the old code fail?

grakn-engine/src/main/java/ai/grakn/engine/controller/response/MetaConcept.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * <p>
+ *     Special wrapper class for the top meta concept {@link ai.grakn.util.Schema.MetaSchema#THING};
+ * </p>
+ *
+ * @author Filipe Peliz Pinto Teixeira
+ */
+@AutoValue
+public abstract class MetaConcept extends SchemaConcept{
```
aelred: This can be more precise and `extend Type`.

grakn-engine/src/main/java/ai/grakn/engine/printer/JacksonPrinter.java
```diff
@@ -0,0 +1,117 @@
...
+        return entries.collect(Collectors.toMap(
+                entry -> graqlString(inner, entry.getKey()),
+                entry -> graqlString(inner, entry.getKey())
+        ));
+    }
+
+    @Override
+    public Object graqlString(boolean inner, Collection collection) {
+        Stream stream = collection.stream();
+        if(limitEmbedded.isPresent()) stream = stream.limit(limitEmbedded.get());
```
aelred: `limitEmbedded` here will mean we limit the number of results in queries and also do other bad stuff (like truncating the returned path in `compute path`). I'm pretty sure that's not what we want: @marco-scoppetta is probably the person to ask about what exactly `limitEmbedded` should do.

grakn-engine/src/test/java/ai/grakn/engine/controller/util/JsonConceptBuilder.java
```diff
@@ -0,0 +1,94 @@
...
+import ai.grakn.util.Schema;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import mjson.Json;
+
+import java.io.IOException;
+
+/**
+ * <p>
+ *     Takes a string representing a {@link ai.grakn.engine.controller.response.Concept} and serialises it into
+ *     it's correct object
```
aelred: I'd add an explanation (or just [the issue link](https://github.com/FasterXML/jackson-annotations/issues/88#issuecomment-219078857)) for why we can't do this the "proper" way.

grakn-engine/src/test/java/ai/grakn/engine/controller/util/JsonConceptBuilder.java
```diff
@@ -0,0 +1,94 @@
...
+ * </p>
+ *
+ * @author Filipe PEliz Pintio Teixeira
```
aelred: The hooves must make it difficult to type 😞

marco-scoppetta:
@fppt Ok, I will make it a priority for next week

grakn-graql/src/test/java/ai/grakn/graql/internal/printer/JsonPrinterTest.java
```diff
@@ -1,145 +0,0 @@
...
-import org.junit.ClassRule;
-import org.junit.Test;
-
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Optional;
-
-import static org.junit.Assert.assertEquals;
-
-public class JsonPrinterTest {
```
aelred: This class still exists! If we delete the test, we should delete `JsonPrinter` too.

grakn-client/src/main/java/ai/grakn/client/QueryResponse.java
fppt: because `json` in certain cases would only contain a single entry so when it tried to iterate over all entries with `i` it would get an index out of bounds exception.

fppt: I am an idiot. I can clean this up MUCH more. Hold on . . .

grakn-engine/src/test/java/ai/grakn/engine/controller/util/JsonConceptBuilder.java
fppt: Will link the issue

grakn-engine/src/test/java/ai/grakn/engine/controller/util/JsonConceptBuilder.java
fppt: Don't judge centaurs!

grakn-graql/src/test/java/ai/grakn/graql/internal/printer/JsonPrinterTest.java
fppt: Murder time!

fppt: Cancel the murder time will try to clean this up in a different PR.

grakn-engine/src/main/java/ai/grakn/engine/printer/JacksonPrinter.java
marco-scoppetta: @aelred is right, don't use it to limit results, they should be already limited because the limit x; is already specified in the graql string. `limitEmbedded` was used for limiting the "embedded" nodes we were embedding in a response e.g. if user asks for a relationship instance which has thousands of roleplayers (IHS Markit) we would limit the number of roleplayers to return in a single HAL object)

fppt: Alright I will kill it for the moment. Thanks guys!

fppt:
@aelred re-review please?

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
```diff
@@ -137,13 +130,6 @@ private String executeGraql(Request request, Response response) {
...
 
-            JacksonPrinter printer;
-            if(limitEmbedded == -1){
-                printer = JacksonPrinter.create();
-            } else {
-                printer = JacksonPrinter.create(limitEmbedded);
-            }
-
             response.status(SC_OK);
             return executeQuery(printer, tx, queryString, multiQuery, parser);
```
aelred: 🎵 super pickiness 🎵 
now `printer` is a field, you don't need to pass it 👍

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
```diff
@@ -119,6 +127,26 @@ public GraqlController(
         spark.exception(InvalidKBException.class, (e, req, res) -> handleError(422, e, res));
     }
 
+    @POST
+    @Path("/kb/{keyspace}/explain")
```
aelred: Does this endpoint also include the query results? _Maybe_ it should instead be a flag on `/kb/{keyspace}/graql`. Not totally sure.

grakn-engine/src/main/java/ai/grakn/engine/controller/ConceptController.java
```diff
@@ -89,9 +91,47 @@ public ConceptController(EngineGraknTxFactory factory, Service spark,
...
+
+            //If the concept was not found or is not a thing there are no results to return;
+            if(concept == null || !concept.isThing()){
+                response.status(SC_NOT_FOUND);
+                return "[]";
+            }
+
+            //Get the attributes including keys if needed
+            Stream<Attribute<?>> attributes;
+            if(isKey){
```
aelred: It might be clearer to pass in the method to get the attributes/keys rather than a boolean.

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
```diff
@@ -173,6 +175,9 @@ private String executeGraql(Request request, Response response) throws RetryExce
         //Define all anonymous variables in the query
         Optional<Boolean> defineAllVars = queryParameter(request, DEFINE_ALL_VARS).map(Boolean::parseBoolean);
 
+        //Used to check if serialisation of results is needed. When loading we skip this for the sake of speed
+        boolean skipSerialisation = parseBoolean(queryParameter(request, LOADING_DATA).orElse("false"));
```
aelred: Maybe we could tie this to the whether "batch" is set as the mode? Or is that too magic.

grakn-engine/src/main/java/ai/grakn/engine/controller/ConceptController.java
```diff
@@ -93,10 +96,26 @@ public ConceptController(EngineGraknTxFactory factory, Service spark,
...
+        spark.get(WebPath.CONCEPT_RELATIONSHIPS, this::getRelationships);
 
         spark.get(WebPath.TYPE_INSTANCES, this::getTypeInstances);
     }
 
+    private String getRelationships(Request request, Response response) throws JsonProcessingException {
+        //TODO: Figure out how to incorporate offset and limit
+        Function<ai.grakn.concept.Thing, Stream<Jacksonisable>> collector = thing -> {
+            Set<Role> roles = thing.plays().collect(Collectors.toSet());
+            return roles.stream().flatMap(role -> {
```
aelred: No reason to collect and then stream again.

grakn-engine/src/main/java/ai/grakn/engine/controller/ConceptController.java
```diff
@@ -120,15 +156,7 @@ private String getAttributes(Request request, Response response, boolean isKey)
...
-            //Get the attributes including keys if needed
-            Stream<Attribute<?>> attributes;
-            if(isKey){
-                attributes = concept.asThing().keys();
-            } else {
-                attributes = concept.asThing().attributes();
-            }
-
-            return objectMapper.writeValueAsString(attributes.map(EmbeddedAttribute::create).collect(Collectors.toSet()));
+            return objectMapper.writeValueAsString(collector.apply((X) concept).collect(Collectors.toList()));
```
aelred: If we do `<X extends Concept>` can we avoid this cast?

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
fppt: We decided against that for the sake of isolation. This implementation is imperfect and we will rework it for sure and when that happens having it like this makes it easier to change.

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
fppt: Too much magic I think. We mixing meanings up.

fppt:
@aelred Seperate PR for updating the API docs. I think this PR is big enough 😛

fppt:
@aelred re-review again please

PR #2423, opened 2017-11-24
Fix bug with `graql` command passing arguments

aelred:


aelred:
@fppt this should fix `master` so an early review would be helpful!

PR #2425, opened 2017-11-24
Add missing method to be able to create inferred attribute links on r…

fppt:
…elationships

Fun fact: This will mean that rules can cause reification. Not scary at all.


kasper-piskorski:
Fun fact II: They already can :)

fppt:
Oh yes you right. It just means that they will be marked as inferred now.

PR #2433, opened 2017-11-28
Add query validity tests

kasper-piskorski:


grakn-graql/src/test/java/ai/grakn/graql/internal/reasoner/QueryValidityTest.java
```diff
@@ -0,0 +1,161 @@
...
+        String queryString = "match $x has name $y; $x id 'V123'; get;";
+        String queryString2 = "match $x has name $y; $y id 'V123'; get;";
+        String queryString3 = "match $x has name $y via $r; $r id 'V123'; get;";
+        assertThat(qb.<GetQuery>parse(queryString).execute(), empty());
+        assertThat(qb.<GetQuery>parse(queryString2).execute(), empty());
+        assertThat(qb.<GetQuery>parse(queryString3).execute(), empty());
+    }
+
+    @Test
+    public void whenQueryingForInexistentEntityTypeLabel() throws GraqlQueryException{
```
aelred: Even though it says `throws` at the end, we need to put what should happen in the name of the test so we can see it when it fails on IntelliJ or Jenkins. I usually just put `whenBlah_Throw` because I think that's enough info.

PR #2434, opened 2017-11-28
Delay execution of post processing

fppt:


fppt:
@aelred it already passed. The latest change is just removing the todo.

fppt:
I set PP to be delayed by 30 mins. That should be moooorrrrreee than enough to allow indicies to settle

PR #2435, opened 2017-11-28
Improve support for ontological queries

kasper-piskorski:


grakn-graql/src/main/java/ai/grakn/graql/internal/query/QueryAnswer.java
```diff
@@ -156,11 +155,40 @@ public Answer merge(Answer a2, boolean mergeExplanation){
...
+                a2.entrySet()
+        )
+                .stream()
+                .filter(e -> !varIntersection.contains(e.getKey()))
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+        varIntersection
+                .forEach(var -> {
+                    Concept concept = this.get(var);
+                    Concept otherConcept = a2.get(var);
+                    if (concept.equals(otherConcept)) entryMap.put(var, concept);
```
fppt: Maybe we could put this in another method? The indentation and the embedding in the lambda makes it very hard to understand. I had to read this one a few times.

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/ResolutionPlan.java
```diff
@@ -167,22 +168,26 @@ public String toString(){
...
-                        .distinct()
-                        .iterator())
-                .build();
+
+        //TODO: need to double check correctness of this
+        ImmutableList.Builder<Atom> builder = ImmutableList.builder();
+        builder.addAll(query.getAtoms(OntologicalAtom.class).iterator());
+        builder.addAll(fragments.stream()
+                .map(Fragment::varProperty)
+                .filter(Objects::nonNull)
```
fppt: It terrifies me that your stream can contain nulls 😮

PR #2436, opened 2017-11-28
Cleanup unused task code in engine

fppt:


grakn-engine/src/main/java/ai/grakn/engine/tasks/manager/TaskState.java
```diff
@@ -150,57 +115,15 @@ private TaskState(TaskState taskState) {
...
-
-    public TaskState markScheduled(){
-        this.status = SCHEDULED;
-        this.statusChangeTime = now();
-
-        return this;
-    }
-
-    public TaskState markStopped(){
-        this.status = STOPPED;
```
aelred: Do we even use `status` now?

fppt: Probably not. TIME FOR MORE KILLING!

PR #2438, opened 2017-11-29
Copy Jenkinsfile from `master`

aelred:
This was because I had been accidentally/lazily applying changes to Jenkinsfile in `master` when they should have gone to `stable`.


burukuru:
yassssss

PR #2441, opened 2017-11-29
Merge release 0.18 into stable

fppt:


fppt:
Immediately merging because this is only being done to maintain the history and changelog

PR #2444, opened 2017-11-30
Grakn Nexus profile

burukuru:
* Add Grakn Nexus profile
* Skip tar.gz archive from Maven upload due to Central filesize limit
* Push snapshot to Grakn Maven when tests pass


Jenkinsfile
```diff
@@ -280,6 +280,16 @@ def runBuild() {
     graknNode { workspace ->
         checkout scm
 
+        // Push to Grakn Maven if tests pass
+        if (!isMainBranch()) {
```
aelred: This means we only deploy if it's a pull request. Is that right?

burukuru: copy/pasta fail

aelred:
!rtg - time to test SNB!

PR #2445, opened 2017-11-30
Replace all tabs at the start of lines with four spaces

aelred:
The inconsistency was bothering me, also I'm loading SNB


grakn-test/test-integration/src/test/java/ai/grakn/test/migration/owl/TestSubProperties.java
```diff
@@ -35,16 +35,16 @@
 import static ai.grakn.graql.Graql.var;
 
 public class TestSubProperties extends TestOwlGraknBase {
-	private final IRI baseIri = IRI.create("http://www.workingontologist.org/Examples/Chapter3/shakespeare.owl");
```
aelred: @fppt did we just unearth an ancient OWL migrator test?

fppt: Looks like it. Kill!

PR #2446, opened 2017-11-30
Ghosts now lead to explicit failure which require retrying

fppt:


grakn-kb/src/main/java/ai/grakn/kb/internal/structure/AbstractElement.java
```diff
@@ -147,7 +147,13 @@ public boolean equals(Object object) {
     public void propertyUnique(P key, String value){
         if(!tx().isBatchTx()) {
             GraphTraversal<Vertex, Vertex> traversal = tx().getTinkerTraversal().V().has(key.name(), value);
-            if(traversal.hasNext()) throw PropertyNotUniqueException.cannotChangeProperty(element(), traversal.next(), key, value);
+
+            if(traversal.hasNext()){
+                Vertex vertex = traversal.next();
+                if(traversal.hasNext() || !vertex.equals(element())){
```
aelred: We should flip this `||` for efficiency reasons (since the equals check is much faster)

grakn-kb/src/main/java/ai/grakn/kb/internal/concept/AttributeTypeImpl.java
```diff
@@ -115,28 +119,36 @@ private void checkInstancesMatchRegex(@Nullable String regex){
     @SuppressWarnings("unchecked")
     @Override
     public Attribute<D> putAttribute(D value) {
-        return putAttribute(value, false);
+        return putAttribute(value, false, 1);
     }
 
     public Attribute<D> putAttributeInferred(D value) {
-        return putAttribute(value, true);
+        return putAttribute(value, true, 1);
```
aelred: Starting at 1 and not 0? 😱

fppt: Did it for logging reasons :P

grakn-kb/src/main/java/ai/grakn/kb/internal/structure/AbstractElement.java
fppt: Good point

fppt:
@aelred re-review please

grakn-kb/src/main/java/ai/grakn/kb/internal/concept/ElementFactory.java
```diff
@@ -154,8 +155,7 @@ public RoleImpl buildRole(VertexElement vertex, Role type){
         try {
             type = getBaseType(vertexElement);
         } catch (IllegalStateException e){
-            LOG.warn("Invalid vertex [" + vertexElement + "] due to " + e.getMessage(), e);
-            return Optional.empty();
+            throw TemporaryWriteException.indexOverlap(vertexElement.element(), e);
```
aelred: This method now always returns something! It doesn't have to be optional if we want.

PR #2447, opened 2017-11-30
Add batch mode option to GraqlController

aelred:


grakn-core/src/main/java/ai/grakn/util/REST.java
```diff
@@ -119,6 +119,7 @@ public static String resolveTemplate(String pathTemplate, String... pathParams)
             public static final String QUERY = "query";
             public static final String INFER = "infer";
             public static final String MULTI = "multi";
+            public static final String TYPE = "type";
```
fppt: ooooo I really don't like type as the name of the parameter. Type is so overloaded in our world. Can we not use `tx-type` or `tx-mode` ?

fppt: `TxType` ?

aelred:
@fppt ready for re-review

PR #2448, opened 2017-11-30
Make behaviour of atomic states consistent wrt type inference

kasper-piskorski:


fppt:
@kasper-piskorski What's up with this PR? It's quite old and is tagged as WIP?

kasper-piskorski:
I grew more pessimistic about it over time ;p I'm no longer convinced that's the right approach, probably will scrap it

fppt:
Fair enough :)

kasper-piskorski:
ok, no longer WIP but mergeable

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
```diff
@@ -110,7 +109,7 @@
     private ImmutableMultimap<Role, Var> roleVarMap = null;
     private ImmutableMultimap<Role, Type> roleTypeMap = null;
     private ImmutableMultimap<Role, String> roleConceptIdMap = null;
-    private ImmutableList<RelationshipType> possibleRelations = null;
+    private ImmutableList<Type> possibleRelations = null;
```
aelred: Is it `Type` because it can include `thing`?

PR #2449, opened 2017-11-30
New path query that can find all shortest paths

duckofyork:


pouledodue:
👍

duckofyork:
!rtg

duckofyork:
!rtg

burukuru:
!rtg

duckofyork:
!rtg

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/PathQueryImpl.java
```diff
@@ -68,48 +69,119 @@
...
+        if (includeAttribute) { // this can be slow
+            return getExtendedPaths(allPaths);
+        }
+
+        LOGGER.info("Number of paths: " + allPaths.size());
+        LOGGER.info("ShortestPathVertexProgram is done in " + (System.currentTimeMillis() - startTime) + " ms");
+        return allPaths;
+    }
+
+    private List<List<Concept>> getExtendedPaths(List<List<Concept>> allPaths) {
```
aelred: I'd put a Javadoc explaining what this method is for. It's the one for handling attribute edges right?

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/PathQueryImpl.java
```diff
@@ -68,48 +69,119 @@
...
+        List<List<Concept>> extendedPaths = new ArrayList<>();
+        for (List<Concept> currentPath : allPaths) {
+            boolean hasAttribute = false;
```
aelred: STREAM MAGIC
```
boolean hasAttribute = currentPath.stream().anyMatch(Concept::isAttribute);
```

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/PathQueryImpl.java
```diff
@@ -68,48 +69,119 @@
...
-                .collect(Collectors.toList()));
-        path.add(destinationId);
-
-        List<ConceptId> fullPath = new ArrayList<>();
-        for (int index = 0; index < path.size() - 1; index++) {
-            fullPath.add(path.get(index));
-            if (includeAttribute) {
-                ConceptId resourceRelationId = getResourceEdgeId(tx.get(), path.get(index), path.get(index + 1));
+
+        int numExtensionAllowed = extendedPaths.isEmpty() ? Integer.MAX_VALUE : 0;
```
aelred: I'm not sure I get what this is for. Maybe add a comment?

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/ShortestPathVertexProgram.java
```diff
@@ -48,35 +52,36 @@
...
     private static final int MAX_ITERATION = 50;
 
-    public static final String FOUND_IN_ITERATION = "shortestPathVertexProgram.foundInIteration";
-    public static final String MIDDLE = "shortestPathVertexProgram.middle";
+    private static final int FROM_SOURCE = 1;
+    private static final int FROM_DESTINATION = -1;
+    private static final int FROM_MIDDLE = 0;
+
+    private static final int DIRECTION = 0;
+    private static final int ID = 1;
```
aelred: All these magic numbers and strings below could do with a description. Are they keys into something?

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/ShortestPathVertexProgram.java
```diff
@@ -102,140 +106,153 @@ public ShortestPathVertexProgram(ConceptId sourceId, ConceptId destinationId) {
...
+                vertex.property(VISITED_IN_ITERATION, 1);
+                sendMessage(messenger, Pair.with(FROM_SOURCE, id));
+            } else if (persistentProperties.get(DESTINATION).equals(id)) {
+                LOGGER.debug("Found destination vertex");
+                vertex.property(VISITED_IN_ITERATION, -1);
+                sendMessage(messenger, Pair.with(FROM_DESTINATION, id));
+            }
+        } else {
+            if (memory.<Boolean>get(FOUND_PATH)) {
+                if (messenger.receiveMessages().hasNext() && vertex.property(VISITED_IN_ITERATION).isPresent()) {
```
aelred: Maybe extract the body of this `if` into a method like `recordPaths`.

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/Utility.java
```diff
@@ -142,4 +142,11 @@ public static ConceptId getResourceEdgeId(GraknTx graph, ConceptId conceptId1, C
         }
         return null;
     }
+
+    /**
+     * Get the id of given vertex.
+     */
+    static String getVertexId(Vertex vertex) {
+        return vertex.value(Schema.VertexProperty.ID.name());
+    }
```
aelred: Can we wrap this in `ConceptId#of` immediately? If I remember right there were serialisation issues, but I still think it would be better to unwrap it before putting it in memory like `memory.put(KEY, cid.getValue());`

duckofyork: I could, but then we need to do this wrapping and unwrapping a couple of times for each of the vertex in each iteration. Or do you meaning doing this only when adding to memory?

fppt:
!rtg - Checking with CI Tests

fppt:
!ci

PR #2450, opened 2017-11-30
Changes for enterprise

micheleorsi:


lolski:
LGTM

grakn-dist/src/main/java/ai/grakn/dist/GraknProcess.java
```diff
@@ -155,11 +160,7 @@ private boolean graknCheckIfReady(String host, int port, String path) {
             connection.connect();
 
             int code = connection.getResponseCode();
-            if (code == 200) {
-                return true;
-            } else {
-                return false;
-            }
+            return code == 200;
```
aelred: haha, much better 👍

micheleorsi: (IntelliJ IDEA Analyze -> Inspect Code)

PR #2451, opened 2017-11-30
Fix bug with location of CSV data in SNB

aelred:
This is incorrect when using generated datasets


fppt:
I am guessing this only affected local runs? Otherwise how did anything ever work :P

aelred:
It only matters because this is where Hadoop puts the generated data, so not an issue on Jenkins 👍

PR #2452, opened 2017-11-30
Simplify behaviour of `SchemaConcept#sups`

aelred:


aelred:
Actually come to think of it, maybe the `sups` method should return `Stream<Type>`.

grakn-kb/src/main/java/ai/grakn/kb/internal/GraknTxAbstract.java
```diff
@@ -348,6 +348,18 @@ public ElementFactory factory() {
         }
     }
 
+    @Override
+    public final Stream<SchemaConcept> sups(SchemaConcept schemaConcept) {
+        Set<SchemaConcept> superSet= new HashSet<>();
+
+        while(schemaConcept != null) {
```
fppt: How barbaric!

PR #2453, opened 2017-12-01
Fix Jenkinsfile so it correctly tears down Grakn

aelred:


aelred:
!rtg - it failed but looks fine

aelred:
!rtg - I'm praying this was a ~one~ two-off

PR #2454, opened 2017-12-01
Get rid of Optionals when building concept from elements. This will lead to more explicit failures

fppt:


fppt:
Because this is quite high impact I am engaging 🐘 mode

fppt:
!rtg

fppt:
!rtg - weirdness fixed

fppt:
I am shocked that this passed

grakn-kb/src/main/java/ai/grakn/kb/internal/concept/ElementFactory.java
```diff
@@ -204,20 +201,12 @@ public RoleImpl buildRole(VertexElement vertex, Role type){
...
-
-        try {
-            label = Schema.EdgeLabel.valueOf(edgeElement.label().toUpperCase(Locale.getDefault()));
-        } catch (IllegalStateException e){
-            LOG.warn("Invalid edge [" + edgeElement + "] due to " + e.getMessage(), e);
-            return Optional.empty();
-        }
-
+    public <X extends Concept> X buildConcept(EdgeElement edgeElement){
+        Schema.EdgeLabel label = Schema.EdgeLabel.valueOf(edgeElement.label().toUpperCase(Locale.getDefault()));
```
aelred: Is removing the `try...catch` deliberate?

fppt: Yep now we just crash like badasses. BOOM!!!

fppt: On a serious note it's just so we can fail when encountering weird edges rather than hide them.

PR #2465, opened 2017-12-04
[DO NOT MERGE] Check if reverting change to loading fixes SNB

aelred:


aelred:
This one seems to have fixed it! @fppt

PR #2466, opened 2017-12-04
Add concept-pattern conversion required for building explanations

kasper-piskorski:


grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/utils/conversion/ConceptConverter.java
```diff
@@ -0,0 +1,49 @@
...
+ * @param <T> {@link Concept} type to convert from
+ *
+ * @author Kasper Piskorski
+ */
+public interface ConceptConverter<T extends Concept> {
+
+    Pattern pattern(T concept);
+
+    static Pattern toPattern(Concept concept) {
+        if (!concept.isThing()) return null;
```
aelred: Instead of returning `null`, I'd recommend `throw`ing here.

But if there's a good reason to return null, make sure to mark the method as `@Nullable`.

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/utils/conversion/EntityConverter.java
```diff
@@ -0,0 +1,41 @@
...
+ * @author Kasper Piskorski
+ */
+class EntityConverter implements ConceptConverter<Entity>{
+
+    @Override
+    public Pattern pattern(Entity concept) {
+            Var entityVar = Graql.var();
+            return entityVar
+                    .isa(concept.type().getLabel().getValue())
+                    .and(entityVar.asUserDefined().id(concept.getId()));
```
aelred: You can skip the `and` here and `getValue`:
```
Var entityVar = Graql.var().asUserDefined();
return entityVar.isa(Graql.label(concept.type().getLabel())).id(concept.getId());
```

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/utils/conversion/AttributeConverter.java
```diff
@@ -0,0 +1,43 @@
...
+ * @author Kasper Piskorski
+ */
+class AttributeConverter implements ConceptConverter<Attribute> {
+
+    @Override
+    public Pattern pattern(Attribute concept) {
+        Var owner = Graql.var().asUserDefined();
+        VarPattern resourceVar = Graql.var().asUserDefined().val(concept.getValue());
+        return owner
+                .has(concept.type().getLabel().getValue(),resourceVar)
```
aelred: You can skip the call to `getValue()` here, so we avoid unwrapping the `Label`.

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/utils/conversion/RelationshipConverter.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    public Pattern pattern(Relationship concept) {
+        VarPattern relationPattern = Graql.var();
+        Set<Pattern> idPatterns = new HashSet<>();
+
+        for (Map.Entry<Role, Set<Thing>> entry : concept.allRolePlayers().entrySet()) {
+            for (Thing var : entry.getValue()) {
+                Var rolePlayer = Graql.var();
+                relationPattern = relationPattern.rel(entry.getKey().getLabel().getValue(), rolePlayer);
```
aelred: `relationPattern.rel(Graql.label(entry.getKey().getLabel()), rolePlayer);`

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/utils/conversion/RelationshipConverter.java
```diff
@@ -0,0 +1,60 @@
...
+                idPatterns.add(rolePlayer.asUserDefined().id(var.getId()));
+            }
+        }
+        relationPattern = relationPattern.isa(concept.type().getLabel().getValue());
```
aelred: `relationPattern.isa(Graql.label(concept.type().getLabel()));`

PR #2467, opened 2017-12-04
Jacksonise and Cleanup commit logs + Some Task API simplification

fppt:


grakn-engine/src/main/java/ai/grakn/engine/tasks/mock/MockBackgroundTask.java
```diff
@@ -88,12 +88,17 @@ public static void clearTasks() {
...
+            throw new RuntimeException("Invalid task config");
+        }
+
         Json id = json.at("id");
         if (id == null) {
             LOG.error("Missing id in {}: {}", this.getClass().getName(), json);
         }
-        this.id = TaskId.of(id.asString());
+        this.id = TaskId.generate();
+        //this.id = TaskId.of(id.asString());
```
aelred: 🐐

grakn-engine/src/test/java/ai/grakn/engine/controller/GraqlControllerInsertTest.java
```diff
@@ -224,7 +234,7 @@ public void POSTGraqlInsertAndNoLogs_PPTaskIsNotSubmitted(){
 
         sendRequest(query);
 
-        verify(taskManager, times(1)).addTask(any(), any());
+        verify(taskManager, times(0)).addTask(any(), any());
```
aelred: Whoops, were these tests wrong all along? 😱

grakn-kb/src/main/java/ai/grakn/factory/GraknSessionImpl.java
```diff
@@ -93,10 +95,7 @@
             ThreadFactory namedThreadFactory = new ThreadFactoryBuilder()
                     .setNameFormat("commit-log-subbmit-%d").build();
             commitLogSubmitter = Executors.newSingleThreadScheduledExecutor(namedThreadFactory);
-            commitLogSubmitter.scheduleAtFixedRate(() -> {
-                submitLogs(tx);
-                submitLogs(txBatch);
-            }, 0, LOG_SUBMISSION_PERIOD, TimeUnit.SECONDS);
+            commitLogSubmitter.scheduleAtFixedRate(this::submitLogs, 0, LOG_SUBMISSION_PERIOD, TimeUnit.SECONDS);
```
aelred: When/how do we submit the commit logs for batch transactions?

fppt: Both the normal logs and batch logs are put in the same `CommitLogHandler` so we only need one submission

grakn-engine/src/test/java/ai/grakn/engine/controller/GraqlControllerInsertTest.java
fppt: Yep. . . They were wrong.

fppt:
@aelred re-review please

grakn-engine/src/main/java/ai/grakn/engine/tasks/mock/MockBackgroundTask.java
```diff
@@ -88,12 +88,16 @@ public static void clearTasks() {
...
+        if(json.equals(Json.object())){
+            throw new RuntimeException("Invalid task config");
+        }
+
         Json id = json.at("id");
         if (id == null) {
             LOG.error("Missing id in {}: {}", this.getClass().getName(), json);
         }
-        this.id = TaskId.of(id.asString());
+        this.id = TaskId.generate();
```
aelred: It looks like we ignore the ID from the JSON configuration here. Is that deliberate? I can't remember how any of this task stuff works anymore.

fppt: The Task ID is kind of in limbo right now. It's not actually not used anywhere anymore properly. The old property tests use it in weird ways, hence the weird mock above.

I could get rid of the old property tests but I am not 100% sure yet if I should.

These odd mocks should die in another PR though. They probably made sense at some point but not anymore

PR #2468, opened 2017-12-05
Docs redesign

excerebrose:
* Redesign of documentation pages
* Self deployment server for heroku


excerebrose:
@fppt @marco-scoppetta

docs/deploy-server.js
```diff
@@ -0,0 +1,18 @@
...
+
+app.use('/images', express.static(path.join(__dirname, 'images')));
+app.use('/', express.static(dist));
+
+
+app.listen(port, (error) => {
+  if (error) {
+    console.log(error); // eslint-disable-line no-console
+  }
+  console.info('Express is listening on port %s.', port); // eslint-disable-line no-console
```
marco-scoppetta: This should probably be an `if-else` block.
Also just FYI you can add `/* eslint no-console: 0 */` at the beginning of the file so that you don't have to comment multiple lines.
Of course these are not important changes, it was just for fun

docs/deploy.sh
```diff
@@ -0,0 +1,15 @@
...
+echo "Checking out temporary branch"
+git checkout -b docs-deploy-temp
+echo "Building _site"
+rake build
+echo "Staging new files"
+git add -A 
+git commit -m "Site update"
+echo "Pushing to Heroku"
+cd ../
+git push https://git.heroku.com/docs-grakn.git `git subtree split --prefix docs docs-deploy-temp`:master --force
```
marco-scoppetta: Nice!

fppt:
@excerebrose Could you also update the docs `README.md` with the new release and testing process?

docs/deploy.sh
```diff
@@ -0,0 +1,15 @@
...
+echo "Deleting temporary branch"
+git checkout docs-redesign
```
burukuru: You might want to checkout `stable` instead.

docs/deploy.sh
burukuru: can we push to git@heroku.com:docs-grakn.git instead so we can use SSH keys for instead?

docs/deploy.sh
excerebrose: shit, yeah - it was for testing

excerebrose:
@fppt @burukuru Your requested changes are addressed in the new commits

fppt:
@excerebrose some merge conflicts have developed on this PR. Could you take a look at them please? 

We recently merged master into stable so I imagine that caused the conflicts.

fppt:
Merging this because:

1. It did indeed pass on CI despite the weird notification issue
2. It is only doc changes
3. We are likely to do more doc changes soon so this should go in to avoid diverting

PR #2469, opened 2017-12-05
Revert changes to `Migrator` to fix SNB issue

aelred:


aelred:
@fppt this looks like it fixed. I'll let it pass, remove the elephant then we can merge.

PR #2471, opened 2017-12-05
Run SNB etc. with all PRs after Junit tests

aelred:


aelred:
@fppt this failed SNB as expected. Do we want to merge it or wait for `master` to pass?

fppt:
In my opinion @aelred we should wait for master to pass. If we merge this then we going to just have a bunch of PRs running CI and failing all the time. Just burning resources for no reason.

fppt:
Merging this in so I can start retesting all the other PRs which need to go in.

PR #2472, opened 2017-12-05
Upgrade redisq version

fppt:


fppt:
@aelred this is safe to merge despite the failing status of CI. All it does is reduce the redisq logs based on this PR: https://github.com/graknlabs/redisq/pull/3

aelred:
@fppt should we merge it before tests pass?

fppt:
Let's merge if it builds. Just want to make sure there are no dependency issues

PR #2475, opened 2017-12-05
Spinner

marco-scoppetta:
- removed Pace progress bar that was shown on every AJAX request
- we now have a new spinner instead


grakn-dashboard/package.json
```diff
@@ -13,9 +13,7 @@
   },
   "homepage": "https://github.com/graknlabs/grakn",
   "dependencies": {
-    "codemirror": "^5.19.0",
-    "jquery": "^3.1.0",
```
aelred: No more JQuery? :o

marco-scoppetta:
@fppt can we merge this? at least one joy for today

aelred:
@marco-scoppetta here is your daily ration of joy. don't spend it all at once!

fppt:
Why am I never allowed to hand out joy?

PR #2478, opened 2017-12-05
Fix the intermittent Jenkins failure properly

aelred:


aelred:
@burukuru you said that last time and things didn't go so great!

PR #2479, opened 2017-12-06
Re-apply memory fix to `BatchExecutorClient`

aelred:
Note: I renamed `BatchExecutorClientTest` to `BatchExecutorClientIT` then made a _new_ `BatchExecutorClientTest`. That's why the diff is dumb.


grakn-migration/base/src/main/java/ai/grakn/migration/base/Migrator.java
```diff
@@ -134,21 +134,21 @@ public void load(String template, Stream<Map<String, Object>> data) {
...
-                        subscribeToReportOutcome(failFast, addObservable);
+                        Observable<QueryResponse> observable = loader.add(q, keyspace, failFast);
+                        subscribeToReportOutcome(failFast, observable, queriesExecuted);
                     });
-            int completed = allObservableWithTimeout(allObservables, OBSERVABLE_TIMEOUT_MINUTES,
-                    TimeUnit.MINUTES).toBlocking().first().size();
-            LOG.info("Loaded {} statements", completed);
         }
+
+        LOG.info("Loaded {} statements", queriesExecuted);
```
fppt: Migrator is client side. This should probably go to `sout`

PR #2490, opened 2017-12-08
Add ConnectedComponentVertexProgram starting from single source

duckofyork:


grakn-graql/src/main/antlr4/ai/grakn/graql/internal/antlr/Graql.g4
```diff
@@ -30,7 +30,7 @@ mean           : MEAN     'of' ofList      ('in' inList)? ';' ;
 std            : STD      'of' ofList      ('in' inList)? ';' ;
 sum            : SUM      'of' ofList      ('in' inList)? ';' ;
 degrees        : DEGREES ('of' ofList)?    ('in' inList)? ';' ;
-cluster        : CLUSTER                   ('in' inList)? ';' clusterParam* ;
+cluster        : CLUSTER ('of' id    )?    ('in' inList)? ';' clusterParam* ;
```
duckofyork: would this work: compute cluster in xxx of v123;

aelred: No, has to be the other way around. But this is true for all the methods so far. We can change it later.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/ClusterQueryImpl.java
```diff
@@ -55,41 +61,38 @@ public T execute() {
             return (T) Collections.emptyMap();
         }
 
-        ComputerResult result;
-
         Set<LabelId> subLabelIds = convertLabelsToIds(subLabels);
 
+        GraknVertexProgram vertexProgram;
```
aelred: Make this `GraknVertexProgram<?> vertexProgram;` otherwise Java moves into ["raw" no generics mode](https://docs.oracle.com/javase/tutorial/java/generics/rawTypes.html) and things get weird :P

duckofyork: good point

PR #2492, opened 2017-12-08
Fix issue with joining explanations

kasper-piskorski:


fppt:
@kasper-piskorski just confirming. Should this still go into master or be transferred to stable?

PR #2494, opened 2017-12-11
Fix session leak

fppt:


fppt:
!rtg - CI is working again

PR #2495, opened 2017-12-11
Fix issue with expanding `grakn-dist*.tar.gz`

aelred:


aelred:
@fppt this is the PR that should fix `stable`, so it should be merged ASAP-ish.

fppt:
Merging because CI on stable it broken anyway so we might as well experiment with this PR

PR #2497, opened 2017-12-11
[Stable] Refine resolution plan based on rule resolvability

kasper-piskorski:


fppt:
!rtg - CI is working again

fppt:
@kasper-piskorski this PR appears to actually be failing on CI. Could you merge with central/stable and push back up? 

I would like to check if that will help.

fppt:
@kasper-piskorski I think this PR has broken CI. The validation queries are taking forever to run. They usually take 20 mins but now it's been over 1 hour.

kasper-piskorski:
ok, this is not safe, saving for later

PR #2499, opened 2017-12-11
Get rid of swagger dependency

fppt:


aelred:
@fppt you're going to get awkward PMD violations with this one. There's a special rule in `pmd-rules.xml` to exclude these methods which we should either change or remove.

If we remove the rule, we'll have to change the methods in question to only accept the `request`/`response` that they use.

PR #2500, opened 2017-12-11
Fix start-long-running-instance script

aelred:


aelred:
@fppt this only impacts the final deployment stage so it can be yolo'd

aelred:
@fppt I'm going to merge this so we can get `stable` passing

PR #2501, opened 2017-12-11
[DO NOT MERGE] Disallow multiple concurrent initialization stable

lolski:
# What is in the PR
## 1. Disallow concurrent bootup invocations
*Problem*:
The accidental case of performing multiple concurrent invocations of `grakn server start` should be prevented because they can cause instability which is hard to debug from the users' perspective.

*Solution*:
Prevent such things from happening. If there is an instance of `grakn server start` running, the subsequent calls to `grakn server start` will terminate with an error. The same rule applies to executing `grakn server start` concurrently with `grakn server stop`.

*Method*:
This is achieved by [a system-wide lock by using `mkdir`](https://unix.stackexchange.com/questions/70/what-unix-commands-can-be-used-as-a-semaphore-lock). It is a well-known approach for solving this exact problem and relies on the fact that `mkdir` is an atomic operation.

Upon the invocation of `grakn server start`, it will do a `mkdir /tmp/.grakn.lock` and if there is a second `grakn server start` execution running concurrently, that one will fail because the attempt of `mkdir /tmp/.grakn.lock` it does will fail.

*Failure Handling*:
1. User aborting the bootup (CTRL+C): TODO

## 2. Bootup script tests
Adding tests for the following scenario, under `ai.grakn.dist.DistGraknTest`
1. Happy path invocation of `grakn server start` and `grakn server start storage|queue|grakn`
2. Happy path invocation of `grakn server stop`
3. Guard against concurrent execution of `grakn server start` and `grakn server stop`

### DIFF
--- a/grakn-dist/src/main/java/ai/grakn/dist/DistGrakn.java
+++ b/grakn-dist/src/main/java/ai/grakn/dist/DistGrakn.java
@@ -36,7 +36,7 @@ import java.util.Scanner;
  * @author Michele Orsi
  */
 public class DistGrakn {
-    private static final String SYNCHRONIZED_BOOTUP_LOCK_PATH = "/tmp/.grakn.synchronizedBootupLock";
+    private static final String SYNCHRONIZED_BOOTUP_LOCK_PATH = "/tmp/grakn-synchronized-bootup.lock";
 
     private static final String GRAKN = "grakn";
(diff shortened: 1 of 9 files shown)


lolski:
Closed - the same goals are achieved with a better approach in https://github.com/graknlabs/grakn/pull/2508

PR #2502, opened 2017-12-11
[Stable] Benchmark integration tests

kasper-piskorski:
This needs the fix to BatchExecutorClient: https://github.com/graknlabs/grakn/pull/2479


kasper-piskorski:
one pass

PR #2508, opened 2017-12-12
Fixing dangling /tmp/grakn.pid in case Grakn could not start

lolski:
When Grakn could not start, the file `/tmp/grakn.pid` is often left dangling. A dangling PID file is annoying because upon subsequent bootup, the script would think that Grakn is still running.

This PR fixes that problem by always deleting the `/tmp/grakn.pid` if Grakn exits. Grakn would correctly delete the file if it exits and should not cause any dangling `/tmp/grakn.pid` which users always complains about.

Other small details for the keen reviewers:
1. Made the `OutputCommand` and `AbstractProcessHandler` class public since we want to reuse them. Additionally, they are moved from `grakn-dist` into a new subproject `grakn-bootup`, as otherwise we would introduce a circular dependency between `grakn-engine` and `grakn-dist`.
2. Retrieve PID via a combination of  `ps -ef`, `grep`, and `awk`. But you said, why couldn't we do something like `java.lang.Runtime.getPid()`? I know. Sadly there's not an easy and portable way of getting a process ID, so we're doing the same thing we've always done in the bash script for now

A follow-up PR (targeted to `master`) will improve the following:
1. Move over bootup script to the new `grakn-bootup` subproject. Right now, they are located at `grakn-dist` which is not ideal
2. Extracting various side-effecting operations in the bootup script (there are a lot of them) so components be easily tested

### DIFF
--- a/conf/main/grakn.properties
+++ b/conf/main/grakn.properties
@@ -120,3 +120,4 @@ cache.db-cache-size=0.25
 queue.host=localhost:6379
 queue.consumers=32
 redis.pool-size=32
+
--- a/conf/main/grakn.properties
+++ b/conf/main/grakn.properties
@@ -120,4 +120,3 @@ cache.db-cache-size=0.25
 queue.host=localhost:6379
 queue.consumers=32
 redis.pool-size=32
-
--- a/grakn-bootup/src/main/java/ai/grakn/bootup/PidRetriever.java
+++ b/grakn-bootup/src/main/java/ai/grakn/bootup/PidRetriever.java
@@ -1,0 +1,45 @@
+/*
+ * Grakn - A Distributed Semantic Database
+ * Copyright (C) 2016  Grakn Labs Limited
+ *
+ * Grakn is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Grakn is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Grakn. If not, see <http://www.gnu.org/licenses/gpl.txt>.
+ */
+
+package ai.grakn.bootup;
+
+/**
+ *
+ * Main class invoked by bash scripting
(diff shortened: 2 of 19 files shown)


pom.xml
```diff
@@ -114,6 +115,7 @@
         <failsafe.version>1.0.4</failsafe.version>
         <redisq.version>0.0.7-SNAPSHOT</redisq.version>
         <reactivex.version>1.3.2</reactivex.version>
+        <sigar.version>1.6.4</sigar.version>
```
lolski: delete this

conf/main/grakn.properties
```diff
@@ -119,5 +119,4 @@ cache.db-cache-size=0.25
 # IP or hostname on which Redis is listening for connections.
 queue.host=localhost:6379
 queue.consumers=32
-redis.pool-size=32
-
+redis.pool-size=32
```
lolski: restore this

grakn-bootup/src/main/java/ai/grakn/bootup/PidRetriever.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ *
+ * Main class invoked by bash scripting
+ *
+ * @author Michele Orsi
+ *
+ */
+public class PidRetriever extends AbstractProcessHandler {
+    static final String getPidCommand = "ps -ef | ps -ef | grep \"ai.grakn.engine.Grakn\" | grep -v grep | awk '{print $2}'";
```
lolski: fix `ps -ef | ps -ef`

grakn-engine/src/main/java/ai/grakn/engine/Grakn.java
```diff
@@ -40,6 +42,9 @@
      */
     public static void main(String[] args) {
         try {
+            GraknPid pidFile = newPidFile_deleteOnExit();
+            pidFile.createPidFile_deleteOnExit();
```
aelred: If we move these lines into `grakn-dist`, can't we remove `grakn-bootup`? All the behaviour can instead go in `grakn-dist`. It makes sense to me that `grakn-engine` isn't responsible for the PID files.

lolski: w.r.t PID file

That was what we used to do, however, PID file has to be managed by the process who owns it. Imagine a scenario where Grakn died due to out of memory long after the bootup script has completed.

aelred: Got it! I can see why it has to be in `grakn-engine` now.

PR #2512, opened 2017-12-13
Update type representations and new REST endpoints

fppt:


grakn-engine/src/main/java/ai/grakn/engine/controller/response/ConceptBuilder.java
```diff
@@ -80,10 +80,10 @@ public static Things buildThings(ai.grakn.concept.Type type, int offset, int lim
     //TODO: This will scale poorly with super nodes. Need to introduce some sort of paging maybe?
```
aelred: I think we just did the TODO!

grakn-engine/src/main/java/ai/grakn/engine/controller/response/MetaConcept.java
```diff
@@ -46,9 +44,9 @@ public static MetaConcept create(
             @JsonProperty("id") ConceptId id,
             @JsonProperty("@id") Link selfLink,
             @JsonProperty("label") Label label,
-            @JsonProperty("super") @Nullable Link sup,
-            @JsonProperty("subs") Set<Link> subs
+            @JsonProperty("super") @Nullable EmbeddedSchemaConcept sup,
+            @JsonProperty("subs") Link subs
     ){
-        return new AutoValue_MetaConcept(Schema.BaseType.TYPE.name(), id, selfLink, label, false, sup, subs, true, Collections.emptySet(), Collections.emptySet(), Collections.emptySet());
+        return new AutoValue_MetaConcept(Schema.BaseType.TYPE.name(), id, selfLink, label, false, sup, subs, true, null, null, null, null);
```
aelred: These could be regular links to empty collections if we want. That feels like a bit less of a special-case to me.

grakn-engine/src/main/java/ai/grakn/engine/controller/response/ConceptBuilder.java
fppt: Our paging is artificial I think, so i will keep the todo for the moment.

aelred: oh yeah! I forgot about that.

grakn-engine/src/main/java/ai/grakn/engine/controller/response/MetaConcept.java
fppt: Ya fair enough. Less nulls running around.

aelred:
@fppt you need to merge with `stable`

fppt:
@aelred thanks for the heads up!

PR #2516, opened 2017-12-13
Get rid of unique flag in docs

fppt:


fppt:
Merging again just because it's a doc change.

PR #2518, opened 2017-12-13
Addition to "Fixing dangling /tmp/grakn.pid" PR

lolski:
Addition to https://github.com/graknlabs/grakn/pull/2508

It breaks the tests because we're killing the engine with `destroyForcibly()` which is the equivalent of a `kill -9`. With a `kill -9`, the PID file won't be deleted and the next time engine starts it will complain that there's a PID file already.

This PR fixes `DistributionContext` used by the `GraqlShellIT`. However, it seems like this particular test isn’t ran during PR, which is why it escaped our attention. We need a way to test this particular case during PR as well.


aelred:
@lolski we also `kill -9` in `distribution-test.sh`, which we saw fail after this PR last got merged. Are we sure that won't happen again? 👍

fppt:
Just realised this is against master. This should be against stable @lolski

lolski:
@aelred @fppt Closing this PR which is incorrectly pointed to `master`. See https://github.com/graknlabs/grakn/pull/2525

PR #2522, opened 2017-12-14
Remove unused dependencies from `grakn-graql`

aelred:


fppt:
@aelred is this essential for 1.0? If not I would ask moving it to master.

fppt:
@aelred lets move them then please. I know they are very unobtrusive but my paranoia has kicked in considering how close we are :)

PR #2523, opened 2017-12-14
Remove dependency on javatuples

aelred:


fppt:
@aelred is this essential for 1.0? If not I would ask moving it to master.

PR #2525, opened 2017-12-14
Addition to "Fixing dangling /tmp/grakn.pid" PR (stable)

lolski:
Addition to #2508

It breaks the tests because we're killing the engine with destroyForcibly() which is the equivalent of a kill -9. With a kill -9, the PID file won't be deleted and the next time engine starts it will complain that there's a PID file already.

This PR fixes DistributionContext used by the GraqlShellIT. However, it seems like this particular test isn’t ran during PR, which is why it escaped our attention. We need a way to test this particular case during PR as well.

### DIFF
--- a/conf/main/grakn.properties
+++ b/conf/main/grakn.properties
@@ -120,3 +120,4 @@ cache.db-cache-size=0.25
 queue.host=localhost:6379
 queue.consumers=32
 redis.pool-size=32
+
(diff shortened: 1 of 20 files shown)


lolski:
@aelred most of the `kill -9` in `distribution-test.sh` only happens during the so-called forceful teardown step done only if the test has failed anyway.

however there is one place where we forgot to clean up redis PID which is fixed here: https://github.com/graknlabs/grakn/pull/2525/commits/01ad80777ab65a932537a817ce166c482b195d49

PR #2528, opened 2017-12-14
Update REST API spec to match new changes

aelred:


fppt:
Only documentation changes so merging early

PR #2531, opened 2017-12-14
Use better error handling when possible in Graql

aelred:


aelred:
I can confirm `grakn-graql` module passes locally if you want to YOLO

fppt:
What could go wrong.

PR #2537, opened 2017-12-15
Add randomised integration tests

kasper-piskorski:


grakn-test/test-integration/src/test/java/ai/grakn/test/graql/reasoner/BenchmarkTests.java
```diff
@@ -0,0 +1,226 @@
...
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static ai.grakn.graql.Graql.var;
+import static ai.grakn.util.SampleKBLoader.randomKeyspace;
+import static org.junit.Assert.assertEquals;
+
+public class BenchmarkTests {
```
aelred: Since this is using a real engine, we should call it `BenchmarkIT`. Unfortunately, putting it in the `test-integration` module isn't enough to get it labelled an integration test :(

PR #2538, opened 2017-12-15
New analytics method: k-core

duckofyork:


grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/KCoreVertexProgram.java
```diff
@@ -0,0 +1,260 @@
...
+    @Override
+    public Set<MemoryComputeKey> getMemoryComputeKeys() {
+        return MEMORY_COMPUTE_KEYS;
+    }
+
+    @Override
+    public void setup(final Memory memory) {
+        LOGGER.debug("KCoreVertexProgram Started !!!!!!!!");
+
+        // K_CORE_STABLE is true by fault, and we reset it after each odd iteration.
```
aelred: "default" right? 👍

grakn-graql/src/main/java/ai/grakn/graql/internal/analytics/KCoreVertexProgram.java
```diff
@@ -0,0 +1,260 @@
...
+                        } else if (vertex.label().equals(Schema.BaseType.RELATIONSHIP.name())) {
+                            relayClusterLabel(messenger, memory);
+                        }
+                    }
+                } else {
+                    if (memory.getIteration() % 2 == 0) {
+                        relayOrSaveMessages(vertex, messenger);
+                    } else {
+                        updateEntityAndAttribute(vertex, messenger, memory);
+                    }
```
aelred: I think I get this: we do this odd-even thing so we relay messages "through" relationships? I'd add a comment explaining it.

aelred:
Looks good! Just some minor things.

PR #2539, opened 2017-12-15
Merge master into stable

fppt:


fppt:
This actually did pass so I am merging

PR #2542, opened 2017-12-19
Extracting most bootup scripts to a new submodule grakn-bootup

lolski:
# Why is this PR needed?
This is a follow up PR to #2525 for cleanup purpose. I mean, who doesn't hate an untidy codebase!

# What does the PR do?

1. Moving most of the grakn bootup script components to the `grakn-bootup` submodule.
2. Updating graql console copyright year to 2018
3. PID file (`/tmp/grakn.pid`) management so it looks cleaner and more readable:
a. `GraknPidRetriever` 
b. `UnixGraknPidRetriever` (uses `ps` to retrieve PID, which is a tool only available in Unix)
c. `GraknPid`
d. `GraknPidManagerFactory`
e. `GraknPidException` (thrown for any PID related failure e.g. unable to retrieve PID or create PID file)
4. Extra bonus: `DistributionContext` now prints `logs/grakn.log` if it fails to start for some reason! This should help us alot in debugging tests.

No, but there are changes which is worth paying attention to:
1. Copyright year visible to the user is now changed to 2018
2. Starting `GraknEngineServer` requires you to pass an additional parameter in java: `java -Dgrakn.pidfile=/tmp/grakn.pid`. This change doesn't affect the user since the option is not visible to the user. This option is used here, if you're curious: https://github.com/graknlabs/grakn/pull/2542/files#diff-63c9b960134b2030ab454afaa391f2dbR54 

None


pom.xml
```diff
@@ -112,6 +112,7 @@
         <failsafe.version>1.0.4</failsafe.version>
         <redisq.version>0.0.7-SNAPSHOT</redisq.version>
         <reactivex.version>1.3.2</reactivex.version>
+        <sigar.version>1.6.4</sigar.version>
```
marco-scoppetta: Is this dependency still used somewhere? Just asking because I see this previous [commit](https://github.com/graknlabs/grakn/pull/2542/commits/7e6a7e2e4bf785082f2fa3b2e589c5be7f373165)

grakn-bootup/src/main/java/ai/grakn/bootup/graknengine/grakn_pid/GraknPidFileStore.java
```diff
@@ -0,0 +1,63 @@
...
+    private void attemptToWritePidFile(long pid, Path pidFilePath) {
+        if (!pidFilePath.toFile().exists()) {
+            String pidString = Long.toString(pid);
+            try {
+                Files.write(pidFilePath, pidString.getBytes(StandardCharsets.UTF_8));
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        } else {
+            String message = "pid file already exists: '" + pidFilePath.toString();
```
marco-scoppetta: For future reference: We might want to add this error message to `ErrorMessage` class situated in `grakn-core/util` so that you it usable with something like `ErrorMessage.PID_ALREADY_EXISTS.getMessage()`

pom.xml
lolski: @marco-scoppetta good catch!

lolski: Fixed in https://github.com/graknlabs/grakn/pull/2542/commits/f0fbe83e06e503067bd2fdb5faec57cecf2e03b1

PR #2545, opened 2017-12-20
FIx several REST API bugs

aelred:
btw I put this against `master` because I thought the changes were kind of serious, even though they are probably technically bug fixes. Let me know if you think it should be `stable`!


aelred:
This PR has probably broke dashboard, so best not merge till we check

marco-scoppetta:
@aelred why do you do this to me Felix?

aelred:
@marco-scoppetta merry christmas!

aelred:
@marco-scoppetta this is ready to review. We can work out if there's another way to do the #page.html redirect thing!

fppt:
I would say move this to be against stable @aelred

PR #2549, opened 2017-12-22
Small changes in order to run kbms version

micheleorsi:
I have just changed the visibility of some methods in order override in the sub-class of the KBMS


micheleorsi:
Unfortunately I cannot add reviewers to this PR, so I just mention you guys that are (hopefully) not on holidays: @lolski @kasper-piskorski 

Thanks

PR #2557, opened 2018-01-04
Add docs for Python and Haskell clients

aelred:


aelred:
@mikonapoli added an example of the `infer` and `multi` params!

PR #2558, opened 2018-01-04
Refine resolution plan to ensure more optimal substitution use

kasper-piskorski:


grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/Atom.java
```diff
@@ -100,6 +100,18 @@ public boolean isRecursive(){
                 .findFirst().isPresent();
     }
 
+    /**
+     * @return true if the atom is ground
```
aelred: If it's not super-complicated, maybe a quick definition of "ground", like "meaning, there are no free variables" (if that is the definition :P )

PR #2560, opened 2018-01-05
Add a test to catch a possible regression

aelred:
Linked to [this TP task](https://work.grakn.ai/entity/13719-non-unique-keyspace-graknsystem).


grakn-factory/janus-factory/src/test/java/ai/grakn/factory/TxFactoryJanusTest.java
```diff
@@ -191,6 +193,29 @@ public void testGraphNotClosed() throws InvalidKBException {
         graph.putEntityType("A Thing");
     }
 
+    @Test
+    public void whenSeveralFactoriesCreateTheSameKeyspace_NoErrorOccurs() throws Exception {
```
fppt: I suspect this case is already covered by `FactoryBuilderTest#whenBuildingFactoriesPointingToTheSameKeyspace_EnsureSingleFactoryIsReturned` and `GraknTest#whenGettingSessionForSameKeyspaceFromMultipleThreads_EnsureSingleSessionIsReturned`

What prompted the need to create this test?

aelred:
@fppt Oh, I should have linked the [TP task](https://work.grakn.ai/entity/13719-non-unique-keyspace-graknsystem).

It's actually a very old branch I had with the regression test already-written. It seems to have fixed itself because now it passes 👍

fppt:
It it likely a redundant test but its minor enough and tests at a different level of abstraction that I am okay with merging it.

PR #2563, opened 2018-01-08
Fix error handling when trying to connect to server

aelred:


aelred:
@fppt I am now semi-confident this will fix `stable` 👍

grakn-client/src/main/java/ai/grakn/client/Client.java
```diff
@@ -121,38 +122,58 @@ private static EngineStatus checkServerRunning() throws IOException {
      * @return true if Grakn Engine running, false otherwise
      */
     public static boolean serverIsRunning(SimpleURI uri) {
+        URL url;
         try {
-            URL url = UriBuilder.fromUri(uri.toURI()).path(REST.WebPath.KB).build().toURL();
+            url = UriBuilder.fromUri(uri.toURI()).path(REST.WebPath.KB).build().toURL();
+        } catch (MalformedURLException e) {
+            assert false : "This will never throw because we're appending a known path to a valid URI";
```
fppt: What magic is this @aelred ?

grakn-test/test-integration/src/test/java/ai/grakn/test/rule/DistributionContext.java
```diff
@@ -59,15 +61,17 @@
     public static final Logger LOG = LoggerFactory.getLogger(DistributionContext.class);
 
     private static final FilenameFilter jarFiles = (dir, name) -> name.toLowerCase().endsWith(".jar");
-    private static final String ZIP = "grakn-dist-" + GraknVersion.VERSION + ".zip";
-    private static final String CURRENT_DIRECTORY = GraknSystemProperty.PROJECT_RELATIVE_DIR.value();
-    private static final String TARGET_DIRECTORY = CURRENT_DIRECTORY + "/grakn-dist/target/";
-    private static final String DIST_DIRECTORY = TARGET_DIRECTORY + "grakn-dist-" + GraknVersion.VERSION;
+    private static final Path ZIP = Paths.get("grakn-dist-" + GraknVersion.VERSION + ".zip");
```
fppt: yay!

grakn-test/test-integration/src/test/java/ai/grakn/test/rule/DistributionContext.java
```diff
@@ -88,10 +92,7 @@ public SimpleURI uri(){
 
     @Override
     protected List<TestRule> testRules() {
-        return ImmutableList.of(
-                SessionContext.create(),
-                InMemoryRedisContext.create(redisPort)
-        );
+        return ImmutableList.of(session, redis);
```
fppt: This is kind of funny when you really think about it :D

grakn-client/src/main/java/ai/grakn/client/Client.java
aelred: This is just an assertion that always fails. Basically we should never hit this code. OH I JUST REMEMBERED WE HAVE `CommonUtil.unreachableStatement()` FOR THIS

grakn-test/test-integration/src/test/java/ai/grakn/test/rule/DistributionContext.java
aelred: Haha 😢

PR #2564, opened 2018-01-08
Several refactors and clean-ups

aelred:


aelred:
@marco-scoppetta yeah, I agree we need to change the name.

It makes `GraknEngineServer`s so it should be `GraknEngineServerFactory` but that's so sad :(

How about `GraknServerFactory` or something?

aelred:
@lolski I'm happy with that idea. It is what we mostly call it.

PR #2568, opened 2018-01-09
Update license

fppt:


aelred:
@fppt this is good to merge right? I'm a bit paranoid about it

fppt:
I think it is

PR #2570, opened 2018-01-09
Rename SNB and biomed as "end-to-end" tests

aelred:


aelred:
If Thanh was here he would want me to merge this

PR #2571, opened 2018-01-09
[Stable] Fix some query cache inconsistencies

kasper-piskorski:
Initially I thought this PR was going to solve more fundamental issues, but what it really does is that it ensures consistent behaviour of the cache API which is tested via tests mimicking common use cases.

### DIFF
--- a/grakn-graql/src/test/java/ai/grakn/graql/internal/reasoner/QueryCacheTest.java
+++ b/grakn-graql/src/test/java/ai/grakn/graql/internal/reasoner/QueryCacheTest.java
@@ -33,8 +33,10 @@ import ai.grakn.graql.internal.reasoner.query.ReasonerQueries;
 import ai.grakn.test.rule.SampleKBContext;
 import ai.grakn.util.GraknTestUtil;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Sets;
 import java.util.Set;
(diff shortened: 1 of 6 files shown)


lolski:
@kasper-piskorski As an unfamiliar reader, more descriptions about the PR as has been outlined by @aelred 's w.r.t the new review process would be helpful, in my opinion

kasper-piskorski:
@fppt @lolski review? nothing ground-breaking in this PR

grakn-graql/src/test/java/ai/grakn/graql/internal/reasoner/QueryCacheTest.java
```diff
@@ -0,0 +1,252 @@
...
+import org.junit.Test;
+
+import static ai.grakn.graql.Graql.var;
+import static java.util.stream.Collectors.toSet;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeTrue;
+
+public class QueryCacheTest {
```
fppt: I think all of the test names here could do with improvement. 

When I looked at `record_Retrieve` I was able to know what your were testing without looking at the code. 

However, when looking at  `stream_Record_Update_Retrieve` I had to look at the details of the method to know what is going on.

PR #2572, opened 2018-01-09
[Stable] Fix inconsistencies of atomic states wrt type inference

kasper-piskorski:
Fixes the following inconsistency in handling atomic states, for an atomic query with a missing type:
- for an atomic query being the top query or atomic rule body, the corresponding AtomicState would not expand possible types (leading to ambiguous entries in cache and other problems)
- in other cases, i.e. atomic query produced as a part of a conjunction will have its possible types expanded

### DIFF
--- a/grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/type/IsaAtom.java
+++ b/grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/type/IsaAtom.java
@@ -21,6 +21,7 @@ package ai.grakn.graql.internal.reasoner.atom.binary.type;
 import ai.grakn.concept.ConceptId;
 import ai.grakn.concept.EntityType;
 import ai.grakn.concept.SchemaConcept;
+import ai.grakn.concept.Type;
 import ai.grakn.graql.Graql;
 import ai.grakn.graql.Pattern;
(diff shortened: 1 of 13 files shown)


marco-scoppetta:
Code-wise it looks neat. I think a description would be specially useful on PRs concerning Reasoner, I know the email is from yesterday night so it's not expected on this PR.

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/type/IsaAtom.java
```diff
@@ -107,12 +111,26 @@ private IsaAtom inferEntityType(Answer sub){
         return this;
     }
 
+    private ImmutableList<Type> inferPossibleEntityTypes(Answer sub){
+        if (getSchemaConcept() != null) return ImmutableList.of(this.getSchemaConcept().asType());
+        if (sub.containsVar(getPredicateVariable())) return ImmutableList.of(sub.get(getPredicateVariable()).asType());
+        return ImmutableList.copyOf(tx().admin().getMetaConcept().subs().iterator());
```
aelred: We have some collectors in `CommonUtil` you can use for this:
```
return tx().admin().getMetaConcept().subs().collect(toImmutableList());
```

PR #2573, opened 2018-01-11
Bug #18096: Relationships can no longer be committed without a roleplayer

fppt:
This adds a new validation rule and cache which makes sure we can't commit relationships without a role player

### DIFF
--- a/grakn-kb/src/main/java/ai/grakn/kb/internal/ValidateGlobalRules.java
+++ b/grakn-kb/src/main/java/ai/grakn/kb/internal/ValidateGlobalRules.java
@@ -388,4 +388,15 @@ class ValidateGlobalRules {
 
         return errors;
     }
+
+    /**
+     * Checks if a {@link Relationship} has at least one role player.
+     * @param relationship The {@link Relationship} to check
+     */
+    static Optional<String> validateRelationshipHasRolePlayers(Relationship relationship) {
+        if(!relationship.rolePlayers().findAny().isPresent()){
+            return Optional.of(ErrorMessage.VALIDATION_RELATIONSHIP_WITH_NO_ROLE_PLAYERS.getMessage(relationship.getId(), relationship.type().getLabel()));
+        }
+        return Optional.empty();
+    }
 }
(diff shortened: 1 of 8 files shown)


grakn-kb/src/test/java/ai/grakn/kb/internal/concept/RelationshipTest.java
```diff
@@ -311,4 +314,17 @@ public void whenAttributeLinkedToRelationshipIsInferred_EnsureItIsMarkedAsInferr
         RelationshipImpl.from(relationship).attributeInferred(attribute);
         assertTrue(relationship.relationships().findAny().get().isInferred());
     }
+
+    @Test
+    public void whenAddingRelationshipWithNoRolePlayers_Throw(){
```
lolski: Just double checking, do we already have a test which verifies the happy path?

grakn-kb/src/main/java/ai/grakn/kb/internal/ValidateGlobalRules.java
```diff
@@ -388,4 +388,15 @@ private static ReasonerQuery combinedRuleQuery(GraknTx graph, Rule rule){
 
         return errors;
     }
+
+    /**
+     * Checks if a {@link Relationship} has at least one role player.
+     * @param relationship The {@link Relationship} to check
+     */
+    static Optional<String> validateRelationshipHasRolePlayers(Relationship relationship) {
```
lolski: Out of curiosity, is it ok to have a relationship with, let's say, 3 roles of which only 1 has a role player?

kasper-piskorski: Yes. You can have a single role-roleplayer mapping for a given relation instance plus you can also have a single roleplayer playing multiple roles.

grakn-kb/src/test/java/ai/grakn/kb/internal/concept/RelationshipTest.java
fppt: Yep we do.

PR #2574, opened 2018-01-11
Provide standard github PR template

lolski:
# Why is this PR needed?
In order to help reviewers review pull requests. It is easier if the PR description follows some standard format with relevant informations

# What does the PR do?
Provides a standard template for every pull requests. This is done following the Github tutorial for setting up standard PR message here: https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/

No

None


docs/PR.md
```diff
@@ -0,0 +1,7 @@
+# Why is this PR needed?
+
+# What does the PR do?
+
+# Does it break backwards compatibility?
+
+# List of future improvements not on this PR
```
aelred: I like these points!

I would make it less wordy though. Making this too heavy might discourage people from making smaller PRs. For example, most of my PRs don't have breaking changes and don't have future improvements (such as refactors). I probably only want to write a short sentence describing "what" and "why".

How about this? Happy to discuss further!

-----


### Description

**What** does this do and **why**?

### Future improvements

Remove if not relevant

- [ ] Breaking change

-----

aelred:
[Here's](https://github.com/devspace/awesome-github-templates#rocket-templates-for-pull-requests) a repo with some PR templates if we want to peek/steal.

lolski:
Cool. This one's bullet points are based on the ones we agreed during the meeting.

docs/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,7 @@
+# Why is this PR needed?
+
+# What does the PR do?
+
+# Does it break backwards compatibility?
```
fppt: We could make this a checkbox to be fancy. . . I don't know how though :)

docs/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,7 @@
...
+
+# List of future improvements not on this PR
```
fppt: This heading is unclear to me. Is this referring to things we can still do relating to this PR to make it better?

lolski: Yes, mainly refers to the technical debt you introduce that you wish you had the time to fix. 
Think of debt that are direcly related to your PR such as these:
- I introduced a `Tuple` library since I need to create a pair of 3 objects. We already have a `Pair` library and this is an overlap we should fix in the future
- This CSV file parser is needed only for migration but at the moment it can only parse file encoded as ASCII and not UTF-8. This should work for now but needs to be improved in the future

What do you think is a more appropriate heading for this? @fppt

docs/PULL_REQUEST_TEMPLATE.md
lolski: I could easily find out, if you think it improves the workflow.

Would it still makes sense though if we sometimes want to put an extra remark such as these: https://github.com/graknlabs/grakn/pull/2542

docs/PULL_REQUEST_TEMPLATE.md
fppt: Maybe `Technical Debt To Be Addressed and Possible Optimisations To make`?

lolski:
!rtg - intermittent error due to downloading failure

docs/PULL_REQUEST_TEMPLATE.md
lolski: sounds good to me. @aelred do you have any opinion on this matter?

aelred: I'm pretty keen on keeping it short, so "Future Work" or "Future Improvements" is clear to me. But I'm OK with @fppt 's suggestion too.

fppt: I am not super fussed. I just had to guess a little what `Future Work` meant. With no context it could mean:

1. Technical Debt Introduced in this PR
2. The next thing I am going to work on which may or may not be related to this PR
3. Performance improvements

Again, not super fussed.

PR #2575, opened 2018-01-11
Bug #18867: Fix clashing rules

fppt:
Sometimes rules with different patterns would clash because of accidentally resulting in the same index. Turns out we don't actually need that index anymore so this PR removes that index.

### DIFF
--- a/grakn-graql/src/test/java/ai/grakn/kb/internal/RuleTest.java
+++ b/grakn-graql/src/test/java/ai/grakn/kb/internal/RuleTest.java
@@ -89,6 +89,22 @@ public class RuleTest {
     }
 
     @Test
+    public void whenCreatingDistinctRulesWithSimilarStringHashes_EnsureRulesDoNotClash(){
+        String productRefused = "productRefused";
+        Pattern when1 = graknTx.graql().parser().parsePattern("{$step has step-id 9; $e (process-case: $case) isa process-record; $case has consent false;}");
+        Pattern then1 = graknTx.graql().parser().parsePattern("{(record: $e, step: $step) isa record-step;}");
+        Rule rule1 = graknTx.putRule(productRefused, when1, then1);
+
+        String productAccepted = "productAccepted";
+        Pattern when2 = graknTx.graql().parser().parsePattern("{$step has step-id 7; $e (process-case: $case) isa process-record; $case has consent true;}");
+        Pattern then2 = graknTx.graql().parser().parsePattern("{(record: $e, step: $step) isa record-step;}");
+        Rule rule2 = graknTx.putRule(productAccepted, when2, then2);
+
+        assertEquals(rule1, graknTx.getRule(productRefused));
+        assertEquals(rule2, graknTx.getRule(productAccepted));
+    }
+
+    @Test
     public void whenAddingRuleWithDisjunctionInTheBody_Throw() throws InvalidKBException {
         validateIllegalRule(
                 graknTx.graql().parser().parsePattern("(role: $x) or (role: $x, role: $y)"),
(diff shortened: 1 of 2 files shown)


grakn-kb/src/main/java/ai/grakn/kb/internal/concept/RuleImpl.java
```diff
@@ -110,13 +109,6 @@ private Pattern parsePattern(String value){
         }
     }
 
-    /**
-     * Generate the internal hash in order to perform a faster lookups and ensure rules are unique
-     */
-    static String generateRuleIndex(Rule type, Pattern when, Pattern then){
-        return "RuleType_" + type.getLabel().getValue() + "_LHS:" + when.hashCode() + "_RHS:" + then.hashCode();
-    }
-
```
kasper-piskorski: So is there a performance hit for rule lookups since the index is removed? Was it a hash for the internal cache? I don't think I am personally using the get-rule-by-label option anyway.

fppt: It was mostly used to ensure that you could not create two rules with the same pattern. But the protection was very artificial since simply using different variable names in the rules would let you bypass it.

Now we just removed that index and let duplicate rules be created. I.e. we allow the user error now.

kasper-piskorski: We could potentially add a validation rule for duplicates. Is there any gain in that? Reasoner recognises duplicates anyway.

fppt: Hmmm we could. I was not aware reasoner could recognise dups. If that's the case we should use that mechanism at commit time. Can you show me where that code is and how I might repurpose it for commit time checks?

kasper-piskorski: I just had a look, so in theory you could compare bodies/heads by doing sth like:

```
ReasonerQuery body1 = Iterables.getOnlyElement(rule1.getWhen().admin().getDisjunctiveNormalForm().getPatterns()).toReasonerQuery(graph);
ReasonerQuery body2 = Iterables.getOnlyElement(rule2.getWhen().admin().getDisjunctiveNormalForm().getPatterns()).toReasonerQuery(graph);
```
and then simply doing

```
body1.equals(body2);
```

However the algorithm behind equals cannot recognise different variable binding patterns so two queries:

```
(chain)
($x, $y) isa some-relation;
($y, $z) isa some relation;
($z, $w) isa some-relation;
```
and

```
(loop)
($x, $y) isa some-relation;
($y, $z) isa some relation;
($z, $x) isa some-relation;
```
will be recognised as equal.

fppt: Ah I already see an issue. `ReasonerQuery` is in `grakn-graql` commit time validation is done only with the presence of `grakn-kb` so we don't have access to those classes at commit time. 

We might need to invert the dependencies again

kasper-piskorski: `ReasonerQuery` is in `grakn-core`

fppt: Oh just to be pedantic I will make that change in another PR

grakn-graql/src/test/java/ai/grakn/kb/internal/RuleTest.java
```diff
@@ -88,6 +88,22 @@ public void whenCreatingRulesWithNonExistentEntityType_Throw() throws InvalidKBE
...
+        Rule rule1 = graknTx.putRule(productRefused, when1, then1);
+
+        String productAccepted = "productAccepted";
+        Pattern when2 = graknTx.graql().parser().parsePattern("{$step has step-id 7; $e (process-case: $case) isa process-record; $case has consent true;}");
+        Pattern then2 = graknTx.graql().parser().parsePattern("{(record: $e, step: $step) isa record-step;}");
+        Rule rule2 = graknTx.putRule(productAccepted, when2, then2);
+
+        assertEquals(rule1, graknTx.getRule(productRefused));
+        assertEquals(rule2, graknTx.getRule(productAccepted));
+    }
```
aelred: This is probably a super-fragile test (since these hashes just _happen_ to be the same). Maybe it can be done with crazy mocking?

PR #2577, opened 2018-01-11
Simplify and cleanup query equivalence

kasper-piskorski:


duckofyork:
@kasper-piskorski Filipe has not approved your PR yet...

PR #2578, opened 2018-01-12
Path concatenation cleanup in DistributionContext

lolski:
# Why is this PR needed?
Cleaning up `DistributionContext`. Path concatenations are done manually by concatenating strings. This is considered to be error-prone.

# What does the PR do?
1. perform cleanup around the building of path. Instead of via string concatenation, it is done via `Paths.get()` which is less error prone.
2. Updating header comment

No

None


grakn-test/test-integration/src/test/java/ai/grakn/test/rule/DistributionContext.java
```diff
@@ -60,11 +63,11 @@
...
     public static final Logger LOG = LoggerFactory.getLogger(DistributionContext.class);
 
-    private static final FilenameFilter jarFiles = (dir, name) -> name.toLowerCase().endsWith(".jar");
-    private static final Path ZIP = Paths.get("grakn-dist-" + GraknVersion.VERSION + ".zip");
-    private static final Path CURRENT_DIRECTORY = Paths.get(GraknSystemProperty.PROJECT_RELATIVE_DIR.value());
-    private static final Path TARGET_DIRECTORY = CURRENT_DIRECTORY.resolve(Paths.get("grakn-dist", "target"));
-    private static final Path DIST_DIRECTORY = TARGET_DIRECTORY.resolve("grakn-dist-" + GraknVersion.VERSION);
+    private static final String ZIP_FILENAME = "grakn-dist-" + GraknVersion.VERSION + ".zip";
+    private static final Path GRAKN_BASE_DIRECTORY = Paths.get(GraknSystemProperty.PROJECT_RELATIVE_DIR.value());
+    private static final Path TARGET_DIRECTORY = Paths.get(GRAKN_BASE_DIRECTORY.toString(), "grakn-dist", "target");
```
aelred: We can avoid breaking the `Path` abstraction with `toString` by using `Path#resolve`:
```
private static final Path TARGET_DIRECTORY = GRAKN_BASE_DIRECTORY.resolve(Paths.get("grakn-dist", "target"));
```
There are a few other places where we can do the same thing.

...but it's also a bit less pretty, so it's up to you if you want to do it or not!

lolski: Yeah that is true. I switched the style since in my personal opinion it is a bit less readable :D

kasper-piskorski: so could `ZIP_FILENAME` use the `Paths.get()` instead of concatenating strings?

aelred: @lolski yeah, I was half-half on which looks better too. I don't think there's much difference, so I'm good with this.

PR #2579, opened 2018-01-12
Introduce new `GraqlConverter` interface

aelred:
# Why is this PR needed?

gRPC will need a way to convert Graql results into streams of gRPC data. Currently, we only support converting results into streams of *strings*.

# What does the PR do?

We introduce a new `GraqlConverter` interface that `Printer` extends.

No. I consider only `Printer#graqlString(Object)` and `Query#resultsString(Printer)` to be part of our public `Printer` API and I have deprecated both methods.

There are other "public" methods in this PR that I've renamed, but I don't consider them part of our API.

Actually using this new abstraction! gRPC will likely use it, but it may come in handy in other places (for example, the `JacksonPrinter` may be able to return something more descriptive than `String`).


grakn-engine/src/main/java/ai/grakn/engine/controller/response/ExplanationBuilder.java
```diff
@@ -59,7 +59,8 @@
                         .flatMap(unifier -> rewrittenQuery.materialise(answer.unify(unifier)))
                         .findFirst().orElse(new QueryAnswer());
             }
-            explanation.add((Answer) printer.graqlString(false, inferredAnswer));
+            // TODO: this looks like a bug: the return type is not `Answer`
```
fppt: Not a bug just poor design. In the context of the explanation builder the return type is actually always an answer.

grakn-core/src/main/java/ai/grakn/graql/GraqlConverter.java
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * An interface for converting objects in Graql responses into other formats.
+ *
+ * @param <Builder> An intermediate builder type that can be changed into a {@link T}
+ * @param <T> The type to convert into
+ *
+ * @author Felix Chapman
+ */
+public interface GraqlConverter<Builder, T> {
```
fppt: The distinction between `GraqlConverter` and `Printer` is kind of vague. Maybe even just changing the to be more explicit to something like:

```
An interface for converting any objects (e.g. Answers, Concepts, etc . . . ) in Graql responses into specific Types `T`.
```

grakn-core/src/main/java/ai/grakn/graql/GraqlConverter.java
```diff
@@ -0,0 +1,147 @@
...
+
+import javax.annotation.CheckReturnValue;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * An interface for converting objects in Graql responses into other formats.
+ *
+ * @param <Builder> An intermediate builder type that can be changed into a {@link T}
```
fppt: Struggled to understand why this was needed. Not sure how to improve on this.

grakn-engine/src/main/java/ai/grakn/engine/controller/response/ExplanationBuilder.java
aelred: Thanks! I get it. So `JacksonPrinter` returns `Object` and therefore sometimes it is an `Answer`. In this context, we know that it must be.

@fppt suggested we can make this a bit clearer by making this method take a `Printer<Object>` or `JacksonPrinter` explicitly.

PR #2582, opened 2018-01-15
Merge stable into master 

kasper-piskorski:


kasper-piskorski:
@fppt waiting for final approval :)

PR #2583, opened 2018-01-15
New Analytics Method : Coreness

duckofyork:
# Why is this PR needed?
New analytics method: centrality based on k-core, or "coreness"
# What does the PR do?
Add coreness query and its impl
No
Unify degree and coreness, as they are both centrality measures.
Unify connected component and k-core, as they are cluster queries.
Update graql syntax.

### DIFF
--- a/grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/CorenessQueryImpl.java
+++ b/grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/CorenessQueryImpl.java
@@ -1,0 +1,157 @@
+/*
+ * Grakn - A Distributed Semantic Database
+ * Copyright (C) 2016  Grakn Labs Limited
+ *
+ * Grakn is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Grakn is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Grakn. If not, see <http://www.gnu.org/licenses/gpl.txt>.
+ */
+
+package ai.grakn.graql.internal.query.analytics;
+
+import ai.grakn.GraknTx;
+import ai.grakn.concept.Label;
+import ai.grakn.concept.LabelId;
+import ai.grakn.exception.GraqlQueryException;
+import ai.grakn.graql.analytics.CorenessQuery;
+import ai.grakn.graql.internal.analytics.ClusterMemberMapReduce;
+import ai.grakn.graql.internal.analytics.KCoreVertexProgram;
+import ai.grakn.graql.internal.analytics.NoResultException;
+import ai.grakn.graql.internal.util.StringConverter;
+import com.google.common.collect.Sets;
+import org.apache.tinkerpop.gremlin.process.computer.ComputerResult;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static java.util.stream.Collectors.joining;
+
+class CorenessQueryImpl extends AbstractComputeQuery<Map<Integer, Set<String>>> implements CorenessQuery {
+
+    private int k = 2;
+    private boolean ofTypeLabelsSet = false;
+    private Set<Label> ofLabels = new HashSet<>();
+
+    CorenessQueryImpl(Optional<GraknTx> graph) {
+        this.tx = graph;
+    }
+
+    @Override
+    public Map<Integer, Set<String>> execute() {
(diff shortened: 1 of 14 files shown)


grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/CorenessQueryImpl.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    @Override
+    public Map<Integer, Set<String>> execute() {
+        LOGGER.info("Coreness query is started");
+        long startTime = System.currentTimeMillis();
+
+        if (k < 2) throw GraqlQueryException.kValueSmallerThanTwo();
+
+        includeAttribute = true; //TODO: REMOVE THIS LINE
```
aelred: Can we remove it now?

grakn-graql/src/test/java/ai/grakn/graql/internal/analytics/CorenessTest.java
```diff
@@ -0,0 +1,349 @@
...
+    @ClassRule
+    public final static SessionContext sessionContext = SessionContext.create();
+
+    @Before
+    public void setUp() {
+        session = sessionContext.newSession();
+    }
+
+    @Test(expected = GraqlQueryException.class)
+    public void testKSmallerThan2() {
```
aelred: We should explain that we expect this to throw in the test name, like `testKSmallerThan2_ThrowsException`

grakn-graql/src/test/java/ai/grakn/graql/internal/analytics/CorenessTest.java
```diff
@@ -0,0 +1,349 @@
...
+        try (GraknTx graph = session.open(GraknTxType.READ)) {
+            graph.graql().compute().coreness().minK(1).execute();
+        }
+    }
+
+    @Test
+    public void testOnEmptyGraph() {
```
aelred: We can also say the expected result here, like `testOnEmptyGraph_ReturnsEmptyMap`. Same for the tests below too!

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/CorenessQueryImpl.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    @Override
+    public CorenessQuery of(String... ofTypeLabels) {
+        if (ofTypeLabels.length > 0) {
+            ofTypeLabelsSet = true;
+            this.ofLabels = Arrays.stream(ofTypeLabels).map(Label::of).collect(Collectors.toSet());
+        }
+        return this;
+    }
```
aelred: Even better is if we define this in terms of the other `of` method below!

```
public CorenessQuery of(String... ofTypeLabels) {
    return of(Arrays.stream(ofTypeLabels).map(Label::of).collect(Collectors.toSet());
}
```

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/CorenessQueryImpl.java
duckofyork: Not yet. I will do it in another pr, so we can remove it from all analytics queries

grakn-graql/src/test/java/ai/grakn/graql/internal/analytics/CorenessTest.java
```diff
@@ -0,0 +1,349 @@
...
+
+            entityId1 = entity1.getId();
+            entityId2 = entity2.getId();
+            entityId3 = entity3.getId();
+            entityId4 = entity4.getId();
+
+            graph.commit();
+        }
+    }
+}
```
kasper-piskorski: Might be worth describing the configuration graphically, saves time when dealing with test failures.

grakn-graql/src/test/java/ai/grakn/graql/internal/analytics/CorenessTest.java
```diff
@@ -0,0 +1,349 @@
...
+            graph.putRelationshipType(veryRelated)
+                    .relates(role3).relates(role4)
+                    .addRelationship()
+                    .addRolePlayer(role3, entity1)
+                    .addRolePlayer(role4, entity2);
+
+            Map<Integer, Set<String>> result = graph.graql().compute().coreness().execute();
+            assertTrue(result.isEmpty());
+        }
+    }
```
kasper-piskorski: Just trying to have a general understanding. I understand the default k is 2. So when executing coreness for what values of k are calculations performed? Single specific value (here the default I guess)? Range of values starting from min k?

duckofyork: A vertex u has coreness c if it belongs to a c-core but not to any (c+1)-core.
In coreness query, we start with k = minK, keep persisting and updating coreness of vertices.

grakn-graql/src/test/java/ai/grakn/graql/internal/analytics/CorenessTest.java
duckofyork: That's a good idea man. I will try to draw this in the test, something like this:
```/*
        root    10
    	(0) -------> (1) \
    	 |  \       /  ^  \
    	 |   \30   |   |20 \
    	 |10  \    |10 |    \10
    	 |     \   |  /      \
    	 V  15  V  V /   20   V
    	(3)<----- (2) -----> (4)
    	  \-------^
    	     40
    	 */

PR #2584, opened 2018-01-17
Make analytics count return the same result as aggregate count

duckofyork:
# Why is this PR needed?
Analytics count and aggregate count may return different numbers
# What does the PR do?
Fix that
No
Remove variable `includeAttributes` from all compute queries


duckofyork:
From our documentation http://dev.grakn.ai/docs/distributed-analytics/compute-queries#available-statistics-algorithms
```Count
The default behaviour of count is to return a single value that gives the number of instances present in the graph. ```

PR #2592, opened 2018-01-23
Add CentralityQueryBuilder

duckofyork:
# Why is this PR needed?
Make analytics queries more consistent
# What does the PR do?
Add `CentralityQueryBuilder`, replacing `compute().degree()` and `compute().coreness()`
No
Remove degree and coreness query from `ComputeQueryBuilder`


grakn-graql/src/test/java/ai/grakn/graql/internal/parser/QueryParserTest.java
```diff
@@ -645,9 +646,10 @@ public void testParseComputeClusterWithSizeThenMembersThenSize() {
         assertEquals(expected, parsed);
     }
 
+    @Ignore
     @Test
     public void testParseComputeDegree() {
-        assertParseEquivalence("compute degrees in movie;");
+        assertParseEquivalence("compute centrality in movie; using degree;");
     }
```
aelred: Is `compute degrees in movie;` still supported? If it is, we can just leave this test as it is and not ignore it.

Otherwise, we should add a `TODO` to say we have to wait until we support it in the parser.

(same for the other tests below)

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractCentralityQuery.java
```diff
@@ -0,0 +1,110 @@
...
+                    .map(StringConverter::typeLabelToString)
+                    .collect(joining(", "));
+        }
+        string += subtypeString() + " using " + getMethod().getName();
+
+        return string;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
fppt: These types of equality and hashcode makes me wonder if this class would be a good candidate for `@AutoValue`. If the sets, labels, etc. . . are never mutated I would consider it

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/CorenessQueryImpl.java
```diff
@@ -95,32 +86,28 @@
...
-    public CorenessQuery minK(int k) {
+    public CorenessQuery minK(long k) {
         this.k = k;
         return this;
     }
 
     @Override
     public CorenessQuery of(String... ofTypeLabels) {
-        return of(Arrays.stream(ofTypeLabels).map(Label::of).collect(Collectors.toSet()));
+        return (CorenessQuery) super.setOfTypeLabels(ofTypeLabels);
```
fppt: Could this sort of casting be solved by having the super method return a generic?

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractCentralityQuery.java
aelred: I thought the same, but it looks like the fields like `ofLabels` do get modified.

grakn-graql/src/test/java/ai/grakn/graql/internal/parser/QueryParserTest.java
duckofyork: We plan to remove them eventually, for now I'll make them deprecated and add a `TODO` as well.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractCentralityQuery.java
```diff
@@ -0,0 +1,110 @@
...
+        @CheckReturnValue
+        public String getName() {
+            return name;
+        }
+    }
+
+    private boolean ofTypeLabelsSet = false;
+    Set<Label> ofLabels = new HashSet<>();
+
+    void initSubGraph() { //TODO: REMOVE THIS METHOD
```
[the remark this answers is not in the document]
duckofyork: Later. I'll try to do it in my next PR

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractCentralityQuery.java
duckofyork: The user can modify these values by calling the same method twice, e.g. `compute().count().in("person").in("dog");`

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractCentralityQuery.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    AbstractCentralityQuery setOfTypeLabels(String... ofTypeLabels) {
+        return setOfTypeLabels(Arrays.stream(ofTypeLabels).map(Label::of).collect(Collectors.toSet()));
+    }
+
+    AbstractCentralityQuery setOfTypeLabels(Collection<Label> ofLabels) {
+        if (!ofLabels.isEmpty()) {
+            ofTypeLabelsSet = true;
+            this.ofLabels = Sets.newHashSet(ofLabels);
```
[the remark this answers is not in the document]
duckofyork: Just wanna make it a bit flexible, so people can use other type of Collection, such as list and

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/CorenessQueryImpl.java
duckofyork: I have removed this method using generic. Please have a look.

duckofyork:
@rtg

duckofyork:
@rtg

duckofyork:
!rtg

duckofyork:
@aelred Do you mind reviewing another commit? I have added more generic stuff.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractCentralityQuery.java
fppt: Awwwww. . . .

PR #2593, opened 2018-01-23
Introduce gRPC server

aelred:
# Why is this PR needed?

To support remote clients, that can execute queries within transactions etc.

# What does the PR do?

Introduces a server that responds to gRPC requests. It supports opening transactions and executing queries against those transactions. Responses are streamed out lazily (so you must send a `Next` message to receive each result).

No.

A lot! We need to support the full `Concept` model on the server - so you can do stuff like `personType.addEntity()`. Also, this doesn't support multiple simultaneous queries - which is probably a fairly important feature.

There is also [this](https://github.com/graknlabs/grakn-spec/pull/6) PR on `grakn-spec`, that will host the proto file. Once that's merged, we can add it as a submodule similar to our clients. That way the proto file is stored in just one place.


aelred:
btw we should probably merge this #2594 first, since it contains some of the change in this PR.

grakn-engine/src/main/java/ai/grakn/engine/GraknCreator.java
```diff
@@ -119,7 +121,21 @@ public synchronized GraknEngineServer instantiateGraknEngineServer(Runtime runti
...
             TaskManager taskManager = instantiateTaskManager(metricRegistry, graknEngineConfig, engineID, factory, jedisPool, postProcessor);
-            HttpHandler httpHandler = new HttpHandler(graknEngineConfig, sparkService, factory, metricRegistry, graknEngineStatus, taskManager, postProcessor);
+
+            int grpcPortPort = graknEngineConfig.getProperty(GraknConfigKey.GRPC_PORT);
+            GrpcServer grpcServer;
+            try {
+                grpcServer = GrpcServer.create(grpcPortPort, engineGraknTxFactory);
+            } catch (IOException e) {
+                // Thrown if unable to bind to the given port
+                throw new IllegalStateException(e);
```
marco-scoppetta: Just out of curiousity: why is the IllegalState exception better than the IO exception here?

grakn-core/src/main/java/ai/grakn/graql/Query.java
```diff
@@ -60,7 +60,7 @@
      * Execute the query and return a converted stream of results
      */
     @CheckReturnValue
-    <T> Stream<T> results(GraqlConverter<?, T> converter);
+    <S> Stream<S> results(GraqlConverter<?, S> converter);
```
marco-scoppetta: why T become S? is it because we are handling Streams here?

aelred: No, it's more boring than that! We were already using `T` in this class (`Query<T>`), so it was confusing.

grakn-engine/src/main/java/ai/grakn/engine/rpc/GrpcServer.java
```diff
@@ -0,0 +1,93 @@
...
+            this.txFactory = txFactory;
+        }
+
+        @Override
+        public StreamObserver<TxRequest> tx(StreamObserver<TxResponse> responseObserver) {
+            return TxObserver.create(txFactory, responseObserver);
+        }
+    }
+
+    private static class StringMarshaller implements Metadata.AsciiMarshaller<String> {
```
marco-scoppetta: Maybe small comment to explain what's this

grakn-engine/src/main/java/ai/grakn/engine/GraknCreator.java
aelred: I said this in person, but I just wanted an unchecked exception. The `IOException` is probably more correct though.

aelred:
@marco-scoppetta I think this is ready for re-review now

grakn-engine/src/main/java/ai/grakn/engine/rpc/TxObserver.java
```diff
@@ -0,0 +1,220 @@
...
+
+        sendNextResult(queryResults);
+    }
+
+    private void next() {
+        if (queryResults == null) {
+            throw error(Status.FAILED_PRECONDITION);
+        }
+
+        sendNextResult(queryResults);
```
marco-scoppetta: Why do we pass around the queryResults as a parameter? Since it's already defined as a class property

grakn-test-tools/src/main/java/ai/grakn/engine/rpc/SynchronousObserver.java
```diff
@@ -0,0 +1,190 @@
...
+ *     }
+ * }
+ * </pre>
+ *
+ * @author Felix Chapman
+ *
+ * @param <Request> The type of requests being received
+ * @param <Response> The type of responses being sent
+ */
+public class SynchronousObserver<Request, Response> implements AutoCloseable {
```
fppt: Lets chat about this one in person. I think I understand it but I want to double check. I think this is used to convert the asynchronous conversation into a synchronous one. I just want to check how.

grakn-test-tools/src/main/java/ai/grakn/engine/rpc/SynchronousObserver.java
```diff
@@ -0,0 +1,190 @@
...
+     *     So, we cannot get at the {@link Request} observer until we provide a {@link Response} observer.
+     *     Unfortunately the latter is created within this method below - so we must pass in a method reference:
+     * </p>
+     *
+     * {@code SynchronousObserver.create(stub::tx)}
+     */
+    public static <Request, Response> SynchronousObserver<Request, Response> create(
+            Function<StreamObserver<Response>, StreamObserver<Request>> createRequestObserver
+    ) {
+        QueueingObserver<Response> responses = new QueueingObserver<>();
```
fppt: Maybe state that you are only wrapping the responses into a blocking approach. I.e. you can still send requests asynchronously.

grakn-engine/pom.xml
```diff
@@ -120,5 +120,49 @@
...
+            <artifactId>grpc-stub</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-netty</artifactId>
+        </dependency>
     </dependencies>
+
+    <build>
+        <extensions>
```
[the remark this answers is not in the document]
aelred: Yes, this is what generates the code from the proto file. It's huge and I don't know which bits are important! I will try removing bits of it and see.

grakn-engine/src/main/java/ai/grakn/engine/GraknCreator.java
```diff
@@ -106,28 +110,28 @@ public static GraknCreator create() {
...
             LockProvider lockProvider = instantiateLock(jedisPool);
             EngineGraknTxFactory factory = instantiateGraknTxFactory(graknEngineConfig, lockProvider);
-
             IndexPostProcessor indexPostProcessor = IndexPostProcessor.create(lockProvider, RedisIndexStorage.create(jedisPool, metricRegistry));
             CountPostProcessor countPostProcessor = CountPostProcessor.create(graknEngineConfig, factory, lockProvider, metricRegistry, RedisCountStorage.create(jedisPool, metricRegistry));
 
             PostProcessor postProcessor = PostProcessor.create(indexPostProcessor, countPostProcessor);
-            HttpHandler httpHandler = new HttpHandler(graknEngineConfig, sparkService, factory, metricRegistry, graknEngineStatus, postProcessor);
-
+            int grpcPortPort = graknEngineConfig.getProperty(GraknConfigKey.GRPC_PORT);
```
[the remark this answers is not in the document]
aelred: Will change to `grpcPortPortPort`

grakn-engine/src/main/java/ai/grakn/engine/rpc/TxObserver.java
```diff
@@ -0,0 +1,220 @@
...
+/**
+ * A {@link StreamObserver} that implements the transaction-handling behaviour for {@link GrpcServer}.
+ *
+ * <p>
+ * Receives a stream of {@link TxRequest}s and returning a stream of {@link TxResponse}s.
+ * </p>
+ *
+ * @author Felix Chapman
+ */
+class TxObserver implements StreamObserver<TxRequest>, AutoCloseable {
```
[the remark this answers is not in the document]
aelred: Yes, one per transaction! Maybe it could be called `TxHandler` or something.

grakn-test-tools/pom.xml
```diff
@@ -92,5 +92,15 @@
             <artifactId>mockito-core</artifactId>
             <scope>compile</scope>
         </dependency>
+
+        <dependency>
```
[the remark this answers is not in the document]
aelred: Because of `GrpcUtil` and `SynchronousObserver`, which use gRPC classes.

grakn-test-tools/src/main/java/ai/grakn/engine/rpc/GrpcUtil.java
```diff
@@ -0,0 +1,81 @@
...
+import static org.hamcrest.Matchers.hasProperty;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.isA;
+
+/**
+ * @author Felix Chapman
+ */
+public class GrpcUtil {
+
+    // TODO: dup code
```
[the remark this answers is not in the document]
aelred: This seems like it's a Gradle deal. So later :(

grakn-engine/src/test/java/ai/grakn/engine/rpc/GrpcServerTest.java
```diff
@@ -0,0 +1,578 @@
...
+    @Rule
+    public final ExpectedException exception = ExpectedException.none();
+
+    // TODO: usePlainText is not secure
+    private final ManagedChannel channel = ManagedChannelBuilder.forAddress("localhost", PORT).usePlaintext(true).build();
+    private final GraknStub stub = GraknGrpc.newStub(channel);
+
+    @Before
+    public void setUp() throws IOException {
+        server = GrpcServer.create(PORT, txFactory);
```
marco-scoppetta: This `server` is never used, do we need it?

aelred: Even though it's not used in the tests, it opens up a gRPC server that the tests talk to.

aelred:
!rtg - I fixed a submodule issue on Jenkins

aelred:
@fppt can you re-review when you have time?
also @marco-scoppetta if you like, you can re-review - there has been some more code since you approved.

PR #2601, opened 2018-01-26
Unify cluster query

duckofyork:
# Why is this PR needed?
Unify cluster query
# What does the PR do?
- change `compute().connectedComponent()` to `compute().centrality().usingConnectedComponent()`, `compute().kCore()` to `compute().centrality().usingKCore()`
- make old methods deprecated
- degree of relationships is now the number of incoming edges
Yes. `compute().cluster()` is no longer valid.
Remove old analytics methods completely?


aelred:
Since this breaks backwards compat, should we bump up to 2.0 @fppt :D

grakn-core/src/main/java/ai/grakn/graql/ComputeQueryBuilder.java
```diff
@@ -106,14 +107,20 @@
...
 
     /**
+     * This method is deprecated. Please use centrality query instead.
+     *
      * @return a cluster query that will find the connected components
      */
+    @Deprecated
     @CheckReturnValue
-    ClusterQuery<Map<String, Long>> cluster();
+    ConnectedComponentQuery<Map<String, Long>> connectedComponent();
```
aelred: Maybe we can just delete this since it wasn't in the 1.0 release 👍

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractClusterQuery.java
```diff
@@ -0,0 +1,73 @@
...
+import ai.grakn.graql.ComputeQuery;
+
+import javax.annotation.CheckReturnValue;
+
+abstract class AbstractClusterQuery<T, V extends ComputeQuery<T>>
+        extends AbstractComputeQuery<T, V> {
+    /**
+     * The centrality measures supported.
+     */
+    enum ClusterMeasure {
```
aelred: Maybe we should use `String`s, unless we have another idea about how to use this

aelred:
I talked with @fppt and the current plan is to wait for 2.0 before we merge.

duckofyork:
So shall I close it?

aelred:
@duckofyork we've decided that the only breaking changes are ones to the data model (that is, ones that prevent you from reading old knowledge graphs).

Therefore we can merge this PR! Once we fix the conflicts :P

duckofyork:
Do you guys wanna re-review or shall I just merge it?

aelred:
It looks fine to me!

PR #2604, opened 2018-01-30
EmbeddedCassandraContext accepts yaml file path as a parameter

marco-scoppetta:
# Why is this PR needed?

So that we can use `EmbeddedCassandraContext` also in KBMS tests.

# What does the PR do?

Overloads `EmbeddedCassandraContext`'s `create()` to also have `create(String yamlFilePath)`.

No.

No.


lolski:
@fppt @marco-scoppetta don't forget that we need to surface this change by deploying the artifact into maven.grakn.ai.

marco-scoppetta:
@fppt can we deploy this to maven.grakn.ai now that it's merged?

PR #2605, opened 2018-01-30
Integrate sonar with travis

fppt:
# Why is this PR needed?

So that we can be more aware of common "bad code" mistakes we are making.  

# What does the PR do?

It runs a sonar analytics passthrough and uploads the results to sonar cloud for easy display and analysis.

No

Getting sonar cloud to break the build would be awesome but this requires a webhook and our own running instance of SonarQube. Which is not impossible just more time consuming,.


fppt:
@aelred We need to check that this works once it is merged. This is a bit of a gamble

PR #2609, opened 2018-01-31
Add `grakn-grpc` module

aelred:
# Why is this PR needed?

There was a lot of code and behaviour shared between the gRPC server, server tests, client and client tests. This is a module where we can put it all.

# What does the PR do?

Adds a new module `grakn-grpc` that `grakn-engine` depends on. `grakn-grpc` also includes the gRPC dependencies and plugin that generates gRPC code.

I also took the opportunity to try and simplify the `SynchronousObserver` class. I specialised it to work only with Tx calls to gRPC, so hopefully it is easier to understand now.

No.

The gRPC client will also use this module in the future.

### DIFF
--- a/grakn-engine/pom.xml
+++ b/grakn-engine/pom.xml
@@ -62,6 +62,10 @@
             <groupId>ai.grakn</groupId>
             <artifactId>grakn-kb</artifactId>
         </dependency>
+        <dependency>
+            <groupId>ai.grakn</groupId>
+            <artifactId>grakn-grpc</artifactId>
+        </dependency>
         <!-- Exposes metrics as a servlet -->
         <dependency>
           <groupId>io.dropwizard.metrics</groupId>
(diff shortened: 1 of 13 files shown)


aelred:
@lolski we talked about it in person, but so it's written down:

Given that:
- We're going to introduce a gRPC client inside the module `grakn-client`
- The gRPC client shares a lot of code and configuration with the gRPC server (in `grakn-engine`)

We need a shared module for `grakn-client` and `grakn-server` containing the common gRPC code - `grakn-core` is not appropriate (because it should not contain implementations), so we need a new module.

Alternatives are to put the client in the same module as the engine (this is bad, because then our "thin" client will inherit all the engine dependencies). We may also want to consider splitting the modules even _more_ - e.g. having `grakn-grpc-server`, `grakn-grpc-client`.

PR #2610, opened 2018-01-31
Remove dead and unused code

fppt:
# Why is this PR needed?

It is needed as routine maintenance so we can keep our code base clean and easier to work with.

# What does the PR do?

Removes Methods which are unused.
Annotates some methods which are part of the public API.
Annotates some methods which are used indirectly, for example the OLAP constructors

Nope. Everything being removed should not be in use.

Automatically pickup dead code and break the build because of it but I have not figured out how to achieve this.

# Note:

Because I had to go through a lot of unfamiliar code I ended up tagging comments for different people to address. I would appreciate it if you guys could help me address these unknowns.


grakn-core/src/main/java/ai/grakn/graql/GetQuery.java
```diff
@@ -48,8 +47,9 @@
     /**
      * Get the transaction this query is running against, if it is set
      */
-    @CheckReturnValue
-    Optional<GraknTx> tx();
+    @SuppressWarnings("unsued") //TODO: @Felix, What's the deal here? Seems totally unused
```
fppt: @aelred

grakn-core/src/main/java/ai/grakn/graql/admin/Atomic.java
```diff
@@ -153,6 +153,9 @@
     @CheckReturnValue
     Atomic inferTypes();
 
+    //TODO @Kapser: This is only used through it's implementation class. Making the method here unneeded
```
fppt: @kasper-piskorski

grakn-core/src/main/java/ai/grakn/graql/analytics/CentralityQueryBuilder.java
```diff
@@ -37,8 +37,11 @@
      * @param tx the {@link GraknTx} to execute the compute query on
      * @return a compute query builder with the graph set
      */
-    @CheckReturnValue
-    CentralityQueryBuilder withTx(GraknTx tx);
+    //TODO: @Jason. What's the deal with this method?
```
fppt: @duckofyork

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/spanningtree/ExclusiveEdge.java
```diff
@@ -44,6 +44,7 @@ private ExclusiveEdge(DirectedEdge<V> edge, List<DirectedEdge<V>> excluded, doub
         return new ExclusiveEdge<>(edge, excluded, weight);
     }
 
+    @SuppressWarnings("unused")//TODO: @Jason, how is this used? I can't find a reference to it?
```
fppt: @duckofyork

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractCentralityQuery.java
```diff
@@ -62,6 +62,8 @@ void initSubGraph() { //TODO: REMOVE THIS METHOD
         includeAttribute = true;
     }
 
+    //TODO: @Jason. This is only ever used by a test. Is it part of the public api? Why do we need this method?
```
fppt: @duckofyork

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractStatisticsQuery.java
```diff
@@ -45,12 +45,16 @@
     Set<Label> statisticsResourceLabels = new HashSet<>();
     Set<Type> statisticsResourceTypes = new HashSet<>();
 
+    //TODO: @Jason. This is only ever used by a test. Is it part of the public api? Why do we need this method?
```
fppt: @duckofyork

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractStatisticsQuery.java
```diff
@@ -45,12 +45,16 @@
...
+    @SuppressWarnings("unused")
     public V of(String... statisticsResourceTypeLabels) {
         this.statisticsResourceLabels =
                 Arrays.stream(statisticsResourceTypeLabels).map(Label::of).collect(Collectors.toSet());
         return (V) this;
     }
 
+    //TODO: @Jason. This is only ever used by a test. Is it part of the public api? Why do we need this method?
```
fppt: @duckofyork

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/cache/Cache.java
```diff
@@ -145,6 +145,9 @@ public void add(Cache<Q, T> c2){
      * @param answers answer stream of the query
      * @return lazy iterator of updated answers
      */
+    //TODO: @Kasper Is this abstract method on `Cache` really needed?
```
fppt: @kasper-piskorski

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/cache/Cache.java
```diff
@@ -153,9 +156,18 @@ public void add(Cache<Q, T> c2){
      * @return unified cached answers
      */
     public abstract T getAnswers(Q query);
+    //TODO: @Kasper Is this abstract method on `Cache` really needed?
```
fppt: @kasper-piskorski

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/cache/Cache.java
```diff
@@ -153,9 +156,18 @@ public void add(Cache<Q, T> c2){
...
+    //      It's only ever called through LazyQueryCache
+    @SuppressWarnings("unused")
     public abstract Pair<T, MultiUnifier> getAnswersWithUnifier(Q query);
     public abstract Stream<Answer> getAnswerStream(Q query);
+    //TODO: @Kasper Is this abstract method on `Cache` really needed?
```
fppt: @kasper-piskorski

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/cache/Cache.java
```diff
@@ -153,9 +156,18 @@ public void add(Cache<Q, T> c2){
...
+    //      It's only ever called through LazyQueryCache
+    @SuppressWarnings("unused")
     public abstract Pair<Stream<Answer>, MultiUnifier> getAnswerStreamWithUnifier(Q query);
+    //TODO: @Kasper Is this abstract method on `Cache` really needed?
```
fppt: @kasper-piskorski

grakn-core/src/main/java/ai/grakn/graql/GetQuery.java
aelred: Just a getter. Arguably part of the API, but I'm not sure either way. Maybe tag as deprecated?

grakn-core/src/main/java/ai/grakn/graql/analytics/CentralityQueryBuilder.java
```diff
@@ -37,8 +37,11 @@
...
+    // I remove it and it's implementation and everything still works
+    //@SuppressWarnings("unused")
+    //@CheckReturnValue
+    //CentralityQueryBuilder withTx(GraknTx tx);
```
duckofyork: The idea was, the user can build a graql query without a tx at the beginning, as long as a tx is provided before execution. So this method is everywhere .

grakn-core/src/main/java/ai/grakn/graql/GetQuery.java
fppt: Can't we just kill it? I doubt it's being used externally?

grakn-bootup/src/main/java/ai/grakn/bootup/ProcessHandler.java
```diff
@@ -19,20 +19,29 @@
 package ai.grakn.bootup;
 
 /**
+ * The methods of this interface are used to define a common expectation of what processes should offer.
+ * The actual {@link ProcessHandler} interface is never directly accessed.
+ * Therefore the need for the suppressions below
```
aelred: I'd rather remove unused interfaces like this - but I think this is a personal opinion so keeping it is all good.

grakn-client/src/main/java/ai/grakn/client/BatchExecutorClient.java
```diff
@@ -178,7 +180,7 @@ public static Builder newBuilder() {
         return new Builder();
     }
 
-    //TODO: Remove this method used only by docs tests
+    @API
```
aelred: I'd deprecate this (or keep the TODO), since we wanted to remove it but it's used by the doc tests.

grakn-core/src/main/java/ai/grakn/GraknTx.java
```diff
@@ -320,6 +320,7 @@
      * @return true if the current transaction is read only
      */
     @CheckReturnValue
+    @API
     boolean isReadOnly();
```
aelred: Since we have an opportunity to define our API now, how about a method that returns the `GraknTxType` instead of this more-specific one?

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/spanningtree/ExclusiveEdge.java
```diff
@@ -44,6 +44,7 @@ private ExclusiveEdge(DirectedEdge<V> edge, List<DirectedEdge<V>> excluded, doub
...
     public static <T> ExclusiveEdge<T> of(DirectedEdge<T> edge, double weight) {
         return ExclusiveEdge.of(edge, ImmutableList.of(), weight);
     }
```
duckofyork: You are right! It's no longer used.

grakn-core/src/main/java/ai/grakn/kb/admin/GraknAdmin.java
```diff
@@ -206,6 +207,7 @@
     /**
      * Closes the root session this graph stems from. This will automatically rollback any pending transactions.
      */
+    @API
     void closeSession();
```
aelred: No one is using this and it's on an admin interface! Sounds like killing time.

grakn-engine/src/main/java/ai/grakn/engine/controller/response/Keyspace.java
```diff
@@ -37,45 +38,53 @@
 @AutoValue
 public abstract class Keyspace implements Jacksonisable {
 
+    @API
```
aelred: Kind of neat how we are accidentally marking what parts of our REST API are API.

grakn-graql-shell/src/main/java/ai/grakn/graql/JsonSession.java
```diff
@@ -145,6 +145,7 @@ void sendJson(Json json) throws WebSocketException, IOException {
     }
 
     @OnWebSocketClose
+    @SuppressWarnings("unused") //Used through socket connection.
```
aelred: IntelliJ magic: you can actually tell IntelliJ that `@OnWebsocketClose` marks an entry point if you want instead.

grakn-graql-shell/src/main/java/ai/grakn/graql/internal/shell/animalia/chordata/mammalia/artiodactyla/hippopotamidae/Hippopotamus.java
```diff
@@ -35,5 +35,6 @@
 
     void submerge();
 
+    @SuppressWarnings("unused")
     boolean isHungryHungry();
```
aelred: `@API` for sure

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
```diff
@@ -539,32 +539,27 @@ else if(compatibleTypesFromRoles.isEmpty() || types.isEmpty()){
      * {@link EntityType}s only play the explicitly defined {@link Role}s (not the relevant part of the hierarchy of the specified {@link Role}) and the {@link Role} inherited from parent
      * @return list of {@link RelationshipType}s this atom can have ordered by the number of compatible {@link Role}s
      */
-    public ImmutableList<Type> inferPossibleTypes(Answer sub) {
+    public ImmutableList<Type> inferPossibleTypes() {
```
aelred: I tried to remove this and stuff started breaking, so @kasper-piskorski suggested I roll it back. But apparently it's working for you soooo

grakn-kb/src/main/java/ai/grakn/kb/internal/GraknTxAbstract.java
```diff
@@ -699,7 +699,6 @@ protected void clearGraph() {
         getTinkerPopGraph().traversal().V().drop().iterate();
     }
 
-    @Override
```
aelred: This method is still on the interface, so this should stay.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractCentralityQuery.java
```diff
@@ -62,6 +62,8 @@ void initSubGraph() { //TODO: REMOVE THIS METHOD
...
+    @SuppressWarnings("unused")
     public V of(String... ofTypeLabels) {
         return of(Arrays.stream(ofTypeLabels).map(Label::of).collect(Collectors.toSet()));
     }
```
duckofyork: It's part of the public API. The idea was to have two methods for specifying types (subgraph), one takes string, one takes label. I can remove one of the two if you want.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractStatisticsQuery.java
```diff
@@ -45,12 +45,16 @@
...
+    @SuppressWarnings("unused")
     public V of(Collection<Label> statisticsResourceLabels) {
         this.statisticsResourceLabels = Sets.newHashSet(statisticsResourceLabels);
         return (V) this;
```
duckofyork: Ditto

grakn-core/src/main/java/ai/grakn/graql/GetQuery.java
aelred: I'm good with that! We toe the line of "breaking change", but you're right it's not likely to be used. Plus, we haven't defined our API yet :P

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/cache/Cache.java
kasper-piskorski: @aelred had the same questions in his cleanup PR :D

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
kasper-piskorski: Well in general things shouldn't break as this deals with type inference. It's just that less will be inferred. I will double check. Probably on the finance dataset.

fppt: `Answer sub` was never used anywhere in the chain so that's why I removed it and then when you do that the whole method looks alot like the existing one.

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerQueryImpl.java
```diff
@@ -360,17 +360,6 @@ public GetQuery getQuery() {
...
-    public ImmutableMap<Var, Type> getVarTypeMap(Answer sub) {
-        return ImmutableMap.copyOf(getVarTypeMap(
-                Stream.concat(
-                        getAtoms(IsaAtom.class),
-                        inferEntityTypes()
-                )
-                )
-        );
-    }
-
```
kasper-piskorski: ok it probably should be `inferEntityTypes(sub)` here so that the sub is actually used in all the infer* methods

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/cache/Cache.java
kasper-piskorski: these are all implemented in `QueryCache` and `LazyQueryCache` so wanted to have a common parent

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
```diff
@@ -539,32 +539,27 @@ else if(compatibleTypesFromRoles.isEmpty() || types.isEmpty()){
...
                     .filter(at -> !Sets.intersection(at.getVarNames(), untypedRoleplayers).isEmpty())
                     .collect(toSet());
 
             ImmutableList.Builder<Type> builder = ImmutableList.builder();
-            //prioritise relations with higher chance of yielding answers
             compatibleConfigurations.asMap().entrySet().stream()
-                    //prioritise relations with more allowed roles
-                    .sorted(Comparator.comparing(e -> -e.getValue().size()))
-                    //prioritise relations with number of roles equal to arity
-                    .sorted(Comparator.comparing(e -> e.getKey().relates().count() != getRelationPlayers().size()))
```
kasper-piskorski: why is this removed? i know that intelliJ doesn't like it but I don't get why

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/plan/SimplePlanner.java
```diff
@@ -114,16 +110,4 @@
...
-     * @param query for which the plan should be constructed
-     * @return list of atoms in order they should be resolved using {@link GraqlTraversal}.
-     */
-    static public ImmutableList<Atom> plan(ReasonerQueryImpl query){
-        return ImmutableList.<Atom>builder().addAll(
-                query.selectAtoms().stream()
-                        .sorted(Comparator.comparing(at -> -at.baseResolutionPriority()))
-                        .iterator())
-                .build();
-    }
```
kasper-piskorski: this our old planner, it is not used in the code however it is useful for debugging, it uses the priority function from `Atom`

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/plan/GraqlTraversalPlanner.java
```diff
@@ -55,17 +55,6 @@
...
-     * @return list of atoms in order they should be resolved using {@link GraqlTraversal}.
-     */
-    public static ImmutableList<Atom> plan(ReasonerQueryImpl query){
-        List<Atom> atoms = query.getAtoms(Atom.class)
-                .filter(Atomic::isSelectable)
-                .collect(Collectors.toList());
-        return planFromTraversal(atoms, query.getPattern(), query.tx());
-    }
-
     /**
```
kasper-piskorski: also simpler plan version, not used but potentially useful when debugging

grakn-core/src/main/java/ai/grakn/graql/admin/Atomic.java
kasper-piskorski: can remove

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
fppt: IntelliJ said it was redundant and did not affect processing. Apparently java is actually smart enough to optimise these cases by itself. So these lines were (as far as I understand it) useless.

aelred: IntelliJ probably thinks the sorts aren't stable, so `stream.sorted(byFoo).sorted(byBar)` is the same as `stream.sorted(byBar)`.

@kasper-piskorski if you want to sort by Foo first, then by Bar second try this:

```
stream.sorted(byFoo.thenComparing(byBar))
```

I don't know if it will make a difference, but it won't make things worse and at least IntelliJ will stop complaining 😛

grakn-core/src/main/java/ai/grakn/graql/analytics/CentralityQueryBuilder.java
fppt: @duckofyork is the idea still valid though? I am not sure if we sitll aiming to provide such on the API?

grakn-bootup/src/main/java/ai/grakn/bootup/ProcessHandler.java
fppt: Killing it.

grakn-client/src/main/java/ai/grakn/client/BatchExecutorClient.java
fppt: This is actually the way we want users to interface with the BEC programatically. So it's actual public API . . . according to the docs at least.

grakn-core/src/main/java/ai/grakn/GraknTx.java
fppt: Good catch! Will do sir!

grakn-graql-shell/src/main/java/ai/grakn/graql/internal/shell/animalia/chordata/mammalia/artiodactyla/hippopotamidae/Hippopotamus.java
fppt: Wow you right. Horrible mistake on my side 😄

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/cache/Cache.java
fppt: If they not accessed through the abstraction then it's an unneeded inheritance. So I am going to kill it. Just shout if you totally disagree.

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerQueryImpl.java
fppt: But it wasn't in use though? I can add it back but nothing used it.

kasper-piskorski: `getVarTypeMap(sub`) is referenced in the infer functions in `RelationshipAtom` and others

duckofyork:
It is still valid, as we do need the method for building a query without a tx. Though we could have something simpler: `query.execute(tx)` or `tx.execute(query)`.

I think we should keep it for now. Felix and I can clean this up in the future.

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerQueryImpl.java
fppt: Yes I saw that, but `sub` was never used so I removed the parameter. At that stage I noticed that `public ImmutableMap<Var, Type> getVarTypeMap()` and `public ImmutableMap<Var, Type> getVarTypeMap(Answer sub) ` pretty much did the same thing.

So now everything calls `getVarTypeMap()` and it all still appears to be working. 

How are these methods different from each other that it's worth keeping both?

fppt:
@aelred @duckofyork @kasper-piskorski Could I get a re-review guys?

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerQueryImpl.java
kasper-piskorski: the one with sub infers types from instance ids

grakn-client/src/main/java/ai/grakn/client/BatchExecutorClient.java
```diff
@@ -120,19 +121,19 @@ private BatchExecutorClient(Builder builder) {
     /**
      * Will block until there is space for the query to be submitted
      */
-    public Observable<QueryResponse> add(Query<?> query, Keyspace keyspace) {
+    public Observable add(Query<?> query, Keyspace keyspace) {
```
aelred: You missed a spot with your raw type fixing! This one is important because it's the entry-point to using the `BatchExecutorClient`.

If we don't want to expose the `QueryResponse` type (since it is useless), we can make this return `Observable<?>`.

grakn-client/src/main/java/ai/grakn/client/QueryResponse.java
```diff
@@ -15,42 +15,19 @@
...
-    }
-
-    public Json getJsonResponse() {
-        return jsonResponse;
-    }
-
-    public Query<?> getQuery() {
-        return query;
-    }
+    public static final QueryResponse INSTANCE = new QueryResponse();
```
aelred: Yay! It's the unit type :D

grakn-test/test-integration/src/test/java/ai/grakn/test/graql/shell/GraqlShellIT.java
```diff
@@ -644,7 +644,7 @@ private void assertShellMatches(List<String> arguments, Object... matchers) thro
 
         ImmutableSet<String> noPromptArgs = ImmutableSet.of("-e", "-f", "-b", "-v", "-h");
         if (Sets.intersection(Sets.newHashSet(arguments), noPromptArgs).isEmpty()) {
-            // Remove first four lines containing license and last line containing prompt
+            // Remove first five lines containing license and last line containing prompt
             outputLines = outputLines.subList(4, outputLines.size() - 1);
```
aelred: But it says 4! Who is the liar??

fppt: Debugging some stupid stuff will fix

grakn-core/src/main/java/ai/grakn/graql/analytics/CorenessQuery.java
```diff
@@ -50,13 +50,6 @@
...
     CorenessQuery of(String... ofTypeLabels);
 
-    /**
-     * @param ofLabels a collection of types in the subgraph to compute coreness of. By default the coreness of all
-     *                 entities and attributes are computed.
-     * @return a CorenessQuery with the subTypeLabels set
-     */
-    CorenessQuery of(Collection<Label> ofLabels);
-
     /**
```
duckofyork: I think we should keep this method

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
kasper-piskorski: From my understanding the sorting here is stable. I have just tested with markit. Removing the extra sorting makes some of the query execution times go through the roof xD

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/atom/binary/RelationshipAtom.java
```diff
@@ -574,7 +569,6 @@ else if(compatibleTypesFromRoles.isEmpty() || types.isEmpty()){
                                 rs.stream().flatMap(Role::playedByTypes).filter(typesFromNeighbour::contains).count()
                         );
                     })
-                    .sorted(Comparator.comparing(p -> -p.getValue()))
```
kasper-piskorski: this needs to stay as well

PR #2611, opened 2018-01-31
New HttpController interfaces fed into HttpHandler

marco-scoppetta:
# Why is this PR needed?

So that KBMS does not need to extend any GRAKN open-source classes and we avoid duplicate code and high coupling.

# What does the PR do?

Makes `HttpHandler` accept a list of `HttpController` (new interface) upon construction and executes the `start()` method on each of them when `startHTTP()` method is invoked.

Also removed deprecated constructor in `GraknCreator` for we have replaced inheritance with delegation in KBMS.

No.

Not for now

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/GraknCreator.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/GraknCreator.java
@@ -19,6 +19,7 @@
 package ai.grakn.engine;
 
 import ai.grakn.GraknConfigKey;
+import ai.grakn.engine.controller.HttpController;
 import ai.grakn.engine.data.RedisWrapper;
 import ai.grakn.engine.factory.EngineGraknTxFactory;
 import ai.grakn.engine.lock.JedisLockProvider;
@@ -38,6 +39,8 @@ import redis.clients.util.Pool;
 import spark.Service;
 
 import java.io.IOException;
+import java.util.Collection;
+import java.util.Collections;
 
 /**
  * Static configurator for classes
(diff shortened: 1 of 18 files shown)


grakn-engine/src/main/java/ai/grakn/engine/GraknCreator.java
```diff
@@ -46,29 +49,16 @@
...
-    protected final EngineID engineID;
-    protected final Service sparkService;
-    protected final GraknEngineStatus graknEngineStatus;
-    protected final MetricRegistry metricRegistry;
-    protected final GraknConfig graknEngineConfig;
+    private final EngineID engineID;
+    private final Service sparkService;
+    private final GraknEngineStatus graknEngineStatus;
+    private final MetricRegistry metricRegistry;
+    private final GraknConfig graknEngineConfig;
```
aelred: 😄

grakn-engine/src/main/java/ai/grakn/engine/HttpHandler.java
```diff
@@ -91,11 +95,13 @@ protected void startCollaborators() {
...
-        // Start all the controllers
-        new GraqlController(factory, spark, postProcessor, printer, metricRegistry);
-        new ConceptController(factory, spark, metricRegistry);
-        new SystemController(spark, prop, factory.systemKeyspace(), graknEngineStatus, metricRegistry);
-        new CommitLogController(spark, postProcessor);
+        // Start all the DEFAULT controllers
+        new GraqlController(factory, postProcessor, printer, metricRegistry).start(spark);
+        new ConceptController(factory, metricRegistry).start(spark);
+        new SystemController(prop, factory.systemKeyspace(), graknEngineStatus, metricRegistry).start(spark);
+        new CommitLogController(postProcessor).start(spark);
```
aelred: Maybe we should pass them in as collaborators? I guess it doesn't make much difference right now.

marco-scoppetta: I thought the same, but I figured this might be a discussion for the future, since for now we always want them

grakn-engine/src/main/java/ai/grakn/engine/GraknCreator.java
fppt: 😄

grakn-engine/src/main/java/ai/grakn/engine/HttpHandler.java
fppt: Ya I kind of like the ides of passing in the defaults as well.

grakn-engine/src/main/java/ai/grakn/engine/controller/HttpController.java
```diff
@@ -0,0 +1,32 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with Grakn. If not, see <http://www.gnu.org/licenses/gpl.txt>.
+ */
+
+package ai.grakn.engine.controller;
+
+import spark.Service;
+
+/**
+ *  Interface implemented by all HTTPControllers
```
fppt: Slight doc poke. I think you should say _why_ this interface is needed. Something like:

`This allows a common manager {@link HttpHandler} to control all controllers in a uniform and consistent manner`

PR #2613, opened 2018-02-05
Fix New Post Processing

fppt:
# Why is this PR needed?

The new PP works but only executes one job per keyspace per second. This means that graph with 10000 attributes will take 10000 seconds to completely post process. That is just dumb. 

# What does the PR do?

Submit multiple jobs for post processing per second rather than just one job.

No. 

Smart job submission. Right now it submits however many jobs there are threads. Ideally it should submit jobs based on load.

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/task/postprocessing/PostProcessingTask.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/task/postprocessing/PostProcessingTask.java
@@ -60,10 +60,12 @@ public class PostProcessingTask implements BackgroundTask{
     public void run() {
         Set<Keyspace> kespaces = factory.systemKeyspace().keyspaces();
         kespaces.forEach(keyspace -> {
-            String index = indexPostProcessor.popIndex(keyspace);
-            if(index != null) {
-                threadPool.schedule(() -> processIndex(keyspace, index), postprocessingDelay, TimeUnit.SECONDS);
-            }
+            String index;
+            do{
+                 index = indexPostProcessor.popIndex(keyspace);
+                 final String i = index;
+                 if(index != null) threadPool.schedule(() -> processIndex(keyspace, i), postprocessingDelay, TimeUnit.SECONDS);
+            } while(index != null);
         });
     }
 
--- a/grakn-engine/src/main/java/ai/grakn/engine/task/postprocessing/PostProcessingTask.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/task/postprocessing/PostProcessingTask.java
@@ -47,12 +47,14 @@ public class PostProcessingTask implements BackgroundTask{
     private final EngineGraknTxFactory factory;
     private final IndexPostProcessor indexPostProcessor;
     private final ScheduledExecutorService threadPool;
+    private final int postProcessingMaxJobs;
     private final int postprocessingDelay;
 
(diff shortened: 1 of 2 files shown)


grakn-engine/src/test/java/ai/grakn/engine/task/postprocessing/PostProcessingTaskTest.java
```diff
@@ -94,4 +94,34 @@ public void whenTheIndexCacheIsEmpty_NothingStarts(){
...
+        when(indexPostProcessor.popIds(keyspaceA, index1)).thenReturn(ids);
+        when(indexPostProcessor.popIds(keyspaceA, index2)).thenReturn(ids);
+        when(indexPostProcessor.popIds(keyspaceA, index3)).thenReturn(ids);
+        when(indexPostProcessor.popIds(keyspaceA, index4)).thenReturn(ids);
+
+        //Run the method
+        postProcessingTask.run();
+
+        //Give time for PP to run
+        Thread.sleep(config.getProperty(GraknConfigKey.POST_PROCESSOR_DELAY) * 2000);
```
aelred: Why 2000?

grakn-engine/src/main/java/ai/grakn/engine/task/postprocessing/PostProcessingTask.java
```diff
@@ -47,23 +47,29 @@
...
-                threadPool.schedule(() -> processIndex(keyspace, index), postprocessingDelay, TimeUnit.SECONDS);
-            }
+            String index;
+            int limit = 0;
+            do{
+                 index = indexPostProcessor.popIndex(keyspace);
+                 final String i = index;
+                 if(index != null) threadPool.schedule(() -> processIndex(keyspace, i), postprocessingDelay, TimeUnit.SECONDS);
+                 limit++;
+            } while(index != null && limit < postProcessingMaxJobs);
```
aelred: Look at beautiful Java 9 possibilities:
```
Stream.generate(() -> indexPostProcessor.popIndex(keyspace))
   .takeWhile(index -> index != null)
   .limit(postProcessingMaxJobs)
   .forEach(index -> threadPool.schedule(() -> processIndex(keyspace, i), ppDelay, SECONDS));
```

grakn-engine/src/test/java/ai/grakn/engine/task/postprocessing/PostProcessingTaskTest.java
lolski: i think this is the part where we wait for cassandra index to be ready?

aelred: I'm just wondering about the exact number - sleeping in tests is a bit dangerous, so we should make sure we're sleeping for the right length.

grakn-engine/src/main/java/ai/grakn/engine/task/postprocessing/PostProcessingTask.java
fppt: Oooooooooooo . . .  pretty

grakn-engine/src/test/java/ai/grakn/engine/task/postprocessing/PostProcessingTaskTest.java
fppt: Ya 2 seconds is just to give time for the index to stabilise. I agree that sleeping is dangerous but we have no other way to check the status of the index right now.

PR #2614, opened 2018-02-05
[Stable] Bug #18869: Refine resolution plan to ensure more optimal substitution use

kasper-piskorski:
**NB: This refinement is already present on master. Pushing to stable as it makes example queries significantly more performant.**

# Why is this PR needed?
There is a range of queries for which highly suboptimal plans can possibly be constructed.

# What does the PR do?
Introduces resolution plan refinements to ensure most subbed atoms are favoured.

No.

Incorporate key enforcement for resolution tree pruning.


fppt:
@kasper-piskorski after this PR: https://github.com/graknlabs/grakn/pull/2615 I will be merging master into stable. Does that make this PR redundant? 

I would rather sort out the differences between the branches via merges than via separate PRs

PR #2615, opened 2018-02-05
Merge stable into master

fppt:
# Why is this PR needed?

To sort out branch histories

# What does the PR do?

Merge stable into master

No

N/A


grakn-client/src/main/java/ai/grakn/client/Client.java
```diff
@@ -18,103 +18,24 @@
...
-        String host = properties.getProperty(GraknConfigKey.SERVER_HOST_NAME.name());
-        int port = Integer.parseInt(properties.getProperty(GraknConfigKey.SERVER_PORT.name()));
-        if (serverIsRunning(new SimpleURI(host, port))) {
-            LOG.info("Server " + host + ":" + port + " is running");
-            return EngineStatus.Running;
-        } else {
-            LOG.info("Server " + host + ":" + port + " is not running");
-            return EngineStatus.NotRunning;
-        }
-    }
```
aelred: For some reason most of this stuff has not been removed in the final diff :(

fppt: @aelred wait more of this needs to die?

aelred: If you look at the changes from all commits, `Client#main` is still there for example. But I removed it in my PR...

fppt:
@aelred re-review please?

PR #2616, opened 2018-02-06
Refactor queries to expose all their fields

aelred:
# Why is this PR needed?

This is primarily motivated by the gRPC client. That will require separating query execution from query representation: so the query classes must become "dumb" value classes and query execution must happen separately. This requires a few things:

- The query classes must expose all their fields (in a safe, read-only manner)
- The query execute methods should be internally cohesive and decoupled from the query representation (so they can be moved easily).

# What does the PR do?

As well as adding a bunch of getters, I also introduce some new interfaces and abstract classes, plus a clean up of the analytics queries to reduce the mutation going on internally. This makes it very easy to move the `execute` methods into another class (in fact I've already done this on a private branch parallel to this).

No.

Introduce a `QueryRunner` class that will be responsible for query execution - then finally a gRPC client!


grakn-core/src/main/java/ai/grakn/graql/ComputeQuery.java
```diff
@@ -61,6 +64,8 @@
     @CheckReturnValue
     ComputeQuery<T> includeAttribute();
 
+    boolean getIncludeAttribute();
```
fppt: Nitpick. Maybe `isAttributeIncluded()` ?

grakn-core/src/main/java/ai/grakn/graql/ComputeQueryOf.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package ai.grakn.graql;
+
+import ai.grakn.concept.Label;
+
+import java.util.Collection;
+
+/**
+ * yes this is a bad name TODO
```
fppt: Todo for now or later?

grakn-core/src/main/java/ai/grakn/graql/Query.java
```diff
@@ -67,4 +66,9 @@
      */
     @CheckReturnValue
     boolean isReadOnly();
+
+    /**
+     * Get the transaction associated with this query
+     */
+    Optional<? extends GraknTx> tx();
```
fppt: Another thing I would like to see die. Either the transactions should be embedded completely or not. This middleground is just silly.

grakn-core/src/main/java/ai/grakn/graql/analytics/ClusterQuery.java
```diff
@@ -43,6 +45,10 @@
      */
     ClusterQuery<Map<String, Set<String>>> members();
 
+    boolean membersSet();
```
fppt: `memberSet` returns a boolean?? Java docs explaining what this means.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/AggregateQueryImpl.java
```diff
@@ -63,25 +48,12 @@ public boolean isReadOnly() {
...
 
     @Override
-    public String toString() {
-        return match.toString() + " aggregate " + aggregate.toString() + ";";
+    public final Optional<GraknTx> tx() {
+        return match().admin().tx();
     }
 
     @Override
-    public boolean equals(Object o) {
```
fppt: This makes me feel warm and fuzzy

grakn-graql/src/main/java/ai/grakn/graql/internal/query/DefineQueryImpl.java
```diff
@@ -56,18 +55,13 @@ static DefineQueryImpl of(ImmutableList<VarPatternAdmin> varPatterns, @Nullable
 
     @Override
     public Answer execute() {
-        GraknTx tx = tx();
-        if (tx == null) throw GraqlQueryException.noTx();
+        GraknTx tx = tx().orElseThrow(GraqlQueryException::noTx);
```
fppt: AHHHH!!! So sick of seeing this handling of optional Tx

grakn-core/src/main/java/ai/grakn/graql/analytics/ClusterQuery.java
duckofyork: I think it indicates whether the query return members of each cluster or just the size of the cluster. Though I think maybe we should delete the method for returning size at some point? what do you guys think?

grakn-core/src/main/java/ai/grakn/graql/ComputeQueryOf.java
```diff
@@ -0,0 +1,47 @@
...
+     * @return a ComputeQuery with the subTypeLabels set
+     */
+    ComputeQueryOf<T> of(String... resourceTypeLabels);
+
+    /**
+     * @param resourceLabels a collection of types of resources to execute the query on
+     * @return a ComputeQuery with the subTypeLabels set
+     */
+    ComputeQueryOf<T> of(Collection<Label> resourceLabels);
+
```
duckofyork: In statistics query, ofLabels must be specified, and the labels must be attribute types. In other queries, such as degree and coreness, ofLables are optional, and they can be any instance type. Maybe we should separate them? It can be difficult to explain what ofLabels are here.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/AggregateQueryImpl.java
duckofyork: 👍

grakn-core/src/main/java/ai/grakn/graql/ComputeQuery.java
aelred: ✅

grakn-core/src/main/java/ai/grakn/graql/ComputeQueryOf.java
aelred: For now I think. Any ideas? @duckofyork you know more precisely what `of` is meant to mean.

grakn-core/src/main/java/ai/grakn/graql/Query.java
aelred: Agreed!

grakn-core/src/main/java/ai/grakn/graql/analytics/ClusterQuery.java
aelred: Renamed to `isMembersSet` and added Javadoc ✅

aelred: @duckofyork does that mean it would only ever return cluster members? Yeah, I agree. Just size seems weird.

duckofyork: The idea was when the graph is huge, the user would have to use java api, so by default graql console returns size of each cluster. I will submit another pr after this one got merged.

grakn-core/src/main/java/ai/grakn/graql/ComputeQueryOf.java
aelred: Yeah, that makes sense! Ideas for names:

- `ComputeQueryOf` -> `StatisticsQuery`
- `StatisticsQuery#ofLabels` -> `attributeLabels`
- `DegreeQuery#ofLabels` -> `targetLabels`
- `CorenessQuery#ofLabels` -> `targetLabels`

duckofyork: 👍

aelred:
@fppt and @duckofyork now it is perfect 👍

PR #2619, opened 2018-02-06
Cleanup reasoner planning and atom priorities

kasper-piskorski:
# Why is this PR needed?
Cause getting rid of unused code helps maintainability and readability.

# What does the PR do?
Removes priorities from atoms and restores the simple plan for debugging as it was useless in the previous form.

No.

Don't envisage any.


grakn-graql/src/test/java/ai/grakn/graql/internal/reasoner/plan/priority/ResourcePriority.java
```diff
@@ -0,0 +1,102 @@
...
+                    } //variable inequality
+                    else {
+                        vpsPriority += SimplePlanner.COMPARISON_VARIABLE_VALUE_PREDICATE;
+                    }
+                } else {
+                    vpsPriority += SimplePlanner.NON_SPECIFIC_VALUE_PREDICATE;
+                }
+            }
+            //normalise
+            vpsPriority = vpsPriority/vps.size();
```
duckofyork: Is it accurate enough to use integer here? I'm asking because I don't really know how big this number is.

PR #2621, opened 2018-02-07
Bug 19309: Get redis to stop when not bound to localhost 

fppt:
# Why is this PR needed?

Without this PR Redis does not stop when not bound to `localhost`

# What does the PR do?

Reads the Redis IP from the config and uses it in killing the redis instance

No

Not sure.

### DIFF
--- a/grakn-bootup/src/main/java/ai/grakn/bootup/QueueProcess.java
+++ b/grakn-bootup/src/main/java/ai/grakn/bootup/QueueProcess.java
@@ -32,7 +32,7 @@ import java.time.LocalDateTime;
  */
 public class QueueProcess extends AbstractProcessHandler {
     private static final String QUEUE_PROCESS_NAME = "redis-server";
-    private static String CONFIG_LOCATION = "/services/redis/redis.conf";
+    private static final String CONFIG_LOCATION = "/services/redis/redis.conf";
     private static final Path QUEUE_PID = Paths.get(File.separator,"tmp","grakn-queue.pid");
     private static final long QUEUE_STARTUP_TIMEOUT_S = 10;
     private static final String NAME = "Queue";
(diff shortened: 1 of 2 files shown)


grakn-bootup/src/main/java/ai/grakn/bootup/QueueProcess.java
```diff
@@ -99,16 +104,29 @@ private void queueStopProcess() {
...
                 "/bin/sh",
                 "-c",
-                homePath + "/services/redis/" + queueBin + " shutdown"
+                homePath + "/services/redis/" + queueBin + " -h " + host + " shutdown"
         }, null, null);
 
         waitUntilStopped(QUEUE_PID,pid);
     }
 
+    private String getHostFromConfig() {
```
lolski: can we read it using `ai.grakn.engine.GraknConfig` instead?

fppt:
@lolski re-review please

grakn-bootup/src/main/java/ai/grakn/bootup/QueueProcess.java
```diff
@@ -29,6 +32,7 @@
  */
 public class QueueProcess extends AbstractProcessHandler {
     private static final String QUEUE_PROCESS_NAME = "redis-server";
+    private static String CONFIG_LOCATION = "/services/redis/redis.conf";
```
lolski: nitpick: let's mark it as final

grakn-bootup/src/main/java/ai/grakn/bootup/QueueProcess.java
```diff
@@ -99,16 +103,22 @@ private void queueStopProcess() {
...
                 "-c",
-                homePath + "/services/redis/" + queueBin + " shutdown"
+                homePath + "/services/redis/" + queueBin + " -h " + host + " shutdown"
         }, null, null);
 
         waitUntilStopped(QUEUE_PID,pid);
     }
 
+    private String getHostFromConfig() {
+        String fileLocation = homePath + CONFIG_LOCATION;
```
lolski: Nitpick #2: Can we do `Paths.get(homePath, CONFIG_LOCATION)`?

fppt:
@lolski The nits have been picked

PR #2622, opened 2018-02-07
Add `QueryRunner` interface for running queries

aelred:
# Why is this PR needed?

The gRPC client will need to provide an alternative way to run queries. This will be done using a new implementation of `QueryRunner` in this PR.

# What does the PR do?

Introduce a `QueryRunner` interface, plus an implementation `TinkerQueryRunner`, that executes queries using Tinkerpop.

...maybe. I removed `ComputeQuery#isStatisticsQuery` but since we haven't got a defined API yet 🤷‍♂️ 

There is still a _lot_ of Tinkerpop-specific code in `grakn-core` and `grakn-graql`. Eventually we will try and isolate this to just `grakn-kb`.


grakn-core/src/main/java/ai/grakn/ComputeJob.java
```diff
@@ -0,0 +1,41 @@
...
+ * @author Felix Chapman
+ *
+ * @param <T> The returned result of the compute job
+ */
+public interface ComputeJob<T> {
+
+    /**
+     * Get the result of the compute query job
+     *
+     * @throws something if the job has been killed
```
fppt: What? Vague javadocs are vague

grakn-core/src/main/java/ai/grakn/QueryRunner.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Stream;
+
+/**
+ * @author Felix Chapman
+ */
+public interface QueryRunner {
```
fppt: I think you might want to update the java docs to say why we need this interface. I.e. So we can decouple the queries and their execution,

grakn-core/src/main/java/ai/grakn/graql/ComputeQuery.java
```diff
@@ -51,7 +51,7 @@
      * @return a ComputeQuery with the subLabels set
      */
     @CheckReturnValue
-    ComputeQuery<T> in(Collection<Label> subLabels);
+    ComputeQuery<T> in(Collection<? extends Label> subLabels);
```
fppt: Why does this need the wildcard now?

grakn-graql/src/main/java/ai/grakn/graql/internal/query/DefineQueryImpl.java
```diff
@@ -55,13 +50,7 @@ static DefineQueryImpl of(Collection<? extends VarPattern> varPatterns, @Nullabl
 
     @Override
     public final Answer execute() {
-        GraknTx tx = tx().orElseThrow(GraqlQueryException::noTx);
```
fppt: Nice ! 👍

grakn-graql/src/main/java/ai/grakn/graql/internal/query/DeleteQueryImpl.java
```diff
@@ -52,9 +46,8 @@ static DeleteQueryImpl of(Collection<? extends Var> vars, Match match) {
     }
 
     @Override
-    public Void execute() {
-        List<Answer> results = match().stream().collect(toList());
-        results.forEach(this::deleteResult);
+    public final Void execute() {
+        queryRunner().run(this);
```
fppt: Out of curiosity is there no way we can make this a default method? Most of the `executes` seem to do this.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractComputeQuery.java
```diff
@@ -73,24 +56,12 @@
 
     @Override
     public final T execute() {
-        GraknTx tx = tx().orElseThrow(GraqlQueryException::noTx);
```
fppt: Just checking TODO now or later? and TODO what?

grakn-graql/src/main/java/ai/grakn/graql/internal/query/runner/TinkerComputeQueryRunner.java
```diff
@@ -0,0 +1,683 @@
...
+            for (Label type : subLabels(query)) {
+                Boolean patternExist = tx.graql().infer(false).match(
+                        Graql.var("x").has(resourceType, Graql.var()),
+                        Graql.var("x").isa(Graql.label(type))
+                ).iterator().hasNext();
+                if (patternExist) return true;
+            }
+        }
+        return false;
+        //TODO: should use the following ask query when ask query is even lazier
```
fppt: TODO now or later?

grakn-graql/src/main/java/ai/grakn/graql/internal/query/runner/TinkerComputeQueryRunner.java
```diff
@@ -0,0 +1,683 @@
...
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ *
+ *
+ * @author Felix Chapman
+ */
+public class TinkerComputeQueryRunner {
```
fppt: Big class is big, maybe we can decompose this a bit? Also javadocs could be better.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/runner/TinkerQueryRunner.java
```diff
@@ -0,0 +1,206 @@
...
+
+import static ai.grakn.util.CommonUtil.toImmutableList;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * A {@link QueryRunner} that runs queries using a Tinkerpop graph.
+ *
+ * @author Felix Chapman
+ */
+@SuppressWarnings("unused") // accessed via reflection in GraknTxAbstract
```
fppt: 👍

grakn-kb/src/main/java/ai/grakn/kb/internal/GraknTxAbstract.java
```diff
@@ -109,6 +111,7 @@
 public abstract class GraknTxAbstract<G extends Graph> implements GraknTx, GraknAdmin {
     final Logger LOG = LoggerFactory.getLogger(GraknTxAbstract.class);
     private static final String QUERY_BUILDER_CLASS_NAME = "ai.grakn.graql.internal.query.QueryBuilderImpl";
```
lolski: can we do `QueryBuilderImpl.getClass().getName()` instead

grakn-kb/src/main/java/ai/grakn/kb/internal/GraknTxAbstract.java
```diff
@@ -109,6 +111,7 @@
...
+    private static final String QUERY_RUNNER_CLASS_NAME = "ai.grakn.graql.internal.query.runner.TinkerQueryRunner";
```
lolski: `TinkerQueryRunner.getClass().getName()`. In my opinion they will be more refactor-proof

grakn-kb/src/main/java/ai/grakn/kb/internal/GraknTxAbstract.java
```diff
@@ -126,6 +129,18 @@
         }
     }
 
+    private static final Method queryRunnerFactory;
+
+    static {
```
lolski: can we add an explanation about the intention of this `static` block? is `method == null` an expected, normal condition?

grakn-kb/src/main/java/ai/grakn/kb/internal/GraknTxAbstract.java
```diff
@@ -957,4 +972,18 @@ public void shard(ConceptId conceptId) {
     public long getShardCount(Type concept){
         return TypeImpl.from(concept).shardCount();
     }
+
+    @Override
+    public final QueryRunner queryRunner() {
+        // TODO apologise
+        if (queryRunnerFactory == null) {
+            throw new RuntimeException("The query runner implementation " + QUERY_RUNNER_CLASS_NAME +
```
lolski: does it make sense to centralise the error under `ErrorMessage` class?

grakn-kb/src/main/java/ai/grakn/kb/internal/GraknTxAbstract.java
lolski: does it make sense to just create a private method for this operation:

```
private static final Method queryRunnerFactory = getCreateMethodFromClass(QUERY_RUNNER_CLASS_NAME);

private Method getCreateMethodFromClass(String className) {
        Method method;
        try {
            method = Class.forName(QUERY_RUNNER_CLASS_NAME).getDeclaredMethod("create", GraknTx.class);
        } catch (NoSuchMethodException | SecurityException | ClassNotFoundException e) {
            method = null;
        }
        return method;
}
```

grakn-graql/src/main/java/ai/grakn/graql/internal/query/DeleteQueryImpl.java
lolski: I actually prefer this style, more of a preference thing since i still prefer keeping interfaces from implementations

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractComputeQuery.java
```diff
@@ -73,24 +56,12 @@
...
-        if (graknComputer == null) {
-            graknComputer = tx.session().getGraphComputer();
-        }
-
-        T result = innerExecute(tx, graknComputer);
-
-        LOGGER.info(toString() + " finished in " + (System.currentTimeMillis() - startTime) + " ms");
-
-        return result;
+        assert job == null; // TODO
```
lolski: can we improve the comment i.e. `// TODO: <what we actually need to do here>`

grakn-graql/src/main/java/ai/grakn/graql/internal/query/runner/TinkerComputeJob.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * A compute query job executed against a {@link GraknComputer}.
+ *
+ * @author Felix Chapman
+ */
+class TinkerComputeJob<T> implements ComputeJob<T> {
```
lolski: <3 the functionalism school of thought here

grakn-graql/src/main/java/ai/grakn/graql/internal/query/runner/TinkerComputeQueryRunner.java
```diff
@@ -0,0 +1,683 @@
...
+    }
+
+    private Set<Label> getHasResourceRelationLabels(Set<Type> subTypes) {
+        return subTypes.stream()
+                .filter(Concept::isAttributeType)
+                .map(resourceType -> Schema.ImplicitType.HAS.getLabel(resourceType.getLabel()))
+                .collect(Collectors.toSet());
+    }// If the sub graph contains attributes, we may need to add implicit relations to the paths
+
+    private List<List<Concept>> getExtendedPaths(List<List<Concept>> allPaths) {
```
lolski: the method is big but you seem to manage to split them into several high-level structure here. maybe actually split them into several methods?

grakn-core/src/main/java/ai/grakn/ComputeJob.java
aelred: lol I forgot about that

Also just below:
> Stop the job executing and make something terrible and unspeakable happen to anyone waiting for a result

grakn-core/src/main/java/ai/grakn/QueryRunner.java
aelred: good point!

grakn-core/src/main/java/ai/grakn/graql/ComputeQuery.java
aelred: A fun sequence of reasons:
- I added `ComputeQuery#subLabels`
- I gave `subLabels` the most generic type (`Collection<? extends Label>`) to protect us from breaking changes etc.
- At some point, code goes `query2.in(query1.subLabels())`, so now `in` must also accept the most generic type.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/DeleteQueryImpl.java
aelred: It would be nice to just have:

```
class AbstractQuery<T> {
    public final T execute() {
        queryRunner().run(this);
    }
}
```

Unfortunately, we can't do this, because Java does not know which `run` method to delegate to. In fact, it doesn't even know if the `Query<?>` class _has_ a `run` method on `QueryRunner`! For example, imagine if the user irritatingly decides to make their own implementation of `Query` and goes `query.execute()`.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/analytics/AbstractComputeQuery.java
aelred: Todo now I think. I will need to check with @duckofyork what the expected behaviour is if you run a compute query twice at once.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/runner/TinkerComputeQueryRunner.java
aelred: I agree. I split it a bit, but I'll try to split it further.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/runner/TinkerComputeQueryRunner.java
aelred: this was already here 👍 
so later!

grakn-kb/src/main/java/ai/grakn/kb/internal/GraknTxAbstract.java
aelred: We can't, because `grakn-kb` does not depend on `grakn-graql` :(

Ideally in the future, we can move `QueryBuilderImpl` and `TinkerQueryRunner` into `grakn-kb` and then we won't need this nonsense.

grakn-kb/src/main/java/ai/grakn/kb/internal/GraknTxAbstract.java
aelred: Ditto above.

grakn-kb/src/main/java/ai/grakn/kb/internal/GraknTxAbstract.java
aelred: The private method is a good idea!

grakn-graql/src/main/java/ai/grakn/graql/internal/query/runner/TinkerComputeJob.java
aelred: 👍 
I even gave it a `map` method so it's a [functor](https://en.wikipedia.org/wiki/Functor)!

aelred:
@lolski there aren't any tests with this PR because it's a refactor. So the tests passing is evidence that the refactor has worked 👍

grakn-graql/src/main/java/ai/grakn/graql/internal/query/runner/TinkerComputeQueryRunner.java
aelred: Yeah, I will try and split them more. `TinkerComputeQueryRunner` is biiiig...

aelred:
I've addressed everything @fppt and @lolski 

Let me know if I've made any new bad things :P

PR #2623, opened 2018-02-07
Optimise query planner: infer relationship type from schema

duckofyork:
# Why is this PR needed?
Better plan means better performance
# What does the PR do?
- Infer relationship types from schema
- Avoid starting queries from implicit types when non-implicit types are available
No
None


grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -136,6 +174,123 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, EmbeddedGrakn
...
+        Iterator<Var> iterator = relationshipRolePlayerMap.keySet().iterator();
+        while (iterator.hasNext()) {
+            Var relationship = iterator.next();
+
+            // the relation should have at least 2 known role players so we can infer something useful
+            if (instanceVarTypeMap.containsKey(relationship) ||
+                    relationshipRolePlayerMap.get(relationship).size() < 2) {
+                iterator.remove();
+            } else {
+                final int[] numRolePlayersHaveType = {0};
```
kasper-piskorski: just wondering, what's the benefit of having an array here?

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -136,6 +174,123 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, EmbeddedGrakn
...
+                    // compute the intersection of possible relationship types
+                    possibleRelationShipTypes.retainAll(getAllPossibleRelationShipTypesOfARolePlayer(
+                            nextRolePlayer, instanceVarTypeMap, relationshipMap));
+                }
+            }
+
+            //TODO: if possibleRelationShipTypes here is empty, the query will not match any data
+            if (possibleRelationShipTypes.size() == 1) {
+
+                Type relationshipType = possibleRelationShipTypes.iterator().next();
```
kasper-piskorski: You can use Iterables#getOnlyElement here.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -136,6 +174,123 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, EmbeddedGrakn
...
+                }
+            }
+        }
+        if (relationshipRolePlayerMap.isEmpty()) return;
+
+        // for each type, get all possible relationship type it could be in
+        Map<Type, Set<RelationshipType>> relationshipMap = new HashMap<>();
+        getAllPossibleRelationships(relationshipMap, tx.admin().getMetaEntityType());
+        getAllPossibleRelationships(relationshipMap, tx.admin().getMetaAttributeType());
+        getAllPossibleRelationships(relationshipMap, tx.admin().getMetaRelationType());
```
kasper-piskorski: Isn't doing these potentially expensive for bigger ontologies? Could we call `getAllPossibleRelationships` on a specific type when we need it in `getAllPossibleRelationShipTypesOfARolePlayer`?

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -136,6 +174,123 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, EmbeddedGrakn
...
+                allFragments.addAll(EquivalentFragmentSets.isa(IsaProperty.of(label(label).admin()),
+                        relationshipVar, labelVar, relationshipType.isImplicit()).fragments());
+            }
+        });
+    }
+
+    private static Set<Type> getAllPossibleRelationShipTypesOfARolePlayer(
+            Var rolePlayer, Multimap<Var, Type> instanceVarTypeMap, Map<Type, Set<RelationshipType>> relationshipMap) {
+
+        Iterator<Type> typeIterator = instanceVarTypeMap.get(rolePlayer).iterator();
```
kasper-piskorski: how do we guarantee that a roleplayer has a type? Does having multiple possibilities allow for types being part of the same hierarchy?

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -136,6 +174,123 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, EmbeddedGrakn
...
+                if (instanceVarTypeMap.containsKey(nextRolePlayer)) {
+
+                    // compute the intersection of possible relationship types
+                    possibleRelationShipTypes.retainAll(getAllPossibleRelationShipTypesOfARolePlayer(
+                            nextRolePlayer, instanceVarTypeMap, relationshipMap));
+                }
+            }
+
+            //TODO: if possibleRelationShipTypes here is empty, the query will not match any data
+            if (possibleRelationShipTypes.size() == 1) {
```
kasper-piskorski: What if we end up with have more than 1 type but all types are in sub hierarchy?

grakn-graql/src/test/java/ai/grakn/graql/internal/query/QueryPlannerTest.java
```diff
@@ -91,6 +116,172 @@ public void setUp() {
...
+                var().rel(x).rel(y));
+        plan = getPlan(pattern);
+        assertEquals(3L, plan.stream().filter(LabelFragment.class::isInstance).count());
+        relationship = plan.get(4).start().getValue();
+
+        // should start from role player
+        assertTrue(relationship.equals(x.getValue()) || relationship.equals(y.getValue()));
+        assertTrue(plan.get(5) instanceof OutIsaFragment);
+    }
+
```
kasper-piskorski: I have a potentially useful test set defined in `typeInferenceTest.gql` and unit tests in `TypeInferenceQueryTest`. You might want to have a look.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
duckofyork: so it can be used inside lambda

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
duckofyork: We do the checking before calling this method

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
duckofyork: We do nothing, for now :-)

grakn-graql/src/test/java/ai/grakn/graql/internal/query/QueryPlannerTest.java
duckofyork: 👍

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -136,6 +175,122 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, EmbeddedGrakn
...
+                    firstRolePlayer, instanceVarTypeMap, relationshipMap);
+            while (rolePlayerVarIterator.hasNext()) {
+                Var nextRolePlayer = rolePlayerVarIterator.next();
+                if (instanceVarTypeMap.containsKey(nextRolePlayer)) {
+
+                    // compute the intersection of possible relationship types
+                    possibleRelationShipTypes.retainAll(getAllPossibleRelationShipTypesOfARolePlayer(
+                            nextRolePlayer, instanceVarTypeMap, relationshipMap));
+                }
+            }
```
aelred: Maybe you can use `Sets#intersection` and `reduce`:

```
Set<Type> possibleRelationshipTypes = rolePlayersVars.stream()
    .filter(instanceVarTypeMap::containsKey)
    .map(rolePlayer -> getAllPossibleRelationshipTypesOfARolePlayer(rolePlayer, instanceVarTypeMap, relationshipMap))
    .reduce(Sets::intersection);
```

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -136,6 +175,122 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, EmbeddedGrakn
...
+
+            //TODO: if possibleRelationShipTypes here is empty, the query will not match any data
+            if (possibleRelationShipTypes.size() == 1) {
+
+                Type relationshipType = possibleRelationShipTypes.iterator().next();
+                Label label = relationshipType.getLabel();
+                Var labelVar = var();
+                allFragments.add(Fragments.label(LabelProperty.of(label), labelVar, ImmutableSet.of(label)));
+                allFragments.addAll(EquivalentFragmentSets.isa(IsaProperty.of(label(label).admin()),
+                        relationshipVar, labelVar, relationshipType.isImplicit()).fragments());
```
aelred: Maybe split this line into a few variables, I have trouble reading it.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -136,6 +175,122 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, EmbeddedGrakn
...
+        Iterator<Type> typeIterator = instanceVarTypeMap.get(rolePlayer).iterator();
+        Type firstRolePlayerType = typeIterator.next(); // the role player here has at least one type
+
+        Set<Type> relationshipTypes = new HashSet<>(relationshipMap.get(firstRolePlayerType));
+
+        // we can stop doing intersection when relationshipTypes is empty
+        while (typeIterator.hasNext() && !relationshipTypes.isEmpty()) {
+            Type nextRolePlayerType = typeIterator.next();
+            relationshipTypes.retainAll(relationshipMap.get(nextRolePlayerType));
+        }
```
aelred: This might be another place you can do `relationshipTypes = stream.reduce(Sets::intersection)`.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -136,6 +175,122 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, EmbeddedGrakn
...
+                Var labelVar = var();
+                allFragments.add(Fragments.label(LabelProperty.of(label), labelVar, ImmutableSet.of(label)));
+                allFragments.addAll(EquivalentFragmentSets.isa(IsaProperty.of(label(label).admin()),
+                        relationshipVar, labelVar, relationshipType.isImplicit()).fragments());
+            }
+        });
+    }
+
+    private static Set<Type> getAllPossibleRelationShipTypesOfARolePlayer(
+            Var rolePlayer, Multimap<Var, Type> instanceVarTypeMap, Map<Type, Set<RelationshipType>> relationshipMap) {
```
aelred: If you move the look-up into `instanceVarTypeMap` outside of this function, the signature can be a bit simpler:
```
private static Set<Type> getAllPossibleRelationshipTypesOfARolePlayer(
    Set<Type> types, Map<Type, Set<RelationshipType>> relationshipMap) {
```

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -136,6 +175,122 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, EmbeddedGrakn
...
+        while (typeIterator.hasNext() && !relationshipTypes.isEmpty()) {
+            Type nextRolePlayerType = typeIterator.next();
+            relationshipTypes.retainAll(relationshipMap.get(nextRolePlayerType));
+        }
+
+        //TODO: if relationshipTypes here is empty, the query will not match any data
+        return relationshipTypes;
+    }
+
+    private static void getAllPossibleRelationships(Map<Type, Set<RelationshipType>> relationshipMap, Type metaType) {
```
aelred: Since this mutates the collection you give it, maybe call it `addAllPossibleRelationships`.

Also, is the second argument always a meta-type?

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -136,6 +175,122 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, EmbeddedGrakn
...
     }
 
+    private static void inferRelationshipTypes(EmbeddedGraknTx<?> tx, Set<Fragment> allFragments) {
+        // find all vars representing types
+        Map<Var, Type> labelVarTypeMap = new HashMap<>();
+        allFragments.stream()
+                .filter(LabelFragment.class::isInstance)
+                .forEach(fragment -> {
+                    // TODO: labels() should return ONE label instead of a set
+                    Type type = tx.getType(Iterators.getOnlyElement(((LabelFragment) fragment).labels().iterator()));
```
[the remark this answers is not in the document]
duckofyork: is it possible to get rules here?

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -136,6 +175,122 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, EmbeddedGrakn
...
+                    if (type != null && !type.isRole()) labelVarTypeMap.put(fragment.start(), type);
```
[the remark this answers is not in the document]
duckofyork: It's actually possible!

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -136,6 +175,122 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, EmbeddedGrakn
...
+            size = instanceVarTypeMap.size();
+            allFragments.stream()
+                    .filter(fragment -> labelVarTypeMap.containsKey(fragment.start()))
+                    .filter(fragment -> fragment instanceof InIsaFragment || fragment instanceof InSubFragment)
+                    .forEach(fragment -> instanceVarTypeMap.put(fragment.end(), labelVarTypeMap.get(fragment.start())));
+        } while (size != instanceVarTypeMap.size());
+
+        // relationship vars and its role player vars
+        Multimap<Var, Var> relationshipRolePlayerMap = HashMultimap.create();
+        allFragments.stream().filter(OutRolePlayerFragment.class::isInstance)
```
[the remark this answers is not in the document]
duckofyork: Can't remember why. It seems that I need all fragments elsewhere in the method, so I just use all fragments here 😄

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -136,6 +175,122 @@ public static GraqlTraversal createTraversal(PatternAdmin pattern, EmbeddedGrakn
...
+        if (labelVarTypeMap.isEmpty()) return;
+
+        // find all vars with direct or indirect out isa edges
+        Multimap<Var, Type> instanceVarTypeMap = HashMultimap.create();
+        int size;
+        do {
+            size = instanceVarTypeMap.size();
+            allFragments.stream()
+                    .filter(fragment -> labelVarTypeMap.containsKey(fragment.start()))
+                    .filter(fragment -> fragment instanceof InIsaFragment || fragment instanceof InSubFragment)
```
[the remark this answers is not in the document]
duckofyork: yes, both `sub` and `isa` are added.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
duckofyork: 👍

duckofyork:
Please re-review. Thanks

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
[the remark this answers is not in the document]
aelred: I think so. You can query for rules like `match $x label grandparent-rule; get` or whatever.

But I think I was wrong about it crashing! If you call `tx.getType("grandparent-rule");`, it will return `null` and you handle that.

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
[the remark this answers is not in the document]
aelred: Is it OK if it just overrides whatever is there?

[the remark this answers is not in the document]
duckofyork: It's kind of okay, but it would be nice to work out which label is the subtype:
`$x label person, $x label mammal;`
Or figure our it wouldn't return anything:
`$x label person, $x label dog;`

PR #2631, opened 2018-02-13
log uncaught exception

lolski:
# Why is this PR needed?
Grakn will log `IOException` and `RuntimeException`, but fails silently in case of other types of exception.

# What does the PR do?
Add a `setDefaultUncaughtExceptionHandler`. 

No

N/A


lolski:
@duckofyork thank you!

PR #2633, opened 2018-02-14
gRPC Client

aelred:
# Why is this PR needed?

We need a Java gRPC client to use our gRPC server - this is our thin/dumb client.

# What does the PR do?

Creates alternative implementations of interfaces such as `GraknTx` and `QueryRunner` that execute using an existing Grakn server.

No.

Loads! There are many methods that simply `throw UnsupportedOperationException()`.

We also will need to make the `Grakn#session` method start a `RemoteGraknSession`, instead of a `(Local)GraknSessionImpl`. For now, I have a different entry point: `RemoteGrakn#session`.


grakn-client/src/main/java/ai/grakn/remote/RemoteGrakn.java
```diff
@@ -0,0 +1,35 @@
...
+package ai.grakn.remote;
+
+import ai.grakn.GraknSession;
+import ai.grakn.Keyspace;
+import ai.grakn.util.SimpleURI;
+
+/**
+ * @author Felix Chapman
+ */
+public final class RemoteGrakn {
```
fppt: Java docs need work here.

grakn-client/src/main/java/ai/grakn/remote/GrpcClient.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private TxResponse responseOrThrow() {
+        Response response;
+
+        try {
+            response = communicator.receive();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw Throwables.propagate(e);
```
fppt: What does this do? Also: https://github.com/google/guava/wiki/Why-we-deprecated-Throwables.propagate . . . Relevant?

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknSession.java
```diff
@@ -0,0 +1,96 @@
...
+import ai.grakn.rpc.generated.GraknGrpc.GraknStub;
+import ai.grakn.util.SimpleURI;
+import com.google.common.annotations.VisibleForTesting;
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+
+/**
+ * @author Felix Chapman
+ */
+class RemoteGraknSession implements GraknSession {
```
fppt: Ditto for docs

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknSession.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    public GraknTx open(GraknTxType transactionType) {
+        return RemoteGraknTx.create(this, transactionType);
+    }
+
+    @Override
+    public GraknComputer getGraphComputer() {
+        throw new UnsupportedOperationException(); // TODO
```
fppt: Ohhhhh this is not going to be fun

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
```diff
@@ -0,0 +1,331 @@
...
+import javax.annotation.Nullable;
+import java.util.Collection;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Stream;
+
+/**
+ * @author Felix Chapman
+ */
+class RemoteGraknTx implements GraknTx, GraknAdmin {
```
fppt: Ditto for docs

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknSession.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    public Keyspace keyspace() {
+        return keyspace;
+    }
+
+    @Override
+    public GraknConfig config() {
+        throw new UnsupportedOperationException(); // TODO
```
fppt: This should be "easy" enough to do. It just needs to mimic the behaviour of the REST controller used to get the config. I.e. read the properties file locally and overwrite some keys.

grakn-client/src/main/java/ai/grakn/remote/RemoteQueryRunner.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * @author Felix Chapman
+ */
+final class RemoteQueryRunner implements QueryRunner {
```
fppt: 🐐

grakn-client/src/main/java/ai/grakn/remote/GrpcClient.java
aelred: Ooh... I didn't know that. So it looks like I should just say `throw new RuntimeException(e);`. I hate `InterruptedException`s 😞

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknSession.java
aelred: If this is a remote API, then we can't expect the config file to be present. I think we should try removing this method from `GraknSession` and make it internal - but that's easier said than done 😛

grakn-engine/src/test/java/ai/grakn/engine/rpc/GrpcServerTest.java
```diff
@@ -83,9 +88,7 @@
 public class GrpcServerTest {
 
     private static final String EXCEPTION_MESSAGE = "OH DEAR";
-    private static final GraknException EXCEPTION = new GraknException(EXCEPTION_MESSAGE) {
-        private static final long serialVersionUID = 8491972914457727509L;
-    };
+    private static final GraknException EXCEPTION = GraqlQueryException.create(EXCEPTION_MESSAGE);
```
duckofyork: I like the exception message 👍

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknSession.java
```diff
@@ -0,0 +1,101 @@
...
+        this.channel = channel;
+    }
+
+    @VisibleForTesting
+    public static RemoteGraknSession create(Keyspace keyspace, SimpleURI uri, ManagedChannel channel) {
+        return new RemoteGraknSession(keyspace, uri, channel);
+    }
+
+    public static RemoteGraknSession create(Keyspace keyspace, SimpleURI uri){
+        // TODO: usePlainText is insecure
```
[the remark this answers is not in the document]
aelred: It means we're explicitly not using encryption - like http vs https. I'll add a comment explaining.

PR #2635, opened 2018-02-14
marking a constructor argument as @Nullable. Adding method javadocs.

lolski:
# Why is this PR needed?
Adding method-level Javadocs also to get myself familiar with the code.

# What does the PR do?
1. Marking the constructor argument `config`  `@Nullable`
2. Adding method-level Javadocs

No. I am marking the constructor argument `config`  `@Nullable`, which shouldn't break any compatibility.

N/A

### DIFF
--- a/grakn-kb/src/main/java/ai/grakn/factory/GraknSessionImpl.java
+++ b/grakn-kb/src/main/java/ai/grakn/factory/GraknSessionImpl.java
@@ -39,6 +39,7 @@ import org.apache.tinkerpop.gremlin.structure.Graph;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nullable;
 import javax.ws.rs.core.UriBuilder;
 import java.net.URI;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


grakn-kb/src/main/java/ai/grakn/factory/GraknSessionImpl.java
```diff
@@ -115,11 +123,27 @@ public CommitLogHandler commitLogHandler(){
         return commitLogHandler;
     }
 
+    /**
+     * This methods creates a {@link GraknSessionImpl} object for the remote API.
+     * A user should not call this method directly.
+     * See {@link Grakn#session(String, String)} for creating a {@link GraknSession} for the remote API
+     * @param keyspace
```
fppt: If you not going to describe the parameter (and in this case I don't think you should) then I would say no need to list it here. It just makes the code file longer with no gain.

lolski: that is true. fixed in https://github.com/graknlabs/grakn/pull/2635/commits/4ae0304dcad296ebbdefd60fd19bacdec0b5b4e5

PR #2636, opened 2018-02-14
Cleaning up some code smells

fppt:
# Why is this PR needed?

General code quality improvements

# What does the PR do?

Cleans up some code smells and bugs identified by [Sonar](https://sonarcloud.io/dashboard?id=ai.grakn%3Agrakn)

No

More code smell cleanup

### DIFF
--- a/grakn-bootup/src/main/java/ai/grakn/bootup/AbstractProcessHandler.java
+++ b/grakn-bootup/src/main/java/ai/grakn/bootup/AbstractProcessHandler.java
@@ -33,7 +33,8 @@ import java.util.Optional;
  */
 public abstract class AbstractProcessHandler {
 
-    public final static long WAIT_INTERVAL_S=2;
+    public static final long WAIT_INTERVAL_S=2;
+    public static final String BIN = "/bin/sh";
 
     public OutputCommand executeAndWait(String[] cmdarray, String[] envp, File dir) {
 
@@ -41,36 +42,25 @@ public abstract class AbstractProcessHandler {
         int exitValue = 1;
 
         Process p;
-        BufferedReader reader = null;
         try {
             p = Runtime.getRuntime().exec(cmdarray, envp, dir);
             p.waitFor();
             exitValue = p.exitValue();
-            reader =
-                    new BufferedReader(new InputStreamReader(p.getInputStream(), StandardCharsets.UTF_8));
-
-            String line;
-            while ((line = reader.readLine()) != null) {
-                outputS.append(line).append("\n");
-            }
-
-        } catch (InterruptedException | IOException e) {
-            // DO NOTHING
-        } finally {
-            if (reader != null) {
-                try {
-                    reader.close();
-                } catch (IOException e) {
-                    // DO NOTHING
+            try(BufferedReader reader = new BufferedReader(new InputStreamReader(p.getInputStream(), StandardCharsets.UTF_8))){
+                String line;
+                while ((line = reader.readLine()) != null) {
+                    outputS.append(line).append("\n");
                 }
             }
+        } catch (InterruptedException | IOException e) {
+            // DO NOTHING
         }
         return new OutputCommand(outputS.toString().trim(), exitValue);
     }
 
     public Optional<String> getPidFromFile(Path fileName) {
         String pid=null;
-        if (Files.exists(fileName)) {
+        if (fileName.toFile().exists()) {
             try {
                 pid = new String(Files.readAllBytes(fileName),StandardCharsets.UTF_8).trim();
             } catch (IOException e) {
@@ -82,7 +72,7 @@ public abstract class AbstractProcessHandler {
 
     public String getPidFromPsOf(String processName) {
         return executeAndWait(new String[]{
-                    "/bin/sh",
+                    BIN,
                     "-c",
                     "ps -ef | grep " + processName + " | grep -v grep | awk '{print $2}' "
             }, null, null).output;
@@ -90,7 +80,7 @@ public abstract class AbstractProcessHandler {
 
     private void kill(int pid) {
         executeAndWait(new String[]{
-                "/bin/sh",
+                BIN,
                 "-c",
                 "kill " + pid
         }, null, null);
@@ -98,14 +88,14 @@ public abstract class AbstractProcessHandler {
 
     private OutputCommand kill(int pid, String signal) {
         return executeAndWait(new String[]{
-                "/bin/sh",
+                BIN,
                 "-c",
                 "kill -"+signal+" " + pid
         }, null, null);
     }
 
     public int retrievePid(Path pidFile) {
-        if(!Files.exists(pidFile)) {
+        if(!pidFile.toFile().exists()) {
             return -1;
         }
         try {
@@ -132,18 +122,15 @@ public abstract class AbstractProcessHandler {
             }
         } while (outputCommand.succes());
         System.out.println("SUCCESS");
-        try {
-            if(Files.exists(pidFile)) {
-                Files.delete(pidFile);
-            }
-        } catch (IOException e) {
-            // DO NOTHING
+        File file = pidFile.toFile();
+        if(file.exists()) {
+            file.delete();
         }
     }
 
     public String selectCommand(String osx, String linux) {
         OutputCommand operatingSystem = executeAndWait(new String[]{
-                "/bin/sh",
+                BIN,
                 "-c",
                 "uname"
         },null,null);
@@ -153,14 +140,14 @@ public abstract class AbstractProcessHandler {
     public boolean processIsRunning(Path pidFile) {
         boolean isRunning = false;
         String processPid;
-        if (Files.exists(pidFile)) {
+        if (pidFile.toFile().exists()) {
             try {
                 processPid = new String(Files.readAllBytes(pidFile),StandardCharsets.UTF_8);
                 if(processPid.trim().isEmpty()) {
                     return false;
                 }
                 OutputCommand command = executeAndWait(new String[]{
-                        "/bin/sh",
+                        BIN,
                         "-c",
                         "ps -p "+processPid.trim()+" | grep -v CMD | wc -l"
                 },null,null);
--- a/grakn-bootup/src/main/java/ai/grakn/bootup/AbstractProcessHandler.java
+++ b/grakn-bootup/src/main/java/ai/grakn/bootup/AbstractProcessHandler.java
@@ -124,7 +124,11 @@ public abstract class AbstractProcessHandler {
         System.out.println("SUCCESS");
         File file = pidFile.toFile();
         if(file.exists()) {
-            file.delete();
+            try {
+                Files.delete(pidFile);
+            } catch (IOException e) {
+                // DO NOTHING
+            }
         }
     }
 
--- a/grakn-bootup/src/main/java/ai/grakn/bootup/AbstractProcessHandler.java
+++ b/grakn-bootup/src/main/java/ai/grakn/bootup/AbstractProcessHandler.java
@@ -34,7 +34,7 @@ import java.util.Optional;
 public abstract class AbstractProcessHandler {
 
     public static final long WAIT_INTERVAL_S=2;
-    public static final String BIN = "/bin/sh";
+    public static final String SH = "/bin/sh";
 
     public OutputCommand executeAndWait(String[] cmdarray, String[] envp, File dir) {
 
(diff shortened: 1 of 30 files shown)


grakn-bootup/src/main/java/ai/grakn/bootup/AbstractProcessHandler.java
```diff
@@ -33,44 +33,34 @@
  */
 public abstract class AbstractProcessHandler {
 
-    public final static long WAIT_INTERVAL_S=2;
+    public static final long WAIT_INTERVAL_S=2;
+    public static final String BIN = "/bin/sh";
```
aelred: This should be called `SH`, because it points to the `sh` command 👍

grakn-bootup/src/main/java/ai/grakn/bootup/AbstractProcessHandler.java
```diff
@@ -33,44 +33,34 @@
...
-                }
-            }
         }
         return new OutputCommand(outputS.toString().trim(), exitValue);
     }
 
     public Optional<String> getPidFromFile(Path fileName) {
         String pid=null;
-        if (Files.exists(fileName)) {
+        if (fileName.toFile().exists()) {
```
aelred: Did Sonarcloud say to do this one? Just curious.

grakn-bootup/src/main/java/ai/grakn/bootup/AbstractProcessHandler.java
kasper-piskorski: maybe rename it to SHELL? BIN is a bit vague

grakn-engine/src/main/java/ai/grakn/engine/data/RedisWrapper.java
```diff
@@ -150,7 +150,7 @@ public RedisWrapper build() {
             if (useSentinel) {
                 jedisPool = new JedisSentinelPool(masterName, uriSet, poolConfig, TIMEOUT);
             } else {
-                String uri = uriSet.stream().findFirst().get();
+                String uri = uriSet.iterator().next();
```
kasper-piskorski: If we want a single element then`Iterables.getOnlyElement(uriSet)` might be useful.

grakn-core/src/main/java/ai/grakn/GraknSession.java
```diff
@@ -64,9 +62,8 @@
     /**
      * Closes the main connection to the graph. This should be done at the end of using the graph.
      *
-     * @throws GraknTxOperationException when more than 1 transaction is open on the graph
      */
-    void close() throws GraknTxOperationException;
```
aelred: I think it's good to keep `RuntimeException`s in the Javadocs (but not the signatures). An example of this [here](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#sort(java.util.List)) to prove I'm not crazy.

grakn-core/src/main/java/ai/grakn/Keyspace.java
```diff
@@ -49,6 +49,7 @@
 
     @Override
     public int compareTo(Keyspace o) {
+        if(equals(o)) return 0;
```
aelred: Did this make Sonarcloud sad also?

grakn-engine/src/main/java/ai/grakn/engine/data/RedisWrapper.java
aelred: Maybe `Iterables#getOnlyElement` is better here? I guess it depends if we expect there to be more than one or not.

grakn-engine/src/main/java/ai/grakn/engine/lock/JedisLock.java
```diff
@@ -33,7 +34,7 @@
  */
 public class JedisLock implements Lock {
 
-    private static final long TIMEOUT_MS = 10 * 1000;
+    private static final long TIMEOUT_MS = 10L * 1000L;
```
aelred: I quite like `10_000L`

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/plan/GraqlTraversalPlanner.java
```diff
@@ -105,14 +108,19 @@ private static Atom optimalCandidate(List<Atom> candidates){
             List<Atom> atomsToPlan = new ArrayList<>(atoms);
             atomsToPlan.remove(first);
```
kasper-piskorski: do we want to remove a possible `null` from a list?

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/GreedyTraversalPlan.java
```diff
@@ -221,10 +221,12 @@ private static void processFragmentWithFixedCost(List<Fragment> plan,
             plan.add(fragment);
             double logInstanceCount = -1D;
             if (fragment.getShardCount(tx).isPresent()) {
-                long shardCount = fragment.getShardCount(tx).get();
-                if (shardCount > 0) {
-                    logInstanceCount = Math.log(shardCount - 1D + SHARD_LOAD_FACTOR) +
-                            Math.log(tx.admin().shardingThreshold());
+                Optional<Long> shardCount = fragment.getShardCount(tx);
+                if(shardCount.isPresent()) {
```
aelred: We're making this check twice now! I think you can remove the first check on line 223.

grakn-kb/src/main/java/ai/grakn/kb/internal/concept/RelationshipImpl.java
```diff
@@ -121,11 +121,8 @@ public Relationship attributeRelationship(Attribute attribute) {
 
     @Override
     public Stream<Attribute<?>> keys(AttributeType[] attributeTypes) {
-        if(reified().isPresent()){
-            return reified().get().attributes(attributeTypes);
-        } else {
-            return Stream.empty();
-        }
+        Optional<RelationshipReified> reified = reified();
+        return reified.map(relationshipReified -> relationshipReified.attributes(attributeTypes)).orElseGet(Stream::empty);
```
aelred: You should totally one-liner this: `return reified().map(...)` - only two extra characters!

grakn-test/test-snb/src/main/java/ai/grakn/GraknQueryHandlers.java
```diff
@@ -156,14 +158,14 @@ public void executeOperation(LdbcQuery8 ldbcQuery8, GraknDbConnectionState dbCon
             GraknSession session = dbConnectionState.session();
             try (GraknTx graknTx = session.open(GraknTxType.READ)) {
                 // for speed the query is again split into the ordering and limit phase
-                Var $reply = var("aReply");
-                Var $responder = var("responder");
-                Var $responderId = var("responderId");
+                Var reply = var("aReply");
+                Var responder = var("responder");
+                Var responderId = var("responderId");
```
aelred: ;_;

grakn-test/test-snb/src/main/java/ai/grakn/GraknQueryHandlers.java
```diff
@@ -156,14 +158,14 @@ public void executeOperation(LdbcQuery8 ldbcQuery8, GraknDbConnectionState dbCon
...
                 Match orderQuery = match(
                         $person.has(PERSON_ID, ldbcQuery8.personId()),
                         var().rel($person).rel($message).isa(HAS_CREATOR),
-                        var().rel($message).rel(REPLY, $reply).isa(REPLY_OF),
-                        $reply.has(CREATION_DATE, $date).has(MESSAGE_ID, $messageId)
+                        var().rel($message).rel(REPLY, reply).isa(REPLY_OF),
+                        reply.has(CREATION_DATE, $date).has(MESSAGE_ID, $messageId)
```
aelred: We only changed some of the dollars. Should probably be all or nothing.

grakn-engine/src/main/java/ai/grakn/engine/data/RedisWrapper.java
fppt: That will throw an exception is there is more than one element in the collection, so I can't use it here.

grakn-bootup/src/main/java/ai/grakn/bootup/AbstractProcessHandler.java
fppt: Good point 👍

grakn-bootup/src/main/java/ai/grakn/bootup/AbstractProcessHandler.java
fppt: Yep. Apparently `File.exists` is VERY slow with JDK 8

grakn-core/src/main/java/ai/grakn/GraknSession.java
fppt: Sonar does not like it and neither do I to be honest. It will be do difficult to maintain as internal implementation changes. So I would say leave it removed but can put it back if you feel strongly about it.

grakn-core/src/main/java/ai/grakn/Keyspace.java
fppt: Yep. The `compareTo` was not obeying the equality contract

grakn-engine/src/main/java/ai/grakn/engine/data/RedisWrapper.java
fppt: We can expect more than one so I can't use`Iterables`

grakn-engine/src/main/java/ai/grakn/engine/lock/JedisLock.java
fppt: hahahaha good point :)

grakn-test/test-snb/src/main/java/ai/grakn/GraknQueryHandlers.java
fppt: Working on the rest in a separate PR. I am going according to the order provided by sonar

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/plan/GraqlTraversalPlanner.java
kasper-piskorski: It shouldn't have nulls. `first` can be null only if `candidates` are empty. Basically we can move the if statement two lines up.

grakn-core/src/main/java/ai/grakn/GraknSession.java
aelred: Are you sure it doesn't like it in the Javadoc? Most libraries I can find includes `RuntimeException`s in the Javadoc.

To be clear, I agree with removing them from the method signature - since that does nothing.

grakn-core/src/main/java/ai/grakn/Keyspace.java
aelred: Oh, [this one](https://sonarcloud.io/project/issues?id=ai.grakn%3Agrakn&issues=AWFnFYlYr0KnSpTG8FT6&open=AWFnFYlYr0KnSpTG8FT6)?

Findbugs got upset about the same thing. I don't _think_ your change will make it happy. It's upset because we have implemented `compareTo` but we haven't implemented `equals`. This is because we expect `@AutoValue` to implement `equals` for us.

For Findbugs I just disabled that check for certain classes... I don't know what we can do here though.

fppt:
@kasper-piskorski @aelred re-review please

aelred:
@fppt I'm good to go, excepting removing the `@throws` from the Javadoc. Sorry for pickiness :P

grakn-core/src/main/java/ai/grakn/Keyspace.java
aelred: Random thought... you might be able to fix the issue by explicitly declaring `public abstract boolean equals(...) {` and `public abstract int hashCode() {` in `Keyspace`. That way findbugs/sonar will understand that subclasses _must_ provide an `equals` and `hashCode` implementation.

grakn-core/src/main/java/ai/grakn/GraknSession.java
fppt: I would say all the APIs should have `@throws` for runtime exceptions or none of them should. This middle ground is kind of silly and does not provide much information. 

Technically nearly every method on the core api can throw `GraknTxOperationException`. So we should be adding it everywhere.

Sorry but I am being a stickler here as well. If it's hard to maintain and the middle ground is meaningless then we should remove it.

fppt:
@aelred How is this for a middleground: If you create a task on TP to document all the runtime exceptions on the public apis then I will put back the one I removed. Otherwise I am going to be sticky here as well.

aelred:
@fppt that's fair! You don't have to re-add the one here, but I'll add a TP task to put them on all the API methods.

PR #2647, opened 2018-02-20
Grakn Transaction via Grpc - Basic Schema Concept Puts

fppt:
# Why is this PR needed?

It is needed in order to flesh out the GRPC api to support the embedded java API.

# What does the PR do?

Adds *very* basic support for the api by allowing you to add schema concepts.

No

1. Do not parse into graql queries. Rather use the embedded API directly
2. Return the created concepts.
3. Getters lots and lots of getters


grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
```diff
@@ -75,27 +88,35 @@ static RemoteGraknTx create(RemoteGraknSession session, GraknTxType txType) {
...
     @Override
     public <V> AttributeType<V> putAttributeType(Label label, AttributeType.DataType<V> dataType) {
-        throw new UnsupportedOperationException(); // TODO
+        return putSchemaConcept(label, ATTRIBUTE, var("x").datatype(dataType));
     }
 
     @Override
     public Rule putRule(Label label, Pattern when, Pattern then) {
-        throw new UnsupportedOperationException(); // TODO
+        return putSchemaConcept(label, RULE, var("x").when(when).then(then));
```
aelred: Maybe we can extract `var("x")` into a constant, since otherwise it looks a bit magic.

Or go full functional and pass a lambda 😛 
```
return putSchemaConcept(label, RULE, var -> var.when(when).then(then));
```

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
```diff
@@ -75,27 +88,35 @@ static RemoteGraknTx create(RemoteGraknSession session, GraknTxType txType) {
...
     public Role putRole(Label label) {
-        throw new UnsupportedOperationException(); // TODO
+        return putSchemaConcept(label, ROLE);
+    }
+
+    private <X extends Concept> X putSchemaConcept(Label label, Schema.MetaSchema meta, VarPattern ... optionalVars){
+        List<VarPattern> vars = new ArrayList<>();
+        vars.add(var("x").label(label).sub(meta.getLabel().getValue()));
+        vars.addAll(Arrays.asList(optionalVars));
+        queryRunner().run(Graql.withoutGraph().define(vars));
```
aelred: You can skip `withoutGraph` if you like and do `Graql.define(vars)`

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
```diff
@@ -75,27 +88,35 @@ static RemoteGraknTx create(RemoteGraknSession session, GraknTxType txType) {
...
 
     @Override
     public Role putRole(Label label) {
-        throw new UnsupportedOperationException(); // TODO
+        return putSchemaConcept(label, ROLE);
+    }
+
+    private <X extends Concept> X putSchemaConcept(Label label, Schema.MetaSchema meta, VarPattern ... optionalVars){
+        List<VarPattern> vars = new ArrayList<>();
+        vars.add(var("x").label(label).sub(meta.getLabel().getValue()));
```
aelred: The "more correct" but less nice-looking thing to do is this:
```
vars.add(var("x").label(label).sub(var().label(meta.getLabel())));
```
Then we don't extract the string from the label.

Ideally `sub` should also accept a `Label` though...

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
```diff
@@ -75,27 +88,35 @@ static RemoteGraknTx create(RemoteGraknSession session, GraknTxType txType) {
...
+        vars.addAll(Arrays.asList(optionalVars));
+        queryRunner().run(Graql.withoutGraph().define(vars));
+        return null;
```
aelred: I guess this is waiting on my PR?

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
```diff
@@ -178,17 +199,18 @@ public QueryBuilder graql() {
...
     @Override
     public void close() {
-        client.close();
+        if(!isClosed()) client.close();
     }
 
     @Override
     public void abort() {
-        throw new UnsupportedOperationException(); // TODO
+        if(!isClosed()) client.close();
```
aelred: Is `abort` always the same as `close`? Sounds like a `default` implementation :D

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
fppt: yep

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
fppt: yep it is and that is a good suggestion!

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
fppt: What's the point of the `withoutGraph` method then? I.e. what's the diff?

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
fppt: I actually like the functional approach

fppt:
@aelred re-review please sir

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
aelred: Basically as a means to get at a `QueryBuilder`, since some methods might accept a `QueryBuilder` and then build queries with it. The only example I can find in the code is `Migrator`, which gets a `QueryBuilder` so it can toggle inference off.

fppt:
@aelred can you provide more details? When I make those changes the tests just hang forever

aelred:
@fppt I cleaned it up in [my PR](https://github.com/graknlabs/grakn/pull/2656/commits/bff90a8b88426dff8e552fbc4048d5fcd5281cfb). It worked locally, anyhow.

PR #2648, opened 2018-02-20
Add remote concept implementations

aelred:
# Why is this PR needed?

We need representations of the concepts for the gRPC client.

# What does the PR do?

Adds `Remote_` implementations for every `Concept` class, e.g. `RemoteEntityType`. Each class contains and is identified by a transaction with a concept ID.

No.

A lot of methods to implement as always.

### DIFF
--- a/grakn-client/src/main/java/ai/grakn/remote/GrpcClient.java
+++ b/grakn-client/src/main/java/ai/grakn/remote/GrpcClient.java
@@ -21,18 +21,28 @@ package ai.grakn.remote;
 import ai.grakn.GraknTx;
 import ai.grakn.GraknTxType;
 import ai.grakn.Keyspace;
+import ai.grakn.concept.Concept;
+import ai.grakn.concept.ConceptId;
+import ai.grakn.graql.Graql;
 import ai.grakn.graql.Query;
+import ai.grakn.graql.Var;
+import ai.grakn.graql.admin.Answer;
+import ai.grakn.graql.internal.query.QueryAnswer;
 import ai.grakn.grpc.GrpcUtil;
 import ai.grakn.grpc.GrpcUtil.ErrorType;
 import ai.grakn.grpc.TxGrpcCommunicator;
 import ai.grakn.grpc.TxGrpcCommunicator.Response;
+import ai.grakn.remote.concept.RemoteConcepts;
 import ai.grakn.rpc.generated.GraknGrpc;
+import ai.grakn.rpc.generated.GraknOuterClass;
 import ai.grakn.rpc.generated.GraknOuterClass.TxResponse;
 import ai.grakn.util.CommonUtil;
 import com.google.common.collect.AbstractIterator;
+import com.google.common.collect.ImmutableMap;
 import io.grpc.Metadata;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
+import mjson.Json;
 
 import javax.annotation.Nullable;
 import java.util.Iterator;
@@ -83,7 +93,7 @@ final class GrpcClient implements AutoCloseable {
 
                 switch (response.getResponseCase()) {
                     case QUERYRESULT:
-                        return GrpcUtil.getQueryResult(tx, response.getQueryResult());
+                        return convert(tx, response.getQueryResult());
                     case DONE:
                         return endOfData();
                     default:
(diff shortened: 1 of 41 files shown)


grakn-client/src/main/java/ai/grakn/remote/GrpcClient.java
```diff
@@ -139,4 +149,54 @@ private static RuntimeException convertStatusRuntimeException(StatusRuntimeExcep
...
+            map.put(Graql.var(grpcVar), convert(tx, grpcConcept));
+        });
+
+        return new QueryAnswer(map.build());
+    }
+
+    private static Concept convert(RemoteGraknTx tx, GraknOuterClass.Concept concept) {
+        ConceptId id = ConceptId.of(concept.getId());
+
+        switch (concept.getBaseType()) {
```
fppt: HAHAHAHAHAHAHA. . . Where have I seen this before :P

grakn-client/src/main/java/ai/grakn/remote/concept/RemoteAttributeType.java
```diff
@@ -0,0 +1,68 @@
...
+import com.google.auto.value.AutoValue;
+
+import javax.annotation.Nullable;
+
+/**
+ * @author Felix Chapman
+ *
+ * @param <D> The data type of this attribute type
+ */
+@AutoValue
```
fppt: For this and all the other Remote concept maybe we want to override the equality to be based on the id? That's how the embedded API work and maybe this should do the same thing?

fppt: Also maybe the current concepts can be autovalued? Not relevant to this PR of coourse

grakn-client/src/main/java/ai/grakn/remote/concept/RemoteConcepts.java
```diff
@@ -0,0 +1,76 @@
...
+import ai.grakn.concept.RelationshipType;
+import ai.grakn.concept.Role;
+import ai.grakn.concept.Rule;
+import ai.grakn.concept.Type;
+import ai.grakn.remote.RemoteGraknTx;
+
+/**
+ * @author Felix Chapman
+ */
+public class RemoteConcepts {
```
fppt: These Java docs make me sad. Also maybe the class should be called `RemoteConceptFactory` ?

grakn-client/src/main/java/ai/grakn/remote/concept/RemoteSchemaConcept.java
```diff
@@ -0,0 +1,81 @@
...
+import ai.grakn.concept.Rule;
+import ai.grakn.concept.SchemaConcept;
+
+import javax.annotation.Nullable;
+import java.util.stream.Stream;
+
+/**
+ * @author Felix Chapman
+ *
+ * @param <Self> The exact type of this class
```
fppt: `Self` ? That's an interesting convention you introducing

grakn-client/src/main/java/ai/grakn/remote/concept/RemoteThing.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.util.stream.Stream;
+
+/**
+ * @author Felix Chapman
+ *
+ * @param <Self> The exact type of this class
+ * @param <MyType> the type of an instance of this class
+ */
+abstract class RemoteThing<Self extends Thing, MyType extends Type> extends RemoteConcept implements Thing {
```
fppt: Ditto for `MyType` not against it . . . just find it interesting

grakn-client/src/test/java/ai/grakn/remote/GrpcServerMock.java
```diff
@@ -0,0 +1,124 @@
...
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Felix Chapman
+ */
+public final class GrpcServerMock extends CompositeTestRule {
```
fppt: Better java docs needed here for sure. This class is quite complex

grakn-engine/src/main/java/ai/grakn/engine/rpc/GrpcConverter.java
```diff
@@ -95,6 +96,33 @@ public Object buildDefault(Object object) {
...
     private GraknOuterClass.Concept makeConcept(Concept concept) {
-        return GraknOuterClass.Concept.newBuilder().setId(concept.getId().getValue()).build();
+        return GraknOuterClass.Concept.newBuilder()
+                .setId(concept.getId().getValue())
+                .setBaseType(getBaseType(concept))
+                .build();
+    }
+
+    private GraknOuterClass.BaseType getBaseType(Concept concept) {
+        if (concept.isEntityType()) {
```
fppt: Deja vu anyone?

grakn-client/src/main/java/ai/grakn/remote/concept/RemoteSchemaConcept.java
aelred: Haha, I was so happy when I realised I could do this!

The only problem I can think of is that `Self` (and `MyType`) look like they might be classes. In an IDE this isn't an issue though:
image

grakn-client/src/main/java/ai/grakn/remote/concept/RemoteConcepts.java
aelred: You're right on the Javadocs! I keep forgetting.

I picked this naming convention because it's what I do in Graql - and the same one used in Guava.

grakn-client/src/main/java/ai/grakn/remote/concept/RemoteAttributeType.java
aelred: I wondered about if we could autovalue the old concepts... maybe? It looks tough. `ConceptImpl` contains all these fields:

```
public abstract class ConceptImpl implements Concept, ConceptVertex, CacheOwner{
    private final Set<Cache> registeredCaches = new HashSet<>();
    //WARNING: DO not flush the current shard into the central cache. It is not safe to do so in a concurrent environment
    private final Cache<Shard> currentShard = Cache.createTxCache(this, Cacheable.shard(), () -> {
        String currentShardId = vertex().property(Schema.VertexProperty.CURRENT_SHARD);
        Vertex shardVertex = vertex().tx().getTinkerTraversal().V().has(Schema.VertexProperty.ID.name(), currentShardId).next();
        return vertex().tx().factory().buildShard(shardVertex);
    });
    private final Cache<Long> shardCount = Cache.createSessionCache(this, Cacheable.number(), () -> shards().count());
    private final Cache<ConceptId> conceptId = Cache.createPersistentCache(this, Cacheable.conceptId(), () -> ConceptId.of(vertex().property(Schema.VertexProperty.ID)));
```

and they're not immutable and not part of object equality either... so it may be a bad idea.

aelred: On equality, personally I think concepts from different transactions should not be equal - since they are not interchangeable with each other (imagine two concepts with the same ID from totally different keyspaces or even servers being equal).

On the other hand, the embedded concepts work this way and we _should_ make it consistent 🤷‍♂️

aelred:
@fppt I think I'm done on this, except the `Concept` equality - still not sure which way to go on that one.

grakn-client/src/main/java/ai/grakn/remote/concept/RemoteAttributeType.java
fppt: Maybe we should be changing the embedded concepts to not use id? That sounds better based on your valid points?

aelred: I think that's a good idea! Something _may_ subtly break - we'll see.

aelred: actually I changed my mind everything is perfect as it is right now

aelred:
@fppt re-review time 👍

PR #2649, opened 2018-02-20
log unhandled exception which causes nasty breakage in post-processing

lolski:
# Why is this PR needed?
If background tasks such as post processing failed due to an exception, the consequences are quite nasty:
1. Fails silently. No logs.
2. Subsequent task will be not be scheduled. This particular behavior is from Java's scheduled executor. In practice, this means that if for any reason (such as the exception due to network error), the scheduled task which perform post processing dies, it will fail, and no more post processing will be scheduled unless grakn is restarted.

Basically this is the same issue as what these guys had: https://stackoverflow.com/a/24902026/1622847

# What does the PR do?
This PR does one thing: logs the exception and informs the user to restart Grakn. Post-processing will still stop working, but at least the user is now informed and advised to restart Grakn.

No

Define a better behavior instead of simply logging and rethrowing the exception

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/task/BackgroundTaskRunner.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/task/BackgroundTaskRunner.java
@@ -20,8 +20,11 @@ package ai.grakn.engine.task;
 
 import ai.grakn.GraknConfigKey;
 import ai.grakn.engine.GraknConfig;
+import ai.grakn.util.ErrorMessage;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableSet;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


grakn-engine/src/main/java/ai/grakn/engine/task/BackgroundTaskRunner.java
```diff
@@ -54,7 +59,15 @@ public BackgroundTaskRunner(GraknConfig graknConfig) {
     public void register(BackgroundTask backgroundTask){
         if(!registeredTasks.contains(backgroundTask)) {
             registeredTasks.add(backgroundTask);
-            threadPool.scheduleAtFixedRate(backgroundTask::run, backgroundTask.period(), backgroundTask.period(), TimeUnit.SECONDS);
+            threadPool.scheduleAtFixedRate(() -> {
+                try {
+                    backgroundTask.run();
+                } catch (Throwable e) {
```
lolski: Pay attention to how I deliberately catching every kind of exceptions / errors so that they get logged.

PR #2656, opened 2018-02-23
Add support for getting properties in gRPC

aelred:
# Why is this PR needed?

To support methods such as `SchemaConcept#getLabel` over gRPC.

# What does the PR do?

Adds support for all the getter methods on concepts - in particular the ones that cannot be handled with standard queries.

No.

All the other concept methods! The rest will need queries written.


grakn-test/test-integration/src/test/java/ai/grakn/test/engine/GrpcServerIT.java
```diff
@@ -189,7 +189,6 @@ public void whenExecutingTwoParallelQueries_Throw() throws Throwable {
         }
     }
 
-    @Ignore // TODO: re-enable after implement methods
     @Test
```
lolski: 👍

grakn-test/test-integration/src/test/java/ai/grakn/test/engine/GrpcServerIT.java
```diff
@@ -305,7 +304,6 @@ public void whenGettingARule_TheInformationOnTheRuleIsCorrect() {
         }
     }
 
-    @Ignore // TODO: re-enable after implement methods
```
lolski: 👍

grakn-test/test-integration/src/test/java/ai/grakn/test/engine/GrpcServerIT.java
```diff
@@ -334,7 +332,6 @@ public void whenGettingARelationshipType_TheInformationOnTheRelationshipTypeIsCo
...
```
lolski: 👍

grakn-test/test-integration/src/test/java/ai/grakn/test/engine/GrpcServerIT.java
```diff
@@ -349,7 +346,6 @@ public void whenGettingAnAttributeType_TheInformationOnTheAttributeTypeIsCorrect
...
```
lolski: 👍

grakn-grpc/src/main/java/ai/grakn/grpc/ConceptProperty.java
```diff
@@ -0,0 +1,200 @@
...
+                return IS_ABSTRACT;
+            case When:
+                return WHEN;
+            case Then:
+                return THEN;
+            case Regex:
+                return REGEX;
+            default:
+            case UNRECOGNIZED:
+                throw new IllegalArgumentException("Unrecognised " + conceptProperty);
```
lolski: does it makes sense to put the error message in `ErrorMessage` class (which is somewhere in `grakn-engine` if i am not mistaken)

aelred: Maybe. This will only happen if the gRPC server you're connecting to is bugged - so it's kind of meant to be "unreachable".

PR #2659, opened 2018-02-26
GRPC GraknTx Getters

fppt:
# Why is this PR needed?

To expand on the GRPC api functionality

# What does the PR do?

Addes getters for schema concepts

No

More Methods on GRPC

### DIFF
--- a/grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
+++ b/grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
@@ -139,7 +139,7 @@ public final class RemoteGraknTx implements GraknTx, GraknAdmin {
     @Nullable
     @Override
     public <T extends SchemaConcept> T getSchemaConcept(Label label) {
-        return getSchemaConcept(label);
+        return getSchemaConcept(label, null);
     }
 
     @Nullable
--- a/grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
+++ b/grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
@@ -154,31 +154,31 @@ public final class RemoteGraknTx implements GraknTx, GraknAdmin {
     @Nullable
     @Override
     public EntityType getEntityType(String label) {
-        return getSchemaConcept(Label.of(label), Schema.MetaSchema.ENTITY);
+        return getSchemaConcept(Label.of(label), ENTITY);
     }
 
     @Nullable
     @Override
     public RelationshipType getRelationshipType(String label) {
-        return getSchemaConcept(Label.of(label), Schema.MetaSchema.RELATIONSHIP);
+        return getSchemaConcept(Label.of(label), RELATIONSHIP);
     }
 
     @Nullable
     @Override
     public <V> AttributeType<V> getAttributeType(String label) {
-        return getSchemaConcept(Label.of(label), Schema.MetaSchema.ATTRIBUTE);
+        return getSchemaConcept(Label.of(label), ATTRIBUTE);
     }
 
     @Nullable
     @Override
     public Role getRole(String label) {
-        return getSchemaConcept(Label.of(label), Schema.MetaSchema.ROLE);
+        return getSchemaConcept(Label.of(label), ROLE);
     }
 
     @Nullable
     @Override
     public Rule getRule(String label) {
-        return getSchemaConcept(Label.of(label), Schema.MetaSchema.RULE);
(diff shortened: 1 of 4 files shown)


grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
```diff
@@ -90,38 +94,40 @@ public GrpcClient client() {
...
 
-    private <X extends Concept> X putSchemaConcept(Label label, Schema.MetaSchema meta, @Nullable Function<VarPattern, VarPattern> extender){
+    private <X extends SchemaConcept> X putSchemaConcept(Label label, Schema.MetaSchema meta,
+                                                   Supplier<X> supplier,
+                                                   @Nullable Function<VarPattern, VarPattern> extender){
         VarPattern var = var().label(label).sub(var().label(meta.getLabel()));
         if(extender != null) var = extender.apply(var);
         queryRunner().run(Graql.define(var));
-        return null;
+        return supplier.get();
```
aelred: Haha, I didn't even notice this was returning `null` before!

Is it possible to just return the answer from the `define` query?

```
return queryRunner.run(Graql.define(var)).get("x").asSchemaConcept();
```

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
```diff
@@ -215,32 +234,32 @@ public void commit() throws InvalidKBException {
 
     @Override
     public Type getMetaConcept() {
-        throw new UnsupportedOperationException(); // TODO
+        return getSchemaConcept(Schema.MetaSchema.THING.getLabel());
```
aelred: more default methods? :D

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
```diff
@@ -133,13 +139,13 @@ public Role putRole(Label label) {
     @Nullable
     @Override
     public <T extends SchemaConcept> T getSchemaConcept(Label label) {
-        throw new UnsupportedOperationException(); // TODO
+        return getSchemaConcept(label);
```
aelred: giphy

grakn-test/test-integration/src/test/java/ai/grakn/test/engine/GrpcServerIT.java
```diff
@@ -112,6 +112,19 @@ public void whenPuttingEntityType_EnsureItIsAdded(){
...
+    @Test
+    public void whenGettingEntityType_EnsureItIsReturned(){
+        String label = "Oliver";
+        try (GraknTx tx = localSession.open(GraknTxType.WRITE)) {
+            tx.putEntityType(label);
+            tx.commit();
+        }
+
+        try (GraknTx tx = remoteSession.open(GraknTxType.WRITE)){
+            assertNotNull(tx.getEntityType(label));
```
aelred: We could assert the label is correct to be more sure we have the right thing... but I guess that has to wait for my PR #2656

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
```diff
@@ -150,31 +156,44 @@ public Role putRole(Label label) {
...
+    }
+
+    @Nullable
+    private <X extends SchemaConcept> X getSchemaConcept(Label label, @Nullable Schema.MetaSchema meta){
+        VarPattern pattern = var().label(label);
+        if(meta != null) pattern = pattern.sub(var().label(meta.getLabel()));
+        Optional<Answer> result = queryRunner().run(Graql.match(pattern).get()).findAny();
+        if(result.isPresent()){
+            Collection<Concept> answer = result.get().concepts();
+            if (!answer.isEmpty()) return (X) Iterables.getOnlyElement(answer);
```
aelred: Maybe this is simpler:
```
Var var = var("x");
VarPattern pattern = var.label(label);
...
if (result.isPresent()) {
     return (X) result.get().get(var);
}
```

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
fppt: SOLD!

grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
fppt: ya ya PMD already yelled at me :D

PR #2660, opened 2018-02-27
Make get meta's default methods

fppt:
# Why is this PR needed?

Cleans up the code a bit and removes some redundant implementations.

# What does the PR do?

Makes the getting of meta concepts default methods

No

none

### DIFF
--- a/grakn-core/src/main/java/ai/grakn/util/Schema.java
+++ b/grakn-core/src/main/java/ai/grakn/util/Schema.java
@@ -31,10 +31,10 @@ import ai.grakn.concept.Role;
 import ai.grakn.concept.Rule;
 import ai.grakn.concept.SchemaConcept;
 import ai.grakn.concept.Type;
-import com.google.common.collect.ImmutableSet;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
 
 import javax.annotation.CheckReturnValue;
+import javax.annotation.Nullable;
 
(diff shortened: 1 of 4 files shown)


grakn-core/src/main/java/ai/grakn/util/Schema.java
```diff
@@ -116,6 +113,18 @@ public static boolean isMetaLabel(Label label) {
...
+        @CheckReturnValue
+        public static MetaSchema valueOf(Label label){
+            if(THING.getLabel().equals(label)) return THING;
+            if(ENTITY.getLabel().equals(label)) return ENTITY;
+            if(ROLE.getLabel().equals(label)) return ROLE;
+            if(ATTRIBUTE.getLabel().equals(label)) return ATTRIBUTE;
+            if(RELATIONSHIP.getLabel().equals(label)) return RELATIONSHIP;
+            if(RULE.getLabel().equals(label)) return RULE;
+            return null;
+        }
```
aelred: How about using something like `isMetaLabel` above?

```
public static MetaSchema valueOf(Label label) {
    for (MetaSchema metaSchema : MetaSchema.values()) {
        if (metaSchema.getLabel().equals(label)) return metaSchema;
    }
    return null;
}
```
and bonus:
```
public static boolean isMetaLabel(Label label) {
    return valueOf(label) != null;
}
```

fppt: Good point. Not sure why I missed that

PR #2663, opened 2018-02-28
Final Getters On RemoteGraknTx

fppt:
# Why is this PR needed?

Adds missing getters to RemoteGraknTX which is needed for the GRPC client.

# What does the PR do?

Allows the user to get concepts via id and attributes via value

No

Final methods on RemoteGraknTx are still needed


grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
```diff
@@ -131,7 +133,10 @@ public Role putRole(Label label) {
     @Nullable
     @Override
     public <T extends Concept> T getConcept(ConceptId id) {
-        throw new UnsupportedOperationException(); // TODO
+        Var var = var("x");
+        VarPattern pattern = var.id(id);
+        Optional<Answer> answer = queryRunner().run(Graql.match(pattern).get(ImmutableSet.of(var))).findAny();
```
aelred: Actually it's easier just to do `match(pattern).get()` (no arguments).

grakn-client/src/test/java/ai/grakn/remote/RemoteGraknTxTest.java
```diff
@@ -350,20 +356,39 @@ public void whenPuttingRole_EnsureCorrectQueryIsSent(){
...
+        Var var = var("x");
+        ConceptId id = ConceptId.of(V123.getValue());
+        GetQuery getQuery = match(var.id(id)).get(ImmutableSet.of(var));
+        verifyCorrectQuerySent(getQuery, GraknOuterClass.BaseType.Entity, tx -> assertNotNull(tx.getConcept(id)));
+    }
+
+    @Test
+    public void whenGettingAttributesViaID_EnsureCorrectQueryIsSent(){
+        Var var = var("x");
+        String value = "Hello Oli";
```
aelred: 🐶

PR #2665, opened 2018-02-28
Direct Isa

duckofyork:
# Why is this PR needed?
So our users can query/insert/delete direct instances of a certain type, without worrying about its supertypes or subtypes (REQUEST #17333)
# What does the PR do?
Add java API and Graql syntax for direct isa
No
None


grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/DirectIsaProperty.java
```diff
@@ -0,0 +1,135 @@
...
+ * <p>
+ * When matching, any subtyping is ignored. For example, if we have {@code $bob isa man}, {@code man sub person},
+ * {@code person sub entity} then it only follows {@code $bob isa person}, not {@code bob isa entity}.
+ * </p>
+ *
+ * @author Felix Chapman
+ * @author Jason Liu
+ */
+@AutoValue
+public abstract class DirectIsaProperty extends AbstractVarProperty implements UniqueVarProperty, NamedProperty {
```
kasper-piskorski: Would it make sense to make `DirectIsaProperty` and `IsaProperty` have a common parent? They mainly differ by the `type()`/`directType()` function which is pretty much a naming difference. Only the `match` function is different to exclude the subs. Other member functions only differ by type()<->directType() change.

grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/AbstractIsaProperty.java
```diff
@@ -0,0 +1,75 @@
...
+import com.google.common.collect.ImmutableSet;
+
+import java.util.Collection;
+import java.util.stream.Stream;
+
+/**
+ * @author Jason Liu
+ */
+
+abstract class AbstractIsaProperty extends AbstractVarProperty implements UniqueVarProperty, NamedProperty {
```
aelred: I always make mistakes and override the wrong things when I make abstract classes. So I label all the methods I don't want to override as `final`:

```
public final String getProperty() {
    return type().getPrintableName();
}
```

Maybe do that here for ones you don't want to override?

grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/DirectIsaProperty.java
```diff
@@ -0,0 +1,95 @@
...
+ * Represents the {@code direct-isa} property on a {@link Thing}.
+ * <p>
+ * This property can be queried and inserted.
+ * </p>
+ * <p>
+ * THe property is defined as a relationship between an {@link Thing} and a {@link Type}.
+ * </p>
+ * <p>
+ * When matching, any subtyping is ignored. For example, if we have {@code $bob isa man}, {@code man sub person},
+ * {@code person sub entity} then it only follows {@code $bob isa person}, not {@code bob isa entity}.
```
aelred: I think the docs are not quite right! It should only follow that `$bob isa! man`.

Both `$bob isa! person` and `$bob isa entity` should be false.

grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/DirectIsaProperty.java
```diff
@@ -0,0 +1,95 @@
...
+
+        Var varName = var.var().asUserDefined();
+        VarPatternAdmin typePattern = this.type();
+        Var typeVariable = typePattern.var();
+
+        IdPredicate predicate = getIdPredicate(typeVariable, typePattern, vars, parent);
+        ConceptId predicateId = predicate != null ? predicate.getPredicate() : null;
+
+        //isa part
+        VarPatternAdmin isaVar = varName.directIsa(typeVariable).admin();
```
aelred: Since this is the only line that's different, maybe you can move `mapToAtom` into the `AbstractIsaProperty` and add a new abstract method `varPatternForAtom`:

```
@Override
protected final VarPattern varPatternForAtom(Var varName, Var typeVariable) {
    return varName.directIsa(typeVariable);
}
```

grakn-graql/src/main/java/ai/grakn/graql/internal/query/runner/ConceptBuilder.java
```diff
@@ -291,6 +292,7 @@ public String toString() {
     }
 
     private static final BuilderParam<Type> TYPE = BuilderParam.of(IsaProperty.NAME);
+    private static final BuilderParam<Type> DIRECT_TYPE = BuilderParam.of(DirectIsaProperty.NAME);
```
aelred: I think you can remove this completely! No one is using it.

grakn-graql/src/test/java/ai/grakn/graql/internal/gremlin/fragment/DirectIsaTest.java
```diff
@@ -0,0 +1,222 @@
...
+
+        getQuery = queryBuilder.parse("match $x isa! thingy1; get;");
+        assertEquals(queryBuilder.match(x.directIsa(thingy1)), getQuery.match());
+
+        getQuery = queryBuilder.parse("match $x isa! $y; get;");
+        assertEquals(queryBuilder.match(x.directIsa(y)), getQuery.match());
+    }
+
+    @Test
+    public void testInsert() {
```
aelred: You can split this test in two! I'd go for `testInsertSyntax` and `whenInsertDirectIsa_InsertsADirectInstanceOfAType`.

grakn-graql/src/test/java/ai/grakn/graql/internal/gremlin/fragment/DirectIsaTest.java
```diff
@@ -0,0 +1,222 @@
...
+        assertTrue(plan.get(0) instanceof LabelFragment);
+        assertTrue(plan.get(1) instanceof LabelFragment);
+        assertTrue(plan.get(2) instanceof LabelFragment);
+        assertTrue(plan.get(3) instanceof InIsaFragment); // start from relationship type
+        assertTrue(plan.get(4) instanceof OutRolePlayerFragment); // go to a role player with direct isa
+        assertTrue(plan.get(5) instanceof OutIsaFragment); // check the role player's type
+        assertTrue(plan.get(6) instanceof OutRolePlayerFragment); // go to the other role player
+        assertTrue(plan.get(7) instanceof NeqFragment); // check two players are different
+        assertTrue(plan.get(8) instanceof OutIsaFragment); // check the role player's type
+        assertTrue(plan.get(9) instanceof OutSubFragment); // check the subtypes
```
aelred: There's a method `Matchers.instanceOf` that will help you for this:

```
assertThat(plan, contains(
    instanceOf(LabelFragment.class),
    instanceOf(LabelFragment.class),
    instanceOf(LabelFragment.class),
    instanceOf(InIsaFragment.class),
    instanceOf(OutRolePlayerFragment.class),
    instanceOf(OutIsaFragment.class),
    instanceOf(OutRolePlayerFragment.class),
    instanceOf(NeqFragment.class),
    instanceOf(OutIsaFragment.class),
    instanceOf(OutSubFragment.class)
));
```

This will also give you nicer error messages if it goes wrong.

grakn-graql/src/test/java/ai/grakn/graql/internal/query/AdminTest.java
```diff
@@ -105,12 +105,18 @@ public void testMutateMatch() {
     public void testInsertQueryMatchPatternEmpty() {
         InsertQuery query = qb.insert(var().id(ConceptId.of("123")).isa("movie"));
         assertFalse(query.admin().match().isPresent());
+
+        query = qb.insert(var().id(ConceptId.of("123")).directIsa("movie"));
+        assertFalse(query.admin().match().isPresent());
```
aelred: Did something go wrong that meant you added this? If not, I sort of think we can remove these new tests.

grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/DirectIsaProperty.java
kasper-piskorski: :+1:

grakn-graql/src/test/java/ai/grakn/graql/internal/gremlin/fragment/DirectIsaTest.java
duckofyork: what's contains?

duckofyork:
Please re-review guys. Thanks

grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerQueryImpl.java
```diff
@@ -131,7 +131,7 @@
     }
 
     ReasonerQueryImpl(Atom atom) {
-        // TODO: This cast is unsafe - ReasonerQuery should return an EmbeddedGraknTx
+        // TODO: This cast is unsafe - ReasonerQuery should return an EmbeisRuleResolvableddedGraknTx
```
aelred: 🐐

PR #2666, opened 2018-03-01
Add getter methods to remote concepts

aelred:
# Why is this PR needed?

So you can perform operations like `type.instances()` over gRPC.

# What does the PR do?

Adds support for a lot of getters on `RemoteConcept` classes. These are mostly implemented with Graql queries.

No.

There are some operations that are difficult or impossible to implement with Graql queries. These should probably be done with dedicated gRPC messages.

### DIFF
--- a/grakn-client/src/main/java/ai/grakn/remote/concept/RemoteAttributeType.java
+++ b/grakn-client/src/main/java/ai/grakn/remote/concept/RemoteAttributeType.java
@@ -20,6 +20,7 @@ package ai.grakn.remote.concept;
 
 import ai.grakn.concept.Attribute;
 import ai.grakn.concept.AttributeType;
+import ai.grakn.concept.Concept;
 import ai.grakn.concept.ConceptId;
 import ai.grakn.grpc.ConceptProperty;
 import ai.grakn.remote.RemoteGraknTx;
@@ -64,5 +65,10 @@ abstract class RemoteAttributeType<D> extends RemoteType<AttributeType<D>, Attri
     @Override
     public final String getRegex() {
         return getProperty(ConceptProperty.REGEX);
+    }
+
+    @Override
+    final AttributeType<D> asSelf(Concept concept) {
+        return concept.asAttributeType();
     }
 }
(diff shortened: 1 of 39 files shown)


grakn-client/src/main/java/ai/grakn/remote/concept/RemoteAttributeType.java
```diff
@@ -65,4 +66,14 @@
     public final String getRegex() {
         return getProperty(ConceptProperty.REGEX);
     }
+
+    @Override
+    final AttributeType<D> asSelf(Concept concept) {
```
fppt: because `from` is a bad word :P

grakn-client/src/main/java/ai/grakn/remote/concept/RemoteConcept.java
```diff
@@ -47,10 +60,22 @@ public final void delete() throws GraknTxOperationException {
 
     @Override
     public final boolean isDeleted() {
-        return false; // TODO: implement
+        return !tx().graql().match(var().id(getId())).aggregate(ask()).execute();
```
fppt: hahahahahahah wonderful!

grakn-client/src/main/java/ai/grakn/remote/concept/RemoteSchemaConcept.java
```diff
@@ -50,17 +56,25 @@ public final Self setLabel(Label label) {
 
     @Override
     public final Self sup() {
-        throw new UnsupportedOperationException(); // TODO: implement
+        // TODO: We use a trick here because there's no "direct super" in Graql and we don't want to use gRPC for this.
```
fppt: I had to stare at this one for a while to get it. What a hack!

grakn-client/src/main/java/ai/grakn/remote/concept/RemoteAttributeType.java
aelred: I love this method btw. It's like your `getThis()` but also:
1. Type-safe! No way it can crash.
2. Because it accepts a `Concept`, you can use it to cast the result of `sup()` or `subs()`.

PR #2668, opened 2018-03-05
Fix issue when error occurs during gRPC transaction

aelred:
# Why is this PR needed?

There was a bug where a `RemoteGraknTx` would hang forever after an error occurred. Now an error will cause the transaction to close and future operations to immediately throw.

# What does the PR do?

Adds a check that will throw when the connection is terminated, instead of polling indefinitely.

Also there was a dependency issue I fixed at the same time #efficiency

No.

We may want to align the behaviour of `EmbeddedGraknTx` and `RemoteGraknTx` when an error occurs. Right now I think `EmbeddedGraknTx` will allow you to keep working. I don't know which behaviour is better: `EmbeddedGraknTx` is more permissive, but it could also cause odd bugs with broken or invalid transactions etc.

### DIFF
--- a/grakn-client/src/test/java/ai/grakn/remote/RemoteGraknTxTest.java
+++ b/grakn-client/src/test/java/ai/grakn/remote/RemoteGraknTxTest.java
@@ -25,6 +25,8 @@ import ai.grakn.concept.AttributeType;
 import ai.grakn.concept.ConceptId;
 import ai.grakn.concept.Label;
 import ai.grakn.exception.GraknBackendException;
+import ai.grakn.exception.GraknTxOperationException;
+import ai.grakn.exception.GraqlQueryException;
 import ai.grakn.exception.InvalidKBException;
 import ai.grakn.graql.DefineQuery;
 import ai.grakn.graql.GetQuery;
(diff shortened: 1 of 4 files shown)


grakn-client/src/test/java/ai/grakn/remote/RemoteGraknTxTest.java
```diff
@@ -315,6 +318,47 @@ public void whenCommittingATxFails_Throw() {
         }
     }
 
+    @Test
+    public void whenAnErrorOccurs_TheTxCloses() {
```
lolski: 👍 test which describes the behavior

grakn-client/src/test/java/ai/grakn/remote/RemoteGraknTxTest.java
```diff
@@ -315,6 +318,47 @@ public void whenCommittingATxFails_Throw() {
...
+            } catch (GraqlQueryException e) {
+                // Ignore
+            }
+
+            assertTrue(tx.isClosed());
+        }
+    }
+
+    @Test
+    public void whenAnErrorOccurs_AllFutureActionsThrow() {
```
lolski: 👍

PR #2669, opened 2018-03-05
remove the ability of overriding GRAKN_HOME

lolski:
# Why is this PR needed?
There's nothing good out of being able to override GRAKN_HOME. By definition, GRAKN_HOME should always points to, well, Grakn's home directory

# What does the PR do?
1. Remove the clause which says "if GRAKN_HOME is defined from outside, use that instead"

No

N/A

### DIFF
--- a/grakn-dist/src/grakn
+++ b/grakn-dist/src/grakn
@@ -18,10 +18,8 @@
 # <http://www.gnu.org/licenses/gpl.txt>.
 
 # Grakn global variables
-if [ -z "${GRAKN_HOME}" ]; then
-    [[ $(readlink $0) ]] && path=$(readlink $0) || path=$0
-    GRAKN_HOME=$(cd "$(dirname "${path}")" && pwd -P)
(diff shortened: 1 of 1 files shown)


grakn-dist/src/grakn
```diff
@@ -18,10 +18,8 @@
 # <http://www.gnu.org/licenses/gpl.txt>.
 
 # Grakn global variables
-if [ -z "${GRAKN_HOME}" ]; then
-    [[ $(readlink $0) ]] && path=$(readlink $0) || path=$0
-    GRAKN_HOME=$(cd "$(dirname "${path}")" && pwd -P)
-fi
+[[ $(readlink $0) ]] && path=$(readlink $0) || path=$0
+GRAKN_HOME=$(cd "$(dirname "${path}")" && pwd -P)
```
kasper-piskorski: so this basically always overrides instead of doing that only if empty?

PR #2671, opened 2018-03-06
Support all methods on remote Java client

aelred:
# Why is this PR needed?

Because we need a fully-featured Java client!

# What does the PR do?

Supports all the methods on the Java client, some using Graql queries, some using pure gRPC.

This PR also has a fairly significant refactor, mainly changing `ConceptProperty` into `ConceptMethod` and splitting out `ConceptResponseType`.

No.

Change everything to use real gRPC.


aelred:
@marco-scoppetta this is blocking my work this week, so I'm going to merge it with one approval. If you have time, I would still appreciate a review when you're back!

duckofyork:
👍

PR #2673, opened 2018-03-07
Change Graql shell to use gRPC

aelred:
# Why is this PR needed?

We don't want to support the websocket API. This is also a good test of the gRPC client.

# What does the PR do?

Switches the shell from websockets to `RemoteGrakn`.

Also, completely deletes the websocket server from engine!

No. Although, old versions of the Graql shell will not be able to connect to this server.

A general refactor of the shell is still overdue and it should be much easier now.


grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
```diff
@@ -222,10 +208,17 @@ public static boolean runShell(String[] args, String version, String historyFile
         }
 
         Keyspace keyspace = Keyspace.of(cmd.getOptionValue("k", DEFAULT_KEYSPACE));
-        SimpleURI location = Optional.ofNullable(cmd.getOptionValue("r")).map(SimpleURI::new).orElse(Grakn.DEFAULT_URI);
+
+        int defaultGrpcPort = 48555;
```
marco-scoppetta: Should we this value it from config file? Maybe not now, but we will probably need a `client-grakn.properties` at some point so that we can expose only a subset of settings to the client's user.

grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
```diff
@@ -222,10 +208,17 @@ public static boolean runShell(String[] args, String version, String historyFile
...
+        SimpleURI defaultGrpcUri = new SimpleURI(Grakn.DEFAULT_URI.getHost(), defaultGrpcPort);
+
+        Optional<SimpleURI> location = Optional.ofNullable(cmd.getOptionValue("r")).map(SimpleURI::new);
         String outputFormat = cmd.getOptionValue("o", DEFAULT_OUTPUT_FORMAT);
 
-        if (!client.serverIsRunning(location)) {
+        SimpleURI httpUri = location.orElse(Grakn.DEFAULT_URI);
+        SimpleURI grpcUri = location.orElse(defaultGrpcUri);
```
marco-scoppetta: Question: `location` can be something like `localhost:48555` if specified via `-r` so then `httpUri` would become `localhost:48555`. Basically if I specify a uri via `-r` then `httpURI==grpcUri`, is that intended? What is the difference between the two then?

grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
```diff
@@ -502,75 +487,31 @@ private void printLicense() {
...
-                throw new RuntimeException("Unrecognized message: " + message);
-        }
     }
 
     private void setDisplayOptions(Set<String> displayOptions) throws IOException {
-        session.sendJson(Json.object(
-                ACTION, ACTION_DISPLAY,
-                DISPLAY, displayOptions
-        ));
+        displayAttributes = displayOptions.stream().map(tx::getAttributeType).collect(toImmutableSet());
```
marco-scoppetta: Curiosity: what does this method do? The name is maybe a bit vague, but I am sure it's just because I don't have a context, so I am just curious.

grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
aelred: Maybe we can get it from the config file! I'll try it and see.

grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
aelred: One of them is the HTTP port for our REST API (so 4567 by default), the other is for gRPC (so 48555 by default).

I don't know if that behaviour is OK, I'll have to check.

grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
aelred: Update: it's not OK because we use that URI to check if the server is running or not, which must be over HTTP.

I don't know what to do about that :(

grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
aelred: Update: it's not OK because we use that URI to check if the server is running or not, which must be over HTTP.

I don't know what to do about that :(

grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
aelred: It lets you display attributes on the results:
```
>>> display name
>>> match $x isa person;
$x isa person has name "Alice";
$x isa person has name "Bob";
...
```

Very handy! No one uses it.

aelred:
@marco-scoppetta I think I solved all the things

grakn-graql-shell/src/main/java/ai/grakn/graql/GraqlShell.java
marco-scoppetta: 🥇 👍 💯

aelred:
`test-biomed` is failing because `Relation#allRolePlayers` is very slow. My [other PR](https://github.com/graknlabs/grakn/pull/2671) ought to solve the issue.

PR #2675, opened 2018-03-07
Re-usable RemoteGraknTx

marco-scoppetta:
# Why is this PR needed?

To make `RemoteGraknTx` reusable in KBMS

# What does the PR do?

Makes `RemoteGraknSession` extendible.
`GrpcClient.open()` now accepts an already built `TxRequest openRequest`, so that `RemoteGraknSession` or any subclass of it can build the request using different fields, i.e. in KBMS we extend the class and build an openRequest using also username and password. 

no

We might want to make `RemoteGraknTx.create()` method accept an interface instead of a `RemoteGraknSession` instance so that in KBMS we can use `RemoteGraknSession` with delegation instead of inheritance


grakn-client/src/main/java/ai/grakn/remote/RemoteGraknTx.java
```diff
@@ -82,11 +84,11 @@ private RemoteGraknTx(GraknSession session, GraknTxType txType, GrpcClient clien
         this.client = client;
     }
 
-    static RemoteGraknTx create(RemoteGraknSession session, GraknTxType txType) {
+    public static RemoteGraknTx create(RemoteGraknSession session, TxRequest openRequest) {
```
aelred: Maybe a TODO about saying this should accept an interface? Or not pass in the session at all.

PR #2676, opened 2018-03-08
use commitSubmitNoLogs() instead of commit()

lolski:
# Why is this PR needed?
The `commit()` in Remote Java Client calls `EmbeddedGraknTx`'s `commit()` instead of `commitSubmitNoLogs()`.

This won't work as the Remote Java Client communicates with an `EmbeddedGraknSession` instance configured with the `remoteSubmissionNeeded == false`. See: https://github.com/graknlabs/grakn/blob/master/grakn-engine/src/main/java/ai/grakn/engine/factory/EngineGraknTxFactory.java#L102

# What does the PR do?
1. Call `commitSubmitNoLogs()` instead of `commit()`

*Yes*: KGMS dependency will need to be updated. This should not be too hard though.

N/A


grakn-engine/src/main/java/ai/grakn/engine/GraknEngineServerFactory.java
```diff
@@ -66,21 +66,20 @@ public static GraknEngineServer createGraknEngineServer() throws IOException {
...
     public static GraknEngineServer createGraknEngineServer(
             EngineID engineID, Service sparkService, GraknEngineStatus graknEngineStatus, MetricRegistry metricRegistry, GraknConfig graknEngineConfig,
             RedisWrapper redisWrapper, IndexStorage indexStorage, CountStorage countStorage, LockProvider lockProvider,
-            Runtime runtime, Collection<HttpController> collaborators, EngineGraknTxFactory factory, GrpcServer grpcServer) {
+            Runtime runtime, Collection<HttpController> collaborators, EngineGraknTxFactory factory) {
 
         IndexPostProcessor indexPostProcessor = IndexPostProcessor.create(lockProvider, indexStorage);
         CountPostProcessor countPostProcessor = CountPostProcessor.create(graknEngineConfig, factory, lockProvider, metricRegistry, countStorage);
         PostProcessor postProcessor = PostProcessor.create(indexPostProcessor, countPostProcessor);
+        GrpcServer grpcServer = configureGrpcServer(graknEngineConfig, factory, postProcessor);
```
marco-scoppetta: Why are we instantiating the gRPC server here instead of taking it as a parameter? I think we should keep it as a param, so that we can reuse this method also in KGMS - passing an authenticated grpc server

PR #2677, opened 2018-03-08
Storage config settable from grakn properties

kasper-piskorski:
# Why is this PR needed?
* to be able to overwrite storage params from grakn properties
* to be able to change storage db dir
# What does the PR do?
Introduce extra `StorageConfig` and `StorageConfigProcessor` classes for manipulating storage config.
No
Changeable logdir.


grakn-bootup/src/main/java/ai/grakn/bootup/StorageConfigProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ *
+ * @author Kasper Piskorski
+ */
+public class StorageConfigProcessor {
+
+    private static final String CONFIG_PATH = "services/cassandra/";
+    private static final String CONFIG_NAME = "cassandra.yaml";
+
```
marco-scoppetta: Let's try to find a way to inject this path

grakn-bootup/src/main/java/ai/grakn/bootup/StorageConfigProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+    public static void updateConfig() {
```
marco-scoppetta: let's get Grakn config as a parameter in this method

grakn-bootup/src/main/java/ai/grakn/bootup/StorageConfig.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @author Kasper Piskorski
+ */
+public class StorageConfig {
+
+    private final ImmutableMap<String, Object> params;
+    private static final String EMPTY_VALUE = "";
+    private static final String CONFIG_PARAM_PREFIX = "storage.internal.";
+
+    private StorageConfig(Map<String, Object> p){
```
marco-scoppetta: Rename `p` to something similar to `yamlParams`

grakn-bootup/src/main/java/ai/grakn/bootup/StorageConfig.java
```diff
@@ -0,0 +1,103 @@
...
+        this.params = ImmutableMap.copyOf(
+                p.entrySet().stream()
+                        .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue() == null ? EMPTY_VALUE : e.getValue()))
```
marco-scoppetta: Try to unify this operation with `mapper.readValue()` in a method called something like `parseStringtoMap`

grakn-bootup/src/main/java/ai/grakn/bootup/StorageConfig.java
```diff
@@ -0,0 +1,103 @@
...
+                "commitlog_directory", dbDir + "cassandra/commitlog"
+        );
+        Map<String, Object> updatedParams = Maps.newHashMap(params);
+        dataParams.keySet().stream()
+                .filter(updatedParams::containsKey)
+                .forEach(dp -> updatedParams.put(dp, dataParams.get(dp)));
+        return new StorageConfig(updatedParams);
+    }
+
+    public StorageConfig update(GraknConfig config){
```
marco-scoppetta: rename with something like `updateYamlFromConfig`

grakn-bootup/src/main/java/ai/grakn/bootup/StorageConfigProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+            yamlString = new String(bytes, StandardCharsets.UTF_8);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+
+        GraknConfig graknConfig = GraknConfig.create();
+        String updatedYamlString = StorageConfig.of(yamlString)
+                .update(graknConfig)
+                .updateDataParams(graknConfig)
+                .toYaml();
```
marco-scoppetta: `toString()` maybe or similar

grakn-bootup/src/main/java/ai/grakn/bootup/StorageConfigProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @author Kasper Piskorski
+ */
+public class StorageConfigProcessor {
+
+    private static final String CONFIG_PATH = "services/cassandra/";
+    private static final String CONFIG_NAME = "cassandra.yaml";
+
+    public static void updateConfig() {
+        Path configPath = Paths.get(CONFIG_PATH, CONFIG_NAME);
```
marco-scoppetta: rename to `storageConfigPath`

conf/main/grakn.properties
```diff
@@ -100,6 +100,8 @@ log.dirs=./logs/
 log.level=INFO
 
 ############################# Persisted Knowledge Base Configuration #############################
+# Directory in which the data will be persisted
+storage.db-dir=db/
```
lolski: I think this should just be`db-dir`, meaning that we have 1 directory setting for everything.

For example, if people set `db.dir=/opt/grakn-data`, Cassandra data would be in `/opt/grakn-data/cassandra` (and in the future Redis data would be in `/opt/grakn-data/redis`

kasper-piskorski:
@lolski  rereview please

grakn-bootup/src/main/java/ai/grakn/bootup/StorageConfig.java
```diff
@@ -0,0 +1,118 @@
...
+            ByteArrayOutputStream outputstream = new ByteArrayOutputStream();
+            mapper.writeValue(outputstream, yamlParams);
+            return outputstream.toString(StandardCharsets.UTF_8.name());
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private StorageConfig updateDataDirs(GraknConfig config) {
+        String dbDir = config.properties().containsKey(GraknConfigKey.DATA_DIR)?
```
lolski: In my opinion, we should avoid having this logic of falling back to some default value. We should just let the program crash if the user mistakenly delete the `db-dir` property.

While this behavior is nice, it can be a little inconsistent since the user may then expect that other parameters will fall back to some default value if removed as well. And that might not always be the case.

grakn-bootup/src/main/java/ai/grakn/bootup/StorageConfig.java
```diff
@@ -0,0 +1,116 @@
...
+        this.yamlParams = ImmutableMap.copyOf(yamlParams);
+    }
+
+    private static Map<String, Object> parseStringToMap(String yaml){
+        ObjectMapper mapper = new ObjectMapper(new YAMLFactory().enable(YAMLGenerator.Feature.MINIMIZE_QUOTES));
+        try {
+            TypeReference<Map<String, Object>> reference = new TypeReference<Map<String, Object>>(){};
+            Map<String, Object> yamlParams = mapper.readValue(yaml, reference);
+            return yamlParams.entrySet().stream()
+                    .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue() == null ? EMPTY_VALUE : e.getValue()));
```
aelred: I think there's a Guava method for transforming map values:
```
return Maps.transformValues(yamlParams, value -> value == null ? EMPTY_VALUE : value);
```

grakn-bootup/src/main/java/ai/grakn/bootup/StorageConfig.java
```diff
@@ -0,0 +1,116 @@
...
+
+        ImmutableMap<String, Object> dataParams = ImmutableMap.of(
+                "data_file_directories", Collections.singletonList(dbDir + DATA_SUBDIR),
+                "saved_caches_directory", dbDir + SAVED_CACHES_SUBDIR,
+                "commitlog_directory", dbDir + COMMITLOG_SUBDIR
+        );
+        Map<String, Object> updatedParams = Maps.newHashMap(yamlParams);
+        dataParams.keySet().stream()
+                .filter(updatedParams::containsKey)
+                .forEach(dp -> updatedParams.put(dp, dataParams.get(dp)));
```
aelred: This might be the same:
```
updatedParams.putAll(dataParams);
```
unless the `containsKey` check is important!

grakn-bootup/src/main/java/ai/grakn/bootup/StorageConfig.java
```diff
@@ -0,0 +1,116 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ *
+ * @author Kasper Piskorski
+ */
+public class StorageConfig {
```
aelred: This and `StorageConfigProcessor` could use some Javadoc to say what they're for 👍

grakn-bootup/src/main/java/ai/grakn/bootup/StorageConfig.java
```diff
@@ -0,0 +1,116 @@
...
+                .stringPropertyNames()
+                .stream()
+                .filter(prop -> prop.contains(CONFIG_PARAM_PREFIX))
+                .forEach(prop -> {
+                    String param = prop.replaceAll(CONFIG_PARAM_PREFIX, "");
+                    if (updatedParams.containsKey(param)) {
+                        updatedParams.put(param, config.properties().getProperty(prop));
+                    }
+                });
+        return new StorageConfig(updatedParams);
```
aelred: So immutable 😍

PR #2678, opened 2018-03-08
Provide "as" syntax for overriding roles

duckofyork:
# Why is this PR needed?
```parenthood sub relationship, relates parent, relates child;
fatherhood sub parenthood, relates father as parent, relates son as child;
=
fatherhood sub parenthood, relates father, relates son;
father sub parent;
son sub child;
 ```
This can eventually allow us to remove sub'ing roles entirely.
https://work.grakn.ai/entity/19005-provideas-syntax-for-overriding-roles
# What does the PR do?
- Provide `as` syntax for overriding roles
- Update documentation
No
N.A

### DIFF
--- a/docs/pages/docs/9-api-references/ddl.md
+++ b/docs/pages/docs/9-api-references/ddl.md
@@ -56,6 +56,10 @@ one will be created.
 
 In the case where `$B` does not have a defined `sub`, this will also implicit define `$B sub role`.
 
+## as
+
+`$A relates $C as $D` is equivalent to `$A relates $C; $C subs $D`. It can can only be used together with `relates`. 
+
 ## plays
 
 `$A plays $B` defines the _type_ `$A` to directly play the _role_ `$B`.
--- a/docs/pages/docs/9-api-references/dml.md
+++ b/docs/pages/docs/9-api-references/dml.md
@@ -47,6 +47,10 @@ Within the [patterns](#pattern), the following [properties](../querying-data/ove
 
 `$x isa $A` is _satisfied_ if `$x` is indirectly an _instance_ of the _type_ `$A`.
 
+### isa!
+
+`$x isa! $A` is _satisfied_ if `$x` is directly an _instance_ of the _type_ `$A`.
+
 ### relationship
 
 `$r ($A1: $x1, ..., $An: $xn)` is _satisfied_ if `$r` is a _relation_ where for each `$Ai: $xi`, `$xi` is a role-player
@@ -215,6 +219,10 @@ supported:
 ## isa
 
 `$x isa $A` creates a new direct instance of the _type_ `$A` and binds it to `$x`.
+
+## isa!
+
+`$x isa! $A` creates a new direct instance of the _type_ `$A` and binds it to `$x`.
 
 ## relationship
 
--- a/grakn-core/src/main/java/ai/grakn/graql/VarPattern.java
+++ b/grakn-core/src/main/java/ai/grakn/graql/VarPattern.java
@@ -89,7 +89,7 @@ public interface VarPattern extends Pattern {
     /**
      * the variable must have a resource of the given type with an exact matching value
      *
-     * @param type a resource type in the schema
+     * @param type  a resource type in the schema
      * @param value a value of a resource
      * @return this
      */
@@ -99,7 +99,7 @@ public interface VarPattern extends Pattern {
     /**
      * the variable must have a resource of the given type that matches the given atom
      *
-     * @param type a resource type in the schema
+     * @param type      a resource type in the schema
      * @param predicate a atom on the value of a resource
      * @return this
      */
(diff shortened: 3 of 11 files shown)

### CODE REVIEW

File: grakn-core/src/main/java/ai/grakn/graql/VarPattern.java [hunk 1]
```diff
@@ -180,6 +180,20 @@
     @CheckReturnValue
     VarPattern relates(VarPattern type);
 
+    /**
+     * @param roleType a {@link Role} id that this relation type variable must have
+     * @return this
+     */
+    @CheckReturnValue
+    VarPattern relates(String roleType, String superRoleType);
```
aelred: Since the argument can be null, we can say: `..., @Nullable String superRoleType);`

(and same for below)

grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/RelatesProperty.java
```diff
@@ -33,45 +33,64 @@
...
  * This property relates a {@link RelationshipType} and a {@link Role}. It indicates that a {@link Relationship} whose
  * type is this {@link RelationshipType} may have a role-player playing the given {@link Role}.
  *
  * @author Felix Chapman
  */
 @AutoValue
-public abstract class RelatesProperty extends AbstractVarProperty implements NamedProperty {
+public abstract class RelatesProperty extends AbstractVarProperty {
+
+    public static RelatesProperty of(VarPatternAdmin role, VarPatternAdmin superRole) {
```
aelred: Another `@Nullable` here!

grakn-graql/src/main/java/ai/grakn/graql/internal/pattern/property/RelatesProperty.java
```diff
@@ -96,12 +115,23 @@ public String getProperty() {
...
+                Role superRole = executor.get(superRoleVar).asRole();
+                executor.builder(roleVar).sub(superRole);
+            };
+
+            PropertyExecutor subExecutor = PropertyExecutor.builder(subMethod)
+                    .requires(superRoleVar).produces(roleVar).build();
+
+            return ImmutableSet.of(relatesExecutor, isRoleExecutor, subExecutor);
+        }
         return ImmutableSet.of(relatesExecutor, isRoleExecutor);
```
aelred: It might be a bit neater to make this an `if ... else`:
```
if (superRoleVarPattern != null) {
    ...
    return ImmutableSet.of(relatesExecutor, isRoleExecutor, subExecutor);
} else {
    return ImmutableSet.of(relatesExecutor, isRoleExecutor);
}
```

but maybe it's just personal preference 😛

grakn-graql/src/main/java/ai/grakn/graql/internal/parser/QueryVisitor.java
```diff
@@ -536,7 +536,8 @@ public VarPattern visitVarPattern(GraqlParser.VarPatternContext ctx) {
 
     @Override
     public UnaryOperator<VarPattern> visitRelates(GraqlParser.RelatesContext ctx) {
-        return var -> var.relates(visitVariable(ctx.variable()));
+        if (ctx.superRole == null) return var -> var.relates(visitVariable(ctx.role));
+        return var -> var.relates(visitVariable(ctx.role), visitVariable(ctx.superRole));
```
aelred: We could take advantage of the fact the argument can be null:
```
VarPattern superRole = ctx.superRole != null ? visitVariable(ctx.superRole) : null;
return var -> var.relates(visitVariable(ctx.role), superRole);
```

duckofyork:
Should I submit another pr for the doc or do you usually just do it in one pr? @aelred

aelred:
@duckofyork it's up to you. I do it in one PR because then it's all "one thing". And I don't forget to do it 👍

docs/pages/docs/9-api-references/ddl.md
```diff
@@ -56,6 +56,10 @@ one will be created.
 
 In the case where `$B` does not have a defined `sub`, this will also implicit define `$B sub role`.
 
+## as
+
+`$A relates $C as $D` is equivalent to `$A relates $C; $C subs $D`. It can can only be used together with `relates`.
```
aelred: Since this is really part of `relates`, it might be simpler to just remove the `## as` header.

```
$C subs $D
```
should be:
```
$C sub $D
```

docs/pages/docs/9-api-references/dml.md
```diff
@@ -216,6 +220,10 @@ supported:
 
 `$x isa $A` creates a new direct instance of the _type_ `$A` and binds it to `$x`.
 
+## isa!
+
+`$x isa! $A` creates a new direct instance of the _type_ `$A` and binds it to `$x`.
+
```
aelred: 👍

PR #2679, opened 2018-03-08
Adjustable storage log dir

kasper-piskorski:
# Why is this PR needed?
Storage log dir is hardcoded.
# What does the PR do?
Add extra param to adjust log dir.
No
No


kasper-piskorski:
For queue we would need another class to handle the config similarly to `StorageConfig`. I haven't touched the queue cause I'm not sure what are the plans for it.

PR #2683, opened 2018-03-12
Fix a bug with toString on equals predicate

aelred:
# Why is this PR needed?

There was a bug where `query.toString()` would create a string that was not a valid query (in that it could not be parsed again with `Graql.parse(string)`).

# What does the PR do?

Fixes the bug! And also refactors the predicates because they were a bit messy, which probably caused the bug in the first place.

No.

None. All the bugs are fixed forever now.


duckofyork:
I like the commit msgs in this pr @aelred

PR #2684, opened 2018-03-12
disable janus cache

lolski:
# Why is this PR needed?
Caching would lead to stale result when performing a query, which is unwanted.

My stance on performance optimisation (such as caching) is to enable them only when we have a clear benchmark showing that the said optimisation greatly improves the experience.

With this in mind I am disabling the cache which may cause staleness issue in distributed systems.

# What does the PR do?
Disable Janus cache

No 

N/A

### DIFF
--- a/conf/main/grakn.properties
+++ b/conf/main/grakn.properties
@@ -112,7 +112,7 @@ storage.connection-timeout=20000
 # Enabling this option speeds up traversals by holding hot elements in memory,
 # but also increases the likelihood of reading stale data. Disabling it forces each transaction
 # to independently fetch elements from storage before reading/writing them.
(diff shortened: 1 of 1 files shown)


aelred:
Actually I'm wrong! The tests are much slower here because we don't have the history to run them in parallel.

Now we're running Jenkins a second time. If the tests go in parallel, we will have comparable numbers.

lolski:
this is interesting. ok, trying that out now

lolski:
@aelred the [test we had just run](https://jenkins.grakn.ai/blue/organizations/jenkins/grakn/detail/PR-2684/2/pipeline) with paralellism yielded similar results: 1h 40m 35s

PR #2686, opened 2018-03-13
Fix an issue with port allocation in tests

aelred:
# Why is this PR needed?

The way we allocate unused ports within tests is not perfect:

1. Ask for a socket to an unused port
2. Record the port and close the socket
3. Write the port into the config file
4. Start up gRPC/spark/redis with the port

After the socket is closed, there is a potential race condition: something else can jump in and steal the port before we start a service using it.

We see this occasionally (when we get tests saying "address already in use"). It's not super-common but it has been a problem for a long time.

# What does the PR do?

Eliminates this race condition. We don't generate ports ourselves but instead tell gRPC and Redis to use port 0, meaning "pick your own port!". Then we write the port they choose into the config file. 

Spark is not so easy, because they don't report the port they choose correctly. For them, we just minimise the risk of the race condition.

No.

If we switch from Spark, we can avoid that race condition too.

### DIFF
--- a/grakn-engine/src/test/java/ai/grakn/engine/controller/SparkContext.java
+++ b/grakn-engine/src/test/java/ai/grakn/engine/controller/SparkContext.java
@@ -31,10 +31,8 @@ import org.junit.rules.TemporaryFolder;
 import org.junit.rules.TestRule;
 import spark.Service;
 
-import java.util.Collection;
 import java.util.List;
 import java.util.function.BiConsumer;
-import java.util.function.Consumer;
 import java.util.stream.Stream;
 
 import static ai.grakn.engine.HttpHandler.configureSpark;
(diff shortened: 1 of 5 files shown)


grakn-engine/src/test/java/ai/grakn/engine/controller/SparkContext.java
```diff
@@ -54,21 +52,24 @@
     private SparkContext(BiConsumer<Service, GraknConfig> createControllers) {
         this.createControllers = createControllers;
 
-        int port = GraknTestUtil.getEphemeralPort();
-        config.setConfigProperty(GraknConfigKey.SERVER_PORT, port);
-
         if ("0.0.0.0".equals(config.getProperty(GraknConfigKey.SERVER_HOST_NAME))) {
```
lolski: is there any case where the server is not localhost? i'm inclined to remove the "if" and just do `config.setConfigProperty(GraknConfigKey.SERVER_HOST_NAME, "localhost");` straight up

lolski: @aelred i am merging this for now. am i making sense though, w.r.t removing the "if"?

PR #2692, opened 2018-03-14
Queue config override and config cleanup

kasper-piskorski:
# Why is this PR needed?
- to make log and db dirs consistent
- to add ability to override queue config from grakn.properties

# What does the PR do?
Adds an extra QueueConfig class allowing for storage and manipulation of parameters.

No.
None?


kasper-piskorski:
@aelred really? no comments ;p?

aelred:
@kasper-piskorski now I feel like I should look again :P

kasper-piskorski:
@marco-scoppetta centralised the config change

Issue #2367, opened 2017-11-12
[Bug] Wrong data type


idealley:
I declared a datatype date in a `.gql` file. Then I have imported the schema into grakn `0.17.1` then I did a file with few insert queries and I thought I declared a timestamp (as a long) and used a dummy value `987654321` I got the following error:

```
$ ./graql console -f ~/Projects/grakn-gdpr/schema/queries.gql
The value [in one year] must be of datatype [java.lang.Long]
``` 

The error message should have probably been `[java.util.Date]`

aelred:
Thanks for submitting this issue! I've made a PR to fix it.

PR #2707, opened 2018-03-19
Fix incorrect error message regarding datatypes

aelred:
# Why is this PR needed?

Fixes issue #2367 

# What does the PR do?

Fixes up the error message to report the user-facing type of the attribute (not the type that is persisted).

No.

None.


aelred:
I'm hoping if I re-open it everything will work again

PR #2708, opened 2018-03-19
add java-opts

lolski:
# Why is this PR needed?
This will allow you to add Java options before starting engine.

For example: `ENGINE_JAVAOPTS='-Xdebug ...' ./grakn server start`

# What does the PR do?
`grakn` command now accepts `ENGINE_JAVAOPTS`, which will be passed down to `GraknBootup` via `-Dengine.javaopts`

No

N/A

### DIFF
--- a/grakn-bootup/src/main/java/ai/grakn/bootup/EngineProcess.java
+++ b/grakn-bootup/src/main/java/ai/grakn/bootup/EngineProcess.java
@@ -23,6 +23,7 @@ import ai.grakn.bootup.graknengine.Grakn;
 import ai.grakn.engine.GraknConfig;
 import ai.grakn.util.REST;
 import ai.grakn.util.SimpleURI;
+import org.slf4j.LoggerFactory;
 
 import javax.ws.rs.core.UriBuilder;
 import java.io.File;
@@ -35,6 +36,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.LocalDateTime;
 import java.util.Comparator;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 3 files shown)


grakn-bootup/src/main/java/ai/grakn/bootup/EngineProcess.java
```diff
@@ -47,6 +48,7 @@
     private static final String GRAKN_NAME = "Grakn";
     private static final long GRAKN_STARTUP_TIMEOUT_S = 300;
     public static final Path ENGINE_PID = Paths.get(File.separator,"tmp","grakn-engine.pid");
+    public static final Optional<String> javaOpts = Optional.ofNullable(System.getProperty("engine.javaopts"));
```
marco-scoppetta: this could be `String javaOpts = Optional.ofNullable(System.getProperty("engine.javaopts")).orElse("")` so we don't carry around Optionals

grakn-dist/src/grakn
```diff
@@ -49,6 +49,6 @@ update_classpath_global_var
 
 GRAKN_CONFIG="conf/grakn.properties"
 
-java -cp ${CLASSPATH} -Dgrakn.dir="${GRAKN_HOME}" -Dgrakn.conf="${GRAKN_HOME}/${GRAKN_CONFIG}" -Dgrakn.pidfile=/tmp/grakn.pid ai.grakn.bootup.GraknBootup $@
+java -cp ${CLASSPATH} -Dgrakn.dir="${GRAKN_HOME}" -Dgrakn.conf="${GRAKN_HOME}/${GRAKN_CONFIG}" -Dengine.javaopts="${ENGINE_JAVAOPTS}" -Dgrakn.pidfile=/tmp/grakn.pid ai.grakn.bootup.GraknBootup $@
```
marco-scoppetta: where is ENGINE_JAVAOPTS set in the script?

lolski: It is supplied by the user when invoking grakn via the command line, ie.,:
```
$ ENGINE_JAVAOPTS='-Xdebug ...' ./grakn server start
Starting Storage.....
```

grakn-bootup/src/main/java/ai/grakn/bootup/EngineProcess.java
aelred: There's a class `GraknSystemProperty` where we put all our system properties. This could go there too.

We could even add a new method there like `getOrDefault`, or one that returns an optional 👍

lolski: fixed 👍

PR #2709, opened 2018-03-20
Remove SystemKeyspace from EngineTxFactory

marco-scoppetta:
# Why is this PR needed?

So that we inject the `GraknKeyspaceStore` (previously called `SystemKeyspace`) down to Engine.
This way we can customise the way we access the `graknsystem` keyspace.
This is necessary because `GraknKeyspaceStore` is the only part in Grakn where we access the database without a direct input from the user, therefore this was causing problem when enabling authentication on the database.

# What does the PR do?

 rename `SystemKeyspace` to `GraknKeyspaceStore`, given that this class actually represents a set of available keyspaces.

Small refactoring to code an tests - nothing major

No

A couple of TODOs

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/GraknKeyspaceStoreImpl.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/GraknKeyspaceStoreImpl.java
@@ -62,30 +62,15 @@ public class GraknKeyspaceStoreImpl implements GraknKeyspaceStore {
         return new GraknKeyspaceStoreImpl(session);
     }
 
-    @Override
-    public boolean putKeyspace(Keyspace keyspace) {
-        //Check the local cache to see which keyspaces we already have open
-        if(existingKeyspaces.contains(keyspace)){
-             return false;
-        }
-
-        //If the cache does not contain the keyspace check the persisted data
-        if(containsKeyspace(keyspace)){
-            existingKeyspaces.add(keyspace);
-            return false;
-        }
-
-        return true;
-    }
-
     /**
      * Logs a new {@link Keyspace} to the {@link GraknKeyspaceStore}.
      *
      * @param keyspace The new {@link Keyspace} we have just created
      */
     @Override
-    public void persistNewKeyspace(Keyspace keyspace){
-        //Log that we have created the keyspace
+    public void addKeyspace(Keyspace keyspace){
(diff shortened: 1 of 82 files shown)


grakn-engine/src/main/java/ai/grakn/engine/GraknKeyspaceStoreImpl.java
```diff
@@ -108,19 +80,25 @@ private void logNewKeyspace(Keyspace keyspace){
...
-        return "/creating-new-keyspace-lock/" + keyspace.getValue();
-    }
-
     @Override
     public boolean containsKeyspace(Keyspace keyspace){
-        try (GraknTx graph = factory.tx(SYSTEM_KB_KEYSPACE, GraknTxType.READ)) {
-            return graph.getAttributeType(KEYSPACE_RESOURCE.getValue()).getAttribute(keyspace) != null;
+        //Check the local cache to see which keyspaces we already have open
+        if(existingKeyspaces.contains(keyspace)){
+            return true;
```
aelred: 👍

grakn-engine/src/main/java/ai/grakn/engine/SystemKeyspaceSessionProvider.java
```diff
@@ -0,0 +1,53 @@
...
+import ai.grakn.kb.internal.EmbeddedGraknTx;
+
+/**
+ * Provides transactions to "graknsystem" which is Grakn system keyspace - used by GraknKeyspaceStore to keep track
+ * of available keyspaces
+ *
+ * @author Marco Scoppetta
+ */
+
+public class SystemKeyspaceSessionProvider implements SystemKeyspaceSession {
```
aelred: Calling this `_Provider` confused me. I thought it was a factory :(

I don't know what a good name is though. `SystemKeyspaceSessionImpl` is a bit silly 🤷‍♂️

marco-scoppetta: @aelred, maybe `GraknSystemKeyspaceSession` ?

aelred: 👍 
Makes sense, and that's the convention you used before I think.

PR #2711, opened 2018-03-20
Fix bug where migrator says "Loaded 0 statements"

aelred:
# Why is this PR needed?

There was a bug where when migrating small files, the migrator would report "Loaded 0 statements".
This fixes that issue.

# What does the PR do?

This sequence of events caused the issue:

1. `BatchExecutorClient` opens
2. `Migrator` registers callback on `BatchExecutorClient` for completed queries
3. `Migrator` submits one query to `BatchExecutorClient`
4. `Migrator` waits for `BatchExecutorClient` to close (waits for all permits to release)
5. On thread 2: The query completes and releases a permit
6. `BatchExecutorClient` closes, `Migrator` continues and prints "Loaded 0 statements"
7. On thread 2: Callback is called, incrementing number of loaded queries

Now we optimistically increment the `queriesLoaded` number when the query is submitted, not when it completes.

No.

None.

### DIFF
--- a/grakn-migration/base/src/main/java/ai/grakn/migration/base/Migrator.java
+++ b/grakn-migration/base/src/main/java/ai/grakn/migration/base/Migrator.java
@@ -113,7 +113,7 @@ public class Migrator {
     public void load(String template, Stream<Map<String, Object>> data) {
         GraknClient graknClient = GraknClient.of(uri);
 
-        AtomicInteger queriesExecuted = new AtomicInteger(0);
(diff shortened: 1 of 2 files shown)


duckofyork:
👍👍👍

aelred:
I think there might still be an issue with this solution: if a query fails to load, it will still increment the "Loaded queries" number.

I can't think how to solve this particular issue :(

aelred:
In fact, I'm closing this because I've found a better way 😛

PR #2713, opened 2018-03-20
Skip post processing if system keyspace isn't ready

lolski:
# Why is this PR needed?
Handle a scenario where post-processing kicks in before system keyspace is ready

# What does the PR do?
1. Skip post processing when system keyspace isn't ready
2. Add a bunch of post-processing logs

No

N/A

### DIFF
--- a/grakn-engine/src/main/java/ai/grakn/engine/task/postprocessing/PostProcessingTask.java
+++ b/grakn-engine/src/main/java/ai/grakn/engine/task/postprocessing/PostProcessingTask.java
@@ -98,9 +98,13 @@ public class PostProcessingTask implements BackgroundTask{
      */
     public void processIndex(Keyspace keyspace, String index, UUID executionId){
         Set<ConceptId> ids = indexPostProcessor.popIds(keyspace, index);
-        LOG.info("post-processing '" + executionId + "': processing " + ids.size() + " concept ids...");
         //No need to post process if another engine has beaten you to doing it
-        if(ids.isEmpty()) return;
+        if(ids.isEmpty()) {
+            LOG.info("post-processing '" + executionId + "': there " + ids.size() + " concept ids to post-process.");
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


grakn-engine/src/main/java/ai/grakn/engine/task/postprocessing/PostProcessingTask.java
```diff
@@ -81,23 +96,28 @@ public void run() {
      * @param keyspace The {@link Keyspace} requiring post processing for a specific index
      * @param index the index to be post processed
      */
-    public void processIndex(Keyspace keyspace, String index){
+    public void processIndex(Keyspace keyspace, String index, UUID executionId){
```
kasper-piskorski: how does this id help?

grakn-engine/src/main/java/ai/grakn/engine/task/postprocessing/PostProcessingTask.java
```diff
@@ -81,23 +96,28 @@ public void run() {
      * @param keyspace The {@link Keyspace} requiring post processing for a specific index
      * @param index the index to be post processed
```
duckofyork: Can you please update the javadoc as well

grakn-engine/src/main/java/ai/grakn/engine/task/postprocessing/PostProcessingTask.java
lolski: @kasper-piskorski make logs more understandable from the user perspective

grakn-engine/src/main/java/ai/grakn/engine/task/postprocessing/PostProcessingTask.java
lolski: Oops, fixed in https://github.com/graknlabs/grakn/pull/2713/commits/07574450c587475c554bdd3255aade27e2fd3b97

PR #2714, opened 2018-03-20
Supply storage host to spark

lolski:
# Why is this PR needed?
Compute count does not work when storage.hostname is set to other than 'localhost'.

# What does the PR do?
Tell Janus to talk to Cassandra on the right host

No

N/A

### DIFF
--- a/grakn-factory/src/main/java/ai/grakn/factory/TxFactoryJanusHadoop.java
+++ b/grakn-factory/src/main/java/ai/grakn/factory/TxFactoryJanusHadoop.java
@@ -26,8 +26,6 @@ import org.apache.tinkerpop.gremlin.structure.util.GraphFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Properties;
-
 /**
  * <p>
  *     A {@link ai.grakn.GraknTx} on top of {@link HadoopGraph}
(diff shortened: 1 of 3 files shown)


grakn-factory/src/main/java/ai/grakn/factory/TxFactoryJanusHadoop.java
```diff
@@ -43,15 +44,22 @@
  * @author fppt
  */
 public class TxFactoryJanusHadoop extends TxFactoryAbstract<EmbeddedGraknTx<HadoopGraph>, HadoopGraph> {
-    private static final String CLUSTER_KEYSPACE = "janusmr.ioformat.conf.storage.cassandra.keyspace";
-    private static final String INPUT_KEYSPACE = "cassandra.input.keyspace";
+    private static final String JANUSMR_IOFORMAT_CONF_STORAGE_CASSANDRA_KEYSPACE = "janusmr.ioformat.conf.storage.cassandra.keyspace";
```
lolski: yes, yes. various configuration that essentially meant the same thing. don't ask me why.

grakn-factory/src/main/java/ai/grakn/factory/TxFactoryJanusHadoop.java
```diff
@@ -43,15 +44,22 @@
...
+    private static final String JANUSMR_IOFORMAT_CONF_STORAGE_HOSTNAME = "janusmr.ioformat.conf.storage.hostname";
+    private static final String JANUSGRAPHMR_IOFORMAT_CONF_STORAGE_HOSTNAME = "janusgraphmr.ioformat.conf.storage.hostname";
+    private static final String JANUSGRAPHMR_IOFORMAT_CONF_STORAGE_CASSANDRA_KEYSPACE = "janusgraphmr.ioformat.conf.storage.cassandra.keyspace";
+    private static final String CASSANDRA_INPUT_KEYSPACE = "cassandra.input.keyspace";
```
kasper-piskorski: can we define the `janusgraphmr.ioformat.conf.` as a prefix as in `TxFactoryJanus`? probably worth introducing the override map as done there as well

grakn-factory/src/main/java/ai/grakn/factory/TxFactoryJanusHadoop.java
```diff
@@ -43,15 +44,22 @@
...
     TxFactoryJanusHadoop(EmbeddedGraknSession session) {
         super(session);
 
-        session().config().properties().setProperty(CLUSTER_KEYSPACE, session().keyspace().getValue());
-        session().config().properties().setProperty(INPUT_KEYSPACE, session().keyspace().getValue());
+        session().config().properties().setProperty(JANUSMR_IOFORMAT_CONF_STORAGE_CASSANDRA_KEYSPACE, session().keyspace().getValue());
+        session().config().properties().setProperty(CASSANDRA_INPUT_KEYSPACE, session().keyspace().getValue());
+        session().config().properties().setProperty(JANUSGRAPHMR_IOFORMAT_CONF_STORAGE_HOSTNAME, session().config().getProperty(GraknConfigKey.STORAGE_HOSTNAME));
+        session().config().properties().setProperty(JANUSMR_IOFORMAT_CONF_STORAGE_HOSTNAME, session().config().getProperty(GraknConfigKey.STORAGE_HOSTNAME));
+        session().config().properties().setProperty(JANUSGRAPHMR_IOFORMAT_CONF_STORAGE_CASSANDRA_KEYSPACE, session().keyspace().getValue());
```
kasper-piskorski: Again probably worth introducing something like the `overrideMap` as in `TxFactoryJanus`. Would make things clearer.

PR #2718, opened 2018-03-21
Refactor TxFactoryBuilder

marco-scoppetta:


grakn-engine/src/test/java/ai/grakn/engine/controller/GraqlControllerTest.java
```diff
@@ -130,7 +130,7 @@ public void whenRunningGetQuery_EnsureAllInstancesAreReturned() {
         assertEquals(expectedInstances, instances);
     }
 
-    //TODO: THis test should be improved
+    //TODO: This test should be improved
```
marco-scoppetta: Improved! 👍

grakn-kb/src/main/java/ai/grakn/factory/TxFactoryAbstract.java
```diff
@@ -25,99 +25,113 @@
...
 
 import javax.annotation.CheckReturnValue;
+import javax.annotation.Nullable;
 
 import static javax.annotation.meta.When.NEVER;
 
 /**
  * <p>
- *     Defines the construction of Grakn Graphs
+ * Defines the construction of {@link GraknTx} based on
```
aelred: based on what? 🤔

grakn-kb/src/main/java/ai/grakn/factory/EmbeddedGraknSession.java
```diff
@@ -129,27 +137,19 @@ public CommitLogHandler commitLogHandler(){
      * A user should not call this method directly.
      * See {@link Grakn#session(String, String)} for creating a {@link GraknSession} for the remote API
      */
-    @SuppressWarnings("unused")//This must remain public because it is accessed via reflection
-    public static EmbeddedGraknSession create(Keyspace keyspace, SimpleURI engineUri){
-        return create(keyspace, engineUri.toString());
-    }
```
aelred: We should keep this method, since it takes a `SimpleURI`, not a `String`

marco-scoppetta: It's unused for now, I think we should have just one method and  in the future this method should accept `SimpleURI` - but for that we should do a full refactor (given that `EmbeddedGraknSession` uses `uri` as a String )- for now having two methods was just confusing to me 🤔 🥇

marco-scoppetta:
!rtg

PR #2720, opened 2018-03-22
On `match ... insert`, project results over vars in `insert`

aelred:
# Why is this PR needed?

The [TP task](https://work.grakn.ai/entity/19733-on-match-insert-project-results-over) summarises the new behaviour.

The motivation is this will help prevent duplicate insertions within `match ... insert` queries, by de-duplicating the results of the `match`.

For example:

```
match ($x, $y) isa friendship; ($y, $z) isa friendship;
inert (fof: $x, fof: $z) isa friend-of-a-friend;
```

In this case, if Alice and Bob have many friends in common, we probably only want to add a friend-of-a-friend relationship _once_. Therefore, we project the `match` results only on `$x` and `$z` (removing `$y`) and de-duplicate.

# What does the PR do?

Changes the `match ... insert` behaviour to project on the `match` results.

It changes query semantics, so yes. But it doesn't break the data model.

None.


grakn-graql/src/main/java/ai/grakn/graql/internal/query/runner/TinkerQueryRunner.java
```diff
@@ -90,11 +93,18 @@ public static TinkerQueryRunner create(EmbeddedGraknTx<?> tx) {
...
-        );
+        return query.admin().match()
+                .map(match -> runMatchInsert(match, varPatterns))
+                .orElseGet(() -> Stream.of(QueryOperationExecutor.insertAll(varPatterns, tx)));
+    }
+
+    private Stream<Answer> runMatchInsert(Match match, Collection<VarPatternAdmin> varPatterns) {
+        Set<Var> varsInMatch = match.admin().getSelectedNames();
+        Set<Var> varsInInsert = varPatterns.stream().map(VarPatternAdmin::var).collect(toImmutableSet());
+        Set<Var> projectedVars = Sets.intersection(varsInMatch, varsInInsert);
```
kasper-piskorski: aren't variables from insert enough?

aelred: You also need the ones in the `match`. For example:
```
match $x has name "Alice";
insert (wife: $x, husband: $y) isa marriage; $y isa person, has name "Bob";
```
If we used only `insert` variables, we would project the `match` answers on `$x` and `$y`. The `match` part doesn't actually contain `$y`, so it'll throw.

kasper-piskorski: :+1:

PR #2728, opened 2018-03-26
updated docs deploy script

marco-scoppetta:
# Why is this PR needed?

Add `_jekyll` and `_site` folders even though they are ignored in the main `.gitignore`
Checkout to previous branch instead of `stable`

# What does the PR do?

Updates the `./deploy.sh`

No

Maybe

### DIFF
--- a/docs/deploy.sh
+++ b/docs/deploy.sh
@@ -5,6 +5,7 @@ echo "Building _site"
 rake build
 echo "Staging new files"
 git add -A
+# Forcefully add the _jekyll and _site folder before pushing to Heroku, this is necessary because both folders are ignored under the main .gitignore
 git add --force -- _jekyll _site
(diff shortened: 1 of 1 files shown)


docs/deploy.sh
```diff
@@ -4,11 +4,12 @@ git checkout -b docs-deploy-temp stable
 echo "Building _site"
 rake build
 echo "Staging new files"
-git add -A 
+git add -A
+git add --force -- _jekyll _site
```
lolski: This could be non-obvious to unfamiliar readers. I would propose adding a comment:
```
git add --force -- _jekyll _site # Forcefully add the _jekyll and _site folder before pushing to Heroku, this is necessary because both folders are ignored under the main .gitignore
```

PR #2733, opened 2018-03-28
Change GetQuery and InsertQuery to return streams by default

aelred:
# Why is this PR needed?

The main reason is related to gRPC. Originally, all queries over gRPC would yield streams of results.
e.g. a `GetQuery` would get a stream of `Answer`s, `compute count` would return a stream containing just one number, a `DeleteQuery` would return an empty stream...

The Python client (and any other clients we implement) cannot tell the difference between these - for example, if the Python client receives an empty stream, is that:
1. An empty stream of answers from a `get` query?
2. An indication of "nothing" such as from a `delete` query?

There's no way to say! This PR changes the behaviour such that "proper" streaming queries return a stream of responses over gRPC, while ones that return a single result actually just return one result.

# What does the PR do?

As well as the fairly simple gRPC change, this also required a big refactor of `GetQuery` and `InsertQuery` to be parameterised on `Stream<Answer>` instead of `List<Answer>`. This is actually a very simple refactor, even if it's big.

I added in a new `toList` method, since the old behaviour was still helpful.

Yes, since `GetQuery#execute` and `InsertQuery#execute` return a different type now! But it's not data model, so it's fine.

None.

### DIFF
--- a/grakn-client/src/test/java/ai/grakn/remote/RemoteGraknTxTest.java
+++ b/grakn-client/src/test/java/ai/grakn/remote/RemoteGraknTxTest.java
@@ -208,7 +208,7 @@ public class RemoteGraknTxTest {
         TxResponse response =
                 TxResponse.newBuilder().setQueryResult(QueryResult.newBuilder().setOtherResult("true")).build();
 
-        server.setResponseSequence(GrpcUtil.execQueryRequest(query), response);
+        server.setResponse(GrpcUtil.execQueryRequest(query), response);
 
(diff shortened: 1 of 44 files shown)


aelred:
@marco-scoppetta @kasper-piskorski [This](https://github.com/graknlabs/grakn/pull/2733/commits/2afff7e52ec0b58aa91ba582b0d558db69b7e563) is the only "interesting" commit that isn't dull refactor stuff.

PR #2736, opened 2018-03-28
Explanation fixes

kasper-piskorski:
# Why is this PR needed?
Fixes issues:
- not differentiating partial answers if they come from two equivalent queries
- visualising implicit attribute relations when not asked to

# What does the PR do?
Switch to list storage in explanations instead of a set. Adds an extra condition when building concepts for visualisation.

No.
No.


grakn-graql/src/main/java/ai/grakn/graql/internal/reasoner/query/ReasonerQueryEquivalence.java
```diff
@@ -68,6 +68,19 @@ public static boolean containsEquivalentAtom(ReasonerQuery query, Atom atom, BiF
...
+        @Override
+        protected boolean doEquivalent(ReasonerQuery q1, ReasonerQuery q2) {
+            return equivalence(q1, q2, Atomic::equals);
+        }
+
+        @Override
+        protected int doHash(ReasonerQuery q) {
+            return equivalenceHash(q, Atomic.class, Atomic::hashCode);
+        }
+    };
```
aelred: Is this being used anywhere?

PR #2756, opened 2018-04-09
Rename directIsa -> isaExplicit

duckofyork:
# Why is this PR needed?
More intuitive?
# What does the PR do?
Rename
For java version, yes. Graql syntax stays the same
NA


haikalpribadi:
so we think *Explicit is better *Explicitly, Jason? @duckofyork

duckofyork:
I'm not sure I understand your questions @haikalpribadi

PR #2761, opened 2018-04-13
Updating Outdated Grakn License to AGPL In Many Places

lolski:
# Why is this PR needed?
There are many places in the code which still indicates Grakn Core is a GPL product. This is incorrect as we're now AGPL.

# What does the PR do?
1. Added `license-maven-plugin` so we don't have to update the license files by hand
2. Updated Grakn to Be Licensed under AGPL in the `LICENSE.txt` file, in all of our subprojects. The change is done automatically via `mvn license:update-project-license` commands.
3. Updated Grakn to Be Licensed under AGPL in the `pom.xml`.

No

Update `*.java` file headers


kasper-piskorski:
why not stable?

PR #2768, opened 2018-04-16
Update header license

lolski:
# Why is this PR needed?
Update header license to AGPL

# What does the PR do?
Update header license to AGPL

No

N/A


kasper-piskorski:
not stable?

lolski:
@kasper-piskorski only docs and hotfix should go to `stable`

PR #2770, opened 2018-04-16
Syntax optimisation: make commas strictly needed for 'compute cluster' query parameters

haikalpribadi:
# Why is this PR needed?

We need to make commas to be strictly required across the graql language, to make it more consistent. So we're starting with this part of the graql query first. 

# What does the PR do?

Make compute cluster query parameters to require commas.

It breaks syntax compatibility for users that don't put commas between parameters of compute cluster queries.

None


grakn-graql/src/test/java/ai/grakn/graql/internal/parser/QueryParserTest.java
```diff
@@ -717,11 +717,21 @@ public void testParseComputeClusterUsingKCoreWithK() {
...
+    public void testParseComputeClusterUsingCCWithoutCommas() {
+        parse("compute cluster in movie, person; using connected-component where size = 10 members = true");
+    }
+
+    @Test(expected = GraqlSyntaxException.class)
+    public void testParseComputeClusterUsingKCoreWithoutCommas() {
+        parse("compute cluster in movie, person; using k-core where k = 0 k = 1");
+    }
+
     @Test
```
duckofyork: It's better to test comma without parameter as well:  ```compute cluster; using k-core where k = 0,;```

grakn-graql/src/test/java/ai/grakn/graql/internal/parser/QueryToStringTest.java
```diff
@@ -176,7 +180,13 @@ public void testCCSizeToString() {
     }
 
     @Test
-    public void testKCoreToString() {
+    public void testClisterUsingKCoreToString() {
```
duckofyork: I think you mean ```testClusterUsingKCoreToString```

docs/pages/docs/05-distributed-analytics/compute-cluster.md
```diff
@@ -39,7 +39,7 @@ Here, the [subgraph](./overview) functionality has been used to get more meaning
 If you only wanna find out which cluster contains the given entity, you can use the modifier `source`.
```
duckofyork: It might be better to emphasise comma is needed in between parameters here, as this is a syntax change

grakn-graql/src/main/antlr4/ai/grakn/graql/internal/antlr/Graql.g4
```diff
@@ -32,15 +32,15 @@ std            : STD      'of' ofList      ('in' inList)? ';' ;
...
+kCore                 : CLUSTER            ('in' inList)? ';' USING 'k-core'              (WHERE kcParam (',' kcParam)*)? ';';
 path           : PATH    'from' id 'to' id ('in' inList)? ';' ;
 paths          : PATHS   'from' id 'to' id ('in' inList)? ';' ;
 count          : COUNT                     ('in' inList)? ';' ;
 
 ccParam        : MEMBERS       '='      bool            # ccClusterMembers
                | SIZE          '='      INTEGER         # ccClusterSize
-               | 'source'      '='      id              # ccStartPoint
+               | SOURCE      '='      id              # ccStartPoint
                ;
```
duckofyork: It would prettier if columns are aligned!

haikalpribadi:
I'm gonna cancel this PR because we want to change the comma behaviour.

PR #2771, opened 2018-04-17
Make comparisons on values less ambiguous

kasper-piskorski:
# Why is this PR needed?
To disambiguate and clean up syntax for comparing attribute values.

# What does the PR do?
* value equality and inequality syntax update:
  - `$x != $y` -> `$x !== $y`
  - `$x = $y` -> `$x == $y`
* comparisons with single equals, e. g. `$x = $y`, are not valid syntax anymore
* `val` keyword is now optional for specifying attribute values. Could remove it completely but leaving to maintain backwards compatibility. 
If we want to specify a value we can either do`$x == 'value';` or `$x 'value';`. 

### DIFF
--- a/grakn-graql/src/main/antlr4/ai/grakn/graql/internal/antlr/Graql.g4
+++ b/grakn-graql/src/main/antlr4/ai/grakn/graql/internal/antlr/Graql.g4
@@ -92,9 +92,9 @@ casting        : variable (':' VARIABLE)?
 
 variable       : label | VARIABLE ;
 
-predicate      : '='? value                     # predicateEq
-               | '=' VARIABLE                   # predicateVariable
-               | '!=' valueOrVar                # predicateNeq
+predicate      : '=='? value                     # predicateEq
+               | '==' VARIABLE                   # predicateVariable
+               | '!==' valueOrVar                # predicateNeq
                | '>' valueOrVar                 # predicateGt
                | '>=' valueOrVar                # predicateGte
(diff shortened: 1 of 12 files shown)


haikalpribadi:
@kasper-piskorski I don't understand the description you wrote in the above.

Did you follow the summary at the bottom of the comments in this ticket? https://work.grakn.ai/entity/19952-graql-operator-for-comparing-attribute-values

It says that the equality operators should behave such that:
```
$x = $y       -> $x and $y are the same concept
$x != $y      -> $x and $y are different concepts
$x == $y     -> $x and $y have the same value
$x !== $y    -> $x and $y have different values
```
Does your PR do this? 

Also, only Jason has accepted this PR. You should wait for me / @lolski :)

kasper-piskorski:
I trust Jason :) Yeah, I followed the summary. It does the change above apart from the `$x = $y` for concepts cause it is redundant and we already have this functionality.

haikalpribadi:
> It does the change above apart from introducing the $x = $y predicate for concepts cause it is redundant and we already have this functionality.

Thank god! I know we already have it, but I thought you removed it when in the description you said:

> comparisons with single equals, e. g. $x = $y, are not valid syntax anymore

...

> I trust Jason's judgement :)

I'm sure you do :) as do I :) But still, wait for other reviewers, or nudge us, so I can learn about the changes too now.

@kasper-piskorski

PR #2773, opened 2018-04-18
another update for gc docs

kasper-piskorski:
# Why is this PR needed?
Update guidelines.
# What does the PR do?
Removes dashboard mentions.
No.
No.


docs/pages/docs/01-cloud-deployment/gc-deployment.md
```diff
@@ -74,14 +53,44 @@ To check cluster health, execute the `grakn cluster status` command. The output
...
+
+Grakn console allows to perform the following actions:
+
+* create a new user:
+
+`CREATE USER username WITH PASSWORD userpassword WITH ROLE admin`
+
+* update existing user's password and/or role:
+
+`UPDATE username WITH PASSWORD newpassword WITH ROLE newrole`
```
marco-scoppetta: can we change the update section to:
```
update existing user's password

UPDATE username WITH PASSWORD newpassword

```
Let's not mention the role for now, as we only support `admin` users , until the day authorisation will be added.
Also maybe we should mention that only `admin` user will work when creating a new user. 🤔 
Sorry for not mentioning this before @kasper-piskorski

kasper-piskorski:
@marco-scoppetta  updated, have a look

PR #2776, opened 2018-04-19
Update docs so the example uses correct data types

duckofyork:
# Why is this PR needed?
Examples don't show all the data types we support
# What does the PR do?
Update docs so the example uses correct data types
No
None


duckofyork:
please re-review guys @jmsfltchr @kasper-piskorski

PR #2780, opened 2018-04-20
Reset all Attribute labels on Dashboard

Irtazaraza:
# Why is this PR needed?
Add Reset button to reset all node attributes and color

# What does the PR do?
Checks for all the attributes currently selected and resets to the default label and color

No

Have a tab in the node settings to remove selective preferences individually


grakn-dashboard/src/components/graphPage/footer/nodeLabelPanel.vue
```diff
@@ -160,7 +165,17 @@ export default {
...
+      if (p == null){   
+        const nodeProps = NodeSettings.getLabelProperties(this.nodeType);
+
+        // Reset all node attributes
+        for(var i in nodeProps){
+          NodeSettings.removeTypeLabel(this.nodeType, nodeProps[i]);;
+        }   
+        // Reset color
+        visualiser.setColourOnNodeType(this.node.baseType, this.node.type);
+      }
```
marco-scoppetta: Why are we resetting also the colour?

grakn-dashboard/src/components/graphPage/footer/nodeLabelPanel.vue
```diff
@@ -38,6 +38,7 @@ along with Grakn. If not, see <http://www.gnu.org/licenses/gpl.txt>.
             <div class="side-column"><colour-picker :node="node"></colour-picker></div>
         </div>
         <div class="panel-footer">
+            <button class="btn reset" @click="configureNode()">Reset</button>
```
marco-scoppetta: Maybe this button should be "Reset Label" to make it more explicit? Not sure, but it feels less ambiguous

grakn-dashboard/src/components/graphPage/footer/nodeLabelPanel.vue
marco-scoppetta: Given that we already have a button for resetting colour I think we should keep the functionalities separated

Irtazaraza: So should we also have a similar button for resetting the colour since right now we have a cross on top of the colours? Or should we keep that as it is?

grakn-dashboard/src/components/graphPage/footer/nodeLabelPanel.vue
Irtazaraza: If we keep the functionalities for resting the colours and labels separate then you are right.

PR #2782, opened 2018-04-23
Add an "internal-release" maven repository for Grakn Core internal release (ie., for KGMS)

lolski:
# Why is this PR needed?
Let's not mix the repository for normal release with the internal release for KGMS's use.

# What does the PR do?
Add an "internal-release" maven repository for Grakn Core internal release.

Grakn Core release from `master` will go to this repository, and have a version number based on its commit id which will look like `v1.1.0-174-g3b30ebaa7c5605bb1b28d27aacd3b6f410aad9da`. The version is generated via `git describe --abbrev=40 head`.

No

N/A

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1258,14 +1258,14 @@
             </distributionManagement>
         </profile>
         <profile>
-            <id>grakn-snapshots</id>
+            <id>internal-releases</id>
             <properties>
                 <skip.deploy.dist>false</skip.deploy.dist>
             </properties>
             <distributionManagement>
                 <repository>
-                    <id>releases</id>
-                    <url>https://maven.grakn.ai/content/repositories/snapshots/</url>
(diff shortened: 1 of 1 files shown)


lolski:
I set up another repository `internal-releases` in the maven.grakn.ai repository for internal release purpose (ie., Core builds used by KGMS).

We couldn't use our snapshot repository, since internal release is technically a "release", not a "snapshot release". Maven don't allow mixing non-snapshots with snapshots in the same repository.

We also shouldn't use our release repository to host internal releases, for clarity purpose (imagine a user getting confused if he sees our internal releases).

kasper-piskorski:
@lolski what constitutes an internal release?

lolski:
@kasper-piskorski Grakn Core build that is to be used for KGMS.

PR #2783, opened 2018-04-24
Analytics query syntax upgrade: use commas (not semicolon) and use square brackets for arguments

haikalpribadi:
_Note: this PR is big, but not because of refactorings, nor is it because it contains multiple features. It does contain refactorings, but you can see that they are only changes in function names and they are actually required because we want to align the terminology with the new grammar (and with the rest of the Graql Java API which already has it). This PR is actually big because of feature we are implementing (new analytics syntax) means that we need to change the way Graql Compute queries are being parsed. And given that we are implementing a new way of parsing the syntax, we have the opportunity to provide better error messages to our users and therefore this PR contains that too (rightly so)._

# Why is this PR needed?

We want to implement a new feature in Graql syntax, and that is the ability to provide (a set of) arguments to a given graql statement. Such as:
`define person sub entity, has email [unique, max=3];`
This goal requires us to first fix analytics so that the `graql compute` queries:
1. use commas (`,`) properly instead of (`;`), and 
2. use `[` and `]` to provide a list of arguments.

As a result, analytics queries will look like: 
`compute cluster in [person, marriage], using connected-component, where [start = "V123", members = true];`

However, the above is change is not possible in `graql compute` queries unless we change the grammar to be more flexible (composing it of generic patterns) rather than hard-coded patterns for each type of `graql compute` queries. The generic pattern for `graql compute` can now be simply described as:

Compute query -> `compute <method> <conditions>? ;`

The `Graql.g4` file contains a thorough description of the new grammar and should guide you to understand the whole implementation.

# What does the PR do?

Thus, this PR contains changes in:
1. Graql grammar definition (`Graql.g4`)
2. `GraqlConstructor` (renamed from `GraqlVisitor`) and `GraqlComputeConstructor` to implement the new way of parsing our `graql compute` queries.
3. New exceptions in `GraqlQueryException` along with the new error messages in `ErrorMessage`.
4. New implementations of `toString()` method for `graql compute` queries to also match the new syntax. The `toString()` method is composed of the same generic patterns: `method` and `conditions`, each of which are implemented uniquely in every `graql compute` query class.
5. Changes in function names in various `graql compute` query classes, to use the same terminology as the grammar file.

Yes, it does. Old `graql compute` syntax will not work with this new change.

Nothing is known so far.


haikalpribadi:
Docs is now updated to match the new syntax, so Travis should pass. Jenkins has already passed before. Changed the function argument names that I missed, that @duckofyork pointed out. @kasper-piskorski @lolski if you can review this today that would be cool, guys!

grakn-core/src/main/java/ai/grakn/util/ErrorMessage.java
```diff
@@ -204,6 +204,114 @@
...
 
     //--------------------------------------------- Analytics Errors -----------------------------------------------
+    INVALID_COMPUTE_METHOD("Invalid compute method. " +
+            "The available compute methods are: count, min, max, median, mean, std, sum, path, paths, centrality, and cluster. " +
+            "A valid compute method contains 'compute <method> <conditions>', such as: 'compute min of age, in person;"),
+
+    // Graql compute count errors ------------------
+
+    INVALID_COMPUTE_COUNT_CONDITION("Invalid 'compute count' condition. " +
+            "Only 'in <types>' is accepted, such as: 'compute count in [person, movie]];'."),
```
lolski: small typo: double closing bracket.

haikalpribadi: I will fix this in the next PR, @lolski. So it doesn't have to trigger another test.

lolski: sure

haikalpribadi: okay i fixed it in the end, cause there were more improvements I wanted to add to the Graql.g4 grammar file. @lolski

docs/pages/docs/05-distributed-analytics/compute-centrality.md
```diff
@@ -7,29 +7,29 @@ permalink: /docs/distributed-analytics/compute-centrality
...
-The degree of an instance gives the number of other instances directly connected to the instance. 
-For an entity, such as a `person`, this means the number of connections, 
+The degree of an instance gives the number of other instances directly connected to the instance.
+For an entity, such as a `person`, this means the number of connections,
 for example instances of a `marriage` relationship.
 The `person` with the highest degree, or the greatest number of marriages, is arguably the most "interesting"
 ([Glynn Wolfe](https://en.wikipedia.org/wiki/Glynn_Wolfe)).
 
-In analyics, the `degree` algorithm computes how many arrows (edges) there are attached to instances in the graph. 
+In analyics, the `degree` algorithm computes how many arrows (edges) there are attached to instances in the graph.
```
kasper-piskorski: typo in analytics

docs/pages/docs/05-distributed-analytics/compute-centrality.md
```diff
@@ -50,14 +50,14 @@ if it plays a role in another relationship." %}
...
-An instance have coreness k if the instance belongs to a 
+An instance have coreness k if the instance belongs to a
 [k-core](https://en.wikipedia.org/wiki/Degeneracy_(graph_theory)#k-Cores) but not to any
 (k+1)-core.
 
 We can compute the coreness centrality using the following:
 
 ```graql
-compute centrality; using k-core;
+compute centrality using k-core;
```
kasper-piskorski: no comma here?

grakn-core/src/main/java/ai/grakn/util/GraqlSyntax.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+package ai.grakn.util;
+
+/**
+ * Graql syntax keywords
+ *
+ * @author Haikal Pribadi
+ */
+public class GraqlSyntax {
```
kasper-piskorski: what's this for? how do we ensure it's in sync with the antlr grammar file?

haikalpribadi: just to document my answer to @kasper-piskorski in person:

We need this class so we can replace all usage of hard-coded strings of Graql's syntax across our code base. It will be much less error-prone and much easier to manage when we change the syntax. 

It may appear that we have GraqlParser (produced by ANTLR) which contains Graql Syntax' constants, but these constants are integers that are generated by ANTLR, which may change after compiling updates from the grammar file.

@lolski @duckofyork

PR #2784, opened 2018-04-24
Dashboard ask query bug

Irtazaraza:
# Why is this PR needed?
Ask query output in dashboard console has odd characters

# What does the PR do?
checks if the the query response contains the odd characters and trims the response string to exclude them

No.

1) Check when the query is an ask query and then remove the unwanted characters
2) Check at the backend if the query is for the console page of the dashboard and does not send the unwanted characters along with the response.

### DIFF
--- a/grakn-dashboard/src/components/consolePage.vue
+++ b/grakn-dashboard/src/components/consolePage.vue
@@ -130,6 +130,12 @@ export default {
          * EngineClient callbacks
          */
     shellResponse(resp, err) {
+
+      // Remove unwanted characters during ask query
+      if(resp.includes('[')){
+        resp = resp.slice(5,-4);
+      } 
+
       if (resp.length === 0) {
         this.state.eventHub.$emit('warning-message', 'No results were found for your query.');
       } else {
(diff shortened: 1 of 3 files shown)


grakn-dashboard/src/components/consolePage.vue
```diff
@@ -130,6 +130,12 @@ export default {
          * EngineClient callbacks
          */
     shellResponse(resp, err) {
+
+      // Remove unwanted characters during ask query
+      if(resp.includes('[')){
+        resp = resp.slice(5,-4);
+      }
```
marco-scoppetta: This check is a bit too optimistic, at this point in the code we don't know what type of shell response we are receiving, so if for example we receive an analytics query that looks like `[1,4,5]` this code would trim the response yielding strange results - even if one day we want to change the output of ask queries to be something like `[true, false, true]` this will also break. As you correctly stated in the PR description this whole issue should be fixed at server side, but that being said, this is a good use case for Regex:
```
let askRegex = /^\[32m(.*)\[0m$/;
if(askRegex.test(resp)){
   resp = askRegex.exec(resp)[1]; // we take group1 in the regex matchers
}
```
or more simply:
```
if(resp.startsWith('[32m')&&resp.endsWith('[0m')){
  ...trim the string
}
```

PR #2786, opened 2018-04-25
add quotes around classpath variable

lolski:
# Why is this PR needed?
Grakn won't be able to boot if the distribution is located on a path that has whitespace, such as `/home/user/grakn location/". This is because we didn't put a quote around the `$CLASSPATH` variable when passing it to the java bootup class

# What does the PR do?
add quotes around classpath variable

No

N/A


duckofyork:
are we adding the fix to stable as well? @lolski

lolski:
@duckofyork good idea, let's merge this into `stable` instead.

PR #2789, opened 2018-04-26
Fix not running tests

kasper-piskorski:
# Why is this PR needed?
Some tests were not run because travis regex wouldn't pick them up.
# What does the PR do?
Change test names and test regex.
No.
No.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,10 @@ addons:
 
 env:
   # Split tests into grakn-test and everything else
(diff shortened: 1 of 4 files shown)


haikalpribadi:
I've approved this PR, but I think you should also make the change for Jenkins, @kasper-piskorski ?

kasper-piskorski:
@haikalpribadi tests are run correctly on jenkins

PR #2790, opened 2018-04-27
Remove the optional val keyword for comparing attribute values

kasper-piskorski:
# Why is this PR needed?
Make the syntax less verbose and confusing and kill the `val`.
# What does the PR do?
Remove the optional `val` from syntax.
Yes. Queries with `val` are no longer valid.
No


haikalpribadi:
Sorry I didn't see this email buried in my inbox, team! @graknlabs/developers

PR #2797, opened 2018-05-02
Dashboard commands/shortcuts modal

Irtazaraza:
# Why is this PR needed?
To add a page in dashboard to show all the commands/shortcuts for interacting with dashboard's graph page

# What does the PR do?
Show and hide a modal to list all possible ways to interact with graph on dashboard

No

Include any new shortcuts or commands


grakn-dashboard/src/components/mainTemplate/commandsModal.vue
```diff
@@ -0,0 +1,326 @@
...
+                        <td>
+                            Organize nodes and reveal their connections without them jumping out of place
+                        </td>                             
+                    </tr>
+                    <tr>
+                        <td style="text-align:center;">
+                            Activate inference
+                        </td>  
+                        <td>
+                            It is off by default per query, but turn this on when you need to run queries that use inference
```
marco-scoppetta: Inference is ON by default

grakn-dashboard/src/js/User.js
```diff
@@ -62,6 +62,20 @@ export default {
...
+  
+    getCmdModalShown() {
+      const cmdModalShown = localStorage.getItem('commands_modal');
+      if (!cmdModalShown) {
+        this.setCmdModalShown(false);
+        return false;
+      }
+      return (cmdModalShown === 'true');
+    },
+
```
marco-scoppetta: Why do we need to remember if the user has already seen the command modal?

marco-scoppetta: We do that only for the join community because we want to show that modal immediately to the user the first time the user open dashboard, without waiting for him to click on the link. So we don't need it for commands, we should show this modal only if the user clicks on it.

Irtazaraza:
@marco-scoppetta rework after review completed

PR #2798, opened 2018-05-02
fix owl error message

lolski:
# Why is this PR needed?
Fix grammar error and centralise the error message into `GraknUtil.ErrorMessage`.

# What does the PR do?
1. Centralise error message into `GraknUtil.ErrorMessage`.
2. Fix minor grammar error.

No

N/A

### DIFF
--- a/grakn-bootup/src/main/java/ai/grakn/bootup/Graql.java
+++ b/grakn-bootup/src/main/java/ai/grakn/bootup/Graql.java
@@ -29,6 +29,7 @@ import ai.grakn.migration.export.Main;
 import ai.grakn.migration.json.JsonMigrator;
 import ai.grakn.migration.sql.SQLMigrator;
 import ai.grakn.migration.xml.XmlMigrator;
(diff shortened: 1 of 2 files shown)


grakn-bootup/src/main/java/ai/grakn/bootup/Graql.java
```diff
@@ -103,7 +104,7 @@ private void migrate(String[] args) {
                 JsonMigrator.main(valuesFrom(args, 1));
                 break;
             case "owl":
-                System.out.println("Owl migration not supported anymore");
+                System.err.println(ErrorMessage.OWL_NOT_SUPPORTED.getMessage());
```
kasper-piskorski: so why don't we remove it completely? BO reasons?

lolski: I think it's good to keep, as there are people coming over from OWL asking about if they can migrate automatically

PR #2799, opened 2018-05-04
Highlight '[', ']', 'using', and 'where' in graql queries

Irtazaraza:
# Why is this PR needed?
Highlight '[', ']', 'using', and 'where' in graql queries

# What does the PR do?
Added the following characters to the patterns to match while highlighting queries and response.
[
]
using
where

No. 

Currently we are using two ways to highlight text.
We can remove codemirrorGraql.js and use PrismGraql.js to highlight in all places: Graql editor, console page, and while saving the current query in addCurrentQuery.vue

### DIFF
--- a/grakn-dashboard/src/components/graphPage/modules/QueryBuilder.js
+++ b/grakn-dashboard/src/components/graphPage/modules/QueryBuilder.js
@@ -35,7 +35,7 @@ export default {
   },
 
   shortestPathBuilder(nodeIds) {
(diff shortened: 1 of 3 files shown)


Irtazaraza:
Hi @marco-scoppetta : I tried removing one of the highlighters however i ran into some Problems. Lets go over this once your're back.

Irtazaraza:
Hey @marco-scoppetta : Do you know why Travis failed?

marco-scoppetta:
@Irtazaraza the built stalled - I will restart it

PR #2800, opened 2018-05-06
correct typo

ShalokShalom:
# Why is this PR needed?

# What does the PR do?

correct typo

no


ShalokShalom:
you do this on other places too, like here: https://grakn.ai/grakn-core

Is this page also hosted here on Github?

jmsfltchr:
The team members at Grakn all use British English rather than American English. 
UK - modelling
US - modeling
It would be nice if our website didn't look odd to those accustomed to American English spellings, but I think for the sake of consistency we should stick to British English. We'll keep this in mind for the time when we support multiple languages on the website!

PR #2801, opened 2018-05-07
Complexity Refactor: phase 1

haikalpribadi:
# Why is this PR needed?

Many parts of our code/architecture are unnecessarily complex. Complex code is hard to maintain and slow to improve. The consequence of a complex code can be very significant in terms of developer productivity and quality of life. It also puts the company and technology at risk when someone is unavailable/leaves (i.e. bus factor is too high). Thus, we need to simplify our code/architecture, to improve productivity, quality of life, and reduce our bus factor to ZERO (yes, ZERO). 

This PR is far from the end. After many iterations of refactors we will start seeing a significant drop in package numbers, package dependencies, and thus, a drop in complexity of the architecture. 

Here's a nice 14 minutes talk on why maintaining a simple architecture is important, by Martin Fowler: https://www.youtube.com/watch?v=DngAZyWMGR0

# What does the PR do?

Admittedly, there are 3 refactors in this PR. 1 major (Analytics) and 2 minors (GRPC and Printers). They should have been made through 3 separate PRs but the 2 smaller refactors were need to smoothen the work, and it's a waste of time to try to split them up now. There will be more refactor in Printers and GRPC separately to simplify them more radically, like what we've done with analytics.

### grakn-client
- RemoteQueryExecutor
	* we no longer need to have individual methods for diff compute queries
	* we also don't need granular methods with one line and complicated generics

### grakn-core
- ConceptId
	* no longer uses Auto_Value
	* we need a default no*argument constructor for serialisation within OLAP frameworks

- GraqlQueryException and ErrorMessage
	* we now have generic exception messages for analytics
	* but taking in arguments, which makes the message very specific
	
- Answer
	* renamed immutableMap() to just map(). "immutable" prefix is implementation detail.

- Analytics (17 compute query interfaces)
	* GONE
	* Too much granular interfaces
	* Too complicated to manage and error prone
	* Dissolves any structure that compute queries could have
	
- ComputeQuery
	* Replacement of 17 old Compute Query interfaces!!!!!
	* Removal of complicated Generics
	- One return type: ComputeQuery.Answer

- GraqlSyntax
	* a class to contain all hard-coded Graql literal, values and constants
	* literally the Java Grammar file for Graql
	* we need to think of a better name (perhaps around the word "lang")

- QueryExecutor
	* removed 12 individual compute query methods and replace with just one
	* new method takes in new ComputeQuery

### grakn-graql
- ClusterMemberMapReduce
	* now returns ConceptIds rather than Strings (of ConceptIds)

- DegreeDistributionMapReduce
	* now returns ConceptIds rather than Strings (of ConceptIds)

- ShortestPathVertexProgram
	* rewritten (algorithm) to align with the algorithm pseudo-code
	* this improves readability and understandability, which is critical for such complex piece of code

- GreedyTraversalPlan
	* a lot of comments were added to explain the algorithm inline with the code

- 19 Analytics Compute Query and Compute Query Builder Classes
	* GONE!! GONE GONE GONE!!
	* all of the Compute Query classes can now be implemented in one simple class: ComputeQueryImpl
	* all of the Compute Query Builder classes are also not needed now given we only have 1 simple class to build

- ComputeQueryImpl
	* ComputeQuery can now take in 6 conditions: from, to, fo, in, using and where
	* where() takes in a set of Argument (singular) and returns an Arguments (plural) object
	* ComputeQuery itself returns an Answer object, this removes the need of overusing generics
	* Arguments and Answer are only used by ComputeQuery
		* they change together
		* they share inner properties
		* thus they are nested
	* We now have one generic method to validate every type of ComputeQuery
		* grammar rules are taken from GraqlSyntax
	* We now have one generic method to get the string representation of ComputeQuery
	* The ComputeQuery class now becomes an expressive framework to contain all type of queries
		* rarely needs to be extended, as we only need to extend GraqlSyntax for new type of queries

- TinkerComputeQuery, TinkerComputeQueryExecutor, TinkerStatisticsQuery
	* ALL GONE!
	* TinkerComputeQueryExecutor contained way too many convoluted lambdas
		* absolutely difficult to maintain
	* TinkerComputeQuery and TinkerStatisticsQuery are all helper methods
		* methods are poorly named
		* some are redundant
		* most are hard to follow
	* all are rewritten into TinkerComputeJob

- TinkerComputeJob
	* initially was just a class to hold the "function" that represents the Compute Query execution
	* now properly contains the execution methods for each of type of Compute Query
	* all helper methods are properly named and used universally across all execution methods

- TinkerQueryExecutor
	* removed all the run() methods for different classes of Compute Queries
		* replaced with just one run(ComputeQuery) method
		* applied the change onto interface, QueryExecutor, and RemoteQueryExeuctor
			* QueryExecutor was renamed from QueryRunner, as it contains one responsibility: execute()
			* Further refactors, such as renaming methods, is required for QueryExecutor

- GraqlConstructor (which was initially added to separate out ComputeQuery construction)
	* GONE!
	* now that there is only one ComputeQuery with a simple query pattern
	* construction can happen with a simple method, written in GraqlConstructor

- GraqlConstructor
	* renamed from GraqlParser to disambiguate the responsibility with ANTLR (which has a GraqlParser)
	* implementation of any ComputeQuery construction (or so called 'parsing') is now done with 2 simple methods
		* visitComputeQuery() will construct the overall ComputeQuery
		* visitComputeArgs() will construct the Arguments for ComputeQuery

- QueryParserImpl
	* some lambdas are slightly expanded to improve understandability

### grakn-spec
- Migrated into main Grakn repo
	* submodule has been removed

- Grakn.proto
	* changed to accomodate new ComputeQuery.Answer and it's internal data structures

- Concept.proto
	* changed to accomodate ConceptIds, to represent a Collection of ConceptIds

### grakn-grpc
- ResponseBuilder
	* contains the new methods too build ComputeQuery.Answer responses for GRPC messages

- GrpcClient
	* contains new methods to read ComputeQuery.Answer reponses for GRPC messages
	* these new methods really should sit in a RespnoseReader class
	* but to build one requires much more significant refactoring of grakn-grpc

### grakn-engine (RPC) and grakn-grpc
- GrpcConverter and GrpcUtil
	* GONE! (replacement classes are in ai.grakn.rpc.util)
	* One often gets lost in GrpcConverter and GrpcUtil
	* every method is either called build() or convert()
		* abuse of method overloading
		* Few are recursively calling each other
		* Most only seem like they are recursive
		* Most are just one liner methods
		* They're called from hundreds of places, and 
		* none are obvious what they're doing because they all have the same name!! 
	* Does not actually contain all the methods to do GRPC conversion
	* Most of the conversion methods are in GrpcUtil and they are all called convert(),
	* Even though GrpcConverter implements GraqlConverter
		* it actually only properly implements ONE METHOD! 
		* The rest are just dummy methods. 
		* And that one method also OVERRIDES A DEFAULT METHOD form the interface.
		* At this point, implementing the interface pointless.

- ConceptBuilder
	* New class to contain refactored convert() methods from GrpcConverter/GrpcUtil
	* responsible for building concepts for GRPC messages
	* all abused overloaded method names are renamed to their proper name

- ConceptReader
	* New class to contain refactored convert() methods from GrpcConverter/GrpcUtil
	* responsible for reading concepts from GRPC messages
	* all abused overloaded method names are renamed to their proper name

- ResponseBuilder
	* New class to contain refactored convert() methods from GrpcConverter/GrpcUtil
	* responsible for building responses for GRPC messages
	* all abused overloaded method names are renamed to their proper name

- RequestBuilder
	* New class to contain refactored convert() methods from GrpcConverter/GrpcUtil
	* responsible for building requests for GRPC messages
	* all abused overloaded method names are renamed to their proper name

- TxGrpcCommunicator
	* variable names changes, to clarify what they do
	* e.g. requestSender and responseListener

### grakn-core
- GraknTxType
	* by having the identity method, we can avoid other complex converter classes just to retrieve GraknTxType enum

### grakn-core
- Query<T>
	* removal of results() and resultString() method
	* Query object should not depend on Printer methods. The reverse should be true.

- Printer<T>
	* GONE!
	* This class is a fake subclassification of a group of GraqlConverters which are supposed to act as Printers
	* Meanwhile, GraqlConverter itself is just Printer
	* It had one (deprecated) method graqlString() that calls convert()
	* It introduces more unecessary generics
	* It's a wasteful middleman (i.e. middle class)

- GraqlConverter
	* moved to ai.grakn.graql.internal.printer.Printer
	* further explanation below

### grakn-graql
- Printer
	* renamed from GraqlConverter because:
	* that's exactly what it is left to do (and meant to be), after removing GrpcConverter from it's subclasses
	* removed abusive method overloading and illusion of recursion
	* refactored from an interface to abstract class, because:
		* it carries too many encapsulated behavior rather than just a promise of methods
		* i.e. it is an implementation
	* extended to take in a graql query execution result and print them
		* because Printers should depend on what it prints, not the other way around. For example:
		* Canon Printers should depend on MS Word library. MS Word should not depend on Canon Drivers.
		* The core product should not depend on accessories. Accessories should depend on the core product.
	* The wrong dependency structure causes further complication such as:
		* fake stream() methods on non-streamable classes, this needs to eventually be cleaned up

- JsonPrinter
	* Extended to print ComputeQuery.Answer, but still using default JSON printer

- StringPrinter
	* renamed from GraqlPrinter - because all printers are Graql printers!
	* StringPrinter represents the fact that it simply prints to String
		* i.e. with no standard formatting such as JSON, XML, YAML and so on
	* Extended to pring ComputeQuery.Answer, in a style we define ourselves

### grakn-graql-shell
- OutputFormat
	* This is a wrong use of enum, and therefore confusing
	* Refactored into a normal abstract class with two inheritance

### grakn-dist
- Grakn ASCII
	* New Grakn Ascii!! :)


grakn-core/src/main/java/ai/grakn/graql/admin/Answer.java
```diff
@@ -67,7 +66,7 @@
     Concept get(Var var);
 
     @CheckReturnValue
-    ImmutableMap<Var, Concept> map();
```
kasper-piskorski: why?

grakn-core/src/main/java/ai/grakn/graql/admin/ReasonerQuery.java
```diff
@@ -72,7 +72,7 @@
     /**
      * @param type the class of {@link Atomic} to return
      * @param <T> the type of {@link Atomic} to return
-     * @return stream of atoms of specified type defined in this query
```
kasper-piskorski: why are we changing `stream` to `output` in comments? did search-replace went too far?

grakn-graql-shell/src/main/java/ai/grakn/graql/shell/GraqlShell.java
```diff
@@ -47,32 +48,6 @@
 import static java.util.stream.Collectors.toSet;
 import static org.apache.commons.lang.StringEscapeUtils.unescapeJavaScript;
 
-/* uncover the secret
```
kasper-piskorski: we getting rid of the hippos ;(?

grakn-graql/src/main/java/ai/grakn/graql/internal/query/AbstractQuery.java
```diff
@@ -40,17 +38,7 @@
     @CheckReturnValue
     protected abstract Stream<S> stream();
 
-    @Override
-    public final Stream<String> resultsString(Printer<?> printer) {
-        return results(printer);
-    }
-
-    @Override
-    public final <U> Stream<U> results(GraqlConverter<?, U> converter) {
```
kasper-piskorski: why are we getting rid of these?

grakn-graql/src/main/java/ai/grakn/graql/internal/query/runner/TinkerComputeJob.java
```diff
@@ -20,44 +20,820 @@
...
+import static ai.grakn.util.GraqlSyntax.Compute.Algorithm.DEGREE;
+import static ai.grakn.util.GraqlSyntax.Compute.Algorithm.K_CORE;
+import static ai.grakn.util.GraqlSyntax.Compute.Method.MEDIAN;
 
 /**
- * A compute query job executed against a {@link GraknComputer}.
+ * A Graql Compute query job executed against a {@link GraknComputer}.
  *
- * @author Felix Chapman
+ * @author Haikal Pribadi
```
kasper-piskorski: Add Felix back.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/ComputeQueryImpl.java
```diff
@@ -0,0 +1,647 @@
...
+import static ai.grakn.util.GraqlSyntax.Compute.Parameter.MEMBERS;
+import static ai.grakn.util.GraqlSyntax.Compute.Parameter.MIN_K;
+import static ai.grakn.util.GraqlSyntax.Compute.Parameter.SIZE;
+import static java.util.stream.Collectors.joining;
+
+
+/**
+ * Graql Compute Query: to perform distributed analytics OLAP computation on Grakn
+ *
+ * @author Haikal Pribadi
```
kasper-piskorski: add Jason

grakn-grpc/src/main/java/ai/grakn/rpc/util/ConceptBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+import java.time.ZoneId;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import static java.util.stream.Collectors.toList;
+
+/**
+ * A utility class to build RPC Concepts by converting them from Grakn Concepts.
+ *
+ * @author Haikal Pribadi
```
kasper-piskorski: Add Felix as well.

grakn-grpc/src/main/java/ai/grakn/rpc/util/ConceptReader.java
```diff
@@ -0,0 +1,130 @@
...
+import ai.grakn.util.CommonUtil;
+
+import javax.annotation.Nullable;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+/**
+ * A utility class to read RPC Concepts and convert them into Grakn Concepts.
+ *
+ * @author Haikal Pribadi
```
kasper-piskorski: ditto

grakn-grpc/src/main/java/ai/grakn/rpc/util/RequestBuilder.java
```diff
@@ -0,0 +1,132 @@
...
+import ai.grakn.rpc.generated.GrpcIterator.IteratorId;
+import ai.grakn.rpc.generated.GrpcIterator.Next;
+import ai.grakn.rpc.generated.GrpcIterator.Stop;
+
+import javax.annotation.Nullable;
+
+/**
+ * A utility class to build RPC Requests from a provided set of Grakn concepts.
+ *
+ * @author Haikal Pribadi
```
kasper-piskorski: ditto

grakn-grpc/src/main/java/ai/grakn/rpc/util/ResponseBuilder.java
```diff
@@ -0,0 +1,240 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * A utility class to build RPC Responses from a provided set of Grakn concepts.
+ *
+ * @author Haikal Pribadi
```
kasper-piskorski: ditto

grakn-core/src/main/java/ai/grakn/graql/admin/Answer.java
haikalpribadi: renamed immutableMap() to just map(). "immutable" prefix is implementation detail, like HashMap, or HashSet - we always call for "Map".

grakn-core/src/main/java/ai/grakn/graql/admin/ReasonerQuery.java
haikalpribadi: When using intellij's refactoring tool (not just changing the code by "search and replace"), it highlights to you the blocks of code that will automatically be refactored when executed. I missed this block that I should have removed. I will patch this.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/ComputeQueryImpl.java
haikalpribadi: I actually rewrote the complete implementation of the new ComputeQuery class (which didn't exist before). But we've decided to remove authors now.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/AbstractQuery.java
haikalpribadi: Because we don't want Graql to depend on Printers. It makes no sense for the Graql classes to depend on Printers. Printers should take in Graql Answers. And thus, Printers depend on Graql, not the other way around. The current way causes complexity, which I will explain more in the code review.

grakn-core/src/main/java/ai/grakn/concept/ConceptId.java
```diff
@@ -21,34 +21,45 @@
...
- * </p>
+ * A class which represents an id of any {@link Concept} in the {@link GraknTx}.
+ * Also contains a static method for producing concept IDs from Strings.
  *
- * <p>
- *     A class which represents an id of any {@link Concept} in the {@link GraknTx}.
- *     Also contains a static method for producing concept IDs from Strings.
- * </p>
- *
- * @author Filipe Peliz Pinto Teixeira
```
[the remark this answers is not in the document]
haikalpribadi: renamed

grakn-core/src/main/java/ai/grakn/concept/ConceptId.java
```diff
@@ -21,34 +21,45 @@
...
+ * @author Grakn Warriors
  */
-@AutoValue
-public abstract class ConceptId implements Comparable<ConceptId>, Serializable {
-    private static final long serialVersionUID = -1723590529071614152L;
+public class ConceptId implements Comparable<ConceptId>, Serializable {
+
+    private final String value;
```
kasper-piskorski: I think if you make it @Nullable autovalue will make it a null by default which is the behaviour you want. Then you could add a create() method to get the non-argument constructor behaviour.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/QueryAnswer.java
```diff
@@ -60,7 +60,7 @@
  */
 public class QueryAnswer implements Answer {
 
-    private final ImmutableMap<Var, Concept> map;
+    private final Map<Var, Concept> map;
```
kasper-piskorski: I think the point of having an explicit `ImmutableMap` here was that you open the class and you can expect it to be immutable without looking at the details.

grakn-graql/src/main/java/ai/grakn/graql/internal/query/AbstractQuery.java
kasper-piskorski: But converters were part of graql anyway. That way we wouldn't have to have the type checking in `GraqlShell` for instance.

kasper-piskorski:
have we ensured there are no performance regressions?

grakn-core/src/main/java/ai/grakn/concept/ConceptId.java
haikalpribadi: Still doesn't work for the OLAP framework, i've tried

grakn-graql/src/main/java/ai/grakn/graql/internal/query/QueryAnswer.java
haikalpribadi: I'll change in the next PR

PR #2802, opened 2018-05-09
Dashboard ask query bug

Irtazaraza:
# Why is this PR needed?
-Ask query output in dashboard console has odd characters

# What does the PR do?
-Checks if the the query response contains the odd characters and trims the response string to exclude them

-No.

-Check at the back-end if the query is for the console page of the dashboard and does not send the unwanted characters along with the response.


Irtazaraza:
@marco-scoppetta 
There is an "OR" statement because the odd characters for 'True' and 'False' are different.

Issue #2811, opened 2018-05-21
Bootup issue - java.lang.RuntimeException: Couldn’t get PID of Grakn. Received '


lolski:
Users have reported that they are unable to start Grakn with the `java.lang.RuntimeException: Couldn’t get PID of Grakn. Received '` error.

This issue has been reported multiple times:
https://discuss.grakn.ai/t/starting-engine-failed/760
https://discuss.grakn.ai/t/starting-grakn-failed/723
https://discuss.grakn.ai/t/start-grakn-server-with-error-couldnt-get-pid-of-grakn-received/651

A user has reported one possible culprit:
“After debugging, found the root cause. The Grakn process startup command line’s java classpath is too long that exceeds the max length of CMD column of the “ps -ef | grep …” result.”

lolski:
Issue fixed

PR #2815, opened 2018-05-23
Bugfix for java.lang.RuntimeException: Couldn’t get PID of Grakn. Received '

lolski:
# Why is this PR needed?
Addresses issue https://github.com/graknlabs/grakn/issues/2811

# What does the PR do?
1. Replace `ps -ef | grep` with `ManagementFactory.getRuntimeMXBean().getName()`
2. If an Engine PID file is left but the process isn't running we know that it is a left-over PID file which can safely be overwritten.

No

N/A

### DIFF
--- a/grakn-bootup/src/main/java/ai/grakn/bootup/graknengine/pid/UnixGraknPidRetriever.java
+++ b/grakn-bootup/src/main/java/ai/grakn/bootup/graknengine/pid/UnixGraknPidRetriever.java
@@ -18,10 +18,7 @@
 
 package ai.grakn.bootup.graknengine.pid;
 
-import java.io.BufferedReader;
-import java.io.IOException;
-import java.io.InputStreamReader;
-import java.nio.charset.StandardCharsets;
+import java.lang.management.ManagementFactory;
 
 /**
  *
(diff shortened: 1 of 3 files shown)


marco-scoppetta:
what could go wrong?

grakn-bootup/src/main/java/ai/grakn/bootup/graknengine/pid/UnixGraknPidRetriever.java
```diff
@@ -32,36 +29,13 @@
...
-            } else {
-                throw new RuntimeException("a non-zero exit code '" + exitValue + "'returned by the command '" + psEfCommand + "'");
-            }
-        } catch (InterruptedException | IOException e) {
-            // DO NOTHING
-        }
-
-        String pidString = outputS.toString().trim();
+        String[] pidAndHostnameString = ManagementFactory.getRuntimeMXBean().getName().split("@");
+        String pidString = pidAndHostnameString[0];
```
kasper-piskorski: Do we know how portability is affected? isn't the name platform-specific? That's what the `RuntimeMXBean` spec says.

marco-scoppetta: @kasper-piskorski There exists no platform-independent way that can be guaranteed to work in all jvm implementations. ManagementFactory.getRuntimeMXBean().getName() looks like the best (closest) solution. It's short, and probably works in every implementation in wide use.

On linux+windows it returns a value like 12345@hostname (12345 being the process id).

PR #2828, opened 2018-05-31
Complexity Refactor: phase 2 (removing noise from Optionals)

haikalpribadi:
> Many parts of our code/architecture are unnecessarily complex. Complex code is hard to maintain and slow to improve. The consequence of a complex code can be very significant in terms of developer productivity and quality of life. It also puts the company and technology at risk when someone is unavailable/leaves (i.e. bus factor is too high). Thus, we need to simplify our code/architecture, to improve productivity, quality of life, and increase our bus factor to infinity (yes, no number of buses will bring Grakn down). This PR is part of many iterations of refactors we will start seeing a significant drop in package numbers, package dependencies, and thus, a drop in complexity of the architecture. Here's a nice 14 minutes talk on why maintaining a simple architecture is important, by Martin Fowler: https://www.youtube.com/watch?v=DngAZyWMGR0

# Why is this PR needed?

We use overuse `Optional`s way too much, such that it creates noise in our code. Not only it adds extra effort to understand when they are used to wrap too many things and combined with streams, it convolutes the code and sometimes even hide bugs to not be obvious. It's better to try as hard as possible to just use basic concepts, it forces us to keep our code simple.

# What does the PR do?

## Replaced Optional\<X\> with just X, as the type of field or parameter in many classes, and rewrote the code that implements it or depends on it

As IntelliJ clearly warns you: 
_Reports of any usesof Java.util.Optional<T>, as the type for a field or parameter. __Optional was designed to provide a limited mechanism for a library method to return types where there needed to be a clear way to represent "no result".__ Using a field with type java.util.Optional is also problematic if the class needs tobe Serializable, which java.util.Optional is not._

## An example of a bug we found, that was not obvious because of wrapped up in optionals: 
```
EmbeddedGraknTx<?> tx = optionalOr(optionalGraph, inner.tx()
    .filter(t -> t instanceof EmbeddedGraknTx)
    .map(t -> (EmbeddedGraknTx<?>) t))
    .orElseThrow(GraqlQueryException::noTx);

if (!RuleUtils.hasRules(tx)) return inner.stream(optionalGraph);
```
- The above code in `MatchInfer.java` should have ended with `inner.stream(tx)`. And it could have been easily caught/avoided if the code was written simply:
```
if (tx != null) embeddedTx = tx;
else if (inner.tx() instanceof EmbeddedGraknTx) embeddedTx = (EmbeddedGraknTx) inner.tx();
else throw GraqlQueryException.noTx();

if (!RuleUtils.hasRules(embeddedTx)) return inner.stream(embeddedTx);
```


haikalpribadi:
I'm closing this ticket for now, because we need to remove AutoValue could not accept Null objects as properties. Another reason to remove AutoValue classes, but I don't wanna work on that mess for now. @marco-scoppetta @lolski @kasper-piskorski

haikalpribadi:
damn, reopening it again, we can work around it. (Still, we gotta remove AutoValues for Query Classes. I don't mind keeping them for proper classes representing values). @marco-scoppetta @lolski @kasper-piskorski

haikalpribadi:
lol, scratch that, I can't work around the AutoValue limitation

grakn-graql/src/main/java/ai/grakn/graql/internal/gremlin/spanningtree/datastructure/FibonacciHeap.java
```diff
@@ -217,46 +228,54 @@ public int size() {
...
+
+        List<Entry> output = new LinkedList<>();
+        for (Entry entry : getCycle(oEntry)) {
+            if (entry != null) {
+                output.add(entry);
+                output.addAll(siblingsAndBelow(entry.oFirstChild));
+            }
+        }
+
+        return output;
```
marco-scoppetta: 👍

grakn-core/src/main/java/ai/grakn/GraknConfigKey.java
```diff
@@ -25,15 +25,14 @@
...
 import java.util.List;
-import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
  * Class for keys of properties in the file {@code grakn.properties}.
  *
  * @param <T> the type of the values of the key
- * @author Felix Chapman
```
kasper-piskorski: what happened here?

grakn-engine/src/main/java/ai/grakn/engine/controller/ConceptController.java
```diff
@@ -65,11 +64,9 @@
 import static org.apache.http.HttpStatus.SC_OK;
 
 /**
- * <p>
- *     Endpoints used to query for {@link ai.grakn.concept.Concept}s
- * </p>
+ * Endpoints used to query for {@link ai.grakn.concept.Concept}s
  *
- * @author Filipe Peliz Pinto Teixeira
```
kasper-piskorski: and here?

PR #2831, opened 2018-06-03
ComputeQueryImpl: extracting rules for default algorithm & argument into GraqlSyntax

haikalpribadi:
__This PR sits on top of PR #2828. Please make sure that PR is already merged first.__ Effectively this PR is just the last commit: f15c155.

# Why is this PR needed?

`ComputeQueryImpl` contain rules for default algorithm and arguments, which are specific behaviours of some instance(s) of Compute Query (in this case Centrality and Cluster). They need to be defined in `GraqlSyntax` instead. These are the hard-coded rules that need to be removed:

- Default algorithm check:
```
if (method.equals(CLUSTER) && algorithm == null) return Optional.of(ALGORITHMS_DEFAULT.get(method));
if (method.equals(CENTRALITY) && algorithm == null) return Optional.of(ALGORITHMS_DEFAULT.get(method));
```
- Default argument check:
```
if ((method.equals(CENTRALITY) || method.equals(CLUSTER)) && arguments == null) arguments = new ArgumentsImpl();
```
- Default `min_k` check:
```
if (method.equals(CENTRALITY) && algorithm.equals(K_CORE) && !argumentsOrdered.containsKey(MIN_K)) {
    return Optional.of(DEFAULT_MIN_K);
}
```
- Default `k` check:
```
if (method.equals(CLUSTER) && algorithm.equals(K_CORE) && !argumentsOrdered.containsKey(K)) {
    return Optional.of(DEFAULT_K);
}
```
- Default `members` check:
```
if (method.equals(CLUSTER) && algorithm.equals(CONNECTED_COMPONENT) && !argumentsOrdered.containsKey(MEMBERS)) {
    return Optional.of(DEFAULT_MEMBERS);
}
```

# What does the PR do?

- Replaced default algorithm check above with `GraqlSyntax.ALGORITHMS_DEFAULT`
- Replaced default argument check above with `GraqlSyntax.ARGUMENTS_DEFAULT`
- Replaced default `min_k`, `k`, and `members` checks from above with a new `getDefaultArgument()` helper method in `ComputeQueryImpl.Arguments`, which uses `GraqlSyntax.ARGUMENTS_DEFAULT`

No.

None.


grakn-core/src/main/java/ai/grakn/util/GraqlSyntax.java
```diff
@@ -200,6 +201,18 @@ public String toString() {
             return ImmutableMap.copyOf(accepted);
         }
 
+        private static Map<Method, Map<Algorithm, Map<Parameter, Object>>> argumentsDefault() {
+            Map<Method, Map<Algorithm, Map<Parameter, Object>>> defaults = new HashMap<>();
```
kasper-piskorski: if we are making the value maps immutable maybe let's make this map immutable as well?

haikalpribadi: The map is returned via `ImmutableMap.copyOf(...)` now.

PR #2843, opened 2018-06-15
Don't hardcode executable paths.

mgttlinger:
# Why is this PR needed?
`bash` does not reside under `/bin/` on all operating systems.

# What does the PR do?
- replaces `/bin/bash` shebangs with `/usr/bin/env bash` ones
- replaces hardcoded `"/bin/bash"` in java code with `"/usr/bin/env", "bash"`

Shouldn't

build/package using the [nix](https://nixos.org/nix/) language


mgttlinger:
I don't see the correspondence between the integration-test failure and my changes. The exceptions don't seem to suggest a missing executable either.

marco-scoppetta:
@mgttlinger , thank you for your PR. The tests pass. Just waiting for a colleague to review the changes. Also could you explain what problem is this PR suppose to fix? Thank you!

mgttlinger:
@marco-scoppetta The problem that `bash` is not specified to reside in `/bin/` and e.g. in NixOS those shebangs don't work because of that.

grakn-bootup/src/main/java/ai/grakn/bootup/BootupProcessExecutor.java
```diff
@@ -42,7 +42,7 @@
 public class BootupProcessExecutor {
 
     public static final long WAIT_INTERVAL_SECOND = 2;
-    public static final String SH = "/bin/sh";
+    public static final String SH = "/usr/bin/env sh";
```
lolski: @mgttlinger this change would not work because the program would try to find a binary named "env sh" rather than using the "sh" part as the argument to "env" correctly.

I would suggest creating two variables
```
public static final String ENV_BIN = "/usr/bin/env";
public static final String SHELL = "sh";
```
which will be used like so, from the other codes:
```
Arrays.asList(ENV_BIN, SHELL, "-c", "...")
```

lolski: Also, have a look at the Jenkins / Travis test in the section below, it would help you.

lolski:
@mgttlinger there are two problems with the latest commit:
1. There is a syntax error: `[ERROR] /home/ubuntu/workspace/grakn_PR-2843-UY3ALC3AWDTG6HJUS3PFRSQLTU6PV3HKHNQAVQQPDHII3YTYUDBQ/grakn-engine/src/main/java/ai/grakn/engine/bootup/BootupProcessExecutor.java:[45,52] ';' expected`
2. You will need to resolve the conflict by merging with the latest `master`

mgttlinger:
Sorry for the slow turnaround time but after the first failed CI build I lost confidence in your CI and given that I don't use grakn anymore didn't pay much attention.

lolski:
We're sorry to hear that, do give us a shout if you want to give Grakn a second chance in the future.

mgttlinger:
Of the databases, I evaluated that were capable of reasoning it was one of the nice ones. 

You have a pretty UI and the setup is straightforward but everything feels a bit unpolished. When entering commands in the UI the last error message was replayed even on successful commands; unsuccessful commands created nodes in the graph you couldn't interact with; and the Haskell client library is built to be extendable but doesn't expose crucial types and functions you would need to extend it e.g. to handle writing in a controlled fashion. It's these little annoyances that by themselves don't prevent usage of the database but give an impression of it not being ready for serious use.

marco-scoppetta:
We will consider adding this changes to the next release, but we have to test this change more to make sure it does not produce new issues on other OSs. Closing for now

PR #2845, opened 2018-06-18
Bug: Match.get(Var var) will crash on RemoteGraknSession

haikalpribadi:
# Why is this PR needed?

For some reason, we have 2 (odd) `Match.get(var)` behaviours:
1. `Match.get(Var var)`, which takes a single `Var` and returns `Stream<Concept>`, and
2. `Match.get(Var var, Var.. vars)`, which takes one or more `Var` and returns `Stream<Answer>`.

The critical issue is that: although the (2) `Match.get(Var var, Var.. vars)` (plural) executes fine over GRPC, the (1)`Match.get(var)` (singular) does not. It forces the `match` clause to execute an internal `AbstractMatch.stream()` which forces a cast of `RemoteGraknTx` into `EmbeddedGraknTx` (see MatchTx.java:51). This causes the query to crash as it hits a casting exception.

# What does the PR do?

While it's possible to try and fix the underlying casting issue, we're better off fixing the odd API issue. I.e. Just remove the singular (1)`Match.get(var)`, as we already have the plural (2) `Match.get(Var var, Var.. vars)` covering all cases.

Nope.

Note that the `ai.grakn.graql.internal.query.match` is way too convoluted to serve such a simple thing. The whole package will be simplified in future refactors.

### DIFF
--- a/grakn-core/src/main/java/ai/grakn/graql/Match.java
+++ b/grakn-core/src/main/java/ai/grakn/graql/Match.java
@@ -19,14 +19,12 @@
 package ai.grakn.graql;
 
 import ai.grakn.GraknTx;
-import ai.grakn.concept.Concept;
 import ai.grakn.graql.admin.Answer;
 import ai.grakn.graql.admin.MatchAdmin;
 
 import javax.annotation.CheckReturnValue;
 import java.util.Collection;
 import java.util.Set;
-import java.util.stream.Stream;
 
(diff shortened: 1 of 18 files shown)


haikalpribadi:
Good idea, I've made a PR on stable now (#2847), I'm closing this PR. @kasper-piskorski @marco-scoppetta

PR #2859, opened 2018-06-26
[Stable] fix unnecessarily caching types when adding specific type

kasper-piskorski:
Same as https://github.com/graknlabs/grakn/pull/2852 on master.


kasper-piskorski:
merging as it is the same as #2852 on master which was already revievew

PR #2862, opened 2018-06-28
[Stable] Fix query cache inconsistency problems

kasper-piskorski:
# Why is this PR needed?
Solves following problems:
- cache unifier eagerness problem - cache unifiers were initialised eagerly which could lead to invalid entries in query cache.
- equivalence of relation atoms with user defined type variables. Two atoms user- and non-user-defined type variables were treated as equivalent.
- dangling ids when converting queries to positive ones

# What does the PR do?
- Makes sure unifiers are retrieved lazily.
- Fixes equivalence check on relation atoms.
- Removes dangling ids on `positive()` conversion.
No.
None.


kasper-piskorski:
@marco-scoppetta re-review please

PR #2875, opened 2018-07-10
[stable] academy fix: match delete queries in the insert-delete-queries.md page

lolski:
# Why is this PR needed?
academy fix: match delete queries in the insert-delete-queries.md page

# What does the PR do?
Fixes [this ticket](https://work.grakn.ai/entity/20286-delete-syntax-is-outdated-in-academy): `match $x isa company has name "Grakn"; delete;` should be `match $x isa company has name "Grakn"; delete $x;`

No

N/A


kasper-piskorski:
why no squash?

lolski:
@kasper-piskorski good point.

PR #2893, opened 2018-07-17
Numerical Aggregate Fix

haikalpribadi:
# Why is this PR needed?

What we mean by Numerical Aggregates are `sum`, `min`, `max`, `mean`, `median` and `std`.
1. These aggregate functions return 0 when there are no attributes to apply the calculation over. This is wrong, it should return null.
2. These aggregate functions (except for `sum`) returns optional, which does not work with GRPC (before and after the new GRPC refactor).

# What does the PR do?

1. Removed `Optional<T>` return types from aggregate functions.
2. Modify the aggregate functions to return `null` when there are no attributes to calculate over.
3. Fix `RemoteQueryExecutor.run(AggregateQuery)` to consider `null` results.
4. Add tests for `null` results in `AggregateTest`
5. Add tests for numerical aggregate functions in `ServerRPCIT`

Additionally, we also did some cleaning:
1. Removed the first Generics parameter `T` in `Aggregate<T, S>`.
2. Added tests for ordering `String`s and `Date`s in `MatchTest.java`

Yes. Aggregate `min` and `max` no longer apply to `String` and `Date`. This is because making the change allows us to simplify numerical aggregates, while still being able to get the same functionality of `min/max string/date` through `order string/date asc/desc`.

This PR will be followed by another PR where we implement a proper GRPC message for Aggregate responses, and also fix the Remote `GroupAggregate` query which is currently broken too on the client side.

### DIFF
--- a/client-java/src/main/java/ai/grakn/client/executor/RemoteQueryExecutor.java
+++ b/client-java/src/main/java/ai/grakn/client/executor/RemoteQueryExecutor.java
@@ -31,6 +31,7 @@ import ai.grakn.graql.admin.Answer;
 import ai.grakn.client.Grakn;
 import com.google.common.collect.Iterators;
 
+import java.util.Iterator;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
@@ -76,7 +77,9 @@ public final class RemoteQueryExecutor implements QueryExecutor {
 
     @Override
     public <T> T run(AggregateQuery<T> query) {
-        return (T) Iterators.getOnlyElement(tx.query(query));
+        Iterator iterator = tx.query(query);
+        if (iterator.hasNext()) return (T) Iterators.getOnlyElement(iterator);
+        else return null;
     }
 
 
--- a/grakn-graql/src/main/java/ai/grakn/graql/internal/query/aggregate/MinAggregate.java
+++ b/grakn-graql/src/main/java/ai/grakn/graql/internal/query/aggregate/MinAggregate.java
@@ -30,7 +30,7 @@ import static java.util.Comparator.naturalOrder;
 /**
  * Aggregate that finds minimum of a {@link Match}.
  */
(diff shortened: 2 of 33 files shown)


haikalpribadi:
I don't know why Circle-CI is not showing all the tests on this page. But we can see them here: https://circleci.com/gh/haikalpribadi/grakn/tree/aggregate-numeric-fix

grakn-graql/src/main/java/ai/grakn/graql/internal/query/aggregate/MinAggregate.java
```diff
@@ -39,25 +36,29 @@
     }
 
     @Override
-    public Optional<T> apply(Stream<? extends Answer> stream) {
-        return stream.map(this::getValue).min(naturalOrder());
+    public Number apply(Stream<? extends Answer> stream) {
+        NumberPrimitiveTypeComparator comparator = new NumberPrimitiveTypeComparator();
```
jmsfltchr: Design not ideal, but probably a non-issue

jmsfltchr: Ideally the comparator instance would be passed in via constructor, however in this case this may make sense since the comparator implementation is unlikely to need to be swapped out for something else with the same interface in the future. Let's discuss!

client-java/src/main/java/ai/grakn/client/executor/RemoteQueryExecutor.java
```diff
@@ -76,7 +77,9 @@ public void run(UndefineQuery query) {
 
     @Override
     public <T> T run(AggregateQuery<T> query) {
-        return (T) Iterators.getOnlyElement(tx.query(query));
+        Iterator iterator = tx.query(query);
```
kasper-piskorski: is it better than not throwing exception?

PR #2902, opened 2018-07-22
[Stable] Add extra rule pruning

kasper-piskorski:
# Why is this PR needed?
To prune rules that even though applicable, yield no answers during resolution.
# What does the PR do?
Introduces an extra rule cache so that we don't keep retrying fruitless rules like maniacs.
No.
No?


marco-scoppetta:
It would be nice to have some explanation in the PR description

PR #2904, opened 2018-07-23
[Stable] GC docs update

kasper-piskorski:
# Why is this PR needed?

# What does the PR do?


kasper-piskorski:
Merging as docs only.

PR #2905, opened 2018-07-23
Removed <scope>test</scope> from grpc-netty

haikalpribadi:
# Why is this PR needed?

Java applications that use (and thus depend on) `client-java` does not "see" the `grpc-netty` dependency, which is required for their `Grakn.Session` to work (which is over GRPC). This is because we mistakenly scoped the `grpc-netty` dependency for just tests.

# What does the PR do?

Removed `<scope>test</scope>` from `pom.xml` of `client-java`.

### DIFF
--- a/client-java/pom.xml
+++ b/client-java/pom.xml
@@ -103,7 +103,6 @@
         <dependency>
             <groupId>io.grpc</groupId>
             <artifactId>grpc-netty</artifactId>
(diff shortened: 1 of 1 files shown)


haikalpribadi:
could you believe, @marco-scoppetta ?! this was the culprit!

PR #2906, opened 2018-07-24
[Stable] Rule cache lvl2 - WIP

kasper-piskorski:
# Why is this PR needed?
To reduce rule handling time.
# What does the PR do?
Cache rules applicable to schema concept and their conversion to `InferenceRule` object (parsing ain't cheap).
No.
Will see.


lolski:
Let's talk about how to invalidate the cache in case the rule is deleted later.

PR #2910, opened 2018-07-30
Unification of Query Answers through common data structures - Part 1

haikalpribadi:
# Why is this PR needed?

It has been difficult to maintain and guarantee that various type of Graql Queries (e.g. get, insert, delete, aggregates min/max/sum/groups/etc, compute cluster/path/centrality/statistics and a few others) is always maintained through remote communication APIs (previously REST, now GRPC). Prior to this PR, `AggregateQuery` and `ComputeQuery` are error-prone and not always supported over GRPC. The main reason for this problem is because we have so many different return types for every query, that it becomes hard to guarantee that all queries are supported in a generic and reliable way.

In order to solve this problem, we need to consider how to unify all queries to return a common set of data structures. By guaranteeing that every query returns a common set of data structures, we can guarantee that every query is supported over the remote communication API (i.e. GRPC) because we have implemented the generic and common data structures.

Because of the nature of the problem, the solution will be a pretty big change in the code base. Thus, we will split it into multiple PRs and this is the first one.

# What does the PR do?

The data structures we have identified so far are:
- `Value` :  a numerical value, i.e a Number (`int`, `long`, `double`, `float`)
- `ConceptMap`: a Map of Concepts
- `ConceptList`: a List of Concepts
- `ConceptSet`:  a Set of Concepts
- `ConceptSetMeasure`: a Set of Concepts associated with a "measurement" value. 
* (`ConceptList`, `ConceptSet`, and `ConceptSetMeasure` are currently implemented using `ConceptId`s, but will be replaced in the next PR).

Each of the data structure above is unified by an `Answer` interface, that allows us to provide an inference `Explanation` to every answer. We only provide them for `ConceptMap`s for now, but it is likely we would want to provide them for answers of different data structures as well in the future.

In order to implement the above data structures across all queries, we need to start with (what this PR contains): 
- Removed all optional return values from Aggregate functions
- Make Aggregate functions return null where calculations are appropriate
- Implement all data structures as an extension of a super `Answer` class
- Change how all `ComputeQuery` methods return their results through the various new `Answer`s
- Make ComputeQuery stream their results, and thus over GRPC too
- Change how all numerical `AggregateQuery`s return their results through `Value` answer
- We also needed to remove several `AggregateQuery`s because it does not fit with any data structure, and does not deliver a strong and unique use case:
    - `AskAggregate`
    - `SelectAggregate`
    - `NamedAggregate` (used only for `SelectAggregate`)

The removal of some `AggregateQuery`s breaks backwards compatibility.

Yes, plenty, such as (not limited to):
- Implement a data structure for GroupAggregate
- Simplify Java Generics parameter of Graql Query classes by unifying all `.stream()` to return `Answer`, and `execute()` to return `List<Answer>`
- Make analytics return `Concept` rather than just `ConceptId`


haikalpribadi:
I'm not sure why not all the tests are showing here, perhaps because it ran on my fork prior to the PR being made. In case it doesn't show up tomorrow, you can see here @marco-scoppetta @kasper-piskorski https://circleci.com/workflow-run/c97c44e2-27d0-4caa-93c6-f8b3d7131374

haikalpribadi:
I'll merge this to speed up, but you can leave any comments here and I'll add them to the subsequent PR @kasper-piskorski

PR #2916, opened 2018-08-01
Fixed RemoteRelationship.rolePlayersMap()

haikalpribadi:
# Why is this PR needed?

When we perform `match $x isa some-relationship; get;`, and `some-relationship` has multiple role players for the same role, the query will throw a null pointer exception. This is because the in the construction of the `Map<Role, Set<Thing>>`, where `Set<Thing>` represents the role players for a given `Role`, the `Set` was initialised with an IMMUTABLE set through `Collections.singleton(..)`.

# What does the PR do?

Fixed RemoteRelationship.rolePlayersMap() to be able to add multiple players to the same role.


lolski:
👍 regression test

PR #2918, opened 2018-08-02
Client python dev

flyingsilverfin:
# Why is this PR needed?
Merge client-python with master

# What does the PR do?
Merge separate development into master

Nope

* Date attribute test
* Test for match...insert query string


client-python/grakn/service/Keyspace/KeyspaceService.py
```diff
@@ -0,0 +1,43 @@
...
+# specific language governing permissions and limitations
+# under the License.
+#
+
+import grpc
+
+from grakn.service.Keyspace.autogenerated.Keyspace_pb2_grpc import KeyspaceServiceStub
+import grakn.service.Keyspace.autogenerated.Keyspace_pb2 as keyspace_messages
+
+class KeyspaceService(object):
```
haikalpribadi: Why is this class called `KeyspaceService`? The "service" is provided by the server (`grakn-engine`). The client "consumes" the service and provides the functionality to the user. This effectively serves as a `KeyspaceManager` and should probably be named that. But because we like to find a one-word name for these most outward facing APIs, @marco-scoppetta and I thought `Keyspaces` is a good compromise for a name. @flyingsilverfin @lolski

client-python/grakn/service/Session/TransactionService.py
```diff
@@ -0,0 +1,160 @@
...
+
+import queue
+from typing import Type
+
+from grakn.service.Session.util.RequestBuilder import RequestBuilder
+import grakn.service.Session.util.ResponseConverter as ResponseConverter # for circular import issue
+from grakn.service.Session.util import enums
+from grakn.exception.ClientError import ClientError
+
+class TransactionService(object):
```
haikalpribadi: Following the same logic in the feedback given for `KeyspaceService`, `TransactionService` should just be called `Transaction`. 

ps: Where's the `Session` class?

@flyingsilverfin @lolski

client-python/grakn/service/Session/util/ResponseConverter.py
```diff
@@ -0,0 +1,384 @@
...
+# under the License.
+#
+
+import abc
+from grakn.service.Session.util import enums
+from grakn.service.Session.Concept import ConceptFactory
+from grakn.exception.ClientError import ClientError
+
+
+class ResponseConverter(object):
```
haikalpribadi: We could probably find a better name than`*Converter`. The term "converter" is too generic and used way too often. It is not immediately clear to some readers/maintainers, which direction / who are is the converter converting it to? What is it converting it to? 

Or, is "converter" the most specific we could get it to? Then perhaps the methods contained in the class could be separated to create more cohesive classes with a more meaningful name. @flyingsilverfin @lolski

client-python/grakn/service/Session/Concept/ConceptHierarchy.py
```diff
@@ -0,0 +1,579 @@
...
+from grakn.service.Session.util.RequestBuilder import RequestBuilder
+# from grakn.service.Session.util.ResponseConverter import ResponseConverter # cannot get name from module because circular imports
+# import grakn.service.Session.util.ResponseConverter as ResponseConverter # toplevel only allowed
+from grakn.service.Session.util import ResponseConverter
+from grakn.service.Session.Concept import ConceptFactory
+from grakn.exception.ClientError import ClientError 
+
+
+
+class Concept(object):
```
haikalpribadi: Why is this class not just called `Concept.py` since all the nested classes are effectively just further extensions of the `Concept` class? @flyingsilverfin @lolski

client-python/grakn/exception/ClientError.py
```diff
@@ -0,0 +1,21 @@
...
+# 
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+class ClientError(Exception):
```
haikalpribadi: Is there a special reason for why we're naming this class `ClientError`? If this class is extending `Exception`, in order to represent exceptions in the `Grakn` client driver, so that we can inform the user about an exception from their `Grakn` driver, perhaps we should call it `GraknException`? @flyingsilverfin @lolski

client-python/grakn/service/Keyspace/KeyspaceService.py
flyingsilverfin: Agreed to keep KeyspaceService

client-python/grakn/service/Session/TransactionService.py
flyingsilverfin: Agreed to keep TransactionServiec

Session is in the __init__.py

client-python/grakn/service/Session/util/ResponseConverter.py
flyingsilverfin: Rename to ResponseReader

client-python/grakn/service/Session/Concept/ConceptHierarchy.py
flyingsilverfin: Yeah can be renamed to Concept.py

client-python/grakn/exception/ClientError.py
flyingsilverfin: Python convention is `*Error` which I will rename to: `GraknError`

PR #2932, opened 2018-08-07
fix batch loading due to incorrectly pointing to gRPC URL instead of HTTP

lolski:
# Why is this PR needed?

# What does the PR do?


client-java/src/main/java/ai/grakn/client/Grakn.java
```diff
@@ -74,6 +74,9 @@
 public final class Grakn {
     public static final SimpleURI DEFAULT_URI = new SimpleURI("localhost:48555");
 
+    @Deprecated
```
marco-scoppetta: fancy!

lolski: propa Javah

PR #2941, opened 2018-08-09
inlcude workbase in main repo

marco-scoppetta:
# Why is this PR needed?

# What does the PR do?


haikalpribadi:
Do the commit histories from the workbase repo matter to you guys, @marco-scoppetta , @Irtazaraza ? It got me thinking because it's 34k commit. It's actually possible to merge all the commits from a separate repository into a folder into this repository. Check this out: http://gbayer.com/development/moving-files-from-one-git-repository-to-another-preserving-history/
cc: @lolski what do you think?

marco-scoppetta:
@haikalpribadi there are quite a few files or folders we cab remove, we will clean up in following commits after reviewing file by file and no I don't think we need to maintain history

PR #2956, opened 2018-08-15
[Workbase] - Implicit relationship bug

Irtazaraza:
# Why is this PR needed?
Implicit relationships were being loaded to the graph

# What does the PR do?
Filter outs all implicit types

No

No


workbase/src/renderer/components/DataManagement/DataManagementUtils.js
```diff
@@ -59,9 +74,19 @@ async function loadRolePlayers(relationship, limitRolePlayers, limit, offset) {
...
+          await prepareEntity(thing);
+          break;
+        case 'ATTRIBUTE':
+          await prepareAttribute(thing);
+          break;
+        case 'RELATIONSHIP':
+          await prepareRelationship(thing);
+          break;
+        default:
+          // do nothing
```
marco-scoppetta: `throw "Unrecognised baseType of thing ["+thing.baseType+"]"` we should never get anything that is not entity/rel/attribute - if we ever do let's throw exception so that these things don't go unnoticed

workbase/test/unit/components/DataManagement/DataManagementUtils.test.js
```diff
@@ -177,3 +177,12 @@ describe('loadAttribtues', () => {
...
   });
 });
+
+describe('filter implicit types', () => {
+  test('implicit relationship', async () => {
+    let concepts = [mockConcepts.getMockImplicitRelationship()];
+    expect(concepts).toHaveLength(1);
+    concepts = await DataManagementUtils.filterImplicitTypes(concepts);
+    expect(concepts).toHaveLength(0);
+  });
```
marco-scoppetta: YAAAYY testssss 👍 💯

PR #2968, opened 2018-08-20
Add rule pruning for relation rules based on attributes

kasper-piskorski:
# Why is this PR needed?

# What does the PR do?


grakn-core/src/main/java/ai/grakn/graql/admin/UnifierComparison.java
```diff
@@ -53,4 +53,12 @@
      * @return true if typing the typeVar with type is compatible with role configuration of the provided query
      */
     boolean typePlayability(ReasonerQuery query, Var var, Type type);
+
+    /**
+     *
+     * @param parent
+     * @param child
+     * @return
+     */
```
marco-scoppetta: 👍

PR #2971, opened 2018-08-22
[Workbase] - Refactor Data Management

Irtazaraza:
# Why is this PR needed?
Data Management was becoming increasingly complex and hard to follow.

# What does the PR do?
-Separate all node and edge construction into a separate class: VisualiserGraphBuilder
-Load neighbours using a query
-Load attributes using a query

No

-


workbase/src/renderer/components/DataManagement/DataManagementUtils.js
```diff
@@ -1,239 +1,24 @@
...
+function loadNeighbours(node, neighboursLimit) {
+  let query;
   switch (node.baseType) {
     case 'ENTITY_TYPE':
     case 'ATTRIBUTE_TYPE':
     case 'RELATIONSHIP_TYPE':
-      return showTypeInstances(node, neighboursLimit, offset);
     case 'ENTITY':
-      return showRelationshipsOfEntity(node, neighboursLimit, offset);
+      query = `match $x id "${node.id}"; $y isa entity; $r ($x, $y); offset ${node.offset}; limit ${neighboursLimit}; get;`;
```
marco-scoppetta: case 'ENTITY' : return `ciao`;
case 'RELATIONSHIP' : return `ciao1`;
case 'ATTRIBUTE' : return `ciao2`;

PR #2973, opened 2018-08-23
[workbase] - tweak failing e2e tests [Do not merge]

Irtazaraza:
# Why is this PR needed?
e2e would fail irregularly due to timing constraints 

# What does the PR do?
increase the time out before failing tests

no

n/a


Irtazaraza:
e2e tests will be updated as per the new Visualiser UI

PR #2974, opened 2018-08-23
The Long Goodbye

lolski:
The Long Goodbye

# Why is this PR needed?
The old post-processing was neither **real-time** nor **fault-tolerant** and may lose data if it crashes in the middle of deduplication. 

# What does the PR do?

### 1. Added the new real-time, fault-tolerant post-processing called the "Attribute Deduplicator"
It works in real-time and can resume where it left off if Grakn crashes in the middle of a deduplication process.

#### How It Works
There are three main components in Attribute Deduplicator: the [AttributeDeduplicatorDaemon](https://github.com/graknlabs/grakn/pull/2974/files#diff-c39b3d8862f8074930200ae221705f7fR43), [RocksDbQueue](https://github.com/graknlabs/grakn/pull/2974/files#diff-85f4e76f337221a9068d119eb65f109eR51), and [AttributeDeduplicator](https://github.com/graknlabs/grakn/pull/2974/files#diff-c39b3d8862f8074930200ae221705f7fR43) components.

`AttributeDeduplicatorDaemon` is a daemon which runs in the background listening for incoming new attributes. As new attributes are received, the daemon sends the attribute into a persistent queue (`RocksDbQueue`). 

The `AttributeDeduplicatorDaemon` has another thread which dequeues items as they come. The dequeued items are processed by the `AttributeDeduplicator`, which is where the deduplication actually takes place.

`RocksDbQueue` is a persistent queue and that is what provides fault-tolerance. If we don't store the item in a persistent queue, we won't know where to resume processing if Grakn crashes during processing. With a persistent queue, we only delete items from the queue once they are deduplicated. If the deduplication crashes in the middle, you can simply re-process the un-acked items.

#### Tests
- [AttributeDeduplicatorIT](https://github.com/graknlabs/grakn/pull/2974/files#diff-bed173517b87d99a2d2e1caff78ad724R29) contains the tests for the `AttributeDeduplicator` class, which contains the deduplicate algorithm
- [AttributeDeduplicatorE2E](https://github.com/graknlabs/grakn/pull/2974/files#diff-14bd928eed562ad881df899ee4a7f345R45) contains the test for the whole algorithm against the TinkerGraph

### 2. Removing the "unique index" guard when inserting attributes:
Currently we have two "guards" which attempt to maintain attribute uniqueness, the **unique index** and the **write-if-not-exists** check. Each of them only works roughly 90% of the time as they're not immune to race-conditions and eventual consistency. We disable the unique index but keep write-if-not-exists. Ideally, we should remove them so we avoid the situation of having too many components working in an indeterministic way. However I am keeping the write-if-not-exists check as it'll break the genealogy tests.

- disabling uniqueness index: [here](https://github.com/graknlabs/grakn/pull/2974/files#diff-bd2d55000d2e304c4901240c7d595d48R22), and [here](https://github.com/graknlabs/grakn/pull/2974/files#diff-e5d85ee73a9e652fd5224355214edb36L67)

### 3. Removed post-processing + Redis completely
This means that from now on Grakn only depends on Cassandra.

No

Remove the write-if-not-exists check.


grakn-core/src/main/java/ai/grakn/util/REST.java
```diff
@@ -56,6 +56,7 @@ public static String reformatTemplate(String pathTemplate) {
         public static final String KEYSPACE_GRAQL = "/kb/:keyspace/graql";
         public static final String KEYSPACE_EXPLAIN = "/kb/:keyspace/explain";
         public static final String COMMIT_LOG_URI = "/kb/:keyspace/commit_log";
+        public static final String MERGE_URI = "/merge";
```
lolski: TODO: remove

grakn-engine/src/main/java/ai/grakn/engine/attribute/deduplicator/AttributeDeduplicator.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.tinkerpop.gremlin.structure.Property;
+import org.apache.tinkerpop.gremlin.structure.Vertex;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+
+/**
+ *
+ * TODO
```
lolski: TODO: fix

grakn-engine/src/main/java/ai/grakn/engine/attribute/deduplicator/queue/Attribute.java
```diff
@@ -16,31 +16,23 @@
...
+import ai.grakn.concept.ConceptId;
+import com.google.auto.value.AutoValue;
 
 /**
- * <p>
- *     Defines the API which must be implemented in order to be able to run the task in the background
- * </p>
- *
- * @author Filipe Peliz Pinto Teixeira
+ * TODO
```
lolski: TODO: fix

grakn-engine/src/main/java/ai/grakn/engine/attribute/deduplicator/queue/QueueException.java
```diff
@@ -16,17 +16,14 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-package ai.grakn.engine.data;
+package ai.grakn.engine.attribute.deduplicator.queue;
 
 /**
- * The {@link QueueSanityCheck} class is responsible for performing sanity check of the Queue component
- *
+ * TODO
```
lolski: TODO: fix

grakn-engine/src/main/java/ai/grakn/engine/controller/GraqlController.java
```diff
@@ -241,6 +241,59 @@ private static void handleError(int status, Exception exception, Response respon
         response.type(ContentType.APPLICATION_JSON.getMimeType());
     }
 
+//    /**
```
lolski: TODO: fix

grakn-engine/src/test/java/ai/grakn/engine/ScratchpadTest.java
```diff
@@ -0,0 +1,164 @@
+package ai.grakn.engine;
```
lolski: TODO: remove

grakn-test/test-distribution/src/test/java/ai/grakn/distribution/AttributeDeduplicatorE2E.java
```diff
@@ -0,0 +1,180 @@
...
+import static ai.grakn.distribution.DistributionE2EConstants.assertGraknStopped;
+import static ai.grakn.distribution.DistributionE2EConstants.assertZipExists;
+import static ai.grakn.distribution.DistributionE2EConstants.unzipGrakn;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static ai.grakn.graql.Graql.count;
+import static ai.grakn.graql.Graql.label;
+import static ai.grakn.graql.Graql.var;
+
+public class AttributeDeduplicatorE2E {
```
lolski: TODO: review

grakn-test/test-integration/src/test/java/ai/grakn/test/engine/attribute/deduplicator/AttributeDeduplicatorIT.java
```diff
@@ -0,0 +1,336 @@
+package ai.grakn.test.engine.attribute.deduplicator;
```
lolski: TODO: review

.ruby-version
```diff
@@ -0,0 +1 @@
+2.4.2
```
lolski: TODO: remove

PR #2979, opened 2018-08-29
Updated deployment workflow

marco-scoppetta:
# Why is this PR needed?

To have a working deployment workflow

# What does the PR do?

Updates CircleCI config to have a working deployment workflow, triggered by tagged commits (vx.y.z)
The build+test workflow is not triggered when pushing tagged commits.
Updated Workbase `package.json`

no
Add Github personal token to CircleCI `graknlabs` in order to enable auto-creation of draft releases


.circleci/config.yml
```diff
@@ -239,45 +239,76 @@ jobs:
...
 workflows:
   version: 2
   run-all-tests:
     jobs:
-      - build
-      - workbase-unit-tests
+      - build:
+          filters:
+            tags:
+              ignore: /.*/
```
lolski: question: what are we ignoring here?

marco-scoppetta: We ignore the workflow when we attach a tag to the current commit. 
We currently run all the tests every time someone submits a PR, so when we push the a commit with only a tag attached to it (and no changes) we only run the deployment workflow and we don't re-run all the tests again

PR #2980, opened 2018-08-30
The Cassandra Crossing (from 2.x to 3.11.3)

marco-scoppetta:
Crossing

# Why is this PR needed?

We were still relying on Cassandra 2.x

# What does the PR do?

Settings, dependencies and classes tweaks to accommodate the latest version of Cassandra:

 - renamed `cassandra-logback.xml` to `logback.xml` so that other nested dependencies can find the file and we can control their logging level from the file (netty needs this). The file is contained inside the cassandra folder so there is no ambiguity for the name

- added `hints_directory` and `cdc_raw_directory` to `cassandra.yaml` - to make Cassandra 3 happy

- removed `slf4j-nop` from `grakn-factory` as it was silencing too many logs and some exception from CassandraEmbedded were hidden (fun times)

- updated hadoop configuration in `grakn-factory/src/main/resources/default-configs.properties` (which is the file we feed to GraknSparkComputer) to use new input format class `ai.grakn.kb.internal.computer.GraknCassandra3InputFormat`

- overridden 3 classes from JanusGraph to completely remove any dependency from Cassandra 2 `ColumnFamilyInputFormat` - which was still giving us `ClassNotFound` exception when using Janusgraph classes

- updated versions in `pom.xml` and also updated exclusions (FUUUNN TIMESS)

- minor tweaks here and there to make everyone happy

nope

We can now use Cassandra 3 but we are still using it through the old Astyanax driver - we should at some point move to CQL driver. For now CQL driver breaks too many things in our beautiful caches - so probably worth trying to switch to that after we have removed all caching mechanisms


grakn-factory/src/main/java/ai/grakn/kb/internal/GraknTxJanus.java
```diff
@@ -74,7 +75,11 @@ public int numOpenTx() {
 
     @Override
     public void clearGraph() {
-        JanusGraphCleanup.clear(getTinkerPopGraph());
+        try {
+            JanusGraphCleanup.clear(getTinkerPopGraph());
+        } catch (BackendException e) {
+            e.printStackTrace();
```
lolski: Is it perhaps better to rethrow: `throw new RuntimeException(e)` ?

marco-scoppetta: yes it’s a wip - thats why nobody is still assogned. also there is a PR waiting for you

PR #2981, opened 2018-09-03
WIP - Attribute deduplicator remove unnecessary log

lolski:
# Why is this PR needed?
There are some leftover post-processing things that weren't removed properly.

# What does the PR do?
1. removing a scary-sounding warning, which isn't supposed to be a warning at all
2. removed an unneeded function `fixDuplicateResources()` and other functions which made it up.
3. removed PostProcessingTest
4. removed old post-processing logs

No

N/A


marco-scoppetta:
@lolski - sorry I just noticed you had added "wip" to the title, need to revert or you can add on top of this one?

PR #2993, opened 2018-09-12
[WIP] Open Windows an try Grakn

marco-scoppetta:
Open Windows
# Why is this PR needed?
Windows was not supported
# What does the PR do?
Add full support to Windows


grakn-engine/src/main/java/ai/grakn/engine/bootup/BootupProcessExecutor.java
```diff
@@ -140,24 +144,30 @@ public void stopProcessIfRunning(Path pidFile, String programName) {
...
+            return Arrays.asList(SH, "-c", "kill " + pid.trim());
+        }
     }
 
-    private BootupProcessResult kill(int pid, String signal) {
-        return executeAndWait(Arrays.asList(SH, "-c", "kill -" + signal + " " + pid), null);
+    private void kill(String pid) { executeAndWait(killProcessCommand(pid), null); }
+
+    private boolean isWindows() {
+        return System.getProperty("os.name").toLowerCase().contains("win");
```
kasper-piskorski: do we know what will happen if run under cygwin?

grakn-engine/src/main/java/ai/grakn/engine/bootup/StorageBootup.java
```diff
@@ -116,34 +119,25 @@ public boolean isRunning() {
 
     /**
      * Attempt to start Storage and perform periodic polling until it is ready. The readiness check is performed with nodetool.
-     *
+     * <p>
```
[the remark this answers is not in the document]
marco-scoppetta: this is by Intellij, when you use the auto-linting function to improve code, so I am not sure what's the corrrect way to go 🤔

README.md
```diff
@@ -46,15 +46,15 @@ With the expressivity of the schema, inference through OLTP and distributed algo
...
-- Yarn (in order to build Dashboard)
+
+- If running on Windows version prior to 10, please make sure to have [Visual Studio C++ Runtime for Visual Studio 2015](https://www.microsoft.com/en-us/download/details.aspx?id=48145) installed
 
 You can build Grakn using Maven:
 ```
 $ mvn package -DskipTests
 ```
-
+NOTE: you need to have [Yarn](https://yarnpkg.com) installed in order to compile Grakn from source.
```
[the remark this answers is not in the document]
marco-scoppetta: Yarn requires nodejs, so I figured it would be enough to say Yarn

grakn-engine/src/main/java/ai/grakn/engine/bootup/BootupProcessExecutor.java
marco-scoppetta: nope, we should test it at some point. I took this check directly from another library - I will check if there is a more thorough and correct check

marco-scoppetta: apparently it yields `posix` https://stackoverflow.com/questions/32612235/what-is-os-name-on-cygwin

Issue #4354, opened 2018-09-22
Add a delay on hover animation on the graph


haikalpribadi:
The graph visualiser right now "highlights" the hovered sugraph, and dims the rest of the graph. However, the hover animation is too sensitive. This results in quite an uncomfortable experience. If we give a delay of 1 second for a mouse to be hovering on a concept before we highlight the subgraph around the concept, this would make the animation smoother.

PR #4362, opened 2018-09-24
Add shortcut to clear graph and add delay before dimming of nodes when hovering on a node

Irtazaraza:
# Why is this PR needed?
Issue: #4354
1) User could only clear graph by clicking on a button
2) Hovering animation on nodes was too sensitive

# What does the PR do?
1) Allow user to clear graph by doing 'cmd' + 'g'
2) Add a 500 ms delay before dimming all other nodes when hovering on a specific node

No

Modify clearing graph shortcut

### DIFF
--- a/workbase/src/renderer/components/Visualiser/TopBar/GraqlEditor/GraqlEditor.vue
+++ b/workbase/src/renderer/components/Visualiser/TopBar/GraqlEditor/GraqlEditor.vue
@@ -217,12 +217,6 @@
     },
     created() {
       this.renderIcons();
-
-      window.addEventListener('keydown', (e) => {
-        const gKey = 71;
-        // meta key -> cmd
-        if ((e.keyCode === gKey) && e.metaKey) { this.clearGraph(); }
-      });
     },
     computed: {
       currentQuery() {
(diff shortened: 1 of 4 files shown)


Irtazaraza:
#4354

workbase/src/renderer/components/Visualiser/TopBar/GraqlEditor/GraqlEditor.vue
```diff
@@ -217,6 +217,12 @@
     },
     created() {
       this.renderIcons();
+
+      window.addEventListener('keydown', (e) => {
+        const gKey = 71;
```
marco-scoppetta: Can we move this const at the top of the file? 
If you declare it inside the function it gets re-declared every time the keydown event is triggered.
Also let's use the convention for global constants, which is all CAPS, smth like: 
`const LETTER_G_KEYCODE = 71`

Irtazaraza: I also moved it to the visualiser store

Issue #4370, opened 2018-09-25
Remove watchers and computed properties based on `isInit()` from all components


marco-scoppetta:
All the components should register events handlers in the `created` or `mounted` method and let `localStore` figure out when it's the right time to attach them to the visualiser facade

PR #4378, opened 2018-09-26
refactor canvas event handling

Irtazaraza:
# Why is this PR needed?
Issue - #4370

# What does the PR do?
-Remove isInit flag and replace with eventCache to keep track of unregistered events.
-Move all event registering from watchers to created in vue child components.

No

N/A

### DIFF
--- a/workbase/src/renderer/components/shared/CanvasStoreMixin.js
+++ b/workbase/src/renderer/components/shared/CanvasStoreMixin.js
@@ -11,6 +11,8 @@ import {
   LOAD_METATYPE_INSTANCES,
 } from './StoresActions';
 
+const eventCache = {};
+
 const actions = {
   async [CURRENT_KEYSPACE_CHANGED](keyspace) {
     if (keyspace !== this.currentKeyspace) {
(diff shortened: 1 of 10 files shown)


workbase/src/renderer/components/shared/CanvasStoreMixin.js
```diff
@@ -79,7 +84,11 @@ const methods = {
   // Setters
   setSelectedNodes(nodeIds) { this.selectedNodes = (nodeIds) ? this.visFacade.getNode(nodeIds) : null; },
   setSelectedNode(nodeId) { this.selectedNodes = (nodeId) ? [this.visFacade.getNode(nodeId)] : null; },
-  registerCanvasEventHandler(event, fn) { this.visFacade.registerEventHandler(event, fn); },
+  registerCanvasEventHandler(event, fn) {
+    if (this.visFacade) { debugger; this.visFacade.registerEventHandler(event, fn); } else {
```
marco-scoppetta: 👯 👯 👯 👯

PR #4410, opened 2018-10-03
Refactor Context Menu

Irtazaraza:
# Why is this PR needed?
closes: #4413
Context menu was not synchronised with the refactor of vuex architecture and did not work 

# What does the PR do?
- Move context menu logic outside of the component
- Store showContextMenu and contextMenuOptions in the vuex state
- e2e tests for context menu

No

N/A


workbase/src/renderer/components/Visualiser/VisualiserCanvasEventsHandler.js
```diff
@@ -39,7 +42,20 @@ export default {
           commit('selectedNodes', null);
           state.visFacade.getNetwork().unselectAll();
         }
-      } });
+
+        /* Context Menu */
```
marco-scoppetta: No need to use comments, you can just add another:
```
commit('registerCanvasEvent', {
       event: 'oncontext',	       event: 'oncontext',
       callback: (params) => {	       callback: (params) => {
        ....handle context menu actions here     
}}
```

workbase/src/renderer/components/Visualiser/ContextMenuUtils.js
```diff
@@ -0,0 +1,27 @@
+import dataVisualiserStore from '@/components/Visualiser/store';
```
marco-scoppetta: instead of importing the store, it's probably better to pass the `state` as a param to the methods:

`function verifyEnableExplain( state ){ }`

it's also easier to test (you can pass a mock object to the functions instead of a real state)

workbase/src/renderer/components/shared/StoresActions.js
```diff
@@ -16,6 +16,7 @@ export const RUN_CURRENT_QUERY = 'run-current-query';
 export const EXPLAIN_CONCEPT = 'explain-concept';
 export const UPDATE_NODES_LABEL = 'update-nodes-label';
 export const UPDATE_NODES_COLOUR = 'update-nodes-colour';
+export const DELETE_SELECTED_NODES = 'delete-nodes';
```
marco-scoppetta: let's name it `delete-selected-nodes`

workbase/src/renderer/components/Visualiser/store/getters.js
```diff
@@ -9,4 +9,6 @@ export default {
   selectedNode: state => ((state.selectedNodes) ? state.selectedNodes[0] : null),
   canvasData: state => state.canvasData,
   isActive: state => (state.currentKeyspace !== null),
+  contextMenu: state => state.contextMenu,
+  visFacade: state => state.visFacade,
```
marco-scoppetta: remote it from getters

Issue #4433, opened 2018-10-07
Client Python: attribute.data_type() doesn't return a string


jmsfltchr:
The Client Python docs pledge that `my_attribute.data_type()` should return a string, but it doesn't. Presently it returns an object with an attribute `name` that holds the capitalised name of the data type.

At present I do the following to get the result I expected to see:
```
my_attribute.data_type().name.lower()
```

flyingsilverfin:
resolution: Docs are wrong. Returns an enum that is a grakn.DataType, same as during creation.

needed by #4240

Issue #4240, opened 2018-09-22
Update drivers documentation to explain better a few concepts


grabl:
_This issue was originally posted by @marco-scoppetta  on 2018-08-22 14:12_.

1. Specify that if a transaction fails (throws exception) it gets automatically closed 2. Specify that concepts are only usable within the tx used to obtain them (maybe snippet example) Python docs: python client doc broken example - answer_iterator

PR #4446, opened 2018-10-09
Client python 1.4.2 update

### LINKED ISSUE #4423: Client Python: file structure naming is unclear
To use the class `Concept` in Python it is necessary to import as follows:
```
from grakn.service.Session.Concept import Concept

is_a_concept = isInstance(my_concept, Concept.Concept)
```
From a user perspective this is confusing since there is a package called `Concept`, containing a file called `Concept`, containing a class called `Concept`.

I believe ideal best practice would be to use lower case names for packages, and that they should be named according to a summary of their content. I don't know whether this would conflict with our naming across our other drivers but something should be done to limit the confusion here if possible.

flyingsilverfin:
# Why is this PR needed?
Update some docs that haven't been changed for a while

Addresses #4240 and #4433 

# What does the PR do?
Bump version number and change README

no

(for v1.6): rename modules/folders to fix #4423


flyingsilverfin:
@marco-scoppetta I've added a commit for the `Concept.id` doc, but decided not to explain base_type (unlikely people will be needing it directly, mostly used via `Concept.is_relationship()`... etc), even though its unlikely to change, it's still tied to GRPC at the minute

marco-scoppetta:
@flyingsilverfin yeah it makes sense!

PR #4477, opened 2018-10-12
Add pre-commit check for unwanted code

Irtazaraza:
# Why is this PR needed?
So that the developer does not commit:
1) un-eslinted files
2) debugger
3) console.log()

# What does the PR do?
1) Import husky and lint-stager to check code before committing 
2) Will only display the files where debugger and console.log are present (Currently will not stop the commit) 

No

Replace debugger and console check with a hook to not allow commit to go through if they exist in the code.

### DIFF
--- a/workbase/package.json
+++ b/workbase/package.json
@@ -161,7 +161,7 @@
   },
   "husky": {
       "hooks": {
-        "pre-commit": "lint-staged $$ grep -E -r 'debugger|console' src/"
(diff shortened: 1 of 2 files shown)


Irtazaraza:
@marco-scoppetta 
Do you think we should run all the tests before committing?

marco-scoppetta:
I am going to close this, we need to find a better solution that puts an hard stop on committing. Future!

Issue #4337, opened 2018-09-22
Variable role not working in some cases


grabl:
_This issue was originally posted by @jmsfltchr  on 2018-09-17 15:40_.

**Description:** Using a variable for a roleplayer works in one case but not in another. **Grakn Version:** 1.4 **Reproducible Steps:** Load genealogy The following is successful but slow (so I'm linking to the relevant ticket for that: "Supplying an id but query still slow"): 
```
 match $relationship id V94432; $relationship($role:$x); get; 
```
 (where the id is that of a `person` in the db, "Jacob J. Niesz" in this case) **Actual Results:** This query throws an exception: 
```
 match $x id V57392; $relationship($role:$x); get; INVALID_ARGUMENT: GraqlQueryException-the concept [role] could not be found in results. Please check server logs for the stack trace. 
```
 (where the id is that of a `parentship` relationship in the db) Stack trace is as follows: 
```
[...]

Issue #4024, opened 2018-09-22
Possibly inaccurate representation of resources


grabl:
_This issue was originally posted by @kasper-piskorski  on 2018-07-02 17:03_.

`ResourceAtom`: `predicateVariable` maps to both the resource concept as well as its type.

PR #4516, opened 2018-10-30
Fix inconsistencies in resource representation

kasper-piskorski:
# Why is this PR needed?
Addresses issues:
#4337 
#4024 
#4525 

# What does the PR do?
Makes corrections to the resource representation and handling of resource materialisation.
No.
None.


haikalpribadi:
Let's start calling "resource" as "attributes", @kasper-piskorski - it's been more than 2 years we changed it :)

PR #4538, opened 2018-11-02
Rules for deployment of Maven artifacts

vmax:
# Why is this PR needed?

So we are able to release artifacts built with `bazel`; resolved graknlabs/grakn#4500

# What does the PR do?

Introduces a custom rule `deploy_maven_jar` which allows pushing JARs to Maven Nexus repo

No, it shouldn't

— Use `bazel-common` from upstream: https://github.com/google/bazel-common/pull/36 
If they refuse to merge the PR, there's an option of copying implemenation locally
— ~How shall we specify credentials for deployment is a question to discuss~ previously not specifying credentials in command line failed the whole build but now we `warn` instead


dependencies/deployment/maven/deploy.sh
```diff
@@ -23,8 +23,8 @@ ARTIFACT="$ARTIFACT"
 MAVEN_PASSWORD="$MAVEN_PASSWORD"
 MAVEN_URL="$MAVEN_URL"
 MAVEN_USERNAME="$MAVEN_USERNAME"
-PARENT="$PARENT"
+COORDINATES="$COORDINATES"
```
lolski: The value is a single maven coordinate like `ai.grakn:core.server:1.5.0`, right?  If yes we should call it `$COORDINATE` (singular) rather than `$COORDINATES` (plural)

lolski: Actually, dismiss this comment.

We asked @jmsfltchr our english expert and we should use the plural form even when referring to just one since even for a single one it consists of a triplet (`ai.grakn`, `core.server`, `1.5.0`).

vmax:
@lolski I successfully ran `grakn-java-client-example` on `1.5.0-SNAPSHOT`. These are the differences between Maven-compiled JAR and `bazel`-compiled JAR (from your screenshot):

1. `*.proto` files are **not** needed in the archive -- generated proto classes are sufficient
1. `META-INF/maven/` directory - see [Maven docs](https://maven.apache.org/guides/mini/guide-archive-configuration.html). From what I understand, you don't need it to be able to use the JAR (because `pom.xml` is stored in repository itself)

dependencies/deployment/maven/deploy.sh
```diff
@@ -0,0 +1,89 @@
...
+    MD5_BINARY="md5sum"
+    SHA1_BINARY="sha1sum"
+else
+    echo "Error: there are no md5 and sha1 binaries for your platform: $platform"
+    exit 1
+fi
+
+# Update the JAR contents to simulate Maven structure
+DIRECTORY_INSIDE_JAR=META-INF/maven/$GROUP_ID/$ARTIFACT/
+cp lib.jar _lib.jar # un-symlink it
```
lolski:
```suggestion
cp lib.jar lib-with-maven-metadata.jar # un-symlink it and add the maven metadata to the new copy afterwards
```

Issue #4520, opened 2018-10-30
Build the target "bazel run //:deploy-github-zip" for "grakn-core-all"


lolski:
- `bazel run //:deploy-github-zip` should build `grakn-core-all-$VERSION.zip`

lolski:
Initially I thought we also need to be able to release `server` and `console` separately. In practice, this isn't true and we will always release both of them together as `grakn-core-all`.

PR #4547, opened 2018-11-04
Add a Bazel target `deploy-github-zip` for creating a draft release containing the zip distribution in Github 

lolski:
# Why is this PR needed?
Solves https://github.com/graknlabs/grakn/issues/4520

We need to be able to deploy to Github. It is now supported via this command in Bazel:

```
bazel run //:deploy-github-zip -- <github-username> <github-token>
```

When the command is invoked, it takes in the github-username github-token credentials needed for the deployment. Additionally the rule which is defined in `BUILD`  takes in `VERSION` and `deployment.properties`, the latter of which specifies a configuration `github.repository=<the-target-repository>`. Given those input:
1. Rename `grakn-core-all.zip` to `grakn-core-all-<version-from-the-version-file>.zip`
2. Create a draft release in Github containing`grakn-core-all-<version-from-the-version-file>.zip`

# What does the PR do?
The major things are:
1. Define `//:deploy-github-zip` in `BUILD`
2. Create `deployment.properties` which contains deployment-related configurations
3. Created `/dependencies/deployment/github/deployment.sh` which actually performs the Github deployment
4. Added `ghr` - the command line tool used to perform the Github deployment

Minor things:
1. Refactored `distribution.sh` on `//server` ,`//console` and `//`. All of them is ran using `set ex` to make it easier to troubleshoot when things go wrong. 

No

N/A


dependencies/deployment/github/deployment.sh
```diff
@@ -0,0 +1,36 @@
...
+fi
+
+set -ex # exit immediately when there's a failure. run the script in verbose mode (ie., print all the commands to stdout)
+
+# command-line arguments
+github_user="$1"
+github_token="$2"
+
+# configurations
+github_repository="cat deployment.properties | grep github.repository | cut -d '=' -f 2"
```
vmax: @lolski 
```suggestion
github_repository="grep github.repository deployment.properties | cut -d '=' -f 2"
```

dependencies/deployment/github/dependencies.bzl
```diff
@@ -0,0 +1,25 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+
+def dependencies_for_github_deployment():
+    native.http_file(
+        name = "ghr",
+        executable = False,
```
vmax: @lolski `executable = False` is already default, [src1](https://docs.bazel.build/versions/master/be/workspace.html#http_file_args), [src2](https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/repo/http.bzl#L283-L284)

dependencies/deployment/github/deployment.sh
```diff
@@ -0,0 +1,36 @@
...
+distribution_name="grakn-core-all"
+distribution_zipfile_path="dist/grakn-core-all.zip"
+distribution_version=`cat VERSION`
+distribution_basedir_name="$distribution_name-$distribution_version"
+distribution_zipfile_name="$distribution_basedir_name.zip"
+github_tag="v$distribution_version" # e.g, v1.5.0
+ghr_executable_file_name="ghr"
+ghr_basedir_name="ghr_v0.10.2_darwin_386"
```
vmax: @lolski you are limiting deployment to Darwin-only machines. While this might be fine, this is arbitrary imposed limit. It's better to also support Linux; we have a handy snippet of easily switching between platforms [here](https://github.com/graknlabs/grakn/blob/master/dependencies/tools/unused-deps.sh#L9-L16)

lolski: nice catch, I haven't fixed this one. Wait for a bit

lolski: We need to pull in different dependencies depending on whether you're on Mac or on Linux. How would we adjust the code, in your opinion?

```
def dependencies_for_github_deployment():
    native.http_file(
        name = "ghr",
        urls = ["https://github.com/tcnksm/ghr/releases/download/v0.10.2/ghr_v0.10.2_darwin_386.zip"]
    )
```

vmax: @lolski we should download them as `ghr_mac` and `ghr` and depend (via `sh_binary`'s `data`) on both of them

dependencies/deployment/github/deployment.sh
```diff
@@ -0,0 +1,70 @@
...
+#!/usr/bin/env bash
+
+set -ex # exit immediately when there's a failure. run the script in verbose mode (ie., print all the commands to stdout)
+
+if [[ $# -ne 2 ]]; then
+    echo "Error: needs two arguments, <github-username> <github-token>"
+    exit 1
+fi
+
+if ! [[ "$(uname)" == "Darwin" ]] && ! [[ "$(uname)" == "Linux" ]]; then
```
vmax:
```suggestion
if [[ "$(uname)" != "Darwin" ]] && [[ "$(uname)" != "Linux" ]]; then
```

Issue #4392, opened 2018-10-01
Relationship panel


Irtazaraza:
Create a new panel in the right bar to display which roles and relationships a node takes part in.

PR #4565, opened 2018-11-08
Relationships panel

Irtazaraza:
# Why is this PR needed?
closes: #4392
To display the roles played by a concept and the relationships it takes part in

# What does the PR do?
Implement a relationships panel component which computes the roles and relationships whenever a concept is clicked.

(Concept -> Roles -> Relationships -> Other role players in the relationships)

No.

1) Improve UI
2) Refactor computing of relationships to be more efficient


Irtazaraza:
> why did we stop using the `await app.client.waitUntil`? is it not applicable here?

This is another issue. the test fails because the position of the node on circle ci is different from my machine so it doesnt click on it to load the relationships panel. I dont think this test will work, ill be removing it.

Issue #4499, opened 2018-10-26
Migrate dependency management of client-nodejs and workbase to use Bazel


haikalpribadi:
At the moment, `client-nodejs` and `workbase` manages their dependency through NPM using `package.json` and `yarn.lock`. We need to migrate the dependency management of these packages to be fully managed using Bazel instead of `package.json` and `yarn.lock`.

Issue #4272, opened 2018-09-22
Integrate workbase to our build system


grabl:
_This issue was originally posted by @haikalpribadi  on 2018-09-01 17:04_.

PR #4568, opened 2018-11-09
Integrate Workbase with bazel workflow

vmax:
# Why is this PR needed?
closes: #4499
closes: #4272 

# What does the PR do?
Integrates Workbase with `bazel` build system

No, current workflow of Workbase developer won't be disturbed is any way

Replacing `npm` scripts (`webpack` and `electron-build`) with `bazel` rules though we'll need to figure out whether it's needed


workbase/wrappers-bazel/BUILD
```diff
@@ -0,0 +1,22 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
```
haikalpribadi: can we name the package to just be `//workbase/builder` ?

workbase/wrappers-bazel/npm_run_build.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+export PATH=`pwd`/external/nodejs/bin:$PATH
+export NODE_PATH=`pwd`/workbase/node_modules/
+
+pushd workbase
```
haikalpribadi: What happens when you run `bazel build` not from the root workspace directory? will this script still work?

And can we have the filenames use hyphens instead of underscores?

workbase/wrappers-bazel/npm_test.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+export PATH=`pwd`/external/nodejs/bin:$PATH
+export NODE_PATH=`pwd`/workbase/node_modules/
+
+pushd workbase
```
haikalpribadi: Same question as the above: What happens when you run `bazel test` not from the root workspace directory? will this script still work?

workbase/BUILD.bazel
```diff
@@ -0,0 +1,94 @@
...
+    srcs = glob([
+        "src/**/*",
+        "build/**/*",
+        "static/**/*",
+        ".electron-vue/**/*",
+        "package.json",
+        ".babelrc",
+        "config.js",
+        "jsconfig.json"]) + [":node_modules"],
+    outs = ["workbase.zip"],
```
haikalpribadi: what's the name of the current distribution zip? can we name it `grakn-core-workbase`? When we deploy, it should also include `VERSION`. Basically it should follow the same naming convention as our jars.

vmax: @haikalpribadi currently it is `grakn-workbase-{VERSION}-{platform}.zip`
We'll be able to rename it while storing artifact on CircleCI or when uploading with `ghr` but not in `bazel` -- `genrule`'s `outs` is `nonconfigurable` which basically means "has to have static name"

workbase/wrappers-bazel/npm_run_build.sh
vmax: @haikalpribadi 
>What happens when you run bazel build not from the root workspace directory? will this script still work?

Yes, because `bazel` replicates workspace structure in sandbox

>And can we have the filenames use hyphens instead of underscores?

 Yup

workbase/wrappers-bazel/npm_test.sh
vmax: @haikalpribadi similarly, yes, it will work

workbase/wrappers-bazel/BUILD
haikalpribadi: oh wait, how many files and folders start with the string `build`? it sounds like there are many now?

workbase/BUILD.bazel
haikalpribadi: Okay, I see. Let's name it `grakn-workbase.zip` at least then, @vmax

workbase/wrappers-bazel/BUILD
vmax: @haikalpribadi we can discuss it on a call. I didn't want to refactor it together with build integration to separate task scopes

.circleci/config.yml
```diff
@@ -115,44 +115,44 @@ jobs:
...
-        - run: cd workbase && npm install
+        - run: bazel run @nodejs//:npm install
         - run: Xvfb :99 &
         - run: export DISPLAY=:99
-        - run: cd workbase && npm run e2e
+        - run: bazel test -s --test_output=streamed //workbase:tests-e2e
 
   build-workbase-linux:
-      docker:
-        - image: circleci/node:8.11.2
```
Irtazaraza: I think the fail is attributed to the node version. I think we will need to revert it back to 8.11.2. What do you think @marco-scoppetta ?

Issue #4597, opened 2018-11-14
Refactor `grakn-core` to refer to `benchmark.lib`


flyingsilverfin:
Following the move of instrumentation code to the benchmarking repository, this ticket ensures that benchmarking files are incorporated via Bazel into `grakn-core`.

PR #4598, opened 2018-11-14
Extract Tracing Instrumentation out of `grakn.core` into `benchmark.lib`

flyingsilverfin:
# Why is this PR needed?
Moves most of the Benchmark Instrumentation required in `client-java` and `server` into the benchmarking repository, then imports required libs as distinct libraries into `grakn.core`

# What does the PR do?
* Move and refactor Client and Server tracing code away from `grakn.core`
* Clean up implementations of Tracing

finishes #4597


flyingsilverfin:
# Why is this PR needed?
Moves most of the Benchmark Instrumentation required in `client-java` and `server` into the benchmarking repository, then imports required libs as distinct libraries into `grakn.core`

# What does the PR do?
* Move and refactor Client and Server tracing code away from `grakn.core`
* Clean up implementations of Tracing

finishes #4597

PR #4611, opened 2018-11-15
Bring back Reasoning tests

kasper-piskorski:
# Why is this PR needed?

# What does the PR do?


test-integration/graql/reasoner/BUILD
```diff
@@ -1,6 +1,6 @@
 java_test(
     name = "reasoning-it",
-    size = "large",
+    size = "enormous",
```
marco-scoppetta: maybe we should split the test in the future

kasper-piskorski: will do, just trying to find a point where it passes :)

marco-scoppetta: 👍

PR #4617, opened 2018-11-17
Extracted Daemon (aka. Bootup) to its own module

haikalpribadi:
# Why is this PR needed?

# What does the PR do?


daemon/GraknDaemon.java
```diff
@@ -226,8 +226,8 @@ private void help() {
     }
 
     private void clean() {
-        boolean storage = storageBootup.isRunning();
-        boolean grakn = engineBootup.isRunning();
+        boolean storage = this.storageDaemon.isRunning();
+        boolean grakn = serverDaemon.isRunning();
```
lolski: `grakn` should be `server`

lolski: this is very minor and we can do it on the next PR

Issue #3441, opened 2018-09-21
Semantic (subsumptive) query cache


grabl:
_This issue was originally posted by @kasper-piskorski  on 2018-04-16 12:26_.

Make the query cache recognise relations between queries so that the query subsumption can be taken advantage of and subsequently we can reduce the number of redundant db lookups.

marco-scoppetta:
Done

PR #4633, opened 2018-11-21
Introduce semantic query cache

kasper-piskorski:
# Why is this PR needed?
Ticket: #3441 
# What does the PR do?
Introduces the following concepts:
- Query Subsumption:
```
Subsumption relation between a query C (child) and a provided query P (parent) holds if:

C <= P,

is true, meaning that P is more general than C (C specialises P) 
and their respective answer sets meet:
 
answers(C) subsetOf answers(P)
    
i. e. the set of answers of C is a subset of the set of answers of P.
```
- Semantic Difference 
```
Semantic difference between query C and P defines a specialisation operation
required to transform query P into a query equivalent to C.
```

Those concepts form the basis of the operation of the semantic cache: if we are looking for answers to query C and the cache already contains query P which has all DB entries (db-complete), we can propagate the answers, possibly specialising them (applying the semantic difference).
No.
Make it transaction-level.


haikalpribadi:
This PR now fails to merge because of this change @kasper-piskorski https://github.com/graknlabs/grakn/pull/4634

server/src/graql/internal/reasoner/ResolutionIterator.java
```diff
@@ -49,7 +47,7 @@
     private final ReasonerQueryImpl query;
     private final Set<ConceptMap> answers = new HashSet<>();
 
-    private final SimpleQueryCache<ReasonerAtomicQuery> cache = new SimpleQueryCache<>();
+    private final IndexedSemanticCache cache = new IndexedSemanticCache();
```
flyingsilverfin: Rename `Indexed` to `Multilevel`

server/src/graql/internal/reasoner/cache/Index.java
```diff
@@ -0,0 +1,70 @@
...
+package grakn.core.graql.internal.reasoner.cache;
+
+import com.google.common.collect.ImmutableSet;
+import grakn.core.graql.Var;
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * TODO
```
flyingsilverfin: what's this TODO supposed to be?

server/src/graql/internal/reasoner/cache/IndexedAnswerSet.java
```diff
@@ -0,0 +1,142 @@
...
+import grakn.core.graql.answer.ConceptMap;
+import grakn.core.graql.query.answer.ConceptMapImpl;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Set;
+import java.util.stream.Stream;
+
+/**
+ * TODO
```
flyingsilverfin: Same documentation missing

Issue #4456, opened 2018-10-10
Extract Graql query language library as a separate module


haikalpribadi:
There should be 2 parts of Graql: 
- `graql.lang`:  the _"language"_, the library that a USER uses to build a Graql string
- `grakn.core.graql`: Grakn's _"implementation"_ of the Graql language. The library that contains Graql algorithms & data structures (fragments, planner, analytics, reasoner) and transaction library that the SERVER uses to execute a given query and produce a result to the user. 

The two components above currently sit together, and they really should not.

> Note: the term _"implementation"_ might is not to be confused with Java _"interface"_ and _"implementation"_ classes. They're not the same. (Do suggest if you know a better one)

So given that we're currently trying to rewrite our session and transaction architecture (issue #3944), we also need to extract Graql language classes (what should be `graql.lang`) into a separate module (and repository), as these two issues go hand in hand. 

Before we can complete this issue, we need to remove any dependencies of `grakn.core.graql.*` classes from `graql.lang` classes. In other words: 

- `graql.lang` should not depend on anything. It's a language. 
- `grakn.core.graql` should depend on `graql.lang`. The language implementation should depend on the language library itself.

However, in the current Graql code base, such structure above does not exist. Every class just depends on every other class. Thus, to create such architecture, it requires many of the Graql (language and implementation) classes to be refactored and rewritten.

After we split the classes to fall into the two categories defined above, we then split out the code just for Graql language into `graql` repository (`graknlabs/graql`), which may look like:
- `language`: artifact id `graql:lang`(contains things like`graql.g4`)
- `java`: artifact id `graql:java` (Graql Java parser and query builder)
- `javascript`: artifact id `graql:javascript` (Graql JavaScript parser and query builder)
- `python`: artifact id `graql:python` (Graql Python parser and query builder)
- `plugins`: artifact id `graql:plugins`
    - `intellij`
    - `atom`
    - `visual-studio`
    - ...

This issue needs #4432

PR #4635, opened 2018-11-21
Refactor Graql Query classes

haikalpribadi:
# Why is this PR needed?

As part of the process of splitting Graql Query classes from to not depend on Transactions (issue #4456), we need to first simplify the Graql Query classes. For this instance where the Query classes will simply be a "query string builder" for Graql, and will unlikely have a second implementation, the Query interfaces do not add value and we can remove them to simplify the codebase.

# What does the PR do?

- merged `AggregateQuery` interface and implementation (`AggregateQueryImpl`)
- _(on going)_


server/src/graql/answer/ConceptMap.java
```diff
@@ -18,129 +18,311 @@
...
      * NB:assumes answers are compatible (concept corresponding to join vars if any are the same)
      *
-     * @param a2 answer to be merged with
+     * @param a2          answer to be merged with
+     * @param explanation flag for providing explanation
      * @return merged answer
      */
-    @CheckReturnValue
-    ConceptMap merge(ConceptMap a2);
+    @javax.annotation.CheckReturnValue
```
lolski: we're refering to the annotation directly rather than importing like this:

```
import javax.annotation.CheckReturnValue;
...

@CheckReturnValue
```

PR #4667, opened 2018-11-28
Refactor Graql query language library - Part 2

haikalpribadi:
# Why is this PR needed?

# What does the PR do?

This PR is an intermediate step in simplifying the architecture of Graql Pattern Property classes (`grakn.core.graql.query.pattern.property.*`). We first simplify the unnecessary modularisation to make it easier to see the behaviour of the code.

- Make `IsAbstractProperty`, `RelatesProperty` and `RelationshipProperty` implement `NamedProperty`, so then we can:
- Merge `VarPropertyInternal` and `NamedProperty` interfaces
- Removed the usage `UniqueVarProperty`, in which `VarProperties` are down-casted to, and:
- Made sure all classes implemented `isUnique()` method
- Finally, we merged `VarProperty` (inteface), `VarPropertyInternal` (interface) and `AbstractVarProperty` (class), into just one abstract class: `VarProperty`

The result of the above change is depicted in the images below.

> Note: green lines indicate interface implementations, blue lines indicate class extensions

**Before**

before

**After**

after

This PR does not change any Graql behaviours.


lolski:
LGTM

PR #4703, opened 2018-12-12
Split //common:common into multiple targets, deploy as one

### LINKED ISSUE #4579: The deploy_maven_jar rule should know how to create a single JAR from a package along with their subpackages
Often we have a Bazel package that is to big for it to exist by itself. Therefore it makes sense to split it into multiple sub-packages.

However we still want to deploy a single JAR consisting of all of them aggregated into one file, rather than having 1 JAR file for each of the sub-package.

vmax:
# Why is this PR needed?

Fixes #4579
Needs graknlabs/deployment#5 graknlabs/deployment#6 (update link to `@graknlabs_rules_deployment` when merged)

# What does the PR do?

* Breaks `//common` into mutliple subpackages:
```
//common/config/test:grakn-config-test
//common:deploy-maven-jar
//common:common
//common/templates/util:grakn-version
//common/http:http
//common/config:config
//common/util:util
//common/exception:exception
```
* Allows to merge subtargets of `//common` to be deployed as one JAR to Maven

Nope

Breaking other Maven targets into subpackages (they may need to be reorganized to avoid circular dependencies)


common/BUILD
```diff
@@ -41,28 +35,17 @@ java_library(
...
     name = "deploy-maven-jar",
-    targets = [":common"],
+    target = ":common",
+    package = "common",
+    deps = [
+        "//common/exception",
+        "//common/config",
+        "//common/http",
+        "//common/util"
+    ],
```
haikalpribadi: Let's move the `deps` to be declared in the originating `target`, and have `deploy_maven_jar` to traverse the dependency tree and build the jar. One idea is to use an `aspect`, but we can look into other ways too.

common/BUILD
```diff
@@ -41,28 +35,17 @@ java_library(
     tags = ["maven_coordinates=grakn.core:common:{pom_version}"],
```
haikalpribadi: Let's move `runtime_deps` to the actual package that needs it. In this case: `//common/config`

common/BUILD
```diff
@@ -17,22 +17,16 @@
...
-            "**/test/*.java",
-            "templates/**",
-        ]
-    ),
-    deps = [
-        "//dependencies/maven/artifacts/com/google/auto/value:auto-value",
-        "//dependencies/maven/artifacts/com/google/code/findbugs:annotations",
-        "//dependencies/maven/artifacts/com/google/guava:guava",
-        "//dependencies/maven/artifacts/org/slf4j:slf4j-api",
+    srcs = ["//common/templates/util:grakn-version"],
```
haikalpribadi: This source `//common/templates/util:grakn-version` should go into sources of `//common/util`

common/BUILD
```diff
@@ -41,28 +35,17 @@ java_library(
...
+    package = "common",
+    deps = [
+        "//common/exception",
+        "//common/config",
+        "//common/http",
+        "//common/util"
+    ],
     version_file = "//:VERSION",
     deployment_properties = "//:deployment.properties",
 )
```
haikalpribadi: Let's move `//:VERSION` and `//:deployment.properties` into `deploy_maven_jar()` as default values. You may override it if needed, but in general you should always use the one provided by the repository`

dependencies/maven/BUILD
```diff
@@ -14,4 +14,11 @@
...
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
-#
\ No newline at end of file
+#
+
+genrule(
+    name = "deployment_rules",
+    srcs = ["@graknlabs_rules_deployment//maven/templates:rules.bzl", "//:deployment.properties"],
+    cmd = "MAVEN_PACKAGES=`grep 'maven.packages' $(location //:deployment.properties) | cut -d '=' -f 2`; sed -e s/{maven_packages}/$$MAVEN_PACKAGES/g $(location @graknlabs_rules_deployment//maven/templates:rules.bzl) >> $@",
+    outs = ["rules.bzl"]
```
haikalpribadi: This will be the place where we provide the `deployment.properties` file as input to generate `rules.bzl`.

common/BUILD
haikalpribadi: In the end, we won't even need `//:VERSION` file as input, once we move `version` as a property in `//:deployment.properties`

dependencies/maven/BUILD
haikalpribadi: Temporarily we will still need to provide `//:VERSION` file as input too, until we move `version` as a property into `deployment.properties`

dependencies/maven/rules.bzl
```diff
@@ -0,0 +1,261 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
```
haikalpribadi: We should make a comment at the top of every auto-generated file, to let the use know that this should not be edited. For example, see: https://github.com/graknlabs/grakn/blob/29b7c8a8e79b8e85b41cbf060b84d95adf1643d2/dependencies/maven/dependencies.bzl#L1

common/BUILD
vmax: @haikalpribadi solved it by having `jarjar_library` which takes a number of `java_library` targets and produces one JAR out of them

common/BUILD
vmax: @haikalpribadi fixed, see https://github.com/vmax/grakn/blob/split-common-to-multiple-pkgs/common/util/BUILD#L4

common/config/BUILD
```diff
@@ -0,0 +1,17 @@
+java_library(
+    name = "config",
+    srcs = ["Config.java", "ConfigKey.java", "SystemProperty.java"],
```
haikalpribadi: Is there a reason why we don't use a `glob([**/*.java])` here, @vmax ?

dependencies/maven/deployment_rules_builder.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+
+import sys
+import re
+
+
+def parse_deployment_properties(fn):
+    deployment_properties = {}
```
haikalpribadi: Why is this file `deployment_rules_builder.py` not placed in `graknlabs/deployment`, @vmax ? Isn't this something generic that we want all users of `graknlabs/deployment` to be able to use?

dependencies/maven/rules.bzl
```diff
@@ -0,0 +1,278 @@
...
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
```
haikalpribadi: Is this license header provided from the template file? What if the user of `graknlabs/deployment` would want to use a different license? For example, we want to use `graknlabs/deployment` rules in `graknlabs/graql` soon, and that repo will be licensed under Apache. Can we make it configurable by the user? This doesn't need to be in this PR.

common/config/BUILD
vmax: @haikalpribadi apart from explicit file lists being a bit faster than recursive `glob`, there is none

dependencies/maven/deployment_rules_builder.py
vmax: @haikalpribadi Makes sense 🤔 The `genrule` part invoking it will still stay in `graknlabs/grakn`, right?

dependencies/maven/rules.bzl
vmax: @haikalpribadi 

>Is this license header provided from the template file?

Yes, it is in the template

>What if the user of `graknlabs/deployment` would want to use a different license?

License comment states that user of `graknlabs/deployment` must adhere to a particular license (GNU AGPL v3). If you _don't_ include license header, strictly speaking, the user cannot use your rules [legally]. License header is template does *not* enforce you to have the _same_ license for your project — correct me if I'm wrong here

>This doesn't need to be in this PR.

I will remove it -- but licensing is still an issue

dependencies/maven/deployment_rules_builder.py
haikalpribadi: yes, the `genrule()` stays in the user's repository (in this case `graknlabs/grakn`

dependencies/maven/rules.bzl
haikalpribadi: What I mean by _"This doesn't need to be in this PR"_ is that any changes to make licensing generic does not need to be in this PR. But the license itself can/should still go in this PR.

haikalpribadi: Well, if a user of `graknlabs/deployment` wants to use it outside of Grakn Labs, it should Apache. I.e. the `graknlabs/deployment` itself will have an Apache License in the root of the repository, and the template files should not contain a license header. Can we make it such that the license header is (optionally) provided by the user through the `genrule` upon generating the file from template?

PR #4724, opened 2018-12-14
Version and license options for server

### LINKED ISSUE #4698: Have certain console commands as grakn options
At the moment, the following requirements are implemented as part of the console:
- `>>> license`: **command** to print out the license
- `--version`: **option** to print out Grakn's version

Both of these requirements have nothing to do with the purpose of the console which is to be an interface for interacting with a keyspace.

It'd make more sense to have them as options of `grakn`:
- `./grakn --version`
- `./grakn --license`

vmax:
# Why is this PR needed?

Fixes #4698

# What does the PR do?

* Removes version (`-v`) as option of `console`
* Adds `--version` and `--license` as options to `server`

No

Testing this on Windows (`grakn.bat`) — I wrote the script 'blindfolded'

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -16,7 +16,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-exports_files(["grakn", "VERSION", "deployment.properties"], visibility = ["//visibility:public"])
+exports_files(["grakn", "VERSION", "deployment.properties", "LICENSE"], visibility = ["//visibility:public"])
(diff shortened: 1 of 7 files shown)


vmax:
@haikalpribadi as we removed `--version ` from `console`, test `when_startingConsoleWithOptionVersion_expect_printVersion` now fails.

What should I do?

* remove the test completely
* rewrite the test so it tests that there's *no* option called (`--version`)
* make a test for `server` which verifies the same thing

PR #4728, opened 2018-12-14
Refactor schema design actions

Irtazaraza:
# Why is this PR needed?
Code reusability

# What does the PR do?
Reuse code to load schema when defining new schema concepts
build and remove edges for adding anding removing attribute types

no

N/A

### DIFF
--- a/workbase/src/renderer/components/SchemaDesign/store/actions.js
+++ b/workbase/src/renderer/components/SchemaDesign/store/actions.js
@@ -19,7 +19,6 @@ import logger from '@/../Logger';
 import Grakn from 'grakn';
 import SchemaHandler from '../SchemaHandler';
 import {
-  META_CONCEPTS,
   computeSubConcepts,
   relationshipTypesOutboundEdges,
   updateNodePositions,
@@ -29,6 +28,19 @@ import {
   ownerHasEdges,
 } from '../SchemaUtils';
 import SchemaCanvasEventsHandler from '../SchemaCanvasEventsHandler';
+
+async function buildSchema(nodes) {
+  // Find nodes that are subconcepts of existing types - these nodes will only have isa edges
+  const subConcepts = await computeSubConcepts(nodes);
+
+  // Draw all edges from relationships to roleplayers
+  const relEdges = await relationshipTypesOutboundEdges(nodes);
+
+  // Draw all edges from owners to attributes
+  const hasEdges = await ownerHasEdges(nodes);
+
+  return { nodes, edges: relEdges.concat(subConcepts.edges, hasEdges) };
+}
 
 export default {
   async [OPEN_GRAKN_TX]({ state, commit }) {
@@ -87,17 +99,12 @@ export default {
         .filter(x => x.label !== 'entity')
         .filter(x => x.label !== 'attribute')
         .filter(x => x.label !== 'relationship');
-      // Find nodes that are subconcepts of existing types - these nodes will only have isa edges
-      const subConcepts = await computeSubConcepts(nodes);
 
-      // Draw all edges from relationships to roleplayers
-      const relEdges = await relationshipTypesOutboundEdges(nodes);
+      const data = await buildSchema(nodes);
 
-      // Draw all edges from owners to attributes
-      const hasEdges = await ownerHasEdges(nodes);
+      data.nodes = updateNodePositions(data.nodes);
 
-      nodes = updateNodePositions(nodes);
-      state.visFacade.addToCanvas({ nodes, edges: relEdges.concat(subConcepts.edges, hasEdges) });
+      state.visFacade.addToCanvas({ nodes: data.nodes, edges: data.edges });
       state.visFacade.fitGraphToWindow();
 
       nodes = await computeAttributes(nodes);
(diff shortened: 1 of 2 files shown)


workbase/src/renderer/components/SchemaDesign/store/actions.js
```diff
@@ -189,6 +185,22 @@ export default {
...
+
+        const data = await buildSchema([type]);
+
+        data.edges.push(...((await typeInboundEdges(type, state.visFacade)).filter(x => x.to === type.id)));
+
+        state.visFacade.addToCanvas({ nodes: data.nodes, edges: data.edges });
+
+        data.nodes = await computeAttributes(data.nodes);
+        state.visFacade.updateNode(data.nodes);
+        graknTx.close();
```
marco-scoppetta: what's going on here? can you make it a method or add comments? so that in the future we remember

workbase/src/renderer/components/SchemaDesign/store/actions.js
```diff
@@ -189,6 +185,22 @@ export default {
     await dispatch(COMMIT_TX, graknTx)
       .then(async () => {
         await dispatch(UPDATE_METATYPE_INSTANCES);
+
```
marco-scoppetta: also we are using `await` everywhere, so why we have the following code inside a `then`? maybe we can just keep the code flat with comments/methods

workbase/src/renderer/components/SchemaDesign/store/actions.js
```diff
@@ -211,10 +223,17 @@ export default {
...
         const node = state.visFacade.getNode(state.selectedNodes[0].id);
 
+        const edges = [];
         await Promise.all(payload.attributeTypes.map(async (attributeType) => {
-          const dataType = await (await graknTx.getSchemaConcept(attributeType)).dataType();
+          const type = await graknTx.getSchemaConcept(attributeType);
+          const dataType = await type.dataType();
           node.attributes = [...node.attributes, { type: attributeType, dataType }];
+
+          edges.push({ from: state.selectedNodes[0].id, to: type.id, label: 'has' });
```
marco-scoppetta: this is ok, but for future reference you can also do:

```
const edges = await Promise.all(
    .
    .
   return { from: state.selectedNodes[0].id, to: type.id,  label: 'has'}
)
```

I am adding this comment because it's a refactoring PR, so tips and trick for JS here and there 😄

marco-scoppetta: this is because `.map()` returns an array already

workbase/src/renderer/components/SchemaDesign/store/actions.js
```diff
@@ -226,19 +245,32 @@ export default {
...
+
+      const attributeTypeId = (await graknTx.getSchemaConcept(payload.attributeLabel)).id;
+      const edgesIds = state.visFacade.edgesConnectedToNode(state.selectedNodes[0].id);
+      edgesIds.forEach((edgeId) => {
+        const edge = state.visFacade.getEdge(edgeId);
+        if (edge.to === attributeTypeId) {
+          state.visFacade.deleteEdge(edgeId);
+        }
+      });
+      graknTx.close();
```
marco-scoppetta: also here, not important but for FYI:
```
edgesIds.
.filter(edgeId => state.visFacade.getEdge(edgeId).to === attributeTypeId)
.forEach((edgeId) => { state.visFacade.deleteEdge(edgeId); });
```

workbase/src/renderer/components/SchemaDesign/store/actions.js
```diff
@@ -276,26 +308,19 @@ export default {
...
+      Object.assign(type, { label: payload.relationshipLabel });
+
+      const data = await buildSchema([type]);
+
+      data.edges.push(...((await typeInboundEdges(type, state.visFacade)).filter(x => x.to === type.id)));
+
+      state.visFacade.addToCanvas({ nodes: data.nodes, edges: data.edges });
+
+      data.nodes = await computeAttributes(data.nodes);
+      state.visFacade.updateNode(data.nodes);
```
marco-scoppetta: comments/methods if possibruuuu

workbase/src/renderer/components/SchemaDesign/store/actions.js
```diff
@@ -253,30 +251,29 @@ export default {
...
-      const edgesIds = state.visFacade.edgesConnectedToNode(state.selectedNodes[0].id);
-      edgesIds.forEach((edgeId) => {
-        const edge = state.visFacade.getEdge(edgeId);
-        if (edge.to === attributeTypeId) {
-          state.visFacade.deleteEdge(edgeId);
-        }
-      });
-      graknTx.close();
-    })
+    await dispatch(COMMIT_TX, graknTx)
```
marco-scoppetta: I think we should remove the:

```
if (await (await type.instances()).next()) throw Error('Cannot remove attribute type from schema concept with instances.');
```

It is not Workbase duty to check the correctness of the graph, the server should return an exception if we try to delete types that have instances, the workbase duty is to correctly handle those exceptions and show them in a fancy way

PR #4738, opened 2018-12-17
Schema Designer - Define new rule

Irtazaraza:
# Why is this PR needed?
Allow user to define a new rule

# What does the PR do?
UI and implementation to define a new rule

No

Replace with interactive rule builder

### DIFF
--- a/workbase/src/renderer/components/SchemaDesign/store/actions.js
+++ b/workbase/src/renderer/components/SchemaDesign/store/actions.js
@@ -12,6 +12,7 @@ import {
   DELETE_SCHEMA_CONCEPT,
   REFRESH_SELECTED_NODE,
   DELETE_ATTRIBUTE,
+  DEFINE_RULE,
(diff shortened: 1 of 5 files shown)


workbase/src/renderer/components/SchemaDesign/store/actions.js
```diff
@@ -388,6 +389,20 @@ export default {
...
+
+    // define rule
+    await state.schemaHandler.defineRule(payload);
+
+    await dispatch(COMMIT_TX, graknTx)
+      .catch((e) => {
+        graknTx.close();
+        logger.error(e.stack);
+        throw e;
+      });
```
marco-scoppetta: I see that you have this block of code (that commit tx and handles exceptions if any) in many places, if it becomes too verbose maybe you can just make a `async commitTx()` helper method that you can call and it wil contain the exception handling and everything

PR #4741, opened 2018-12-18
Schema display settings

Irtazaraza:
# Why is this PR needed?
Large Schemas gets too complicated when showing all three types. Allow user to toggle visibility of three meta types

# What does the PR do?
Add settings tab in right bar and display settings panel
UI/UX/Implementation to toggle visibility of certain meta-types

no

N/A


workbase/src/renderer/components/SchemaDesign/RightBar/SettingsTab/DisplaySettingsPanel.vue
```diff
@@ -0,0 +1,100 @@
...
+        this.showTypes(val, 'ATTRIBUTE_TYPE');
+      },
+      showRelationships(val) {
+        this.showTypes(val, 'RELATIONSHIP_TYPE');
+      },
+    },
+    methods: {
+      showTypes(showType, baseType) {
+        const schemaConcepts = this.$store.state['schema-design'].visFacade.getAllNodes().filter((x => x.baseType === baseType)).map(x => Object.assign(x, { hidden: !showType }));
+        this.$store.state['schema-design'].visFacade.updateNode(schemaConcepts);
```
marco-scoppetta: to we always access the schema design state using: `this.$store.state['schema-design']`?

Irtazaraza: ah no. We can use a getter. Change has been done.

PR #4743, opened 2018-12-18
Move distribution from genrules to new custom rule

### LINKED ISSUE #4494: Reimplement distribution scripts in Python
Building the Grakn distributions requires Bazel to call the distribution scripts: 
- `./console/distribution.sh`
- `./server/distribution.sh`
- `./distribution.sh`

They are bash scripts and therefore does not run on Windows. Meanwhile, we need `bazel build //:distribution` to run on all major platforms: Linux, Mac, and Windows. 

Thus, let's re-implement `distribution.sh` scripts in Python, and make sure that `bazel build //:distribution` runs on Linux, Mac and Windows.

The distribution script should not use the `_deploy.jar` _"fat JAR"_ as it causes two issues:
1. Our license becomes cobbled together with the licenses of our dependencies.
2. As the JAR becomes a monolithic JAR, the user will have less control when dealing with transitive dependency issues

We also need to simplify the way we copy files and directories, which should come naturally in Python.

vmax:
# Why is this PR needed?

Fixes #4494 

# What does the PR do?

Uses new `deployment` rule to build distribution ZIPs:
```
//server:distribution
//console:distribution
//:distribution
```

No

—


haikalpribadi:
Do you still need me to review this, or has @lolski reviewed it with you, @vmax ?

vmax:
@haikalpribadi we agreed to merge it with @lolski when I fix e2e tests

haikalpribadi:
Then you can add @lolski to the reviewer

PR #4744, opened 2018-12-18
Refactor Graql query language library - Part 7

haikalpribadi:
# Why is this PR needed?

# What does the PR do?

This PR focuses on splitting Graql patterns into Language VS Execution (in pursuit of #4456).
1. Removed class references in JavaDoc comments, as this produces _"dirty imports"_ into the class (imports of classes that are not actually used, and thus misleading).
2. Mapping of `VarProperty` to `Atomic` reasoner queries are moved to  `AtomicFactory`. This may or may not be the best place to place them, but it's a good enough place, for now, to remove them from `VarProperty` classes. We can move them again to a more appropriate place once we get to tidying up the architecture for reasoner, @kasper-piskorski.
3. Define, Undefine and Insert execution for Graql `VarProperty` classes are now extracted out of each `VarProperty classes, and reimplemented through new `PropertyExecutor` classes, sitting under `grakn.core.graql.internal.executor.property`. 
    - `PropertyExecutor` can now be either:
        * `Definable`: must implement `defineExecutors()` and `undefineExecutors()`
        * `Insertable`: must implement `insertExecutors()`
    - `PropertyExecutor`s will have several inner classes that implements of `PropertyExecutor.Writer`, which will contain the logic if `Define`, `Undefine` or `Insert` write operation to the database.


server/src/graql/internal/reasoner/atom/AtomicFactory.java
```diff
@@ -54,4 +98,237 @@
...
+    /**
+     * maps a var property to a reasoner atom
+     *
+     * @param varProperty   the var property to be converted to a reasoner atom
+     * @param var           Statement this property belongs to
+     * @param vars          Statements constituting the pattern this property belongs to
+     * @param parent        reasoner query this atom should belong to
+     * @return created atom
+     */
+    private static Atomic mapToAtom(VarProperty varProperty, Statement var, Set<Statement> vars, ReasonerQuery parent) {
```
kasper-piskorski: good Lord

haikalpribadi: Haha, I'm sure this will can be refactored into a better shape, @kasper-piskorski. But let's do that once we get to reasoner architecture. Right now we'll have to [carefully] push the mess out of Graql query and pattern classes.

kasper-piskorski: It's how it's been in the prototype (0.3?). Definitely not a good place!

kasper-piskorski: `AtomicFactory`. That way we break the Open/Closed principle. If we add a new property we need to modify an existing class. Possibly in multiple places.

haikalpribadi: So if a wants to add a new Atom, the developer changes AtomFactory which is used to build Atoms - what's weird in that? 

Don't just look at one specific software engineering principle when designing an architecture. Look at the overall design and cost. As it stands right now: Graql (language) depends on reasoner, Reasoner depends on Transaction, and Transaction depends on Graql (language). This is just one among many circular dependencies that exist in the current code base, @kasper-piskorski.

Start with the higher level architecture that makes sense, then work downwards on deciding the implementation. Everything else is secondary.

PR #4749, opened 2018-12-19
Rule applicability bug fixes

kasper-piskorski:
# Why is this PR needed?
- Applicable rules copying is potentially error prone. If we change the atom context, the applicable rules may change.
- possible for cache to miss an answer if following conditions are met:
    * a query is ground,
    * an answer to the query is inferred, 
    * an entry doesn't exist in the cache and answers are propagated from parent
# What does the PR do?
- Removes the copying of applicable rules.
- fixes caching issue
No.
No.


server/src/graql/internal/reasoner/atom/AtomicFactory.java
```diff
@@ -0,0 +1,59 @@
...
+ * Factory class for creating {@link Atomic} objects.
+ */
+public class AtomicFactory {
+
+    /**
+     * @param pattern conjunction of patterns to be converted to atoms
+     * @param parent query the created atoms should belong to
+     * @return set of atoms
+     */
+    public static Stream<Atomic> createAtoms(Conjunction<Statement> pattern, ReasonerQuery parent) {
```
haikalpribadi: Why did you extract this method to its own class again @kasper-piskorski ? I moved it into `ReasonerQueryImpl` because it's one method made only used by that class, and it doesn't need its own class. I'm trying to minimise the number of classes that don't carry significant weight/purpose.

kasper-piskorski: Adding more things later.

haikalpribadi: Okay, you should mention that (e.g. in the description) and tag relevant person (in this case me), otherwise it looks _rather impolite_ to just undo the change that someone just made in the previous PR. I'll approve this PR once my other PR is in.

PR #4775, opened 2018-12-29
Debian deployment

### LINKED ISSUE #3093: Repackage distributions for Grakn Core and KGMS (enable both installed together)
_This issue was originally posted by @haikalpribadi  on 2017-06-13 12:45_.

We are in the process of redesigning our architecture after we redefine our package architecture, we want to refine our distributions.

1. Differentiate product names (Grakn Core VS Grakn KGMS)
2. Differentiate package services (Server VS Console)
3. Unify entry point through the shell script (just have one console per product)

For the binary/jar distribution, we want to have the following:

### For Grakn Core:

1. `grakn-core-bin` : `opt/grakn/core/grakn`: the shell script for running server or console
2. `grakn-core-server` : `opt/grakn/core/server/grakn-core-server.jar`
    - `opt/grakn/core/server/services/...`
3. `grakn-core-console`: `opt/grakn/core/console/grakn-core-console.jar`
    - `opt/grakn/core/console/services/...`

`grakn` will symlink to `opt/grakn/core/grakn`. Users can then install grakn through package managers in such a way (example using APT, but should apply for every package manager):

1. `apt-get install grakn-core-all` which installs 1, 2, and 3 from above.
2. `apt-get install grakn-core-server` which installs 1 and 2 from above.
3. `apt-get install grakn-core-console` which installs 1 and 3 from above.

### For Grakn KGMS

1. `grakn-kgms-bin` : `opt/grakn/kgms/grakn`: the shell script for running server or console on Grakn KGMS.
2. `grakn-kgms-server` : `opt/grakn/kgms/server/grakn-core-server.jar`
    - `opt/grakn/core/server/services/...`
3. `grakn-kgms-console`: `opt/grakn/kgms/console/grakn-core-console.jar`
    - `opt/grakn/core/console/services/...`

`grakn` will symlink to `opt/grakn/kgms/grakn`. Users can then install grakn through package managers in such a way (example using APT, but should apply for every package manager):

1. `apt-get install grakn-kgms-all` which installs 1, 2, and 3 from above.
2. `apt-get install grakn-kgms-server` which installs 1 and 2 from above.
3. `apt-get install grakn-kgms-console` which installs 1 and 3 from above.

Note:

1. There is no package manager command to install both grakn-core and grakn-kgms distributions all at once, because it is not needed.
2. grakn-kgms distributions can only be installed through a package manager, when the user is provided access to our internal APT server, or other package managers.

This issue needs #4260, needs #3419

vmax:
# Why is this PR needed?

Fixes #3093
Needs graknlabs/deployment#12

# What does the PR do?

Uses rules from `@graknlabs_rules_deployment` to build Debian packages

Most probably it does -- I'll need to modify existing ZIP distributions to adapt to new db location and remove unused files
PR is to just to review the initial structure

Backwards-compat fixes
Deployment process (*uploading* the built package to the repo)


grakn-debian
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
```
lolski: do we really need `grakn-debian`? I wonder if we can just use the existing`grakn`.

server/BUILD
```diff
@@ -131,3 +132,27 @@ distribution(
     output_filename = "grakn-core-server",
     visibility = ["//visibility:public"]
 )
+
+deploy_deb(
+    name = "grakn-core-server",
+    target = ":server-binary",
+    package_dir = "/opt/grakn/core/server/",
```
lolski: Would it make sense to call it `installation_dir`?

vmax: @lolski it mirrors the name of [the attribute](https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/pkg/pkg.bzl#L215) of `pkg_tar` (which we use in implementation)

vmax: @lolski fixed in https://github.com/graknlabs/grakn/pull/4775/commits/d5853651e4308a331e4df8b2074f1df76025a8be

grakn-debian
vmax: @lolski fixed in https://github.com/graknlabs/grakn/pull/4775/commits/eb01ae4f3977747ac1c36cf1f1071530f342e7ea

vmax:
Don't review, will be replaced by a new PR

PR #4778, opened 2019-01-02
New Graql Grammar (in progress)

haikalpribadi:
# Why is this PR needed?

We are in the process of refactoring Graql (starting from PR #4646), with the goal to split Graql from Grakn (issue #4456). However, this journey involved refactoring most of Graql's implementation and data structures, which are not as efficient and robust as it should be. The current constraint in getting Graql's data structure model right is the Graql (ANTLR) Grammar (written in `Graql.g4). The Graql Grammar is too loosely defined, which allows room for error in parsed Graql queries (left to be caught during runtime) and leads to the wrong data structure implementation.

# What does the PR do?

This PR focuses on rewriting the Graql Grammar (`Graql.g4`) to be more higher level in its grammar definition but at the same time more strict in parsing the queries. As you can see in `Graql.g4`, changes are made across the whole grammar, but here are some of the key points (which may not be exhaustive) are:
- A `pattern` is now defined as a `statement`, `conjunction` (of patterns), and `disjunction` (of patterns) 
- We split a `statement` into `statement_type` or `statement_instance`
- Define Queries now only accepts `statement_type`
- Insert Query now only accepts `statement_instance`
- Match Clause (in which Delete, Get, Aggregate and Group queries depend on) can take in any statement
- A `statement_type` is a statement to describe types in the schema, defined by a set of `type_property`s you can provide for a given type
- A `statement_instance` is a statement to describe data instances, defined as `instance_thing`, `instance_relation`, and `instance_attribute` - higher level constructs that describe Grakn Concepts more naturally
- Each type of `statement_*` can be followed by a list of `attributes`, separated by a comma which is no longer optional
- A variable is always required when writing a `statement_thing`
- We introduce `$_` as _"anonymous variables"_ for when users do not want/need to declare a variable name to refer to

Additionally:
- We introduce the keyword `like` to query attributes by regular expression
- We merge `REGEX` lexer (token) into `STRING`

_(IN PROGRESS)_


haikalpribadi:
Closing due to conflict with https://github.com/graknlabs/grakn/pull/4781. Will redo the work taking into account new Graql negation grammar and patterns.

PR #4780, opened 2019-01-03
Move RBE config to .bazelrc

lolski:
# Why is this PR needed?
Bazel RBE config is in CircleCI's `.circleci/config.yml`. It needs to be moved to `.bazelrc`.

# What does the PR do?
1. Moves RBE configurations from `.circleci/config.yml` to `.bazelrc`
2. Bazel can now be invoked like this in `.circleci/config.yml`:
```
client-java:
  ...
  steps:
    - bazel:
        command: bazel test //client-java/... --test_size_filters=small --test_output=errors
  ...
```

No

N/A


.circleci/config.yml
```diff
@@ -23,49 +23,22 @@ commands:
...
-                --auth_enabled=true \
-                --spawn_strategy=remote \
-                --strategy=Javac=remote \
-                --strategy=Closure=remote \
-                --genrule_strategy=remote \
-                --define=EXECUTOR=remote \
-                --action_env=BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1 \
-                --experimental_strict_action_env=true
-                "
+              RBE_FLAGS="--config=rbe"
```
haikalpribadi: We should inline this `RBE_FLAGS` variable since it's so short now. I.e. just do 
```
if credentials exist
    CMD="<< parameters.command >> --config=rbe"
else
    CMD="<< parameters.command >>"
fi
```

lolski: done: https://github.com/graknlabs/grakn/pull/4780/files#diff-1d37e48f9ceff6d8030570cd36286a61

PR #4792, opened 2019-01-07
Update readme for clients

sorsaffari:
# Why is this PR needed?
To update the readme for clients Java, Node.js and Python to keep them consistent with the documentation.

# What does the PR do?
Updates the `README.MD` of each client, to provide:
- installation instructions (inc. dependencies)
- a quickstart example
- references to the examples queries in the docs
- reference to the API Reference in the docs
- reference to the Concept API in the docs

No

The minimal redundant content that exists both in the `README.MD` files and in the corresponding page in the documentation, need to stay in sync.

### DIFF
--- a/client-nodejs/README.md
+++ b/client-nodejs/README.md
@@ -81,7 +81,6 @@ async function runBasicQueries() {
   }
 
   // Or query and consume the iterator immediately collecting all the results
-  // - consume it all immediately
   answerIterator = await readTransaction.query("match $x isa person; limit 10; get;");
   const persons = await answerIterator.collectConcepts();
   persons.forEach( person => { console.log("Retrieved person with id "+ person.id) });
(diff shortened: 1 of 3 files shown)


client_python/README.md
```diff
@@ -1,433 +1,116 @@
----
-title: Grakn Python Client
-keywords: setup, getting started, download, driver
-tags: [getting-started]
-sidebar: documentation_sidebar
-permalink: /docs/language-drivers/client-python
-folder: docs
-symlink: true
----
+# Grakn Client Python
```
flyingsilverfin: Not sure about @marco-scoppetta  but I think it's more readable as `Grakn Python Client`

client-nodejs/README.md
```diff
@@ -1,377 +1,116 @@
...
+
+  while ( aConceptMapAnswer != null) {
+    // get the next `x`
+    const person = aConceptMapAnswer.map().get("x");
+    console.log("Retrieved person with id "+ person.id);
+    aConceptMapAnswer = await answerIterator.next();
+  }
+
+  // Or query and consume the iterator immediately collecting all the results
+  // - consume it all immediately
```
marco-scoppetta: we can remove this comment

client_python/README.md
marco-scoppetta: Yeah it reads slightly better with the language name first. Grakn Java Client - Grakn Python Client - Grakn Node.js Client

PR #4793, opened 2019-01-07
Support multiple targets in checkstyle_test; fix #4767

### LINKED ISSUE #4767: checkstyle_test should take multiple targets and glob("*.java"), so that we only need a single checkstyle_test rule covering multiple rules per package
For example, consider the following snippet where we have a single `checkstyle_test` covering `java_library` and `java_test`:
```
java_library(
    name = "rocks-db-queue-library",
   ...
)

java_test(
    name = "rocks-db-queue-test",
   ...
)
checkstyle_test(
 name = "checkstyle",
 targets = [":rocks-db-queue-test", ":rocks-db-queue-lib"] // or alternatively glob("*.java")
 config = "//config/checkstyle:checkstyle.xml",
 suppressions = "//config/checkstyle:checkstyle-suppressions.xml",
 licenses = ["//config/checkstyle:licenses"],

)
```

vmax:
# Why is this PR needed?

Fixes #4767

# What does the PR do?

* Adds `targets` attribute to `checkstyle_test` while still supporting *both* new `targets` and old `target`
* Adds support for `targets`-covered Java targets to `check-for-missing-checkstyle-rule.py`
* Makes messages in `check-for-missing-checkstyle-rule.py` more user-friendly

No

* Introducing naming convention for `checkstyle_test` targets with multiple packages (`checkstyle`?) (`package-name-checkstyle`?)
* Merging several `-checkstyle` targets in one package into one
* Removing support for `target` (attribute for checking single target)

### DIFF
--- a/check-for-missing-checkstyle-rule.py
+++ b/check-for-missing-checkstyle-rule.py
@@ -19,28 +19,44 @@
 #
 
 from __future__ import print_function
+from itertools import chain
+import os
 from pprint import pprint
 from xml.etree import ElementTree
 import subprocess
 import sys
 
 
-java_targets = set(subprocess.check_output([
+def check_output_discarding_stderr(*args, **kwargs):
+    with open(os.devnull, 'w') as devnull:
+        return subprocess.check_output(*args, stderr=devnull, **kwargs)
+
+
+print('Checking if there are any source files not covered by checkstyle...')
+
+java_targets = set(check_output_discarding_stderr([
     'bazel', 'query',
     '(kind(java_library, //...) union kind(java_test, //...)) '
     'except //dependencies/... except attr("tags", "checkstyle_ignore", //...)'
 ]).split())
 
-checkstyle_targets_xml = subprocess.check_output([
+checkstyle_targets_xml = check_output_discarding_stderr([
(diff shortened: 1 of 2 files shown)


check-for-missing-checkstyle-rule.py
```diff
@@ -19,28 +19,44 @@
...
 ]).split())
 
-checkstyle_targets_xml = subprocess.check_output([
+checkstyle_targets_xml = check_output_discarding_stderr([
     'bazel', 'query', 'kind(checkstyle_test, //...)', '--output', 'xml'
 ])
 checkstyle_targets_tree = ElementTree.fromstring(checkstyle_targets_xml)
 checkstyle_targets = set(map(lambda x: x.get('value'),
-                             checkstyle_targets_tree.findall(".//label[@name='target'][@value]")))
+                             chain(
```
lolski: Make the line less terse:
```
a = checkstyle_targets_tree.findall(".//label[@name='target'][@value]")
b = checkstyle_targets_tree.findall(".//list[@name='targets']//label[@value]")
checkstyle_targets = list(map(lambda x: x.get('value'), a + b))
```

PR #4795, opened 2019-01-09
RPMs should be built locally, without RBE

vmax:
# Why is this PR needed?

Needs graknlabs/deployment#23
Fixes RPM failing on RBE

# What does the PR do?

Uses new rules which include `no-remote` tag

—

—


vmax:
Will be replace by another one

PR #4802, opened 2019-01-12
Refactor Graql query language library - Part 11

haikalpribadi:
# Why is this PR needed?

## Refining Graql's Grammar

We are in the process of refactoring Graql (starting from PR #4646), with the goal to split Graql from Grakn (issue #4456). However, this journey involves refactoring of Graql's data structures, which are not as consistent and robust as it should be. The current constraint in correcting Graql's data structures is the Graql Grammar (written in `Graql.g4` ANTLR4). The Graql Grammar is too loosely defined, which allows room for invalid queries to be parsed, and left to the responsibility of the compiler to catch during runtime. It also leads to the wrong data structures being designed, and this PR is just an initial attempt in fixing them while allowing future PRs to continue the work without having to change the grammar too much.

# What does the PR do?

This PR focuses on extracting higher-level patterns to be "lifted" in the grammar and be made apparent, and at the same time more strict in parsing the queries. As you can see in `Graql.g4`, changes are made across the whole grammar, but here are some of the key points (which may not be exhaustive):
- A `pattern` is now defined as a `pattern_statement`, `pattern_conjunction`, `pattern_disjunction`, or `pattern_negation`
- We split a `statement` into `statement_type` or `statement_instance`
- Define Queries now only accepts `statement_type`
- Insert Query now only accepts `statement_instance`
- Match Clause (in which Delete, Get, Aggregate and Group queries depend on) can take in any statement
- A `statement_type` is a statement to describe types in the schema, defined by a set of `type_property`s you can provide for a given type
- A `statement_instance` is a statement to describe data instances, defined as `statement_thing`, `statement_relation`, and `statement_attribute` - higher level constructs that describe Grakn Concepts more naturally
- Each type of `statement_intance` can be followed by a list of `attributes`, separated by a comma which is no longer optional
- A variable is always required when writing a `statement_thing`
- We introduce `$_` as _"anonymous variables"_ for when users do not want/need to declare a variable name to refer to

Additionally:
- We introduce the keyword `like` to query attributes by a regular expression in Match Queries
- We merge `REGEX` lexer (token) into `STRING`

_This PR also changes how Graql Statement classes are implemented. However, they are still an intermediary state which will evolve again in the next PR._


server/src/graql/query/pattern/statement/StatementInstance.java
```diff
@@ -0,0 +1,371 @@
...
+        String message = "Not allowed to provide Statement Property: [" + varProperty.toString() + "] ";
+        message += "to " + statement.getClass().getSimpleName() + ": [" + statement.toString() + "]";
+        throw new IllegalArgumentException(message);
+    }
+
+    private StatementInstance(Variable var, LinkedHashSet<VarProperty> properties, Sign sign) {
+        super(var, properties, sign);
+    }
+
+    @SuppressWarnings("Duplicates")
```
kasper-piskorski: why supress?

server/src/graql/query/pattern/statement/StatementInstance.java
```diff
@@ -0,0 +1,371 @@
...
+            } else if (getProperty(NeqProperty.class).isPresent()) {
+                return getProperty(NeqProperty.class).get().toString();
+
+            } else {
+                return "";
+            }
+        }
+
+        @Override @SuppressWarnings("Duplicates")
+        public String toString() {
```
kasper-piskorski: we have three copies of this tostring

server/src/graql/query/pattern/statement/StatementInstance.java
haikalpribadi: this was left over from when I was copying it from `Statement.java`.

haikalpribadi: you can remove it

server/src/graql/query/pattern/statement/StatementInstance.java
haikalpribadi: they're not the same, I may or may not modularise them when I change Statement classes to be strict typing.

Issue #4234, opened 2018-09-22
Graql: Pattern Negation


grabl:
Ability to negate patterns in order to exclude some results from conjunctive queries

PR #4812, opened 2019-01-14
Pattern negation pt 1.5 - unified negated pattern behaviour

kasper-piskorski:
# Why is this PR needed?
Addresses #4234

Extends the set difference semantics to all patterns, i.e. we interpret executing queries with negated patterns in terms of performing set operations with negation corresponding to a set difference.

For a conjunctive query with negated patterns:

```
Q = A ∧ {...} ∧ ¬B ∧ ¬C ∧ {...}
```
Each answer x to query Q has to belong to the set:

``` 
{x : x ∈ A ∧ x !∈ B ∧ x !∈ C ∧ {...}}
``` 
or equivalently:

``` 
{x : x ∈ A x ∈ B^C ∧ x ∈ C^C ∧ {...}}
``` 
where upper C letter marks set complement.
 
As a result the answer set ans(Q) is equal to:
``` 
 ans(Q) = ans(A) \ [ ( ans(A) ∩ ans(B) ) ∪ ( ans(A) ∩ ans(C) ) ]
```
or equivalently
``` 
 ans(Q) = ans(A) ∩ ans(B^C) ∩ ans(C^C)
```

This aligns the semantics of a negated pattern with defining a negative literal with an atom corresponding to the body of the negation.
# What does the PR do?
Introduces `CompositeQuery` and `CompositeState` extending the ideas of `ReasonerQueryImpl` and `NegatedConjunctiveState` to all patterns.
No.
Add support for negated disjunctions.


graql/grammar/Graql.g4
```diff
@@ -111,20 +111,17 @@ pattern             :   pattern_statement
                     |   pattern_disjunction
                     |   pattern_negation
                     ;
+
 pattern_conjunction :   '{' patterns '}' ';' ;
 pattern_disjunction :   '{' patterns '}'  ( OR '{' patterns '}' )+  ';' ;
-
-pattern_negation    :   NOT '{' patterns '}' ';' ;
+pattern_negation    :   NOT '{' pattern '}' ';' ;
```
haikalpribadi: As discussed, `pattern_negation` should be `NOT { patterns }`, taking in one or more patterns. If there is more than one pattern provided, it will be a negation of a conjunction as it should.

server/src/graql/parser/Parser.java
```diff
@@ -409,7 +411,7 @@ public Pattern visitPattern(GraqlParser.PatternContext ctx) {
     }
 
     @Override
-    public Disjunction<?> visitPattern_disjunction(GraqlParser.Pattern_disjunctionContext ctx) {
+    public Pattern visitPattern_disjunction(GraqlParser.Pattern_disjunctionContext ctx) {
```
haikalpribadi: It's good practice to return the strict type whenever known. So this method should be returning `Disjunction<?>`.

server/src/graql/parser/Parser.java
```diff
@@ -421,20 +423,13 @@ public Pattern visitPattern(GraqlParser.PatternContext ctx) {
     }
 
     @Override
-    public Conjunction<?> visitPattern_conjunction(GraqlParser.Pattern_conjunctionContext ctx) {
+    public Pattern visitPattern_conjunction(GraqlParser.Pattern_conjunctionContext ctx) {
```
haikalpribadi: It's good practice to return the strict type whenever known. So this method should be returning `Conjunction<?>`.

server/src/graql/query/pattern/Pattern.java
```diff
@@ -65,19 +65,18 @@
...
+    @CheckReturnValue
+    Disjunction<Conjunction<Pattern>> getNegationDNF();
+
     /**
-     * @return this pattern negated
+     * @return true if this pattern corresponds to a negation pattern
      */
     @CheckReturnValue
-    Pattern negate();
+    default boolean isNegation(){ return false; }
```
haikalpribadi: We don't need this method as it does exactly the same thing as `instanceof Negation`.

server/src/graql/query/pattern/Pattern.java
```diff
@@ -65,19 +65,18 @@
...
-     * @return true if this pattern contains only positive patterns
+     * @return this Pattern as a Negation, if it is one.
      */
     @CheckReturnValue
-    default boolean isPositive(){
-        return getDisjunctiveNormalForm().getPatterns().stream()
-                .flatMap(p -> p.getPatterns().stream())
-                .allMatch(Pattern::isPositive);
-    }
+    default Negation<?> asNegation(){ throw new UnsupportedOperationException(); }
```
haikalpribadi: We also don't need to "casting helper method". It hides the possibility of an exception being thrown. I would recommend doing an explicit act of casting when you need to. We've also started removing all methods like this in Grakn.

PR #4832, opened 2019-01-21
Simplify checkstyle integration (fix #4825)

### LINKED ISSUE #4825: Simplify CheckStyle rule definition
Right now, this is how we define CheckStyle rules in our `BUILD` files:
```
load("//dependencies/tools/checkstyle:checkstyle.bzl", "checkstyle_test")
checkstyle_test(
    name = "exception-checkstyle",
    target = ":exception",
    config = "//config/checkstyle:checkstyle.xml",
    suppressions = "//config/checkstyle:checkstyle-suppressions.xml",
    licenses = ["//config/checkstyle:licenses"],
)
```

1) We should only declare one CheckStyle target per `BUILD` file. Once we can do this, the target name can just be `"checkstyle"`; in the example above it would be without `exception-`.
2) You will notice that `config`, `suppressions`, and `licenses` will repeat for every `BUILD` file. Is it possible to define this once per `//package`?

vmax:
# Why is this PR needed?

So checkstyle targets are easier to read; fixes #4825

# What does the PR do?

1. Utilizes new attribute `targets`, merging all `checkstyle` targets to one per `BUILD` file
2. Adapt `check-for-missing-checkstyle-rule.py ` to be Python 3 compatible
3. Moves `config`, `suppressions`, `license_files` (former `licenses`, renamed due to bazelbuild/bazel#7194) to `default` of attributes

—

—

### DIFF
--- a/check-for-missing-checkstyle-rule.py
+++ b/check-for-missing-checkstyle-rule.py
@@ -30,13 +30,23 @@ def check_output_discarding_stderr(*args, **kwargs):
         return subprocess.check_output(*args, stderr=devnull, **kwargs)
 
 
+def try_decode(s):
+    if type(s) == bytes:
+        return s.decode()
+    return s
+
+
+def lmap(func, seq):
+    return list(map(func, seq))
+
+
 print('Checking if there are any source files not covered by checkstyle...')
 
-java_targets = check_output_discarding_stderr([
+java_targets = lmap(try_decode, check_output_discarding_stderr([
     'bazel', 'query',
     '(kind(java_library, //...) union kind(java_test, //...)) '
     'except //dependencies/... except attr("tags", "checkstyle_ignore", //...)'
-]).split()
+]).split())
 
 checkstyle_targets_xml = check_output_discarding_stderr([
     'bazel', 'query', 'kind(checkstyle_test, //...)', '--output', 'xml'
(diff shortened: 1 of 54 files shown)


check-for-missing-checkstyle-rule.py
```diff
@@ -30,22 +30,32 @@ def check_output_discarding_stderr(*args, **kwargs):
         return subprocess.check_output(*args, stderr=devnull, **kwargs)
 
 
+def try_decode(s):
+    if type(s) == bytes:
+        return s.decode()
+    return s
+
+
+def lmap(func, seq):
```
lolski: Is there a reason why we need to abstract `list(map(func, seq))` into `lmap`?

check-for-missing-checkstyle-rule.py
```diff
@@ -30,22 +30,32 @@ def check_output_discarding_stderr(*args, **kwargs):
         return subprocess.check_output(*args, stderr=devnull, **kwargs)
 
 
+def try_decode(s):
```
lolski: `try_decode` does not seem to abstract the steps in a way that is meaningful - and rather just something which exists as a way to minimise repetition. What do you think?

check-for-missing-checkstyle-rule.py
vmax: @lolski in this case, it's just shorter to read

check-for-missing-checkstyle-rule.py
lolski: Is `check_output_discarding_stderr` the only place which uses `subprocess.check_output`? If yes, I would actually make sure it always returns string and we can remove `try_decode`:
```
def check_output_discarding_stderr(*args, **kwargs):
    with open(os.devnull, 'w') as devnull:
        out = subprocess.check_output(*args, stderr=devnull, **kwargs)

        # subprocess.check_output return 'str' on Python 2 but 'bytes' on Python 3.
        #Therefore, do an optional conversion to string if necessary
        out = type(out) == bytes ? s.decode() : s

        return out
```

PR #4844, opened 2019-01-23
Remove client-nodejs and explain where it's now (fix #4821)

### LINKED ISSUE #4821: Split //client-nodejs into a separate repo: graknlabs/client-nodejs
We need to split client drivers from `grakn-core` so that they can have their own independent release cycles. The dependency to `//protocol` needs to be done using `git_repository` dependency to `grakn-core`.

vmax:
# Why is this PR needed?

Fixes #4821

# What does the PR do?

Removes `client-nodejs` as it has been moved

—

—


vmax:
Another one will be submitted after proper migration

PR #4855, opened 2019-01-29
Refactor Graql query language library - part 15

haikalpribadi:
# Why is this PR needed?

# What does the PR do?

Graql Predicates (based on how we implemented it) are operations we do on an attribute. For example:
- `$x 18 isa age`: which includes an assignment predicate of `val 18` to the variable `$x`
- `$x >= 20`: which includes GTE (greater-than-or-equals) comparison `>= 18` on the  variable `$x`
- `$x has name contains "Bob"`: which includes a String check `contains "Bob"` on the invisible variable which will become the `name` of `$x`

This PR focuses on rewriting Graql Predicates, such that:
1. We rename `Predicates` to something else that does not clash with reasoner's terminology. We've picked `Operation`, which is rather generic but satisfactory in the current case given that the class is nested scoped as a nested class: `ValueProperty.Operation`
2. We differentiate two different usages of predicates: `Operation.Assignment` and `Operation.Comparison`
3. The language and execution data structure are separated:
    - `ValueProperty.Operation` represents the language data structure
    - `ValueExecutor.Operation` represents the execution data structure
4. The types of literal values that can be provided into a predicate/operation are strictly checked at compile time, by enumerating all possible attribute operations in the `AttributeProperties` interface in which `Statements` will implement. This seems rather verbose, but highly valuable as it makes the language more reliable and intuitive to use.
5. The types of literal values that can be stored in the data structure are also strictly guaranteed by modelling it into the `Generic` types of the data structure, as seen in `ValueProperty.Operation` and `ValueExecutor`


server/src/graql/query/Query.java
```diff
@@ -95,7 +95,30 @@ public String toString() {
...
+            for (Operator c : values()) {
+                if (c.operator.equals(value)) {
+                    return c;
+                }
+            }
+            return null;
+        }
+    }
+
+    enum Comparator {
```
kasper-piskorski: why don't we make it a class? we would avoid all the procedural code and multiple ifs in ValueExecutor

server/src/graql/internal/reasoner/atom/predicate/IdPredicate.java
```diff
@@ -18,65 +18,104 @@
...
- *
- * <p>
  * Predicate implementation specialising it to be an id predicate. Corresponds to {@link IdProperty}.
- * </p>
- *
- *
  */
-@AutoValue
-public abstract class IdPredicate extends Predicate<ConceptId>{
+public class IdPredicate extends Predicate<ConceptId> {
```
kasper-piskorski: Still think it's a perfect autovalue use case here. Instead we are removing one line and adding forty for no gain.

server/src/graql/internal/reasoner/atom/predicate/IdPredicate.java
```diff
@@ -18,65 +18,104 @@
...
+
+    private final Variable varName;
```
kasper-piskorski: Without autovalue, these should go into the parent Predicate class.

server/src/graql/internal/reasoner/atom/predicate/NeqPredicate.java
```diff
@@ -18,39 +18,73 @@
...
  * Predicate implementation specialising it to be an inequality predicate. Corresponds to graql {@link NeqProperty}.
- * </p>
- *
- *
  */
-@AutoValue
-public abstract class NeqPredicate extends Predicate<Variable> {
+public class NeqPredicate extends Predicate<Variable> {
+
+    private final Variable varName;
```
kasper-piskorski: ditto

server/src/graql/internal/reasoner/atom/predicate/ValuePredicate.java
```diff
@@ -18,60 +18,94 @@
...
- *
- *
  */
-@AutoValue
-public abstract class ValuePredicate extends Predicate<grakn.core.graql.query.predicate.ValuePredicate> {
+public class ValuePredicate extends Predicate<ValueProperty.Operation<?>> {
 
-    @Override public abstract Statement getPattern();
-    @Override public abstract ReasonerQuery getParentQuery();
+    private final Variable varName;
```
kasper-piskorski: ditto

server/src/graql/query/Query.java
haikalpribadi: The actual comparison class is in `(ValueProperty|ValueExecutor).Operation.Comparison`. These comparators you're referring to is just the syntax token.

server/src/graql/internal/reasoner/atom/predicate/IdPredicate.java
haikalpribadi: Yes, that makes sense!

PR #4866, opened 2019-02-05
CI pipeline - automated release workflow

lolski:
# Why is this PR needed?
In order to minimise the manual steps needed in order to perform a release

# What does the PR do?
Introduces two workflows: `ci` and `ci-release`.

The `ci` workflow consists of jobs which:
   - executes tests (executed on pull requests and `master`)
   - syncs docs/development to the latest grakn version (executed on `master` only)
   - waits for a manual approval from an administrator for Grakn to be released (executed on `master` only)

The `ci-release` workflow is triggered by the creation of `trigger-ci-release` branch in `graknlabs/grakn`. It consists of jobs which:
  - creates a Github draft release containing the Grakn distribution
  - _in the future, prepares docs to be released. but it's not included yet in this PR_
  - cleans up the `trigger-ci-release` branch which was created by the approve-release job

**Confidence in the code**:
Either simplify the code to like 5 lines so they don't need tests, or otherwise consider adding tests as the code isn't that simple:

- Implement `approve-release-e2e.py`: it should start Grabl locally, and perform the test against it
- Implement `deploy-docs-e2e.py`: it should test deploying docs including the failure cases
- Implement `docs-auto-merge-to-master-e2e.py`: it should test both happy path and non-happy path (ie., merging)


.circleci/config.yml
```diff
@@ -175,66 +172,89 @@ jobs:
...
       - test-integration-reasoner:
           filters:
             branches:
-              ignore: master
+              ignore: /master|trigger-ci-release/
       - test-integration-analytics:
           filters:
             branches:
-              ignore: master
+              ignore: /master|trigger-ci-release/
```
marco-scoppetta: Check if possible to use references to trigger all tests without the need of repeating every command:
```
ci-pull-request:
    jobs:
      - reference_here
          filters:
      	      branches:
                 ignore: /master|trigger-ci-release/
```

.circleci/config.yml
```diff
@@ -279,6 +299,9 @@ workflows:
             branches:
               only: master
       - docs-update-grakn-dependency-on-dev-branch:
+          filters:
```
marco-scoppetta: check if possible to trigger this job before merging rather than after

.circleci/approve-release.py
```diff
@@ -0,0 +1,53 @@
...
+                output = output.decode()
+            return output
+        except subprocess.CalledProcessError as e:
+            print('An error occurred when running "' + str(e.cmd) + '". Process exited with code ' + str(
+                e.returncode) + ' and message "' + e.output + '"')
+            raise e
+
+workflow_id = os.getenv('CIRCLE_WORKFLOW_ID')
+
+GRABL_DATA = {
```
marco-scoppetta: not constant

.circleci/approve-release.py
```diff
@@ -0,0 +1,53 @@
...
+GRABL_URL_STATUS = '{GRABL_HOST}/release/{commit}/status'.format(GRABL_HOST=GRABL_HOST, commit=workflow_id)
+
+print('Tests have been ran and everything is in a good, releasable state. It is possible to proceed with the release process')
+_ = check_output_discarding_stderr([
+    'curl', '-X', 'POST', '--data', json.dumps(GRABL_DATA), '-H', 'Content-Type: application/json', GRABL_URL_NEW
+])
+
+status = 'no-status'
+
+while status == 'no-status':
```
marco-scoppetta: do we sleep?

.circleci/deploy-docs.py
```diff
@@ -0,0 +1,40 @@
...
+        sp.check_output(["git", "config", "--global", "user.email", git_email])
+        sp.check_output(["git", "config", "--global", "user.name", git_username])
+
+        # --recursive clones web-dev as well as the docs submodule
+        print('Cloning ' + web_dev_url + ' (' + web_dev_master_branch + ' branch)')
+        sp.check_output(["git", "clone", "--recursive", "https://"+grabl_credential+"@"+web_dev_url, web_dev_clone_location], stderr=sp.STDOUT) # redirect stderr to silence the output from git
+        
+        print('Updating submodule at "' + docs_submodule_location + '"')
+        sp.check_output(["git", "pull", "origin", web_dev_master_branch], cwd=docs_submodule_location) # TODO: replace origin
+        sp.check_output(["git", "add", "."], cwd=web_dev_clone_location)
```
marco-scoppetta: git add -A

.circleci/deploy-docs.py
```diff
@@ -0,0 +1,40 @@
...
+        
+        should_commit = sp.check_output(["git", "status"], cwd=web_dev_clone_location).find('nothing to commit, working tree clean') == -1
```
marco-scoppetta: maybe there are other ways to check if git status is empty

.circleci/config.yml
lolski: Fixed.

Using reference wasn't possible because the workflow in `ci-pull-request` and `ci-master-branch` isn't exactly the same - they're mostly the same but needs to ignore different branches in the filter.

The solution is to just have one workflow `ci` which is executed on `master` and pull requests where the approve-release and sync docs jobs are only executed in `master`

.circleci/config.yml
lolski: will have a discussion with Haikal and probably create a new PR to address it.

.circleci/approve-release.py
lolski: fixed

.circleci/approve-release.py
lolski: we didn't. it's fixed now.

.circleci/deploy-docs.py
lolski: after giving some thought, it's better to use `git add .` since this job works on a repository which has submodules. using `git add .` will clearly show the intent that we only want to include this directory. what do you think?

.circleci/deploy-docs.py
lolski: fixed, now we use `git diff` which can return different status code.

.circleci/deploy-docs.py
marco-scoppetta: 👍

PR #4869, opened 2019-02-08
Unify processing of neq predicates

### LINKED ISSUE #4528: Incorrect behaviour when reasoning with !==
Example:

Dataset: resourceAttachment.gql

with
```
derivable-resource-string sub attribute datatype string;
reattachable-resource-string sub derivable-resource-string datatype string;
derived-resource-string sub derivable-resource-string datatype string;
```

Example queries:
```
match
$x has derived-resource-string $val;
$unwanted 'unattached';
$val !== $unwanted;
get;
```

```
match
$x has derived-resource-string $val;
$y has reattachable-resource-string $anotherVal;
$val !== $anotherVal;
get;
```
```
match
$x has derivable-resource-string $value;
$x has derivable-resource-string $unwantedValue;
$unwantedValue 'unattached';
$value !== $unwantedValue;
$value isa $type;
$unwantedValue isa $type;
$type != $unwantedType;
$unwantedType label 'derivable-resource-string';
get;
```

ValuePredicate with !== should exhibit set comlement behaviour similarly to NeqPredicate. 
Instead, once negation is fully implemented it can generalise both.

kasper-piskorski:
# Why is this PR needed?
Fixes #4528 
# What does the PR do?
Treats `!==`, i. e. inequality of attribute values, in the same manner as `!=` - answers are obtained by calculating relevant set complements and checking that condition is satisfied.
No.
No.


haikalpribadi:
In the PR description, instead of saying `Addresses #4528`, you should say `Fixes #4528`, @kasper-piskorski. That way GitHub will automatically `close` the issue when this PR is merged. I've changed the description for this one.

server/src/graql/query/pattern/property/ValueProperty.java
```diff
@@ -119,9 +116,12 @@ public T value() {
         }
 
         public boolean isValueEquality() {
-            return comparator.equals(Query.Comparator.EQV) && !(this instanceof Comparison.Variable);
+            return comparator.equals(Query.Comparator.EQV) && innerStatement() == null;
```
haikalpribadi: Although this is "programmatically" correct, it hides the explicit meaning of the method which makes it one step harder to understand. For example:

When a reader reads this method, and ask: When is a ValueExecutor `isValueEquality`?
The answer, in this case, would be: when it holds a Value Equality comparator, and the `innerStatements()` is NULL.

However, the reader will still be left with a second/nested question: Why does having NULL `innerStatements()` make a ValueExecutor isValueEquality? The answer, in this case, would be: because when the `innerStatements()` is NULL, it means that this ValueExecutor is an instance of Variable Comparison. However, having NULL is an implementation detail.

Not only that the understanding of the method requires a dependency (thus adding "complexity"), it's also not the canonical meaning, and also not guaranteed to be true all the time. What if a different subclass of ValueExecutor implements innerStatements() to not be NULL, and still be `isValueEquality`? How would that affect the `isValueEquality` implementation? It would have to change too.

The change, in this case, is obviously very minor, but I believe this is a coding principle we should apply to improve the understandability of our code base. That is: to define the behaviour of a class/method as canonical as possible, which simplifies the understanding, and have it not be affected by the change of other non-related classes/methods.

As I start to understand more of our low-level implementations, I will highlight other places where this occurs. For now, I would recommend changing this method to the previous implementation, unless it would programmatically wrong/error.

server/src/graql/internal/reasoner/atom/predicate/NeqIdPredicate.java
```diff
@@ -0,0 +1,78 @@
...
+    public Atomic copy(ReasonerQuery parent) { return create(this.getPattern(), parent);}
+
+    @Override
+    public String toString(){
+        IdPredicate idPredicate = this.getIdPredicate(this.getVarName());
+        IdPredicate refIdPredicate = this.getIdPredicate(this.getPredicate());
+        return "[" + getVarName() + "!=" + getPredicate() + "]" +
+                (idPredicate != null? idPredicate : "" ) +
+                (refIdPredicate != null? refIdPredicate : "");
+    }
```
haikalpribadi: Can we have syntax-creation like this be implemented using the `Query` syntax enums, Kasper? These things will change in the future and we want to make sure we manage them consistently. (In fact, I'm about to make a change to the inequality operators syntax in the next PR).

server/src/graql/internal/reasoner/atom/predicate/NeqValuePredicate.java
```diff
@@ -0,0 +1,92 @@
...
+    @Nullable public Object getValue(){ return value;}
+
+    @Override
+    public Atomic copy(ReasonerQuery parent) { return create(this.getVarName(), this.getPredicate(), this.getValue(), parent);}
+
+    @Override
+    public String toString(){
+        return "[" + getVarName() + " !== " + getPredicate() + "]"
+                + (getValue() != null? "[" + getPredicate() + "/" + getValue() + "]" : "");
+    }
```
haikalpribadi: Ditto.

server/src/graql/internal/reasoner/atom/predicate/NeqIdPredicate.java
kasper-piskorski: It's for output and printing when debugging only

haikalpribadi: Still, though. It's not just "good practice" to always manage string values in constant variables, but also, in this case, is required so we can ensure syntax updates are synchronised across the codebase, even for debug statements. Otherwise, the debug statements would become misleading. @kasper-piskorski

haikalpribadi: the new syntax enums are at `graql.util.Token` bw (which will moved to `graql.lang.util.Token` in the next PR)

Issue #4874, opened 2019-02-08
GraknTestServer should not hard-code configuration path in order to be generic


lolski:
`GraknTestServer` expects:
- `grakn.properties` to be located exactly at `"server/conf/grakn.properties"`
- `embedded-cassandra.yaml` to be located exactly at `"test-integration/resources/cassandra-embedded.yaml"`

This isn't ideal and the parameters should be configurable by the class which instantiates it. For example, if `client-java` have tests which needs `GraknTestServer`, the hardcoded paths will get in the way.

PR #4875, opened 2019-02-08
make GraknTestServer takes server config and cassandra config in the constructor

lolski:
# Why is this PR needed?
Fixes https://github.com/graknlabs/grakn/issues/4874

# What does the PR do?

Adds a new constructor which accepts paths to config files:
```
public GraknTestServer(String serverConfigPath, String cassandraConfigPath) {
 ...
}
```

No

https://github.com/graknlabs/grakn/issues/4876

### DIFF
--- a/test-integration/rule/GraknTestServer.java
+++ b/test-integration/rule/GraknTestServer.java
@@ -59,8 +59,8 @@ import java.util.UUID;
  * It allows all the integration tests to run concurrently on the same machine.
  */
 public class GraknTestServer extends ExternalResource {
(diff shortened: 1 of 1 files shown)


test-integration/rule/GraknTestServer.java
```diff
@@ -75,7 +75,13 @@
     private SessionStore sessionStore;
 
     public GraknTestServer() {
+        this("server/conf/grakn.properties", "test-integration/resources/cassandra-embedded.yaml");
```
haikalpribadi: this is minor, but we should put this in a constant variable and refer to it, @lolski. makes it more obvious that it's default variable.

PR #4908, opened 2019-02-12
Revise query plan string output

kasper-piskorski:
# Why is this PR needed?
Current output has readability problems and obfuscates variables.
# What does the PR do?
Revises the output so it can be followed easily.
No.
No.


haikalpribadi:
and don't forget to `label` and `milestone` the PRs too @kasper-piskorski :)

PR #4918, opened 2019-02-14
do not stratify rules when they are positive

kasper-piskorski:
# Why is this PR needed?
No rule stratification is needed when we are dealing with positive rules.
# What does the PR do?
Shortcircuits startification if no negative rules are present.
No.
NO.


haikalpribadi:
👍

Issue #4818, opened 2019-01-17
Split //client-java into a separate repo: graknlabs/client-java


haikalpribadi:
We need to split client drivers from `grakn-core` so that they can have their own independent release cycles. The dependency to `//protocol` needs to be done using `git_repository` dependency to `grakn-core`. 

However, we can only work on this after complete issue #4456 and #4513.

Issue #4513, opened 2018-10-30
Remove transitive/database dependencies from client-java


haikalpribadi:
Right now, `client-java` pulls all the database dependencies, such as JanusGraph, Cassandra, and so on. This is because of `grakn-graql`, which `client-java` depends on, contains language + database implementation. This issue will be automatically resolved when we complete issue #4456.

Issue #4511, opened 2018-10-30
split client-java distribution for benchmarking vs public release


flyingsilverfin:
* Don't want to pull in Brave and Zipkin into `client-java`'s public release distribution (want a small, light client-java distribution)
* alternative 2nd distribution that creates client-java for benchmarking
* Will need to create abstraction allowing client-java public release distirbution to compile without depending on benchmarking & Brave/Zipkin

flyingsilverfin:
Plan: Create an extended `benchmark.Grakn` that enables instrumentation and build a separate a Bazel distribution jar/maven deploy which exclues `benchmark.*` for normal usage and includes it for the client-java-benchmark version.

haikalpribadi:
@flyingsilverfin I just realised you don't even need to extend `GraknClient` to add tracing for benchmarking. Given that `GraknClient` has a `.overrideChannel(ManagedChannel)` method (which was previously provided as an overloaded constructor `GraknClient(ManagedChannel)`, you can just override the `ManagedChannel` from your benchmark application, with one that you constructed to have tracing with Zipkin. The code would look like something like this:

```
GraknClient client = new GraknClient(address);
        SimpleURI parsedURI = new SimpleURI(address);
        ManagedChannel channel = ManagedChannelBuilder.forAddress(parsedURI.getHost(), parsedURI.getPort())
                .intercept(new ClientTracingInstrumentationInterceptor("client-java-instrumentation"))
                .usePlaintext(true).build();
        client.overrideChannel(channel);
```

Right? I'll remove benchmarking tracing in `GraknClient` for now.

flyingsilverfin:
The approach of overriding the client's `Channel` has been shown to work and moves the responsibility of instrumenting the client to packages outside the client

PR #4926, opened 2019-02-17
Refactor client-java to not depend on grakn-core server

haikalpribadi:
# Why is this PR needed?

1) Issue #4513: we need to remove any direct and transitive dependencies from //client-java (`grakn.core.client`) to //server (`grakn.core.server` and `grakn.core.graql`)
2) Issue #4818: we need to split //client-java into its own repository so that it can have its own release cycle, independent of grakn-core
3) Issue #4511: we need to split //client-java for 2 usages: benchmarking and normal use

# What does this PR do?

Before we can achieve the three goals above, we need to refactor //client-java significantly.

_(IN PROGRESS)_


haikalpribadi:
It turns out issue #4513 was resolved with PR #4976 and #4977; issue #4511 doesn't seem to need solving at all; and issue #4818 will be done in a separate PR.

PR #4932, opened 2019-02-18
Readd client server tracing

flyingsilverfin:
Fix linux and MacOS errors where the `protocol` java dependencies were added twice, once via the `grakn.benchmark.lib` indirect dependencies. The solution was to exclude `grakn.core:protocol` in the `dependencies.yml` file where `lib` is brought in.

Re-adds tracing implementation for `client-java` and `server` packages in core.


client-java/BUILD
```diff
@@ -31,6 +31,8 @@ java_library(
         "@graknlabs_graql//java:graql",
         "//server:server",
 
+         "//dependencies/maven/artifacts/grakn/benchmark:lib",
```
haikalpribadi: This should be replaced with Bazel, @flyingsilverfin. Make sure to plan the migration to Bazel soon.

PR #4934, opened 2019-02-19
Improve README

vmax:

- `README` becomes cleaner

- Specify the target to build (`//...` ["compile everything"] is not really helping the user)
- Remove installing Workbase node dependencies (it's now [in a separate repo](https://github.com/graknlabs/workbase))
- Remove `rpmbuild` requirement — it's not needed if you're building `//:distribution`

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,17 +52,11 @@ You can visit the [Setup Guide](https://dev.grakn.ai/docs/running-grakn/install-
 1. Make sure you have the following dependencies installed on your machine:
     - Java 8
     - Python >= 2.7 and Pip >= 18.1
-    - [Bazel >= 0.18.0](https://docs.bazel.build/versions/master/install-os-x.html)
-    - `rpmbuild`, available via `apt install rpm` (Debian derivatives) or `brew install rpm` (macOS)
-
-2. Install Workbase requirements:
+    - [Bazel >= 0.18.0](http://docs.bazel.build/install.html)
+1. Compile (you should get `bazel-genfiles/grakn-core-all.zip` as a result) 
 ```
-$ bazel run @nodejs//:npm install
-```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,17 +52,11 @@ You can visit the [Setup Guide](https://dev.grakn.ai/docs/running-grakn/install-
 1. Make sure you have the following dependencies installed on your machine:
     - Java 8
     - Python >= 2.7 and Pip >= 18.1
-    - [Bazel >= 0.18.0](https://docs.bazel.build/versions/master/install-os-x.html)
```
haikalpribadi: Do we still need Python for `@graknlabs_grakn_core` @vmax? And does Bazel have to be 0.18.0 and above? Bazel is so much later than that now, I think there's no need to put the version requirement.

vmax: @haikalpribadi 

>Do we still need Python for `@graknlabs_grakn_core`

We do — scripts that build distribution are in Python (https://github.com/graknlabs/bazel-distribution/tree/master/distribution)

>And does Bazel have to be 0.18.0 and above? Bazel is so much later than that now, I think there's no need to put the version requirement.

Alright then, let's remove the version altogether & assume users install the latest. Alternatively, we could utilize [`bazelisk`](https://github.com/philwo/bazelisk) which lets you to put version in `.bazelversion` and always use that

haikalpribadi: Nah, Bazel is evolving (and releasing) so rapidly at the moment, I'd rather be on the latest Bazel all the time.

vmax: @haikalpribadi fixed in https://github.com/graknlabs/grakn/pull/4934/commits/2e1b567732e8c8b2c0a75a5771f9f20705cd2946

Issue #4940, opened 2019-02-19
client-java does not set a client-side timeout, and will stall forever if the network gets interrupted


lolski:
**Grakn version**:
1.5

**Reproducible step**:
1. Start grakn: `grakn server start`
2. Run the following queries using client-java:
```
try (GraknClient.Session session = new GraknClient("localhost:48555").session("grakn")) {
        while (true) {
                GraknClient.Transaction tx = session.transaction(Transaction.Type.WRITE);
                tx.execute(Graql.match(var("x").sub("thing")).get());
                System.out.println("query");
                tx.commit();
        }
}
```
3. Pause grakn in the middle: `kill -STOP <grakn process id>`
4. Observe that the queries will stall and never times out

**TODO**
1. The client should be able to set a timeout when instantiating the client. We need to set a default, eg., 15 seconds.
```
client1 = new GraknClient(DEFAULT_URI, 20) // timeout set to 20 seconds
client2 = new GraknClient(DEFAULT_URI) // timeout set to the default timeout, eg., 15 seconds
```
2. The provided timeout value should be used to set the deadline of GRPC requests.

krishnangovindraj:
The server now has a default timeout, clients have a timeout option as well.

PR #4941, opened 2019-02-19
[DO NOT MERGE] Add timeout setting to client-java so it doesn't stall forever in case the network is interrupted

lolski:

Closes https://github.com/graknlabs/grakn/issues/4940. This adds a proper timeout of 15 seconds on the client side so it does not stall forever if the network is interrupted.

Adds a timeout of 15 seconds to `KeyspaceService` and `SessionService`'s blocking and non-blocking stub (the latter used by `Transceiver`)

### DIFF
--- a/client-java/GraknClient.java
+++ b/client-java/GraknClient.java
@@ -84,7 +84,7 @@ public final class GraknClient {
 
     public static final String DEFAULT_URI = "localhost:48555";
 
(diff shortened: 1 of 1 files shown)


lolski:
@marco-scoppetta actually I think it's better if the client can configure it him/herself, right?

PR #4951, opened 2019-02-22
Remove  JoinExplanation dep from ConceptMap

kasper-piskorski:


test-integration/graql/reasoner/benchmark/BUILD
```diff
@@ -13,7 +13,7 @@ java_test(
             "//dependencies/maven/artifacts/org/slf4j:slf4j-api",
             "//dependencies/maven/artifacts/com/google/guava:guava",
         ],
-    size = "large",
+    size = "enormous",
```
haikalpribadi: Why did this test size had to be increased to `enormous` @kasper-piskorski ? did it cause any CI job to fail?

PR #4967, opened 2019-02-26
Disable automatically updating docs repo when building on forks

vmax:

Currently CI on `grakn-core` forks is red because `grakn-docs-update` is failing. Moreover, it only fails because forks tend to not have `GRABL_CREDENTIAL` — and this is implicit.

- checks whether CircleCI is building mainline `grakn-core`
- if we're building a fork, successfully exit prematurely
- if we're building mainline repo and we *don't* have `GRABL_CREDENTIAL` set, print out descriptive error message

### DIFF
--- a/.circleci/grakn-docs-update.py
+++ b/.circleci/grakn-docs-update.py
@@ -2,6 +2,16 @@ import os
 import re
 import subprocess as sp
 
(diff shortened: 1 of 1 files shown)


vmax:
@haikalpribadi 
>I think this is a patch that we don't want to keep as a long term solution

Agreed. Just wanted to unbreak things temporarily.

PR #4984, opened 2019-02-27
Bring back Grakn Core logo

marco-scoppetta:

grakn core logo was missing due to new distribution package structure

Update path to logo file and minor cleanups
Also updated the timer for Grakn bootup

### DIFF
--- a/server/BUILD
+++ b/server/BUILD
@@ -50,7 +50,6 @@ java_library(
         "//dependencies/maven/artifacts/com/google/code/findbugs:jsr305",
         "//dependencies/maven/artifacts/com/google/guava:guava",
         "//dependencies/maven/artifacts/commons-configuration:commons-configuration", # PREVIOUSLY UNDECLARED
-        "//dependencies/maven/artifacts/commons-lang:commons-lang",
         "//dependencies/maven/artifacts/commons-cli:commons-cli", # PREVIOUSLY UNDECLARED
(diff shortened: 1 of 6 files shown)


server/BUILD
```diff
@@ -50,7 +50,6 @@ java_library(
         "//dependencies/maven/artifacts/com/google/code/findbugs:jsr305",
         "//dependencies/maven/artifacts/com/google/guava:guava",
         "//dependencies/maven/artifacts/commons-configuration:commons-configuration", # PREVIOUSLY UNDECLARED
-        "//dependencies/maven/artifacts/commons-lang:commons-lang",
```
haikalpribadi: @marco-scoppetta removing `:commons-lang` in `//server` breaks the build for me. Did break for you when you tested it? I'm not sure if it's your-commit, or if it your-commit-AND-my-commit that breaks this. My commit in Graql (https://github.com/graknlabs/graql/pull/22) removes `:commons-lang` from `@graknlabs_graql//java:graql`, which is a transitive dependency of `//server`. If your-commit alone breaks it, then it's a compile-time deps. If your-commit-AND-my-commit breaks it then `:commons-lang` must be a runtime deps.

haikalpribadi: Okay it turns out it's a "transitive dependency" of server (i.e. we're not using it directly), but it is needed during "compile time", and prior to my Graql PR (https://github.com/graknlabs/graql/pull/22) it was brought in by Graql.

PR #5018, opened 2019-03-11
Explicitly close RocksDB objects

marco-scoppetta:

Close RocksDB objects used in AttributeDeduplicator

- update RocksDb version
- make sure to close all RocksDb object (they are C++ objects which need explicit `close()`)
- disable SparkUI (which creates unused threads)
- add check before removing caches Session


server/resources/default-configs.properties
```diff
@@ -67,4 +67,5 @@ gremlin.spark.persistContext=true
 spark.master=local[*]
 spark.serializer=org.apache.spark.serializer.KryoSerializer
 spark.kryo.registrator=org.apache.tinkerpop.gremlin.spark.structure.io.gryo.GryoRegistrator
-spark.ui.showConsoleProgress=false
\ No newline at end of file
+spark.ui.showConsoleProgress=false
+spark.ui.enabled=false
```
lolski: is this needed?

marco-scoppetta: yes apparently (discovered via VisualVM) the Spark UI creates a lot of threads (~10) when we use analytics queries, given that we never access the Spark web UI dashboard in production it's probably best to disable it for the time being

PR #5025, opened 2019-03-13
Remove unscalable rule features

kasper-piskorski:

Refine features that scale poorly with the number of rules:
- rule priority - simplified it. Now we only check locally for rule resolvability (instead of looking at number of dependent rules).
- do not include priority when printing
- cached conclusion atom calculation
- avoiding composite query construction when rewriting rules


haikalpribadi:

Refine features that scale poorly with the number of rules:
- rule priority - simplified it. Now we only check locally for rule resolvability (instead of looking at number of dependent rules).
- do not include priority when printing
- cached conclusion atom calculation
- avoiding composite query construction when rewriting rules

PR #5036, opened 2019-03-17
Distribution (Assembly and Deployment) Tests

lolski:
Added test assembly and test deployments. These tests are designed to verify whether Grakn distributions which comes in various formats (Zip, Tar, Deb, RPM and Docker Image) works on the following target operating systems: Mac, Windows, Ubuntu, CentOS and Docker.

1. `test-assembly-windows-zip` (currently disabled due to [this particular error](https://circleci.com/gh/lolski/grakn/3869?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) from `build-tools`: `error: invalid command 'bdist_wheel'`
2. `test-assembly-mac-zip`
3. `test-assembly-linux-targz`
4. `test-assembly-linux-apt` and `test-deployment-linux-apt`
5. `test-assembly-linux-rpm` and `test-deployment-linux-rpm`
6. `test-assembly-docker`

## Remaining Issues 
1. test-deployment-linux-rpm - The yum install command needs to specify the exact Grakn version #5037
2. test-deployment-linux-rpm - fix infinite symlink expansion issue in CentOS and enable Bazel test #5038
3. test-assembly-windows-zip - enable the test after fixing the python issue regarding missing bdist_wheel #5039
4. test-assembly-windows-zip - need to verify that Grakn can be started from within a directory of which the name contains whitespace #5040
5. test-deployment-linux-apt - figure out why repo.grakn.ai only keeps the latest deb package #5041
6. test-assembly-docker - needs to add timeout #5042
7. Move test-end-to-end and test-integration to //test package for consistency #5043


.circleci/config.yml
```diff
@@ -16,6 +16,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
+
 version: 2.1
 commands:
   install-bazel:
```
haikalpribadi: Now that there is `install-bazel-mac`, let's rename this to `install-bazel-linux`, Ganesh.

.circleci/config.yml
```diff
@@ -125,6 +139,132 @@ jobs:
...
+      - run:
+          command: |
+            bazel build //:assemble-mac-zip
+            unzip bazel-genfiles/grakn-core-all-mac.zip -d bazel-genfiles/
+            nohup bazel-genfiles/grakn-core-all-mac/grakn server start
+      - run: bazel test //test-deployment:test-deployment --test_output=streamed --spawn_strategy=standalone --cache_test_results=no
+      - run: bazel-genfiles/grakn-core-all-mac/grakn server stop
+
+# TODO: add after fixing the python issue regarding missing bdist_wheel
+#  test-assembly-windows-zip:
```
haikalpribadi: Should we fix the python issue first so we can re-enable this? or is it too big and we want to skip it for now?

.circleci/config.yml
```diff
@@ -125,6 +139,132 @@ jobs:
...
+      - install-bazel
+      - run:
+          command: |
+            bazel build //:assemble-linux-targz
+            tar -xf bazel-genfiles/grakn-core-all-linux.tar.gz -C bazel-genfiles
+            nohup bazel-genfiles/grakn-core-all-linux/grakn server start
+      - run: bazel test //test-deployment:test-deployment --test_output=streamed --spawn_strategy=standalone --cache_test_results=no
+      - run: bazel-genfiles/grakn-core-all-linux/grakn server stop
+
+  test-assembly-apt:
```
haikalpribadi: I'm starting to see some inconsistencies in the namings of our distribution Bazel targets and CI jobs. Do you think we should make them consistent? 
E.g. should we rename this `test-assembly-linux-apt` because we have `test-assembly-linux-targz`? Or should we rename `test-assembly-linux-targz` to be `test-assembly-targz-linux`? The same question applies to `assemble-*` and `deploy-*` targets. I'm actually not so sure of the answer... 🤔

.circleci/config.yml
```diff
@@ -193,6 +333,42 @@ workflows:
...
+            branches:
+              only: test-deployment # TODO: change back to master before merging the PR
+#      - test-assembly-windows-zip:
+#          filters:
+#            branches:
+#              only: test-deployment # TODO: change back to master before merging the PR
+      - test-assembly-linux-targz:
+          filters:
+            branches:
+              only: test-deployment # TODO: change back to master before merging the PR
```
haikalpribadi: We forgot to set these branch filters back to `master`

.circleci/test/assembly/windows-zip/README.md
```diff
@@ -0,0 +1,61 @@
+# Workbase CI on Windows
```
haikalpribadi: We need to put these code in `//test/assembly/...` rather than `.circleci/`, because `.circleci/` is hidden away from the main development environment and make it seem that this code is note "first class citizen" in the code base. And because we also want to treat `//test` as main code to take care of.

haikalpribadi: And does the package name need to be `../windows-zip`? I.e. Do/Will we have tests for `../windows-<non-zip>`? If so, will they belong inside or outside of this package? 

Given that:
a) we don't have any other assembly tests for windows,
b) if we do they will go in the same package, and
c) you already name the python test script `windows-zip.py`, therefore...
I think this package should be called just `../windows`

test-deployment/BUILD
```diff
@@ -0,0 +1,21 @@
...
+        "@graknlabs_grakn_core//concept"
+    ]
+)
+
+checkstyle_test(
+    name = "checkstyle",
+    targets = [
+        ":test-deployment"
+    ],
+)
```
haikalpribadi: I think this package should be call `//test/deployment` instead of `//test-deployment`, then `//test` can also contain `../assembly` from `./circleci` as commented above.

.circleci/config.yml
```diff
@@ -125,6 +139,132 @@ jobs:
...
+      - attach_workspace:
+          at: ~/circleci-workspace
+      - install-bazel
+      - run: echo "deb [ arch=all ] https://repo.grakn.ai/repository/test-deb/ trusty main" | sudo tee -a /etc/apt/sources.list.d/grakn-core.list
+      - run: sudo apt-key adv --keyserver keyserver.ubuntu.com --recv 8F3DA4B5E9AEF44C
+      - run: sudo apt update
+      - run:
+          command: |
+            mv ~/circleci-workspace/VERSION VERSION
+            cat VERSION
```
haikalpribadi: I find it easier to read the config files when we just write these commands like this, ganesh:
```
- run: mv ~/circleci-workspace/VERSION VERSION
- run: cat VERSION
```
yes, it creates more "CI" steps in the web browser, but it makes the config file more "aligned" and "uniform".

.circleci/config.yml
```diff
@@ -125,6 +139,132 @@ jobs:
...
+      - install-bazel
+      - run: sudo apt install rpm
+      - run:
+          command: |
+            echo $(cat VERSION)-$CIRCLE_SHA1 > VERSION
+            sed -i -e 's/-/_/g' VERSION
+            cat VERSION
+      - run: bazel build //bin:assemble-rpm
+      - run: bazel build //server:assemble-rpm
+      - run: bazel build //console:assemble-rpm
```
haikalpribadi: I think we don't need to call `bazel build <target>`, given that `bazel run <target>` will trigger them too. It will simplify the CI file. Unless if you need to do something in the middle, like in `test-assembly-apt` above.

.circleci/config.yml
lolski: Yes, I want to submit a follow-up PR with other improvements to the test as a whole

.circleci/config.yml
lolski: Fixed in 49716f26ca94d41f3d86c8bf6a4082512f23ca8b

test-deployment/BUILD
lolski: Can we do it on a follow up PR? In that PR we can also move `test-integration` and `test-end-to-end` so everything is consistent:
```
test/
    integration
    end-to-end
    deployment
```

.circleci/config.yml
lolski: good point, fixed in cc8d8f2a3b2b4bf079e842c37e9340a2df58459e

.circleci/config.yml
lolski: Fixed in 49716f26ca94d41f3d86c8bf6a4082512f23ca8b

.circleci/config.yml
lolski: good point, fixed in cc8d8f2a3b2b4bf079e842c37e9340a2df58459e

.circleci/config.yml
lolski: fixed in 39c0b57d6e8d9e7b2e83d70ccfeaaf5e70757609

.circleci/config.yml
lolski: Personally I think `test-assembly-linux-apt` makes sense as the naming becomes OS-centric: "this is a test for a Grakn distribution for Linux of which the format is Apt"

test-deployment/BUILD
haikalpribadi: Well, I think we can make this first PR to be correct without stretching the scope, and still fix the whole thing in the next PR. I.e.:
1. Move `//test/deployment` in this PR.
2. Move  the old tests (which we may remove soon) to `//test/end-to-end` and `//test/integration` in the next PR

.circleci/config.yml
haikalpribadi: Actually, even `cat VERSION` doesn't need a `name:`, Ganesh. It's still a very basic bash command that it doesn't need to be too verbose.

.circleci/test/assembly/windows-zip/README.md
lolski: I agree we should call the folder `windows`. Should then the `py` file be called `zip.py`? eg., `test/assembly/windows/zip.py`?

lolski: Fixed in f7c3ce3daa038fe4dc7aff7f125f620e64e3dfa9

.circleci/config.yml
lolski: fixed in f7c3ce3daa038fe4dc7aff7f125f620e64e3dfa9

.circleci/config.yml
haikalpribadi: That's what I'm feeling too. would this apply to build target names too, @lolski?

.circleci/config.yml
haikalpribadi: there's another in `test-assembly-rpm` @lolski

.circleci/config.yml
lolski: I think so.

test-deployment/BUILD
lolski: Done in 1e1eb2e79b68e3d3952a045e5c6f9752ae37815d.

The structure becomes like this:
```
test/
    assembly
    deployment
    common # contains 'GraknApplicationTest' which is shared by both assembly and deployment
```

.circleci/config.yml
lolski: Whoops! Fixed ccf8b3ec1019daf7ec8bf6d9a0258786868a43b4

.circleci/test/assembly/windows-zip/README.md
lolski: Fixed in 1e1eb2e79b68e3d3952a045e5c6f9752ae37815d

The structure becomes like this:
```
test/
  assembly
  deployment
  common     # for things shared by both test-assembly and test-deployment (eg., 'GraknApplicationTest')
```

.circleci/config.yml
haikalpribadi: okay cool, can you make the changes in this PR, @lolski?

lolski: fixed in 54e8de318bec4742eb2b48e82a33c73fb263b41d

test-deployment/BUILD
lolski: There will be a follow-up PR for this issue: https://github.com/graknlabs/grakn/issues/5043

PR #5046, opened 2019-03-19
Introduce extra rule pruning in RuleCache

kasper-piskorski:

Improve and test query execution time scaling with number of rules.

- introduced rule pruning based on type instance absence in the rule graph
- added conjunctive query shortcircuiting based on the type instance absence
- added a scalable non-trivial test stressing the rule application aspect
- refactored `RuleApplicabilityIT`

## Future work
- the prunable types should be stored/extracted from db statistics layer


test-integration/graql/reasoner/cache/RuleCacheIT.java
```diff
@@ -128,11 +132,21 @@ public void whenDeletingARule_cacheContainsUpdatedEntry(){
...
+            Type description = tx.getType(Label.of("description"));
+            Set<Rule> filteredRules = description.thenRules()
+                    .filter(r -> tx.stream(Graql.match(r.when())).findFirst().isPresent())
+                    .collect(Collectors.toSet());
+            Set<Rule> fetchedRules = tx.ruleCache().getRulesWithType(description).collect(Collectors.toSet());
+            //NB:db lookup filters more aggressively, hence we check for containment
+            assertTrue(fetchedRules.containsAll(filteredRules));
+
+            //even though rules are filtered, the type has instances
+            assertTrue(!fetchedRules.isEmpty());
```
marco-scoppetta: or `assertFalse(fetchedRules.isEmpty())`  😄

PR #5050, opened 2019-03-19
Fix ghost vertices issue with concurrent insertions

marco-scoppetta:

Resolve exception `TemporaryWriteException: Index overlap has led to the accidental sharing of a partially complete vertex {v[...]}`
which is often occurring when migrating (inserting) a lot of attributes into Grakn using multiple concurrent threads.

In this PR we serialise the `commit`s to Janus to avoid sharing incomplete vertices between concurrent transactions.
Apparently Janus makes indices immediately available to all transactions while committing,
even to transactions that were already open before starting the `commit` process.

Example
Scenario: two concurrent transactions are trying to insert an attribute instance `name=Marco`

- Before inserting the attribute, Transaction A will do a Janus lookup by `INDEX` property (`hasProperty(INDEX, "ATTRIBUTE-name-Marco")`)
- This will return null, so Transaction A will proceed to create the attribute Vertex and add the property `INDEX` onto the vertex
- While Transaction A is committing, Transaction B executes the same lookup by `INDEX` property,
at this point Janus will return the vertex (as `CacheVertex`) that Transaction A is committing
- Transaction B then will try to access `BASE_TYPE` vertex label or any other property of the vertex, but they are all NULL -> Exception is thrown because it looks like we are handling a ghost vertex here

The above is an educated guess, further investigation is needed.

For the time being we are solving the issue by using a ReadWriteLock so that concurrent transaction will never be able to end up in the situation described in the example.
Each transaction will acquire a read lock (non exclusive) when looking up any vertex by property (properties on vertices are all indexed) and will acquire a write lock (exclusive) when committing to Janus.

### DIFF
--- a/server/src/server/session/TransactionOLTP.java
+++ b/server/src/server/session/TransactionOLTP.java
@@ -19,7 +19,6 @@
 package grakn.core.server.session;
 
 import brave.ScopedSpan;
-import brave.Span;
 import grakn.benchmark.lib.serverinstrumentation.ServerTracingInstrumentation;
 import grakn.core.api.Transaction;
 import grakn.core.common.config.ConfigKey;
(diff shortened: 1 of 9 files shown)


server/src/server/session/TransactionOLTP.java
```diff
@@ -381,15 +404,24 @@ public ElementFactory factory() {
...
-        } else {
-            return Optional.empty();
+            Vertex vertex;
+            // Get read lock before accessing graph so that if a commit is occurring, we wait
+            // Without lock, when inserting a lot of attributes this traversal might return ghost vertices
+            graphLock.readLock().lock();
+            try {
+                vertex = vertices.next();
+            } finally {
+                graphLock.readLock().unlock();
```
flyingsilverfin: good catch with this `finally` could be very bad otherwise!

PR #5087, opened 2019-04-04
Query planner refactor 1

flyingsilverfin:
Begin cleaning up very procedural and overly generic code in Query Planner to be more understandible.

* Remove Generic from `DirectedEdge<T>` and all associated classes, including tests. Usage was early exclusively `DirectedEdge<Node>`. Now none of the classes in the Query Planner accept a generic. It seems like the generic originally existed to allow using an Integer as a Node, rather than a Node object in tests - tests for now use `Node` with the Integer encoded as a variable string.
* `Node` had methods `addIfPresent(a, b, allNodes)`, which did a create-if-present in the `allNodes` map and returned the Node. This functionality is moved out. 
* `NodeId` now uses a factory constructor because the instances are not meant to be unique
* Move a large chunk of isolated code from `GreedyTraversalPlan.java` into `RelationTypeInference.java` - all to do with inferring types from role players to generate more Label fragments.
* `Fragment`s provide the `Node`s that  go into the internal QueryPlanner traversal graph - Fragments that represent Janus edges for instance provide a `middle` node that is not a valid starting point. These are collected into the set of all Nodes at the start of query planning now rather than deep in the chain of c-like function calls.
* Move chunks of code into their own methods to deal with later (`chooseStartingNodes`, `buildDependenciesBetweenNodes`)
* _important_ experimental change to NOT access all the indexed vertices first thing in the query plan - do them as a normal step in the flattened minimum spanning tree as required (starting point is still almost always an indexed vertex) 
* Remove `lowPriorityStartingNodes` as we currently always reify nodes, so implicit relationships are as good of a starting point as any


server/src/graql/gremlin/GreedyTraversalPlan.java
```diff
@@ -103,80 +89,69 @@ public static GraqlTraversal createTraversal(Pattern pattern, TransactionOLTP tx
      * @return a semi-optimal traversal plan to execute the given conjunction
      */
     private static List<Fragment> planForConjunction(ConjunctionQuery query, TransactionOLTP tx) {
+        // a query plan is an ordered list of fragments
+        final List<Fragment> plan = new ArrayList<>();
 
-        final List<Fragment> plan = new ArrayList<>(); // this will be the final plan
-        final Map<NodeId, Node> allNodes = new HashMap<>(); // all the nodes in the spanning tree
+        // flatten all the possible fragments from the conjunction query (these become edges in the query graph)
+        final Set<Fragment> allFragments = query.getEquivalentFragmentSets().stream()
```
kasper-piskorski: if these are not changed later we might consider making this set immutable

server/src/graql/gremlin/GreedyTraversalPlan.java
```diff
@@ -195,139 +170,65 @@ public static GraqlTraversal createTraversal(Pattern pattern, TransactionOLTP tx
...
-                            Iterators.getOnlyElement(((LabelFragment) fragment).labels().iterator()));
-                    if (schemaConcept != null && !schemaConcept.isRole() && !schemaConcept.isRule()) {
-                        labelVarTypeMap.put(fragment.start(), schemaConcept.asType());
-                    }
-                });
-        return labelVarTypeMap;
-    }
+                // it's either neq or value fragment
+                Node start = allNodes.get(NodeId.of(NodeId.NodeType.VAR, fragment.start()));
+                Node other = allNodes.get(NodeId.of(NodeId.NodeType.VAR, fragment.dependencies().iterator().next()));
```
kasper-piskorski: if we are sure there's is only one element we can do `Iterables.getOnlyElement`

server/src/graql/gremlin/fragment/AbstractRolePlayerFragment.java
```diff
@@ -83,9 +88,34 @@ final String innerName() {
...
+    public final Set<Weighted<DirectedEdge>> directedEdges(Map<NodeId, Node> nodes,
+                                                           Map<Node, Map<Node, Fragment>> edges) {
+
+        // this is a somewhat special case, where the middle node being converted to a vertex
+        // may be addressed by a variable
+
+        Node start = nodes.get(NodeId.of(NodeId.NodeType.VAR, start()));
+        Node end = nodes.get(NodeId.of(NodeId.NodeType.VAR, end()));
+        Node middle = nodes.get(NodeId.of(NodeId.NodeType.VAR, edge()));
+        middle.setInvalidStartingPoint();
```
kasper-piskorski: maybe we could use `getNodes` here?

server/src/graql/gremlin/fragment/AbstractRolePlayerFragment.java
```diff
@@ -83,9 +88,34 @@ final String innerName() {
...
+
+        if (!edges.containsKey(middle)) {
```
kasper-piskorski: just wondering, what's the scenario when edges contain `middle` key?

server/src/graql/gremlin/fragment/Fragment.java
```diff
@@ -144,18 +146,58 @@ Variable end() {
...
+    /**
+     * When building the query plan spanning tree, every fragment has a start defined with a variable
+     * Some fragments are actually edges in JanusGraph (such as isa, sub, etc.)
+     * These require another variable for the end() variable, and to force the MST algorithm to
+     * traverse these JanusGraph edges too, we insert a fake middle node representing the edge
+     * @return
+     */
+    public Set<Node> getNodes() {
+        NodeId startNodeId = NodeId.of(NodeId.NodeType.VAR, start());
+        return new HashSet<>(Arrays.asList(new Node(startNodeId)));
```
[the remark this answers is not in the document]
haikalpribadi: Unless you strictly need Guava's Immutable collections, you can use our collection builders at: https://github.com/graknlabs/grakn/blob/master/common/util/Collections.java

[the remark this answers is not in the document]
haikalpribadi: And actually in most cases, our Collections util already behave as an unmodifiable collection, given that set is constructed within the native `Collections.unmodifiableSet(...)` wrapper, and no one has access to the internal object to modify it.

server/src/graql/gremlin/RelationTypeInference.java
```diff
@@ -0,0 +1,184 @@
...
+                    if (instanceVarTypeMap.containsKey(rolePlayer)) {
+                        numRolePlayersHaveType++;
+                    }
+                }
+                if (numRolePlayersHaveType < 2) {
+                    iterator.remove();
+                }
+            }
+        }
+        return relationRolePlayerMap;
```
marco-scoppetta: when possible it's better to avoid this pattern of creating a collection and then retrieve the iterator and manually remove what's not needed, we have many tools that can help us building the correct map in one go

server/src/graql/gremlin/spanningtree/graph/DenseWeightedGraph.java
```diff
@@ -35,50 +35,45 @@
 import static com.google.common.collect.Range.closedOpen;
 
 /**
- * @param <V> the type of the nodes stored
+ * Used for testing
```
marco-scoppetta: I don't have a context, but why do we have a full class that's only used for tests (according to your comment) in a folder that is not dedicated to tests? 
It was already in here but it might be worth considering to move it in the appropriate test folder

server/src/graql/gremlin/fragment/AbstractRolePlayerFragment.java
flyingsilverfin: This is a remaining piece of procedural code that modifies arguments in place - `edges` actually maps from `middle` node to `Fragment`s and the mapping is created here. Hoping to remove.

server/src/graql/gremlin/spanningtree/graph/DenseWeightedGraph.java
flyingsilverfin: Was wondering about this! Implemented in QP refactor 2

server/src/graql/gremlin/RelationTypeInference.java
flyingsilverfin: Implemented in QP refactor 2

PR #5100, opened 2019-04-09
Convert types if possible on WRITE

kasper-piskorski:

Allow for implicit type conversion if possible when doing WRITEs. As a result queries like:

```
define resource sub attribute, datatype double;
insert $x 10 isa attribute;
```

Will succeed as there exists a unambiguous conversion. Previously we didn't allow that.


server/src/server/kb/concept/AttributeImpl.java
```diff
@@ -52,11 +52,13 @@ private AttributeImpl(VertexElement vertexElement, AttributeType<D> type, D valu
     }
 
     public static <D> AttributeImpl<D> create(VertexElement vertexElement, AttributeType<D> type, D value) {
```
haikalpribadi: How can `value` ever have a different `Class` than was is defined in the schema when `type` is defined to be `AttributeType<D>`? This method signature already bounds the type, no? @kasper-piskorski @marco-scoppetta

server/src/server/kb/concept/TypeConverter.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    public static class DateConverter extends TypeConverter<LocalDateTime> {
+        @Override
+        public Object serialise(LocalDateTime value) {
+            return value.atZone(ZoneId.of("Z")).toInstant().toEpochMilli();
+        }
+
+        @Override
+        public LocalDateTime deserialise(Object value) {
```
haikalpribadi: Would it be better to restrict the input argument to always be `Long`?

server/src/server/kb/concept/TypeConverter.java
```diff
@@ -0,0 +1,133 @@
...
+            if (value instanceof LocalDateTime) return (LocalDateTime) value;
+            if (value instanceof Long) return LocalDateTime.ofInstant(Instant.ofEpochMilli((Long) value), ZoneId.of("Z"));
+            throw new ClassCastException();
+        }
+    }
+
+    public static class DoubleConverter extends TypeConverter<Double> {
+        @Override
+        public Double deserialise(Object value) {
```
haikalpribadi: I think the input arguments here should definitely be restricted to `Number`? Along with every other `TypeConverter`, for that matter.

server/src/server/kb/concept/TypeConverter.java
```diff
@@ -0,0 +1,133 @@
...
+            if (value instanceof Number){
```
haikalpribadi: what happens when you cast a `Long` to a `Double` or `Float`?

test-integration/graql/query/BUILD
```diff
@@ -20,161 +20,159 @@ load("@graknlabs_build_tools//checkstyle:rules.bzl", "checkstyle_test")
...
 )
 
 java_test(
     name = "graql-delete-it",
-    test_class = "grakn.core.graql.query.GraqlDeleteIT",
+    size = "medium",
     srcs = ["GraqlDeleteIT.java"],
+    test_class = "grakn.core.graql.query.GraqlDeleteIT",
     deps = [
-        "//common:common",
```
haikalpribadi: you should put back the target names, @kasper-piskorski . As you can see in other target definitions, we're making it a convention to write the full `//package/name:target-name`

server/src/server/kb/concept/TypeConverter.java
kasper-piskorski: We still want to throw if it is an unsupported conversion/not a Number, so I don't think so. Also that way the whole conversion logic is in one place.

test-integration/graql/query/BUILD
kasper-piskorski: it's not me, it's bazel/intellij

server/src/server/kb/concept/TypeConverter.java
kasper-piskorski: I don't get the question. If it's a Number (so Long as well) it gets converted to a Double.

server/src/server/kb/concept/TypeConverter.java
kasper-piskorski: Same argument as for Double. If we constrained it to Long we would need to convert somewhere else/make sure it's a Long. Here we treat general objects so the whole conversion logic sits in a single place.

server/src/server/kb/concept/AttributeImpl.java
kasper-piskorski: D is `Object` at run-time as generics in java only work at compile-time, there's no way to enforce the specific type at run-time.

server/src/server/kb/concept/TypeConverter.java
kasper-piskorski: Ok, i changed my mind!

server/src/server/kb/concept/ValueConverter.java
```diff
@@ -0,0 +1,96 @@
...
+        public TARGET convert(SOURCE value) { return (TARGET) value;}
+    }
+
+    public static class DateConverter extends ValueConverter<Object, LocalDateTime> {
+
+        @Override
+        public LocalDateTime convert(Object value) {
+            if (value instanceof LocalDateTime) return (LocalDateTime) value;
+            else if (value instanceof Long) {
+                LocalDateTime.ofInstant(Instant.ofEpochMilli((Long) value), ZoneId.of("Z"));
```
haikalpribadi: Shouldn't there be a `return` statement here, @kasper-piskorski ?

server/src/server/kb/concept/Serialiser.java
```diff
@@ -78,26 +72,23 @@ public VALUE serialise(VALUE value) {
...
 
-    public static class Date extends Serialiser<LocalDateTime, Long> {
-
-        Date() {
-            super();
-        }
+    public static class Date extends Serialiser<java.time.LocalDateTime, Long> {
 
         @Override
         public java.lang.Long serialise(LocalDateTime value) {
```
haikalpribadi: this seems to be remnant from old code, but might as well remove `java.lang.` prefix too.

server/src/server/kb/concept/AttributeImpl.java
haikalpribadi: So if `type` has class `AttributeType<Long>`, the last argument `value` is not bound to `Long` as well?

server/src/server/kb/concept/ValueConverter.java
kasper-piskorski: good spot!

PR #5103, opened 2019-04-10
Query planner refactor 3

flyingsilverfin:
Reduce redundant code introduced by QP refactors 1 and 2 and finish reorganizing the QP code into more distinct and well-named portions. Also add runtime checks on `match` queries for unbound variable-variable comparisons (both for concepts and values) and fix some bugs.

* Introduce new `EdgeFragment` from which all fragments that represent Janus edges inherit from, eliminating re-produced code between these classes
* Add check to ensure that statements must have at least one property (runtime check to disallow `match $x; get;`)
* Stricter checking for variables in comparisons. Now we also disallow comparisons between two unbound variables (such as `match $x !== $y` or `match $x != $y` that would lead to n^2 full DB checks). However, `match $x == "john"` is still allowed (ie. value comparisons between a single unbound variable and a value). We may want to be more strict with this in the future as well. 
* Wrote a few tests for the above two cases and fixed other tests that failed the new checks
* Fix a nastly bug introduced by the new `EdgeFragment` supertype in the QP to do with ordering graph nodes and validity of starting points
* Move tests from reasoner to non-reasoner packages, since they had nothing to do with reasoner
* Fix bug that occurs when querying `match $x == 100`, which can return SHARD vertices


server/src/graql/gremlin/GreedyTreeTraversal.java
```diff
@@ -64,7 +64,7 @@
 
             assert nodeWithMinCost != null : "reachableNodes is never empty, so there is always a minimum";
 
-            // add edge fragment first, then node fragments
+            // add edge fragment first, then node fragmentwhenQueryingForRelationWithNonRoleRoles_Throwss
```
marco-scoppetta: interesting comment

server/src/graql/gremlin/TraversalPlanner.java
```diff
@@ -56,18 +58,17 @@
...
  * Class for generating greedy traversal plans
  */
-public class GreedyTraversalPlan {
+public class TraversalPlanner {
 
-    protected static final Logger LOG = LoggerFactory.getLogger(GreedyTraversalPlan.class);
+    protected static final Logger LOG = LoggerFactory.getLogger(TraversalPlanner.class);
 
-    /**
+    /**```
```
marco-scoppetta: comments police

server/src/graql/gremlin/TraversalPlanner.java
```diff
@@ -110,33 +111,42 @@ public static GraqlTraversal createTraversal(Pattern pattern, TransactionOLTP tx
...
-            // find and include all the nodes not touched in the MST in the plan
-            Set<Node> connectedNodes = connectedFragments.stream()
-                    .flatMap(fragment -> fragment.getNodes().stream())
-                    .map(node -> queryGraphNodes.get(node.getNodeId()))
-                    .collect(Collectors.toSet());
-            plan.addAll(fragmentsForUnvisitedNodes(queryGraphNodes, connectedNodes));
+        // this shouldn't be necessary, but we keep it just in case of an edge case that we haven't thought of
+        List<Fragment> remainingFragments = fragmentsForUnvisitedNodes(queryGraphNodes, queryGraphNodes.values());
+        if (remainingFragments.size() > 0) {
+            LOG.error("Expected all fragments to be handled, but found these: " + remainingFragments);
```
marco-scoppetta: Given that we don't interrupt the execution of the code if `remainingFragments.size() > 0` we might want to print the message as a WARN rather than ERROR.

I am assuming that this case does not undermine the correct execution of the QueryPlanner

flyingsilverfin: Ahh warn is what i was looking for - better than debug or error!

PR #5123, opened 2019-04-18
[DO NOT MERGE] Janus micro benchmark

flyingsilverfin:

{ Please describe the goal of this PR, why they are valuable to achieve, and reference the related GitHub issues. }

{ Please explain what you implemented, why your changes are the best way to achieve the goal(s) above, and reference the GitHub issues to be automatically closed, such like 'closes #number'. }


marco-scoppetta:
Not sure we should add all the benchmarks to Grakn repo, they are not really tests 🤔 
and they are very specific for Janus.

Maybe add these things as module of benchmark?
I think in general we should have a discussion to decide where to put all benchmarks, given that we also need them for Reasoner.

What do you think @flyingsilverfin ?

flyingsilverfin:
This PR is being closed, with the goal of moving out of grakn core (doesn't fit as a standard test)

PR #5131, opened 2019-04-23
Enable query cache at tx level

### LINKED ISSUE #3449: Increase (reasoner) query cache lifetime
_This issue was originally posted by @kasper-piskorski  on 2018-04-16 18:16_.

Currently the lifetime of the cache is single input query. We could either increase it to be transaction or even longer (multiple transaction). NB: look at how core caches things.

### LINKED ISSUE #4215: Graql: query caching per transaction
_This issue was originally posted by @haikalpribadi  on 2018-08-14 20:09_.

This issue needs #3061, needs #3097, needs #3261, needs #3441, needs #3449, needs #3775

kasper-piskorski:
Increase the query cache lifetime to transaction. Previously the answers were reused only within a scope of the same disjunction. Now if we execute multiple queries within the same tx, answers will be cached and reused. 

Caveats: 
- only queries executed with reasoning on will be cached
- we only cache small queries (atomic)

- removed cache flush at the beginning of each disjunction processing
- closes #3449
- closes #4215


kasper-piskorski:
BAM!

PR #5132, opened 2019-04-24
Restrict the token patterns for type labels and ID values

haikalpribadi:

Graql now restricts the use of type labels and IDs in queries to no longer accept and string value, which could be arbitrary. To comply with this change, we've made sure that Grakn uses and implements this language behaviour appropriately, which predominantly changes the way tests are written in the codebase.

- `StringPrinter` for type labels and ID values no longer output the values in quotes (`"` or `'`)
- Integration tests for Graql (query, analytics and reasoner) are updated to comply with the new syntax

### DIFF
--- a/concept/printer/StringPrinter.java
+++ b/concept/printer/StringPrinter.java
@@ -40,8 +40,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import static graql.lang.util.StringUtil.escapeLabelOrId;
-
 /**
  * Default printer that prints results in Graql syntax
  *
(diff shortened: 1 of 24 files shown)


haikalpribadi:
I think that should be in Graql, rather than Grakn, @kasper-piskorski. The design and implementation all ends in Graql side, we don't need to test this in Grakn.

haikalpribadi:
Although, there should probably be a check that Grakn's Concept IDs are always within the bounds of Graql ID token patterns. @kasper-piskorski.

haikalpribadi:
Can you add the test in a follow-up PR and approve this one, @kasper-piskorski ?

kasper-piskorski:
The reason i'm asking is we seem to parse the ids in the tests everywhere. There's no pattern building with ids being tested.

haikalpribadi:
> There's no pattern building with ids being tested.

Yes, but this should be in https://github.com/graknlabs/graql - not in this repo. @kasper-piskorski

haikalpribadi:
Thanks, @kasper-piskorski. Can you add issue/PR on graknlabs/graql so we don't forget this one.

Issue #4666, opened 2018-11-28
"has attribute-label $var" in an insert throws error


sorsaffari:
Grakn version: 1.4.2

## Expected outcome:
```
>>> $ match $p id V90296 has last-name $ln; insert $np isa person has last-name $ln;
{$np id V28848 isa person;}
```

## Actual outcome
```
>>> $ match $p id V90296 has last-name $ln; insert $np isa person has last-name $ln;
INVALID_ARGUMENT: GraqlQueryException-cannot overwrite properties `$ln isa last-name` on  concept `Base Type [ATTRIBUTE] - Id [V28848] - Value [Lilywhite] `. Please check server logs for the stack trace.
```

sorsaffari:
@haikalpribadi same outcome with 1.5.0 on the `social_network` keyspace:

```
social_network> match $p isa person, has email "leona.rice@gmail.com", has name $n; insert $np isa person, has email "some@email.com", has name $n;
```

```
Error: INVALID_ARGUMENT: GraqlQueryException-cannot overwrite properties `$n isa name;` on  concept `Base Type [ATTRIBUTE] - Id [V32968] - Value [Loni] `. Please check server logs for the stack trace.
All uncommitted data is cleared
```

kasper-piskorski:
@flyingsilverfin didn't your last PR fix this? Can we double-check?

PR #5145, opened 2019-05-01
Allow reusing attributes in match-inserts

### LINKED ISSUE #5138: Inserting a new owner for an existing attribute (using a variable) throws an error
## Description

When making a match...insert query, if an attribute in the `match` is given a new owner in the `insert`, then the query will fail with an error.

## Environment

2. Grakn version (and platform): Grakn Core 1.5.1
3. Grakn client: console
4. Other environment details: -

## Reproducible Steps

```
define

name sub attribute,
    datatype string;

person sub entity,
    has name;

insert $x isa person, has name "bill";
```

## Expected Output

```
match $x isa person, has name $np; insert $e isa person, has name $np;
```
The above should successfully insert a new person who owns the pre-existing name attribute `"bill"`.

## Actual Output

```
match $x isa person, has name $np; insert $e isa person, has name $np;
Error: INVALID_ARGUMENT: GraqlQueryException-cannot overwrite properties `$np isa name;` on  concept `Base Type [ATTRIBUTE] - Id [V4208] - Value [bill] `. Please check server logs for the stack trace.
```
 
It seems the insert tries to overwrite the name rather than create ownership of the existing.

flyingsilverfin:
Previously, we disallowed the following: `match $x isa person; $x has name $n; insert $y isa person, has name $n` because Grakn tries to re-create the same name again, when it already exists. Now, this is a valid query and doesn't throw an exception when re-using attributes between the match and the insert statements. This means Grakn now also allows `match $x isa name; insert $x isa name;`, but silently does no insertions under the hood (however, it does retrieve all the names and iterates over them still). `match $x isa entity; insert $x;` and the equivalent for `relation` is still banned. 

Note that assigning another owner to an attribute instance works without issues using the ConceptAPI, so this is a bug/change restricted to Graql executors. 

In Summary, the following semantics are now implemented:

    1. `match $x isa name; insert $x isa first-name;` is a downcast, and throws an exception

    2. `match $x isa first-name; insert $x isa name;` Silently does nothing - semantically valid so it should'nt be an error. There's an argument for re-inserting the first-names as name instances again.

    3. `match $x isa person, has first-name $a; insert $y isa person, has name $a;` This is an upcast of the ownership, so semantically is ok. The current implementation still holds and inserts a @has-first-name implicit relation rather than a @has-name as it matches the concrete instance type.

    4. `match $x isa person, has name $a; insert $y isa person, has first-name $a;` Is an interesting case because it iterates over possible $a concrete values that are name or first-name. If we get a first-name, we can validly insert a new person with that same first-name. On the other hand, should we get a concrete name instance, we cannot downcast it to first-name so an exceptoin is thrown. The way this is data dependant is a bit confusing... Indicates these checks should be done on the query level against some kind of formal semantics.

* Updates to `IsaExecutor` that first checks if the variable being referenced already has a concept mapped to it. If yes, and its an attribute, pass, otherwise throw an exception.
* A set of tests that checks variants on the behavior described above has been added to `GraqlInsertIT`.

Closes #5138 and #4666


haikalpribadi:
First of all, very good PR description @flyingsilverfin !!! It made it very easy for me to understand the problem, goal and implementation, and I think it would go well once compiled into the release notes. cc: @graknlabs/developers 

I'll provide more feedback on the implementation in the PR page.

flyingsilverfin:
Couple of further investigations. Previous to this change:
* `match $x isa name; insert $x isa first-name;` throws the same `cannot overwrite properties` error. This behavior is downcasting and should ideally throw an exception
*  `match $x isa first-name; insert $x isa name;` throws the same `cannot overwrite properties` error. This behavior is upcasting, and theoretically is allowed. In Grakn however we have decided it should also throw an exception (?)
* `match $x isa person, has first-name $a; insert $y isa person, has name $a;` throws the same error. It should create a new `@has-name` relationship to the `first-name` instance
* `match $x isa person, has name $a; insert $y isa person, has first-name $a;' should throw an error as this is downcasting

PR #5151, opened 2019-05-02
Re-add Gremlin indexed nodes steps first

flyingsilverfin:
During the QP refactors done previously, indexed fragments were sometimes ordered after the variable's first occurence in the plan, which may have resulted in some slower plans. Now we explicltly put the indexed fragments at the start of the query plan again. Further investigation on the Janus level is required to see if this is strictly faster.

* When encountering a fixed cost fragment such as any indexed fragment, add it right to the start of the query plan instead of just appending it with other fragments
* remove special case in `nodeToPlanFragments` for visited/unvisited switch that was only being used in one place, move this code up into the caller.

### DIFF
--- a/test-integration/graql/query/QueryPlannerIT.java
+++ b/test-integration/graql/query/QueryPlannerIT.java
@@ -420,7 +420,7 @@ public class QueryPlannerIT {
                 z.isa(thingy3),
                 var().rel(x).rel(y).rel(z));
         plan = getPlan(pattern);
-        assertEquals(x.var(), plan.get(1).end());
+        assertEquals(x.var(), plan.get(3).end());
         assertEquals(3L, plan.stream().filter(fragment -> fragment instanceof NeqFragment).count());
 
         //TODO: should uncomment the following after updating cost of out-isa fragment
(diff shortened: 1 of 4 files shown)


test-integration/graql/query/QueryPlannerIT.java
```diff
@@ -300,13 +302,37 @@ public void noIndexedStartNodeGeneratesValidPlans() {
...
+
+        List<Fragment> plan = getPlan(pattern);
+        boolean priorFragmentHasFixedCost = true;
+        for (Fragment fragment : plan) {
+            if (!fragment.hasFixedFragmentCost()) {
+                priorFragmentHasFixedCost = false;
+            } else {
+                // if the current fragment is fixed cost
+                // and the prior one was not, throw and fail the test
+                if (!priorFragmentHasFixedCost) {
```
kasper-piskorski: Could we have an assertTrue or similar here? It can go with a message as well.

flyingsilverfin:
@kasper-piskorski  and I have decided that due to statistics making some cases of nondeterministic query plans (soon) non-ambiguous, and that this use case is mostly restricted to symmetric query patterns, and that the overhead of the proper implementation of the full message passing algorithm consumes about 10-40% of the query planning time (can be 0.5-10 ms, which is quite high)

kasper-piskorski:
excluding indexed things, does it preserve the ordering?

PR #5154, opened 2019-05-03
More Deterministic Query Plans

flyingsilverfin:
Make Grakn's query plans more deterministic - previously the plan chosen was as random as java's set stream implementation which allows any random order (no ordering guarantees). Now, the query plans are far more deterministic (the caveats for "far more" are explained further down). This means that when a query is slow, it will likely always be slow, and fast queries will likely always be fast. This should be very helpful for debugging, as nondeterminism is hard to solve and pin down.

Cases where we may still end up with nondeterministic plans:
Any case where the two-hop neighborhood of a query variable is symmetrical, but beyond that it isn't.

In other words, if we have a pseudocode query like (illustrated with one-hop rather than two-hops):
```
$x isa $A; $x isa $B; ($x, $z);
$B sub $K; 
($z, $y); $y isa $A; $y isa $B; 
$B sub $L;
```
In this query, the graph that is formed for query planning induces local symmetry/identical neighborhoods about $x and $y. Only if we go to two-hops do we find a difference: $K versus $L.

This is analogous to the solution implemented here, except we take into account two hops rather than 1 hop. However, quite a few cases are identical even with their two-hop neighborhood (imagine x ISA y, which has some attribute value via an implicit relation with some value -- the actual value is not included in the numerical representation of x.

This may be more comprehensible after reading the next section, which explains how this change was achieved.

* Adding a hashing mechanism to query planner `Node`s that identifies the Node using a hash of the fragments it represents, plus the hashes of one and two neighbors.
* Adding a hash to `Fragment`s that ignores variables -- only taking into account the fragment's class name and any deterministic information such as labels, comparators, IDs, etc. This is means that Fragments are considered equal by this hash function up to _alpha equivalence_, in other words ignoring variables. The public method for this is called `Fragment.variableAgnosticHash()`, which hashes a combination of the protected abstract method `Fragment.alphaEquivalentHash()` and the fragment's class name the `variableAgnosticHash()` value.
* `Node` has two new hash functions - `localHashCode()` which is the has of fragments it contains, and `globalHashCode()` which is set from externally.
* `TraversalPlanner.injectDeterministicNodeHashes(graph)` computes a one-hop hash for each node, using neighbor's Nodes' local hashes, then repeats the process combining the neighbor's one-hop hashes, implicitly bringing the values from two hops away into each node.
 
## Follow up
A complete solution that would be entirely deterministic (if implemented without errors) would implement a pregel/network broadcast algorithm that doesn't stop at one-hop  or two hops away, but instead hashes a combination of every other node in the graph in combination with the distance. This way nodes only have the same hash if the entire graph is. This is a time/quality tradeoff - the cost of such an algorithm would grow quickly with the size of the query, though some efficient network routing algorithms probably exist.

**note: the index fragment re-ordering will be added in from another PR


server/src/graql/gremlin/TraversalPlanner.java
```diff
@@ -181,11 +182,14 @@ public static GraqlTraversal createTraversal(Pattern pattern, TransactionOLTP tx
...
             // sparse graph for better performance
             SparseWeightedGraph sparseWeightedGraph = SparseWeightedGraph.from(weightedGraph);
+            injectDeterministicNodeHashes(sparseWeightedGraph);
             Set<Node> startingNodes = chooseStartingNodeSet(connectedFragments, nodes, sparseWeightedGraph);
 
             // find the minimum spanning tree for each root
             // then get the tree with minimum weight
             Arborescence<Node> arborescence = startingNodes.stream()
+                    .sorted(Comparator.comparing(Node::globalHashCode)) // this sort ensures we determinstically order the nodes
+                    .collect(Collectors.toList()).stream()
```
kasper-piskorski: hmmm ;p

server/src/graql/gremlin/fragment/AbstractRolePlayerFragment.java
```diff
@@ -112,4 +115,14 @@ static void traverseToRole(
...
     }
+
+    @Override
+    public int alphaEquivalentHash() {
+        // convert sets into ordered lists, based on the hashcodes
+        List<Label> labels = relationTypeLabels().stream().sorted().collect(Collectors.toList());
+        labels.addAll(roleLabels().stream().sorted().collect(Collectors.toList()));
+
+        Label[] labelsArray = labels.toArray(new Label[] {});
+        return Arrays.hashCode(labelsArray);
```
kasper-piskorski: why do we use arrays? we can get the hashcode from the list as well

server/src/graql/gremlin/fragment/LabelFragment.java
```diff
@@ -87,4 +89,10 @@ public Long getShardCount(TransactionOLTP tx) {
                 .mapToLong(schemaConcept -> tx.getShardCount(schemaConcept.asType()))
                 .sum();
     }
+
+    @Override
+    public int alphaEquivalentHash() {
+        Label[] labels = labels().stream().sorted().collect(Collectors.toList()).toArray(new Label[] {});
+        return Arrays.hashCode(labels);
```
kasper-piskorski: ditto

server/src/graql/gremlin/spanningtree/graph/Node.java
```diff
@@ -18,16 +18,20 @@
...
+import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * An node in a directed graph.
  *
  */
-public class Node {
+public class Node implements Comparable {
```
kasper-piskorski: Why do we need to implement Comparable? is overriding `hashCode` and `equals` not enough?

server/src/graql/gremlin/spanningtree/graph/Node.java
```diff
@@ -117,4 +124,42 @@ public int hashCode() {
...
+        internalFragments.addAll(fragmentsWithDependency);
+        internalFragments.addAll(fragmentsWithDependencyVisited);
+        internalFragments.addAll(dependants);
+
+        // sort into a list based on the fragments' hashes
+        Integer[] orderedInternalFragmentHashes = internalFragments.stream()
+                .map(Fragment::variableAgnosticHash)
+                .sorted()
+                .collect(Collectors.toList())
+                .toArray(new Integer[] {});
```
kasper-piskorski: We can use list hash.

server/src/graql/gremlin/TraversalPlanner.java
```diff
@@ -196,6 +200,49 @@ public static GraqlTraversal createTraversal(Pattern pattern, TransactionOLTP tx
...
+        for (Node node : oneHopHashes.keySet()) {
+            List<Integer> orderedNeighborNodeHashes = weightedGraph.getIncomingEdges(node)
+                    .stream()
+                    .map(weightedEdge -> weightedEdge.val.source)
+                    .map(oneHopHashes::get)
+                    .sorted()
+                    .collect(Collectors.toList());
+            orderedNeighborNodeHashes.add(0, oneHopHashes.get(node));
+            Integer[] oneHopOrderedNodeHashes = orderedNeighborNodeHashes.toArray(new Integer[] {});
+            int twoHopGlobalHash = Arrays.hashCode(oneHopOrderedNodeHashes);
```
kasper-piskorski: also here

server/src/graql/gremlin/fragment/Fragment.java
```diff
@@ -304,4 +305,15 @@ public final String toString() {
...
 
+    /**
+     * Compute a hash that is alpha-equivalent -
+     * this means that it doesn't depend on the vars contained in the
+     * fragments, but only information that cannot be random - labels, comparators, and
+     * any flags that may be set in the fragment
+     */
+    protected abstract int alphaEquivalentHash();
+
+    public int variableAgnosticHash() {
```
kasper-piskorski: why do we have two methods?

server/src/graql/gremlin/spanningtree/graph/Node.java
flyingsilverfin: You're right this isn't actually needed anymore

server/src/graql/gremlin/fragment/ValueFragment.java
```diff
@@ -129,4 +129,10 @@ public int hashCode() {
         h ^= this.operation.hashCode();
         return h;
     }
+
+    @Override
+    public int alphaEquivalentHash() {
+        // convert to string to ensure we compare the representation not the memory address
+        return operation.comparator().toString().hashCode();
```
kasper-piskorski: we can use the comparator hash here

flyingsilverfin:
@kasper-piskorski and I are likely NOT to merge this that due to statistics making some cases of nondeterministic query plans (soon) non-ambiguous, and that this use case is mostly restricted to symmetric query patterns, and that the overhead of the proper implementation of the full message passing algorithm consumes about 10-40% of the query planning time (can be 0.5-10 ms, which is quite high)

PR #5156, opened 2019-05-06
Cache distribution test results

### LINKED ISSUE #5079: Improve the speed of the CI on master branch
## Problem to Solve

The CI on `master` takes around 50 minutes to complete. For example, here's one instance: https://circleci.com/workflow-run/8330e7e5-40a0-4796-9b71-e7f2cf96dedf

It doesn't really impact us on a day-to-day basis, however it does when we're releasing.
Ideally, it should be somewhere around 10-15 minutes to maintain the smoothness of the release process.

### Caching Tests Results
@haikalpribadi has suggested caching the SHA of our distributions. This would improve the speed by never re-running test-assembly / test-deployment more than once for a given distribution.

The logic is as follows:
1. Before commencing any test-assembly / test-deployment, check if we have the test results in the cache for the given distribution.
2a. If we don't: commence the test and add the result to the cache
2b. If we do: skip the test and return the results from the cache

Needs graknlabs/grabl#20

vmax:

Fix #5079

Results of testing assembled distribution are now cached. Therefore, tests for same distribution archive are not triggered more than once, which speeds up overall CI execution for `master` branch.

For `test-assembly-mac-zip` and `test-assembly-linux-targz` CI jobs
- Cache on `sha256` of archive is queried
- Failure is stored on test failure
- Success is stored at the very last step of job

For tests that depend on distribution (targz/zip for linux/mac currently):
- build distribution
- ping @grabl: “is there a cached result for sha256 <…>“?
--> has stored ‘result: true’ — exit the CircleCI job with success status
--> has stored ‘result: false’ — exit the CircleCI job with failure status
--> has no stored results — execute test
- test results -> store success/failure for that `sha256`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,10 +131,12 @@ jobs:
       - checkout
       - run-bazel-rbe:
           command: bazel build //:assemble-mac-zip
+      - run: test/common/check_cache.py --file bazel-genfiles/grakn-core-all-mac.zip --get
       - run: unzip bazel-genfiles/grakn-core-all-mac.zip -d bazel-genfiles/
       - run: nohup bazel-genfiles/grakn-core-all-mac/grakn server start
-      - run: bazel test //test/common:grakn-application-test --test_output=streamed --spawn_strategy=standalone --cache_test_results=no
+      - run: bazel test //test/common:grakn-application-test --test_output=streamed --spawn_strategy=standalone --cache_test_results=no || test/common/check_cache.py --file bazel-genfiles/grakn-core-all-mac.zip --save-failure
       - run: bazel-genfiles/grakn-core-all-mac/grakn server stop
+      - run: test/common/check_cache.py --file bazel-genfiles/grakn-core-all-mac.zip --save-success
 
   test-assembly-windows-zip:
     machine: true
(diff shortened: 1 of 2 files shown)


vmax:
@haikalpribadi 

>1. Can we have /test/common/check_cache.py to be part of graknlabs_build_tools, @vmax ?

We can. How about we review it here and then submit the version we're happy with to `@graknlabs_build_tools`?

>2. How does the CI job know to "skip" running the test if it's already cached? I can't seem to identify this from looking at the code.

It does so by executing `circleci step halt`, see [these lines](https://github.com/vmax/grakn/blob/cache-test-distribution-results/test/common/check_cache.py#L77-L83). It's not well-documented, but it's a way to mark whole job as successful finishing at current step, [ref](https://discuss.circleci.com/t/ability-to-return-successfully-from-a-job-before-completing-all-the-next-steps/12969)

lolski:
This logic won't allow us to test flaky tests because when you want to run a test the second time it's already cached. The logic should be rewritten in this way to allow re-running failing test:

```
- build distribution
- ping @grabl: "is there a cached result for sha256 <…>"?
--> has stored 'result: true' — exit the CircleCI job with success status
--> otherwise — execute test and cache `sha256` **only** if test succeeded
```

vmax:
Another PR will be submitted to [graknlabs/build-tools](https://github.com/graknlabs/build-tools)

PR #5165, opened 2019-05-08
DFS Query Planning replacing Greedy Traversal

flyingsilverfin:
Replacing the Greedy Tree -> Plan conversion with DFS Tree -> plan conversion, should generally lead to the same speed or faster query plans. 

It should be much faster when there is a query that is branching, and the further down the branches you go the fewer matching results the query has. For instance, if we have `x rel y, y rel z, x rel a`, and there are 5000 of the `x rel y` and `x rel a`, doing those two steps first would lead to needing to check `5000 * 5000` pairs of edges. However, if `y rel z` reduces the set of 5000 to 1, doing the two steps `x rel y, y rel z`, then `x rel a` leads to 1*5000 pairs of edges. In general, we don't pay for going depth first because we will have to come back and deal with the rest of the deeper query portions later anyway, so we may as well do it first in the hope it reduces the result set quickly. 

Another way of thinking about this is that sibling branches in the query plan graph cannot reduce the set of results each other branch creates (imagine the `y` branch above results in 5k and the `a` branch results in 5k - one cannot reduce the size of the other 5k**). So, there's no point flipping between traversing one branch before the other - we just do each depth first and hope they lead to small result sets to create n-tuple results out of.

** The two exceptions where this doesn't hold are:
1. Short circuiting - in the case the query doesn't return any results, we can aim to compute the most likely to return 0 results branches first, rather than doing many computations to realise there are 0 results. In fact, the only way to return 0 results is that at least one path from the root node returns 0 results. We may want to consider how to order the branches in order of likelihood to return 0 results. A good start may be to do the cheapest branches first so we can process as many as possible to hit a 0 result first? On the other hand a more expensive branch my process more QP nodes, one of which my result in 0 matching results.
2. Comparisons. We still need to consider how doing a comparison on the Janus level  (== or !=, or even in the future >, < etc.) may be optimised, possibly including statistics in the DFS query planning as well. 

Before this change, we tried to estimate some cost of every possible next step (this is what made the linearisation of the query plan tree greedy) and picked the next cheapest step, which sometimes results in a BFS-type gremlin query, a DFS one, or a mix of the two.

* Scrap the Greedy Query Planning, replace it with a simpler DFS query planning step after the minimum spanning tree computation.


server/src/graql/gremlin/DepthFirstTreeTraversal.java
```diff
@@ -0,0 +1,103 @@
...
+        arborescence.getParents().forEach((child, parent) -> {
+            if (!edgesParentToChild.containsKey(parent)) {
+                edgesParentToChild.put(parent, new HashSet<>());
+            }
+            edgesParentToChild.get(parent).add(child);
+        });
+
+        Node root = arborescence.getRoot();
+
+        List<Node> reachableNodes = new ArrayList<>();
```
kasper-piskorski: We could use a 'Stack' collection here and pop instead of doing `reachableNodes.remove(reachableNodes.size()-1);` to make it more readable.

flyingsilverfin:
Turns out my understanding that went into this change was flawed. One branch CAN affect another by filtering (reducing) or multiplying (increasing) the size of the answer set that is multiplied with by future nodes. So, closing this PR until we re-think a better approach

PR #5172, opened 2019-05-09
Cache distribution test results

### LINKED ISSUE #5079: Improve the speed of the CI on master branch

vmax:

Fix #5079

Results of testing assembled distribution are now cached. Therefore, once successfully passed, tests for same distribution archive are not triggered more than once, which speeds up overall CI execution for `master` branch.

For `test-assembly-mac-zip` and `test-assembly-linux-targz` CI jobs
- Cache on `sha256` of archive is queried; if successful execution is found, tests are skipped
- Success is stored at the very last step of job

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,10 +131,12 @@ jobs:
       - checkout
       - run-bazel-rbe:
           command: bazel build //:assemble-mac-zip
+      - run: RELEASE_APPROVAL_TOKEN=$REPO_GITHUB_TOKEN bazel run @graknlabs_build_tools//ci:test-cache -- --file $(pwd)/bazel-genfiles/grakn-core-all-mac.zip --get
       - run: unzip bazel-genfiles/grakn-core-all-mac.zip -d bazel-genfiles/
       - run: nohup bazel-genfiles/grakn-core-all-mac/grakn server start
       - run: bazel test //test/common:grakn-application-test --test_output=streamed --spawn_strategy=standalone --cache_test_results=no
       - run: bazel-genfiles/grakn-core-all-mac/grakn server stop
+      - run: RELEASE_APPROVAL_TOKEN=$REPO_GITHUB_TOKEN bazel run @graknlabs_build_tools//ci:test-cache -- --file $(pwd)/bazel-genfiles/grakn-core-all-mac.zip --save-success
 
   test-assembly-windows-zip:
     machine: true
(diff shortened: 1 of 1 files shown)


haikalpribadi:
I know I asked this before (in #5156), @vmax, but by the time I saw your answer the PR was closed and the link to the answer no longer worked.

How does the CI job know to "skip" running the test if it's already cached? I can't seem to identify this from looking at the code.

vmax:
@haikalpribadi copying over the old answer 🙂:

>2. How does the CI job know to "skip" running the test if it's already cached? I can't seem to identify this from looking at the code.

It does so by executing `circleci step halt`, see [these lines](https://github.com/graknlabs/build-tools/blob/master/ci/test-cache.py#L85-L89). It's not well-documented, but it's a way to mark whole job as successful finishing at current step, [ref](https://discuss.circleci.com/t/ability-to-return-successfully-from-a-job-before-completing-all-the-next-steps/12969)

haikalpribadi:
Thanks, @vmax !

vmax:
@lolski 
- `$RELEASE_APPROVAL_TOKEN` is specified for HMAC-signing the request that goes to Grabl, must match the value stored in Heroku config of Grabl
- `--` is needed to separate command-line args of `bazel` and of the target that we're running (`@graknlabs_build_tools//ci:test-cache`)
- `--file` argument specifies the file path to compute checksum of. Note that `$(pwd)` is important because `.` will be relative to sandbox
- `--get` (as opposed to `--save-success`) queries stored info in Grabl and either executes:
* `exit 0` — proceed with test execution
* `circleci step halt` — mark the job as completed

PR #5189, opened 2019-05-14
Fix issue with unattaching attributes from relations

### LINKED ISSUE #4670: Removing an attribute from a Relationship Instance fails with ReifiedRelationship error
As outlined in the community post https://discuss.grakn.ai/t/error-running-unhas-relationshipreified-cannot-be-cast-to-relationship/1019

Basically the problem boils down to being unable to cast `ai.grakn.kb.internal.concept.RelationshipReified` to `ai.grakn.concept.Relationship` and seems like a proper bug, as this isn't a problem when trying to `unhas` on an entity instance.

kasper-piskorski:

Fix problem with unattaching attributes from relations - due to casting exception. The exception was thrown due to a implicit casting happening when executing a lambda - lambda tries to silently cast the result to return.

- Avoid calling the return.
- change 'reified()` return value from Optional to a Nullable 
- Closes #4670.

### DIFF
--- a/server/src/server/kb/concept/RelationImpl.java
+++ b/server/src/server/kb/concept/RelationImpl.java
@@ -31,6 +31,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Stream;
+import javax.annotation.Nullable;
 
 /**
  * Encapsulates relations between Thing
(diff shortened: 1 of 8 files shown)


server/src/server/kb/concept/RelationImpl.java
```diff
@@ -63,13 +61,13 @@ public static RelationImpl from(Relation relation) {
      *
      * @return The RelationReified if the Relation has been reified
      */
-    public Optional<RelationReified> reified() {
-        if (!relationStructure.isReified()) return Optional.empty();
-        return Optional.of(relationStructure.reify());
+    public RelationReified reified() {
```
marco-scoppetta: nit-pick: we should probably add `@Nullable` for correctness

PR #5198, opened 2019-05-16
[DO NOT MERGE YET] Statistics 1

flyingsilverfin:
Implement persisted statistics that count 0th order statistics - counts of each type. We now are able to query for the count of committed concepts of each type via the `KeyspaceStatistics` object, that is shared across Sessions. Accessing the counts requires a transaction, as the `KeyspaceStatistics` operate as a cache that on miss query from Janus. `AttributeDeduplicator` also propagates deletions into the overal statistics count.

We also do a first step to using statistics in query planning - eliminating all but the top 3 starting points in a query, if more than 3 "good" starting points are available. This is the first minor step towards a better query planner - we hope some queries now run faster.

* `KeyspaceStatistics` - read and write (persistently) statistics for a specific keyspace
* `UncomittedStatisticsDelta` - lives within a Transaction and on commit is merged into `KeyspaceStatistics`
* Tests for both of these ensuring each type of concept (including implicit attr relations) are counted correctly, and persisted across openings/closing of non-overlapping sessions to a keyspace. Includes a concurrent transaction test to ensure `AttributeDeduplicator` negative counts are included
* `Fragment.estimatedCostAsStartingPoint` -- provides a heuristic cost - based on number of owners or instances of that type, for the cost of the first step from that fixed cost starting node. Throws an exception if the fragment is not a fixed cost fragment
* Propagate attribute labels through the commitlog to AttributeDeduplicator to avoid having to do extra lookups

## Shortcomings
* No awareness of distributed system possibility - caching is done locally. This would need to be addressed to read accurate instance counts in a distributed environment
* Unsure of count behavior in race conditions when two transactions manipulate the same schema concept - for instance one deleting a schema concept and the other deleting and recreating the same name as a different type. For now, we leave the counts in the counts cache forever until the sessions are all closed and the cache is rebuilt
* The `estimateInitialCost` method portion that computes cost of `ValueFragment` is expensive (retrieves all subs of `attribute` and `@has-attribute`) and probably not the best heuristic for the cost -- the first can be solved by storing base type of each concept type as well, providing quick aggregation of the number of entities/attrs/rels without traversing all the subs()


server/src/graql/gremlin/TraversalPlanner.java
```diff
@@ -228,6 +240,68 @@ public static GraqlTraversal createTraversal(Pattern pattern, TransactionOLTP tx
...
 
+    /**
+     * Estimate the "cost" of a starting point for each type of fixed cost fragment
+     * These are heuristic proxies using statistics
+     * @param fixedCostFragment
+     * @return
+     */
+    private static long estimateInitialCost(Fragment fixedCostFragment, TransactionOLTP tx) {
+        KeyspaceStatistics statistics = tx.session().keyspaceStatistics();
+        if (fixedCostFragment instanceof  LabelFragment) {
```
kasper-piskorski: Maybe let's put all this instance specific behaviours into respective classes. I know we want to do the big refactor in the end, but it should be still beneficial to avoid cluttering things now if we can.

server/src/graql/gremlin/TraversalPlanner.java
```diff
@@ -207,15 +217,17 @@ public static GraqlTraversal createTraversal(Pattern pattern, TransactionOLTP tx
...
-    private static Set<Node> chooseStartingNodeSet(Set<Fragment> fragmentSet, Map<NodeId, Node> allNodes, SparseWeightedGraph sparseWeightedGraph) {
+    private static Set<Node> chooseStartingNodeSet(Set<Fragment> fragmentSet, Map<NodeId, Node> allNodes, SparseWeightedGraph sparseWeightedGraph, TransactionOLTP tx) {
         final Set<Node> highPriorityStartingNodeSet = new HashSet<>();
 
-        fragmentSet.forEach(fragment -> {
-            if (fragment.hasFixedFragmentCost()) {
+        fragmentSet.stream()
+            .filter(Fragment::hasFixedFragmentCost)
+            .sorted(Comparator.comparing(fragment -> estimateInitialCost(fragment, tx)))
+            .limit(3)
```
kasper-piskorski: maybe let's record it as a variable with a meaningful name instead of a magic number. Will help in the future.

server/src/server/statistics/KeyspaceStatistics.java
```diff
@@ -0,0 +1,84 @@
...
+            });
+        }
+
+        persist(tx);
+    }
+
+    private void persist(TransactionOLTP tx) {
+        for (String label : instanceCountsCache.keySet()) {
+            Long count = instanceCountsCache.get(label);
+            Concept schemaConcept = tx.getSchemaConcept(Label.of(label));
```
kasper-piskorski: Maybe do a null check here? What if we remove a type that has a key label in the cache?

server/src/server/statistics/KeyspaceStatistics.java
```diff
@@ -0,0 +1,84 @@
...
+            janusVertex.property(Schema.VertexProperty.INSTANCE_COUNT.name(), count);
+        }
+    }
+
+    private long retrieveCountFromVertex(TransactionOLTP tx, String label) {
+        Concept schemaConcept = tx.getSchemaConcept(Label.of(label));
+        Vertex janusVertex = ConceptVertex.from(schemaConcept).vertex().element();
+        VertexProperty<Object> property = janusVertex.property(Schema.VertexProperty.INSTANCE_COUNT.name());
+        // VertexProperty is similar to a Java Optional
+        return (Long) property.orElse(0L);
```
kasper-piskorski: Might be more beneficial to throw if we don't find the property. Should make debugging easier and won't silently allow incorrect behaviours.

server/src/graql/gremlin/TraversalPlanner.java
flyingsilverfin: changed to MAX_STARTING_POINTS

server/src/server/statistics/KeyspaceStatistics.java
flyingsilverfin: Type vertices dont' have this property until a non-zero count is added - if no property exists we default to 0, so this is correct behavior rather than throwing an exception.

server/src/graql/gremlin/TraversalPlanner.java
flyingsilverfin: great plans

server/src/server/deduplicator/AttributeDeduplicator.java
```diff
@@ -47,13 +47,13 @@
      * finally deleting that other duplicates.
      *
      * @param sessionFactory    the factory object for accessing the database
-     * @param keyspaceIndexPair the pair containing information about the attribute keyspace and index
+     * @param keyspaceAttributeTriple the pair containing information about the attribute keyspace and index
```
marco-scoppetta: let's try to keep the comments up to date when possible

server/src/server/deduplicator/AttributeDeduplicatorDaemon.java
```diff
@@ -80,16 +80,16 @@ public AttributeDeduplicatorDaemon(SessionFactory sessionFactory) {
...
-        Attribute attribute = Attribute.create(keyspace, index, conceptId);
+    public void markForDeduplication(KeyspaceImpl keyspace, String label, String index, ConceptId conceptId) {
+        Attribute attribute = Attribute.create(keyspace, label, index, conceptId);
         LOG.trace("insert({})",  attribute);
         queue.insert(attribute);
     }
 
     /**
      * Starts a daemon which performs deduplication on incoming attributes in real-time.
      * The thread listens to the RocksDbQueue queue for incoming attributes and applies
```
marco-scoppetta: I know this is not your change, but can you rename the rocksdbqueue to just queue - I forgot to change this

server/src/server/statistics/KeyspaceStatistics.java
```diff
@@ -0,0 +1,105 @@
...
+import grakn.core.server.kb.concept.ConceptVertex;
+import grakn.core.server.session.TransactionOLTP;
+import org.apache.tinkerpop.gremlin.structure.Vertex;
+import org.apache.tinkerpop.gremlin.structure.VertexProperty;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class KeyspaceStatistics {
```
marco-scoppetta: probably worth writing a high level overview of this class?

server/src/server/statistics/UncomittedStatisticsDelta.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package grakn.core.server.statistics;
+
+import java.util.HashMap;
+
+public class UncomittedStatisticsDelta {
```
marco-scoppetta: quick explanation of the class

test-integration/server/statistics/KeyspaceStatisticsIT.java
```diff
@@ -0,0 +1,295 @@
...
+        tx = localSession.transaction().write();
+        long personCount = localSession.keyspaceStatistics().count(tx, "person");
+        long ageCount = localSession.keyspaceStatistics().count(tx, "age");
+        long friendshipCount = localSession.keyspaceStatistics().count(tx, "friendship");
+        long implicitAgeCount = localSession.keyspaceStatistics().count(tx, "@has-age");
+        tx.close();
+
+        localSession.close();
+
+        // at this point, the graph and keyspace should be deleted from Grakn server
```
marco-scoppetta: just by closing the session graph and keyspace are deleted?

flyingsilverfin: yep!

flyingsilverfin: What happens is that there's an `onSessionClose` callback that deletes the Keyspace, graph, statistics etc. when the last session closes

marco-scoppetta: ok so we are referring to caches, not to graph/keyspace itself, ok

PR #5203, opened 2019-05-20
Statistics 1.1 - improve ValueFragment starting point cost estimate efficiency

flyingsilverfin:
Improves the efficiency of evaluating the starting point of a ValueFragment by about 50%. Previously, we were retrieving the concept for `@has-attribute` and all of its subs, then querying statistics using these labels. On top of that, we were also retrieving the entire `attribute` subs() hierarchy. Only one of these is required (`attribute` subs() are more likely to be cached) and we can convert these labels into the corresponding implicit labels, which then goes to the statistics lookup.

* Add test logback files to integration tests 
* Change `ValueFragment` to not retrieve both hierarchies of `attribute`.subs() and `@has-attribute`.subs()
* remove temporary test that slipped through

### DIFF
--- a/server/src/graql/gremlin/fragment/ValueFragment.java
+++ b/server/src/graql/gremlin/fragment/ValueFragment.java
@@ -140,7 +140,6 @@ public class ValueFragment extends Fragment {
 
     @Override
     public double estimatedCostAsStartingPoint(TransactionOLTP tx) {
-        long startTime = System.currentTimeMillis();
         KeyspaceStatistics statistics = tx.session().keyspaceStatistics();
 
         // compute the sum of all @has-attribute implicit relations
@@ -163,7 +162,6 @@ public class ValueFragment extends Fragment {
             totalImplicitRels += statistics.count(tx, implicitAttrRelLabel.toString());
         }
 
-        System.out.println("Time for valueFragment: " + (System.currentTimeMillis() - startTime) + " ms");
         if (totalAttributes == 0) {
             // short circuiting can be done quickly if starting here
             return 0.0;
--- a/server/src/graql/gremlin/fragment/ValueFragment.java
+++ b/server/src/graql/gremlin/fragment/ValueFragment.java
@@ -36,6 +36,7 @@ import org.apache.tinkerpop.gremlin.structure.Vertex;
 import javax.annotation.Nullable;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Stream;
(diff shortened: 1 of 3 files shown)


server/src/graql/gremlin/fragment/ValueFragment.java
```diff
@@ -139,6 +140,7 @@ public int hashCode() {
 
     @Override
     public double estimatedCostAsStartingPoint(TransactionOLTP tx) {
+        long startTime = System.currentTimeMillis();
```
kasper-piskorski: :)

server/src/graql/gremlin/fragment/ValueFragment.java
```diff
@@ -147,19 +149,21 @@ public double estimatedCostAsStartingPoint(TransactionOLTP tx) {
...
+        for (Iterator<AttributeType> it = attributeSubs.iterator(); it.hasNext(); ) {
+            AttributeType attrType = it.next();
+            Label attrLabel = attrType.label();
+            Label implicitAttrRelLabel = Schema.ImplicitType.HAS.getLabel(attrLabel);
+            totalAttributes += statistics.count(tx, attrLabel.toString());
+            totalImplicitRels += statistics.count(tx, implicitAttrRelLabel.toString());
         }
 
-        double totalAttributes = attributeSubs.map(t -> statistics.count(tx, t.label().toString())).reduce((a,b) -> a+b).orElse(1L);
+        System.out.println("Time for valueFragment: " + (System.currentTimeMillis() - startTime) + " ms");
```
kasper-piskorski: If we want to log it then let's use a logger. Otherwise let's remove it.

server/src/graql/gremlin/fragment/ValueFragment.java
```diff
@@ -147,19 +149,21 @@ public double estimatedCostAsStartingPoint(TransactionOLTP tx) {
...
-        Label implicitAttributeRelation = Schema.ImplicitType.HAS.getLabel(attributeLabel);
-        SchemaConcept implicitAttributeRelationType = tx.getSchemaConcept(implicitAttributeRelation);
-        double totalImplicitRels = 0.0;
-        if (implicitAttributeRelationType != null) {
-            Stream<RelationType> implicitSubs = implicitAttributeRelationType.asRelationType().subs();
-            totalImplicitRels = implicitSubs.map(t -> statistics.count(tx, t.label().toString())).reduce((a,b) -> a+b).orElse(1L);
+        for (Iterator<AttributeType> it = attributeSubs.iterator(); it.hasNext(); ) {
+            AttributeType attrType = it.next();
+            Label attrLabel = attrType.label();
+            Label implicitAttrRelLabel = Schema.ImplicitType.HAS.getLabel(attrLabel);
```
kasper-piskorski: I think we might want to look at the `KEY` variant as well for implicit relations.

kasper-piskorski: actually, since it is an estimate, it might be ok

PR #5206, opened 2019-05-21
Add logback file for console distribution

marco-scoppetta:

Add logback file to set to `OFF` logging level when using console alone without the full distribution.

As of now when packaging `grakn-core-console.zip` the package contains unused empty folders and a very verbose logback file (comes from server) which is also misplaced and not picked up by console JVM, so we see all the DEBUG statements in the console.

- remove unused folders when packaging just console
- add `logback.xml` inside `console/conf` to be packaged with console when `bazel build //console:assemble-*`

### DIFF
--- a/console/BUILD
+++ b/console/BUILD
@@ -76,7 +76,7 @@ assemble_targz(
     output_filename = "grakn-core-console-linux",
     targets = [":console-deps", "//bin:assemble-bash-targz"],
     additional_files = {
-           "//console:conf/logback.xml": "conf/logback.xml"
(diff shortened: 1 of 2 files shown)


lolski:
@marco-scoppetta makes sense to me. how would this change affect the error reporting behavior? ie., when there's an exception in the console, will they be printed to the screen and not swallowed silently?

marco-scoppetta:
@lolski yes exceptions are always reported because in the code we redirect everything to `stdout` and `sterr`, so logging is really not doing anything in Console.

PR #5210, opened 2019-05-22
QP Nodes - label propagation and count estimates

flyingsilverfin:
Prepare for rewriting the Tree to Query plan stage of the query planning process, which is currently implemented in `GreedyTreeTraversal` and it currently uses mostly hardcoded weights to choose the best query ordering. 

With this PR, we have the basic infrastructure to estimate how many vertices in the stored graph will be matched against a query planner node. Each query planner `Node` is one of 4 sub classes corresponding to 4 possible node specifications: a schema concept, a vertex matched by ID, a janus edge (vs a vertex), or an instance node. 
In the case of an instance (`InstanceNode`), we do a basic estimation of the label applicable to the node by propagating any `LabelFragments` connected over an `ISA` edge to any instance nodes -- this can be made cleverer in the future.

The method `Node.matchingElementsEstimate` implemented in subclasses utilizes the new information - `NodeType` and label (if available) to calculate about how many elements this node will match as an upper bound.

* `Node` is now abstract which is implemented by `InstanceNode`, `IdNode`, `EdgeNode`, and `SchemaNode` which indicates if it's a schema node, node with ID, instance node, or edge node. This is type is provided by each applicable Fragment, so there are corresponding changes to construct the right class.
* `InstanceNode` also has a field for instance label, if one can be deduced. `null` indicates unknown, so leads to assuming any instance may match
* `Node` has a method `matchingElementsEstimate` that returns 1 if it's an ID node; 1 if it's a Schema node. If it's an `InstanceNode` we compute `instanceCount(label)`, which defaults to some large number if we have no label (will be implemented properly in a follow up PR); `EdgeNode` returns 1 as well. This is because for now we don't have second order statistics (eg. `x` edges of this type per `y` of the starting vertex), if we want to go this way at all. It must be 1 instead of 0 because we will be multiplying the costs of vertices together, and 0 would lead to all 0's
* `NodeUtils.propagateLabels` is called right before the Tree traversal produces a plan to propagate labels across ISA nodes if possible


server/src/graql/gremlin/NodesUtil.java
```diff
@@ -130,4 +144,58 @@ private static void buildDependenciesBetweenNodes(Set<Fragment> allFragments, Ma
...
+                    .map(Map.Entry::getKey)
+                    .findFirst()
+                    .orElseThrow(() ->  GraqlSemanticException.create("QueryPlanner node: " + node.toString() + " has no parent"));
+
+            Set<Fragment> parentFragments = parent.getFragmentsWithoutDependency();
+            LabelFragment parentLabelFragment = (LabelFragment)parentFragments.stream().filter(f -> f instanceof LabelFragment).findFirst().orElse(null);
+            if (parentLabelFragment != null) {
+                // propagate the label to the children
+                Label label = parentLabelFragment.labels().iterator().next();
+                children.stream().filter(childNode -> childNode instanceof InstanceNode).forEach(child -> ((InstanceNode) child).setInstanceLabel(label));
```
kasper-piskorski: just FYI, to filter for type you can also do `.filter(InstanceNode.class::isInstance)`, and to cast '.map(InstanceNode.class::cast)`. Looks a bit neater in my mind.

flyingsilverfin: Just did something like this with marco! Explicitly casting now rather than ::cast however

PR #5212, opened 2019-05-23
Statistics - track total Thing count

flyingsilverfin:
Maintain a count of the total number of Things committed to the graph (this includes implicit relations). This should be equal to the number of concepts counted by an aggregate count in a new transaction. This means that retrieving the total number of things persisted no longer requires traversing all the `subs()` of `Thing` and adding the counts of each of these, as this can be an expensive operation if the schema is large.

* On commit, we sum the total of the concrete instance types changed and add them to the label `Thing`, which is persisted into the meta `Thing` concept just like any other schema concept would be handled.
* Tests updated to check the count of `Thing`
* Replace hardcoded upper bound in `InstanceNode` to use the `Thing` count from statistics (it was too expensive to do all the `subs()` before --  see above)

### DIFF
--- a/server/src/server/statistics/KeyspaceStatistics.java
+++ b/server/src/server/statistics/KeyspaceStatistics.java
@@ -39,6 +39,10 @@ import java.util.concurrent.ConcurrentHashMap;
  *
  * On cache miss, we read from JanusGraph schema vertices, which only have the INSTANCE_COUNT property if the
  * count is non-zero or has been non-zero in the past. No such property means instance count is 0.
+ *
+ * We also store the total count of all concepts the same was as any other schema concept, but on the `Thing` meta
+ * concept. Note that this is different from the other instance counts as it DOES include counts of all subtypes. The
+ * other counts on user-defined schema concepts are for for that concrete type only
  */
 public class KeyspaceStatistics {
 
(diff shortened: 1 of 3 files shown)


server/src/server/statistics/KeyspaceStatistics.java
```diff
@@ -57,6 +61,15 @@ public long count(TransactionOLTP tx, Label label) {
...
         HashMap<Label, Long> deltaMap = statisticsDelta.instanceDeltas();
 
+        // precompute the delta for all Thing's and update the delta map
+        long thingDelta = deltaMap.values().stream()
+                .reduce((a,b) -> a+b)
+                .orElse(0L);
+        Label thingLabel = Schema.MetaSchema.THING.getLabel();
+        // it shouldn't exist but just in case we insert it and do an update
+        deltaMap.putIfAbsent(thingLabel, 0L);
+        deltaMap.put(thingLabel, deltaMap.get(thingLabel) + thingDelta);
```
kasper-piskorski: Why double put? Also, put does an overwrite if an entry already exists.

flyingsilverfin: The second put does a get first :) so we do an update in the second, line, and initialise a value on the first line

PR #5211, opened 2019-05-22
Upgrade @graknlabs_build_tools and matching .bazelrc settings

vmax:
## What is the goal of this PR?

Upgrading the CI configuration in pull request graknlabs/grakn#5208 broke the CI due to a bug in `bazel` (only local builds on Linux are affected). Now `bazel` is upgraded to `0.25.2` through the means of upgrading `@graknlabs_build_tools` which fixes the CI build status.

RPMs of Grakn Core are now built remotely:
* Maintainability is improved by having consistency across all targets: no targets are built locally on CI
* Performance is improved by caching the RPM target of Grakn Core

## What are the changes implemented in this PR?

Fixes #4797 by removing `--strategy_regexp=MakeRpm=local` option from `.bazelrc`

* `bazel` is upgraded to `0.25.2` to fix broken CI via upgrading `@graknlabs_build_tools`
* contents of `.bazelrc` are now matching latest version in `@graknlabs_build_tools`

PR #5215, opened 2019-05-24
 Fix `test-assembly-mac-zip` and `test-assembly-windows-zip` CI jobs

vmax:

- Fix `test-assembly-mac-zip` which was broken because of upgrading `bazel` to latest version in #5211

- Fix `test-assembly-windows-zip` which was broken by a recent introduction of custom java toolchain in #5208

This pull request addresses both of these problems which enables us to continue testing Grakn Core on Mac and Windows platforms.

* `test-assembly-mac-zip`: upgrading XCode to latest version and explicitly installing JDK8 (bazelbuild/bazel#7304): resolved in graknlabs/build-tools#54
* `test-assembly-windows-zip`'s failure is resolved in graknlabs/build-tools#53, upgraded link to latest `@graknlabs_build_tools` manually.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,7 @@ commands:
     steps:
       - run: curl -OL https://raw.githubusercontent.com/graknlabs/build-tools/master/ci/install-bazel-mac.sh
       - run: bash ./install-bazel-mac.sh && rm ./install-bazel-mac.sh
+      - run: brew cask install homebrew/cask-versions/adoptopenjdk8
 
   install-rpm:
     steps:
@@ -124,7 +125,7 @@ jobs:
 
   test-assembly-mac-zip:
     macos:
-      xcode: "9.0"
+      xcode: "10.2.1"
     working_directory: ~/grakn
     steps:
       - install-bazel-mac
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -29,6 +29,7 @@ commands:
     steps:
       - run: curl -OL https://raw.githubusercontent.com/graknlabs/build-tools/master/ci/install-bazel-mac.sh
       - run: bash ./install-bazel-mac.sh && rm ./install-bazel-mac.sh
+      - run: brew cask install homebrew/cask-versions/adoptopenjdk8
```
lolski: why do we need this @vmax ? if this is always needed for mac, we should bundle it into `./install-bazel-mac.sh`

vmax: @lolski mac builds will fail without JDK installed (JRE only), see https://github.com/bazelbuild/bazel/issues/7304. I'll add it to `install-bazel-mac` then.

vmax: @lolski addressed in https://github.com/graknlabs/grakn/pull/5215/commits/2e599edfd186b9bfcf4865d6a3e1bb880f4f9886

PR #5219, opened 2019-05-28
Persist inferred concepts if we explicitly ask to do so

kasper-piskorski:
Previously we defined the commit behaviour of inferred concepts for MATCH queries (#5142). Now we define a commit behaviour for INSERT queries:

Inferred concepts will be persisted provided they occur in an explicitly executed insert (or insert-match) query.

- on write we mark inferred concepts for persistence
- based on those concepts and all tracked inferred concepts, we only discard those not marked for persistence

### DIFF
--- a/server/src/graql/executor/QueryExecutor.java
+++ b/server/src/graql/executor/QueryExecutor.java
@@ -303,7 +303,8 @@ public class QueryExecutor {
 
             Stream<ConceptMap> answers = transaction.stream(match.get(projectedVars), infer);
             answerStream = answers
-                    .map(answer -> WriteExecutor.create(transaction, executors.build()).write(answer));
+                    .map(answer -> WriteExecutor.create(transaction, executors.build()).write(answer))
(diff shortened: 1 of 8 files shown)


server/src/graql/executor/QueryExecutor.java
```diff
@@ -296,15 +296,14 @@ public ConceptMap undefine(GraqlUndefine query) {
...
 
             LinkedHashSet<Variable> projectedVars = new LinkedHashSet<>(matchVars);
             projectedVars.retainAll(insertVars);
 
             Stream<ConceptMap> answers = transaction.stream(match.get(projectedVars), infer);
-            answerStream = answers.map(answer -> WriteExecutor
-                    .create(transaction, executors.build()).write(answer))
-                    .collect(toList()).stream();
+            answerStream = answers
+                    .map(answer -> WriteExecutor.create(transaction, executors.build()).write(answer));
```
marco-scoppetta: 😮

PR #5229, opened 2019-06-03
Minor cleanup in TransactionOLTP

marco-scoppetta:

Minor cleanup in TransactionOLTP in attempt to make logic easier to follow.

- simplify logic of `getConcept(ConeptId id)`
- refactor `operateOnOpenGraph` to be only a check instead of receiving a supplier function

### DIFF
--- a/server/src/server/session/TransactionOLTP.java
+++ b/server/src/server/session/TransactionOLTP.java
@@ -196,7 +196,7 @@ public class TransactionOLTP implements Transaction {
     // Register new committed Attributes in AttributesMap (if the map doesnt already contain the same INDEX),
     // so that all following Transactions will be able to use the same ConceptId
     // when in need of same attribute with same value.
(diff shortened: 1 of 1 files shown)


server/src/server/session/TransactionOLTP.java
```diff
@@ -657,37 +657,39 @@ public Rule putRule(Label label, Pattern when, Pattern then) {
...
-                //uses an its previous EdgeRelation as an id so property must be fetched
-                return concept == null?
-                        this.getConcept(Schema.VertexProperty.EDGE_RELATION_ID, id.getValue()) :
-                        concept;
-            }
-        });
+        checkGraphIsOpen();
+
+        // fetch concept from cache if it's cached
+        if (transactionCache.isConceptCached(id)) {
```
kasper-piskorski: probably it's negligible, but this does a cache GET twice

PR #5245, opened 2019-06-11
Refactor reiteration condition

kasper-piskorski:
Relax the reiteration condition. Currently we often reiterate the reasoning algorithm even if it's not needed. Relaxing the condition allows for performance gains for some queries.

- refactored the use of `resolve` function of reasoner queries - instead of checking whether reiteration is required at the beginning of processing, we do it after the first iteration when we have more information in the query cache.
- we make the reiteration check more detailed - not only we check for loops in the rule dependency graph but also in the corresponding graph of relation instances.
- the check is more expensive but allows to finish processing early if no loops are found
- if we do not have full information to access the presence of loops, we still reiterate


server/src/graql/reasoner/rule/RuleUtils.java
```diff
@@ -128,11 +133,41 @@
...
+                return answers.stream().anyMatch(ans ->
+                        varPairs.stream().anyMatch(p -> {
+                            Concept from = ans.get(p.getKey());
+                            Concept to = ans.get(p.getValue());
+                            if (conceptMap.get(to).contains(from)) return true;
+                            conceptMap.put(from, to);
+                            return false;
+                        }));
+            });
+        });
```
marco-scoppetta: maybe in the future the above method can be made a bit more readable? even though I understand it might be complex

PR #5246, opened 2019-06-12
Refactor attribute atom materialisation pt II

kasper-piskorski:
When materialising attributes we want to follow a similar procedure to materialising relations. This means we try to use graql for checking for answer existence. In that way we can cache the checks and reuse them later.

Employs a similar materialisation procedure for attributes as we currently have for relations - we employ the put behaviour by having a cache lookup before a potential insert.


server/src/graql/reasoner/atom/Atom.java
```diff
@@ -38,6 +39,8 @@
 import grakn.core.graql.reasoner.atom.predicate.VariablePredicate;
 import grakn.core.graql.reasoner.cache.SemanticDifference;
 import grakn.core.graql.reasoner.cache.VariableDefinition;
+import grakn.core.graql.reasoner.query.ReasonerAtomicQuery;
+import grakn.core.graql.reasoner.query.ReasonerQueries;
```
marco-scoppetta: are these import needed? I don't see any other code change in this class

kasper-piskorski: good catch! they are unused

PR #5253, opened 2019-06-14
Make 'grakn server clean' command to clean the correct data directory as specified in grakn.properties

### LINKED ISSUE #5248: 'grakn server clean' hard-codes the path to the db directory instead of reading from the configuration file
https://github.com/graknlabs/grakn/blob/master/daemon/executor/Storage.java#L66

vmax:

Fix #5248

Previously a hardcoded directory (`server/db/`) was used when running `grakn server clean`. This resulted in possible problems if custom directory was used for data storage. Now, `data-dir` config value is used to determine the correct directory to clean.

- Use `data-dir` from `grakn.properties` to determine what directory to clean instead of hardcoding it to `STORAGE_DATA` constant
- Remove `STORAGE_DATA` constant since it's no longer needed

### DIFF
--- a/daemon/executor/Storage.java
+++ b/daemon/executor/Storage.java
@@ -63,7 +63,6 @@ public class Storage {
     private static final String DISPLAY_NAME = "Storage";
     private static final long STORAGE_STARTUP_TIMEOUT_SECOND = 60;
     private static final Path STORAGE_PIDFILE = Paths.get(System.getProperty("java.io.tmpdir"), "grakn-storage.pid");
-    private static final Path STORAGE_DATA = Paths.get("server", "db", "cassandra");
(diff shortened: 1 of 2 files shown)


lolski:
We don't need to read grakn properties as we have done it and have the config object: https://github.com/graknlabs/grakn/blob/master/daemon/executor/Storage.java#L83

lolski:
@vmax can we add a test which tests this functionality in https://github.com/graknlabs/grakn/blob/master/test-end-to-end/distribution/GraknGraqlCommands_WithARunningGraknE2E.java ?

The test should verify that the clean command properly reads the path by doing this steps:
0. modify the path to the `db` folder in `grakn.properties`
1. `grakn server start`
2. insert some data
3. `grakn server stop`
4. `grakn server clean`
5. `grakn server start`
6. verify that the keyspace is empty

PR #5260, opened 2019-06-17
Farewell AttributeDeduplicator

marco-scoppetta:

Remove `AttributeDeduplicator` and instead serialise commits that contain new attributes and merge in real-time when needed.

Also set JanusGraph cache to `false` by default. The user will need to manually enable it and handle with care.

- Remove all references to AttributeDeduplicator
- set to `false` the use of JanusGraph cache, given that it might lead to OutOfMemory if not handled carefully
- re-introduced `graphLock` used to lock when committing new attributes
- added `removedAttributes` map in `TransactionCache` so that we can evict deleted indexes from `attributesMap`
-`commitInternal` method inside `TransactionOLTP` now handles 3 cases, see comments in code
- removed `commitWithLogs` and `commitAndGetLogs` methods
- removed `CommitLog` class, as there is nothing to deduplicate asynchronously anymore
- renamed tests and updated them, to check attribute uniqueness is guaranteed right after commit


marco-scoppetta:
@flyingsilverfin I think we should at some point yes, but the idea is that we are disabling the Janus cache by default because we haven't profile it enough yet to know what to suggest to people. So only the more daring users will have to enabled it manually and handle it with care. Once we'll learn more about best practices on how to use it we should write a piece of documentation

flyingsilverfin:
@marco-scoppetta Sure, but we should probably start a document or something now or we will forget stuff like this :)

PR #5266, opened 2019-06-18
Core Bugfixes

flyingsilverfin:
Fix and add tests for:
* querying for nonreified implicit attribute ownership and the owner in the same query not returning any concepts fixed by moving an edge traversal step into `IdFragment`
* remove check for deleting inferred concepts (undoing #5054, the added tests in that PR still passes)
* return deleted concepts' IDs without re-querying deleted Janus element (was causing vertex already deleted bugs with reified relations)

* Bugfixes for above small issues and some relevant tests

### DIFF
--- a/server/src/graql/executor/QueryExecutor.java
+++ b/server/src/graql/executor/QueryExecutor.java
@@ -22,6 +22,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
 import grakn.benchmark.lib.instrumentation.ServerTracing;
 import grakn.core.concept.Concept;
+import grakn.core.concept.ConceptId;
 import grakn.core.concept.answer.Answer;
 import grakn.core.concept.answer.AnswerGroup;
 import grakn.core.concept.answer.ConceptList;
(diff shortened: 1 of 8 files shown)


server/src/graql/executor/QueryExecutor.java
```diff
@@ -378,10 +379,12 @@ public ConceptSet delete(GraqlDelete query) {
             } else {
                 throw GraknServerException.create("Unhandled concept type isn't a schema concept or a thing");
             }
-        });
+            // TODO rewrite to either throw or return an ID
+            return null; // we shouldn't be hitting this case without throwing an exception
+        }).filter(Objects::nonNull).collect(Collectors.toSet());
```
flyingsilverfin: @marco-scoppetta @kasper-piskorski maybe you guys have an idea how to reorganise this loop to avoid the `nonNull`

marco-scoppetta: if you want to avoid the `null` thing just keep the `.forEach` instead of the `stream().map()` and instead of `return id` you can `set.add(id)`

flyingsilverfin: better!

PR #5274, opened 2019-06-20
Fix creation and deletion of attributes bugs

marco-scoppetta:

Resolve 2 bugs concerning creation and deletion of attributes in the same transaction.
Specifically:

- when deleting an attribute, the `newAttributes` map in `TransactionCahce` was not properly evicted
- when deleting and re-creating an attribute with same value in the same Transaction, `attributesMap` in `SessionImpl` was not updated in the correct order.

- deleting attributes from `newAttributes` map now correctly creates a `Pair` containing `Label, String` instead of `String, String`
- when committing, removed deleted attributes from `attributesMap` before checking if merging is needed
- add new regression tests and update test names

### DIFF
--- a/server/src/server/session/TransactionOLTP.java
+++ b/server/src/server/session/TransactionOLTP.java
@@ -221,6 +221,7 @@ public class TransactionOLTP implements Transaction {
     private void mergeAttributesAndCommit() {
         session.graphLock().writeLock().lock();
         try {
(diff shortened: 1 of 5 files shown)


marco-scoppetta:
@flyingsilverfin I am going to merge this, if you have comments I will address them in another PR

PR #5280, opened 2019-06-24
Define atom conversion behaviours explicitly

kasper-piskorski:
We add atom conversion tests which explicitly specify the expected behaviour of atom conversions.

- Added 'ConversionIT' test defining explicit atom conversion behaviours
- fixed exisiting inconsistencies in atom conversion to make it conform to the specified behaviours

### DIFF
--- a/test-integration/graql/reasoner/atomic/ConversionIT.java
+++ b/test-integration/graql/reasoner/atomic/ConversionIT.java
@@ -1,0 +1,176 @@
+/*
+ * GRAKN.AI - THE KNOWLEDGE GRAPH
+ * Copyright (C) 2018 Grakn Labs Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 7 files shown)


test-integration/graql/reasoner/atomic/ConversionIT.java
```diff
@@ -0,0 +1,176 @@
...
+    private static Conjunction<Statement> attributePattern;
+    private static Conjunction<Statement> relationPattern;
+    private static Conjunction<Statement> implicitRelationPattern;
+    private static Variable attributeVar;
+    private static Variable relationVar;
+
+    @BeforeClass
+    public static void loadContext(){
+        final String resourcePath = "test-integration/graql/reasoner/resources/";
+        session = server.sessionWithNewKeyspace();
```
marco-scoppetta: let's close this session in an AfterClass so we don't risk upsetting Grakn

PR #5285, opened 2019-06-25
Use transaction.close instead of transaction.rollback

marco-scoppetta:

Revert recent change and use `janusTransaction.close()` again, instead of `janusTransaction.rollback()`

- also remove the size large from the test as it actually completes in 60s, so 300s should be more than enough

### DIFF
--- a/server/src/server/session/TransactionOLTP.java
+++ b/server/src/server/session/TransactionOLTP.java
@@ -914,7 +914,7 @@ public class TransactionOLTP implements Transaction {
             return;
         }
         try {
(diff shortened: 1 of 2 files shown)


lolski:
LGTM, we may need to revisit using `rollback()` at some point because that's what Janusgraph recommends: https://docs.janusgraph.org/latest/tx.html#_common_transaction_handling_problems

```
Transactions must be terminated manually with commit() or rollback()
```

marco-scoppetta:
@lolski yes, you're right - let first see if this is the actual fix

PR #5293, opened 2019-06-27
Cleanup code in TransactionOLTP and ElementFactory

marco-scoppetta:

Cleanup code in TransactionOLTP and ElementFactory.
Centralise operations on `janusTransaction` (i.e. `addVertex()`, 'close()' and `commit()`) inside TransactionOLTP.

- moved creation of new Vertex from `ElementFactory` to `TransactionOLTP`
- other minor refactoring and cleanup

### DIFF
--- a/server/src/server/session/TransactionOLTP.java
+++ b/server/src/server/session/TransactionOLTP.java
@@ -97,12 +97,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
- * A TransactionOLTP using JanusGraph as a vendor backend.
- * <p>
- * Wraps a TinkerPop transaction (the graph is still needed as we use to retrieve tinker traversals)
- * <p>
- * With this vendor some issues to be aware of:
- * 2. Clearing the graph explicitly closes the connection as well.
+ * A TransactionOLTP that wraps a Tinkerpop OLTP transaction, using JanusGraph as a vendor backend.
  */
 public class TransactionOLTP implements Transaction {
     private final static Logger LOG = LoggerFactory.getLogger(TransactionOLTP.class);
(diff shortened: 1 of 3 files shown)


server/src/server/session/TransactionOLTP.java
```diff
@@ -292,37 +279,29 @@ private static void mergeAttributeEdge(Vertex mergeTargetV, Vertex ent, GraphTra
...
-        return executeLockingMethod(() -> factory().addVertexElementWithEdgeIdProperty(baseType, conceptId));
-    }
-
-    /**
-     * Executes a method which has the potential to throw a TemporaryLockingException or a PermanentLockingException.
-     * If the exception is thrown it is wrapped in a GraknServerException so that the transaction can be retried.
-     *
-     * @param method The locking method to execute
-     */
-    private <X> X executeLockingMethod(Supplier<X> method) {
```
marco-scoppetta: removed this method as it adds complexity to code without any particular added value, first of all because there is no re-try policy anymore anywhere and the exception class is currently lost we reported to the client via gRPC.

The retry policy can be implemented by catching a BackendException from Janus where needed in the future.
And we should also figure out a way to report the correct exception type back to the client.

server/src/server/session/TransactionOLTP.java
```diff
@@ -982,17 +955,17 @@ private void validateGraph() throws InvalidKBException {
...
      * @param conceptId the id of the concept to shard
      */
+    @VisibleForTesting
     public void shard(ConceptId conceptId) {
         ConceptImpl type = getConcept(conceptId);
         if (type == null) {
-            LOG.warn("Cannot shard concept [" + conceptId + "] due to it not existing in the graph");
-        } else {
-            type.createShard();
+            throw new RuntimeException("Cannot shard concept [" + conceptId + "] due to it not existing in the graph");
```
marco-scoppetta: given that this method is only used in tests, it makes more sense to throw an exception if something is wrong rather than `LOG.warn`

PR #5326, opened 2019-07-11
Change Cassandra driver from Thrift to CQL

marco-scoppetta:

Move away from Thrift driver which is now deprecated and adopt CQL driver which is currently the only supported driver by Datastax.
The distribution has now ~40MB less dependencies.

- removed `janusgraph-cassandra` dependency
- updated config keys to use `cql` instead of `cassandra`
- removed customised InputFormat classes (such as `GraknCassandra3InputFormat`) and use default provided by Janus: `CqlInputFormat`
- refactored `HadoopGraphFactory` to immediately instantiate an HadoopGraph (before we were lazily creating one for no reason)
- updated `GraknTestServer` to use nativePort (CQL), instead of  `rpc` (Thrift)
- updated dependencies (farewell Astyanax!)

We now depend on a new customised version of Janusgraph which has all the improvements needed to make opening/closing of a graph as fast as when using Thrift.


server/src/server/session/SessionImpl.java
```diff
@@ -76,7 +77,7 @@
      * @param config   config to be used.
      */
     public SessionImpl(KeyspaceImpl keyspace, Config config, KeyspaceCache keyspaceCache, StandardJanusGraph graph, KeyspaceStatistics keyspaceStatistics, Cache<String, ConceptId> attributesCache, ReadWriteLock graphLock) {
-        this(keyspace, config, keyspaceCache, graph, keyspaceStatistics, new HadoopGraphFactory(config, keyspace), attributesCache, graphLock);
+        this(keyspace, config, keyspaceCache, graph, null, keyspaceStatistics, attributesCache, graphLock);
```
lolski: Based on the discussion with @marco-scoppetta , the need of passing down `null` can be removed once the code is refactored, which will be done in a follow-up PR.

PR #5332, opened 2019-07-15
Remove the use of assert

kasper-piskorski:

To remove the use of `assert` in the code. The motivation is that it is only meant to be used during development and it requires a VM option to be enabled during run time. Additionally, if the assert condition is false, the assert throws an AssertionError with just the message text.

- Removes the use of `assert` and uses `Preconditions` in place.

### DIFF
--- a/server/src/graql/executor/WriteExecutor.java
+++ b/server/src/graql/executor/WriteExecutor.java
@@ -18,6 +18,7 @@
 
 package grakn.core.graql.executor;
 
+import com.google.common.base.Preconditions;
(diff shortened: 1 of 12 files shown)


kasper-piskorski:
The following exceptions are thrown:
checkArgument - IllegalArgumentException 
checkNotNull - NullPointerException
checkState - IllegalStateException

All those checks are very low level and mostly sanity check the algorithm logic - if any of them throws it highlights unforeseen execution path - a bug somewhere. As a result I don't think in this case we can provide a meaningful grakn error to the user.

PR #5349, opened 2019-07-30
QP stage 2 Replacement (not faster yet)

flyingsilverfin:
Save the new second stage of the query planner as a complete unit of work, even if performance needs work.

Includes the new statistically driven optimal tree-to-plan algorithm. It is optimal with regards to the defined simple cost function if it is allowed to run to completion, though we add a timeout. Worst case complexity grows exponentially, so we also prune and optimise the core algorithm as much as possible.

Cost function:
For each _query graph_ node, we have an estimated number of matching vertices that match in Janus (ignoring edge relations for now). We have to choose an order of visiting nodes in a tree with nodes `A,B,C, D, E...`
The cost of a given order `ABCDE` of these nodes is:
`cost = A + AB + ABC + ABCD + ABCDE`

This is the cost of finding and traversing the nodes of A, then pairs (A,B), then triples (A,B,C) etc, which mostly seems to match the way JanusGraph processes nodes. A good validation of this informed guess would help increase confidence in this approach in the future.

The algorithm determines the best ordering that produces the smallest possible cost. This implicitly weights the later choices more than the earlier ones because they costs multiply to very large numbers that dwarf the early terms. Another issue is that the numbers can grow extremely fast to approach the limits of the storage of a Double and leads to numerical accuracy issues when adding small numbers to extremely large ones (eg. `1000 + 4E234`)

Thus, alternative considered costs are using:
* the sum of logs `log(A) + log(B) + ...`, but this means any ordering has the same cost
* Linearly decreasing logs: `1 log(A) + 2 log(B) + 3 log(B) + ...` which seems like a good compromise of weighting later tuples generated more than early ones, which is what we want as these dominate the number of process items
* Alternative variations of these linearly weighted terms (possible even the inverse: `n log(A) + (n-1) log(B)...` if empirically shown to be better

Other design decisions:
* `EdgeNode` vertices are handled together with the other `VertexNode` attached to the edge node. `EdgeNode` cost 5 rather than 1, though this may not matter
* Timeout mechanism that can force stop as soon as the first solution is found, which we do greedy-first as this is a high-scoring starting plan
* Plenty of custom optimisations to increase performance so we can handle more nodes

* new class `OptimalTreeTraversal` that replaces `GreedyTreeTraversal` and simplifies the way nodes and fragments are processed, plus optimisations within in this class
* testing for `OptimalTreeTraversal` using mocks that can also be used for testing speed of the new planner
* Statistic counts for query planning nodes are now estimated from sets of labels, using inferred information on `RolePlayer` fragments among other places


haikalpribadi:
@flyingsilverfin this PR got closed when I deleted the branch `experimental-qp` (it was 9 months old, stale, and had no diverging commits to master). Do we still need work in your fork? Should we reopen the PR to `master`?

PR #5350, opened 2019-07-31
Minor cleanup of properties/settings

marco-scoppetta:

Cleanup use and location of properties and configurations used to instantiate new graphs.

- remove unused config keys (`cql.storage.port` and `storage.batch-loading`)
- rename `default-configs.properties` to `default-OLAP-configs.properties` to only store OLAP related configurations which are quite low level and should not be configurable by users
- cleanup `JanusGraphFactory` which was using and reading many configs that are not needed
- removed unused `rpcPort` from `GraknTestServer`

### DIFF
--- a/server/src/server/session/JanusGraphFactory.java
+++ b/server/src/server/session/JanusGraphFactory.java
@@ -18,10 +18,8 @@
 
 package grakn.core.server.session;
 
-import com.google.common.collect.ImmutableMap;
 import grakn.core.common.config.Config;
 import grakn.core.common.config.ConfigKey;
-import grakn.core.common.exception.ErrorMessage;
 import grakn.core.server.kb.Schema;
 import grakn.core.server.session.optimisation.JanusPreviousPropertyStepStrategy;
 import org.apache.tinkerpop.gremlin.process.traversal.Order;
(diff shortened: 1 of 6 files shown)


server/src/server/session/JanusGraphFactory.java
```diff
@@ -61,25 +55,9 @@
...
-        String DEFAULT_CONFIG = "resources/default-configs.properties";
-        DEFAULT_PROPERTIES = new Properties();
-        try (InputStream in = JanusGraphFactory.class.getClassLoader().getResourceAsStream(DEFAULT_CONFIG)) {
-            DEFAULT_PROPERTIES.load(in);
-        } catch (IOException e) {
-            throw new RuntimeException(ErrorMessage.INVALID_PATH_TO_CONFIG.getMessage(DEFAULT_CONFIG), e);
-        }
-    }
+    private static final String STORAGE_BACKEND = ConfigKey.STORAGE_BACKEND.name();
+    private static final String CQL_BACKEND = "cql";
```
marco-scoppetta: Moved `storage.backend` configuration back inside the code as it is not a configuration that can be actually changed at run-time, as everything would break

PR #5352, opened 2019-08-01
Simplify the architecture by removing the 'graknsystem' keyspace

lolski:

We have removed the `graknsystem` keyspace. It is not needed and the removal leads to a simpler architecture and faster load time.

- Remove the `graknsystem` keyspace. The keyspace stores a list of keyspaces that exist. This functionality is replaced by simply querying Cassandra using CQL
- We've been depending on JanusGraph's behaviour of maintaining main thread to run, by accident. This behaviour is lost as `KeyspaceManager` no longer opens a graph. Therefore we're adding similar functionality: `Server::awaitTermination`.
- Some small refactors and removal of dead-codes
- Modify visibility modifiers of various classes as needed by KGMS

### DIFF
--- a/server/src/server/Grakn.java
+++ b/server/src/server/Grakn.java
@@ -63,13 +63,12 @@ public class Grakn {
             Stopwatch timer = Stopwatch.createStarted();
             boolean benchmark = parseBenchmarkArg(args);
             Server server = ServerFactory.createServer(benchmark);
-            Runtime.getRuntime().addShutdownHook(new Thread(server::close));
             server.start();
             LOG.info("Grakn started in {}", timer.stop());
             try {
-                server.block();
+                server.awaitTermination();
             } catch (InterruptedException e) {
-                //
+                server.close();
             }
         } catch (RuntimeException | IOException e) {
             LOG.error(ErrorMessage.UNCAUGHT_EXCEPTION.getMessage(e.getMessage()), e);
--- a/server/src/server/Grakn.java
+++ b/server/src/server/Grakn.java
@@ -63,9 +63,14 @@ public class Grakn {
             Stopwatch timer = Stopwatch.createStarted();
             boolean benchmark = parseBenchmarkArg(args);
             Server server = ServerFactory.createServer(benchmark);
+            Runtime.getRuntime().addShutdownHook(new Thread(server::close));
             server.start();
-
             LOG.info("Grakn started in {}", timer.stop());
+            try {
+                server.block();
+            } catch (InterruptedException e) {
+                //
+            }
         } catch (RuntimeException | IOException e) {
             LOG.error(ErrorMessage.UNCAUGHT_EXCEPTION.getMessage(e.getMessage()), e);
             System.err.println(ErrorMessage.UNCAUGHT_EXCEPTION.getMessage(e.getMessage()));
--- a/server/src/server/Grakn.java
+++ b/server/src/server/Grakn.java
@@ -68,6 +68,7 @@ public class Grakn {
             try {
                 server.awaitTermination();
             } catch (InterruptedException e) {
+                // grakn server stop is called
                 server.close();
             }
         } catch (RuntimeException | IOException e) {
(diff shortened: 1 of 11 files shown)


server/src/server/Grakn.java
```diff
@@ -64,8 +64,13 @@ public static void main(String[] args) {
             boolean benchmark = parseBenchmarkArg(args);
             Server server = ServerFactory.createServer(benchmark);
             server.start();
-
             LOG.info("Grakn started in {}", timer.stop());
+            try {
+                server.awaitTermination();
+            } catch (InterruptedException e) {
+                // grakn server stop is called
+                server.close();
```
marco-scoppetta: I think we need to add `Thread.currentThread().interrupt()`, base on [SonarCloud rule](https://rules.sonarsource.com/java/tag/error-handling/RSPEC-2142)

server/src/server/Server.java
```diff
@@ -30,38 +30,30 @@
...
 
     public void start() throws IOException {
-        initialiseSystemSchema();
         serverRPC.start();
     }
 
+    // NOTE: this method is used by Grakn KGMS and should be kept public
+    public void awaitTermination() throws InterruptedException {
+        serverRPC.awaitTermination();
+    }
```
marco-scoppetta: serverRPC is created by KGMS, so KGMS should be able to call `serverRPC.awaitTermination()`, I am probably missing something here, it just looks like an odd method

server/src/server/keyspace/KeyspaceManager.java
```diff
@@ -18,149 +18,29 @@
...
-            thing.delete();
-            attribute.delete();
-
-            existingKeyspaces.remove(keyspace);
-            tx.commit();
-        }
+    private final Cluster storage;
+    private final Set<String> internals = new HashSet<>(Arrays.asList("system_traces", "system", "system_distributed", "system_schema", "system_auth"));
+    
+    public KeyspaceManager(int storageNativeTransportPort) {
```
marco-scoppetta: can we inject the `Cluster` directly here? so that we can reuse this class in KGMS with no need to re-implementation

server/src/server/ServerFactory.java
```diff
@@ -37,22 +37,23 @@
  * This is a factory class which contains methods for instantiating a Server in different ways.
  */
 public class ServerFactory {
+    private static final int STORAGE_NATIVE_TRANSPORT_PORT = 9042;
```
marco-scoppetta: why don't we read this from config?

server/src/server/keyspace/KeyspaceManager.java
```diff
@@ -18,149 +18,29 @@
...
-    /**
-     * Loads the system schema using the provided Transaction.
-     */
-    private void loadSystemSchema(TransactionOLTP tx) {
-        AttributeType<String> keyspaceName = tx.putAttributeType("keyspace-name", AttributeType.DataType.STRING);
-        tx.putEntityType("keyspace").key(keyspaceName);
+        Set<String> result = storage.connect().getCluster().getMetadata().getKeyspaces().stream()
+                .map(KeyspaceMetadata::getName).collect(Collectors.toSet());
+        result.removeAll(internals);
+        return result.stream().map(KeyspaceImpl::of).collect(Collectors.toSet());
```
marco-scoppetta: minor, but ideally we could just use `filter()` rather than collecting in a set and then re-stream it

server/src/server/Server.java
lolski: As per our offline discussion, will keep this for now as it'll make more sense to solve it by re-architecting `Server` and `ServerFactory` as they have high overlap in terms of functionality.

server/src/server/keyspace/KeyspaceManager.java
lolski: Why can't we simply use it in KGMS as is? Is it because we will need to supply credentials?

marco-scoppetta: yes indeed

lolski: fixed

PR #5354, opened 2019-08-05
Make Core extensible by KGMS

marco-scoppetta:

Minor refactors to make Core extensible by KGMS

- inject `HadoopGraphFactory` in `SessionFactory`
- change visibility of variables

### DIFF
--- a/server/src/server/session/HadoopGraphFactory.java
+++ b/server/src/server/session/HadoopGraphFactory.java
@@ -39,10 +39,13 @@ import java.util.Properties;
  */
 public class HadoopGraphFactory {
 
-    private Config config;
+    protected Config config;
     //These properties are loaded in by default and can optionally be overwritten
     private static final Properties DEFAULT_OLAP_PROPERTIES;
     private static final String DEFAULT_OLAP_PATH = "resources/default-OLAP-configs.properties";
+    private static final String STORAGE_HOSTNAME = ConfigKey.STORAGE_HOSTNAME.name();
+    protected static final String STORAGE_KEYSPACE = ConfigKey.STORAGE_KEYSPACE.name();
+    protected static final String JANUSGRAPHMR_IOFORMAT_CONF = "janusgraphmr.ioformat.conf.";
 
     static {
         DEFAULT_OLAP_PROPERTIES = new Properties();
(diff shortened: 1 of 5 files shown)


server/src/server/session/HadoopGraphFactory.java
```diff
@@ -39,10 +39,13 @@
  */
 public class HadoopGraphFactory {
 
-    private Config config;
+    protected Config config;
```
lolski: is it protected for use by KGMS? If yes let's add this comment so the next person won't change it to `private`:
```
// NOTE: this field is used by Grakn KGMS and should be kept as 'protected'
```

server/src/server/session/HadoopGraphFactory.java
```diff
@@ -39,10 +39,13 @@
...
     //These properties are loaded in by default and can optionally be overwritten
     private static final Properties DEFAULT_OLAP_PROPERTIES;
     private static final String DEFAULT_OLAP_PATH = "resources/default-OLAP-configs.properties";
+    private static final String STORAGE_HOSTNAME = ConfigKey.STORAGE_HOSTNAME.name();
+    protected static final String STORAGE_KEYSPACE = ConfigKey.STORAGE_KEYSPACE.name();
```
lolski: same with these two

server/src/server/session/SessionFactory.java
```diff
@@ -46,32 +46,29 @@
...
-    // NOTE: this field is used by Grakn KGMS and should be kept public
-    protected final static int TIMEOUT_MINUTES_ATTRIBUTES_CACHE = 2;
+    private final static int TIMEOUT_MINUTES_ATTRIBUTES_CACHE = 2;
+    private final static int ATTRIBUTES_CACHE_MAX_SIZE = 10000;
 
-    // NOTE: this field is used by Grakn KGMS and should be kept public
-    protected final static int ATTRIBUTES_CACHE_MAX_SIZE = 10000;
-
-    private final JanusGraphFactory janusGraphFactory;
+    protected final JanusGraphFactory janusGraphFactory;
```
lolski: same comments as above for these `protected` fields

PR #5377, opened 2019-08-16
Prepare for supporting authenticated KeyspaceService operations

vmax:

In order to support authenticated operations in `KeyspaceService` in enterprise version, certain fields of class should be protected.

Make `keyspaceManager` and `janusGraphFactory` protected and non-final such that it's possible to rewrite them in child classes.


vmax:
After design discussion w @marco-scoppetta we decided to go with a different solution.

PR #5381, opened 2019-08-20
Convert attribute values to persisted data type

kasper-piskorski:
We are parsing attributes as they are - we do not take into account the validity or compatibility of the provided value with respect to the underlying attribute data type. This is fine as the QueryParser's job is to just parse patterns and it doesn't have any information about the specific data model. However, during query processing this potential lack of agreement between datatypes leads to problems, especially when reasoning (equivalence checks, rule applicability checks). As now all query processing happens via reasoning facilities, it makes sense to introduce a value conversion when we define atoms. In this way all attribute atoms always are defined in terms of the defined persisted data type. This PR accomplishes that.

- introduce an extra conversion step when building `AttributeAtom` ensuring the attribute value is consistent with the persisted data type
- make the `comparableDataTypes` a property of `DataType` instead of `ValueOperation`. In that way we can have a single place where compatible data types of attributes are defined. When defining a new datatype it is then required to define the immediate comparable data types.


haikalpribadi:
Don't forget to describe the PR, @kasper-piskorski

PR #5387, opened 2019-08-23
Fix an issue in CI caused by the custom flag in the scp command

lolski:

We've removed the `-T` option when calling `scp`. It is apparently a custom flag which is no longer supported.


lolski:
For how we came to the conclusion that `-T` is a custom flag, please refer to the following post that @vmax found : https://unix.stackexchange.com/questions/499958/why-does-scps-strict-filename-checking-reject-quoted-last-component-but-not-oth/500099#500099

PR #5390, opened 2019-08-26
Exit with correct exit code on invalid args

### LINKED ISSUE #5389: 'grakn' should exit with exit code 1 if it is called with unrecognised option in Linux, Mac, and Windows
This is the expected behavior:
```
$ grakn srv start
$ echo $?
1
```

This is the current behavior:
```
$ grakn srv start
$ echo $?
0
```

vmax:
## What is the goal of this PR?'

Fix #5389 

- Exit with `1` when an argument is missing or incorrect
- Remove unneeded `popd`

### DIFF
--- a/bin/grakn
+++ b/bin/grakn
@@ -43,7 +43,7 @@ exit_if_java_not_found() {
 # main routine
 # =============================================
 
-pushd "$GRAKN_HOME" > /dev/null
+cd "$GRAKN_HOME"
 exit_if_java_not_found
 
 if [ -z "$1" ]; then
--- a/bin/grakn
+++ b/bin/grakn
@@ -59,6 +59,7 @@ elif [ "$1" = "console" ]; then
     else
         echo "Grakn Core Console is not included in this Grakn distribution."
         echo "You may want to install Grakn Core Console or Grakn Core (all)."
+        exit 1
     fi
 elif [[ "$1" = "server" ]] || [[ "$1" = "version" ]]; then
     if [[ -f "${SERVER_JAR_FILENAME}" ]]; then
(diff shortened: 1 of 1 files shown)


bin/grakn
```diff
@@ -69,15 +69,14 @@ elif [[ "$1" = "server" ]] || [[ "$1" = "version" ]]; then
...
 else
     echo "Invalid argument: $1. Possible commands are: "
     echo "  Server:          grakn server [--help]"
     echo "  Console:         grakn console [--help]"
+    exit 1
 fi
 
-exit_code=$?
-
-popd > /dev/null
```
lolski: Can we keep the `popd`? Since we're doing a `pushd` at the top I think it's nice to keep it (though I know it's not neccessary for correctness since it's located at the end of the file)

vmax: @lolski `popd` effectively is a no-op since our script runs in a subshell. It has no effect on the calling shell

lolski: Yeah, I just find it weird that we're doing something and not put it back. It's like `open()`-ing a file but not calling `close()` afterwards.

If you prefer to remove it, can you remove the on in KGMS as well to maintain consistency?

vmax: @lolski 
>If you prefer to remove it, can you remove the on in KGMS as well to maintain consistency?

I'll do the same in PR to KGMS

lolski: As discussed, let's also change `pushd` to plain old `cd` in both this PR and the one to KGMS.

vmax: @lolski fixed in https://github.com/graknlabs/grakn/pull/5390/commits/52189c5ddae1958cee0ca364b8e73a45d9c52f32

PR #5412, opened 2019-09-09
Re-implement and adapt type sharding to Grakn Core 1.5.x

lolski:

Type sharding is a feature which is intended to maintain good query performance when working with large graph.  
 The main aim of type sharding is to alleviate a well-known graph problem known as the "supernode" problem. A supernode refers to a vertex which has many incidental edges. Queries that need to touch such a vertex will be slow. The bigger the graph is, the more likely it is for a supernode to exist in the graph.

In Grakn, a supernode can exist when there exists a single type with millions of instances. In this scenario, the type vertex is the supernode as it will have millions of edges to each and every instance.

With type sharding, the type vertex will be connected to several "shards", each of which connected to a portion of the instances.

Type sharding is performed when a transaction is committed.

- Perform type sharding on `TransactionOLTP::commit` when the threshold is reached. Determining when the next time we should shard is done using checkpoints. The checkpoints are persisted in the graph (see `TypeShardCheckpoint` for the implementation).
- Added tests in `TransactionOLTPIT` which test shard creation and threshold configuration
- Refactor: moved `TransactionOLTPIT` from the `kb` package into the more appropriate `session` package.


lolski:
comments from discussions with @flyingsilverfin @marco-scoppetta and @kasper-piskorski 
- `TypeShardCheckpoint` should be static given that it holds no state
- Add more comments to `TypeShardCheckpoint`
- Simplify `Optional` return type

lolski:
Thanks @kasper-piskorski ! Don't merge it yet as I'm still adding more commits to address suggestions from @marco-scoppetta

server/src/server/session/TransactionOLTP.java
```diff
@@ -101,6 +103,8 @@
  */
 public class TransactionOLTP implements Transaction {
     private final static Logger LOG = LoggerFactory.getLogger(TransactionOLTP.class);
+    private long TYPE_SHARD_THRESHOLD;
```
marco-scoppetta: I think this should be `final` (as it's initialised in the constructor) and not be all caps as it's not `static`

lolski: fixed on d67fe2b093939600a77f12ab93fbf6baeb46b5e4

PR #5416, opened 2019-09-10
Extracted //console and //bin to separate repos

### LINKED ISSUE #5269: Split //console into its own repo: graknlabs/console
## Problem to Solve

We need to move the `//console` package into its own repository as it holds Grakn Console: an application that serves as an "interface" for users to interact with Grakn. Like Grakn Workbase and all Grakn Clients, Grakn Console should work for both Grakn Core and KGMS. Its source code and development should be independent of Grakn Core and KGMS codebase.

## Proposed Solution

Break all dependency from `//console` to any other package in `@graknlabs_grakn_core`, and migrate the source code to a new repository: graknlabs/console.

haikalpribadi:

We need to move the `//console` package into its own repository as it holds Grakn Console: an application that serves as an "interface" for users to interact with Grakn. Like Grakn Workbase and all Grakn Clients, Grakn Console should work for both Grakn Core and KGMS. Its source code and development should be independent of Grakn Core and KGMS codebase.

Given that both Grakn Core and Console depends on one binary (`grakn.sh` and `grakn.bat`) to start the JAR, we also need to move the `//bin` package to its own repository in which Grakn Core and Console could both depend on without creating a circular dependency.

1) We moved `//console` package to https://github.com/graknlabs/console (fixes #5269)
2) We moved `//bin` package to https://github.com/graknlabs/common
3) Refactored this repository to depend on the new repositories


lolski:
@haikalpribadi there may be console tests which you need to remove:
- https://github.com/haikalpribadi/grakn/blob/refactor-console/test-end-to-end/distribution/GraknGraqlCommandsE2E.java#L235
- https://github.com/haikalpribadi/grakn/blob/refactor-console/test-end-to-end/distribution/GraknGraqlCommandsE2E.java#L241
- https://github.com/haikalpribadi/grakn/blob/refactor-console/test-end-to-end/distribution/GraknGraqlCommandsE2E.java#L265
- https://github.com/haikalpribadi/grakn/blob/refactor-console/test-end-to-end/distribution/GraknGraqlCommands_WithARunningGraknE2E.java#L82
- https://github.com/haikalpribadi/grakn/blob/refactor-console/test-end-to-end/distribution/GraknGraqlCommands_WithARunningGraknE2E.java#L102

PR #5417, opened 2019-09-10
Cache attributes within transaction

kasper-piskorski:
In situations when we insert multiple attributes within a transaction, especially when the attribute value is a super node (for example `male/female` gender), it is beneficial to cache the inserted attribute instead of doing a db lookup which requires an index check. Otherwise we need to fetch the same attribute multiple times. This PR introduces an extra transaction-based attribute cache to allow to cache attributes. 

Introduce attribute cache in `TransactionCache` to store attributes created in that transaction.

### DIFF
--- a/test-integration/server/session/cache/TransactionCacheIT.java
+++ b/test-integration/server/session/cache/TransactionCacheIT.java
@@ -61,6 +61,7 @@ import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertThat;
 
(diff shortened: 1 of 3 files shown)


test-integration/server/session/cache/TransactionCacheIT.java
```diff
@@ -371,6 +372,19 @@ public void whenInsertingAndDeletingInferredAttribute_instanceIsTracked(){
...
+        AttributeType<String> attributeType = tx.putAttributeType("resource", AttributeType.DataType.STRING);
+        String value = "banana";
+        Attribute attribute = AttributeTypeImpl.from(attributeType).create(value);
+        String index = Schema.generateAttributeIndex(attributeType.label(), value);
+
+        Attribute cachedAttribute = tx.cache().getAttributeCache().get(index);
+        assertNotNull(cachedAttribute);
+        assertEquals(attribute, cachedAttribute);
+        assertEquals(attribute, attributeType.attribute(value));
+    }
```
marco-scoppetta: to avoid memory leaks in the future can you add a trivial test to check the attribute gets removed from cache if deleted?

test-integration/server/session/cache/TransactionCacheIT.java
```diff
@@ -371,6 +373,22 @@ public void whenInsertingAndDeletingInferredAttribute_instanceIsTracked(){
         assertFalse(tx.cache().getInferredInstances().anyMatch(inst -> inst.equals(attribute)));
     }
 
+    @Test
+    public void whenInsertingAndDeletingAttribute_attributeCachedIsUpdated(){
+        AttributeType<String> attributeType = tx.putAttributeType("resource", AttributeType.DataType.STRING);
+        String value = "banana";
```
marco-scoppetta: yes!

PR #5422, opened 2019-09-12
Optimise instance check in ComputeExecutor

kasper-piskorski:
Similarly to #5397, we want to remove recursive `tx.stream` calls. When streaming compute queries, some of them perform extra checks to verify whether any instances are present. These currently perform stream calls on a transaction leading to multiple `tx.stream` calls within a single transaction stream method. This leads to poor performance and potential cycling problems.

- rename `QueryExecutor::traversal` method to `QueryExecutor::traverse` to avoid confusion with query planner traversals
- remove `tx.stream` calls in ComputeExecutor and replace them with safer `tx.executor().traverse()` calls


server/src/graql/executor/ComputeExecutor.java
```diff
@@ -687,24 +686,19 @@ private VertexProgram initStatisticsVertexProgram(GraqlCompute query, Set<LabelI
...
-//                .map(type -> var("x").isa(Graql.label(type))).collect(Collectors.toList());
-//
-//        return tx.get().graql().infer(false)
-//                .match(or(checkResourceTypes), or(checkSubtypes)).get().aggregate(ask()).execute();
+        Set<Label> targetLabels = targetTypeLabels(query);
+        ImmutableSet<Label> scopeLabels = scopeTypeLabels(query);
+        BiFunction<Label, Label, Pattern> patternFunction = (attributeType, type) -> Graql.and(
+                Graql.var("x").has(attributeType.getValue(), Graql.var()),
+                Graql.var("x").isa(Graql.type(type.getValue()))
+        );
```
marco-scoppetta: why is this one defined separately if it's used only once?

kasper-piskorski: yes and to avoid a cast to Pattern

test-integration/server/session/TransactionOLTPIT.java
```diff
@@ -335,6 +336,7 @@ public void whenShardingSuperNode_EnsureNewInstancesGoToNewShard() {
         assertThat(s3.links().collect(toSet()), containsInAnyOrder(s3_e1, s3_e2));
     }
 
+    @Ignore("flaky")
```
flyingsilverfin: why is sharding flaky?

kasper-piskorski: ask @lolski. I've already sent him my warmest regards :) and created an issue.

PR #5426, opened 2019-09-13
Install APT packages via Bazel

vmax:

Previously, all dependencies had the same version (equal to that of `grakn-core-server`). After splitting `grakn-core-bin` and `grakn-core-console` out to separate repositories (`@graknlabs_common` and `@graknlabs_console`) they got their own release cycles.

Introduce a runnable target `//test-deployment:apt-install` which installs `grakn-core-server` with dependencies (console, bin). Their versions are taken from workspace references file.

Additionally, `test-assembly-linux-apt` is removed since all of its functions are covered by deployment test


haikalpribadi:
You can describe the PR better by explaining what was wrong before and what is the approach of the solution, @vmax . That way we can understand the full context of this PR not only when we're reviewing it right now, but also when we revisit this PR again in the future.

PR #5430, opened 2019-09-17
Fix eager execution due to flatMap

flyingsilverfin:
We currently have a use of `flatMap()` on the stream of answers that is produced from Janus via our various transformations to produce Concepts and Answers. The `flatMap()` at the end of this sequence of stream manipulations is *not* necessarily lazy.

See:
https://bugs.openjdk.java.net/browse/JDK-8075939

To fix this we introduced [`Vavr`](vavr.io) which is able to properly handle lazy streams and `flatMap` operations. We introduce a minimal change using this library as required here, converting to a Vavr stream and back once the flatMap is added to the stream.

* Introduce a light dependency `io.vavr.vavr` that has no transitive dependencies beyond one of its own further packages
* Replace `flatMap` in `QueryExecutor.match` with a few steps to utilise `vavr`'s stream implementation that actually lazy


server/src/graql/executor/QueryExecutor.java
```diff
@@ -107,10 +109,17 @@ public QueryExecutor(TransactionOLTP transaction, boolean infer) {
             validateClause(matchClause);
 
             if (!infer) {
-                answerStream = matchClause.getPatterns().getDisjunctiveNormalForm().getPatterns().stream()
+
```
haikalpribadi: Mark this code with a `TODO: <description>` so we remember to remove them in the future once we upgrade to Java 10, @flyingsilverfin

lolski:
@flyingsilverfin LGTM! don't forget to set the appropriate labels and milestone next time ;)

PR #5436, opened 2019-09-20
Experimental KeyspaceStatistics using Cassandra's counter column

lolski:

This is the experimental `KeyspaceStatistics` implementation which uses Cassandra counters. The purpose of this experiment is to test if we can use it as a cache for `match; aggregate;` and/or `compute;` to speed up the query.

- The function `commit` and `count` behaves identically to the current keyspace statistics
- However, the atomicity behavior (or the lack thereof) may behave differently, given that we're not using the transaction object, but instead, a separate `Cluster`. We know that Janusgraph's `commit` to Cassandra isn't really atomic either, but this one is, even "less atomic", so to say.
- Counter table is `int` only. We can't have `float` or `double`, meaning we can't store other statistics such as the mean, median, min, and max of an attribute (which maybe useful for certain types of query: `match; get; aggregate mean;`)
- Need to do more research on counter precision - there seem to be failure conditions which can cause it to be incorrect: https://stackoverflow.com/questions/53794343/race-conditions-with-cassandra-counters
- Does not implement the in-memory cache count, therefore it is 10-20 times slower. It can't be easily implemented in a cluster scenario.
- Changes the data structure in an incompatible way such that you can no longer use data from 1.5.8.

### DIFF
--- a/server/src/server/ServerFactory.java
+++ b/server/src/server/ServerFactory.java
@@ -55,16 +55,15 @@ public class ServerFactory {
         LockManager lockManager = new LockManager();
 
         // CQL cluster used by KeyspaceManager to fetch all existing keyspaces
-        Cluster cluster = Cluster.builder()
+        Cluster.Builder clusterBuilder = Cluster.builder()
                 .addContactPoint(config.getProperty(ConfigKey.STORAGE_HOSTNAME))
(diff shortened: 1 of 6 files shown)


flyingsilverfin:
@lolski  i think we can close this no...

haikalpribadi:
did you ask my review on this by accident, @lolski ? Something triggered a codeownership review on this PR 2 minutes ago?

PR #5439, opened 2019-09-23
Use release-validate-deps to ensure that Grakn depends on released versions of common, console, graql, and protocol

lolski:

We have added a validation step using `//ci:release-validate-deps` in order to ensure that Grakn is releasable only if it depends on released versions of Graql, protocol, common, and console.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -284,8 +284,8 @@ jobs:
       - run: |
           set +e
           bazel run @graknlabs_build_tools//ci:release-validate-deps \
-            graknlabs_common graknlabs_graql \
-            graknlabs_protocol graknlabs_console
+            graknlabs_graql \
+            graknlabs_protocol
           is_releasable=$?
           set -e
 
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -286,6 +286,15 @@ jobs:
           export RELEASE_APPROVAL_TOKEN=$REPO_GITHUB_TOKEN
           bazel run @graknlabs_build_tools//ci:release-approval
 
+  release-validate:
+    machine: true
+    steps:
+      - install-bazel-linux-rbe
+      - checkout
+      - run: |
+          bazel run @graknlabs_build_tools//ci:release-validate-deps graknlabs_common \
```
haikalpribadi: Do we not always need to provide arguments after `--` with `bazel run`?

haikalpribadi: Can we have all the arguments in the next line? It'll just read better.

lolski: Both addressed in https://github.com/graknlabs/grakn/pull/5439/commits/38f962b87dfceccc9097eb855c7842f126ccc980

PR #5448, opened 2019-09-25
Use port 587 instead of the default 25 for sending release-notification

lolski:

We have used a more appropriate port `587` for sending SMTP email. The default one (port `25`) is throttled by default ([reference article](https://support.circleci.com/hc/en-us/articles/360007444314-Sending-email-from-a-container)) and therefore should not use that.


vmax:
LGTM :+1:

PR #5449, opened 2019-09-26
Implement Custom Lazy Stream Merging

flyingsilverfin:
Replace io.vavr (#5430) that was introduced to solve the lack of lazy `flatMap` that is mentioned in #5430. We then determined that conversion to the `io.vavr` Stream implementation is more lazy, but the conversion step itself does a single call to `.next()`. In practical terms, this means Grakn always precomputed the first answer on query, which is still not fully lazy.
This led to the decision to restore prior behavior and solve the eagerness problem by implementing our own lazy stream merging solution.

* Implement a lazy stream merging solution that uses `stream.iterator()` conversions interally
* Remove `io.vavr` as a dependency
* tests for lazy stream merging


flyingsilverfin:
Yeah i was just thinking that especially since i just found a bug with null pointers. Will add

server/test/graql/util/LazyMergingStreamTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+        List<Integer> numbers = flatStream.limit(3).collect(Collectors.toList());
+
+        verify(mockIntegerIterator, times(2)).next();
+        verify(mockIntegerIterator, times(3)).hasNext();
+
+        verify(mockSecondIntegerIterator, times(1)).next();
+        verify(mockSecondIntegerIterator, times(1)).hasNext();
+    }
+}
```
marco-scoppetta: very nice tests!!! 🚀 🚀 🚀 🚀

PR #5458, opened 2019-10-03
Disable janus consistency checks

kasper-piskorski:
Janus does optional consistency checking to avoid duplicate edges and duplicate properties on vertices.

The checking behaviour in case of encountering already existing equivalents is to throw.
Those checks are redundant for us as Grakn already ensures uniqueness in these cases.
When building transactions, disable consistency checking at Janus level.


marco-scoppetta:
super duper

PR #5459, opened 2019-10-03
Break client-java cyclic dependency -- big

flyingsilverfin:
Client-java no longer has code dependencies on Grakn Core. However, we still have interfaces in `grakn.core.api` that are no longer being used externally.
These are collapsed into `Keyspace` (previously `KeyspaceImpl`), `Session` (previously `SessionImpl`), and `TransactionOLTP` classes and the `grakn.core.api` package is deleted.

We also synchronize all tests that depend on client-java's new API which are no longer identical the those in `grakn.core` -- notable `grakn.client.concept.api` and the `Transaction` class on the client.

* Flatten keyspace, session and transaction interfaces in `grakn.core.api` into their implementations as they are not used externally
* Fix tests dependending on refactored client-java packages and classes


haikalpribadi:
Exciting!

flyingsilverfin:
Closing - this will be re-written into a large PR that attempts a large package refactor as well

Issue #5454, opened 2019-09-27
Migrate //common code to @graknlabs_common


haikalpribadi:
# Problem to Solve

//common contains a few nifty utility tools that are used in other repos. We started off by copying them to other repos because the codebase is still small, but now that we actually have a repo for @graknlabs_common, we might as well move it there already.

## Proposed Solution

Migrate //common code to @graknlabs_common

PR #5466, opened 2019-10-11
Rearchitect Grakn Packages

flyingsilverfin:
Grakn server has a tighly coupled and very circular dependency structure, which is why the main BUILD file contains both `server` and `graql`.

The primary goal is to split `server` and `graql` into separate bazel packages and introduce a better story and flow of the code. We imagine that:
`server` -- the implementation the Grakn server -- should receive RPC calls, instantiate sessions and transactions, required caches, etc. Queries are handed (via transactions) to `graql`.
`graql` -- the implementation of the Graql language -- can invoke its subpackages - `reasoner`, `executor`,  which in turn may hit `planning`.
Finally, we receive `Concept`s and `answers` that are returned again via the `server`. 

Getting to this flow involves introducing a set set of interfaces that reflect Grakn's overall architecture and capabilities. This in turn lets us break circular package dependencies between packages. This should discourage future circular dependencies, and encourage significantly better encapsulation by preventing us from over-exposing methods from implementations of classes to each other which was (and still is) prevalent.

Temporarily removes `console` from the binaries because of non-backwards compatible changes

* Create a set of interfaces for `server` and `graql`
* Split `server` and `graql` into their own top-level packages, and remove dependency from `graql` on `server`. Both now depend on the set of interfaces as an intermediate refactoring state


common/util/LazyMergingStream.java
```diff
@@ -14,9 +14,10 @@
  *
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
  */
 
-package grakn.core.graql.util;
+package grakn.core.common.util;
```
haikalpribadi: Make sure to split this class out of `//common` before we split `//common` into its own repository (https://github.com/graknlabs/grakn/issues/5454). We want to make sure that the `@graknlabls_common` repo only contain packages that are used by all other repos on @graknlabs.

PR #5471, opened 2019-10-16
Remove Autovalues from Grakn Core

flyingsilverfin:
We decided that AutoValues add an extra layer of indirection and make code more difficult to read and use, with small added benefits of automatic `equals()` and immutable `value` classes. This change removes them from out code base everywhere.

* Remove AutoValue from `Fragment`
* Remove AutoValue from `FragmentSet`
* Remove AutoValue from `Atomic` classes
* remove AutoValue from one or two other smaller places


graql/gremlin/fragment/FragmentImpl.java
```diff
@@ -108,7 +109,6 @@
     /*
      * This is the memoized result of {@link #vars()}
      */
```
marco-scoppetta: move also the comment

PR #5492, opened 2019-11-08
Selective locking on commit

kasper-piskorski:
When loading concurrently, we want to minimise the amount of locking needed to improve performance whilst retaining correctness and consistency.

Currently we lock the graph if there are any attribute changes in any of the committing transactions. This in practice means that we lock practically always and all the time for mixed loads. This is however very defensive and not required as the only situations that require locking on commit are when two transactions share an inserted attribute or want to create a shard vertex for the same type.

To be able to lock more effectively, we need to be able to a) lock selectively based on the creation of identical attributes among different transaction, b) lock selectively based on the creation of shards of the same type among different transaction. Hence we introduce a basic selective locking behaviour.

- if two transactions insert the same attribute, we require them to acquire graph locks when committing. To be able to recognise this situation taking place, we introduce the `AttributeManager`. The `AttributeManager` is a session-wide object used to manage attribute mutations. All transactions must report their attribute insertions, deletions and commits to the `AttributeManager`. The `AttributeManager` then tracks the attribute mutations as well as txs in which the mutations took place. By logging this information the `AttributeManager` can then find and recognise possible contention. The contention resolution happens in the following way. When about to commit, each transaction polls the `AttributeManager` if it needs to use a lock during commit. If the `AttributeManager` finds at least two transactions with a shared attribute, it will advise the competing transactions to use a lock when committing.

- if two transactions are about to insert a shard for the same type, we require them to acquire graph locks when committing. To do so we introduce the `ShardManager`. The `ShardManager` is a session-wide object used to resolve shard creation contention. Just before committing, all transactions are required to signal their need to create a new shard vertex to the `ShardManager`. The `ShardManager` then tracks the transaction shard requirements for specific types. This way it can find and resolve possible contention. When a transaction is about to commit, it polls the `ShardManager` if it needs to use a lock during commit. If the `ShardManager` finds at least two transactions with a shared shard vertex request, it will advise the competing transaction to use a lock when committing.


kb/server/ShardManager.java
```diff
@@ -0,0 +1,47 @@
...
+ * To be able to lock more effectively, we need to lock selectively based on the creation of shards of the same type.
+ *
+ * The idea is that if two transactions are about to insert a shard for the same type, we require them to acquire graph locks when committing.
+ * To do so we introduce the ShardManager. The ShardManager is a session-wide object used to resolve shard creation contention.
+ * Just before committing, all transactions are required to signal their need to create a new shard vertex to the ShardManager.
+ * The ShardManager then tracks the transaction shard requirements for specific types. This way it can find and resolve possible contention.
+ *
+ * When a transaction is about to commit, it polls the ShardManager if it needs to use a lock during commit.
+ * If the ShardManager finds at least two transactions with a shared shard vertex request, it will advise
+ * the competing transaction to use a lock when committing.
```
marco-scoppetta: boom

server/session/AttributeManagerImpl.java
```diff
@@ -0,0 +1,104 @@
...
+        this.lockCandidates = ConcurrentHashMap.newKeySet();
+    }
+
+    @Override
+    public Cache<String, ConceptId> attributesCache() {
+        return attributesCache;
+    }
+
+    @Override
+    public void ackAttributeInsert(String index, String txId) {
```
marco-scoppetta: quick description of the logic if possible

server/session/AttributeManagerImpl.java
```diff
@@ -0,0 +1,104 @@
...
+            else{
+                entry.add(txId);
+                if (entry.size() > 1) lockCandidates.addAll(entry);
+                return entry;
+            }
+        });
+    }
+
+    @Override
+    public void ackAttributeDelete(String index, String txId) {
```
marco-scoppetta: quick description of the logic if possible

server/session/TransactionOLTP.java
```diff
@@ -171,24 +170,36 @@ public void open(Type type) {
...
+                || shardLockRequired
+        // In this case we need to lock, so that other concurrent Transactions
+        // that are trying to create new attributes will read an updated version of attributesCache
+        // Not locking here might lead to concurrent transactions reading the attributesCache that still
+        // contains attributes that we are removing in this transaction.
+                || !cache().getRemovedAttributes().isEmpty()
+                || cache().modifiedKeyRelations();
+
+        if (lockRequired){
+            LOG.warn(txId + " is about to acquire a " + (shardLockRequired? "shard/" : "") +
```
marco-scoppetta: remember to remove the warns before merging

server/session/TransactionOLTP.java
```diff
@@ -171,24 +170,36 @@ public void open(Type type) {
...
-            // In this case we need to lock, so that other concurrent Transactions
-            // that are trying to create new attributes will read an updated version of attributesCache
-            // Not locking here might lead to concurrent transactions reading the attributesCache that still
-            // contains attributes that we are removing in this transaction.
+        computeShardCandidates();
+        String txId = this.janusTransaction.toString();
+        //TODO: asks for a lock manager
+        boolean attributeLockRequired = session.attributeManager().requiresLock(txId);
+        boolean shardLockRequired = session.shardManager().requiresLock(txId);
+        boolean lockRequired = attributeLockRequired
```
marco-scoppetta: maybe `lockRequired` can become a method with comment

server/session/TransactionOLTP.java
```diff
@@ -197,44 +208,72 @@ private void persistInternal() throws InvalidKBException {
...
+    }
+
+    private void computeShardCandidates() {
+        uncomittedStatisticsDelta.instanceDeltas().entrySet().stream()
+                .filter(e -> !Schema.MetaSchema.isMetaLabel(e.getKey()))
+                .forEach(e -> {
+            Label label = e.getKey();
+            Long uncommittedCount = e.getValue();
+            long instanceCount = session.keyspaceStatistics().count(this, label) + uncommittedCount;
+            long lasShardCheckpoint = getShardCheckpoint(label);
```
marco-scoppetta: last

test-end-to-end/distribution/ConcurrencyE2E.java
```diff
@@ -146,4 +155,101 @@ public void concurrentInsertionOfDuplicateAttributes_doesNotCreateGhostVertices(
...
+            asyncInsertions.add(asyncInsert);
+        }
+        CompletableFuture.allOf(asyncInsertions.toArray(new CompletableFuture[]{})).get();
+        executorService.shutdown();
+    }
+
+    private static String generateString(int length) {
+        final boolean useLetters = true;
+        final boolean useNumbers = true;
+        return RandomStringUtils.random(length, useLetters, useNumbers);
```
marco-scoppetta: this method can just be `return RandomStringUtils.random(length, true, true);`, also no need to make random `final` variables

test-end-to-end/distribution/element/AttributeElement.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+package grakn.core.distribution.element;
+
+import graql.lang.Graql;
+import graql.lang.pattern.Pattern;
+import graql.lang.statement.Variable;
+import java.util.Objects;
+
+public class AttributeElement implements Element{
```
marco-scoppetta: what's this class for?

test-integration/server/kb/structure/EdgeIT.java
```diff
@@ -71,26 +70,21 @@
...
 
         session = new SessionImpl(keyspace, server.serverConfig(), new KeyspaceSchemaCache(), graph,
-                new KeyspaceStatistics(), attributeCache, new ReentrantReadWriteLock());
+                new KeyspaceStatistics(), attributeManager, new ShardManagerImpl(), new ReentrantReadWriteLock());
 
         // create the transaction
         CacheProviderImpl cacheProvider = new CacheProviderImpl(new KeyspaceSchemaCache());
         UncomittedStatisticsDelta statisticsDelta = new UncomittedStatisticsDelta();
-        ConceptObserver conceptObserver = new ConceptObserver(cacheProvider, statisticsDelta);
+        ConceptObserver conceptObserver = new ConceptObserver(cacheProvider, statisticsDelta, attributeManager, graph.toString());
```
marco-scoppetta: graph.toString()?  shouldn't it be `txId`?

test-integration/util/GraqlTestUtil.java
```diff
@@ -115,4 +123,43 @@ public static Thing getInstance(Transaction tx, String id){
         }
         return things.iterator().next();
     }
+
+    public static void insertStatements(Session session, List<Statement> statements,
```
marco-scoppetta: given that this is now in Utils maybe let's explain what's happening and maybe a less generic name, something like `insertStatementsConcurrently` or `withLatch` or `maximiseContention`

concept/impl/ConceptImpl.java
```diff
@@ -39,12 +39,11 @@
  */
 public abstract class ConceptImpl implements Concept, ConceptVertex {
     private final VertexElement vertexElement;
-    final ConceptManagerImpl conceptManager;
+    ConceptManagerImpl conceptManager;
```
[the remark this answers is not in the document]
kasper-piskorski: CAUSE LAMBDA

test-end-to-end/distribution/ConcurrencyE2E.java
kasper-piskorski: yes, but then we don't know what the flags are for, no? It makes it more explicit

marco-scoppetta: it's all one click away, but ok it makes sense, but really no need to be final

concept/impl/ConceptImpl.java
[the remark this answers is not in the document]
marco-scoppetta: what lambda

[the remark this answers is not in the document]
marco-scoppetta: ah I see, but `final` carries a meaning when used as a field - it tells us the it will passed as a collaborator in the constructor. So I would try to find a workaround for the lambda rather than sacrificing the `final`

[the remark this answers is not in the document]
marco-scoppetta: and also the `final` here makes actual sense, so that nobody tries to override it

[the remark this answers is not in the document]
kasper-piskorski: can remove the cache and use a simple null initialiser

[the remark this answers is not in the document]
marco-scoppetta: can we initialise the caches in the constructor?

test-integration/server/kb/structure/EdgeIT.java
kasper-piskorski: graph.toString() is what we currently use as an Id. COuld revise that.

concept/impl/ConceptImpl.java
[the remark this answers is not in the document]
kasper-piskorski: can try that as well

test-integration/server/kb/structure/EdgeIT.java
marco-scoppetta: https://github.com/graknlabs/grakn/pull/5492/files#diff-7226e2759816d4691e666d175f134c4aR157

here we are using transaction.toString()

concept/impl/ConceptImpl.java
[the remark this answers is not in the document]
marco-scoppetta: yeah, I think initialising things outside the constructor makes only sense if `static`

test-integration/server/kb/structure/EdgeIT.java
kasper-piskorski: Ok true.

kb/server/AttributeManager.java
```diff
@@ -0,0 +1,52 @@
...
+    Cache<String, ConceptId> attributesCache();
+
+    void ackAttributeInsert(String index, String txId);
+    void ackAttributeDelete(String index, String txId);
+    void ackAttributeCommit(String index, String txId);
+    void ackCommit(String txId);
+
+    boolean requiresLock(String txId);
+    boolean lockCandidatesPresent();
+    boolean shardRequestsPresent();
```
flyingsilverfin: is this @visiblefortesting?

flyingsilverfin: What is this even for?

kb/server/AttributeManager.java
```diff
@@ -0,0 +1,52 @@
...
+
+    Cache<String, ConceptId> attributesCache();
+
+    void ackAttributeInsert(String index, String txId);
+    void ackAttributeDelete(String index, String txId);
+    void ackAttributeCommit(String index, String txId);
+    void ackCommit(String txId);
+
+    boolean requiresLock(String txId);
+    boolean lockCandidatesPresent();
```
flyingsilverfin: Is this a shortcut method - why do we need this and `requiresLock`?

server/session/AttributeManagerImpl.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+
+public class AttributeManagerImpl implements AttributeManager {
+    private final static int TIMEOUT_MINUTES_ATTRIBUTES_CACHE = 2;
+    private final static int ATTRIBUTES_CACHE_MAX_SIZE = 10000;
+
+    private final Cache<String, ConceptId> attributesCache;
+    //we track txs that insert an attribute with given index
+    private final ConcurrentHashMap<String, Set<String>> ephemeralAttributeCache;
```
flyingsilverfin: Better name might be `attributesUnderModification` - definitely not a cache...

server/session/ShardManagerImpl.java
```diff
@@ -0,0 +1,98 @@
...
+    public void ackShardCommit(Label type, String txId) {
+        shardRequests.merge(type, ConcurrentHashMap.newKeySet(), (existingValue, newValue) -> {
+            if (existingValue.size() == 1) return null;
+            existingValue.remove(txId);
+            return existingValue;
+        });
+    }
+
+    @Override
+    public void ackCommit(String txId) {
```
flyingsilverfin: Do we need both `ackShardCommit` and `ackCommit`?

concept/impl/ConceptManagerImpl.java
```diff
@@ -492,6 +493,18 @@ public Role getRole(String label) {
         return getSchemaConcept(Label.of(label), Schema.BaseType.ROLE);
     }
 
+    Shard getShardWithLock(String typeId) {
```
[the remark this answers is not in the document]
flyingsilverfin: +1 on this

kb/server/AttributeManager.java
kasper-piskorski: last two are for testing

concept/impl/ConceptManagerImpl.java
```diff
@@ -492,6 +493,18 @@ public Role getRole(String label) {
...
+        graphLock.readLock().lock();
+        Vertex shardVertex;
+        try {
+            Object currentShardId = elementFactory.getVertexWithId(typeId).property(Schema.VertexProperty.CURRENT_SHARD.name()).value();
```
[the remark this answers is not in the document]
kasper-piskorski: Yep, we don't know the specific type.

test-end-to-end/distribution/element/AttributeElement.java
kasper-piskorski: testing mixed attribute load

kb/server/AttributeManager.java
kasper-piskorski: I've added a 'VisibleForTesting' annotation.

PR #5511, opened 2019-11-21
Revise attribute merging mechanism

kasper-piskorski:
To revise the attribute merging mechanism and adopt it to the latest locking changes.

- the `attributeCache` now strictly holds attributes that were committed
- revised `getAttribute` behaviour - the read lock was removed instead we check the `attributesCommitted`, `attributeEphemeral` and then the db

### DIFF
--- a/concept/impl/AttributeTypeImpl.java
+++ b/concept/impl/AttributeTypeImpl.java
@@ -156,7 +156,7 @@ public class AttributeTypeImpl<D> extends TypeImpl<AttributeType<D>, Attribute<D
      */
     private Attribute<D> getAttributeWithLock(D value) {
         String index = Schema.generateAttributeIndex(label(), value.toString());
-        return conceptManager.getAttributeWithLock(index);
+        return conceptManager.getAttribute(index);
     }
 
     /**
--- a/concept/impl/AttributeTypeImpl.java
+++ b/concept/impl/AttributeTypeImpl.java
@@ -112,7 +112,7 @@ public class AttributeTypeImpl<D> extends TypeImpl<AttributeType<D>, Attribute<D
 
         if (dataType().equals(DataType.STRING)) checkConformsToRegexes((String) value);
 
-        Attribute<D> instance = getAttributeWithLock(value);
+        Attribute<D> instance = getAttribute(value);
         if (instance == null) {
             // create a brand new vertex and concept
             instance = conceptManager.createAttribute( this, value, isInferred);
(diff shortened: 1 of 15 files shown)


concept/impl/AttributeTypeImpl.java
```diff
@@ -156,7 +156,7 @@ private void checkConformsToRegexes(String value) {
      */
     private Attribute<D> getAttributeWithLock(D value) {
```
marco-scoppetta: change name also to the method

concept/impl/ConceptManagerImpl.java
```diff
@@ -377,18 +378,20 @@ Attribute getCachedAttribute(String index) {
...
         if (concept != null) return concept;
 
-        graphLock.readLock().lock();
-        try {
-            return getConcept(Schema.VertexProperty.INDEX, index);
-        } finally {
-            graphLock.readLock().unlock();
-        }
+        //check EPHA
+        if (attributeManager.attributesEphemeral().containsKey(index)) return null;
```
marco-scoppetta: if not in ephemeral it can still be in the graph 🌞

kasper-piskorski: yes, that's why we keep checking

server/session/TransactionOLTP.java
```diff
@@ -213,31 +216,43 @@ private void persistInternal() throws InvalidKBException {
...
-        Set<String> newAttributeIndices = cache().getNewAttributes().keySet().stream().map(Pair::second).collect(Collectors.toSet());
+    private void ackCommit(){
+        String txId = this.janusTransaction.toString();
         session.shardManager().ackCommit(cache().getNewShards().keySet(), txId);
-        session.attributeManager().ackCommit(newAttributeIndices, txId);
+        //this should ack all inserts so that insert requests are cleared
+        session.attributeManager().ackCommit(cache().getNewAttributes().keySet(), txId);
 
-        LOG.trace("Graph committed.");
+        //this should ack all inserts that weren't deduplicated so that we have correct attributes in attributesCommitted
```
marco-scoppetta: I understand this, but if we are merging 2 attributes A, it's because A was already in `attributesCommitted`, right?
So would it be so bad to re-put?
We could safe a field and some additional logic, but maybe I am missing something. I keep getting confused

kasper-piskorski: if we add all new attributes, we will overwrite legit entries in `attributesCommitted` with the ones that were deduplciated

server/session/TransactionOLTP.java
```diff
@@ -183,26 +183,32 @@ boolean commitLockRequired(){
...
                 // that are trying to create new attributes will read an updated version of attributesCache
                 // Not locking here might lead to concurrent transactions reading the attributesCache that still
                 // contains attributes that we are removing in this transaction.
                 || !cache().getRemovedAttributes().isEmpty()
                 || keyLockRequired;
+        LOG.warn(txId + " needs lock: " +
+                (attributeLockRequired? "attribute" : "") +
+                (shardLockRequired? "shard" : "") +
+                (keyLockRequired? "key" : "")+
+                (!cache().getRemovedAttributes().isEmpty()? "delete" : ""));
```
marco-scoppetta: remember the warns before merging

PR #5512, opened 2019-11-21
Refactor architecture - Removing transaction dependencies

flyingsilverfin:
We remove `Transaction` from throughout reasoner, executors, etc. We make explicit dependencies on `ConceptManager`, `ReasonerQueryFactory`, `ExecutorFactory`, and the various caches. Dependencies are injected as collaborators, unneeded static factories and class hierarchies are removed, and methods live in the correct packages.

* Extract `Transaction` from the `graql` package completely
* New stateful factories:
  * `ReasonerQueryFactory` is stateful and replaces `ReasonerQueries`, which was static
  * `PropertyAtomicFactory` collects the conversions from `Property`s to `Atomic`s, which used to be lumped into the `executors`
  * `ExecutorFactory` returns either new `QueryExecutor` for transactional queries, or a `ComputeExecutor` for OLAP queries. In the future we probably want to split the query executors into further types such as `ReadExecutor`, `SchemaExecutor`, `InsertExecutor`, `DeleteExecutor`, `ComputeExecutor`... etc. This should let us extend the capabilities of Grakn easily without making the generic `QueryExecutor` the catch-all.
  * `JanusTraversalSourceProvider` exposes the Janus/Gremlin traversals, making it explicit who depends on this component directly
* We utilise a `ConceptNotificationChannel` to link `ConceptManager` and `ConceptObserver` (now `ConceptListener`). This naturally leads to a bit of a pub-sub model to update the `ConceptObserver`, but this is a overly generic and hard to debug approach and normally used when dynamically adding and removing listeners and publishers. See the class comment in `ConceptNotificationChannel.java` for ideas
* Few new interfaces as required for dependencies
* `TransactionOLAP`/`TransactionAnalytics` is gone, we only have `TransactionImpl`, which can use the `ComputeExecutor`, which in turn uses the `OLAPOperation` class to do OLAP jobs
* Refactor testing process to be able to spawn `Storage` without `GraknServer`, allowing us to build and inject the new `TransactionProvider` class for tests. The `TestTransaction` that are produced by `Session.transaction()` are then exposing getters for transaction internals that we want to test. Work to be continued as in #5529 

## Future

In future iterations, when we are satisfied with a stable, explicit and verbose code flow, we can look into re-combining factories, deleting uneeded interfaces, and reducing the number of arguments passed into constructors, while keeping the same readability levels.

The end goal is still to have a good flow and dependency in the code that mirrors a query execution:
client -> Graql -> transaction -> rewrite -> validate -> reasoner -> executor -> traversal -> answer [ --> caches, back to reasoner ].
A good benchmark is if we are able to split the interfaces package `kb` into a dependency flow like the one above, rather than a big interdependent set of classes.


concept/answer/BUILD
```diff
@@ -25,12 +25,14 @@ java_library(
...
     srcs = glob(["*.java"]),
     deps = [
+        "//concept/util:util",
         "//kb/concept/api:api",
 
         # External dependenicies from @graknlabs
         "@graknlabs_graql//java:graql",
 
         # External dependencies from Maven
+        "//dependencies/maven/artifacts/com/google/guava",
```
marco-scoppetta: This is minor, but I think the idea was to start collecting `Collection`s and utils in our `common` package so that we don't have to depend on Guava for trivial things, like `Sets.newHashSet()` instead of the native `new HashSet()`, also functions like `intersection` I am sure can be copied over in common.
Again, this is just a comment for the future, but ideally if we are not using fancy Guava functionalities, at some point  we could stop depending on it. But ofc it's a long term thing, because some of our deps still depend on Guava 😄

concept/impl/ConceptImpl.java
```diff
@@ -39,21 +41,21 @@
  */
 public abstract class ConceptImpl implements Concept, ConceptVertex {
     private final VertexElement vertexElement;
-    final ConceptManagerImpl conceptManager;
-    final ConceptObserver conceptObserver;
+    final ConceptManager conceptManager;
+    final ConceptNotificationChannel conceptNotificationChannel;
 
     //WARNING: DO not flush the current shard into the central cache. It is not safe to do so in a concurrent environment
     private final ConceptCache<Shard> currentShard;
```
marco-scoppetta: I know it's not related to your PR, but please delete the above WARNING comment if possible. We already decided to delete it, but we keep forgetting it. It's just outdated now. Fearmongering!

concept/impl/RelationEdge.java
```diff
@@ -49,8 +52,8 @@
     private final Logger LOG = LoggerFactory.getLogger(RelationEdge.class);
 
     private final EdgeElement edgeElement;
-    private final ConceptManagerImpl conceptManager;
-    private ConceptObserver conceptObserver;
+    private final ConceptManager conceptManager;
+    private ConceptNotificationChannel conceptNotificationChannel;
```
marco-scoppetta: final

concept/structure/ElementFactory.java
```diff
@@ -57,12 +55,11 @@
  */
 public final class ElementFactory {
     private final JanusGraphTransaction janusTx;
+    private JanusTraversalSourceProvider traversalSourceProvider;
```
marco-scoppetta: final

graql/executor/ComputeExecutorImpl.java
```diff
@@ -126,31 +136,36 @@
...
+        return new OLAP(hadoopGraph).compute(program, mapReduce, scope, includesRolePlayerEdges);
     }
 
+    @Override
     public final ComputerResult compute(@Nullable VertexProgram<?> program,
                                         @Nullable MapReduce<?, ?, ?, ?, ?> mapReduce,
                                         @Nullable Set<LabelId> scope) {
 
-        return tx.session().transactionOLAP().compute(program, mapReduce, scope);
+        return new OLAP(hadoopGraph).compute(program, mapReduce, scope);
```
marco-scoppetta: is OLAP the new name of TransactionOLAP?

marco-scoppetta: or OLAP operation?

graql/gremlin/TraversalPlanFactoryImpl.java
```diff
@@ -219,19 +227,19 @@ public GraqlTraversal createTraversal(Pattern pattern) {
         return weightedGraph;
     }
 
-    private static Set<Node> chooseStartingNodeSet(Set<Fragment> fragmentSet, Map<NodeId, Node> allNodes, SparseWeightedGraph sparseWeightedGraph, Transaction tx) {
+    private Set<Node> chooseStartingNodeSet(Set<Fragment> fragmentSet, Map<NodeId, Node> allNodes, SparseWeightedGraph sparseWeightedGraph) {
         final Set<Node> highPriorityStartingNodeSet = new HashSet<>();
         final Set<Node> lowPriorityStartingNodeSet = new HashSet<>();
```
marco-scoppetta: I know it's not a change of this PR again and also this is more personal/stylistic decision, but I think we should avoid using random `final` keywords inside methods. The code inside a method is written by us, these are not structures injected by the user nor that will be shared with the outside world. The methods should never be so long or complex that we lose track of local variables, so that we need to declare them final. My reasoning behind this is: let's not start throwing random defensive mechanisms here and there, and also let's avoid adding unnecessary noise to the whole repo. There are programming languages which have native final variables, such as JS or C++, Java unfortunately doesnt, so we should learn how to and when to use `final` and also what is it really protecting us against? The main problems usually happens when people start mutating collections, it rarely happens that someone inside a method tries to re-allocate a variable. But anyway, food for thoughts 👋

graql/gremlin/fragment/FragmentImpl.java
```diff
@@ -253,11 +253,11 @@ Variable end() {
 
     /**
      * @param traversal the traversal to extend with this Fragment
-     * @param tx        the transaction to execute the traversal on
+     * @param conceptManager        the transaction to execute the traversal on
```
marco-scoppetta: probably worth updating the comment or delete it altogether

kb/graql/executor/ExecutorFactory.java
```diff
@@ -0,0 +1,25 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package grakn.core.kb.graql.executor;
+
+public interface ExecutorFactory {
+    ComputeExecutor compute();
+    QueryExecutor transactional(boolean infer);
```
marco-scoppetta: maybe in the future better naming? why not just call a method that returns a new queryExecutor `buildQueryExecutor()` or just `queryExecutor()`? maybe I dont have enough context, just looks strange

marco-scoppetta: same for `createComputeExecutor()`/`buildComputeExecutor()`/`computeExecutor()`

graql/executor/ComputeExecutorImpl.java
flyingsilverfin: has been renamed to OLAPOperation for now, because we can run many of these within one Transaction still

graql/reasoner/query/ReasonerQueries.java
```diff
@@ -35,142 +35,142 @@
  * Factory for reasoner queries.
  */
 public class ReasonerQueries {
```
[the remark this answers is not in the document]
flyingsilverfin: has been deleted

graql/reasoner/query/ReasonerQueryImpl.java
```diff
@@ -90,45 +94,80 @@
...
+        this.atomSet = null;
+        this.reasonerQueryFactory = reasonerQueryFactory;
     }
 
-    ReasonerQueryImpl(Set<Atomic> atoms, Transaction tx){
-        this.tx = tx;
+    /*
+    TODO this should not be needed in a better structure, is a risky split of constructor
+     */
+    void setAtomSet(ImmutableSet<Atomic> atoms) {
```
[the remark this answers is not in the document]
flyingsilverfin: yes, will be resolved in upcoming PRs i hope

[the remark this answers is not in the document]
flyingsilverfin: Links to #5531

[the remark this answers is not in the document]
flyingsilverfin: Has been resolved for now by passing the factory for constructing the atoms into the constructor

PR #5536, opened 2019-12-11
Split //kb/graql into Subpackages

flyingsilverfin:
Split the `//kb/graql` package into subpackages for _reasoner, executor, analytics, planning, and exception_, further cleaning up our dependency graph. The finer granularity allows us to control package access and prevent unneccesarily exposed classes. It also will allow us to change smaller code regions at a time.

* move `//kb/graql/gremlin` into `//kb/graql/planning`
* Create separate BUILD files for `//kb/graql/planning`, `//kb/graql/reasoner`, `//kb/graql/executor`, and `//kb/graql/exception`
* Move `AttributeSerialiser` and `AttributeValueConverter` into `//core` to avoid having everyone depend on `//concept/impl`
* Clean up `Transaction` interface, no more exposed `@VisibleForTesting` getters
* Relocate and split exceptions into sub-packages, notably `GraknAnalyticsException` and `GraknQueryPlanningException`
* Relocate tests as needed to match the new structure

## Dependency Graph
This graph is more verbose and larger, but more granular. We can start to see the flow from server to executor to reasoner to planning!

image


test-integration/graql/query/QueryPlannerIT.java
```diff
@@ -398,6 +399,11 @@ public void sameLabelFragmentShouldNotBeAddedTwice() {
                 .filter(fragment -> fragment instanceof OutIsaFragment || fragment instanceof InIsaFragment).count());
     }
 
+    /**
+     * We no longer rely on shard count to order query plan starting points, using instance counts directly
+     * Whether we want to delete this test is not yet clear, or just keep a similar one
+     */
+    @Ignore
```
kasper-piskorski: If it is not relevant anymore we should probably remove it. Alternatively replace it with an appropriate equivalent.

PR #5537, opened 2019-12-11
Revise query cache behaviours

kasper-piskorski:

To revise the query cache behaviours so that we can increase the cache efficiency. In certain scenarios query results weren't reused when they could have. The specific scenario we address here is when there is a subsumption relation between a cache query and input query up to the choice of ids - which we refer to as structural subsumption.

Previously, we stored query cache entries with specific IDs attached in the lookup key. This led to a second layer of stored IDs that are all valid substitutions into the query. Effectively, the specific IDs attached at the lookup key are placeholders that we now ignore (using structural subsumption), and so can find all the IDs matching a query rather than a specific one.

## What are the changes implement in this PR?
* introduced `STRUCTURAL_SUBSUMTIVE` unifier for determination of structural subsumption relation
* query cache entries are now compared wrt structural subsumption
* revised the `GET` and `PROPAGATE` behaviours to incorporate the new comparison

### DIFF
--- a/graql/reasoner/cache/SemanticCache.java
+++ b/graql/reasoner/cache/SemanticCache.java
@@ -49,7 +49,7 @@ import static java.util.stream.Collectors.toSet;
  * and subsequently reusing their answer sets.
  *
  * Relies on the following concepts:
- * - Query Subsumption {@link ReasonerAtomicQuery#subsumes(ReasonerAtomicQuery)}
+ * - Query Subsumption {@link ReasonerAtomicQuery#isSubsumedBy(ReasonerAtomicQuery)} )}
  * Subsumption relation between a query C (child) and a provided query P (parent) holds if:
  *
  * C <= P,
@@ -89,7 +89,7 @@ public abstract class SemanticCache<
     public boolean isComplete(ReasonerAtomicQuery query){
         if (super.isComplete(query)) return true;
         return getParents(query).stream()
-                .filter(q -> query.subsumes(keyToQuery(q)))
+                .filter(q -> query.isSubsumedBy(keyToQuery(q)))
                 .anyMatch(q -> super.isComplete(keyToQuery(q)));
     }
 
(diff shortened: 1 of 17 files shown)


graql/reasoner/cache/SemanticCache.java
```diff
@@ -219,15 +213,24 @@ private boolean propagateAnswersToQuery(ReasonerAtomicQuery target, CacheEntry<R
...
                         boolean propagateInferred = fetchInferred || parentComplete || child.getAtom().getVarName().isReturned();
                         boolean newAnswers = propagateAnswers(parentMatch, childMatch, propagateInferred);
-                        newAnswersFound[0] = newAnswers;
-                        if (parentDbComplete || newAnswers) ackDBCompleteness(target);
-                        if (parentComplete) ackCompleteness(target);
+                        newAnswersFound[0] = newAnswersFound[0] || newAnswers;
+
+                        boolean targetSubsumesParent = target.subsumes(keyToQuery(parent));
+                        //since we compare queries structurally, freshly propagated answer might not necessarily answer
+                        //the target query - hence this check
```
flyingsilverfin: good to have comment when unusual/confusing action taken!

test-integration/graql/reasoner/cache/QueryCacheIT.java
```diff
@@ -300,19 +301,79 @@ public void whenGettingAndMatchDoesntExist_prospectiveParentCached_childQueriesA
...
+            assertNotNull(cache.getEntry(indirectGroundChildQuery));
+            assertFalse(cache.isDBComplete(indirectGroundChildQuery));
+        }
+    }
+
+    @Test
+    public void whenGettingAndMatchExists_queryGround_queryDBComplete_answerNotFound_answersFetchedFromDbAndCache(){
+        try(Transaction tx = genericSchemaSession.readTransaction()) {
+            TestTransactionProvider.TestTransaction testTx = ((TestTransactionProvider.TestTransaction)tx);
+            MultilevelSemanticCache cache = new MultilevelSemanticCache(testTx.executorFactory(), testTx.traversalPlanFactory());
```
flyingsilverfin: add a little comment to say why a new cache rather than from the testTx

PR #5546, opened 2020-01-08
Correctly interpret type compatibility and playability when unifying queries

kasper-piskorski:
By overpruning rules, we hit cases where no answers would be returned when they should be.

To clarify this, to check rule applicability we perform two tests about unifiability of the query and the rule: 
* INSERT semantics for the `then` clause of a rule
* MATCH semantisc for the `when` clause of the rule. 

Both of these tests need to pass in order for a rule to be resolved. The INSERT semantics check is a correctness test: the `then` clause of a rule is applicable if the `then` clause is more or equivalently specific to the comparable part of the input query.

The MATCH semantics check is a rule pruning optimisation. We check that the `when` part of the rule doesn't contradict the original query. If it does (for instance, disjoint types), the rule may be applicable but will never have any results.

This change explicitly writes these two semantics out and applies them to the `when` and `then` portions of a rule separately, as opposed to using the stricter INSERT semantics for both rule clauses, which overprunes rules.

We differentiate two types of checks having the different semantics:
 - MATCH (queries in general, rule bodies)
MATCH semantics include type hierarchies. We use them to check if a query can actually return any answers by looking at the conjunction of the input query and the `when` part of the rule.
 - INSERT (only rule heads)
INSERT semantics impose stricter type compatibility and playability criteria. They follow from the fact that rule conclusions need to be insertable - the asserted facts need to be well-defined and unambiguous (checked at commit-time). For a rule to be matched to a query, the rule head needs to be a specialisation of the input query (you cannot resolve a query when the `then` is a more general query than the user's input clause).

Example:

Let's define a simple rule:
when{ `(baseRole: $x, baseRole: $y) isa baseRelation;`},
then{ `(baseRole: $x, baseRole: $y) isa derivedRelation;`}.

with a user query:
query: `($x, $y), parentType($x), parentType($y)` 
where: 
`baseRelation relates baseRole, subRole;`
`derivedRelation relates baseRole, subRole;`
`subRole sub baseRole;`
`parentType plays subRole;`

Now we will consider whether the rule can be matched against the query following different semantics.

**MATCH semantics**:
MATCH semantics apply to the rule body, hence we check whether the
when:  `(baseRole: $x, baseRole: $y) isa baseRelation;`
and
query:  `($x, $y), parentType($x), parentType($y)` 

are compatible.

**RESULT**: compatible -> even though` parentType` doesn't play the `baseRole`, its subtypes might, so they need to be taken into account:
MATCHing `(baseRole: $x, baseRole: $y) isa baseRelation, parentType($x), parentType($y)` can potentially return results as matching `baseRole` will return relation instances with `subRole` as well.

**INSERT semantics**:
INSERT semantics apply to the rule head, hence we check whether the
then: `(baseRole: $x, baseRole: $y) isa derivedRelation;`
and
query:  `($x, $y), parentType($x), parentType($y)` 

are compatible.

**RESULT**:  incompatible -> 
`(baseRole: $x, baseRole: $y) isa derivedRelation, parentType($x), parentType($y)` is not a valid INSERT statement as it doesn't conform to the schema (`parentType` can't play `baseRole`)


graql/reasoner/query/ResolvableQuery.java
```diff
@@ -136,7 +136,8 @@ public GraqlGet getQuery() {
     public Stream<ConceptMap> resolve(Set<ReasonerAtomicQuery> subGoals){
         boolean doNotResolve = getAtoms().isEmpty() || (isPositive() && !isRuleResolvable());
         if (doNotResolve) {
-            return executorFactory.transactional(true).traverse(getPattern());
+            //NB: the flag actually doesn't affect the traverse method which doesn't use reasoning
```
flyingsilverfin: Add comment in QueryExecutorImpl that `traverse` should be private or moved!

graql/reasoner/unifier/UnifierComparison.java
```diff
@@ -120,14 +119,44 @@ default boolean predicateCompatibility(Set<Atomic> parent, Set<Atomic> child, Bi
     }
 
     /**
-     * @param query to be checked
+     * NB:
```
flyingsilverfin: align comments here with description in PR

PR #5549, opened 2020-01-14
Uniformly fetch relation concepts if query contains a relation variable

kasper-piskorski:

{ Please explain what you implemented, why your changes are the best way to achieve the goal(s) above, and reference the GitHub issues to be automatically closed, such like 'closes #number'. }


flyingsilverfin:
Is there a way to do this without rewriting all the rules

PR #5573, opened 2020-02-11
Added necessary deps for building and running simulation from grakn repo

adammitchelldev:

To add the dependencies required to build and run the simulation from this repo so that the grabl performance analysis is locked to a version of the simulation defined by this commit (and that can also be updated by grabl).

Added the `@graknlabs_simulation` workspace and necessary rules to load the tracing client dependencies.


haikalpribadi:
Woohoo! First PR onto grakn-core, @adammitchelldev well done! :)

PR #5590, opened 2020-02-25
Abstract semantic processing in atoms

kasper-piskorski:
To delegate the semantic tasks (unification and semantic diff computation) in `Atom`s to a new `SemanticProcessor` entity so that the `Atom` implementations are simpler so that we can eventually make them data classes.

We introduce an `SemanticProcessor` interface with implementations for `IsaAtom`, `RelationAtom` and `AttributeAtom`, which allows to abstract and delegate the semantic processing present in the corresponding `Atom`s.


graql/reasoner/atom/processor/SemanticProcessor.java
```diff
@@ -0,0 +1,44 @@
...
+
+package grakn.core.graql.reasoner.atom.processor;
+
+import grakn.core.graql.reasoner.atom.Atom;
+import grakn.core.graql.reasoner.cache.SemanticDifference;
+import grakn.core.graql.reasoner.unifier.UnifierType;
+import grakn.core.kb.graql.reasoner.unifier.MultiUnifier;
+import grakn.core.kb.graql.reasoner.unifier.Unifier;
+
+public interface SemanticProcessor<T extends Atom> {
```
flyingsilverfin: I think we can find a clearer name for this? TODO

graql/reasoner/atom/processor/SemanticProcessor.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Calculates the semantic difference between the this (parent) and child atom,
+     * that needs to be applied on A(P) to find the subset belonging to A(C).
+     *
+     * @param childAtom child atom
+     * @param unifier    parent->child unifier
+     * @return semantic difference between this and child defined in terms of this variables
+     */
+    SemanticDifference semanticDifference(T parentAtom, Atom childAtom, Unifier unifier);
```
flyingsilverfin: better name of method too I think is possible

graql/reasoner/atom/task/processor/SemanticProcessor.java
```diff
@@ -0,0 +1,44 @@
...
+
+package grakn.core.graql.reasoner.atom.task.processor;
+
+import grakn.core.graql.reasoner.atom.Atom;
+import grakn.core.graql.reasoner.cache.SemanticDifference;
+import grakn.core.graql.reasoner.unifier.UnifierType;
+import grakn.core.kb.graql.reasoner.unifier.MultiUnifier;
+import grakn.core.kb.graql.reasoner.unifier.Unifier;
+
+public interface SemanticProcessor<T extends Atom> {
```
flyingsilverfin: Fine we can keep the name, but lets make a class comment about the the scope of this class is... something like "Computes the unification between Atoms, and the difference between atoms unified via a unifier"

PR #5595, opened 2020-02-26
Abstract atom materialisation

kasper-piskorski:
To delegate the matarialisation tasks in `Atom`s to a new `AtomMaterialiser` entity so that the `Atom` implementations hold less responsibility and are less crowded - easier to maintain and understand.

We introduce an `AtomMaterialiser` interface with implementations for `IsaAtom`, `RelationAtom` and `AttributeAtom`, which allows to abstract and delegate the matarialisation procedures present in the corresponding `Atom`s.


graql/reasoner/atom/materialise/AttributeMaterialiser.java
```diff
@@ -0,0 +1,141 @@
...
+     */
+    private Relation putImplicitRelation(AttributeAtom atom, ConceptMap sub, Concept owner, Attribute attribute) {
+        ConceptMap answer = findAnswer(atom, sub);
+        if (answer == null) return attachAttribute(owner, attribute);
+        Variable relationVariable = atom.getRelationVariable();
+        return relationVariable.isReturned() ? answer.get(relationVariable).asRelation() : null;
+    }
+
+    @Override
+    public Stream<ConceptMap> materialise(AttributeAtom atom) {
```
flyingsilverfin: style - lets put public methods at the top so we see the intrface first thing :)

flyingsilverfin:
You think we'll be able to reduce the complexity of the `Atom` hierarchy after this set of PRs?

PR #5604, opened 2020-02-28
Cleanup atom validation

kasper-piskorski:
To simplify and further abstract atom validation and its implementations across `AtomValidator`s.

- moved `checkValid` implementation to `AtomValidator`
- removed `IdPredicate` dependency on `COnceptManager` - we only inject it when converting to value predicates
- removed validation on IdPredicates, instead we validate ids on query level
- removed `ResolutionPlan` dependency on `ConceptManager`


graql/reasoner/atom/predicate/IdPredicate.java
```diff
@@ -130,16 +118,15 @@ public String toString() {
     /**
      * @return corresponding value predicate if transformation exists (id corresponds to an attribute concept)
      */
-    public ValuePredicate toValuePredicate() {
+    public ValuePredicate toValuePredicate(ConceptManager conceptManager) {
```
flyingsilverfin: Great! Looks like we're removing state that we don't need and passing it in, direction we want to go 💃

PR #5612, opened 2020-03-02
Store labels instead of ids in atoms

kasper-piskorski:
To store `SchemaConcept` labels instead of their ids in atoms. 
Reasons: 
1) ids carry little information and require further manipulation (lookups) to extract their meaning
2) we often want to use the label which then requires doing a lookup via `ConceptManager`. 

As we trim down the number of dependencies on `ConceptManager`, this change also should allow to remove `ReasoningContext` out of atoms and instead inject it to functions that require it.

Instead of storing type `ConceptId`s in atoms, we store their `Label`s.

### DIFF
--- a/graql/reasoner/atom/Atom.java
+++ b/graql/reasoner/atom/Atom.java
@@ -72,6 +72,9 @@ public abstract class Atom extends AtomicBase {
 
     private final Label typeLabel;
 
+    /**
+     * NB: NULL typeLabel signals that the we don't know the specific non-meta type of this atom.
+     */
     public Atom(ReasonerQuery reasonerQuery, Variable varName, Statement pattern, @Nullable Label typeLabel, ReasoningContext ctx) {
         super(reasonerQuery, varName, pattern);
         this.ctx = ctx;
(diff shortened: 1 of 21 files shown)


graql/reasoner/atom/Atom.java
```diff
@@ -70,20 +70,20 @@
     private final AtomValidator<Atom> validator = new BasicAtomValidator();
     private Set<InferenceRule> applicableRules = null;
 
-    private final ConceptId typeId;
+    private final Label typeLabel;
 
-    public Atom(ReasonerQuery reasonerQuery, Variable varName, Statement pattern, ConceptId typeId, ReasoningContext ctx) {
+    public Atom(ReasonerQuery reasonerQuery, Variable varName, Statement pattern, @Nullable Label typeLabel, ReasoningContext ctx) {
```
flyingsilverfin: why is this nullable?

kasper-piskorski: `$x isa $type;` :)

kasper-piskorski: Basically if we do not know the non-meta type of the atom.

Issue #5596, opened 2020-02-26
Atom conversion error when using complex type for relation


jmsfltchr:
## Description

We encounter this scenario: 
For a relation `A`, where `A` has a subtype `B`, and there is a rule to infer that `A` instances can be cast to `B` as a complex type.

In this case, any query for the `A` or `B` relations fails with an error, even without and data instances present.

## Environment

1. OS (where Grakn server runs): Mac OS 10
2. Grakn version (and platform): Grakn Core 1.6.2
3. Grakn client: console

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Load this schema:
```
define

currency sub attribute,
    datatype string,
    plays currency-adoption_currency-adopted;

end-date sub attribute,
    datatype string;

currency-adoption sub relation,
    has end-date,
    relates currency-adoption_adopting-country,
    relates currency-adoption_currency-adopted;

current-currency sub currency-adoption,
    relates currency-adoption_adopting-country,
    relates currency-adoption_currency-adopted;

country-currently-uses-currency sub rule,
when {
    $ca isa currency-adoption;
    not {$ca has end-date $ed;};
}, then {
    $ca isa current-currency;
};
```
2. Run this query
```
match $x isa currency; $c($x); get;
```
3. Receive the error:
```
Error: UNKNOWN: Attempted illegal atom conversion of atom [current-currency($ca)] to type [class grakn.core.graql.reasoner.atom.binary.RelationAtom].. Please check server logs for the stack trace.
All uncommitted data is cleared
```

4. We even get the same error when querying for the non-complex type:
```
match $x isa currency; $c($x) isa currency-adoption; get;
```

## Expected Output

There should be no results as no data has been inserted. I don't see any integrity issue with adding this complex type, so I presume this behaviour is unexpected and a bug.

## Stack trace
```
2020-02-26 15:14:14,398 [transaction-listener] ERROR g.c.s.r.SessionService$TransactionListener - Runtime Exception in RPC TransactionListener:
grakn.core.kb.graql.reasoner.ReasonerException: Attempted illegal atom conversion of atom [current-currency($ca)] to type [class grakn.core.graql.reasoner.atom.binary.RelationAtom].
	at grakn.core.kb.graql.reasoner.ReasonerException.illegalAtomConversion(ReasonerException.java:71)
	at grakn.core.graql.reasoner.atom.Atom.toRelationAtom(Atom.java:91)
[...]

PR #5617, opened 2020-03-03
Improve validation and disallow more rule types

flyingsilverfin:
In the errors flagged in #5596 lead to the issues being fixed in this PR. We are more strict with rule validation, not allowing a rule to change the type of a concept if the meta type is not the same between the `when` and the `then` and not allowing setting a type via `isa` to be a relation, unless specifying role players.

* Implement better validation for rules, which reject newly rules with `isa` as the conclusion

### DIFF
--- a/common/exception/ErrorMessage.java
+++ b/common/exception/ErrorMessage.java
@@ -201,7 +201,6 @@ public enum ErrorMessage {
     INVALID_VARIABLE_PREDICATE_STATE("Invalid state in variable predicate [%s] with answer [%s]: either a concept is missing or not an attribute."),
     NO_ATOMS_SELECTED("No atoms were selected from the query [%s]."),
     INVALID_CACHE_ENTRY("Query cache entry for query [%s] contains an invalid entry: [%s]."),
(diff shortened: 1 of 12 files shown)


common/exception/ErrorMessage.java
```diff
@@ -194,8 +201,7 @@
     INVALID_VARIABLE_PREDICATE_STATE("Invalid state in variable predicate [%s] with answer [%s]: either a concept is missing or not an attribute."),
     NO_ATOMS_SELECTED("No atoms were selected from the query [%s]."),
     INVALID_CACHE_ENTRY("Query cache entry for query [%s] contains an invalid entry: [%s]."),
-    UNIFICATION_ATOM_INCOMPATIBILITY("Attempted unification on incompatible atoms."),
-    NON_EXISTENT_UNIFIER("Could not proceed with thr unification as the unifier doesn't exist."),
+    NON_EXISTENT_UNIFIER("Could not proceed with the unification as the unifier doesn't exist."),
```
kasper-piskorski: do we use it in the end :D?

flyingsilverfin: this is just a rename, got rid of the other one!

PR #5618, opened 2020-03-04
Flatten AttributeAtom

kasper-piskorski:
Instead of extending `Binary`, we make the `AttributeAtom` contain relevant `IsaAtom`s so that the `Atom` hierarchy is simpler and we enable further simplifications.
`AttributeAtom` now simply extends `Atom`. We store the binary information in relevant `IsaAtom`s which we use directly for attribute unification and equivalence.


graql/reasoner/atom/binary/AttributeAtom.java
```diff
@@ -208,10 +214,26 @@ public final boolean equals(Object obj) {
     }
 
     @Override
-    boolean isBaseEquivalent(Object obj){
-        if (!super.isBaseEquivalent(obj)) return false;
+    public boolean isAlphaEquivalent(Object obj) {
```
flyingsilverfin: why do we have new equivalences on AttributeAtom?

kasper-piskorski: We don't. We just call everything explicitly.

flyingsilverfin: ok cool, i think this is more readable :)

PR #5620, opened 2020-03-05
Flatten RelationAtom

kasper-piskorski:
Instead of extending `IsaAtom`, we make the `RelationAtom` contain the associated `IsaAtom`. As a result the Atom hierarchy is more flat and simple and we enable further simplifications.

`RelationAtom` now simply extends `Atom`. We store the binary information in a appropriate `IsaAtom` which we use directly for relation unification and equivalence.


graql/reasoner/atom/task/convert/RelationAtomConverter.java
```diff
@@ -84,7 +84,7 @@ public AttributeAtom toAttributeAtom(RelationAtom atom, ReasoningContext ctx) {
 
     @Override
     public IsaAtom toIsaAtom(RelationAtom atom, ReasoningContext ctx) {
-        IsaAtom isaAtom = IsaAtom.create(atom.getVarName(), atom.getPredicateVariable(), atom.getTypeLabel(), false, atom.getParentQuery(), ctx);
+        IsaAtom isaAtom = atom.isaAtom();
```
flyingsilverfin: Why can't we just return an isaAtom here? Converting a RelationAtom with an `isa` in it into an `isaAtom` can only contain the information that `X isa RelationType`, and nothing else, right?

PR #5635, opened 2020-03-09
Fix the equality comparison of AttributeAtoms with differing attribute variables

jmsfltchr:

We add a fix for the comparison of AttributeAtoms with different variables, but the same owner variable, and the same attribute type.

- Equality comparison of `AttributeAtoms` now considers the attributes` variables.
- Adds a test to check this works as expected

### DIFF
--- a/graql/reasoner/atom/binary/AttributeAtom.java
+++ b/graql/reasoner/atom/binary/AttributeAtom.java
@@ -251,7 +251,7 @@ public class AttributeAtom extends Atom{
     @Override
     public final int hashCode() {
         int hashCode = this.alphaEquivalenceHashCode();
-        hashCode = hashCode * 37 + this.getVarName().hashCode();
+        hashCode = hashCode * 37 + this.getVarName().hashCode() + this.getAttributeVariable().hashCode();
         return hashCode;
     }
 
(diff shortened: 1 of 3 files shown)


graql/reasoner/atom/binary/AttributeAtom.java
```diff
@@ -216,7 +216,9 @@ public final boolean equals(Object obj) {
         AttributeAtom a2 = (AttributeAtom) obj;
         return Objects.equals(this.getTypeLabel(), a2.getTypeLabel())
                 && this.getVarName().equals(a2.getVarName())
+                && this.getAttributeVariable().equals(a2.getAttributeVariable())
```
kasper-piskorski: We might want to update the hashCode as well. Currently we are not breaking the hashCode/equals contract (equal attributes have equal hashcodes, non-equal attributes can have equal hashcodes), however we could avoid doing extra `equals` check for cases when attribute values are different.

jmsfltchr:
Closing in favour of the duplicate [https://github.com/graknlabs/grakn/pull/5645/commits](https://github.com/graknlabs/grakn/pull/5645/commits)

PR #5636, opened 2020-03-09
Merge Query Execution Code Paths for Reasoning and Non-reasoning

flyingsilverfin:
This PR merges the entry point for reasoning and non-reasoning queries to use the type inference code, but then diverges in the `resolve()` method within `ReasonerQueryImpl`. This means we push the `infer=true/false` flag further down the stack, using a `ResolutionIterator` to resolve  with inferece versus `traverse(pattern)` to just read the DB. 

* Merge entry point for query execution to pass the flag directly to `resolve(boolean infer)`, cleaning up the split path in `QueryExecutorImpl`


flyingsilverfin:
Merge after rebasing off master AFTER #5641 is in.

PR #5640, opened 2020-03-10
Extract TraversalExecutor from QueryExecutor

flyingsilverfin:

Refactor `QueryExecutorImpl`, which is instantiated per-query with an `infer` flag. Interally it used to perform traversals on the backend, which was completely independent of inference. This PR extracts `TraversalExecutor` out of `QueryExecutor`, which has the side effect of removing `ExecutorFactory` from much of reasoner and only passing a `TraversalExecutor` directly. This makes more sense than before, as the semantics of holding a `ExecutorFactory` indicates reasoner might perform read, write, or compute queries where it really only needs to perform read traversals using the `TraversalFactory`.

* Split `QueryExecutor` into `QueryExecutor` and `TraversalExecutor`
* `TraversalExecutor` is a dedicated, lower level component that only consumes patterns and executes traversals on the backend without any knowledge of reasoning


graql/executor/TraversalExecutorImpl.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @Override
+    public Stream<ConceptMap> traverse(Conjunction<? extends Pattern> pattern) {
+        return traverse(pattern, traversalPlanFactory.createTraversal(pattern));
+    }
+
+    /**
+     * NB: doesn't depend on the infer flag, the result is always returned directly from the DB without application of rules
+     * TODO: seems out of place: make private or move elsewhere
```
kasper-piskorski: I understand we can remove the TODO now

Issue #5628, opened 2020-03-06
Direct Sub No Longer Works


flyingsilverfin:
## Description
Direct sub `sub!` when used as a Graql keyword no longer operates correctly.

## Environment
Grakn 1.6.2 upwards was detected, probably an issue before too.

## Reproducible Steps
1. `define person sub entity; child sub person; get;`
2. commit
3. `match $x sub! entity; get;`

## Expected Output
Expect to see only `entity` and `person`, but not `child`

## Actual Output
See all of`entity` and `person` AND `child`

flyingsilverfin:
Prerequisite for https://github.com/graknlabs/grakn/issues/4473, related to #4472

Issue #4473, opened 2018-10-12
Update reasoning to handle direct sub


kasper-piskorski:
Reasoner is missing representation for `sub!` property.

flyingsilverfin:
Complted with #5641

PR #5641, opened 2020-03-10
Fix and Implement Direct Sub

flyingsilverfin:
#5628 flags that direct sub is broken, and #4473 indicates reasoning is not able to handle direct Sub at all. This PR implements direct Sub in reasoner, and removes the reasoning code from the non-reasoning code path to fix the direct sub when reasoning is disabled.

* Add tests for direct sub with reasoning on and off
* Implement `OntologicalAtom` enum for direct sub
* Remove reasoning code from non-reasoning code path (`QueryExecutorImpl`)

### DIFF
--- a/test-integration/graql/executor/DirectSubIT.java
+++ b/test-integration/graql/executor/DirectSubIT.java
@@ -19,8 +19,6 @@
 package grakn.core.graql.executor;
 
 import grakn.core.concept.answer.ConceptMap;
-import grakn.core.kb.concept.api.Label;
-import grakn.core.kb.concept.api.SchemaConcept;
 import grakn.core.kb.server.Session;
 import grakn.core.kb.server.Transaction;
 import grakn.core.rule.GraknTestServer;
--- a/test-integration/graql/executor/DirectSubIT.java
+++ b/test-integration/graql/executor/DirectSubIT.java
@@ -65,7 +65,17 @@ public class DirectSubIT {
     }
 
     @Test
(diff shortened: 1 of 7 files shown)


test-integration/graql/executor/DirectSubIT.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Test
+    public void directSubReturnsOnlyItselfAndDirectChildTypesWithoutReasoning() {
+        tx.execute(Graql.parse("define person sub entity; child sub person;").asDefine());
+        tx.commit();
+
+        tx = session.writeTransaction();
+        List<ConceptMap> directSubs = tx.execute(Graql.match(Graql.var("x").subX("entity")).get(), false);
+        assertEquals(2, directSubs.size());
```
kasper-piskorski: maybe we could make it more explicit: check it doesn't contain `child`.

test-integration/graql/executor/DirectSubIT.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Test
+    public void directSubReturnsOnlyItselfAndDirectChildTypesWithReasoning() {
+        tx.execute(Graql.parse("define person sub entity; child sub person;").asDefine());
+        tx.commit();
+
+        tx = session.writeTransaction();
+        List<ConceptMap> directSubs = tx.execute(Graql.match(Graql.var("x").subX("entity")).get());
+        assertEquals(2, directSubs.size());
```
kasper-piskorski: ditto

PR #5645, opened 2020-03-11
Fix the equality comparison of AttributeAtoms with differing attribute variables 2

jmsfltchr:

We add a fix for the comparison of AttributeAtoms with different variables, but the same owner variable, and the same attribute type.

- Equality comparison of `AttributeAtoms` now considers the attributes` variables.
- Adds a test to check this works as expected

### DIFF
--- a/graql/reasoner/atom/binary/AttributeAtom.java
+++ b/graql/reasoner/atom/binary/AttributeAtom.java
@@ -251,7 +251,8 @@ public class AttributeAtom extends Atom{
     @Override
     public final int hashCode() {
         int hashCode = this.alphaEquivalenceHashCode();
-        hashCode = hashCode * 37 + this.getVarName().hashCode() + this.getAttributeVariable().hashCode();
+        hashCode = hashCode * 37 + this.getVarName().hashCode();
+        hashCode = hashCode * 37 + this.getAttributeVariable().hashCode();
         return hashCode;
     }
 
--- a/graql/reasoner/atom/binary/AttributeAtom.java
+++ b/graql/reasoner/atom/binary/AttributeAtom.java
@@ -251,7 +251,7 @@ public class AttributeAtom extends Atom{
     @Override
     public final int hashCode() {
         int hashCode = this.alphaEquivalenceHashCode();
-        hashCode = hashCode * 37 + this.getVarName().hashCode();
+        hashCode = hashCode * 37 + this.getVarName().hashCode() + this.getAttributeVariable().hashCode();
         return hashCode;
     }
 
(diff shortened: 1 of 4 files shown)


graql/reasoner/atom/binary/AttributeAtom.java
```diff
@@ -249,7 +251,7 @@ private boolean multiPredicateEqual(AttributeAtom that){
     @Override
     public final int hashCode() {
         int hashCode = this.alphaEquivalenceHashCode();
-        hashCode = hashCode * 37 + this.getVarName().hashCode();
+        hashCode = hashCode * 37 + this.getVarName().hashCode() + this.getAttributeVariable().hashCode();
```
kasper-piskorski: Could we follow the common hashing approach and make it something like:
```
hashCode = hashCode * 37 + this.getVarName().hashCode();
hashCode = hashCode * 37 + this.getAttributeVariable().hashCode();
```

jmsfltchr: Ah I see, changed

graql/reasoner/atom/binary/AttributeAtom.java
```diff
@@ -250,6 +252,7 @@ private boolean multiPredicateEqual(AttributeAtom that){
     public final int hashCode() {
         int hashCode = this.alphaEquivalenceHashCode();
         hashCode = hashCode * 37 + this.getVarName().hashCode();
+        hashCode = hashCode * 37 + this.getAttributeVariable().hashCode();
         return hashCode;
```
haikalpribadi: 1) The hashed objects do not match the objects in the equality function? Are we sure this is correct? This is currently breaking the equality-and-hash contract.

2) In most (if not all) cases, Java's `Objects.hash(...)` is sufficient and the simplest way to go. It just can't take arrays, and so you wrap them in `Arrays.hashCode(...)` before you pass them in. You can also pass in other hashcodes, as they are just integers - which are their own hashcodes. For the code above, it would like the following (But this still breaks the equals-hash-contract as the objects are not the same as what's being tested in `equals()`).
```
return Objects.hash(this.alphaEquivalenceHashCode(), this.getVarName(), this.getAttributeVariable().hashCode())
```

haikalpribadi: You can find more details [here](objects that are equal to each other must return the same hashCode), but ultimately:

> _"objects that are equal to each other must return the same hashCode"_

Which often implies that the equality function should check variables that are also used to produce the hashCode.

haikalpribadi: You can find more details [here](https://www.baeldung.com/java-equals-hashcode-contracts), but ultimately:

> _"objects that are equal to each other must return the same hashCode"_

Which often implies that the equality function should check variables that are also used to produce the hashCode.

kasper-piskorski: The compared objects in hash and equals do match. We compare on `varName` `attributeVariable`, `multiPredicate` and `typeLabel`. Also not having them matched doesn't break the contract - hashcode can be more lenient than equals - two things that are not equal can have the same hash code.

haikalpribadi: yep, that's true. I just wanted to make sure we're aware they're not identical, but maybe that's the closest they can be.

Issue #5647, opened 2020-02-29
A query using rules returns empty output instead of the expected one non-deterministically


MajaTr:
## Description

Setting a rule which establishes a relation `equal` between two entities if their labels are equal, and then asking for pairs of entities for which the `equal` relation holds (where the entities are not quantified as their entity type, but as whether an equivalent unary relation holds for them), sometimes results in an empty output.

## Environment

1. OS (where Grakn server runs): Ubuntu 16.04
2. Grakn version (and platform): Grakn Core 1.6.2
3. Grakn client: At least client-java and console.

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Run
```
define label sub attribute, datatype string;
define edge sub relation, relates source, relates destination;
define vertex sub entity, has label, plays source, plays destination;

insert $v isa vertex, has label "0";
match
  $s isa vertex, has label "0";
  $t isa vertex, has label "0";
insert $e (source: $s, destination: $t) isa edge;

define hasedge sub relation, relates v;
define hasedge-rule sub rule, 
when {
  $_ (source: $x, destination: $y) isa edge; 
}, then {
  $_ (v: $x) isa hasedge; 
};

define equal sub relation, relates l, relates r;
define vertex sub entity, plays l, plays r;
define equal-rule sub rule, when { 
  $v1 isa vertex; 
  $v2 isa vertex; 
  $v1 has label $l1; 
  $v2 has label $l2; 
  $l1 == $l2;
}, then { 
  $_ (l: $v1, r: $v2) isa equal; 
};

```

Or run in Java

```
GraknClient.Transaction writeTransaction = session.transaction().write();
ArrayList<GraqlQuery> writeQueries = new ArrayList<>();

//Define a graph schema
writeQueries.add(Graql.define(type("label").sub("attribute").datatype("string")));
writeQueries.add(Graql.define(type("edge").sub("relation").relates("source").relates("destination")));
writeQueries.add(Graql.define(type("vertex").sub("entity").has("label").plays("source").plays("destination")));

//Add a vertex and an edge
writeQueries.add(
        Graql.insert(var("v").isa("vertex").has("label", "0"))
);

writeQueries.add(Graql.match(
        var("s").isa("vertex").has("label", "0"),
        var("t").isa("vertex").has("label", "0")
).insert(
        var("e").isa("edge").rel("source", "s").rel("destination", "t")
));

//Add a unary relation that holds iff the argument is a vertex
[...]

PR #5650, opened 2020-03-11
Add stricter rule validation for unplayable roles

flyingsilverfin:
Validate rules more strictly, banning all occurences of rules that infer a role that can never be played for any possible type that the role player variable may be.

* Add stricter rule validation for each role that is mentioned in the `then` of a rule (the head)
* Add a test that validates this throws when this is violated
* Fix client-java-E2E, which had an error in its schema flagged by this stronger validation

### DIFF
--- a/test-end-to-end/client-java/ClientJavaE2E.java
+++ b/test-end-to-end/client-java/ClientJavaE2E.java
@@ -140,7 +140,7 @@ public class ClientJavaE2E {
                     type("parentship").sub("relation").relates("parent").relates("child"),
 
                     type("name").sub("attribute").datatype(Graql.Token.DataType.STRING),
-                    type("lion").sub("entity").has("name").plays("male-partner").plays("female-partner").plays("offspring")
+                    type("lion").sub("entity").has("name").plays("male-partner").plays("female-partner").plays("offspring").plays("parent").plays("child")
             );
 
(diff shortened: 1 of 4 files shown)


test-integration/server/RuleValidationIT.java
```diff
@@ -869,6 +869,27 @@ public void whenAddingDuplicateRulesOfTheSameTypeWithDifferentPatternVariables_R
         }
     }
 
+    @Test
+    public void whenCreatingRuleWithImpossibleRole_rejectRule() {
```
kasper-piskorski: what are we testing here? shouldn't we expect an exception?

flyingsilverfin:
Would close #5647

test-end-to-end/client-java/ClientJavaE2E.java
```diff
@@ -140,7 +140,7 @@ public void clientJavaE2E() {
                     type("parentship").sub("relation").relates("parent").relates("child"),
 
                     type("name").sub("attribute").datatype(Graql.Token.DataType.STRING),
-                    type("lion").sub("entity").has("name").plays("male-partner").plays("female-partner").plays("offspring")
```
kasper-piskorski: hah! did we catch an incomplete schema?

Issue #5651, opened 2020-03-11
Transitive inference doesn't provide full explanation


jmsfltchr:
## Description

A simple case of a transitive rule only gives one layer of explanation, although this is non-deterministic, is seems to occur on the majority of attempts.

## Environment

1. OS (where Grakn server runs): Mac OS 10
2. Grakn version (and platform): Grakn Core 1.6.2
3. Grakn client: client-java

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Load this schema:
```
define
name sub attribute,
    datatype string;

location sub entity,
    abstract,
    key name,
    plays location-hierarchy_superior,
    plays location-hierarchy_subordinate;

area sub location;
city sub location;
country sub location;
continent sub location;

location-hierarchy sub relation,
    relates location-hierarchy_superior,
    relates location-hierarchy_subordinate;

location-hierarchy-transitivity sub rule,
when {
    $lh1(location-hierarchy_superior: $a, location-hierarchy_subordinate: $b) isa location-hierarchy;
    $lh2(location-hierarchy_superior: $b, location-hierarchy_subordinate: $c) isa location-hierarchy;
}, then {
    (location-hierarchy_superior: $a, location-hierarchy_subordinate: $c) isa location-hierarchy;
};
```
2. Load this data:
```
insert
$ar isa area, has name "King's Cross";
$cit isa city, has name "London";
$cntry isa country, has name "UK";
$cont isa continent, has name "Europe";
(location-hierarchy_superior: $cont, location-hierarchy_subordinate: $cntry) isa location-hierarchy;
(location-hierarchy_superior: $cntry, location-hierarchy_subordinate: $cit) isa location-hierarchy;
(location-hierarchy_superior: $cit, location-hierarchy_subordinate: $ar) isa location-hierarchy;
```
3. Execute this query using client-java and examine the explanation:
```
match
$ar isa area, has name $ar-name;
$cont isa continent, has name $cont-name;
$lh(location-hierarchy_superior: $cont, location-hierarchy_subordinate: $ar) isa location-hierarchy;
get;
```

## Expected Output

We expect that there should be 2 nested explanations for this transitive example.

## Actual Output

often we see only one explanation, and all of the answers within it have `.hasExplanation()` `false`

flyingsilverfin:
This is solved as of https://github.com/graknlabs/grakn/pull/5764

PR #5654, opened 2020-03-13
Find Missing Explanations

flyingsilverfin:
Re-querying for explanations on the client side sometimes returns the wrong or missing explanation. This is reproduced in #5651 .

* Search child queries in query cache as well as parents when answering a query


flyingsilverfin:
Thankfully, this has been superceded by https://github.com/graknlabs/grakn/pull/5764

PR #5657, opened 2020-03-16
Include rule in explanation

jmsfltchr:

When returning an explanation, we now return the rule from which the explanation is derived.

Requires:
[https://github.com/graknlabs/protocol/pull/24](https://github.com/graknlabs/protocol/pull/24)
[https://github.com/graknlabs/client-java/pull/81](https://github.com/graknlabs/client-java/pull/81)

- Contruct Explanations by passing a rule rather than a rule id
  - Update other tests 
- Nest a rule in an explanation in an explanation grpc response
  - Include a test for this
- Update deps

### DIFF
--- a/dependencies/graknlabs/dependencies.bzl
+++ b/dependencies/graknlabs/dependencies.bzl
@@ -49,7 +49,7 @@ def graknlabs_client_java():
     git_repository(
         name = "graknlabs_client_java",
         remote = "https://github.com/jmsfltchr/client-java",
-        commit = "d507ddb876b600355751b447beb71479371d2f5d",
+        commit = "cdbc13fe0d0d61f22298b9921e6620dd2127533b",
     )
 
 def graknlabs_console():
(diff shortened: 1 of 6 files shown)


haikalpribadi:
I will leave this to @flyingsilverfin to review, and I can approve afterwards. My code-ownership-review is only triggered due to changes in the dependency file. But I've also skimmed through the code at a high-level, and it looks okay.

dependencies/graknlabs/dependencies.bzl
```diff
@@ -41,15 +41,15 @@ def graknlabs_graql():
...
+        remote = "https://github.com/jmsfltchr/protocol",
+        commit = "87c5b205ab7b8747e05e9310cc8e64bdc34c41bc", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_protocol
     )
 
 def graknlabs_client_java():
     git_repository(
         name = "graknlabs_client_java",
-        remote = "https://github.com/graknlabs/client-java",
-        commit = "c37f67a6a95e93d9647403b08aa3421cc047e1a7",
+        remote = "https://github.com/jmsfltchr/client-java",
```
flyingsilverfin, line 51: don't forget to revert these before merging!

PR #5661, opened 2020-03-18
Rule validate duplicate role players are distinct

### LINKED ISSUE #5649: Inconsistent reasoner behaviour on reflexive relations
## Description

Data inserted by rules does not adhere to same restrictions as data inserted directly (specifically the same entity is allowed to play the same role twice in a relation). Data 'inserted' by rules seems to not adhere to some restrictions placed on regular inserts, specifically the same entity can play the same role twice in any given relation.

## Environment

1. OS (where Grakn server runs): Ubuntu 18.04.4 LTS
2. Grakn version (and platform): Grakn (core?) 1.6.2 ["grakn server version" just says '1.6.2']
3. Grakn client: console
4. Other environment details: fresh/clean keyspace

## Reproducible Steps

1. Start grakn console

2. Paste the following into console

````
define name sub attribute, datatype string;
define friendship sub relation, relates friend;
define person sub entity, has name, plays friend;
define friendship-transitive sub rule, when { (friend: $f1, friend: $f2) isa friendship;(friend: $f2, friend: $f3) isa friendship;}, then {  (friend: $f1, friend: $f3) isa friendship;};
insert $x isa person, has name "Alice";
insert $x isa person, has name "Bob";
insert $x isa person, has name "Charlie";
match $alice isa person, has name "Alice"; $bob isa person, has name "Bob"; insert (friend:$alice,friend:$bob) isa friendship;
match $alice isa person, has name "Alice"; $charlie isa person, has name "Charlie"; insert (friend:$alice,friend:$charlie) isa friendship;
commit
match (friend: $f1, friend: $f2) isa friendship; $f1 has name $f1N; $f2 has name $f2N; get $f1N, $f2N;
````

## Expected Output

I would expect to see
A-B,
A-C,
B-A,
B-C,
C-A,
C-B
as the only 6 results

## Actual Output

I see the above 6, but additionally I see
A-A,
B-B,
C-C
as another 3 results, despite that being one entity playing the "friend" role twice in "friendship", which can't be achieved by directly inserting

### LINKED ISSUE #5658: Roleplayers of a single role are not enforced to be distinct
## Description

It is currently possible to write patterns where either explicitly or implicitly there is a duplicate definition for a (role, roleplayer) tuple in a relation, and they can be executed as if they are valid.

Related to #5649

## Environment

2. Grakn version (and platform): Grakn Core 1.6.2
3. Grakn client: console

## Reproducible Steps

```
case1> match $x sub thing; get;
{$x type thing;}
{$x type relation;}
{$x type entity;}
{$x type attribute;}
case1> define person sub entity, plays friend; friendship sub relation, relates friend;
{}
case1> commit
case1> insert $x isa person;
{$x id V8200 isa person;}
case1> commit
case1> match $x isa person; insert $f(friend: $x, friend: $x) isa friendship;
{$f id V12352 (friend: id V8200) isa friendship; $x id V8200 isa person;}
case1> commit
case1> match $x isa person; $y isa person; insert $f(friend: $x, friend: $y) isa friendship;
{$f id V8320 (friend: id V8200) isa friendship; $x id V8200 isa person; $y id V8200 isa person;}
case1> commit
case1> match (friend: $f1, friend: $f2) isa friendship; get;
case1> match $f(friend: $x, friend: $x) isa friendship; get;
case1> match $f(friend: $x) isa friendship; get;
{$f id V8320 (friend: id V8200) isa friendship; $x id V8200 isa person;}
{$f id V12352 (friend: id V8200) isa friendship; $x id V8200 isa person;}
case1> match $f(friend: $x) isa friendship; $f(friend: $x) isa friendship; get;
{$f id V8320 (friend: id V8200) isa friendship; $x id V8200 isa person;}
{$f id V12352 (friend: id V8200) isa friendship; $x id V8200 isa person;}
case1>
```

## Expected Output vs. Desired Output
From discussion with @haikalpribadi and @flyingsilverfin, we think that:
- `match $x isa person; insert $f(friend: $x, friend: $x) isa friendship;` should throw an error, ideally at parse-time.

- `match $x isa person; $y isa person; insert $f(friend: $x, friend: $y) isa friendship;` should throw an error at runtime, once it has been established that `$x` and `$y` are equivalent.

- `match (friend: $f1, friend: $f2) isa friendship; get;` is currently correct in returning no results here

- `match $f(friend: $x, friend: $x) isa friendship; get;` should throw an error at parse-time

### PR DESCRIPTION
**Pull request by flyingsilverfin**
In order to fix #5658 and #5649 , one required step is to remove the ability to infer relations that can have duplicate role players of the same role in the same relation. We do this by checking whether the rule can _possibly_ infer (as given in the `then`) a duplicate role being played by the same variable, and reject the rule at validation time if the required `!=` are missing. This is over-strict in some cases (eg. if attribute values are given making the role players always distinct in the rule `when`).

Closes #5649 
* add extra rule validation to ensure that duplicate role players that may be inferred can not be identical
* tests for the above

### DIFF
--- a/graql/reasoner/atom/task/validate/RelationAtomValidator.java
+++ b/graql/reasoner/atom/task/validate/RelationAtomValidator.java
@@ -73,7 +73,7 @@ public class RelationAtomValidator implements AtomValidator<RelationAtom> {
         return Sets.union(
                 basicValidator.validateAsRuleHead(atom, rule, ctx),
                 validateRelationPlayers(atom, rule, ctx),
-                validateDuplicateInferredRolesMustBeDifferentConcepts(atom, ctx);
+                validateDuplicateInferredRolesMustBeDifferentConcepts(atom, rule, ctx);
     }
 
     @Override
--- a/graql/reasoner/atom/task/validate/RelationAtomValidator.java
+++ b/graql/reasoner/atom/task/validate/RelationAtomValidator.java
@@ -19,6 +19,7 @@
 package grakn.core.graql.reasoner.atom.task.validate;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.SetMultimap;
 import com.google.common.collect.Sets;
 import grakn.core.common.exception.ErrorMessage;
@@ -27,6 +28,8 @@ import grakn.core.core.Schema;
 import grakn.core.graql.reasoner.ReasoningContext;
 import grakn.core.graql.reasoner.atom.binary.IsaAtom;
 import grakn.core.graql.reasoner.atom.binary.RelationAtom;
+import grakn.core.graql.reasoner.atom.predicate.NeqIdPredicate;
+import grakn.core.graql.reasoner.query.ReasonerQueryImpl;
 import grakn.core.kb.concept.api.Label;
 import grakn.core.kb.concept.api.Role;
 import grakn.core.kb.concept.api.Rule;
@@ -34,21 +37,23 @@ import grakn.core.kb.concept.api.SchemaConcept;
 import grakn.core.kb.concept.api.Type;
 import grakn.core.kb.concept.manager.ConceptManager;
 import grakn.core.kb.graql.exception.GraqlSemanticException;
-import graql.lang.property.PlaysProperty;
+import graql.lang.pattern.Conjunction;
 import graql.lang.property.RelationProperty;
 import graql.lang.statement.Statement;
 import graql.lang.statement.Variable;
 
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 public class RelationAtomValidator implements AtomValidator<RelationAtom> {
 
     private final BasicAtomValidator basicValidator;
 
-    public RelationAtomValidator(){
+    public RelationAtomValidator() {
         this.basicValidator = new BasicAtomValidator();
     }
 
@@ -67,7 +72,8 @@ public class RelationAtomValidator implements AtomValidator<RelationAtom> {
         //can form a rule head if type is specified, type is not implicit and all relation players are insertable
         return Sets.union(
                 basicValidator.validateAsRuleHead(atom, rule, ctx),
-                validateRelationPlayers(atom, rule, ctx));
+                validateRelationPlayers(atom, rule, ctx),
+                validateDuplicateInferredRolesMustBeDifferentConcepts(atom, ctx);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: server/ValidateGlobalRules.java [hunk 1]
```diff
@@ -427,4 +436,44 @@ private static ReasonerQuery combinedRuleQuery(ReasonerQueryFactory reasonerQuer
...
+     * We have an over-strict defensive check we enforce for now, to ensure that a rule will not throw at runtime
+     * Requiring a '!=' between variables that may result in an inferred relation with duplicate role player edges.
+     * For example:
+     *
+     *
+     * @param headQuery
+     * @param bodyQuery
+     * @return
+     */
+    private static Set<String> validateDuplicateInferredRolesMustBeDifferentConcepts(ReasonerQuery headQuery, ReasonerQuery bodyQuery) {
```
kasper-piskorski: it might be better to put this check in `RelationAtomValidator` and call it in `validateAsRuleHead`

server/ValidateGlobalRules.java
```diff
@@ -427,4 +436,44 @@ private static ReasonerQuery combinedRuleQuery(ReasonerQueryFactory reasonerQuer
...
+                        Set<Variable> variables = new HashSet<>(roleVarMap.get(duplicateRole));
+                        Set<List<Variable>> variablePairsWithDuplicates = Sets.cartesianProduct(variables, variables);
+                        // we can remove all the duplicate pairs because (X,Y) and (Y,X) only need to be checked once
+                        // also remove pairs (X,X), these are over-generated
+                        Set<Set<Variable>> pairs = variablePairsWithDuplicates.stream().map(HashSet::new).filter(pair -> pair.size() > 1).collect(Collectors.toSet());
+                        for (Set<Variable> pair : pairs) {
+                            boolean requiredInequalityFound = neqAssertions.stream()
+                                    .anyMatch(neq -> neq.getVarNames().equals(pair));
+                            if (!requiredInequalityFound) {
+                                errors.add("Rule that may infer duplicate roles played by same concept is missing != check between variables: " + pair.toString());
```
kasper-piskorski: we could define a separate `ErrorMessage` for this scenario

graql/reasoner/atom/task/validate/RelationAtomValidator.java
```diff
@@ -65,9 +70,10 @@ public void checkValid(RelationAtom atom, ReasoningContext ctx) {
     @Override
     public Set<String> validateAsRuleHead(RelationAtom atom, Rule rule, ReasoningContext ctx) {
         //can form a rule head if type is specified, type is not implicit and all relation players are insertable
-        return Sets.union(
+        Set<String> errors = Sets.union(
```
kasper-piskorski: ah so that's the bit when you want to have union with 3 sets. It might be clearer to just initialise a set to empty and then do `.addAll` one by one. You would immediately see that you have a union of 3 things.

flyingsilverfin:
We have decided that instead of forcing the rules to be overly strict, and throwing an exception when a duplicate role player is added, that the more logical implementation is to actually remove the check at all and add a duplicate role player. This means this PR, which neither Kasper nor I liked very much as it is very strict and unintuitive!

PR #5666, opened 2020-03-24
Role Subtype Unification Test

flyingsilverfin:
An issue highlighted when debugging #5651 via #5654 , we have a corner case in Graql that has to do with role subtyping in relations. Specifically the unification in cases where we have variable roles with types indicated separately. This test highlights the followign case:
```
child = { ($r1 : $a, $r2: $b); $r1 type baseRole1; };

parent = { ($r1: $x, $r2: $y); $r1 type role;};
```

These queries are NOT in any kind of subsumption relation, because we bind `$r1` to a specific type and inheritance mechanisms are not triggered. It is also NOT a structural unification because $r1 is inferred to be a type and replaced, meaning the structure is not the same. In this interesting case, we DO have  a STRUCTURAL_SUBSUMPTIVE unifier, as replacing $r1 with the inferred role ID leads to satisfying the more relaxed structural_subsumptive condition that IDs are not required to match.

* Write a test for specifically this case, with listing the expected outcomes of each type of unification. 
* Fix one failing case where we expect the RULE unifier to fail when it doesn't, because we did not check the role variable IDs when they were present
* Skip checking role type compatibility if there are also IDs present for the role variables

### DIFF
--- a/test-integration/graql/reasoner/query/AtomicQueryUnificationIT.java
+++ b/test-integration/graql/reasoner/query/AtomicQueryUnificationIT.java
@@ -555,15 +555,12 @@ public class AtomicQueryUnificationIT {
             unification(query, potentialEquivalent, true, UnifierType.RULE, testTx);
             unification(potentialEquivalent, query, true, UnifierType.RULE, testTx);
 
-            // we block subtype checking by tying the type of $role in the potentialEquivalent
+            // 'query' has answers that are not a subset of 'potentialEquivalent' because of the extra returned variable
             unification(query, potentialEquivalent, false, UnifierType.SUBSUMPTIVE, testTx);
             unification(potentialEquivalent, query, true, UnifierType.SUBSUMPTIVE, testTx);
         }
     }
 
-    /**
-     * An example of a test where we have different levels of relaxed-ness of the unifiers
-     */
     @Test
     public void testUnification_RelationRolesNotIdentical() {
         try (Transaction tx = genericSchemaSession.writeTransaction()) {
@@ -584,7 +581,7 @@ public class AtomicQueryUnificationIT {
             // structural_subsumptive is the most relaxed unifier - we don't care about IDs if they don't exist
             // and if they do, then they don't have to match
             unification(query, potentiallyParent, true, UnifierType.STRUCTURAL_SUBSUMPTIVE, testTx);
-            unification(potentiallyParent, query, false, UnifierType.STRUCTURAL_SUBSUMPTIVE, testTx);
+            unification(potentiallyParent, query, true, UnifierType.STRUCTURAL_SUBSUMPTIVE, testTx);
         }
     }
 
--- a/test-integration/graql/reasoner/query/AtomicQueryUnificationIT.java
+++ b/test-integration/graql/reasoner/query/AtomicQueryUnificationIT.java
@@ -571,14 +571,20 @@ public class AtomicQueryUnificationIT {
             String query = "{ ($r1 : $a, $r2: $b); $r1 type baseRole1; };";
             String potentiallyParent = "{ ($r1: $x, $r2: $y); $r1 type role;};";
             unification(query, potentiallyParent, false, UnifierType.EXACT, testTx);
+
             unification(query, potentiallyParent, false, UnifierType.RULE, testTx);
+            unification(potentiallyParent, query, false, UnifierType.RULE, testTx);
             // structural requires IDs to be present in all the same places, but may not have to match
             // this is false because of role inference - we rewrite $r1 to be specific role
             unification(query, potentiallyParent, false, UnifierType.STRUCTURAL, testTx);
+            unification(potentiallyParent, query, false, UnifierType.STRUCTURAL, testTx);
+
             unification(query, potentiallyParent, false, UnifierType.SUBSUMPTIVE, testTx);
+            unification(potentiallyParent, query, false, UnifierType.SUBSUMPTIVE, testTx);
             // structural_subsumptive is the most relaxed unifier - we don't care about IDs if they don't exist
             // and if they do, then they don't have to match
             unification(query, potentiallyParent, true, UnifierType.STRUCTURAL_SUBSUMPTIVE, testTx);
+            unification(potentiallyParent, query, false, UnifierType.STRUCTURAL_SUBSUMPTIVE, testTx);
         }
     }
 
--- a/test-integration/graql/reasoner/query/AtomicQueryUnificationIT.java
+++ b/test-integration/graql/reasoner/query/AtomicQueryUnificationIT.java
@@ -555,7 +555,8 @@ public class AtomicQueryUnificationIT {
             unification(query, potentialEquivalent, true, UnifierType.RULE, testTx);
             unification(potentialEquivalent, query, true, UnifierType.RULE, testTx);
 
-            unification(query, potentialEquivalent, true, UnifierType.SUBSUMPTIVE, testTx);
+            // we block subtype checking by tying the type of $role in the potentialEquivalent
+            unification(query, potentialEquivalent, false, UnifierType.SUBSUMPTIVE, testTx);
             unification(potentialEquivalent, query, true, UnifierType.SUBSUMPTIVE, testTx);
         }
     }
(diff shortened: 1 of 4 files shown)


test-integration/graql/reasoner/query/AtomicQueryUnificationIT.java
```diff
@@ -555,11 +555,39 @@ public void testUnification_RelationsWithVariableRolesAndPotentialTypes() {
             unification(query, potentialEquivalent, true, UnifierType.RULE, testTx);
             unification(potentialEquivalent, query, true, UnifierType.RULE, testTx);
 
-            unification(query, potentialEquivalent, true, UnifierType.SUBSUMPTIVE, testTx);
+            // we block subtype checking by tying the type of $role in the potentialEquivalent
+            unification(query, potentialEquivalent, false, UnifierType.SUBSUMPTIVE, testTx);
```
jmsfltchr, line 559: I don't follow how blocking subtype checking matters - isn't the unification subsumptive if the two would share the exact same set of answers? Which, in this case, it seems they will. I must be missing something.

flyingsilverfin: Key point here is that `query` doesn't have answers that are a subtype of `potentialQuery`, but only the other way around. 
This is because of key point 2: by specifying the type of `$role` ie `$role type baseRole1`, we are preventing the inheritance mechanism from kicking in (exemplifies the edge cases i was talking about reasoner has to handle!)

So, to recap, all the answers of `query` include the answers to `potentialEquivalent`. However, the reverse is not true (see line below for other direction)

jmsfltchr: Aha - in `query` `baseRole1` can be itself or any of its subtypes, whereas in `potentialEquivalent`, `$role1` is constrained to be exactly `baseRole1` and not any of its subtypes. The new test also makes sense given this.

flyingsilverfin:
Further investigation into the behaviors of roles and variable roles has been prompted after some discussion.
Here's some current (Master, 1.6.2) behavior (no rules, so reasoner is off. However, reasoner may still rewrite some of the terms):

```
grakn> define super-rel sub relation, relates super-role; p sub entity, plays super-role;
{}
grakn> define sub-rel sub relation, relates sub-role as super-role; sub-p sub p, plays sub-role;
{}
grakn> insert (super-role: $p) isa super-rel; $p isa p;
{$p id V16608 isa p;}
grakn> insert (sub-role: $q) isa sub-rel; $q isa sub-p;
{$q id V12368 isa sub-p;}
grakn> commit

# 1
grakn> match $m ($r: $x) isa relation; get;
{$r type super-role sub role; $x id V16608 isa p; $m id V8224 (super-role: id V16608) isa super-rel;}
{$r type role; $x id V16608 isa p; $m id V8224 (super-role: id V16608) isa super-rel;}
{$r type sub-role sub super-role; $x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}
{$r type super-role sub role; $x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}
{$r type role; $x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}

# 2
grakn> match $m (role: $x) isa relation ;get;
{$x id V16608 isa p; $m id V8224 (super-role: id V16608) isa super-rel;}
{$x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}

# 3
grakn> match $m (super-role: $x) isa relation; get;
{$x id V16608 isa p; $m id V8224 (super-role: id V16608) isa super-rel;}
{$x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}

# 4
grakn> match $m (sub-role: $x) isa relation; get;
{$x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}

# 5
grakn> match $m ($r: $x) isa relation; $r type role; get;
{$r type role; $x id V16608 isa p; $m id V8224 (super-role: id V16608) isa super-rel;}
{$r type role; $x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}

# 6
grakn> match $m ($r: $x) isa relation; $r type super-role; get;
{$r type super-role sub role; $x id V16608 isa p; $m id V8224 (super-role: id V16608) isa super-rel;}
{$r type super-role sub role; $x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}

# 7
grakn> match $m ($r: $x) isa relation; $r type sub-role; get;
{$r type sub-role sub super-role; $x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}

# 8
grakn> match $m ($r: $x) isa relation; $r sub role; get;
{$r type super-role sub role; $x id V16608 isa p; $m id V8224 (super-role: id V16608) isa super-rel;}
{$r type role; $x id V16608 isa p; $m id V8224 (super-role: id V16608) isa super-rel;}
{$r type sub-role sub super-role; $x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}
{$r type super-role sub role; $x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}
{$r type role; $x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}

# 9
grakn> match $m ($r: $x) isa relation; $r sub super-role; get;
{$r type super-role sub role; $x id V16608 isa p; $m id V8224 (super-role: id V16608) isa super-rel;}
{$r type role; $x id V16608 isa p; $m id V8224 (super-role: id V16608) isa super-rel;}
{$r type sub-role sub super-role; $x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}
{$r type super-role sub role; $x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}
{$r type role; $x id V12368 isa sub-p; $m id V4272 (sub-role: id V12368) isa sub-rel;}

# 10
grakn> match $m ($r: $x) isa relation; $r sub sub-role; get;
Error: INVALID_ARGUMENT: GraqlSemanticException-the concept [x] could not be found in results. Please check server logs for the stack trace.
```

In all examples, the inheritance mechanism is never "blocked" by giving the type of a role using a variable (which is one acceptable behavior, if consistent). This clashes with the understanding on which this PR is written and the tests are updated!

Evidently, I misunderstood or dont have a complete grasp of what Graql does or is supposed to do. Clearly, some things are broken as well (see query #10 above). 

I propose investigating the current and expected behaviors further and handling the, along with this PR, fully in the future.

flyingsilverfin:
TODO:

New test to be added to `VariableRolesIT` to investigate further breakages in a separate PR that should precede this one
```
public void providingRoleTypeUsingSub() {
    try (Transaction tx = variableRoleSession.writeTransaction()) {
        String query = "match $rel ($role: $player) isa binary-base; $role sub role; get;";
        List<ConceptMap> answers = tx.execute(Graql.parse(query).asGet());  
        System.out.println(answers);
// some assertion
    }
}
```
```

**Comment by flyingsilverfin** [comment]
@haikalpribadi  @jmsfltchr @kasper-piskorski please see above two comments!

**Comment by flyingsilverfin** [comment]
Discussion on the above has led to the conclusion that:
```
match $x isa $t; $t type person; get;
```

_should_ trigger the inheritance mechanism in $x, because instances of `man` and `woman` and `person` , because all are of type `person`.

similarly:
```
match ($r: $x) isa relation; $r type arole; get;
```

_should_ retrieve relations like `(sub-arole: $x)` as well.

flyingsilverfin:
TODO: update tests to match new behavior above and see if the PR still passes!

test-integration/graql/reasoner/query/AtomicQueryUnificationIT.java
flyingsilverfin: Ok so this one is actually false because of a different reason - the second query has an extra variable, so it can never be subsumed by the one with fewer variables :)

jmsfltchr: I just had the thought - if this is false because of the extra variable, do we have a test case which models the behaviour when each has the same variables? I feel that would be useful to capture the full context of what's going on here.

flyingsilverfin: yep these exists already im quite sure!

PR #5668, opened 2020-03-25
Add query profiler based on resolution tree

kasper-piskorski:
To add a better way of examining query execution with reasoning by providing a way of outputting the resolution tree in a graphical-friendly form. The tree contains both answer and processing time per node information.

With logging set to `TRACE`, we now explicitly store the full ResolutionTree.
Once the query is finished executing, the tree is then output in Graph Description Laguage format (https://en.wikipedia.org/wiki/DOT_(graph_description_language) to the log directory.


graql/reasoner/tree/TreeWriter.java
```diff
@@ -0,0 +1,81 @@
...
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.nio.file.Path;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.Stack;
+
+public class TreeWriter {
```
flyingsilverfin: class comments on what this is andwhy

graql/reasoner/tree/ResolutionTree.java
```diff
@@ -0,0 +1,97 @@
...
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ *
+ */
+public class ResolutionTree {
```
flyingsilverfin: comment on class

graql/reasoner/tree/ResolutionTree.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public Set<Node> getNodes(){
+        return new HashSet<>(mapping.values());
+    }
+
+    public void clear(){
+        mapping.clear();
+    }
+
+    public void updateTree(ResolutionState state) {
```
flyingsilverfin: Can we rename this to be  abit more descriptive? like `recordNewState` or `addResolutionState`?

graql/reasoner/ResolutionIterator.java
```diff
@@ -115,19 +132,22 @@ public boolean hasNext() {
...
                 return hasNext();
             }
         }
 
         MultilevelSemanticCache queryCache = CacheCasting.queryCacheCast(this.queryCache);
-
         subGoals.forEach(queryCache::ackCompleteness);
         queryCache.propagateAnswers();
 
+        if (LOG.isTraceEnabled()) logTree.outputToFile(Paths.get("/tmp/query.profile"));
```
flyingsilverfin: I wonder if we can find a better place to put this file, would break on windows atm right

kasper-piskorski: ideally it should be in the log dir. Not sure how to access it.

flyingsilverfin: ask @vmax

kasper-piskorski: changed

PR #5671, opened 2020-03-30
Validate resolution profiler answers

kasper-piskorski:
Resolve issues with incomplete/out of sync answers in the resolution profiler.

- We ensure all the information from the `AnswerState` and its parent is propagated into the answer saved in the resolution Profiler. 
- Changed the logging level required to produce resolution profile to DEBUG.
- We test answer validity in the resolution profile as part of the `GeoInferenceIT` test. To do that we added an extra reasoner logback file.
- Fixed issue of comparing Resolution Nodes. We now explicitly test for Equality of the corresponding queries.

### DIFF
--- a/graql/reasoner/tree/NodeImpl.java
+++ b/graql/reasoner/tree/NodeImpl.java
@@ -19,13 +19,20 @@
 
 package grakn.core.graql.reasoner.tree;
 
+import com.google.common.collect.Sets;
 import grakn.core.concept.answer.ConceptMap;
+import grakn.core.graql.reasoner.query.ResolvableQuery;
+import grakn.core.graql.reasoner.state.AnswerPropagatorState;
 import grakn.core.graql.reasoner.state.ResolutionState;
+import grakn.core.kb.graql.reasoner.ReasonerException;
+import grakn.core.kb.graql.reasoner.atom.Atomic;
+import graql.lang.statement.Variable;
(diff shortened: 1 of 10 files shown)


graql/reasoner/tree/NodeImpl.java
```diff
@@ -84,9 +91,26 @@ public void addChild(Node child){
 
     @Override
     public void addAnswer(ConceptMap answer){
+        if (answer.isEmpty()) return;
+        validateAnswer(answer);
         answers.add(answer);
     }
 
+    private void validateAnswer(ConceptMap answer){
```
flyingsilverfin: why do we need this? can we write a comment for the method about why / what its doing?

graql/reasoner/tree/ResolutionTree.java
```diff
@@ -66,20 +74,35 @@ public void clear(){
...
-                ConceptMap sub = state.getSubstitution();
-                parentNode.addAnswer(sub);
-            }
+            addAnswerState((AnswerState) state, parent);
         } else {
             addChildToNode(parent, state);
         }
     }
 
+    private void addAnswerState(AnswerState state, AnswerPropagatorState parent){
```
flyingsilverfin: This feels out of places, we shouldn't have to do much/any logic in the tree recording that is already being done elsewhere :)

PR #5682, opened 2020-04-09
Update console dep

flyingsilverfin:
Due to a breaking change in client-java and console's printer (preventing client-java from merging its update), we have to merge console (with a dependency on my fork of client-java) and update the dependency on console here. We can then go and merge client-java and make console point at a graknlabs client java dependency again.

* Bump console dependency

### DIFF
--- a/dependencies/graknlabs/dependencies.bzl
+++ b/dependencies/graknlabs/dependencies.bzl
@@ -56,7 +56,8 @@ def graknlabs_console():
     git_repository(
         name = "graknlabs_console",
         remote = "https://github.com/graknlabs/console",
-        tag = "1.0.3",
+        # tag = "1.0.3",
+        commit = "27409b68d86b0b08ece1300b62dadd24baefe78b",
     )
 
 def graknlabs_benchmark():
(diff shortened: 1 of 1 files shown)


dependencies/graknlabs/dependencies.bzl
```diff
@@ -56,7 +56,8 @@ def graknlabs_console():
     git_repository(
         name = "graknlabs_console",
         remote = "https://github.com/graknlabs/console",
-        tag = "1.0.3",
+        # tag = "1.0.3",
```
haikalpribadi, line 59: why are we only commenting out the tag reference and not actually removing it, @flyingsilverfin ?

flyingsilverfin: Errr can do, mostly as a reminder to revert it back to a tag when we release console...?

haikalpribadi: That's already part of a `release-validation` check. No java repo can be released without depending on a `tag`, otherwise, release validation will fail. So we can remove the comment as it's part of natural workflow now.

flyingsilverfin: cool

PR #5690, opened 2020-04-17
Deletes accept statements instead of variables

flyingsilverfin:
`match...; delete...`; queries now accept full graql syntax in the `delete` clause.

This means we now:
* can _finally_ remove role players from relations using: `match $r (arole: $x) isa somerelation; delete $r (arole: $x)`
* We don't require `via` to remove attribute ownerships anymore, instead write: `match $x has attribute $a; delete $x has attribute $a;`
* delete an instance using an `isa` clause: `match $x isa person; delete $x isa person;`

* implement `Deletable` in certain property executors.
* Update and extend integration tests for `GraqlDeleteIT` extensively
* Clean up relations with no more RPs when unassigning role players via graql
* Modify the way `delete(DeleteQuery)` operates to utilise the same `Writer`s as Insert operations.
* Break ties when ordering delete statements with new tie breaker: edges first, then relations, then non-relations. This avoid trying to delete instances that are cleaned up via the deletion of other instances (eg. deleting RPs of a relation will clean up the relation, for instance)


flyingsilverfin:
TODO: clean up error messages

haikalpribadi:
Don't forget to refer to the issue that should be closed with this PR, @flyingsilverfin . Also, let's update the Graql dependency first before merging.

PR #5699, opened 2020-04-26
Fix issue with too eager subgoal propagation for queries with negation

### LINKED ISSUE #5697: Inconsistent inferred results
## Description

Grakn infers the head of a rule in one case but not in another.

## Environment

2. Grakn version (and platform): Grakn Core 1.6.2 (NOT tried on 1.7.0 yet)
3. Grakn client: console

## Reproducible Steps

Load this schema
[troubleshooting-schema.txt](https://github.com/graknlabs/grakn/files/4530406/troubleshooting-schema.txt)

Plus this data
[reasoner-bug.txt](https://github.com/graknlabs/grakn/files/4530414/reasoner-bug.txt)

(Apologies it isn't smaller, having a hard time seeing what can be going wrong.)

I find that there are inconsistent results.

Verify that there are no results for the head of the rule `determined-fault` with query A:
```
match (diagnosed-fault: $flt, parent-session: $ts) isa diagnosis; get;
```
Try query B in console:
```
match
  (diagnosed-fault: $flt, parent-session: $ts) isa diagnosis;
  $ts isa troubleshooting-session, has identifier 0;
  $flt has name $flt-name;
  $s($flt, $proc) isa solution;
  $proc isa procedure, has name $proc-name, has description $proc-desc;
get $flt-name, $proc-name, $proc-desc; group $flt-name;
```
Probably there's no results, and there shouldn't be since it contains query A.
Now `rollback` to get a new transaction, run query B again and see that there's an answer :0
This even throws an error when you go back to the query A.

You can see how this looks here, each time I use `editor` I'm executing query B:
```
ts_bug> editor
{"low input volume" isa name: {{$flt-name "low input volume" isa name; $proc-name "Increase mic volume" isa name; $proc-desc "Either in the App's settings or system settings, increase the input volume for your microphone." isa description;}}}
ts_bug> match (diagnosed-fault: $flt, parent-session: $ts) isa diagnosis; get;
Error: UNKNOWN: Illegal index:  indices: [$p1]. Please check server logs for the stack trace.
ts_bug> match (diagnosed-fault: $flt, parent-session: $ts) isa diagnosis; get;
ts_bug> match (diagnosed-fault: $flt, parent-session: $ts) isa diagnosis; get;
ts_bug> match (diagnosed-fault: $flt, parent-session: $ts) isa diagnosis; get;
ts_bug> match (diagnosed-fault: $flt, parent-session: $ts) isa diagnosis; get;
ts_bug> editor
ts_bug> editor
ts_bug> rollback
ts_bug> editor
{"low input volume" isa name: {{$flt-name "low input volume" isa name; $proc-name "Increase mic volume" isa name; $proc-desc "Either in the App's settings or system settings, increase the input volume for your microphone." isa description;}}}
ts_bug>
```

kasper-piskorski:
Negation queries are short circuited when one contradiction is found from the negation block. However, we mark negation sub-queries as fully explored by adding the sub-query to the visited `subgoals` that is shared globally. We then re-visit the same state at another point while exploring the negation query. However, we are missing answers, since the cache will only contain a single answer to the visited sub-query, not all of them. This means we can miss other solutions to the negation block. This is the underlying problem of #5697.

## Detailed Explanation

The issue can be described in terms of the 2 following query profiles which both describe the execution of a variant of the query:

```
($flt, $ques) isa fault-identification;
(question-not-asked: $ques, parent-session: $ts) isa unasked-question;
$flt id ...;
$ts id ...;
```
inside a negation block, for different substitutions for `$flt`.

The first profile shows the first execution of the query for `[$flt/65736]`. The conjunctive query is decomposed into atomic queries, the first query is:

`unasked-question($ts, $ques) [$ts/V53256]`

The query is seen for the first time (not been visited before) hence a rule is applied. Since the query is within a negation block, once an answer is found, we immediately return.

image

The second query shows the execution of a following query with  `[$flt/32960]`. Once we decompose the conjunction, the query produces the same query as in the first profile:

`unasked-question($ts, $ques) [$ts/V53256]`
This query has however been marked as visited and we haven't consumed all its answers as we were interested only in a single one. A quick inspection of the db reveals that there is actually a pair. As a result we only do a table lookup and retrieve a single answer which is then fed into the second atomic query. The atomic query correctly doesn't yield any answers. We have however not processed the second answer from the pair which is the one that satisfies the conjunction. An answer is lost.
image

Make sure we do not reuse global subGoals in negation branches - we do an explicit copy before branching out.

Fixes #5697


haikalpribadi:
Thanks, @kasper-piskorski ! Can you please `rebase` your PR off the latest `master` branch? I've just made `@graknlabs_verification` public so your tests will run.

PR #5703, opened 2020-04-27
Ability to change Storage startup timeout via ENV

harduino:
When running in docker-compose mode, if there are a lot of other docker-services, storage sometimes fails to start and throws error "Unable to start Storage". So, default timeout is increased from 60 to 180 seconds.
Also I've added ability to set custom startup timeout via environment variable STORAGE_STARTUP_TIMEOUT_SECOND

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,7 +87,7 @@ jobs:
       - run-bazel-rbe:
           command: bazel test //common/... --test_output=errors
 
-  test-behavior-connection:
+  test-behaviour-connection:
     machine: true
     working_directory: ~/grakn
     steps:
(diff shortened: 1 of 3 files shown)


haikalpribadi:
Thank you for the PR, @harduino ! Can you please `rebase` your PR off the latest `master` branch? I've just fixed the issue that causes your tests to fail.

haikalpribadi:
Sorry I accidentally closed this while trying to fix it manually from my side, @harduino !

haikalpribadi:
Here's the source code that you initially wrote, rebased on top of the latest `master`, @harduino : https://github.com/haikalpribadi/grakn/commits/storage-startup-timeout

haikalpribadi:
I want to reopen this PR, but GitHub does not seem to let me!

Issue #5043, opened 2019-03-18
Refactor tests into the right packages


lolski:
## Problem to Solve

Currently, there is an inconsistency where the root test structure looks like this:
```
test-end-to-end/
test-integration/
test/
  assembly/
  deployment/
  common/
```

We also have some unit tests sitting across the repositories packages and subpackages.

## Proposed Solution

There should be 4 categories of tests: unit, integration, assembly, deployment. All unit tests should sit inside the package in which the unit test is testing. The rest of the test categories should sit under `test/`:
```
test/
  integration/
  assembly/
  deployment/
```

Additionally, you may need to have `test/common` package for any shared code.

haikalpribadi:
@kasper-piskorski given that most of the tests in Grakn Core for reasoner, can you help me on this issue and make sure we organise our tests neatly in the proposed structure above?

@flyingsilverfin for all the Graql tests, do you think we can migrate them over to the `@graknlabs_behaviour` repo? Basically, can we move the test coverage from the integration tests (where most of them sit right now) to be incorporated in the axiomatic theorem tests? cc: @jmsfltchr

haikalpribadi:
1) Inside `//test-end-to-end` exists `ClientJavaE2E2`. This test should be moved to `@graknlabs_client_java//test` (`/assembly` or `/deployment`)
2) `Deduplicator` test should really be Integration Tests for `grakn.core.server.deduplicator`. We should not test `Deduplicator` using `GraknClient`, and use TransactionOLTP instead.
3) Distribution test should be grakn-core's "assembly tests" (`//test/assembly`)

flyingsilverfin:
https://github.com/graknlabs/client-java/pull/96 will complete 1. in haikal's comment

PR #5708, opened 2020-04-29
Refactor test packages

flyingsilverfin:
Multiple test currently reach across package boundaries to utilise the `GraknTestServer` or the `test-integration/logback.xml`. We can fix this architecture and resolve https://github.com/graknlabs/grakn/issues/5043 at the same time.

* Move `ClientJavaE2E` into client-java repository
* Merge `AttributeUniquenessE2E` into `AttributeUniquenessIT`
* Convert `ConcurrencyE2E` into IT
* Create new package in tests for `test-integration/server/uniqueness` containing `AttributeUniquessIT` and `ConcurrencyIT`


flyingsilverfin:
@haikalpribadi  im not sure `test-assembly` is the best name for tests like `GraknGraqlCommandsE2E` - that used to be E2E tests. It's a bit awkward since we have `test-assembly-mac-zip` that have avery different meaning. It's more of a `test-assembled-commands` or something

haikalpribadi:
I don't have enough context at the moment to know what is in the actual tests, @flyingsilverfin. But the test terminologies that we have decided on is:
- test-unit: for class level
- test-integration: for package or multi-package level
- test-assembly: for testing the software distribution once you've "assembled" the distribution
- test-deployment: for testing the software distribution once you've assembled and "deployed" the distribution

Ideally, each test should be suffixed with the class, package, or distribution it is testing. Additionally, the specific aspect/behaviour of that distribution you're testing.

Does that help?

flyingsilverfin:
I suppose in that case I need to find a name for something that encompasses "user-experience" or "functionality" in an assembly test

haikalpribadi:
I have a feeling that these test are neither here nor there, and it's just a bunch of cases put together with an scoping domain that it's focusing on. I'd say, if you can migrate each case to BDD scenarios, then we should just dissolve this test?

flyingsilverfin:
Then we'd need behaviour tests that handle high level commands such as runnnig console commands haikal, seems a bit out of scope of the behavior steps we have now? Can discuss on a call or im happy to just keep them as is for now and we'll revisit them again in the future

test/assembly/BUILD
```diff
@@ -0,0 +1,70 @@
...
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+load("@graknlabs_build_tools//checkstyle:rules.bzl", "checkstyle_test")
+
+java_test(
+    name = "daemon-test-assembly",
```
haikalpribadi, line 21: Let's reverse the naming, and have `test` be the prefix. So this one we can call `test-daemon`

test/assembly/BUILD
```diff
@@ -0,0 +1,70 @@
...
+        "//server",
+        "//dependencies/maven/artifacts/commons-io",
+        "//dependencies/maven/artifacts/org/hamcrest:hamcrest-library",
+        "//dependencies/maven/artifacts/org/zeroturnaround:zt-exec",
+        "@graknlabs_client_java//:client-java",
+    ],
+)
+
+java_test(
+    name = "distribution-test-assembly",
```
haikalpribadi, line 40: Let's reverse the naming, and have `test` be the prefix. So this one we can call `test-distribution`

haikalpribadi:
Will this PR close #5403 completely, @flyingsilverfin ? The following text in the description will trigger the closing of that issue.

> We can fix this architecture and **resolve #5043** at the same time.

flyingsilverfin:
yes @haikalpribadi it should

PR #5709, opened 2020-04-29
Explanation Fixes and Behaviour Changes

jmsfltchr:

We fix some broken elements of the explanations that Grakn returns with its answers. In doing so we revise the structure of the explanations returned.

- Answers that were obtained via lookup, that were nested in explanations were returned with an incorrect ConceptMap and pattern, now fixed
- Our expectation as users was that of the three types of explanation: `Lookup`, `Rule`, and `Join`, only the `Join` would contain more than one answer. In fact, a `RuleExplanation` would essentially swallow its children. We rectify this with the structure we naively expect. In this way, a `RuleExplanation`'s answer is always a single, full answer to its body (`when`), with the corresponding pattern.


graql/reasoner/explanation/JoinExplanation.java
```diff
@@ -34,11 +32,6 @@ public JoinExplanation(List<ConceptMap> partialAnswers){
         super(partialAnswers);
     }
 
-    @Override
-    public JoinExplanation childOf(ConceptMap ans) {
```
flyingsilverfin, line 38: yaasss bye byee

test-integration/server/GraknClientIT.java
```diff
@@ -262,7 +262,7 @@ public void explanationIsGivenWithRule() {
             Set<ConceptMap> deductions = deductions(answer);
             assertTrue(deductions.stream().anyMatch(ans -> {
                 assertNotNull(ans.queryPattern());
-                if (ans.hasExplanation()) {
+                if (ans.hasExplanation() && ans.explanation().getAnswers().size() == 1) {
```
flyingsilverfin, line 265: what is this special extra check for?

jmsfltchr: This is the non-deterministic failure I was telling you about - the check was previously looking for the single `RuleExplanation` among the answers. Now there are two explanations, a `RuleExplanation` and a `JoinExplanation`, and without this new check it could find either. This test uses the client so we can't check the explanation's type specifically.

Issue #5570, opened 2020-01-30
Unable to start Storage when starting Grakn Docker


neomatrix369:
## Description

Started Grakn docker container and get the below error in the docker console:

```
Version: 1.5.7
Starting Storage......FAILED!
Unable to start Storage. 
```

This is intermittent, does happen all the time. Is this resolved in the later versions? For now, I would like to continue to use 1.5.7.

## Environment

1. OS (where Grakn server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. Grakn version (and platform): 1.5.7 or 1.6.2
3. Grakn client: n/a
4. Other environment details:
Docker container: version 19.03.5
JDK used: GraalVM CE 19.2.1 (I also run it under OpenJDK8 at times)

Some more config related details:
JAVA_HOME=/usr/local/graalvm-ce-19.2.1
openjdk version "1.8.0_232"
OpenJDK Runtime Environment (build 1.8.0_232-20191008104205.buildslave.jdk8u-src-tar--b07)
OpenJDK 64-Bit GraalVM CE 19.2.1 (build 25.232-b07-jvmci-19.2-b03, mixed mode)
STORAGE_JAVAOPTS=-XX:+UseJVMCINativeLibrary -XX:+UseJVMCINativeLibrary
SERVER_JAVAOPTS=-XX:+UseJVMCINativeLibrary -XX:+UseJVMCINativeLibrary
GRAKN_DAEMON_JAVAOPTS=-XX:+UseJVMCINativeLibrary -XX:+UseJVMCINativeLibrary

_Note: please don't get confused by JDK8 strings above, GraalVM is based on JDK8._

## Reproducible Steps

Usual `grakn start server` command inside a container, please try using [this](https://github.com/neomatrix369/awesome-ai-ml-dl/tree/master/examples/data/databases/graph/grakn). 

Use the command:

```
$ GRAKN_VERSION=1.6.2 ./grakn-runner.sh --runContainer
or
$ GRAKN_VERSION=1.6.2 ./grakn-runner.sh --jdk GRAALVM --runContainer
```

to see it try to start the Grakn 1.6.2 inside the container. As I said it's intermittent and might be free space related. 

Will report with the relevant logs when it happens again.

## Expected Output

An up and running Grakn server

## Actual Output

See in description above.
 
## Additional information

[...]

PR #5710, opened 2020-04-29
Ability to change Storage startup timeout via ENV

harduino:
When running in docker-compose mode, if there are a lot of other docker-services, storage sometimes fails to start and throws error "Unable to start Storage". So, default timeout is increased from 60 to 180 seconds.
Also I've added ability to set custom startup timeout via environment variable STORAGE_STARTUP_TIMEOUT_SECOND

### DIFF
--- a/daemon/executor/Storage.java
+++ b/daemon/executor/Storage.java
@@ -187,7 +187,7 @@ public class Storage {
      * @throws GraknDaemonException
      */
     private void start() {
-        System.out.print("Starting " + DISPLAY_NAME + "...");
+        System.out.println("Starting " + DISPLAY_NAME + "...");
         System.out.flush();
 
         // Consume configuration from Grakn config file into Cassandra config file
@@ -195,7 +195,16 @@ public class Storage {
 
         Future<Executor.Result> result = daemonExecutor.executeAsync(storageCommand(), graknHome.toFile());
 
-        LocalDateTime timeout = LocalDateTime.now().plusSeconds(STORAGE_STARTUP_TIMEOUT_SECOND);
+        // form storage timeout (default or ENV-value)
+        long startupTimeoutCalculated = STORAGE_STARTUP_TIMEOUT_SECOND;
+        String timeoutEnvValue = System.getenv("STORAGE_STARTUP_TIMEOUT_SECOND");
+        if (timeoutEnvValue != null && !timeoutEnvValue.isEmpty()) {
+          startupTimeoutCalculated = Long.parseLong(timeoutEnvValue, 10);
+        }
+        System.out.println(DISPLAY_NAME + " timeout is configured to " + startupTimeoutCalculated + " seconds");
+        System.out.flush();
+
+        LocalDateTime timeout = LocalDateTime.now().plusSeconds(startupTimeoutCalculated);
 
         while (LocalDateTime.now().isBefore(timeout) && !result.isDone()) {
             System.out.print(".");
--- a/daemon/executor/Storage.java
+++ b/daemon/executor/Storage.java
@@ -60,7 +60,7 @@ import static grakn.core.daemon.executor.Executor.WAIT_INTERVAL_SECOND;
 public class Storage {
 
     private static final String DISPLAY_NAME = "Storage";
-    private static final long STORAGE_STARTUP_TIMEOUT_SECOND = 60;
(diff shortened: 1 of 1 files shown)


harduino:
Hi, @haikalpribadi

I cherry-picked my commits with the latest master branch

haikalpribadi:
Thanks, @harduino ! 

@lolski please review when you get a chance.

lolski:
Thank you @harduino !

@haikalpribadi I think the convention is to add configuration to `grakn.properties`, rather than environment variables. Therefore, what if we make it a configuration in `grakn.properties` under `daemon` (that's how we refer to the bootup script):

```
# conf/grakn.properties
...
daemon.storage-bootup-timeout=180
```

haikalpribadi:
Sounds good, @lolski 👍

SamuelHassine:
Hello,

This PR seems to be the good solution to narrow the timeout. But as Julien and I requested (https://github.com/graknlabs/grakn/issues/5570#issuecomment-623534292), please adjust it so the `Grakn Server Start... failed` stops the Docker container. For the moment it runs forever.

richard-julien:
Add please make it also configurable through an environment variable, is really simpler than mounting a configuration file just for that.

lolski:
> Add please make it also configurable through an environment variable, is really simpler than mounting a configuration file just for that.

@haikalpribadi right now, `grakn.properties` cannot be overridden with environment variable (unlike in Grabl), and we have no plan for adding that in the near future.

Should we allow overriding the Storage timeout using an environment variable anyway (as a special case) as per @richard-julien 's comment?

richard-julien:
Im always vote for consistency, in that case you should also take care of JAVA_OPTS option :)

But to be pragmatic for user that have this kind of problem (basically docker user) i clearly vote for a quick win with something like that: 
> private static final int STARTUP_TIMEOUT= Integer.parseInt(SystemProperty.STARTUP_TIMEOUT.value() || "300");

And then you can eventually think about push all the config in grakn.properties with the capability to override all the config through environnement variables.

haikalpribadi:
I think it would be better to follow the current behaviour and configure it through `grakn.properties`, and in the future, we introduce a feature to override `grakn.properties` values with an environment variable.

daemon/executor/Storage.java
```diff
@@ -187,15 +187,24 @@ public boolean isRunning() {
...
 
         // Consume configuration from Grakn config file into Cassandra config file
         initialiseConfig();
 
         Future<Executor.Result> result = daemonExecutor.executeAsync(storageCommand(), graknHome.toFile());
 
-        LocalDateTime timeout = LocalDateTime.now().plusSeconds(STORAGE_STARTUP_TIMEOUT_SECOND);
+        // form storage timeout (default or ENV-value)
+        long startupTimeoutCalculated = STORAGE_STARTUP_TIMEOUT_SECOND;
+        String timeoutEnvValue = System.getenv("STORAGE_STARTUP_TIMEOUT_SECOND");
```
lolski, line 200: 1. We should declare the new config in `grakn.properties`:

```
# conf/grakn.properties
...
daemon.storage-bootup-timeout=180
```

2. We should declare the new config in https://github.com/graknlabs/grakn/blob/master/common/config/ConfigKey.java#L63

```
public static final ConfigKey<String> DAEMON_STORAGE_BOOTUP_TIMEOUT = key("daemon.storage-bootup-timeout", INT);
```

3. Finally, we should read the value from the config:

```
Integer timeoutEnvValue = graknProperties.getProperty(ConfigKey.DAEMON_STORAGE_BOOTUP_TIMEOUT)
```

flyingsilverfin:
@harduino any chance you've got a moment to make the requested changes? :)

harduino:
@flyingsilverfin what changes are you talking about?

flyingsilverfin:
https://github.com/graknlabs/grakn/pull/5710#discussion_r422254149

lolski:
Perhaps Harduino is currently busy.

@richard-julien if you need it urgently in OpenCTI, feel free to submit a PR based on the guideline [here](https://github.com/graknlabs/grakn/pull/5710#discussion_r422254149) :D

lolski:
Closing this PR since in 2.0 this won't be an issue.

Still, thank you for the contribution @harduino :)

PR #5711, opened 2020-04-30
Support for Graql Delete and validate keys on deletion of ownership

flyingsilverfin:
Delete BDD scenarios are implemented in https://github.com/graknlabs/verification/pull/38  - this PR implements the new steps required to execute the new scenarios. We also find and fix a bug where validation was not performed for keys when attributes were "un-owned".

* Add `graql delete` and `graql delete throws` steps
* Move query parsing into `try/catch` for each `graql ___ throws` step
* Add objects that un-has keys to be validated at commit time


flyingsilverfin:
TODO revert dependency

PR #5714, opened 2020-05-05
Deletion - fix ordering bug

flyingsilverfin:
In some complex deletion cases, we must delete attribute ownerships and role playing first, followed by relations instances and other instances. We applied an incorrect tie breaker, resulting in deleting relations before attribute ownerships, resulting in `vertex deleted` exceptions from Janus. We also enforce deleting attribute ownerships before role players, as role player deletion could lead to cascading deletions.

* Add explicit ordering of deletes: HAS, ROLE_PLAYER, RELATION, NON_RELATION


flyingsilverfin:
@jmsfltchr this is failing because of explanation tests! Are you going to be able to submit your PR with fixes for these to master anytime soon?

jmsfltchr:
Aha yes @flyingsilverfin we can put that in soon

graql/executor/property/HasAttributeExecutor.java
```diff
@@ -135,20 +134,24 @@ public VarProperty property() {
...
+            vars.add(var);
+            vars.add(property.attribute().var());
+            return vars;
         }
 
         @Override
         public TiebreakDeletionOrdering ordering(WriteExecutor executor) {
-            return TiebreakDeletionOrdering.EDGE;
+            return TiebreakDeletionOrdering.HAS;
         }
```
haikalpribadi, line 155: This method takes in a `WriteExecutor` but not use it. Does the other implementation of the interface where this method comes from uses it properly? Some feels off...

graql/executor/property/RelationExecutor.java
```diff
@@ -170,23 +164,27 @@ public VarProperty property() {
         }
 
         @Override
-        public Set<Variable> requiredVars() {
-            return allVars();
+        public TiebreakDeletionOrdering ordering(WriteExecutor executor) {
+            return TiebreakDeletionOrdering.ROLE_PLAYER;
+        }
```
haikalpribadi, line 169: Here as well...

flyingsilverfin: yeah only in 1 case do we actually need the WriteExecutor - with `isa`. We need to delete `Relation` instances before other types (and for now implicit attr relations are also deletable using `isa @has...`  for a short while

haikalpribadi: We can change it to not take in a `writeexecutor`?

PR #5723, opened 2020-05-18
Change behaviour of explanations and their patterns

jmsfltchr:

We change the behaviour of explanations quite drastically. Previously explanations were only provided to explain answers derived using a rule. Now we introduce explanations for negations and disjunctions, which means that an explanation will be present for any query that contains a disjunction or negation.

Explaining a disjunction now shows the clause of the disjunction that yielded the answer, including the full ConceptMap for that clause.

We alter the format of patterns in answers. Patterns still include ids, which should be limited to the set of ids that are returned in the answer's ConceptMap. These ids now only occur once each within the pattern, and will always sit in the root conjunction of the pattern.

Previously patterns for conjunctive queries were working, but patterns for queries and rules with negation were incorrect. This is necessary for our efforts to build a resolution testing framework.

- Ids in answers' patterns now occur at most once, and in the outermost scope of the pattern.
- Changes the behaviour of how negations and disjunctions are represented in explanations
  - Disjunctions are now represented in the pattern in their DNF'ed form, and explained with the clause that was matched for the given answer
  - Negations now include the omitted `not`, and are explained by the lookup that was made (prior to being checked against the `not` clause)


graql/reasoner/query/CompositeQuery.java
```diff
@@ -311,9 +343,26 @@ public void checkValid() {
 
     @Override
     public Conjunction<Pattern> getPattern() {
-        Set<Pattern> pattern = Sets.newHashSet(getConjunctiveQuery().getPattern());
-        getComplementQueries().stream().map(ResolvableQuery::getPattern).forEach(pattern::add);
-        return Graql.and(pattern);
+        if (pattern == null) {
+            Set<Pattern> conjunctPatterns = Sets.newLinkedHashSet(getConjunctiveQuery().getPattern().getPatterns());
+            getComplementQueries().stream().map(ResolvableQuery::getPattern).forEach(p -> {
+                if (p instanceof Conjunction && ((Conjunction<Pattern>) p).getPatterns().size() == 1) {
```
jmsfltchr, line 349: Investigate whether this can be lived with by adding extra `{}` inside the `not` in the test case

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,304 @@
...
+
+    public DisjunctiveQuery(Set<CompositeQuery> clauses, Set<Variable> bindingVars, TraversalExecutor traversalExecutor, ReasoningContext ctx) {
+        super(traversalExecutor, ctx);
+        this.clauses = clauses;
+        this.bindingVars = bindingVars;
+    }
+
+    @Override
+    public CompositeQuery asComposite() {
+        throw new RuntimeException("Not composite"); //TODO What exception should this throw?
```
jmsfltchr, line 73: Use `GraknSemanticException`

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,304 @@
...
+        return new HashSet<>(clauses);
+    }
+
+    public Set<Variable> getBindingVars() {
+        return new HashSet<>(bindingVars);
+    }
+
+    // Only needed for a disjunction to be the body of a rule
+    @Override
+    public Set<String> validateOntologically(Label ruleLabel) {
```
jmsfltchr, line 91: throw an exception until we enable disjunctions in rules

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,304 @@
...
+                getBindingVars(),
+                traversalExecutor,
+                context()
+        );
+    }
+
+    @Override
+    ResolvableQuery copy() {
+        return new DisjunctiveQuery(
+                new HashSet<>(getClauses()),
```
jmsfltchr, line 159: getClauses is already a copy

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,304 @@
...
+    private final Set<Variable> bindingVars;
+
+    public DisjunctiveQuery(Disjunction<Conjunction<Pattern>> pattern, Set<Variable> bindingVars, TraversalExecutor traversalExecutor, ReasoningContext ctx) {
+        super(traversalExecutor, ctx);
+        ReasonerQueryFactory queryFactory = context().queryFactory();
+        clauses = pattern.getPatterns().stream().map(queryFactory::composite).collect(Collectors.toSet());
+        this.bindingVars = bindingVars;
+    }
+
+    public DisjunctiveQuery(Set<CompositeQuery> clauses, Set<Variable> bindingVars, TraversalExecutor traversalExecutor, ReasoningContext ctx) {
```
jmsfltchr, line 65: Should accept queries other than `CompositeQuery`

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,304 @@
...
+        return new DisjunctiveQuery(
+                new HashSet<>(getClauses()),
+                getBindingVars(),
+                traversalExecutor,
+                context());
+    }
+
+    @Override
+    public boolean isAtomic() {
+        // TODO unclear of the meaning of atomicity in this case
```
jmsfltchr, line 167: @kasper-piskorski what do you think the behaviour of `isAtomic()` should be for a Disjunctive Query?

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    @Override
+    public String toString(){
+        return getPattern().toString();
+    }
+
+    @Override
+    public ReasonerQuery conjunction(ReasonerQuery q) {
+        throw new UnsupportedOperationException();
```
jmsfltchr, line 185: Delete as the method is moved out to be a helper

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    @Override
+    public Set<Variable> getVarNames() {
+        return bindingVars;
+    }
+
+    @Override
+    public Set<Atomic> getAtoms() {
+        return getClauses().stream().flatMap(c -> c.getAtoms().stream()).collect(Collectors.toSet());
```
jmsfltchr, line 215: @kasper-piskorski also here

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    @Override
+    public Disjunction<Pattern> getPattern() {
+        // TODO How can we rebuild the disjunction in proper Graql form (with unbound variables outside the disjunction) once it's already in disjunctive normal form?!
```
jmsfltchr, line 220: Comment now moot

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,304 @@
...
+    public Conjunction<Pattern> getPattern(Map<Variable, Concept> map) {
+        HashSet<Pattern> patterns = getIdPredicatePatterns(map);
+        patterns.add(getPattern());
+
+        return Graql.and(patterns);
+    }
+
+    @Override
+    public ConceptMap getSubstitution() {
+        // TODO What should this do, given that there could be a substitution for any of its clauses
```
jmsfltchr, line 234: If there is a substitution in the query it should be in the form of id predicates. These should be converted out to match the set of `bindingVars`

jmsfltchr: Leave as an exception for now. Will be needed if we don't normalise queries or use disjunctions in rules.

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,304 @@
...
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class DisjunctiveQuery extends ResolvableQuery {
```
jmsfltchr, line 54: We expect DisjunctiveQueries to never directly contain Atoms, they can conceptually contain CompositeQueries, ConjunctiveQueries, AtomicQueries, and they can never contain another DisjunctiveQuery thanks to negationDNF

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    @Override
+    public SetMultimap<Variable, Type> getVarTypeMap() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public SetMultimap<Variable, Type> getVarTypeMap(boolean inferTypes) {
+        return getVarTypeMap(); // TODO Seems a weird thing to do since that throws an exception
```
jmsfltchr, line 260: This is ok

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    @Override
+    public ResolutionState resolutionState(ConceptMap sub, Unifier u, AnswerPropagatorState parent, Set<ReasonerAtomicQuery> subGoals) {
+        return new DisjunctiveState(this, sub, u, parent, subGoals);
+    }
+
+    @Override
+    public Iterator<ResolutionState> innerStateIterator(AnswerPropagatorState parent, Set<ReasonerAtomicQuery> subGoals) {
+//        TODO Does the substitution make sense here?
```
jmsfltchr, line 299: ✅

graql/reasoner/query/ReasonerQueryFactory.java
```diff
@@ -86,12 +88,16 @@ public ResolvableQuery resolvable(Conjunction<Pattern> pattern) {
...
-     * @return resolvable query with the substitution contained in the query
+     * @param pattern disjunctive pattern defining the query
+     * @param bindingVars the set of variables in the outer scope of the disjunction
+     * @return a resolvable reasoner query constructed from provided disjunctive pattern
      */
-    public ResolvableQuery resolvable(ResolvableQuery q, ConceptMap sub) {
-        return q.withSubstitution(sub).inferTypes();
+    public ResolvableQuery resolvable(Disjunction<Conjunction<Pattern>> pattern, Set<Variable> bindingVars) {
+        if (pattern.getPatterns().size() == 1) {
+            return resolvable(pattern.getPatterns().iterator().next());
```
jmsfltchr, line 97: Use `Iterators.getOnlyElement()` instead

graql/reasoner/query/ReasonerQueryImpl.java
```diff
@@ -169,15 +171,31 @@ public CompositeQuery asComposite() {
         return new CompositeQuery(getPattern(), traversalExecutor, context());
     }
 
+    @Override
+    public DisjunctiveQuery asDisjunctive() {
+        throw new RuntimeException("Not disjunctive"); //TODO What exception should this throw?
```
jmsfltchr, line 176: Same as previous

graql/reasoner/query/ReasonerQueryImpl.java
```diff
@@ -169,15 +171,31 @@ public CompositeQuery asComposite() {
         return new CompositeQuery(getPattern(), traversalExecutor, context());
```
jmsfltchr, line 171: We should try to make this throw as we shouldn't be able to have a CompositeQuery without negations. Keep in mind that this class should then become `ConjunctiveQuery`

graql/reasoner/query/ReasonerQueryImpl.java
```diff
@@ -246,7 +264,7 @@ public int hashCode() {
     public void checkValid() {
         getAtoms().forEach(Atomic::checkValid);
         //this creates the corresponding substitution and checks for id validity
-        ConceptMap substitution = getSubstitution();
+        ConceptMap substitution = getSubstitution(); //TODO This is dangling!
```
jmsfltchr, line 267: Delete

graql/reasoner/query/ReasonerQueryImpl.java
```diff
@@ -262,6 +280,13 @@ public void checkValid() {
         return pattern;
     }
 
+    @Override
+    public Pattern getPattern(Map<Variable, Concept> map){
```
jmsfltchr, line 284: Look at moving into the supertype

graql/reasoner/query/ResolvableQuery.java
```diff
@@ -139,14 +158,7 @@ public GraqlGet getQuery() {
...
     @CheckReturnValue
-    public Stream<ConceptMap> resolve(Set<ReasonerAtomicQuery> subGoals, boolean infer) {
-        boolean doNotResolve = !infer || getAtoms().isEmpty() || (isPositive() && !isRuleResolvable());
-        if (doNotResolve) {
-            return traversalExecutor.traverse(getPattern());
-        } else {
-            return new ResolutionIterator(this, subGoals, ctx.queryCache()).hasStream();
-        }
-    }
+    public abstract Stream<ConceptMap> resolve(Set<ReasonerAtomicQuery> subGoals, boolean infer);
```
jmsfltchr, line 161: Move part of this back up again!

graql/reasoner/state/AnswerPropagatorState.java
```diff
@@ -58,6 +58,10 @@
         this.subGoalIterator = generateChildStateIterator();
     }
 
+    public String toString(){
+        return getClass().getSimpleName() + "@" + Integer.toHexString(hashCode()) + "\n" + getQuery().toString();
```
jmsfltchr, line 62: Investigate duplicate output in `toString()`

graql/reasoner/query/CompositeQuery.java
jmsfltchr: This was achievable!

graql/reasoner/query/ReasonerQueryImpl.java
jmsfltchr: The implementation of `public Pattern getPattern(Map<Variable, Concept> map)` across `CompositeQuery`, `DisjunctiveQuery` and `ReasonerQueryImpl` is nearly identical, but to make them the same would nest conjunctions in conjunctions, so leaving them as-is for now

graql/reasoner/query/DisjunctiveQuery.java
jmsfltchr: Actually this fits well with the hierarchy and as a helper may be an anti-pattern of `instanceOf`checks against each type

graql/reasoner/state/DisjunctiveState.java
```diff
@@ -0,0 +1,65 @@
...
+import graql.lang.statement.Variable;
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Set;
+
+public class DisjunctiveState extends AnswerPropagatorState<DisjunctiveQuery> {
+
+    // TODO Surely the subGoals won't necessarily be atomic and may be whole conjunctions
+    public DisjunctiveState(DisjunctiveQuery query, ConceptMap sub, Unifier u, AnswerPropagatorState parent, Set<ReasonerAtomicQuery> subGoals) {
```
kasper-piskorski, line 36: why TODO? subgoals will be atomic unless you go with a different resolution algorithm

graql/reasoner/query/DisjunctiveQuery.java
kasper-piskorski: atomic queries are either `ReasonerAtomicQuery` or conjunctions (`ReasonerQueryImpl`) with a single selectable atom. Since `DisjunctiveQuery` contains always >1 conjunctions it is never atomic. So either return false or throw - disjunctions being expected to be atomic might suggests problems in the code and a behaviour to be investigated.

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    @Override
+    public boolean isAtomic() {
+        // TODO unclear of the meaning of atomicity in this case
+        return getClauses().stream().allMatch(ResolvableQuery::isAtomic);
+    }
+
+    @Override
+    boolean isEquivalent(ResolvableQuery q) {
```
kasper-piskorski, line 167: this is an another interesting function. I think it only carries the expected meaning for atomic queries. Might be worth to throw here to highlight unexpected execution paths.

graql/reasoner/query/DisjunctiveQuery.java
kasper-piskorski: You could either return all atoms or throw. Throwing might help highlighting unexpected execution paths. If you want to actually use it then return all atoms.

graql/reasoner/query/ReasonerQueryImpl.java
```diff
@@ -245,8 +258,6 @@ public int hashCode() {
     @Override
     public void checkValid() {
         getAtoms().forEach(Atomic::checkValid);
-        //this creates the corresponding substitution and checks for id validity
-        ConceptMap substitution = getSubstitution();
```
kasper-piskorski, line 249: just a note, this removes id validation. If that's expected then cool.

jmsfltchr: Aha, this is because `getSubstitution()` throws if ids are not found. No this was not intended.

graql/reasoner/query/ReasonerQueryImpl.java
jmsfltchr: To be addressed in a separate PR as a refactor

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,277 @@
...
+
+    @Override
+    public Set<Variable> getVarNames() {
+        return bindingVars;
+    }
+
+    @Override
+    public Set<Atomic> getAtoms() {
+        return getClauses().stream().flatMap(c -> c.getAtoms().stream()).collect(Collectors.toSet());
+    }
```
flyingsilverfin, line 193: not sure if you resolved kasper's comments :)

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,277 @@
...
+
+    @Override
+    public Disjunction<Pattern> getPattern() {
+        return Graql.or(clauses.stream().map(ResolvableQuery::getPattern).collect(Collectors.toSet()));
+    }
+
+    @Override
+    public Conjunction<Pattern> getPattern(Map<Variable, Concept> map) {
```
flyingsilverfin, line 201: you think this name is sufficient? I would at least call the `map` the `substitution` in line with other reasoner terms?

graql/reasoner/query/DisjunctiveQuery.java
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    @Override
+    public boolean isRuleResolvable() {
+        return clauses.stream().anyMatch(ResolvableQuery::isRuleResolvable);
+    }
+
+    @Override
+    public MultiUnifier getMultiUnifier(ReasonerQuery parent) {
+        throw new UnsupportedOperationException();
```
flyingsilverfin, line 220: Let's put a message in here that we can't unify with Disjunctions

graql/reasoner/state/DisjunctiveState.java
```diff
@@ -0,0 +1,64 @@
...
+    ConceptMap consumeAnswer(AnswerState state) {
+        ConceptMap sub = state.getSubstitution();
+        return new ConceptMap(sub.map(), new DisjunctiveExplanation(sub), getQuery().withSubstitution(sub).getPattern());
+    }
+
+    @Override
+    ResolutionState propagateAnswer(AnswerState state) {
+        ConceptMap answer = consumeAnswer(state);
+
+        HashMap<Variable, Concept> outerScopeVarsSub = getQuery().filterBindingVars(answer.map());
```
flyingsilverfin, line 54: if `filterBindingVars` is not used internally in `DisjunctiveQuery`, I would put this in a `AnswerUtil` or something and then just have a gernric `filterMapVars(ConceptMap map, Set<Variable> varsToKeep)` -- cleans up the query clas a bit

graql/reasoner/state/CompositeState.java
```diff
@@ -94,7 +93,10 @@ public ResolutionState propagateAnswer(AnswerState state) {
                 .noneMatch(q -> q.resolve(subGoals, true).findFirst().isPresent());
 
         return isNegationSatisfied?
-                new AnswerState(answer, getUnifier(), getParentState()) :
-                null;
+                new AnswerState(
+                        new ConceptMap(answer.map(), answer.explanation(), getQuery().getPattern(answer.map())),
```
flyingsilverfin, line 97: hopefully this doesn't have any implications with performance, rewriting the concept maps more often :) we can bench in taxfix

graql/reasoner/state/CompositeState.java
```diff
@@ -62,22 +63,20 @@
     private final Set<ResolvableQuery> complements;
 
     public CompositeState(CompositeQuery query, ConceptMap sub, Unifier u, AnswerPropagatorState parent, Set<ReasonerAtomicQuery> subGoals) {
-        super(query.withSubstitution(sub), sub, u, parent, subGoals);
+        super(query, sub, u, parent, subGoals);
```
flyingsilverfin, line 66: looks much better!

PR #5728, opened 2020-05-27
Block gRPC request observer thread to get back-pressure

### LINKED ISSUE #5726: Many async query requests can cause server OOM and crash
# Problem to solve

When streaming many asynchronous requests, it becomes possible for the client to overload the server's buffer with requests and cause an out of memory crash. This is because the server aggressively pulls and buffers requests, rather than setting up back-pressure to force the client to hold off on additional requests until capacity is available. With the incoming changes to clients to make as many queries async as possible, it is critical that we fix this issue to prevent a bad user experience due to "random" server crashes.

# Proposed Solution

Fortunately, we have a way to control flow in gRPC, as is shown in this example: https://github.com/grpc/grpc-java/blob/master/examples/src/main/java/io/grpc/examples/manualflowcontrol/ManualFlowControlServer.java

This allows us to prevent gRPC from requesting additional messages from the client in our defined cases, since we make those requests explicit. We are currently operating on a decoupled incoming request BlockingQueue for each transaction (freeing up gRPC threads as soon as we can). Logic will need to be included such that when new requests are pulled, the queue is filled up with requests where possible (making the request from the separate transaction thread is okay according to gRPC docs). Some experimentation may be required to understand what the optimal buffer size is here, and the docs do not make it 100% clear but it appears that the `isReady()` method on the `ServerCallStreamObserver` is for knowing whether or not the server buffer is ready to receive more requests. We probably need to account for both but I'd expect with something like 100 queue-limit and 100 message request batches should be acceptable in terms of memory usage, but the example shows 1 request at a time, so I am not sure whether gRPC already optimizes this for us at the channel level.

adammitchelldev:

Fix #5726 by allowing gRPC's natural back-pressure (automatic flow control) to take place.

Request messages are no longer eagerly queued asynchronously for processing in the TransactionService, and instead are blocked by waiting on the returned future when submitting them. This blocks the gRPC thread, which prevents the server buffer from being depleted and allows gRPC to control the back-pressure.

Also fixed dependency on a now non-existent commit of verification (due to a rebase).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -454,6 +454,7 @@ workflows:
               only:
                 - master
                 - rewrite
+                - 1.7.2
       - test-common:
           filters:
             branches:
@@ -496,6 +497,7 @@ workflows:
               only:
                 - master
                 - rewrite
+                - 1.7.2
           requires:
             - build
             - build-checkstyle
@@ -512,6 +514,7 @@ workflows:
               only:
                 - master
                 - rewrite
+                - 1.7.2
           requires:
             - build
             - build-checkstyle
@@ -528,6 +531,7 @@ workflows:
               only:
                 - master
                 - rewrite
+                - 1.7.2
           requires:
             - build
             - build-checkstyle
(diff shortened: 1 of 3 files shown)


haikalpribadi:
what's holding back this PR, @adammitchelldev ?

PR #5729, opened 2020-05-27
Bump Graql Dependency and use `DateTime`

flyingsilverfin:
Re-synchronise dependencies to use `master` of different repositories. This change includes using `datetime` instead of `date` for the Date attribute type.

* Bump depedencies to latest Graql
* use `datetime` instead of `date` for attribute types


graql/executor/property/ValueTypeExecutor.java
```diff
@@ -51,16 +51,16 @@
         }
         this.property = property;
 
-        if (!VALUE_TYPES.containsKey(property.valueType())) {
+        if (!VALUE_TYPES.containsKey(property.ValueType())) {
```
haikalpribadi, line 54: The method name looks weird here, shouldn't it be camel case and thus small `v` at the start?

flyingsilverfin: yeah i made a mistake, ive already fixed it in my next "actual" PR in `graql` that will go in with the implicit relations

flyingsilverfin: didn't think it was worth a wholew new PR

PR #5731, opened 2020-05-29
Re-enable copying attribute values across types

flyingsilverfin:
Allow copying an attribute value across types, using a rule. This was enabled in the past, but at some point between 1.6.2 and 1.7.1, a refactor accidentally caused this feature to be lost. This PR re-adds it by relaxing overly-strict constraints in `AttributeSemanticProcessor`, and adds a test for this case.

* Relax unifier computation between Attribute atoms, allowing unification without applying any constraints to the attribute variable themselves -- we assume the type is compatible due to commit time rule validation
* Add a test for copying attribute values across unrelated types

### DIFF
--- a/test/integration/graql/reasoner/stubs/resourceAttachment.gql
+++ b/test/integration/graql/reasoner/stubs/resourceAttachment.gql
@@ -31,7 +31,7 @@ derived-resource-string sub derivable-resource-string, value string;
 resource-long sub attribute, value long;
 derived-resource-boolean sub attribute, value boolean;
 subResource sub reattachable-resource-string;
-unrelated-reattachable-string sub derivable-resource-string, value string;
+unrelated-reattachable-string sub attribute, value string;
 
 #Rules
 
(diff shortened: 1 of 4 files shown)


test/integration/graql/reasoner/stubs/resourceAttachment.gql
```diff
@@ -30,6 +31,7 @@ derived-resource-string sub derivable-resource-string, value string;
 resource-long sub attribute, value long;
 derived-resource-boolean sub attribute, value boolean;
 subResource sub reattachable-resource-string;
+unrelated-reattachable-string sub derivable-resource-string, value string;
```
jmsfltchr, line 34: I suggest that this should `sub attribute` to be absolutely sure that it works outside of a hierarchy. At present the two types in the new test share a parent

flyingsilverfin: sorted, still works!

PR #5736, opened 2020-06-03
Merge changes from 1.7.2

adammitchelldev:

We made some changes in 1.7.2 to add GRPC back pressure in order to address some client issues with server overloading. Without back-pressure, excessive async client requests (possible in the new client releases) could cause the server to become overloaded and run out of memory, taking down the entire server. Since Grakn currently does not have automatic restart or recovery features, this is considered a serious issue so this fix was patched into 1.7.2 during the 1.8 development cycle and now needs to be merged into 1.8

- Merge change from 1.7.2: Add back-pressure by blocking grpc transaction service RequestObserver whilst server is still processing the last request.


flyingsilverfin:
@adammitchelldev can we add a more through explanation in the PR please? As this is a fundamental, but small, change in the behaviour, we should outline clearly not only why we're doing the change (server could OOM crash!!) but how the clients will behave differently (don't forget to collect your futures once in a while, or client will crash instaed!)

haikalpribadi:
The PR description also needs to be improved for the release notes that this will be included in. Don't forget the tags and milestone.

PR #5738, opened 2020-06-03
Update for new client-java future syntax

adammitchelldev:

For 1.8, `client-java` syntax has changed to return a `Future` from query methods. This requires us to update on our end.

- Updated to new `client-java` syntax


flyingsilverfin:
Changes going to be included during console upgrade in https://github.com/graknlabs/grakn/pull/5744

PR #5751, opened 2020-06-10
Fix too eager query completeness acking for queries with negation

kasper-piskorski:
Fix an occasional reasoning completeness issue encountered when combining negation with other rules.

We track query completeness by explicitly acknowledging that specific queries were exhaustively explored once all iterations were completed in `ResolutionIterator`. To do that we were using the `subGoals` - queries that were visited during query execution. This approach is however not correct when we deal with queries with negation. Each negation query has its own `ResolutionIterator` however the subgoals are injected from the iterator that triggered the negation query. As a result, if the negation query does an exhaustive search which happens when no answers can be found - it can acknowledge completion of queries that were injected but not fully completed. Due to this eager acknowledgment, queries that were actually not exhaustively explored will be omitted during reasoning. We fix this issue in this PR by explicitly tracking queries explored exhaustively within a specific `ResolutionIterator` and using them insted of `subGoals` to ack their completion once all iterations have finished

- `subGoals` are no longer stored in `ResolutionIterator` and used for query completeness acking
- instead we track exhaustive exploration of `AtomicState`s which corresponds to atomic queries that have been explored exhaustively


flyingsilverfin:
The cast is kinda ugly but eh

Issue #5719, opened 2020-05-11
Migrate Graql Native Integration Tests to BDD


flyingsilverfin:
## Problem to Solve
Currently, there are quite a few tests in Grakn Core that are "Graql", ie language tests - these should be moved to BDD and integrated into the `Axiomatic` tests - basic language behaviour tests found at https://github.com/graknlabs/verification/tree/master/behaviour/graql/language.

## Proposed Solution
Tests to migrate from Grakn Core to BDD and implemented:

* `graql/executor`
  * `DirectIsaIT.whenInsertIsaExplicit_InsertsADirectInstanceOfAType`
  * `DirectSubIT.directSubReturnsOnlyItselfAndDirectChildTypesWithoutReasoning`
  * `DirectSubIT.directSubReturnsOnlyItselfAndDirectChildTypesWithReasoning`
==> can be merged into Match BDD

* `graql/query/pattern`
  * `PatternIT.whenValueComparisonIsUnbound_weThrow`
  * `PatternIT.whenStatementDoesntHaveProperties_weThrow`
  * `PatternIT.whenMatchingWithValueInequality_resultsAreFilteredCorrectly`
==> can be merged into Match BDD

* `graql/query/GraqlDefineIT`
  * ALL tests migratable to Define BDD

* `graql/query/GraqlDeleteIT`
  * `afterDeletingAllInstances_TheTypeCanBeUndefined` -> Undefine BDD
  * ALL others can be moved to Delete BDD

* `graql/query/GraqlGetIT`
  * ALL to Get BDD

* `graql/query/GraqlInsertIT`
  * ALL to Insert BDD

* `graql/query/GraqlUndefineIT``
  * ALL to Undefine BDD

* `graql/query/NumberCastingIT`
  * ALL to Attribute Query BDD (might find better name that doesn't clash with Concept API)

* `graql/query/QueryErrorIT`
  * ALL to Error BDD

* `graql/query/QueryValidityIT`
  * Tests that are not designed to throw, merge into Match BDD
  * Tests that are designed to throw, merge into QueryError BDD

Across all of these, it will be interesting to see how to run the Write queries, then commit, and then roll back the state.
May be unucessary in BDD, as we should load a new keyspace or 
Also, many depend on MovieGraph, which may need to be shortened/adapted for BDD!

flyingsilverfin:
Also to migrate are probably the majority of `test/integration/concept`

alexjpwalker:
The migration is complete

PR #5760, opened 2020-06-11
Migrate Graql Insert ITs to BDD

alexjpwalker:

Where possible, we would like our Graql `insert` tests to be migrated to BDD.

This PR forms part of the work towards #5719 .

This PR will gradually delete test cases from GraqlInsertIT as they are tested in BDD.

This PR should NOT be merged until verification is updated with the new BDD scenarios, which is the desired outcome of https://github.com/graknlabs/verification/pull/54.


test/integration/graql/query/GraqlInsertIT.java
```diff
@@ -275,46 +151,11 @@ public void whenInsertingAResourceWithMultipleValues_Throw() {
...
-                var().isa("a-new-type").has("a-new-attribute-type", "hello").has("a-new-attribute-type", "goodbye")
-        ));
-
-        exception.expect(InvalidKBException.class);
-        tx.commit();
-    }
-
     @Ignore // TODO: Un-ignore this when constraints are designed and implemented
     @Test
     public void testKeyUniqueValue() throws InvalidKBException {
```
flyingsilverfin, line 161: let's have a look at the intent of this test?

test/integration/graql/query/GraqlInsertIT.java
```diff
@@ -334,118 +175,6 @@ public void testKeyUniqueValue() throws InvalidKBException {
...
-                allOf(containsString("$y"), containsString("id"), containsString("isa"), containsString("sub"))
-        );
-        tx.execute(Graql.insert(
-                var().rel("genre-of-production", "x").rel("production-with-genre", "y").isa("has-genre"),
-                var("x").isa("genre").has("name", "drama")
-        ));
-    }
-
-    @Test
-    public void testErrorWhenAddingInstanceOfConcept() {
```
flyingsilverfin, line 434: is this migrated?

alexjpwalker:
```
  Scenario: attempting to insert an instance of type 'thing' throws an error
    Then graql insert throws
      """
      insert $x isa thing;
      """
    Then the integrity is validated
```

test/integration/graql/query/GraqlInsertIT.java
alexjpwalker: As it turns out, it's just testing key uniqueness, which we have a BDD scenario for already, so this IT will be deleted.

alexjpwalker:
Closing due to the diff bug

PR #5761, opened 2020-06-12
Fixed implementation of Thing.keys(types)

haikalpribadi:

There was a bug in `ThingImpl.keys(AttributeType<?>... attributeTypes)`, where it would always return all instances of all key types, even when the user provided a subset of the key types in the argument. Additionally, the implementation can be made more readable and efficient, which could have contributed to why the bug existed in the first place.

1. Make sure we get the "edge case" or "easy logic" out of the way quickly. I.e. `if (attributeTypes.length == 0)` get that out of the way, then continue with the rest of the body of the code.
2. We need to use `keyTypes` multiple times, so we get that at the very top.
3. We don't want to convert collections too many times. So rather than converting from `array` to `list` to `set`, we just convert from array to set using `Collections.addAll(set, array)`.
4. Rather than consuming a stream twice, we can just do a `set` operation `set.retainAll()`. It's possible that it could be faster, but it would definitely make the code much clearer on its intention - which is to "retain all".
5. Unrelated to this method, there's the expression of `!<stream>.anyMatch()` that can be expressed as `<stream>.noneMatch()` to improve readability.

### DIFF
--- a/concept/impl/ThingImpl.java
+++ b/concept/impl/ThingImpl.java
@@ -36,6 +36,7 @@ import grakn.core.kb.concept.structure.VertexElement;
 import org.apache.tinkerpop.gremlin.structure.Direction;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


haikalpribadi:
@flyingsilverfin @alexjpwalker do we have a test for this in `@graknlabs_verification`?

alexjpwalker:
> flyingsilverfin alexjpwalker do we have a test for this in `@graknlabs_verification`?

I don't think Graql currently has an equivalent syntax - `match $x key $k;` is not supported.

PR #5779, opened 2020-06-20
Refactor daemon CLI to use picocli

adammitchelldev:

See related PR: https://github.com/graknlabs/console/pull/44

Right now our CLI architecture is fragmented, actual invocation is performed by a shell script which switches between console and server and chooses how to execute the console application. This means we can't do more advanced parameter suggestion, help and tab-completion generation etc. without severe code duplication.

What we'd like to do instead is to unify our CLI architecture across all repos into one Java architecture which can generate the full command tree and invoke subcommands naturally within the same JVM process (and hopefully across processes). This will provide extra benefits when moving to a CLI infrastructure that can install and update individual grakn components independently, and will greatly improve KGMS integration with regular grakn.

Picocli provides us with this CLI architecture, not only does it do a lot of heavy lifting that we were doing before manually (or with apache cli), it allows us to treat one command class from one project as a subcommand class in another project at runtime. It also allows us to extend and mix commands at runtime, useful for enabling new commands exclusive to KGMS whilst still having a single command set shared across both.

- Refactored `GraknDaemon`, the CLI behind `grakn server`, to use picocli.
- Added a `--no-logo` option to `grakn server` for people who hate fun.
- Added a `-y/--yes` option to `grakn server clean` to override the default confirmation prompt (useful for scripts).


lolski:
Closing this PR since we no longer have `GraknDaemon` on 2.0

PR #5780, opened 2020-06-21
CLI architecture refactor, implement top-level command with Java CLI

adammitchelldev:

Prior to this change, the architecture for our CLI subcommands was fragmented and making changes to them or adding new ones would be difficult. Top level subcommands required a bash script to decide between the `daemon` and `console` binaries, since each had a different set of dependencies.

After this change, the bash script will no longer handle CLI logic, all logic is now in a top-level binary using picocli, which we can recursively use to automatically build a subcommand tree across multiple binaries and validate arguments.

This change introduces runtime dependency loading, loading a small set of dependencies in order to build the command structure and only loading the full set for a subcommand when that subcommand is specifically executed. The result is that we can choose which binary to "become" after the process has already launched, avoiding expensive JVM relaunches.

The biggest useful takeaway is that we are now able to code this top-level management in "Java-land", so our Java developers can write more advanced CLI tools without needing to drop down to the bash script (which can't be cross-platform). This is a step closer to our CLI tools being able to manage multiple instances of Grakn, and help upgrade, migrate, backup and otherwise maintain Grakn installations.

- Added the `grakn.core.cli` module, which contains the top-level `grakn` command and the dependency loader.
- Added a runtime dependency loader, which dynamically loads the correct jars depending on which command is invoked.
- Added some prototype `community` commands because picocli is too easy to use.

## Problems to sort out with this PR

- Currently depends on changes to `GraknDaemon`, `GraknConsole` and `graknlabs/common`.
- Dependency loading does not play too nicely with slf4j, and will likely break in some cases that components use static initializers in classes loaded when creating their commands. This should technically be something avoided by the components, but we need to find a way to test this integration, enforce this, and allow definitions to load some additional dependencies where necessary.


adammitchelldev:
An interesting behaviour to note is that the implementation deals fine when certain components are not installed, and even excludes them from the help menus. This is fine for now since we have static distributions, but in the future I think we should manage installations of components from this tool.

adammitchelldev:
Another issue right now is that this change needs to be propagated to all build targets across all platforms, which is some tedious work.

PR #5783, opened 2020-06-23
Remote building with BuildBarn

vmax:

As Google is going to shut down hosted solution for remote builds (RBE), we're migrating to our own self-hosted solution, BuildBarn. This PR updates URLs to point at it instead of RBE.

Update `.bazelrc` to new URLs


vmax:
The experiment with BuildBarn in considered to be unsuccessful, so we're closing the PR.

PR #5785, opened 2020-06-23
Migrate Graql Undefine ITs to BDD

alexjpwalker:

Where possible, we would like our Graql `undefine` tests to be migrated to BDD.

This PR forms part of the work towards #5719 .

This PR will gradually delete test cases from GraqlUndefineIT as they are tested in BDD.

This PR should NOT be merged until verification is updated with the new BDD scenarios, which is the desired outcome of https://github.com/graknlabs/verification/pull/54.


alexjpwalker:
Once https://github.com/graknlabs/verification/pull/61 is merged, update `dependencies.bzl` in this PR to point at the HEAD revision of `verification`, then merge this PR.

PR #5786, opened 2020-06-23
Add dependency on console distribution

adammitchelldev:

Part of the effort to use distributable artefacts instead of building executable dependencies through source from other repos (where the dependencies might leak).

- Depend on distribution version of console


.circleci/config.yml
```diff
@@ -438,6 +450,18 @@ jobs:
           bazel run @graknlabs_build_tools//ci:release-validate-deps -- \
             graknlabs_common graknlabs_graql graknlabs_protocol graknlabs_console
```
haikalpribadi, line 451: This should be updated to take in `@graknlabs_console_distribution` now, and make sure it still works.

dependencies/graknlabs/BUILD
```diff
@@ -13,4 +13,6 @@
 #
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
-#
\ No newline at end of file
+#
+
+exports_files(["graknlabs_console.BUILD"], visibility = ["//visibility:public"])
```
haikalpribadi, line 18: new line

.gitignore
```diff
@@ -59,4 +59,8 @@ grakn-test/test-integration/benchmarks/
 
 
 # Python cache #
-__pycache__
\ No newline at end of file
+__pycache__
+
+
+# bazel-local #
+.local_deps*
```
haikalpribadi, line 66: new line

dependencies/graknlabs/BUILD
haikalpribadi, line 18: Why do we need this exported file, @adammitchelldev ? It doesn't show up in this PR either.

adammitchelldev: Ah, was for a previous method, good catch

Issue #5791, opened 2020-06-25
The value type of an attribute can be overridden in its subtype


alexjpwalker:
## Description

The value type of an attribute can be overridden in its subtype.

## Environment

1. OS (where Grakn server runs): Mac OS 10
2. Grakn version (and platform): Grakn Core 1.8.0
3. Grakn client: client-java

## Reproducible Steps

1. `define name sub attribute, value string;`
2. commit
3. `define code-name sub name, value long;`

## Expected Output

Should throw an error

## Actual Output

Succeeds.
 
## Additional information

Documented by BDD scenario: **Scenario: define attribute subtype throws if you try to override 'value'**

alexjpwalker:
The relevant BDD scenario now passes.

PR #5799, opened 2020-06-29
Don't allow a value type to be defined on a sub-attribute that does not match the parent's value type

alexjpwalker:

To not allow a value type to be defined on a sub-attribute that does not match the parent's value type (fixes https://github.com/graknlabs/grakn/issues/5791)


dependencies/graknlabs/dependencies.bzl
```diff
@@ -62,8 +62,8 @@ def graknlabs_simulation():
 def graknlabs_verification():
     git_repository(
         name = "graknlabs_verification",
-        remote = "https://github.com/graknlabs/verification",
-        commit = "f20731dfe367c53dec67ef098d8dc081632ccfd9",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_verification
+        remote = "https://github.com/alexjpwalker/verification",
```
alexjpwalker, line 65: Update to graknlabs when verification is merged

alexjpwalker:
We are no longer actively working on this repository, so this PR has become redundant. It will need to be addressed in 2.0

PR #5801, opened 2020-06-29
Ensure that two double attributes inserted with the same long value create a single concept

alexjpwalker:

Previously, given: `define length sub attribute, value double`, when you did:

```
insert
$x 2 isa length;
$y 2 isa length;
```

**two** concepts were created instead of one. Now, only a single concept is created.

The value of a `double` attribute needs to be normalised in order to accurately check for equality. This was handled effectively by the `AttributeValueConverter`, which provides a mechanism for taking in an attribute value and its type, and returning the value in a normalised form that ensures, for example, that `2`, `2.0` and `2.000` are all converted to the same output value because they represent the same value.

However, the `AttributeValueConverter` was only used in a small fraction of all the instances where we need to generate the "unique index" used by JanusGraph for deduplication. The method used to generate this index is at a very high level, in `Schema.java`, and cannot be easily replaced. Therefore, this PR aims to address the issue by ensuring that we pass in the correct `value` parameter to `Schema.java` every time, by consistently using `AttributeValueConverter.tryConvert` to convert the attribute value to normal form before it is used.

Also, it has a try-catch block in it which ensures we consistently throw the same error if the conversion could not be performed (e.g: attempting to parse "banana" as a `long`).

### DIFF
--- a/concept/manager/ConceptListenerImpl.java
+++ b/concept/manager/ConceptListenerImpl.java
@@ -20,6 +20,7 @@ package grakn.core.concept.manager;
 
 import grakn.core.concept.impl.RelationImpl;
 import grakn.core.concept.impl.ThingImpl;
+import grakn.core.core.AttributeValueConverter;
 import grakn.core.core.Schema;
 import grakn.core.kb.concept.api.Attribute;
 import grakn.core.kb.concept.api.AttributeType;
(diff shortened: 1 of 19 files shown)


concept/manager/ConceptListenerImpl.java
```diff
@@ -135,20 +136,19 @@ private void thingCreated(Thing thing, boolean isInferred) {
...
-        attributeManager.ackAttributeInsert(index, txId);
+        attributeManager.ackAttributeInsert(uniqueIndex, txId);
     }
 
     private <D> void attributeDeleted(Attribute<D> attribute) {
-        Type type = attribute.type();
+        AttributeType<D> type = attribute.type();
         //Track the attribute by index
-        String index = Schema.generateAttributeIndex(type.label(), attribute.value().toString());
+        String index = Schema.generateAttributeIndex(type.label(), AttributeValueConverter.tryConvert(type, attribute.value()).toString());
```
flyingsilverfin, line 151: again, I don't think we need to apply the (potentially costly? not sure!) conversion everywhere - only where the user provides a value, it comes from a query, or a rule, etc! So I think this and the following cases can be kept as before as the value comes from a pre-existing `Attribute` instance :)

kb/server/cache/TransactionCache.java
```diff
@@ -154,7 +155,7 @@ public void remove(Concept concept) {
             Attribute<?> attr = concept.asAttribute();
             // this is probably slower than reading index from vertex but we don't have access to AttributeImpl here (cyclic dep)
             Label attrLabel = attr.type().label();
-            String attrIndex = Schema.generateAttributeIndex(attrLabel, attr.value().toString());
+            String attrIndex = Schema.generateAttributeIndex(attrLabel, AttributeValueConverter.tryConvert(attr.type(), attr.value()).toString());
```
flyingsilverfin, line 158: here too

alexjpwalker: Yeah this one is definitely unnecessary on second thoughts as the concept was loaded from the KG.

concept/manager/ConceptListenerImpl.java
alexjpwalker: I'm not so convinced about this one. That `Attribute` could have come from many places - going up the stack, we have:

- ConceptNotificationChannel.hasAttributeCreated
- ConceptManager.createHasAttribute
- ThingImpl.attributeOwnership
  - Thing.has
    - HasAttributeExecutor.InsertHasAttribute ...
    - ConceptMethod.run
  - Thing.attributeInferred
    - AttributeMaterialiser.putAttributeOwmership
    - AttributeMaterialiser.materialise
    - AttributeAtom ...

kb/server/cache/TransactionCache.java
alexjpwalker: Reverted

alexjpwalker:
After discussing with Joshua, we will not be merging this PR into master, but rather into 1.9.

PR #5802, opened 2020-06-29
Scoped roles

flyingsilverfin:

Implement scoped roles in Grakn Core, not including inheritance or overriding. Only relation types can be used to create new roles (eg `RelationType.role(String name)`). However, roles can be retrieved and utilised using the full scope (`tx.getRole("spouse", "marriage")`).

- `grakn.core...Label` is removed, replaced with `graql.lang...Label`
- roles are created from RelationType instances
- writing scoped roles treates `label` (unscoped name) and `scope` as two separate properties to add to a concept. This allows us to semi-elegantly merge information from the `relation` statement: `($r: $x) isa marriage) into the concept builder for `$r`, and include information from `$r`: `$r type spouse`. The result is creating the full scoped role `marriage:spouse`


flyingsilverfin:
Conclusion: we will delay this until post-integration of hypergraph, implementing restrictions on roles belonging to a particular relation and implemeting role inheritance first.

PR #5804, opened 2020-06-30
Remove bazel deps

lolski:

We are replacing `bazel-deps` with the new `maven()` rule that is based on `rules_jvm_external`. This new rule is designed in such a way that allows us to centralise dependency versioning in a single repository, `@graknlabs_dependencies`.

Use Maven artifacts that are declared in `graknlabs_dependencies` rather than `dependencies.yaml`


lolski:
testing test deployments and test assemblies by enabling them temporarily on this branch and they seem to pass as well: https://app.circleci.com/pipelines/github/graknlabs/grakn/1384/workflows/84096710-a26a-43d4-bf73-abaced1aff6f

PR #5806, opened 2020-06-30
BuildBuddy and beyond!

siggisim:
Migrating repo to use BuildBuddy Cloud RBE.

The only things you'll need to do on your end is upgrade CI's Bazel version to 3.2.0 download the auth certs and place them on your CI machines:
1. Signing in or creating an account with an @grakn.ai email address here: https://app.buildbuddy.io/
2. Downloading buildbuddy-cert.pem and buildbuddy-key.pem from here by clicking on the "Certificate" button then "Download buildbuddy-cert.pem" and "Download buildbuddy-key.pem: https://app.buildbuddy.io/docs/setup/
3. Put these two files at the root of your workspace on your CI machines (similar to how you placed the Google auth credentials file).

API key approach can also be used if there's trouble updating the Bazel version or placing certificate files. Flags containing API key can be passed into CI only and will override flags in your .bazelrc.

Please don't hesitate to reach out if you have any questions at all - or if there's any way we can be helpful.


vmax:
Superseded with #5810 as we need to test with variables set in our forks
@siggisim I preserved your authorship on the commits

PR #5810, opened 2020-07-02
BuildBuddy and beyond!

vmax:
Migrating repo to use BuildBuddy Cloud RBE.

The only things you'll need to do on your end is upgrade CI's Bazel version to 3.2.0 download the auth certs and place them on your CI machines:
1. Signing in or creating an account with an @grakn.ai email address here: https://app.buildbuddy.io/
2. Downloading buildbuddy-cert.pem and buildbuddy-key.pem from here by clicking on the "Certificate" button then "Download buildbuddy-cert.pem" and "Download buildbuddy-key.pem: https://app.buildbuddy.io/docs/setup/
3. Put these two files at the root of your workspace on your CI machines (similar to how you placed the Google auth credentials file).

API key approach can also be used if there's trouble updating the Bazel version or placing certificate files. Flags containing API key can be passed into CI only and will override flags in your .bazelrc.

Please don't hesitate to reach out if you have any questions at all - or if there's any way we can be helpful.


.bazelrc
```diff
@@ -6,10 +6,12 @@ run --show_result=0 --show_progress=no --incompatible_strict_action_env
...
-build:rbe --remote_cache=remotebuildexecution.googleapis.com
-build:rbe --remote_executor=remotebuildexecution.googleapis.com
-build:rbe --bes_backend="buildeventservice.googleapis.com"
-build:rbe --bes_results_url="https://source.cloud.google.com/results/invocations/"
+build:rbe --remote_cache=cloud.buildbuddy.io
+build:rbe --remote_executor=cloud.buildbuddy.io
+build:rbe --bes_backend=cloud.buildbuddy.io
+build:rbe --bes_results_url=https://app.buildbuddy.io/invocation/
+build:rbe --tls_client_certificate=/home/circleci/.credentials/buildbuddy-cert.pem
+build:rbe --tls_client_key=/home/circleci/.credentials/buildbuddy-key.pem
```
lolski, line 14: Can we change this path to be relative, @vmax ? eg. `.credentials/buildbuddy-cert.pem`

flyingsilverfin:
don't feel particularly qualified to review this, will leave it to ganesh!

.bazelrc
vmax: @lolski we'd have to switch `checkout` and `install-bazel-linux` steps (you cannot `checkout` into non-empty directory) as well as modify `@graknlabs_dependencies`

PR #5811, opened 2020-07-03
Allow tests to print stack traces

flyingsilverfin:
Allow tests to print stack traces -- this should automatically apply to all `GraknExceptions` and its subtypes.

Re-enable stack traces during DEBUG level logging, and update test logback to use DEBUG by default.


lolski:
Approved. Had a short discussion with @flyingsilverfin whether we DEBUG will be too verbose. However, we've agreed that it won't.

PR #5814, opened 2020-07-06
Reduce code duplication between graql query steps in BDD

alexjpwalker:

To reduce code duplication between graql query steps in BDD.

We took the steps "graql define", "graql insert", "graql delete" etc. and refactored them, extracting a common method named "graql query". We repeated this process for "graql define without commit", and "graql define throws", and for the other query types.


haikalpribadi:
Does the failing test need to be rerun, @alexjpwalker ?

alexjpwalker:
> Does the failing test need to be rerun, @alexjpwalker ?

Most likely yes, triggered a rerun.

PR #5815, opened 2020-07-06
Clarify intent of resolution BDD steps

alexjpwalker:

To perform a code cleanup of the resolution BDD steps, setting a "second draft" of how we want the scenarios to look.

- Previously, the graql query steps involved were called "graql define" and "graql insert". This was confusing as they had the same names as the relevant steps in GraqlSteps, but performed different actions - they operated across multiple sessions simultaneously. We renamed them to "for each session, graql define", etc.
- The step: "answer count is: {n}" was too unclear. Unlike the graql queries which operated across multiple keyspaces, this stopgap manual test operates against only one keyspace. To clarify the meaning, we renamed it to "for reasoned keyspace, answer size is: {n}".
- We renamed 'completion' and 'test' to the clearer terms: 'materialised' and 'reasoned' .
- We initialised two keyspaces in the background of each scenario, but it was not at all clear how these were used to derive which one was the 'test' keyspace and which one was the 'completion' one. Naturally, one might assume that it is based on the keyspace name; but in fact, it was simply based on the order of arguments passed to the "connection open sessions for keyspaces" step. This has now been refactored and split into multiple steps, which will look as follows in an actual scenario file:

```
Given connection open sessions for keyspaces:
      | materialised |
      | reasoned     |
Given materialised keyspace is named: materialised
Given reasoned keyspace is named: reasoned
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -155,7 +155,8 @@ jobs:
           command: bazel test //test/behaviour/graql/language/delete:test --test_output=errors
 
   test-behaviour-graql-get:
-    machine: true
+    machine:
+      image: ubuntu-1604:201903-01
     working_directory: ~/grakn
     steps:
       - install-bazel-linux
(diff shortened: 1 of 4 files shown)


flyingsilverfin:
Ready to merge once the verification steps have been updated in `graknlabs/verification`

alexjpwalker:
Merge https://github.com/graknlabs/verification/pull/73 and update dependencies.bzl before merging this PR

PR #5820, opened 2020-07-09
Catch all exceptions thrown in tracing client

lriuui0x0:

Currently, not all exceptions thrown in tracing handling are caught and close the connection properly. Some exceptions are thrown but swallowed, which means we don't see any error reporting when an exception happens.

Raise the closing connection logic to a higher level so that whenever an exception is thrown, we always close the connection.


flyingsilverfin:
classic, got the smorgasbord  of failures :P

PR #5823, opened 2020-07-09
Allow resolution tests to operate without keys on everything

alexjpwalker:

Previously, `isa [thing]` would fail in resolution tests if [thing] was an attribute type. We fixed it.

We added `plays instance` to attribute types in the resolution test framework, and added `match $x has $y; get;` to the completeness check.


test/behaviour/resolution/framework/common/RuleResolutionBuilder.java
```diff
@@ -119,34 +125,52 @@
         return Graql.and(result);
     }
 
-    public LinkedHashMap<String, Statement> statementToResolutionProperties(Statement statement) {
+    public LinkedHashMap<String, Statement> statementToResolutionProperties(final Transaction tx, Statement statement, final Boolean inferred) {
```
flyingsilverfin, line 128: why is this a Boolean instead of a boolean? Do we need a null on top of true/false?

dependencies/graknlabs/dependencies.bzl
```diff
@@ -62,8 +62,8 @@ def graknlabs_simulation():
 def graknlabs_verification():
     git_repository(
         name = "graknlabs_verification",
-        remote = "https://github.com/graknlabs/verification",
-        commit = "15e9132c9cc5d173edd81afe22ac60d04a1abe02",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_verification
+        remote = "https://github.com/alexjpwalker/verification",
```
flyingsilverfin, line 65: need to revert this before merging

test/behaviour/resolution/ResolutionSteps.java
```diff
@@ -97,6 +97,11 @@ public void reasoned_keyspace_all_answers_are_correct() {
         resolution.testResolution(queryToTest);
     }
 
+    @Then("no answers are resolved in reasoned keyspace")
```
flyingsilverfin, line 100: this could either be interpreted as no answers are found (ie, answer size is 0), or that there are no inferred anwers that used reasoning (there may be answers that come form the DB. Let's make sure that the step and name of the methods being callsed are clear about the one thing they do :)

test/behaviour/resolution/framework/Resolution.java
```diff
@@ -154,5 +158,32 @@ public CorrectnessException(String message) {
...
+
+        public int getExpectedAnswers() {
+            return expectedAnswers;
+        }
+
+        public GraqlQuery getInferenceQuery() {
+            return inferenceQuery;
+        }
+
+        private final int expectedAnswers;
```
flyingsilverfin, line 185: we tend to put state at the top of the class, then constructors, public methods, last private methods :)

test/behaviour/resolution/framework/common/RuleResolutionBuilder.java
```diff
@@ -119,34 +125,52 @@
...
 
                     String nextVar = getNextVar("x");
 
-                    StatementInstance propStatement = Graql.var(nextVar).isa("relation-property").rel("rel", statementVar).rel("roleplayer", Graql.var(rolePlayer.getPlayer().var()));
-                    if(role.isPresent()) {
+                    StatementInstance propStatement = Graql.var(nextVar).isa("relation-property").rel("rel", statementVarName).rel("roleplayer", Graql.var(rolePlayer.getPlayer().var()));
+                    if (role.isPresent()) {
                         String roleLabel = ((TypeProperty) getOnlyElement(role.get().properties())).name();
-                        propStatement = propStatement.has("role-label", roleLabel);
+                        final Set<Role> roles = tx.getRole(roleLabel).sups().collect(Collectors.toSet());
```
flyingsilverfin, line 154: we might be able to use a .forEach loop here? not sure because we modify state though hummm. might be best approach already

test/behaviour/resolution/framework/common/RuleResolutionBuilder.java
alexjpwalker: yes - null is passed in as an argument in some places.

flyingsilverfin: if you need to you can do `my_method(@Nullable Boolean my_arg_that_can_be_null)`

test/behaviour/resolution/framework/Resolution.java
alexjpwalker: Rearranged.

test/behaviour/resolution/framework/common/RuleResolutionBuilder.java
alexjpwalker: Yeah I think I prefer the for loop because of the state modification.

test/behaviour/resolution/ResolutionSteps.java
alexjpwalker: We decided to delete the step, as it actually asserts the same thing as the step below it: `answer size in reasoned keyspace is: 0`.

dependencies/graknlabs/dependencies.bzl
alexjpwalker: This will be reverted in a separate PR later - it's tricky to remove at the moment because the PRs have got a bit mixed up.

lolski:
Merging the PR despite the fact that one of the test failed, according to a discussion with @flyingsilverfin and @alexjpwalker .

The reason being that CircleCI is currently down and the test can't be re-ran.

PR #5827, opened 2020-07-13
Refactor atomic and benchmark Reasoner tests

flyingsilverfin:
In preparation of extensive role-behaviour changes, we refactor  reasoner's tests to not use shared roles across relations, and let them be written in a way that is easier to update for future changes: scoping, `as` role overriding, and inheritance of `relates`. The changes include using real-world schema examples and bringing schemas into the test classes, which helps reduce complexity as large parts of each schema is not reused across tests.
We also add a flag to `ReasonerQueryFactory` to disable type inference, which is useful for reducing the scope of tests.

* Refactor tests in `//test/integration/graql/reasoner/atomic` to not use external schemas
* refactor tests to use more accessible schema terms
* fix bug related to pruning rules when a role player ID is present that cannot play a required role
* add flag to enable/disable role inference in `ReasonerQueryFactory`


graql/reasoner/atom/binary/RelationAtom.java
```diff
@@ -453,7 +453,9 @@ private boolean isTypeRoleCompatible(Variable typedVar, Type parentType, boolean
 
         if (roleRequirements.isEmpty()) return true;
 
-        Set<Type> parentTypes = parentType.subs().collect(Collectors.toSet());
+        // in some cases, the parent types are not specified by ISA but by ID directly - then we do no have to search subtypes for compatibility
+        Set<Type> parentTypes = parentTypeIsExact ? Sets.newHashSet(parentType) : parentType.subs().collect(Collectors.toSet());
```
kasper-piskorski, line 457: You added a boolean and pick the `isa` type if id is present. What happens in the following scenario:
```
($x, $y) isa someRelation;" +
$x id instanceIdOfSubTypeThatCanPlayRolesOfX;
$x isa superTypeThatCannotPlayRolesOfX;
```
or the case when the sub-super types are inverted.

Also I've just realised that the MetaType filter at the top might be invalid -> it assumes all meta types can play all the roles which doesn't have to be the case (even for `entity`).

flyingsilverfin: solved by passing the exact type instead of a boolea !

Issue #5831, opened 2020-07-16
Grakn Core gets 5x slower with Updated Netty


flyingsilverfin:
## Description
Upgrading Netty from `4.1.38.Final` to `4.1.50.Final` leads to a 5x slowdown, similar to what is seen when upgrading gRPC to `1.29.0` from `1.24.1` (this is likely due to using a newer netty itself). 

## Environment
MacOS and Ubuntu 16.04
Grakn Core 1.8.x

## Reproducible Steps
Fast Grakn: 
1. check out commit `4a54aed1fc625f818762cba6f8b691e9e3210b82` (13 July 2020)

run a benchmark that uses a client

Slow Grakn:
2. check out the next commit `5d89c681d5127999a9035fcf85bf016ed32b5a41` (13 July 2020)
run the same benchmark.

## Expected Output
Roughly the same performance.

## Actual Output
Using the updated Netty is 5x slower
 
## Additional information
We also tested updating to gRPC `1.30.2` from `1.24.1`, which pulled in netty `4.1.50.Final` transitively (we removed our explicit netty import). This did not help, and was still very slow! We suspect that Netty changed a key functoinality gRPC utilises in a way that impacted performance -- gRPC may need to change the way it uses Netty to achieve the higher performance again.

flyingsilverfin:
The issue appears to have been raised at https://github.com/grpc/grpc-java/issues/6696
the fix is implemented and should be released in 1.30.3 of gRPC

lolski:
This issue is related https://github.com/graknlabs/grakn/issues/5769.

Should we merge these two issues @flyingsilverfin ?

flyingsilverfin:
We've successfully upgraded to Netty 4.1.63.Final and not noticed a performance regression.

PR #5832, opened 2020-07-16
Downgrade netty

flyingsilverfin:
To fix the issue found in https://github.com/graknlabs/grakn/issues/5831 that slows Grakn down drastically, we force downgrade Netty back to `4.1.38.Final`.

* Force downgrade netty to the prior version


flyingsilverfin:
@lolski  do you think we should comment out or delete netty deps? Or should this PR really be in the `dependencies` repo?

lolski:
I think we should delete the netty deps in `artifacts.bzl`

flyingsilverfin:
done

PR #5842, opened 2020-07-21
Add export to 1.8

adammitchelldev:

To add the ability to export and import date and schema to 1.8 grakn in order to support migration into newer versions of grakn (1.8 and beyond).

These features include:
* Exporting all of the data in a keyspace to a file.
* Exporting the schema of a keyspace.
* Importing an exported data file into a keyspace.

We have added these features to the `grakn server` command, they should be run local to grakn server instance.

```
grakn server export [--no-anim] <keyspace> <file>.grakn
grakn server import [--no-anim] <keyspace> <file>.grakn
```

These can be used as a backup solution but are mostly designed to simplify the process of migrating data across Grakn versions.

In this 1.8 version, we have also added label remapping on import, to solve issues where a schema may now use labels that are reserved keywords.

- Add schema export
- Add data export
- Add data import
- Add data import label remapping


server/migrate/proto/migrate.proto
```diff
@@ -0,0 +1,84 @@
...
+    }
+}
+
+message ExportSchema {
+    message Req {
+        string name = 1;
+    }
+
+    message Res {
+        string schema = 1;
```
flyingsilverfin, line 82: Might be worth noting that there is a string limit size of i think 4mb messages? I think this might be about 4 million characters uncompressed, hopefully we don't actually hit this limit at any point!

server/session/TransactionImpl.java
```diff
@@ -1178,6 +1180,10 @@ public void commit() throws TransactionException {
         }
     }
 
+    public void disableCommitValidation() {
```
flyingsilverfin, line 1183: did we have a discussion on how we can do validation since its all disabled? I think this needs some more thought

flyingsilverfin:
Ah! Also, what do you think of us checking that the total count of data in the database before insert is 0? Ie. we only allow import into an empty keyspace :)

.circleci/config.yml
```diff
@@ -22,14 +22,14 @@ orbs:
 commands:
   install-bazel-linux-rbe:
     steps:
-      - run: curl -OL https://raw.githubusercontent.com/graknlabs/build-tools/master/ci/install-bazel-linux.sh
+      - run: curl -OL https://raw.githubusercontent.com/graknlabs/build-tools/715e4802ea375abda44e24b2bb092ee14de72e42/ci/install-bazel-linux.sh
```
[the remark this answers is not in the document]
adammitchelldev: The master branch of dependencies doesn't have these scripts, so we need to download them from an old version of the repo that was available at this time. I believe this commit is the same as the commit of `build-tools` that this repo depends on.

server/migrate/Export.java
```diff
@@ -0,0 +1,355 @@
...
+                    write(read(label, iterator.next(), ownedTypes));
+                }
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+
+            LOG.debug("||||| Finished: {}", label);
+        }
+
+        abstract protected DataProto.Item read(String label, U instance, AttributeType<?>[] ownedTypes);
```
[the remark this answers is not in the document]
adammitchelldev: It is `private` state, making it `protected` is possible but then less clear to implementers that it is expected that their implementation needs to use this value.

server/migrate/proto/migrate.proto
adammitchelldev: I highly doubt it, this should easily support schemas well into the 10's of thousands of schema concepts.

server/session/TransactionImpl.java
adammitchelldev: Uh, did I actually use this anywhere? I think I added it and then ended up not using it.

.circleci/config.yml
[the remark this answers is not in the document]
flyingsilverfin: oh!! okay, i think this PR should be being merged into Grakn Core master right? Not on a new branch. Everything should be normal - this isn't a patch!

server/migrate/Export.java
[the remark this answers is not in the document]
flyingsilverfin: ok just curious

server/session/TransactionImpl.java
flyingsilverfin: you do call it in `newTransaction()`

flyingsilverfin:
Oh and also adding in the PR description the command to print schema

server/session/TransactionImpl.java
adammitchelldev: No, I'm wrong, we do use it. I'm pretty sure all this is only disabling the role player count validation for us, not removing all validation.

lolski:
`--no-anim` is a name that describe what it is. Let's pick a name that describe what the animation is meant to display.

Is it for toggling the progress bar? If yes, what if we rename it to this:

```
grakn server export [--no-progress-bar] <keyspace> <file>.grakn
grakn server import [--no-progress-bar] <keyspace> <file>.grakn
```

adammitchelldev:
> `--no-anim` is a name that describe what it is. Let's pick a name that describe what the animation is meant to display.
> 
> Is it for toggling the progress bar? If yes, what if we rename it to this:
> 
> ```
> grakn server export [--no-progress-bar] <keyspace> <file>.grakn
> grakn server import [--no-progress-bar] <keyspace> <file>.grakn
> ```
> 
> If it is not a progress bar, still think of a name that better describe what it is meant to display rather than what it is

But it is meant to disable all animation, or as much animation as possible, so it also reduces the rate of the update so that it becomes a slow ticker rather than an animated output. It's not just disabling the progress bar.

PR #5849, opened 2020-07-24
Fix Docker test to be compatible with both Python 2 and Python 3

vmax:

Previously, it was only possible to run `test-assembly-docker` with `python` pointing to `python3`. In order to be compatible with both Python 2 and Python 3, test shouldn't use any new language features.

* Don't use `time.perf_counter`: it was introduced in 3.3
* Don't use `raise Exception() from ex` syntax
* Don't use `socket.create_connection` as context manager

### DIFF
--- a/test/assembly/docker.py
+++ b/test/assembly/docker.py
@@ -8,16 +8,16 @@ import time
 
 
 def wait_for_port(port, host='localhost', timeout=30.0):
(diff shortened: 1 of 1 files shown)


lolski:
@vmax Approved. Please create a ticket in Grabl to update it to be more like CircleCI: default to Python 3 rather than Python 2.

PR #5854, opened 2020-07-28
Remove INTEGER and FLOAT value types

flyingsilverfin:

  - its responsibility, 


flyingsilverfin:
TODO: this should merge into 1.9/2.0 branch because it breaks DB compatibility I believe

PR #5869, opened 2020-08-01
Reenable generative operational tests

### LINKED ISSUE #5847: Re-enable `GenerativeOperationalIT`
## Problem to Solve
The test `GenerativeOperationalIT` was commented out in https://github.com/graknlabs/grakn/pull/5843.
The following issues arose:
1. Some of the generalisations result in queries that are semantically invalid (not satisfiable in the given schema)
2. Generates queries with undefined unification requirements, such as  requiring some types to specialise and some to generalise. The query pair generation should only be generating queries that are either generalised or specialised versions of each other, but not both.

## Proposed Solution
Once we fix the issues, we should re-enable it as its a good large scale automated test.

kasper-piskorski:
Reenable the previously ignored `GenerativeOperationalIT`. Fixes #5847
- fixed an issue with `SUBSUMPTIVE_STRUCTURAL` unification when determining role player validity
- added basic filtering of semantically invalid patterns in `GenerativeOperationalIT` - now only semantically valid patterns are compared
- switched off type inference for subsumption tests in `GenerativeOperationalIT` as these can affect the unification outcome
- multithreaded the fuzzying tests for faster processing


graql/reasoner/query/ReasonerQueryFactory.java
```diff
@@ -60,12 +60,14 @@ public ReasonerQueryFactory(ConceptManager conceptManager, QueryCache queryCache
         this.ctx = new ReasoningContext(this, conceptManager, queryCache, ruleCache, statistics);
     }
 
-    public void enableInferTypes() {
+    public ReasonerQueryFactory enableInferTypes() {
```
flyingsilverfin, line 63: not sure i like this change, its not a builder really, or returning a new instance ? Not having the "fluid" pattern here is fine imo

graql/reasoner/atom/task/relate/RelationSemanticProcessor.java
```diff
@@ -155,11 +156,11 @@ public MultiUnifier getMultiUnifier(RelationAtom childAtom, Atom parentAtom, Uni
                     Set<Type> parentTypes = parentVarTypeMap.get(parentRolePlayer);
                     Type parentTypeExact = parentAtom.getPredicates(parentRolePlayer, IdPredicate.class)
                             .filter(idPredicate ->  !idPredicate.isPlaceholder())
-                            .findAny()
                             .map(idPredicate -> (Concept)ctx.conceptManager().getConcept(idPredicate.getPredicate()))
+                            .filter(Objects::nonNull)
```
flyingsilverfin, line 160: how can we have null IDs?

test/integration/graql/reasoner/query/GenerativeOperationalIT.java
```diff
@@ -266,66 +264,90 @@ public void whenGeneralisingAttributes_SubsumptionRelationHoldsBetweenPairs(){
...
+                            });
                 }
             }
+            return null;
+        };
+        processPatterns(new ArrayList<>(binaryRelationPatternTree.keySet()), func, 1);
+    }
+
+    private void processPatterns(List<Pattern> patterns, Function<List<Pattern>, Supplier<Void>> func, int threads) throws ExecutionException, InterruptedException {
+        ExecutorService executorService = Executors.newFixedThreadPool(4);
```
flyingsilverfin, line 333: will it be super slow if we just do one pattern per tx (tx will be super fast in 2.0) - then no risk of cache issues, and this code will be super simple:
```
patterns.parallelStream().map(...)
```

kasper-piskorski: Bit slower, nothing is executed here so there isn't much that is cached. Anyway, changing to simpler.

graql/reasoner/atom/task/relate/RelationSemanticProcessor.java
kasper-piskorski: ids are not null but a lookup can return null if it doesn't exist

test/integration/graql/reasoner/query/GenerativeOperationalIT.java
```diff
@@ -267,50 +263,45 @@ public void whenGeneralisingAttributes_SubsumptionRelationHoldsBetweenPairs(){
...
     @Test
     public void whenFuzzyingIdsWithBindingsPreserved_StructuralEquivalenceIsNotAffected(){
-        try (Transaction tx = genericSchemaSession.transaction(Transaction.Type.READ)) {
-            TestTransactionProvider.TestTransaction testTx = (TestTransactionProvider.TestTransaction) tx;
-            ReasonerQueryFactory reasonerQueryFactory = testTx.reasonerQueryFactory();
-            TransactionContext txCtx = new TransactionContext(tx);
-            Set<Pattern> patterns = binaryRelationPatternTree.keySet();
-            Operator fuzzer = Operators.fuzzIds();
-            for(Pattern pattern : patterns){
+        binaryRelationPatternTree.keySet().parallelStream().forEach(pattern -> {
```
flyingsilverfin, line 298: cool!

flyingsilverfin:
might need a rerun to pass

flyingsilverfin:
lol again

flyingsilverfin:
aaand again rerun??

kasper-piskorski:
much success!

PR #5880, opened 2020-09-02
Update credential paths and fix Grabl config

vmax:

Credential paths needs to be updated according to graknlabs/dependencies#196. In addition, Grabl config is also made up-to-date and cleaned up.

* Use `/opt/credentials/` as RBE credential path
* Use up-to-date image name for Grabl
* Stop fetching script that installs Bazel: it's already backed into the image


haikalpribadi (approved):

flyingsilverfin:
@vmax this will definitely fail - we cannot let core 1.8.x (this repo's master) depend on the `master` of dependencies - if you want to update it, you'll have to put a new commit on the `core-1.8` branch in `dependencies` repo, with the same changes as the one you linked.

I've mentioned this a couple times in the office, but probably not loudly enough over zoom:
all repos except user-facing ones (workbase, docs, examples, kglib) all have a `development` branch that is 2.0-ready (including updated build dependencies). Grakn core (this repo) is strictly 1.8.x, and uses old dependencies. All other `graknlabs` repos are 2.0-ready on their master branches and depend on each other, including `grakn-2.0`.

You'll find that in the repositories that this repo depends on (ie 1.8.x), we usually have a branch called `core-1.8` -- see `dependencies` for an example. Any changes we want make in the build system for this repo have to be added on top of that branch as well. The best thing to do would be to NOT update the dependencies of this repository at all, to avoid any headaches :)

vmax:
Got it, thanks @flyingsilverfin!

PR #5895, opened 2020-11-02
Fix to Schema Export so that rules are also exported

maxbaxt:

We complete the the schema export functionality, including missing rules in the printed schema when calling `./grakn export [keyspace] [filename].grakn`

Schema.java includes TypeHierachy of rules, and logic has been added to print rules in a consistent way.

schema_export_truth.gql updated in order to reflect new output

### DIFF
--- a/server/migrate/Schema.java
+++ b/server/migrate/Schema.java
@@ -88,6 +88,7 @@ public class Schema {
         printTypesHierarchically(writer, "attribute");
         printTypesHierarchically(writer, "entity");
         printTypesHierarchically(writer, "relation");
+        printTypesHierarchically(writer, "rule"); //This is the fix, I think
 
         tx.close();
     }
--- a/server/migrate/Schema.java
+++ b/server/migrate/Schema.java
@@ -88,7 +88,7 @@ public class Schema {
         printTypesHierarchically(writer, "attribute");
         printTypesHierarchically(writer, "entity");
         printTypesHierarchically(writer, "relation");
-        printTypesHierarchically(writer, "rule"); //This is the fix, I think
+        printTypesHierarchically(writer, "rule");
 
         tx.close();
     }
--- a/server/migrate/Schema.java
+++ b/server/migrate/Schema.java
@@ -333,13 +333,18 @@ public class Schema {
                 relates.print(writer);
             }
             if (when != null) {
+                writer.println(",");
                 when.print(writer);
             }
             if (then != null) {
                 writer.println(", then {");
                 writer.indent();
-                writer.println(then.toString());
+                String then_string = then.toString();
+                writer.println(then_string.substring(2, then_string.length() - 3));
                 writer.unindent();
+//                writer.indent();
+//                writer.println(then.toString().);
+//                writer.unindent();
                 writer.print("}");
             }
             writer.unindent();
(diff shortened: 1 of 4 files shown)


flyingsilverfin:
lets use a more reader-friendly wording in the description
```
Schema export now works on rules as well as the rest of the schema.
```
=>
```
We complete the the schema export functionality, including missing rules in the printed schema when calling `./grakn ....`
```

This way the "improvement" and the "problem" off the bat (in your description, it's not clear if rules were broken or missing for instance :)

server/migrate/Schema.java
```diff
@@ -154,6 +157,44 @@ public int hashCode() {
         }
     }
 
+    private class WhenDeclaration {
+        private final Set<Pattern> patterns;
+
+        WhenDeclaration(Pattern when) {
+            patterns = when.getNegationDNF().getPatterns().iterator().next().getPatterns();
```
flyingsilverfin, line 164: why do we need to do negationDNF?

test/assembly/data/schema_export_truth.gql
```diff
@@ -306,5 +306,110 @@ transaction sub relation,
     relates transaction_merchandise,
     relates transaction_vendor;
 
+born-in-location-implies-residency sub rulewhen {
```
flyingsilverfin, line 309: errr where did you get these rules? If the rest of the schema comes from our simulation, we should use those rules. If that's where these were sourced, are they supposed to be `sub rulewhen`? That looks like a bug to me... If this test passes i'm not sure that's correct!!

flyingsilverfin (changes requested):

test/assembly/data/schema_export_truth.gql
maxbaxt: That's not right. There's a println(",") missing after 'rule'. I've fixed that now.

server/migrate/Schema.java
maxbaxt: Because the other normal form gave a Set of Statements, but the statements did not distinguish between positive and negated statements.

server/migrate/Schema.java
```diff
@@ -293,15 +334,18 @@ public void print(IndentPrintWriter writer) {
...
-                writer.println(",");
-                writer.println("then ");
-                String ruleString = then.toString();
-                writer.print(ruleString.substring(0, ruleString.length() - 1));
+                writer.println(", then {");
+                writer.indent();
+                String then_string = then.toString();
+                writer.println(then_string.substring(2, then_string.length() - 3));
+                writer.unindent();
+//                writer.indent();
```
flyingsilverfin, line 345: btw are these meant to still be commented?

test/assembly/MigrationTest.java
```diff
@@ -95,14 +95,18 @@ public void testMigration() throws InterruptedException, TimeoutException, IOExc
 
     @Test
     public void testSchemaExport() throws InterruptedException, IOException, TimeoutException {
+        System.out.println("Loading schema... ");
         loadSimulationSchema("schematest");
```
flyingsilverfin, line 99: and we wanted to load the initial schema here to make sure there's no error in it right?

flyingsilverfin (changes requested):

server/migrate/Schema.java
maxbaxt: No they are not; good catch

test/assembly/MigrationTest.java
maxbaxt: Done.

flyingsilverfin (approved):

PR #5910, opened 2020-12-07
RuleImpl uses constructor to create new instance instead of of()

flyingsilverfin:
To keep in line with our style, `of()` should be a thin wrapper around some state passed in. Work being performed should be in a method called something else or in the constructor. Instead of giving all `ConceptImpl` classes a new `create()` method which does not mesh well with inheritance, we revert `RuleImpl` to use the dual-constructor strategy used elsewhere in `ConceptImpl`s.

* `RuleImpl` has two static `of()` methods, one of which calls a second constructor which actually creates a `logic` instance and performa validation.


flyingsilverfin:
closing as included in https://github.com/graknlabs/grakn/pull/5917

PR #5916, opened 2020-12-07
Make DB Cache instead of SchemaGraph borrowable

lolski:

We are refining the definition of the object lifetime of the db cache (`RocksDatabase.Cache`):

-  A cache is bound to a database
- A cache can be *borrowed* by one more more data transactions. The transactions are responsible to *unborrow* / return it. 
- A cache can be *invalidated* when a schema write transaction have been committed. If it is invalidated while there are no borrowers, it will be closed immediately. Otherwise, it will be closed when the last borrower has returned it.

- Refined `Cache` to follow the model described above
  - adds method `borrow`, `unborrow`, `invalidate`, and `close`
- Provide a thread-safe accessor method for the cache in `RocksDatabase`
  - the methods have guards that throws an exception if called on a database that is already is closed.
- Remove the read lock that is needed to perform schema commit and cache invalidation atomically


rocks/RocksDatabase.java
```diff
@@ -143,14 +144,15 @@ RocksSession createAndOpenSession(Arguments.Session.Type type, Options.Session o
         return session;
     }
 
-    synchronized Cache cache() {
+    synchronized Cache borrowCache() {
```
lolski, line 147: When an object borrows the cache, it calls `cache::borrow` to increment the number of borrower before returning the cache object

graph/SchemaGraph.java
```diff
@@ -418,26 +414,6 @@ public boolean isModified() {
         return isModified;
     }
 
-    // TODO: verify
```
lolski, line 421: We no longer do reference counting on the SchemaGraph. The db Cache is what gets reference-counted

rocks/RocksTransaction.java
```diff
@@ -239,23 +239,18 @@ void closeStorage() {
...
 
         public Data(RocksSession.Data session, Arguments.Transaction.Type type, Options.Transaction options) {
             super(session, type, options);
 
-            long lock = session.database.dataReadSchemaLock().readLock();
-            schemaGraph = session.database.cache().schemaGraph();
-            schemaGraph.incrementReference();
-            session.database.dataReadSchemaLock().unlockRead(lock);
-
+            cache = session.database.borrowCache();
```
lolski, line 248: we no longer to lock with the read lock since the access to the cache and the process of incrementing the reference has been wrapped into a synchronised method `borrowCache`

rocks/RocksDatabase.java
```diff
@@ -278,6 +286,26 @@ public ReasonerCache reasoner() {
         public SchemaGraph schemaGraph() {
             return schemaGraph;
         }
+
+        public void borrow() {
```
lriuui0x0, line 290: We need to put the cache close in `RocksDatabase.close()` into this new structure.

lriuui0x0 (changes requested):
Looks good, one comment.

lriuui0x0:
I also wonder with this new way, is the `dataReadSchemaLock` still needed?

lriuui0x0 (approved):

rocks/RocksDatabase.java
```diff
@@ -260,11 +269,17 @@ public void delete() {
         private final TraversalCache traversalCache;
         private final ReasonerCache reasonerCache;
         private final SchemaGraph schemaGraph;
+        private final RocksStorage schemaStorage;
+        private long borrowerCount;
+        private boolean invalidated;
 
         private Cache(RocksDatabase database) {
```
lolski, line 276: The `Cache` object expects the parent class `RocksDatabase` to call it in a thread safe way. Given that assumption, we no longer use any atomic variables

rocks/RocksDatabase.java
```diff
@@ -143,18 +143,27 @@ RocksSession createAndOpenSession(Arguments.Session.Type type, Options.Session o
         return session;
     }
 
-    synchronized Cache cache() {
+    synchronized Cache borrowCache() {
         if (cache == null) cache = new Cache(this);
```
lolski, line 147: we're using `synchronized` to prevent race conditions in this if block. Should we use an `AtomicReference` instead?

rocks/RocksDatabase.java
```diff
@@ -260,11 +269,17 @@ public void delete() {
         private final TraversalCache traversalCache;
         private final ReasonerCache reasonerCache;
         private final SchemaGraph schemaGraph;
+        private final RocksStorage schemaStorage;
+        private long borrowerCount;
```
lolski, line 273: do they need to be marked `volatile`?

rocks/RocksTransaction.java
lolski: According to @lriuui0x0, volatile is not needed in the context of explicit synchronization: https://stackoverflow.com/a/32974360/1622847

haikalpribadi (approved):

PR #5920, opened 2020-12-08
Variable Mapping Between Reasoner Actors

jmsfltchr:

When passing messages among the actors in reasoner it is necessary to map variables of the ConceptMaps being passed. Determining these mappings is the job of: Alpha Equivalence checking for de-duplication of actors; or unification for mapping from a concludable to a rule's then. This communication was being mocked. Here we un-mock the process, implement aggregation of partial answers, integrate Alpha Equivalence checking, and leave placeholders for the unification to be integrated soon.

- Remove all mocks except for a `TransactionMock` that mocks traversals.
- Add answer aggregators for unification (unimplemented), mapping via alpha equivalence, and no-op for the root actor. These are extremely important, they carry partial solutions around the reasoner, and aggregate these with newly explored partial solutions.
- There was a placeholder to pass unifiers around the reasoner framework, this is encapsulated by the aggregators and are no longer needed.
- The pattern that we now follow is that all requests sent contain a partial solution wrapped in an aggregator that uses the variables of the receiver. Answer responses contain an aggregated solution, and record the solution prior to variable mapping for use in explanations.
- Deduplicate Concludable Actors based on Alpha Equivalence.
- Upgrade resolution test to an integration test that uses real (un-mocked) conjunctions.
- A fix for where `$x = $a` in a rule would throw an exception when building concludables.


reasoner/resolution/ResolverRegistry.java
```diff
@@ -18,51 +18,78 @@
...
 
-    public Actor<ConcludableResolver> registerConcludable(Long pattern, List<List<Long>> rules, long traversalSize) {
-        LOG.debug("Register retrieval for concludable actor: '{}'", pattern);
-        return concludables.computeIfAbsent(pattern, (p) -> Actor.create(elg, self -> new ConcludableResolver(self, p, rules, traversalSize)));
+    public Pair<Actor<ConcludableResolver>, Map<Reference.Name, Reference.Name>> registerConcludable(ConjunctionConcludable<?, ?> concludable) {
+        LOG.debug("Register retrieval for concludable actor: '{}'", concludable.conjunction());
+
+        final Optional<Pair<ConjunctionConcludable<?, ?>, AlphaEquivalence>> alphaEquivalencePair = concludableActorsMap.keySet().stream()
+                .map(k -> new Pair<ConjunctionConcludable<?, ?>, AlphaEquivalence>(k, k.alphaEquals(concludable)))
+                .filter(p -> p.second().isValid()).findAny();
```
jmsfltchr, line 61: Here we're interested in registering deduplicated Concludables, so we want alpha equivalence. We iterate over all concludables already registered. If any is alpha equivalent then we add it and it's AlphaEquivalence to a pair, such that both can be accessed in the next step. Perhaps we can find a clearer flow for this, perhaps just a for loop with a `break` on finding a match?

reasoner/resolution/ResolverRegistry.java
```diff
@@ -18,51 +18,78 @@
...
+            // Then we can use the same ConcludableActor, but with a different variable mapping
+            concludableActor = concludableActorsMap.get(alphaEquivalencePair.get().first());
+            variableMapping = alphaEquivalencePair.get().second().asValid().namedVariableMapping();
+        } else {
+            // Create a new ConcludableActor
+            concludableActor = Actor.create(elg, self -> new ConcludableResolver(self, concludable));
+            concludableActorsMap.put(concludable, concludableActor);
+            variableMapping = MappingAggregator.identityMapping(concludable);
+        }
+        return new Pair<>(concludableActor, variableMapping);
```
jmsfltchr, line 75: If we found a match, return that ConcludableActor and its VariableMapping.

Otherwise, create a new ConcludableActor, with a trivial (identity) variableMapping.

reasoner/resolution/ResolverRegistry.java
jmsfltchr: I've refactored this to use a for loop now - much cleaner

pattern/equivalence/AlphaEquivalence.java
```diff
@@ -113,9 +118,24 @@ public Valid asValid() {
             return this;
         }
 
-        HashMap<Variable, Variable> map() {
+        public HashMap<Variable, Variable> variableMapping() {
             return new HashMap<>(map);
```
flyingsilverfin, line 122: don't we have a helper `map()` function for this? hmmm

pattern/variable/ThingVariable.java
```diff
@@ -199,7 +199,8 @@ public String toString() {
 
         if (reference().isName()) syntax.append(reference()).append(SPACE);
 
-        syntax.append(Stream.of(relationConstraints, set(isaConstraint), hasConstraints, valueConstraints, isConstraints)
+        Set<IsaConstraint> isaConstraintSet = isaConstraint == null ? set() : set(isaConstraint);
+        syntax.append(Stream.of(relationConstraints, isaConstraintSet, hasConstraints, valueConstraints, isConstraints)
```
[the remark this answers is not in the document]
flyingsilverfin: not immediately

reasoner/ReasonerProducer.java
```diff
@@ -17,22 +17,52 @@
...
+        }
     }
 
     @Override
-    public void recycle() {} // no-op
-}
+    public void recycle() {
+
+    }
+}
```
flyingsilverfin, line 68: missing newline

reasoner/resolution/MockTransaction.java
```diff
@@ -47,33 +47,15 @@ public MockTransaction(long computeLength) {
...
-    public Iterator<ConceptMap> query(List<Long> traversalPattern, ConceptMap partialConceptMap) {
-        Map<String, Long> variableSeeds = new HashMap<>();
-        for (Long pattern : traversalPattern) {
-            String name = "var_" + pattern.toString();
-            Long seed = new Random(pattern.hashCode() + traversalPattern.hashCode()).nextLong();
-            variableSeeds.put(name, seed);
+            if (variable.reference().isName()) {
+                String name = variable.identifier().reference().asName().toString().substring(1);
+                Long seed = new Random(variable.hashCode()).nextLong();
+                variableSeeds.put(name, seed);
```
flyingsilverfin, line 53: lol good to see this hack still works haha

reasoner/resolution/ResolutionRecorder.java
```diff
@@ -46,32 +46,32 @@ protected void exception(final Exception e) {
...
+            ResolutionAnswer subAnswer = subAnswers.get(key);
+            ResolutionAnswer mergedSubAnswer = merge(subAnswer);
             mergedSubAnswers.put(key, mergedSubAnswer);
         }
         newDerivation.replace(mergedSubAnswers);
 
         int actorIndex = actorIndices.computeIfAbsent(newAnswer.producer(), key -> actorIndices.size());
         LOG.debug("actor index for " + newAnswer.producer() + ": " + actorIndex);
-        AnswerIndex newAnswerIndex = new AnswerIndex(actorIndex, newAnswer.conceptMap());
+        AnswerIndex newAnswerIndex = new AnswerIndex(actorIndex, newAnswer.aggregated().original());
```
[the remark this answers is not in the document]
flyingsilverfin: i'm pretty sure this is right - we want to record answers in the "space" of the actor that is recording this derivation as its own, ie it should be in the sender's variable space

reasoner/resolution/ResolverRegistry.java
```diff
@@ -18,51 +18,69 @@
...
         rules = new HashMap<>();
         resolutionRecorder = Actor.create(elg, ResolutionRecorder::new);
     }
 
-    public Actor<ConcludableResolver> registerConcludable(Long pattern, List<List<Long>> rules, long traversalSize) {
-        LOG.debug("Register retrieval for concludable actor: '{}'", pattern);
-        return concludables.computeIfAbsent(pattern, (p) -> Actor.create(elg, self -> new ConcludableResolver(self, p, rules, traversalSize)));
+    public Pair<Actor<ConcludableResolver>, Map<Reference.Name, Reference.Name>> registerConcludable(ConjunctionConcludable<?, ?> concludable) {
+        LOG.debug("Register retrieval for concludable actor: '{}'", concludable.conjunction());
+        for (Map.Entry<ConjunctionConcludable<?, ?>, Actor<ConcludableResolver>> c: concludableActorsMap.entrySet()) {
```
flyingsilverfin, line 58: can you put a TODO here that this should be optimised from a linear search to use a alpha hash? for kasper

reasoner/resolution/ResolverRegistry.java
```diff
@@ -18,51 +18,69 @@
...
+            AlphaEquivalence alphaEquality = c.getKey().alphaEquals(concludable);
+            if (alphaEquality.isValid()) {
+                return new Pair<>(c.getValue(), alphaEquality.asValid().namedVariableMapping());
+            }
+        }
+        Actor<ConcludableResolver> concludableActor = Actor.create(elg, self -> new ConcludableResolver(self, concludable));
+        concludableActorsMap.put(concludable, concludableActor);
+        return new Pair<>(concludableActor, MappingAggregator.identity(concludable));
```
flyingsilverfin, line 66: line reads odd, why are using a mapping aggregator to create a mapping hmmm maybe we can find a better naming after all

jmsfltchr: Actually this should be ok, we just need to find a better place for the `identity` method, which returns an identity mapping. Already deliberated on where to put this, couldn't find somewhere better.

jmsfltchr: If we still need to do a deep check after comparing with the hash, then the fastest we can do is to find the set of equal concludables and then run the deep check across those?

flyingsilverfin: yes, exactly, but better than looping. Was this response intended for another comment?

reasoner/resolution/resolver/ConcludableResolver.java
```diff
@@ -45,86 +48,92 @@
...
 
-    private Either<Request, Response> produceMessage(Request fromUpstream, ResponseProducer responseProducer) {
+    private Either<Request, Response> messageToSend(Request fromUpstream, ResponseProducer responseProducer) {
         while (responseProducer.hasTraversalProducer()) {
             ConceptMap conceptMap = responseProducer.traversalProducer().next();
+            Aggregator.Aggregated aggregated = fromUpstream.partialConceptMap().aggregateWith(conceptMap);
+            if (aggregated == null) {
+                // TODO this should be the only place that aggregation can fail, but can we make this explicit?
+                continue;
+            }
```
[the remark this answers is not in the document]
flyingsilverfin: agree, im not sure its the only place!

rocks/RocksTransaction.java
```diff
@@ -115,6 +115,10 @@ public void close() {
         }
     }
 
+    public Reasoner reasoner() {
```
flyingsilverfin, line 118: yup doesn't need to be public, as its in the same package as the test

test/integration/reasoner/ResolutionTest.java
```diff
@@ -0,0 +1,451 @@
...
+        createRootAndAssertResponses(conjunctionPattern, answerCount);
+    }
+
+    @Ignore // TODO Un-ignore, ignored until rules are ready to use
+    @Test
+    public void simpleRule() throws InterruptedException {
+        // TODO This is what the initial test described, but since atomic2 is the conjunction the conjuntion traversal must return the same number of results as the atomic2 traversal.
+        String atomic1 = "$p1 isa person, has name \"Bob\";";
+        String atomic2 = "$p1 isa person, has age 42;";
+        String rulePattern = "rule bobs-are-42: when { $p1 isa person, has name \"Bob\"; } then { $p1 isa person, has age 42; };";
```
flyingsilverfin, line 154: errr this isn't a valid rule pattern?? is having `isa` on `p1` valid in graql??

test/integration/reasoner/ResolutionTest.java
```diff
@@ -0,0 +1,451 @@
...
+            }
+        }
+        long atomic1TraversalAnswerCount = 3L;
+        long atomic2TraversalAnswerCount = 3L;
+        long atomic3TraversalAnswerCount = 3L;
+        long atomic4TraversalAnswerCount = 3L;
+        long atomic5TraversalAnswerCount = 3L;
+        long conjunctionTraversalAnswerCount = 0L;
+        long answerCount = conjunctionTraversalAnswerCount + (atomic5TraversalAnswerCount * atomic4TraversalAnswerCount * atomic3TraversalAnswerCount * atomic2TraversalAnswerCount * atomic1TraversalAnswerCount);
+        Conjunction conjunctionPattern = parseConjunction("{ " + atomic5 + " " + atomic4 + " " + atomic3 + " " + atomic2 + " " + atomic1 + " }");
```
flyingsilverfin, line 311: are you intentionally keeping multiple `isa` on `$p1` after merging? I think it's fine as they get deduplicated, just want to check

test/integration/reasoner/ResolutionTest.java
```diff
@@ -0,0 +1,451 @@
...
+        long atomic1TraversalAnswerCount = 3L;
+        long ruleTraversalAnswerCount = 0L;
+        long atomic2TraversalAnswerCount = 0L;
+        long conjunctionTraversalAnswerCount = 3L;
+        long answerCount = conjunctionTraversalAnswerCount + atomic2TraversalAnswerCount + ruleTraversalAnswerCount + atomic1TraversalAnswerCount;
+        Conjunction conjunctionPattern = parseConjunction("{ " + atomic2 + " }");
+
+        LinkedBlockingQueue<ResolutionAnswer> responses = new LinkedBlockingQueue<>();
+        AtomicLong doneReceived = new AtomicLong(0L);
+        EventLoopGroup elg = new EventLoopGroup(1, "reasoning-elg");
```
flyingsilverfin, line 381: i think even though this test is possibly going to have to change a fair amount, we should update it to be similar style to the others - eg. get `ResolverRegistry` via the `Reasoner`, etc.

flyingsilverfin (changes requested):
I've done high level review of all the stuff NOT including `Aggregator`, since we should discuss these in person

reasoner/resolution/answer/Aggregator.java
```diff
@@ -0,0 +1,86 @@
...
+
+    Aggregator(ConceptMap original, ConceptMap transformed) {
+        this.original = original;
+        this.transformed = transformed;
+    }
+
+    public Aggregated aggregateWith(ConceptMap toAggregate) {
+        ConceptMap unTransformed = unTransform(toAggregate);
+        if (unTransformed == null) return null;
+        if (unTransformed.concepts().isEmpty()) ; // TODO What do we do in this case?
```
flyingsilverfin, line 41: should be illegal state

reasoner/resolution/answer/Aggregator.java
```diff
@@ -0,0 +1,86 @@
...
+    private final ConceptMap original;
+    private final ConceptMap transformed;
+
+    Aggregator(ConceptMap original, ConceptMap transformed) {
+        this.original = original;
+        this.transformed = transformed;
+    }
+
+    public Aggregated aggregateWith(ConceptMap toAggregate) {
+        ConceptMap unTransformed = unTransform(toAggregate);
```
flyingsilverfin, line 39: assert that there are some variables in common

reasoner/resolution/answer/Aggregator.java
```diff
@@ -0,0 +1,86 @@
...
+        if (unTransformed == null) return null;
```
flyingsilverfin, line 40: this is the filter operation that unifier can provide, so we should either implement a `canAggregate()` to check before calling this method, or call this `tryAggregateWith()` and return an Optional

reasoner/resolution/answer/Aggregator.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+import graql.lang.pattern.variable.Reference;
+
+public abstract class Aggregator {
+    private final ConceptMap original;
+    private final ConceptMap transformed;
+
+    Aggregator(ConceptMap original, ConceptMap transformed) {
```
flyingsilverfin, line 33: TODO to try to refactor this to clean up the ownership of operations and variable domains

reasoner/resolution/resolver/RuleResolver.java
```diff
@@ -21,74 +21,43 @@
...
+        super(self, RuleResolver.class.getSimpleName() + "(rule:" + rule + ")", rule.when(), rule.whenConcludables());
     }
 
     @Override
     public Either<Request, Response> receiveRequest(Request fromUpstream, ResponseProducer responseProducer) {
-        return produceMessage(fromUpstream, responseProducer);
-    }
-
-    @Override
-    public Either<Request, Response> receiveAnswer(Request fromUpstream, Response.Answer fromDownstream, ResponseProducer responseProducer) {
```
flyingsilverfin, line 48: let's put this back since max will have to implement `materialise()` in here, which is different between the conjunction actors

reasoner/resolution/resolver/ConjunctionResolver.java
```diff
@@ -68,22 +108,43 @@ protected ResponseProducer createResponseProducer(Request request) {
...
     boolean isLast(Actor<? extends Resolver<?>> actor) {
-        return plannedConcludables.get(plannedConcludables.size() - 1).equals(actor);
+        return plannedConcludables.get(plannedConcludables.size() - 1).first().equals(actor);
     }
 
-    Actor<ConcludableResolver> nextPlannedDownstream(Actor<? extends Resolver<?>> actor) {
-        return plannedConcludables.get(plannedConcludables.indexOf(actor) + 1);
+    Pair<Actor<ConcludableResolver>, Map<Reference.Name, Reference.Name>> nextPlannedDownstream(Actor<? extends Resolver<?>> actor) {
+        boolean match = false;
+        for (Pair<Actor<ConcludableResolver>, Map<Reference.Name, Reference.Name>> planned : plannedConcludables) {
```
flyingsilverfin, line 132: use index

flyingsilverfin (approved):

reasoner/resolution/answer/Aggregator.java
jmsfltchr: Will address this in the rafactor

reasoner/resolution/answer/Aggregator.java
jmsfltchr: I think that wouldn't be correct as there won't be any connected variables if there's a disconnection in the query.

test/integration/reasoner/ResolutionTest.java
jmsfltchr: I missed that one!

test/integration/reasoner/ResolutionTest.java
jmsfltchr: Good spot

reasoner/resolution/ResolverRegistry.java
```diff
@@ -18,51 +18,70 @@
...
         LOG.debug("Creating Conjunction Actor for pattern: '{}'", pattern);
-        return Actor.create(elg, self -> new RootResolver(self, pattern, traversalSize, onAnswer, onExhausted));
+        return Actor.create(elg, self -> new RootResolver(self, pattern, onAnswer, onExhausted));
     }
 
     public Actor<ResolutionRecorder> resolutionRecorder() {
         return resolutionRecorder;
     }
+
+    public void setEventLoopGroup(EventLoopGroup eventLoopGroup) {
```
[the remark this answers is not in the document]
jmsfltchr: Have left `setEventLoopGroup` public

reasoner/resolution/answer/UnifyingAggregator.java
```diff
@@ -0,0 +1,49 @@
...
+
+package grakn.core.reasoner.resolution.answer;
+
+import grakn.core.concept.answer.ConceptMap;
+import graql.lang.pattern.variable.Reference;
+
+import java.util.Map;
+import java.util.Set;
+
+public class UnifyingAggregator extends Aggregator {
```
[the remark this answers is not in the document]
jmsfltchr: Somewhere we need to do the actual logic or transforming ConceptMaps with the unifier, and un-transforming it. Currently expecting that to sit here. Should be clearer in the refactor of aggregators I hope.

reasoner/resolution/ResolverRegistry.java
jmsfltchr: Yes, above

reasoner/resolution/ResolverRegistry.java
jmsfltchr: If we still need to do a deep check after comparing with the hash, then the fastest we can do is to find the set of equal concludables and then run the deep check across those?

pattern/equivalence/AlphaEquivalence.java
jmsfltchr: I don't understand - this is that helper method, but called `variableMapping()` now to be explicit as to which map you'll get back.

rocks/RocksTransaction.java
```diff
@@ -120,6 +120,10 @@ public void close() {
...
```
jmsfltchr, line 123: We need to be able to access the reasoner in order to make assertions in tests about it's behaviour.

test/integration/reasoner/ResolutionTest.java
```diff
@@ -0,0 +1,456 @@
+/*
```
jmsfltchr, line 1: Move the resolution test to `test/integration`.

test/integration/reasoner/BUILD
```diff
@@ -0,0 +1,52 @@
+#
```
jmsfltchr, line 1: Add the necessary BUILD file for the resolution test

reasoner/BUILD
```diff
@@ -62,26 +63,6 @@ native_java_libraries(
     tags = ["maven_coordinates=io.grakn.core:grakn-reasoner:{pom_version}"],
 )
 
-host_compatible_java_test(
```
jmsfltchr, line 65: Remove the test target from the old test location

pattern/variable/ThingVariable.java
```diff
@@ -199,7 +205,8 @@ public String toString() {
 
         if (reference().isName()) syntax.append(reference()).append(SPACE);
 
-        syntax.append(Stream.of(relationConstraints, set(isaConstraint), hasConstraints, valueConstraints, isConstraints)
+        Set<IsaConstraint> isaConstraintSet = isaConstraint == null ? set() : set(isaConstraint);
```
jmsfltchr, line 208: Fix the `toString` method for the case where the `isa` constraint is absent.

pattern/variable/ThingVariable.java
```diff
@@ -157,6 +157,12 @@ public IsConstraint is(ThingVariable variable) {
         return valueDateTimeConstraint;
     }
 
+    public ValueConstraint.Variable valueVariable(GraqlToken.Predicate.Equality comparator, ThingVariable variable) {
```
jmsfltchr, line 160: Add an omitted constraint method - adding a variable value constraint.

pattern/equivalence/AlphaEquivalenceTest.java
```diff
@@ -59,8 +59,8 @@ private Variable parseVariables(String variableName, String... graqlVariables) {
                 .collect(Collectors.toSet());
     }
 
-    private Map<String, String> alphaMapToStringMap(AlphaEquivalence.Valid alphaMap) {
-        return alphaMap.map().entrySet().stream()
+    private Map<String, String> alphaMapToStringMap(AlphaEquivalence.Valid alphaEq) {
```
jmsfltchr, line 62: A single variable rename

pattern/equivalence/AlphaEquivalence.java
```diff
@@ -113,9 +118,24 @@ public Valid asValid() {
             return this;
         }
 
-        HashMap<Variable, Variable> map() {
+        public HashMap<Variable, Variable> variableMapping() {
```
jmsfltchr, line 121: Improved API for accessing the AlphaEquivalence result

rocks/RocksTransaction.java
flyingsilverfin: did this turn out to have to be public?

lolski (approved):

reasoner/resolution/ResolutionRecorder.java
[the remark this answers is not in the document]
flyingsilverfin: @jmsfltchr  i just revisited this as I'm working my way through unification... i had misunderstood and we reached the wrong conclusion. We _definitely_ want the concept map to the be in the "variable space" of the sender, but it should also include the answer found by the child! Will fix this in my next PR

PR #5922, opened 2020-12-08
Type hinter cleanup

flyingsilverfin:
Type hinter was creating concept directly from vertices via the `TypeImpl` class. Instead, it should use the `ConceptManager` class which is added.

* Clean up formatting, such as spacing
* Inject `ConceptManager` to create concepts instead of using `Vertex` directly
* make inner classes static

### DIFF
--- a/logic/tool/TypeHinter.java
+++ b/logic/tool/TypeHinter.java
@@ -21,13 +21,14 @@ package grakn.core.logic.tool;
 import grakn.core.common.concurrent.ExecutorService;
 import grakn.core.common.exception.GraknException;
 import grakn.core.common.parameters.Label;
+import grakn.core.concept.ConceptManager;
 import grakn.core.concept.type.Type;
-import grakn.core.concept.type.impl.TypeImpl;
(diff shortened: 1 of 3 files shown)


maxbaxt:
Looks good and you haven't had to change much of the existing code, which is elegant.

However, you haven't changed any BUILD file, so it is still building the graph package into TypeHinter

logic/tool/TypeHinter.java
```diff
@@ -417,7 +422,7 @@ private void convertIsa(TypeVariable owner, IsaConstraint isaConstraint) {
             else if (isaConstraint.type().reference().isName()) owner.is(isaConstraint.type());
             else if (isaConstraint.type().label().isPresent())
                 owner.label(isaConstraint.type().label().get().properLabel());
-            else GraknException.of(ILLEGAL_STATE);
+            else throw GraknException.of(ILLEGAL_STATE);
```
maxbaxt, line 425: good catch

lolski:
Merged as it has been verbally approved by @haikalpribadi , according to @flyingsilverfin

PR #5923, opened 2020-12-08
Long lived rule cache and rule validation hooks

flyingsilverfin:
Avoid re-loading rules that are in the rule cache by introducing long-lived `LogicCache`. The `LogicCache` contains the prior`TypeHinterCache` and the new `RuleCache`. We also introduce a scan for retrieving all rule structures that are stored in the schema graph, and leave hooks for rule validation points. Rule validation now occurs on both rule `put` and on schema transaction `commit`

* Introduce `LogicCache`, containing `RuleCache` and `TypeHinterCache`
* Introduce a scan for all rules in `SchemaGraph`. We will improve this by indexing rules by types they mention in the future, so we can look up rules that are relevant to a starting rule and the types it can infer
* Introduce points for rule validation: at `put` time, and re-validation at schema transaction commit time (in case the schema has chanced since writing the rule the first time)
* rename `Grakn.Transaction.logics()` to `Grakn.Transaction.logic()`
* Simplify `Rule` by removing `ConceptManager`


graph/SchemaGraph.java
```diff
@@ -166,6 +164,13 @@ public TypeVertex rootRoleType() {
         return getType(ROLE.label(), ROLE.scope());
     }
 
+    public ResourceIterator<RuleStructure> rules() {
```
flyingsilverfin, line 167: scan index for all rules, then create `RuleStructure` objects from them

logic/Rule.java
```diff
@@ -165,57 +146,50 @@ public final int hashCode() {
...
-    private void validateLabelsExist() {
-        Stream<Label> whenPositiveLabels = getTypeLabels(when.variables().stream());
-        Stream<Label> whenNegativeLabels = getTypeLabels(when.negations().stream().flatMap(this::negationVariables));
-        Stream<Label> thenLabels = getTypeLabels(then.stream().flatMap(constraint -> constraint.variables().stream()));
-        Set<String> invalidLabels = invalidLabels(Stream.of(whenPositiveLabels, whenNegativeLabels, thenLabels).flatMap(Function.identity()));
-        if (!invalidLabels.isEmpty()) {
-            throw GraknException.of(TYPES_NOT_FOUND, getLabel(), String.join(", ", invalidLabels));
-        }
+    void validateCycles() {
+        // TODO detect negated cycles in the rule graph
```
flyingsilverfin, line 155: kasper's slot in

logic/LogicManager.java
```diff
@@ -18,43 +18,133 @@
...
-        // TODO detect negated cycles in the rule graph after inserting this rule and updating a rule graph
-
-        return rule;
+        RuleStructure structure = graphMgr.schema().getRule(label);
+        if (structure != null) {
+            // overwriting a rule means we purge it and re-create the rule
+            structure.delete();
+            logicCache.rule().invalidate(label);
+        }
+        return logicCache.rule().get(label, l -> Rule.of(graphMgr, conceptMgr, this, label, when, then));
```
jmsfltchr, line 68: We have ensured the rule has been deleted if it exists; So this last line is creating a rule, putting it in the cache, and storing it to the graph, and then returning it from the cache?

logic/LogicManager.java
```diff
@@ -18,43 +18,133 @@
...
+            return logicCache.rule().get(structure.label(), l -> Rule.of(this, structure));
+        }
         return null;
     }
 
-    public TypeHinter typeHinter() {
+    public ResourceIterator<Rule> rules() {
+        return graphMgr.schema().rules().map(struct -> {
+            Rule rule = logicCache.rule().getIfPresent(struct.label());
+            if (rule == null) rule = logicCache.rule().get(struct.label(), l -> Rule.of(this, struct));
```
jmsfltchr, line 84: We check nulls everywhere in the storage layer rather than use `Optional<>`s?

jmsfltchr (approved):

logic/LogicManager.java
flyingsilverfin: yup - and atomically!

logic/LogicManager.java
flyingsilverfin: yep

flyingsilverfin: because rocks can return null values for a key

haikalpribadi: how can RocksDB return null values for a key, actually, @flyingsilverfin ?

haikalpribadi (approved):

PR #5925, opened 2020-12-08
Re-enable BDD tests for match query

flyingsilverfin:
Uncomment BDD Graql steps and restore them to a buildable state. Running `bazel test //test/behaviour/graql/language/...` should now run, with only the standard assertion failures from the tests.

* Add `Numeric` answer type
* Add `isType()` and `isThing()` on `Concept`
* Add match APIs for aggregate and group queries
* make some tests to size 'medium' that were taking longer than 'small'
* uncomment BDD Graql steps, and make them buildable

### DIFF
--- a/concept/answer/Numeric.java
+++ b/concept/answer/Numeric.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (C) 2020 Grakn Labs
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package grakn.core.concept.answer;
+
+/**
+ * A type of Answer object that contains a Number. Will either be a long or a double
+ */
+public class Numeric implements Answer {
+
+    private final Number number;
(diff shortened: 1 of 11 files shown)


concept/answer/Numeric.java
```diff
@@ -0,0 +1,48 @@
...
+    public Number number() {
+        return number;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this) return true;
+        if (obj == null || getClass() != obj.getClass()) return false;
+        Numeric a2 = (Numeric) obj;
+        return this.number.toString().equals(a2.number.toString());
```
haikalpribadi, line 41: Are we sure of this, @flyingsilverfin ? Is this what we had in the old code? I'm not quite sure this is right.

test/behaviour/graql/language/get/BUILD
```diff
@@ -38,7 +38,7 @@ java_test(
         "@graknlabs_behaviour//graql/language:get.feature",
     ],
     classpath_resources = ["//test/conf:logback"],
-    size = "small",
+    size = "medium",
```
haikalpribadi, line 41: Why are we increasing the size of all the tests?

haikalpribadi (changes requested):
I'm not quite sure why the `define` BDD tests failed, as you didn't change anything that could make it fail. But let's look into it, and address a few of the minor comments here.

test/behaviour/graql/language/get/BUILD
flyingsilverfin: they all take longer now -- im pretty sure its cuz we open twice the number of databases as before!

concept/answer/Numeric.java
flyingsilverfin: updated to make sure the long and double values are the same.

flyingsilverfin:
The `define` test is now failing, because each `define` test did both a `define` and a `match`. Before this PR, we just ignored any steps to do with match, which is why the `define` test passed. Now we expect them to fail until the traversal engine is fully complete :)

flyingsilverfin:
Merged as part of a separate commit

PR #5931, opened 2020-12-09
Type Hinter will no longer give abstract hints for Thing Variables

maxbaxt:

Type Hinter should no longer suggest an abstract type for any Thing Variable. This is because a Thing Variable cannot be a an instance of an abstract type. However, it is still true that a Type Variable can be. 

For example, if we `person sub entity, abstract;`, `man sub person;`, `woman sub person;`. Then `$x isa person;` should have Type Hints `{man, woman}`, whilst `$x sub person;` should have Type Hints `{man, woman, person}`.

A new method in TypeHinter to check for abstract labels and used exclusively when dealing with Thing Variables

### DIFF
--- a/logic/tool/TypeHinter.java
+++ b/logic/tool/TypeHinter.java
@@ -222,6 +222,7 @@ public class TypeHinter {
     }
 
     private void removeAbstractHints(Set<Label> hints) {
+        //TODO: use .getType(label) once ConceptManager can handle labels
         hints.removeIf(label -> {
             Type type = conceptMgr.getType(label.scopedName());
             assert type != null;
--- a/logic/tool/TypeHinter.java
+++ b/logic/tool/TypeHinter.java
@@ -214,20 +214,12 @@ public class TypeHinter {
     }
 
     private void addInferredIsaLabels(ThingVariable variable, Set<Label> hints, Map<Label, TypeVariable> labelMap) {
-        removeAbstractHints(hints);
+        //TODO: use .getType(label) once ConceptManager can handle labels
+        hints.removeIf(label -> conceptMgr.getType(label.scopedName()).isAbstract());
         if (!variable.isa().isPresent()) {
             variable.isa(lowestCommonSuperType(hints, labelMap), false);
         }
         variable.isa().get().addHints(hints);
-    }
-
-    private void removeAbstractHints(Set<Label> hints) {
-        //TODO: use .getType(label) once ConceptManager can handle labels
-        hints.removeIf(label -> {
-            Type type = conceptMgr.getType(label.scopedName());
-            assert type != null;
-            return type.isAbstract();
-        });
     }
 
     private TypeVariable lowestCommonSuperType(Set<Label> labels, Map<Label, TypeVariable> labelMap) {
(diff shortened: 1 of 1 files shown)


logic/tool/TypeHinter.java
```diff
@@ -141,12 +140,6 @@ private TypeVariable above(Variable variable) {
         } else throw GraknException.of(ILLEGAL_STATE);
     }
 
-    private void clearHintLabels(Variable variable) {
-        if (variable.isType()) variable.asType().sub().ifPresent(SubConstraint::clearHintLabels);
-        else if (variable.isThing()) variable.asThing().isa().ifPresent(IsaConstraint::clearHintLabels);
-        else throw GraknException.of(ILLEGAL_STATE);
-    }
-
```
maxbaxt, line 149: This wasn't being used, so I deleted.

logic/tool/TypeHinter.java
```diff
@@ -114,7 +113,7 @@ public Conjunction computeHints(Conjunction conjunction, int parallelisation) {
             Set<Label> hintLabels = localTypeHints.get(variable.reference());
             if (variable.isThing()) {
                 if (hintLabels.size() != numOfThings) {
-                    addInferredIsaLabels(variable.asThing(), localTypeHints.get(variable.reference()), labelMap);
+                    addInferredIsaLabels(variable.asThing(), hintLabels, labelMap);
```
maxbaxt, line 116: Not part of the original scope of the PR, but this pattern is a lot better.

logic/tool/TypeHinter.java
```diff
@@ -221,12 +214,22 @@ private void addInferredSubLabels(TypeVariable variable, Set<Label> hints, Map<L
     }
 
     private void addInferredIsaLabels(ThingVariable variable, Set<Label> hints, Map<Label, TypeVariable> labelMap) {
+        removeAbstractHints(hints);
```
maxbaxt, line 217: It might have been nice to have done this in `retrieveVariableHints`, since that is whence the type hints come. But that method doesn't have the context of whether a variable is Thing or Type, so it has to be filtered later. Putting it here seems the neatest place, since this is exactly then a Thing Variable is assigned its Hint Labels.

flyingsilverfin: i think it'll be fine if you inline this function and just remove the assertion, since we should never have invalid labels in the first place...
`hints.removeIf(label -> conceptMgr.getType(label.scopedName()).isAbstract())`

logic/tool/TypeHinter.java
```diff
@@ -221,12 +214,22 @@ private void addInferredSubLabels(TypeVariable variable, Set<Label> hints, Map<L
...
         if (!variable.isa().isPresent()) {
             variable.isa(lowestCommonSuperType(hints, labelMap), false);
         }
         variable.isa().get().addHints(hints);
```
flyingsilverfin, line 221: is there an `else` missing here??

maxbaxt: No. This code checks if an `isa` edge exists, and if not adds it. If it already exists, we just carry on.

logic/tool/TypeHinter.java
maxbaxt: Alright, made this one line.

logic/tool/TypeHinter.java
flyingsilverfin: cool!

flyingsilverfin (approved):

PR #5933, opened 2020-12-09
Unification from Conjunction Concludables to Head Concludables

maxbaxt:

Given a Conjunction Concludable, we can return all possible unifiers from the Conjunction Concludable to Head Concludable.

In other words, we can find every valid mapping from the named variables of a concludable from a Conjunction or a Query, to a corresponding Concludable that might appear as a Concludable to the Then of a Rule. 

Implemented a `unify` method to Conjunction Concludable that takes in a Head Conjunction and returns a Stream of Unifications.


logic/Rule.java
```diff
@@ -169,7 +169,7 @@ public final int hashCode() {
     private Set<HeadConcludable<?, ?>> createHead(Set<Constraint> thenConstraints, Set<Variable> constraintContext) {
         HashSet<HeadConcludable<?, ?>> thenConcludables = new HashSet<>();
         thenConstraints.stream().filter(Constraint::isThing).map(Constraint::asThing)
-                .flatMap(constraint -> HeadConcludable.of(constraint, constraintContext).stream()).forEach(thenConcludables::add);
+                .map(constraint -> HeadConcludable.of(constraint, constraintContext)).forEach(thenConcludables::add);
```
maxbaxt, line 172: We've gotten rid of generalisations, so the output of `.of` is a single HeadConcludable rather than a Set or Stream.

logic/concludable/Concludable.java
```diff
@@ -74,18 +80,18 @@ static IsaConstraint copyConstraint(IsaConstraint isa) {
     }
 
     static ValueConstraint<?> copyConstraint(ValueConstraint<?> value) {
+        //NOTE: isa can never exist on a Value Concludable (or else it would be a Isa Concludable).
```
maxbaxt, line 80: Added this note because it wasn't obvious imho why we weren't copying the isa of a value.

logic/concludable/Concludable.java
```diff
@@ -105,7 +111,10 @@ else if (toCopy.isString())
             value = toConstrain.valueString(toCopy.asString().predicate(), toCopy.asString().value());
         else if (toCopy.isDateTime())
             value = toConstrain.valueDateTime(toCopy.asDateTime().predicate().asEquality(), toCopy.asDateTime().value());
-        else throw GraknException.of(ILLEGAL_STATE);
+        else if (toCopy.isVariable()) {
+            ThingVariable copyOfVar = copyIsaAndValues(toCopy.asVariable().value());
+            value = toConstrain.valueVariable(toCopy.asValue().predicate().asEquality(), copyOfVar);
+        } else throw GraknException.of(ILLEGAL_STATE);
```
maxbaxt, line 114: The case for ValueConstraint being a Variable was not covered.

logic/concludable/Concludable.java
```diff
@@ -122,6 +131,11 @@ static void copyIsaAndValues(ThingVariable oldOwner, ThingVariable newOwner) {
 
     static void copyLabelAndValueType(TypeVariable copyFrom, TypeVariable copyTo) {
         if (copyFrom.label().isPresent()) copyTo.label(Label.of(copyFrom.label().get().label()));
+        if (copyFrom.sub().isPresent()) {
+            SubConstraint subCopy = copyFrom.sub().get();
+            copyTo.sub(subCopy.type(), subCopy.isExplicit());
+            copyTo.sub().get().addHints(subCopy.typeHints());
+        }
```
maxbaxt, line 135: Need to copy the sub constraints when copying a type.

logic/concludable/ConjunctionConcludable.java
```diff
@@ -50,36 +52,37 @@ private ConjunctionConcludable(CONSTRAINT constraint) {
         return new Extractor(conjunction.variables()).concludables();
     }
 
-    public Stream<Pair<Rule, Unification>> findUnifiableRules(Stream<Rule> allRules) {
+    public Stream<Pair<Rule, Map<Reference, Set<Reference>>>> findUnifiableRules(Stream<Rule> allRules) {
```
maxbaxt, line 56: The unification object will soon just take a mapping at its construction. So for now we'll just return the mappings.

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -181,6 +181,11 @@ public Long(ThingVariable owner, GraqlToken.Predicate.Equality predicate, long v
             super(owner, predicate, value, set());
         }
 
+        @Override
+        public GraqlToken.Predicate.Equality predicate() {
+            return super.predicate().asEquality();
+        }
+
```
maxbaxt, line 188: It's logical that a predicated from Double, Long, Boolean or DateTime Constraint should always return an equality predicate.

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +119,89 @@ public Value asValue() {
...
+            if (conjunctionRolePlayers.isEmpty()) {
+                return set(unifier);
+            }
+            RelationConstraint.RolePlayer conj = conjunctionRolePlayers.remove(0);
+            for (RelationConstraint.RolePlayer head : headRolePlayers) {
+                if (visitedHeadRolePlayers.contains(head)) continue;
+                visitedHeadRolePlayers.add(head);
+                updatedUnifier(conj, head, unifier).ifPresent(newUnifier ->
+                        allUnifiers.addAll(matchRolesAndUnify(conjunctionRolePlayers,
+                                headRolePlayers, visitedHeadRolePlayers, newUnifier)));
```
maxbaxt, line 183: We recurse with the matched roles removed from the set of head and conjunction role-players, and keep going until we run out of conjunction players to match.

logic/concludable/ConjunctionConcludable.java
```diff
@@ -19,22 +19,29 @@
...
 import grakn.common.collection.Pair;
 import grakn.core.common.exception.GraknException;
-import grakn.core.common.parameters.Label;
 import grakn.core.logic.Rule;
-import grakn.core.logic.Unification;
 import grakn.core.pattern.constraint.Constraint;
 import grakn.core.pattern.constraint.thing.HasConstraint;
 import grakn.core.pattern.constraint.thing.IsaConstraint;
 import grakn.core.pattern.constraint.thing.RelationConstraint;
+import grakn.core.pattern.constraint.thing.RelationConstraint.RolePlayer;
```
maxbaxt, line 27: This allows me to write RolePlayer instead of RelationConstraint.RolePlayer, which gets quite verbose. However, having `import X` and `import X.Y` together is perhaps less than desirable?

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +119,86 @@ public Value asValue() {
...
+                        unifyWith.constraint().owner().isa().get().type(), startingUnifier);
+                if (newUnifier.isPresent()) startingUnifier = newUnifier.get();
+                else return Stream.empty();
+            }
+
+            return matchRolesAndUnify(new ArrayList<>(this.constraint().players()), unifyWith.constraint().players(),
+                    startingUnifier, new HashSet<>()).stream();
+        }
+
+        private Set<Map<Reference, Set<Reference>>> matchRolesAndUnify(
```
flyingsilverfin, line 166: we should look into how to make this a lazy algorithm... as this set can be very large, we should be able to stream them lazily as required :)

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +119,86 @@ public Value asValue() {
...
+            Map<Reference, Set<Reference>> startingUnifier = new HashMap<>();
+            Optional<Map<Reference, Set<Reference>>> newUnifier;
+            if (constraint.owner().reference().isName()) {
+                newUnifier = updatedUnifier(this.constraint().owner(), unifyWith.constraint().owner(), startingUnifier);
+                if (newUnifier.isPresent()) startingUnifier = newUnifier.get();
+                else return Stream.empty();
+            }
+
+            if (constraint.owner().isa().isPresent() && constraint.owner().isa().get().type().reference().isName()) {
+                assert unifyWith.constraint().owner().isa().isPresent();
```
flyingsilverfin, line 155: good assertion

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +119,86 @@ public Value asValue() {
...
+                newUnifier = updatedUnifier(this.constraint().owner().isa().get().type(),
+                        unifyWith.constraint().owner().isa().get().type(), startingUnifier);
+                if (newUnifier.isPresent()) startingUnifier = newUnifier.get();
+                else return Stream.empty();
+            }
+
+            return matchRolesAndUnify(new ArrayList<>(this.constraint().players()), unifyWith.constraint().players(),
```
flyingsilverfin, line 162: this should be called `unifyRolePlayers` and you should just use the name `unifier` instead of `startingUnifier`. Check out my later comment on some ideas for this

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +119,86 @@ public Value asValue() {
...
+                    startingUnifier, new HashSet<>()).stream();
+        }
+
+        private Set<Map<Reference, Set<Reference>>> matchRolesAndUnify(
+                List<RolePlayer> conjunctionRolePlayers, List<RolePlayer> headRolePlayers,
+                Map<Reference, Set<Reference>> unifier, Set<RolePlayer> visitedHeadRolePlayers) {
```
flyingsilverfin, line 168: we're moving away from `head` to `then` terminology

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +119,86 @@ public Value asValue() {
...
+                updatedUnifier(conj, head, unifier).ifPresent(newUnifier ->
+                        allUnifiers.addAll(matchRolesAndUnify(conjunctionRolePlayers,
+                                headRolePlayers, newUnifier, visitedHeadRolePlayers)));
+                visitedHeadRolePlayers.remove(head);
+            }
+            conjunctionRolePlayers.add(conj);
+            return allUnifiers;
+        }
+
+        Optional<Map<Reference, Set<Reference>>> updatedUnifier(
```
flyingsilverfin, line 186: don't use past tense in general - should be `extendUnifier` or something

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +119,86 @@ public Value asValue() {
...
+                RolePlayer conjPlayer, RolePlayer headPlayer, Map<Reference, Set<Reference>> originalUnifier) {
+            if (!headPlayer.roleTypeHints().isEmpty() && !conjPlayer.roleTypeHints().isEmpty() &&
```
flyingsilverfin, line 188: can we make this more pseudo-codey by calling this `!contradictionExists(...)` or something?

flyingsilverfin: also, how about unification of this:
```
$x > 10;
```
and
```
$y < 5;
```

this seems like a contradiction that should be derived somehow?

flyingsilverfin (changes requested):

flyingsilverfin:
Ok so some thoughts, lets talk about it more tomorrow

logic/concludable/HeadConcludable.java
```diff
@@ -134,13 +124,11 @@ public Relation asRelation() {
...
         public Has(HasConstraint constraint, Set<Variable> constraintContext) {
-            super(copyConstraint(constraint), constraintContext);
+            super(constraint, constraintContext);
         }
 
-        @Override
-        ResourceIterator<Has> getGeneralisations() {
-            // TODO
-            return Iterators.iterate(Collections.singletonList(this));
+        public static Has copyOf(HasConstraint constraint, Set<Variable> constraintContext) {
```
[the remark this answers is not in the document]
maxbaxt: Changed to `create`

logic/concludable/ConjunctionConcludable.java
maxbaxt: Does that mean we should change the name of the class to `ThenConcludable`?

logic/concludable/ConjunctionConcludable.java
maxbaxt: OK, changed to `extendUnifier`.

logic/concludable/ConjunctionConcludable.java
maxbaxt: I did originally have a method for this but it's only used here, and we seem to prefer doing things inline when we can? Would be good to get a clarification on the style we prefer here.

maxbaxt: As to your second comment, I though about this. It seems that in this case it is not the unfier's job to prune at this point. The HeadConcludable will never be `$y<5`, as that contains an unnamed variable. It will be something like `$y<$num`. So the unifier should return `$x -> $y`, assuming the type hints are not disjoint.

logic/concludable/ConjunctionConcludable.java
maxbaxt: Good point, I will need some assistance in creating a lazy method in Java.

maxbaxt: Update: I've made this lazy, and it makes a nice functional program flow :)

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +119,107 @@ public Value asValue() {
...
+            if (constraint.owner().isa().isPresent() && constraint.owner().isa().get().type().reference().isName()) {
+                assert unifyWith.constraint().owner().isa().isPresent();
+                newUnifier = extendUnifier(this.constraint().owner().isa().get().type(),
+                        unifyWith.constraint().owner().isa().get().type(), variableUnifier);
+                if (newUnifier.isPresent()) variableUnifier = newUnifier.get();
+                else return Stream.empty();
+            }
+
+            Map<Reference, Set<Reference>> finalVariableUnifier = variableUnifier;
+            return unifyRolePlayers(this.constraint().players(), unifyWith.constraint().players(),
```
flyingsilverfin, line 163: can you wrap each of these lists in a `Collections.unmodifableList()` call to signal that they shuoldn't be cut up/modified?

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +119,107 @@ public Value asValue() {
...
+                            newMapping.putIfAbsent(conjRP.player().reference(), new HashSet<>());
+                            newMapping.get(conjRP.player().reference()).add(thenRP.player().reference());
+                        }
+                );
+            });
+
+            return newMapping;
+        }
+
+        private boolean roleHinstDisjoint(RolePlayer conjRP, RolePlayer thenRP) {
```
flyingsilverfin, line 205: typo

logic/concludable/Concludable.java
```diff
@@ -122,6 +126,11 @@ static void copyIsaAndValues(ThingVariable oldOwner, ThingVariable newOwner) {
 
     static void copyLabelAndValueType(TypeVariable copyFrom, TypeVariable copyTo) {
```
flyingsilverfin, line 127: shouldn't this method now be renamed?

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +119,107 @@ public Value asValue() {
         throw GraknException.of(INVALID_CASTING, className(this.getClass()), className(Value.class));
     }
 
+    Optional<Map<Reference, Set<Reference>>> extendUnifier(
+            Variable conjVar, Variable headVar, Map<Reference, Set<Reference>> variableMapping) {
+        if (varHintsDisjoint(headVar, conjVar)) return Optional.empty();
```
flyingsilverfin, line 124: i suspect that this method should not be returning an Optional at all -- everywhere you call it you immediately do an `isPresent()` check to extract the thing inside the optional. If you just do the `varHintsDisjoint()` call as a check before `extendUnifier()`, this won't need to return an optional at all, and the caller will not need to check for optional. It'll just be `if (varHintsDisjoint()) extendUnifier(...) else return Stream.empty()`

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +119,107 @@ public Value asValue() {
...
+                                    .map(thenRP -> extendRolePlayerUnifier(conjRP, thenRP, unifier))
+                                    .filter(Optional::isPresent).map(Optional::get)
+                                    .flatMap(newUnifier ->
+                                            unifyRolePlayers(conjRolePlayers.subList(1,length), thenRolePlayers, newUnifier)
+                                    ))
+            ).orElseGet(() -> Stream.of(unifier));
+
+        }
+
+        Map<Reference, Set<Reference>> convertRolePlayerUnifier(
```
flyingsilverfin, line 186: this meant to be private?

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +119,107 @@ public Value asValue() {
...
+                    && Collections.disjoint(conjRP.roleTypeHints(), thenRP.roleTypeHints())) return true;
+            if (varHintsDisjoint(conjRP.player(), thenRP.player())) return true;
+            if (conjRP.roleType().isPresent()) {
+                assert thenRP.roleType().isPresent();
+                return varHintsDisjoint(conjRP.roleType().get(), thenRP.roleType().get());
+            }
+            return false;
+        }
+
+        Optional<Map<RolePlayer, Set<RolePlayer>>> extendRolePlayerUnifier(
```
flyingsilverfin, line 216: private? can you check access modifiers everwhere? Also, public methods tend to come first and private methods later. This way you can see the class API at the very top, and the implementation follows. Also the "import" private methods i tend to put higher up as well, under the public API

flyingsilverfin (changes requested):

pattern/constraint/thing/RelationConstraint.java
```diff
@@ -177,7 +177,8 @@ public boolean equals(Object o) {
             if (this == o) return true;
             if (o == null || getClass() != o.getClass()) return false;
             final RolePlayer that = (RolePlayer) o;
-            return (Objects.equals(this.roleType, that.roleType)) && (this.player.equals(that.player));
+            return (Objects.equals(this.roleType, that.roleType)) && (this.player.equals(that.player)) &&
+                    this.roleTypeHints.equals(that.roleTypeHints);
```
flyingsilverfin, line 181: be careful with this - the hash hasn't changed... how is this going to play with hashing overall if we update type hitnts?

logic/concludable/ConjunctionConcludable.java
maxbaxt: Made that change.

logic/concludable/ConjunctionConcludable.java
maxbaxt: Fixed

logic/concludable/Concludable.java
maxbaxt: Good point.

logic/concludable/ConjunctionConcludable.java
maxbaxt: But then you've decoupled the failure of unifying. Essentially forcing every calling to know to do that check. I think using the Optional way is safer.

logic/concludable/ConjunctionConcludable.java
maxbaxt: Changed to private

logic/concludable/ConjunctionConcludable.java
maxbaxt: Changed to private

pattern/constraint/thing/RelationConstraint.java
maxbaxt: OK I've moved it back and we'll discuss this later.

logic/concludable/ConjunctionConcludable.java
flyingsilverfin: then the name is misleading, it should be `tryExtendUnifier`

logic/concludable/Concludable.java
```diff
@@ -120,15 +124,47 @@ static void copyIsaAndValues(ThingVariable oldOwner, ThingVariable newOwner) {
...
 
     static TypeVariable copyVariableWithLabelAndValueType(TypeVariable copyFrom) {
         TypeVariable copy = TypeVariable.of(copyFrom.identifier());
-        copyLabelAndValueType(copyFrom, copy);
+        copyLabelSubAndValueType(copyFrom, copy);
         return copy;
     }
 
+    static boolean hasNoHints(Variable variable) {
+        //TODO: refactor once new 'all things' symbol is introduced
```
flyingsilverfin, line 144: this TODO is obsolete no?

logic/concludable/ConjunctionConcludable.java
flyingsilverfin: don't need to call it `find`, it's just `unifiableRules()`

logic/concludable/Concludable.java
maxbaxt: No, that will be implemented when my other PR is merged.

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +118,107 @@ public Value asValue() {
...
+        }
+
+        private Stream<Map<RolePlayer, Set<Integer>>> unifyRP(
+                List<RolePlayer> conjRolePlayers, List<RolePlayer> thenRolePlayers,
+                Map<RolePlayer, Set<Integer>> unifier, Integer startCon) {
+
+            if (startCon == conjRolePlayers.size()) return Stream.of(unifier);
+            RolePlayer conjRP = conjRolePlayers.get(startCon);
+
+            return IntStream.range(0, thenRolePlayers.size())
```
flyingsilverfin, line 175: i think i see what you're doing, and i think using index is ok. I think it can be made marginally cleaner by doing the following:
make a small intermediate method called `withRolePlayers(constraint.players(), unifyWith.constraint().players(), finalVariableUnifier)`
this method internally does:
```
withRolePlayers(players, otherPlayers, baseUnifier) {
Stream<> rolePlayerMappingIndices =  unifyRP(...)
return rolePlayerMappingIndices.map(indexMap -> unifierWithRolePlayers(indexMap, baseUnifier))
}
```
this way the naming helps the reader understand what is going on with the numbers

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +118,107 @@ public Value asValue() {
...
+                    .filter(thenIdx -> unifier.values().stream().noneMatch(players ->
+                            players.contains(thenIdx)))
+                    .mapToObj(thenIdx ->
+                            tryExtendRolePlayerUnifier(conjRP, thenRolePlayers.get(thenIdx), thenIdx, unifier))
+                    .filter(Optional::isPresent).map(Optional::get)
+                    .flatMap(newUnifier ->
+                            unifyRP(conjRolePlayers, thenRolePlayers, newUnifier, startCon + 1));
```
flyingsilverfin, line 182: I think you can also "peel" using `.sublist()` on most lists (just checked arraylist and linkedlist implementations, its efficient) to remove the first element each time and get rid of the "startCon" index that you pass down and increment. This will make it easier to the reader as well.

logic/concludable/ConjunctionConcludable.java
```diff
@@ -112,21 +118,107 @@ public Value asValue() {
...
+        }
+
+        private Map<Reference, Set<Reference>> convertRolePlayerUnifier(
+                Map<RolePlayer, Set<Integer>> rolePlayerUnifier, List<RolePlayer> thenRolePlayers,
```
flyingsilverfin, line 186: the first var should be called "unifiedRolePlayerIndices" or something right?

flyingsilverfin (changes requested):

logic/concludable/ConjunctionConcludable.java
maxbaxt: Good suggestion, I will bare this in mind when using Lists with Java in the future.

logic/concludable/ConjunctionConcludable.java
maxbaxt: Indexing was not my first choice, but all other solutions I came up with didn't work because it kept stumbling on the fact that the `equals` method was too loose, while `==` was too strict (it would see RolePlayers that should have been the same as distinct). I agree that it makes the code more opaque, unfortunately.

maxbaxt: I don't I see the benefit of this intermediate function. I sympathise, however, with the story being unclear, so I've made a number of renamings. The `UnifyRP` is now called `matchRolePlayerIndices`. This is much more accurate to what it actually does. Now it's clear it doesn't return a unifier, just an indices matching. Now we take this index mapping and map it to a unifier. So now the code reads

```
            return matchRolePlayerIndices(conjRolePlayers, thenRolePlayers, new HashMap<>())
                    .map(indexMap -> mapRolePlayerMatchingToUnifier(indexMap, thenRolePlayers, baseVariableUnifier));
```

maxbaxt: The main problem now is that we are slightly two-faced, because the intermediate Stream is from RolePlayers to Set<Indices>. It would feel more consistent to do a mapping from Indiceis -> Set<Indices>. However, this will add more complexity to the code. I therefore think that the current solution a good tradeoff, but let me know what you think.

maxbaxt: There was one other solution I thought of, which is to wrap the RolePlayers into an Object so that they also contained their Index. I thought this was a little over elaborate, but if the current implementation is too terse, I will refactor it with that idea.

logic/concludable/ConjunctionConcludable.java
maxbaxt: Renamed to `mathcedRolePlayerIndices` in accord with the renaming outlined above.

flyingsilverfin (approved):

test/integration/logic/BUILD
```diff
@@ -21,57 +21,75 @@ load("@graknlabs_dependencies//builder/java:rules.bzl", "host_compatible_java_te
...
-        "//common/test:logback"
-    ],
-    resource_strip_prefix = "common/test",
 )
 
 host_compatible_java_test(
     name = "test-type-hinter",
     srcs = ["TypeHinterTest.java"],
-    test_class = "grakn.core.logic.TypeHinterTest",
+    data = [":basic-schema.gql"],
```
flyingsilverfin, line 49: does your auto format move these around?? haha data is not source/deps, so we tend to put it last. test_class is super important so it comes towards the top :D

test/integration/logic/concludable/UnifyTest.java
```diff
@@ -0,0 +1,783 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static grakn.common.collection.Collections.list;
+import static grakn.common.collection.Collections.set;
+import static grakn.core.pattern.variable.VariableRegistry.createFromThings;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class UnifyTest {
```
flyingsilverfin, line 46: I think it would be more our style to call this "UnificationTest" (noun not verb)

test/integration/logic/BUILD
maxbaxt: I think it must be. I thought I'd installed the Grakn xlm Haikal put on discord to format correctly. Thanks for spotting that.

test/integration/logic/concludable/UnifyTest.java
maxbaxt: Agreed changed.

jmsfltchr (approved):

lolski (approved):

lolski:
Merged with the approval from @flyingsilverfin and @jmsfltchr

PR #5936, opened 2020-12-11
TypeHints moved from Constraint Field to Variable Field

maxbaxt:

Moved the Type Hints (the set of possible types that a variable can be), from one of its constraints (`isaConstraint` for `ThingVariable` and `subConstraint` for `TypeVariable`) to `Variable`. `TypeHints` in `RolePlayer` remain unchanged.

We also introduce a new `isSatisfiable` flag onto `Variable`. This acts as a guarantee of the satisfiability of a `Variable`, and is important in the case of an empty `TypeHints` (which can mean either a Variable that can't be anything, or can be anything, depending upon this flag).

Field `typeHints` and its accompanying methods were moved from `isaConstraint` and `subConstraint` to `Variable`. Code calling on typeHints have been changed accordingly.

Added `isSatisfiable` flag, and added some logic to changing it when TypeHints are added to and adjusted in `TypeHinter` and `Rule`.


pattern/variable/Variable.java
```diff
@@ -52,6 +58,7 @@ public Reference reference() {
     }
 
     public void addTo(Traversal traversal) {
+        if (!typeHints.isEmpty()) traversal.types(identifier, typeHints);
```
maxbaxt, line 61: Don't we need to add the isSatisfiable information to the vertex as well?

pattern/variable/Variable.java
```diff
@@ -71,6 +78,34 @@ public ThingVariable asThing() {
...
+            return;
+        }
+        typeHints.retainAll(labels);
+        if (typeHints.isEmpty()) isSatisfiable = false;
+    }
+
+    public void removeHint(Label label) {
+        assert !(typeHints.isEmpty() && isSatisfiable);
+        typeHints.remove(label);
+    }
```
maxbaxt, line 99: removeHint() is only used when removing types that don't conform to being a sub of a hint of the variables sub constraint. When the Variable has no typeHints, this method is not used, and instead all the subTypes of it's sub constraint is inserted as its TypeHints.

logic/Rule.java
```diff
@@ -162,19 +160,12 @@ void validateCycles() {
...
-                        whenVar.get().asType().sub().ifPresent(whenSub -> thenVar.asType().sub().ifPresent(
-                                thenSub -> thenSub.retainHints(whenSub.getTypeHints())));
-                    }
-                });
+        then.variables().stream().filter(variable -> variable.identifier().isNamedReference())
+                .forEach(thenVar ->
+                        when.variables().stream()
+                                .filter(variable -> variable.identifier().equals(thenVar.identifier()))
+                                .findFirst().ifPresent(whenVar -> thenVar.retainHints(whenVar.typeHints()))
+                );
```
maxbaxt, line 168: Do tests exist for this method to ensure it still works as before?

logic/concludable/ConjunctionConcludable.java
```diff
@@ -170,9 +169,6 @@ public Isa(final IsaConstraint constraint) {
 
         @Override
         public Stream<Unification> unify(ThenConcludable.Isa headIsa) {
-            Set<Label> possibleLabels = constraint().getTypeHints();
-            Set<Label> otherLabels = headIsa.constraint().getTypeHints();
-            possibleLabels.retainAll(otherLabels);
             return Stream.of(new Unification(this, headIsa, null)); // TODO Add variable mapping
```
flyingsilverfin, line 172: this is properly repalced by your other PR right?

pattern/variable/ThingVariable.java
```diff
@@ -56,6 +57,7 @@
     private final Set<HasConstraint> hasConstraints;
     private final Set<ValueConstraint<?>> valueConstraints;
     private final Set<ThingConstraint> constraints;
+    private final Set<Label> typeHints;
```
flyingsilverfin, line 60: why is this here and on `Variable`?

flyingsilverfin (changes requested):

pattern/variable/ThingVariable.java
maxbaxt: That shouldn't be there, deleted.

logic/concludable/ConjunctionConcludable.java
maxbaxt: Indeed.

pattern/variable/Variable.java
```diff
@@ -74,6 +81,34 @@ public ThingVariable asThing() {
         throw GraknException.of(INVALID_CASTING, className(this.getClass()), className(ThingVariable.class));
     }
 
+    public void addHints(Set<Label> labels) {
```
flyingsilverfin, line 84: I thought i had pointed this out before hmmm maybe I didn't hit submit:
Unless you can convince me otherwise, I think the API for adding/retaining/modifying type hints should _not_ be overlapping with the `isSatisfiable()` API internally. An external user can `setSatisfiable(false/true)` from outside, but from the variable's point of view, it is disconnected from type hints completely. This means the following methods should be simple adders/removers/setters, and the external user should use them to determine satisfiability (if using hints to do this at all!)

maxbaxt: There was a misalignment with the use of `isSatisfiable` here. I've now moved the logic of changing `isSatisfiable` to outside of `Variable`, which means that we have to do these checks when some of these methods are called. In particular, I've made those checks on `TypeHinter`, and also on `Rule` where the `TypeHints` are pruned based on the `TypeHints` on the when conjunction. I suggest you confer that's the correct behaviour.

logic/Rule.java
```diff
@@ -162,19 +161,16 @@ void validateCycles() {
...
-                    }
-                });
+        then.variables().stream().filter(variable -> variable.identifier().isNamedReference())
+                .forEach(thenVar ->
+                        when.variables().stream()
+                                .filter(whenVar -> whenVar.identifier().equals(thenVar.identifier()))
+                                .filter(whenVar -> !(whenVar.isSatisfiable() && whenVar.typeHints().isEmpty()))
+                                .findFirst().ifPresent(whenVar -> {
+                                    thenVar.retainHints(whenVar.typeHints());
+                                    if (thenVar.typeHints().isEmpty()) thenVar.setSatisfiability(false);
```
maxbaxt, line 171: If the last of the TypeHints are pruned by the When then the variable needs to be set to Unsatisfiable.

logic/Rule.java
```diff
@@ -162,19 +161,18 @@ void validateCycles() {
...
+        then.variables().stream().filter(variable -> variable.identifier().isNamedReference())
+                .forEach(thenVar ->
+                        when.variables().stream()
+                                .filter(whenVar -> whenVar.identifier().equals(thenVar.identifier()))
+                                .filter(whenVar -> !(whenVar.isSatisfiable() && whenVar.typeHints().isEmpty()))
+                                .findFirst().ifPresent(whenVar -> {
+                                    if (thenVar.typeHints().isEmpty() && thenVar.isSatisfiable()) {
+                                        thenVar.addHints(whenVar.typeHints());
+                                    } else thenVar.retainHints(whenVar.typeHints());
+                                    if (thenVar.typeHints().isEmpty()) thenVar.setSatisfiability(false);
```
maxbaxt, line 173: If the last of the TypeHints are pruned by the When then the variable needs to be set to Unsatisfiable.

logic/tool/TypeHinter.java
```diff
@@ -77,17 +77,17 @@ public Conjunction computeHintsExhaustive(Conjunction conjunction, int paralleli
         Map<Label, TypeVariable> labelMap = labelVarsFromConjunction(conjunction);
         Map<Reference, Set<Label>> referenceHintsMapping =
                 retrieveVariableHints(new HashSet<>(variableHints.getVariableHints()), parallelisation);
-        long numOfThings = traversalEng.graph().schema().stats().thingTypeCount();
+        long numOfTypes = traversalEng.graph().schema().stats().thingTypeCount();
```
maxbaxt, line 80: Realised this name was inaccurate.

logic/tool/TypeHinter.java
```diff
@@ -177,15 +153,15 @@ private void removeHintsViolatingSuper(Variable subVar) {
         if (supVar == null) return;
         if (supVar.reference().isLabel()) return;
 
-        Set<Label> supLabels = getHintLabels(supVar);
-        Set<Label> subLabels = getHintLabels(subVar);
-        if (supLabels == null || supLabels.isEmpty()) return;
+        Set<Label> supLabels = supVar.typeHints();
```
flyingsilverfin, line 156: woop woop!

logic/tool/TypeHinter.java
```diff
@@ -177,15 +153,15 @@ private void removeHintsViolatingSuper(Variable subVar) {
...
+        Set<Label> subLabels = subVar.typeHints();
+        if (supLabels.isEmpty()) return;
 
-        if (subLabels == null) return;
         if (subLabels.isEmpty()) {
             Set<Label> subHintsOfSupLabels = supLabels.stream().flatMap(label -> getType(label).getSubtypes())
                     .map(Type::getLabel).collect(Collectors.toSet());
-            addHintLabels(subVar, subHintsOfSupLabels);
+            subVar.addHints(subHintsOfSupLabels);
+            if (subVar.typeHints().isEmpty()) subVar.setSatisfiability(false);
```
flyingsilverfin, line 164: how's it possible to add hints and its set to `false`??

pattern/constraint/thing/IsaConstraint.java
```diff
@@ -63,31 +58,10 @@ public boolean isExplicit() {
...
-        typeHints.clear();
-    }
-
-    public Set<Label> getTypeHints() {
-        return typeHints;
-    }
-
     @Override
     public void addTo(Traversal traversal) {
         // TODO: assert !(type.reference().isLabel() && typeHints.isEmpty());
```
flyingsilverfin, line 63: probably this assertion needs to move to wherever the type hints are now too?

logic/tool/TypeHinter.java
maxbaxt: `subHintsOfSupLabels` could be empty.

It's possible I might need to rename `addHintLabels` to `setHintLabels`

pattern/variable/Variable.java
```diff
@@ -74,6 +81,31 @@ public ThingVariable asThing() {
...
+
+    public Set<Label> typeHints() {
+        return typeHints;
+    }
+
+    public boolean isSatisfiable() {
+        return isSatisfiable;
+    }
+
+    public void setSatisfiability(boolean isSatisfiable) {
```
flyingsilverfin, line 105: i'd just call this `setSatisfiable` to only have 1 terminology

pattern/variable/Variable.java
```diff
@@ -74,6 +81,31 @@ public ThingVariable asThing() {
...
+    public void addHints(Set<Label> labels) {
+        typeHints.addAll(labels);
+    }
+
+    public void retainHints(Set<Label> labels) {
+        typeHints.retainAll(labels);
+    }
+
+    public void removeHint(Label label) {
+        assert !(isSatisfiable && typeHints.isEmpty());
```
flyingsilverfin, line 93: i think it'll be more readable with two negated ORs, but this is subjective :D

flyingsilverfin (approved):

pattern/variable/Variable.java
maxbaxt: Changed to `setIsSatisfiable`.

pattern/variable/Variable.java
maxbaxt: In general, I think `!(A ^ B)` and `(!A v !B)` are equally acceptable imho.

logic/tool/TypeHinter.java
maxbaxt: On second thoughts, we've already checked if supLabels is empty by this point, so I will remove this line.

pattern/constraint/thing/IsaConstraint.java
flyingsilverfin: @maxbaxt you did not remove this assertion........

maxbaxt: You mean the comment? Deleted

PR #5939, opened 2020-12-11

lriuui0x0:

Currently we save the string length as an unsigned byte (0-255). When we deserialize it, we cannot simply get the number as a byte, otherwise it will be treated as negative number during comparison.

- Convert unsigned byte to int during deserialization.

### DIFF
--- a/common/collection/Bytes.java
+++ b/common/collection/Bytes.java
@@ -169,8 +169,12 @@ public class Bytes {
     }
 
     public static String bytesToString(byte[] bytes, Charset encoding) {
-        final byte[] x = Arrays.copyOfRange(bytes, 1, 1 + bytes[0]);
+        final byte[] x = Arrays.copyOfRange(bytes, 1, 1 + unsignedByteToInt(bytes[0]));
         return new String(x, encoding);
+    }
+
+    public static int unsignedByteToInt(byte x) {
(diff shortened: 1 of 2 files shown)


common/collection/Bytes.java
```diff
@@ -169,10 +169,14 @@ public static double sortedBytesToDouble(byte[] bytes) {
     }
 
     public static String bytesToString(byte[] bytes, Charset encoding) {
-        final byte[] x = Arrays.copyOfRange(bytes, 1, 1 + bytes[0]);
+        final byte[] x = Arrays.copyOfRange(bytes, 1, 1 + unsignedByteToInt(bytes[0]));
         return new String(x, encoding);
     }
 
+    public static int unsignedByteToInt(byte x) {
+        return x & 0xff;
```
lolski, line 177: Can we write `ByteBuffer.wrap(content).getInt()` instead to make it look less h4ck3r?

haikalpribadi: I think this fine, and would be more performant, @lolski

haikalpribadi (approved):
haha well done, @lriuui0x0 :D

common/collection/Bytes.java
lolski: ok cool

PR #5940, opened 2020-12-11
Allow importing data into 2.0

lriuui0x0:

To let people migrate data from older version of grakn into 2.0, we added importer to import exported data from old grakn. The command line interface for invoking the importer largely stays the same, and the algorithm stays unchanged except for some code simplification.

- Add importer for the exported data from old grakn.
- Add command line interface for importer.
- Introduced various simplification over the the old importer.
    - Refactor names in the `migrator.proto`
    - Removed `Job.Completion` message and use `onCompleted` to communicate the same thing
    - Various code simplification in `ProgressPrinter`
    - Removed cache in the old importer, this is because we're already doing cache in the graph package. There's no need to cache again.


server/GraknServer.java
```diff
@@ -147,52 +151,64 @@ private static Properties parseProperties() {
...
-            final GraknServer server = new GraknServer(options);
-            server.start();
-
-            final long end = System.nanoTime();
-            LOG.info("Grakn Core version: {}", Version.VERSION);
-            LOG.info("Grakn Core Server has been started (in {} ms)",
-                     String.format("%.3f", (end - start) / 1_000_000.00));
-
-            server.serve();
+            final ServerCommand command = parseCommandLine(parseProperties(), args);
```
lriuui0x0, line 191: Here the server command line now may mean multiple commands, we created an interface `ServerCommand`, to abstract multiple commands. See `ServerCommand` class.

server/migrator/Importer.java
```diff
@@ -0,0 +1,311 @@
...
+    private final List<Pair<byte[], List<String>>> missingOwnerships = new ArrayList<>();
+    private final List<Pair<byte[], List<Pair<String, List<String>>>>> missingRolePlayers = new ArrayList<>();
+    private long totalThingCount = 0;
+    private long entityCount = 0;
+    private long relationCount = 0;
+    private long attributeCount = 0;
+    private long ownershipCount = 0;
+    private long playerCount = 0;
+    private Grakn.Transaction tx;
+    private int commitWriteCount = 0;
```
lriuui0x0, line 77: We removed `caches` compared with https://github.com/graknlabs/grakn/blob/1.8.4/server/migrate/Import.java, since hypergraph is already doing the cache.

server/migrator/MigratorClient.java
```diff
@@ -0,0 +1,80 @@
...
+        stub = MigratorGrpc.newStub(channel);
+    }
+
+    public boolean importData(String database, String filename, Map<String, String> remapLabels) {
+        MigratorProto.ImportData.Req req = MigratorProto.ImportData.Req.newBuilder()
+                .setDatabase(database)
+                .setFilename(filename)
+                .putAllRemapLabels(remapLabels)
+                .build();
+        CountDownLatch latch = new CountDownLatch(1);
```
lriuui0x0, line 48: latch used to wait for import completed.

server/migrator/proto/migrator.proto
```diff
@@ -0,0 +1,71 @@
...
+option java_package = "grakn.core.server.migrator.proto";
+option java_outer_classname = "MigratorProto";
+
+import "server/migrator/proto/data.proto";
+
+package migrator;
+
+service Migrator {
+  rpc exportData (ExportData.Req) returns (stream Job.Res);
+  rpc importData (ImportData.Req) returns (stream Job.Res);
```
lriuui0x0, line 33: Only `importData` is implemented so far.

server/rpc/MigratorRPCService.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public void exportData(MigratorProto.ExportData.Req request, StreamObserver<MigratorProto.Job.Res> responseObserver) {
+    }
+
+    @Override
+    public void importData(MigratorProto.ImportData.Req request, StreamObserver<MigratorProto.Job.Res> responseObserver) {
+        try {
+            Importer importer = new Importer(grakn, request.getDatabase(), Paths.get(request.getFilename()), request.getRemapLabelsMap());
+            CompletableFuture<Void> importerJob = CompletableFuture.runAsync(importer::run);
```
lriuui0x0, line 55: Here we run the importer on a different thread, and the main grpc thread keeps polling the progress and send to client to report the progress.

server/util/ServerCommand.java
```diff
@@ -0,0 +1,162 @@
...
+        @Parameters(  index = "0", description = "Database to import data into")
+        private String database;
+
+        @Parameters(index = "1", description = "File containing the data to import")
+        private String filename;
+
+        @Parameters(index = "2..*", arity = "0..*", description = "Schema concept remap labels")
+        private Map<String, String> remapLabels = new LinkedHashMap<>();
+
+        public ImportData(Start startCommand) {
```
lriuui0x0, line 136: `ImportData` command takes in `Start` command since we may want to do something like:

```
./grakn server --port 2000 import grabl grabl.data
```

We use the options at `Start` command to know where the server is.

server/migrator/proto/migrator.proto
```diff
@@ -0,0 +1,71 @@
...
+
+message ExportSchema {
+  message Req {
+    string database = 1;
+  }
+
+  message Res {
+    string schema = 1;
+  }
+}
```
alexjpwalker, line 71:
```suggestion
}

```

server/GraknServer.java
```diff
@@ -66,10 +68,11 @@
 
     private final Grakn grakn;
     private final Server server;
-    private final ServerOptions options;
+    private final ServerCommand.Start options;
     private final GraknRPCService graknRPCService;
+    private final MigratorRPCService migratorRPCService;
 
-    private GraknServer(ServerOptions options) throws IOException {
+    private GraknServer(ServerCommand.Start options) throws IOException {
```
alexjpwalker, line 75: Let's rename `options` (and the matching private field) to `command`.

server/migrator/MigratorClient.java
```diff
@@ -0,0 +1,80 @@
...
+        boolean[] result = new boolean[1];
```
alexjpwalker, line 49: Does this need to be a `boolean[1]`? I think a simple `boolean` would work.

server/rpc/MigratorRPCService.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(MigratorRPCService.class);
+    private final Grakn grakn;
+
+    public MigratorRPCService(Grakn grakn) {
+        this.grakn = grakn;
+    }
+
+    @Override
+    public void exportData(MigratorProto.ExportData.Req request, StreamObserver<MigratorProto.Job.Res> responseObserver) {
```
alexjpwalker, line 48: Are you planning to implement the exports immediately?

If not, these methods should probably throw exceptions - something like
`throw new UnsupportedOperationException("Export data|schema is not currently supported.");`

server/rpc/MigratorRPCService.java
```diff
@@ -0,0 +1,81 @@
...
+        try {
+            Importer importer = new Importer(grakn, request.getDatabase(), Paths.get(request.getFilename()), request.getRemapLabelsMap());
+            CompletableFuture<Void> importerJob = CompletableFuture.runAsync(importer::run);
+            try {
+                while (true) {
+                    try {
+                        importerJob.get(1, TimeUnit.SECONDS);
+                        break;
+                    } catch (TimeoutException e) {
+                        responseObserver.onNext(MigratorProto.Job.Res.newBuilder().setProgress(importer.getProgress()).build());
```
alexjpwalker, line 62: This flow is pretty strange. Is it intended that we only send a progress response on catching a `TimeoutException` or is that a typo?

alexjpwalker (changes requested):
Requested some changes and clarifications

server/migrator/MigratorClient.java
lriuui0x0: @alexjpwalker It's being modified in the lambda, that's why I put it in the array.

server/rpc/MigratorRPCService.java
lriuui0x0: It's the `CompletableFuture.get` that throws `TimeoutException`. Because we want to send progress repetitively to the client in the rate of one second.

server/rpc/MigratorRPCService.java
lriuui0x0: Done

lriuui0x0: Done

server/GraknServer.java
lriuui0x0: done.

server/migrator/MigratorClient.java
alexjpwalker: Why does that require it to go in an array, though? It's totally valid for a simple boolean to be modified by a method in that StreamObserver.

server/rpc/MigratorRPCService.java
alexjpwalker: I see. Is there a cleaner way to do this that doesn't involve throwing an exception by design? In general we try to follow the rule of "exceptions should be for unexpected scenarios" (at least in Java)

server/migrator/MigratorClient.java
lriuui0x0: Yep I see what you mean, I have fixed that.

server/BUILD
```diff
@@ -61,11 +72,15 @@ native_java_libraries(
 
         # External dependencies from Maven
         "@maven//:com_google_code_findbugs_jsr305",
+        "@maven//:com_google_guava_guava",
```
flyingsilverfin, line 75: hmmm we should not be importing guava unless we _really_ have to! It's a large dependency we've moved away from. When we need something from Guava we either re-implement it or find another way right now

server/GraknServer.java
```diff
@@ -147,52 +151,64 @@ private static Properties parseProperties() {
...
-            final long end = System.nanoTime();
-            LOG.info("Grakn Core version: {}", Version.VERSION);
-            LOG.info("Grakn Core Server has been started (in {} ms)",
-                     String.format("%.3f", (end - start) / 1_000_000.00));
-
-            server.serve();
+            final ServerCommand command = parseCommandLine(parseProperties(), args);
+            if (command == null) System.exit(0);
+
+            if (command instanceof ServerCommand.Start) {
```
flyingsilverfin, line 194: our pattern would be to implement `ServerCommand.isStart()` in the interface as a default boolen returning false, then overriding it in each subclass once

server/GraknServer.java
```diff
@@ -147,52 +151,64 @@ private static Properties parseProperties() {
...
+                        String.format("%.3f", (end - start) / 1_000_000.00));
+
+                server.serve();
+            } else if (command instanceof ServerCommand.ImportData) {
+                ServerCommand.ImportData importDataCommand = command.asImport();
+                MigratorClient migrator = new MigratorClient(importDataCommand.port());
+                boolean success = migrator.importData(importDataCommand.database(), importDataCommand.filename(), importDataCommand.remapLabels());
+                System.exit(success ? 0 : 1);
+            } else {
+                assert false;
```
flyingsilverfin, line 210: I think you show throw an exception here, such as invalid state, or something similar rather than use an assertion

server/migrator/Importer.java
```diff
@@ -0,0 +1,311 @@
...
+    private final Map<String, byte[]> idMap = new HashMap<>();
+    private final List<Pair<byte[], List<String>>> missingOwnerships = new ArrayList<>();
+    private final List<Pair<byte[], List<Pair<String, List<String>>>>> missingRolePlayers = new ArrayList<>();
+    private long totalThingCount = 0;
+    private long entityCount = 0;
+    private long relationCount = 0;
+    private long attributeCount = 0;
+    private long ownershipCount = 0;
+    private long playerCount = 0;
+    private Grakn.Transaction tx;
```
flyingsilverfin, line 76: can we put all the `count` variables together :P looks odd to see the tx in the middle here

alexjpwalker (approved):

server/migrator/Importer.java
```diff
@@ -0,0 +1,311 @@
...
+                Collectors.toMap(x -> x.getLabel().scopedName(), x -> x));
+    }
+
+    private String relabel(String label) {
+        return remapLabels.getOrDefault(label, label);
+    }
+
+    private void mayCommit() {
+        commitWriteCount++;
+        if (commitWriteCount >= BATCH_SIZE) {
```
flyingsilverfin, line 298: i think `txWriteCount` is clearer name as it signals that its used only for lifetime of one commit?

server/migrator/Importer.java
```diff
@@ -0,0 +1,311 @@
...
+            commit();
+        }
+    }
+
+    private void commit() {
+        LOG.debug("Commit start, inserted {} things", commitWriteCount);
+        long time = System.nanoTime();
```
flyingsilverfin, line 305: still using nanotime here instead of Instant, is that intended?

server/rpc/MigratorRPCService.java
flyingsilverfin: discussed in person

flyingsilverfin (changes requested):
i actually liked your original title using the word "import" somewhere in it!

server/GraknServer.java
lriuui0x0: Actually I can just delete the else clause.

server/migrator/Importer.java
lriuui0x0: No, that's copied over, will change to `Instant`.

server/BUILD
```diff
@@ -72,7 +72,6 @@ native_java_libraries(
 
         # External dependencies from Maven
         "@maven//:com_google_code_findbugs_jsr305",
-        "@maven//:com_google_guava_guava",
```
lriuui0x0: Actually without guava, project fails to build.

alexjpwalker: What is the error?

Generally speaking Guava methods can be trivially implemented using our own code.

lriuui0x0: @alexjpwalker It fails to build the GRPC generated code, there's nothing we can do to implement.

flyingsilverfin (approved):

server/BUILD
lriuui0x0: unfortunately this is needed because we need it to compile grpc.

PR #5943, opened 2020-12-14

lriuui0x0:

Recently we have changed the string length deserialization using two bytes to unsigned short. However, Java performs numeric promotion during integer operations (see [ref](https://docs.oracle.com/javase/specs/jls/se7/html/jls-5.html#jls-5.6.2)), which means `0xff` of a byte value will be first promoted into `0xffffffff` of an int value before performing any operations. This results in the unexpected result of high order bytes not being cleared.

- Force clear the high order bytes during unsigned short deserialization.


haikalpribadi (approved):
LOL

PR #5946, opened 2020-12-15
Fix traversal scopes to return all combinations of role players

flyingsilverfin:
The traversal engine was not correctly providing both (x,y) and (y,x) answers to relations, where roles duplicate or not specified. This PR sets uniqueness bounds and and clears them again (which we had not done before).

* clear up the scoping sets when iterating to the next answer
* fix `TraversalTest2` to match the BDD scenario `all combinations of players in a relation can be retrieved`, notably including the missing `answers.next()`!

### DIFF
--- a/test/integration/traversal/TraversalTest2.java
+++ b/test/integration/traversal/TraversalTest2.java
@@ -33,6 +33,7 @@ import graql.lang.query.GraqlDefine;
 import graql.lang.query.GraqlInsert;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


test/integration/traversal/TraversalTest2.java
```diff
@@ -131,46 +131,57 @@ public void all_combinations_of_players_in_a_relation_can_be_retrieved_0_repeat(
...
         proc.backwardIsa(1, relation, r, true);
         proc.forwardRolePlayer(2, r, x, set());
-        proc.forwardIsa(3, x, person, true);
+        proc.forwardRolePlayer(3, r, y, set());
 
         try (RocksTransaction transaction = session.transaction(READ)) {
             ResourceIterator<VertexMap> vertices = transaction.traversal().iterator(proc.build(), params);
             ResourceIterator<ConceptMap> answers = transaction.concepts().conceptMaps(vertices);
-            assertTrue(answers.hasNext());
+            AttributeType.Long ref = transaction.concepts().getAttributeType("ref").asLong();
```
flyingsilverfin, line 170: add same assertions as in above variant `_0`

haikalpribadi (approved):

PR #5948, opened 2020-12-15
Allow exporting data from 2.0

lriuui0x0:

To let people migrate data from 2.0 to future version of grakn, we added exporter to export data from current grakn database. The command line interface for invoking the exporter largely stays the same, and the algorithm stays unchanged except for some code simplification.

- Add exporter for exporting data from grakn database.
- Add command line interface for exporter.
- Reduced number of classes in the old exporter.

### DIFF
--- a/concept/type/ThingType.java
+++ b/concept/type/ThingType.java
@@ -35,8 +35,6 @@ public interface ThingType extends Type {
 
     Stream<? extends Thing> getInstances();
 
-    long getInstancesCount();
-
     void setAbstract();
 
     void unsetAbstract();
--- a/concept/type/ThingType.java
+++ b/concept/type/ThingType.java
@@ -35,6 +35,8 @@ public interface ThingType extends Type {
 
     Stream<? extends Thing> getInstances();
 
+    long getInstancesCount();
+
     void setAbstract();
 
     void unsetAbstract();
--- a/server/migrator/Exporter.java
+++ b/server/migrator/Exporter.java
@@ -34,6 +34,7 @@ public class Exporter implements Migrator {
     private static final Logger LOG = LoggerFactory.getLogger(Exporter.class);
     private final Grakn.Session session;
     private final Path filename;
+    private long totalThingCount = 0;
     private final AtomicLong entityCount = new AtomicLong(0);
     private final AtomicLong relationCount = new AtomicLong(0);
     private final AtomicLong attributeCount = new AtomicLong(0);
@@ -50,8 +51,7 @@ public class Exporter implements Migrator {
         long current = attributeCount.get() + relationCount.get() + entityCount.get();
         return MigratorProto.Job.Progress.newBuilder()
                 .setCurrent(current)
-                // TODO
-                .setTotal(100000000)
+                .setTotal(totalThingCount)
                 .build();
     }
 
@@ -60,6 +60,7 @@ public class Exporter implements Migrator {
         LOG.info("Exporting {} from Grakn {}", session.database().name(), Version.VERSION);
         try (OutputStream outputStream = new BufferedOutputStream(Files.newOutputStream(filename))) {
             try (Grakn.Transaction tx = session.transaction(Arguments.Transaction.Type.READ)) {
+                totalThingCount = tx.concepts().getRootThingType().getInstancesCount();
                 DataProto.Item header = DataProto.Item.newBuilder()
                         .setHeader(DataProto.Item.Header.newBuilder()
                                 .setGraknVersion(Version.VERSION)
--- a/server/migrator/Exporter.java
+++ b/server/migrator/Exporter.java
@@ -59,14 +59,14 @@ public class Exporter implements Migrator {
     private final AtomicLong playerCount = new AtomicLong(0);
     private long totalThingCount = 0;
 
-    public Exporter(Grakn grakn, String database, Path filename) {
+    public Exporter(final Grakn grakn, final String database, final Path filename) {
         this.session = grakn.session(database, Arguments.Session.Type.DATA);
         this.filename = filename;
     }
 
     @Override
     public MigratorProto.Job.Progress getProgress() {
-        long current = attributeCount.get() + relationCount.get() + entityCount.get();
+        final long current = attributeCount.get() + relationCount.get() + entityCount.get();
         return MigratorProto.Job.Progress.newBuilder()
                 .setCurrent(current)
                 .setTotal(totalThingCount)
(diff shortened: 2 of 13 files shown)


concept/type/ThingType.java
```diff
@@ -35,6 +35,8 @@
 
     Stream<? extends Thing> getInstances();
 
+    long getInstancesCount();
```
alexjpwalker, line 38: I wonder if this is really the right place for such a method. Do you know what the scope of statistics are? Presumably they will be used to optimise far more things than just retrieving instance counts, right?

If so I wonder if it warrants the creation of a new separate class, called say `ConceptStatistics` that is responsible for storing all statistics to do with concepts, and would include (for example) `instancesCount`, `ownsCount`, `linkedConceptCount` .... etc. Or is it just going to be `getInstancesCount` for the foreseeable future?

server/migrator/Exporter.java
```diff
@@ -0,0 +1,197 @@
...
+                .setTotal(totalThingCount)
+                .build();
+    }
+
+    @Override
+    public void run() {
+        LOG.info("Exporting {} from Grakn {}", session.database().name(), Version.VERSION);
+        try (OutputStream outputStream = new BufferedOutputStream(Files.newOutputStream(filename))) {
+            try (Grakn.Transaction tx = session.transaction(Arguments.Transaction.Type.READ)) {
+                totalThingCount = tx.concepts().getRootThingType().getInstancesCount();
```
alexjpwalker, line 81: Something worth getting into the habit of is marking as `final` local variables whose value never changes, such as this `totalThingCount`.

Not only does this make algorithms easier to parse (as we don't have to consider the possibility of variables being mutated further down the method) but it will have an additional benefit if/when we migrate to a stricter language such as Rust: in Rust, variables are immutable by default. So we will give special attention to variables that are not marked as `final` in Java.

This can actually be refactored very easily in IntelliJ using:

1. Analyze
2. Run Inspection by Name
3. Local variable or parameter can be final
4. Uncheck "Report method parameters"
5. Fix All

server/migrator/Exporter.java
```diff
@@ -0,0 +1,197 @@
...
+            valueObject.setDatetime(attribute.asDateTime().getValue().atZone(ZoneId.of("Z")).toInstant().toEpochMilli());
+        } else {
+            throw GraknException.of(ILLEGAL_STATE);
+        }
+        return valueObject;
+    }
+
+    private Stream<DataProto.Item.OwnedAttribute.Builder> readOwnerships(Thing thing) {
+        return thing.getHas().map(attribute -> {
+            ownershipCount.incrementAndGet();
```
alexjpwalker, line 184: I'm not a huge fan of doing this `incrementAndGet` inside a `map` lambda expression. It could have unexpected side effects if, say, the code were changed to call `readOwnerships` multiple times. As a rule of thumb, performing state changes inside a `map` is dangerous.

Could we move it outside the map - perhaps into `readEntity`, `readAttribute` and `readRelation`? I know it would mean some code duplication but I think the flow would be clearer and safer.

server/migrator/Importer.java
```diff
@@ -139,6 +141,7 @@ public void run() {
         insertMissingOwnerships();
         insertMissingRolePlayers();
         commit();
+        tx.close();
```
alexjpwalker, line 144: The recommended way of closing Grakn Transactions in Java is using try-with-resources:
```java
try (tx = session.transaction(Arguments.Transaction.Type.WRITE)) {
   ...
}
```

server/migrator/MigratorClient.java
```diff
@@ -57,6 +57,23 @@ public boolean importData(String database, String filename, Map<String, String>
...
+        MigratorProto.ExportData.Req req = MigratorProto.ExportData.Req.newBuilder()
+                .setDatabase(database)
+                .setFilename(filename)
+                .build();
+        CountDownLatch latch = new CountDownLatch(1);
+        ProgressPrinter progressPrinter = new ProgressPrinter("export");
+        ResponseObserver streamObserver = new ResponseObserver(progressPrinter, latch);
+        stub.exportData(req, streamObserver);
+        try {
+            latch.await();
```
alexjpwalker, line 70: I know the `ResponseObserver` isn't currently touched in this PR but is there any reason for the `CountDownLatch` to be initialized _outside_ the Observer?

I think it would be cleaner if the `CountDownLatch` was initialized inside the constructor of `ResponseObserver`, and `ResponseObserver` provided its own `await()` method that simply called `latch.await()`.

alexjpwalker (changes requested):
Requested some changes

concept/type/ThingType.java
lriuui0x0: I think this question should be raised to Haikal

server/migrator/Exporter.java
lriuui0x0: `totalThingCount` is not a final variable, it's a field in class that will be read by the progress reporting thread.

server/migrator/Exporter.java
lriuui0x0: Nah, this needs to be incremented for each edge, will see what I can do there.

server/migrator/Importer.java
lriuui0x0: Actually `commit` will also `close` it, I will just delete that line.

server/migrator/MigratorClient.java
lriuui0x0: Could do that.

server/migrator/Exporter.java
lriuui0x0: Do we have a strict rule on this? Having everything to be final is quite noisy.

server/migrator/Exporter.java
```diff
@@ -0,0 +1,197 @@
...
+                }));
+                workers.add(() -> tx.concepts().getRootRelationType().getInstances().forEach(relation -> {
+                    DataProto.Item item = readRelation(relation);
+                    write(outputStream, item);
+                }));
+                workers.add(() -> tx.concepts().getRootAttributeType().getInstances().forEach(attribute -> {
+                    DataProto.Item item = readAttribute(attribute);
+                    write(outputStream, item);
+                }));
+                workers.parallelStream().forEach(Runnable::run);
```
flyingsilverfin, line 102: can you confirm this actually runs in parallel? I think we've seen that this sometimes is parallel and sometimes is not. May be better to use an Async executor service to guarantee this happens as desired.

server/migrator/Migrator.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package grakn.core.server.migrator;
+
+import grakn.core.server.migrator.proto.MigratorProto;
+
+public interface Migrator {
+
+    void run();
+    MigratorProto.Job.Progress getProgress();
```
flyingsilverfin, line 26: newline between interface APIs

server/migrator/Exporter.java
lriuui0x0: This is copied from old code, will double check.

server/migrator/Migrator.java
lriuui0x0: ok

server/util/ServerCommand.java
```diff
@@ -49,6 +49,14 @@ default ImportData asImportData() {
         throw GraknException.of(ILLEGAL_CAST);
     }
 
+    default boolean isExportData() {
+        return false;
+    }
+
+    default ExportData asExportData() {
+        throw GraknException.of(ILLEGAL_CAST);
```
flyingsilverfin, line 57: doesn't our ILLEGAL_CAST take parameters saying which class was being cast to which other class?

flyingsilverfin (approved):
3 comments individually

server/migrator/Exporter.java
lriuui0x0: `parallelStream` does return a stream with `isParallel` set to true.

server/util/ServerCommand.java
lriuui0x0: Yep I missed that, will add.

server/migrator/Exporter.java
lriuui0x0: Ok moving them into `readEntity/...` seems to be the easiest.

server/migrator/Exporter.java
alexjpwalker: It is Java's fault that they chose to make variables non-final by default. Adding `final` to variables is definitely a workaround.

But it shouldn't be seen as noise - code implemented using final variables is usually a lot more readable!

server/migrator/Importer.java
alexjpwalker: Ah I see that now. Would be nice if we could refactor it to be a local variable but... if it's a lot of work then don't worry about it.

concept/type/ThingType.java
alexjpwalker: did Haikal respond?

lriuui0x0: Not yet, I already sent him a discord message

server/migrator/Exporter.java
lriuui0x0: OK I finalized everything.

alexjpwalker (approved):

PR #5949, opened 2020-12-15
Rule Then's Convert Variables to Named

maxbaxt:

- Whenever a rule is inserted, its Then clause is made to have only named variables.

* Implemented methods in Rule to, after the conversion of a Then Clause from Graql to Grakn, we reconstruct it if necessary to a named version
* In effect, this means changing anonymous variables to temporary named ones, using the SystemReference object.


logic/Rule.java
```diff
@@ -63,7 +74,7 @@ private Rule(LogicManager logicManager, RuleStructure structure) {
     }
 
     private Rule(GraphManager graphMgr, ConceptManager conceptMgr, LogicManager logicManager, String label,
-                 graql.lang.pattern.Conjunction<? extends Pattern> when, ThingVariable<?> then) {
+                 graql.lang.pattern.Conjunction<? extends Pattern> when, graql.lang.pattern.variable.ThingVariable<?> then) {
```
maxbaxt, line 77: I think that when we use both analogous objects from both Grakn and Graql, we should use the "long import" for the Object that is foreign to the current repo. Therefore, I've changed all `ThingVariable<?>` to `graql.lang.pattern.variable.ThingVariable<?>` in this PR.

logic/Rule.java
```diff
@@ -200,8 +208,100 @@ private Conjunction whenPattern(graql.lang.pattern.Conjunction<? extends Pattern
...
-    private Conjunction thenPattern(ThingVariable<?> thenVariable) {
-        return new Conjunction(VariableRegistry.createFromThings(list(thenVariable)).variables(), set());
+    private Conjunction thenPattern(graql.lang.pattern.variable.ThingVariable<?> thenVariable) {
+        Set<Variable> unnamedThenVars = VariableRegistry.createFromThings(list(thenVariable)).variables();
+        return new Conjunction(nameThenVars(unnamedThenVars), set());
+    }
+
+    private Set<Variable> nameThenVars(Set<Variable> variables) {
+        for (Variable variable : variables) {
+            if (variable.isThing() && !variable.asThing().has().isEmpty()){
```
flyingsilverfin, line 218: can move the second conditions inside of a single `if` block

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -317,7 +317,7 @@ public void addTo(Traversal traversal) {
 
     public static class Variable extends ValueConstraint<ThingVariable> {
 
-        public Variable(ThingVariable owner, GraqlToken.Predicate.Equality predicate, ThingVariable variable) {
+        public Variable(ThingVariable owner, GraqlToken.Predicate predicate, ThingVariable variable) {
```
flyingsilverfin, line 320: This change is not good - now it makes it look like we can do a say `substring` or `like` (regex) comparision between two variables, which we don't actually support! I'd revert this and undo moving the big `if` switch back into its prior place

pattern/variable/ThingVariable.java
```diff
@@ -91,6 +92,14 @@ public void addTo(Traversal traversal) {
         constraints().forEach(constraint -> constraint.addTo(traversal));
     }
 
+    public static ThingVariable createTemp(String name) {
```
flyingsilverfin, line 95: these should go into the `ConstraintCopier` imo

pattern/variable/ThingVariable.java
```diff
@@ -138,7 +147,23 @@ public IsConstraint is(ThingVariable variable) {
         return valueConstraints;
     }
 
-    public ValueConstraint.Long valueLong(GraqlToken.Predicate.Equality comparator, long value) {
+    public <T> ValueConstraint<?> value(GraqlToken.Predicate comparator, T value) {
```
flyingsilverfin, line 150: these are the big switch statements im talking about

pattern/variable/TypeVariable.java
```diff
@@ -82,6 +83,14 @@ public static TypeVariable of(Identifier.Variable identifier) {
         return new TypeVariable(identifier);
     }
 
+    public static TypeVariable createTemp(String name) {
```
flyingsilverfin, line 86: also to `ConstraintCopier` i think, not general enough

flyingsilverfin (changes requested):

flyingsilverfin:
Superceded by #5977

PR #5952, opened 2020-12-16
Allow 2.0 to dump currently defined schema and rules

lriuui0x0:

In order for people to easily get the currently defined schema in query format for them to easily rerun, we implemented a new server command to dump the currently defined schema and rules. 

- Implemented a new server command `./grakn server schema <database>`.
- Implemented a new blocking rpc call `getSchema`.
- Implemented the schema pretty printer that convert schema into a string format.


server/migrator/MigratorClient.java
```diff
@@ -106,4 +118,80 @@ public boolean success() {
             return success;
         }
     }
+
+    private static class ProgressPrinter {
```
lriuui0x0, line 122: This class is only moved here, not modified. This is to make the top level classes less noisy.

server/migrator/Importer.java
```diff
@@ -142,6 +142,7 @@ public void run() {
         insertMissingRolePlayers();
         commit();
         tx.close();
+        session.close();
```
lriuui0x0, line 145: Missed a session close here.

server/GraknServer.java
```diff
@@ -214,6 +216,10 @@ public static void main(String[] args) {
                 MigratorClient migrator = new MigratorClient(exportDataCommand.port());
                 boolean success = migrator.exportData(exportDataCommand.database(), exportDataCommand.filename());
                 System.exit(success ? 0 : 1);
+            } else if (command.isGetSchema()) {
```
alexjpwalker, line 219: Now that we have three different commands here, I think it would make sense to extract each `if` block into its own method. Something like: `runGraknServer`, `importData`, `getSchema`.

server/migrator/Schema.java
```diff
@@ -0,0 +1,181 @@
...
+                    .getSubtypes()
+                    .filter(x -> !x.isRoot())
+                    .sorted(Comparator.comparing(x -> x.getLabel().name()))
+                    .forEach(x -> writeRelationType(builder, x));
+            tx.concepts()
+                    .getRootEntityType()
+                    .getSubtypes()
+                    .filter(x -> !x.isRoot())
+                    .sorted(Comparator.comparing(x -> x.getLabel().name()))
+                    .forEach(x -> writeEntityType(builder, x));
```
alexjpwalker, line 57: Let's extract `.getSubtypes().filter(x -> !x.isRoot()).sorted(Comparator.comparing(x -> x.getLabel().name()))` into its own method, something like:
```java
private <TYPE extends Type> Stream<TYPE> subtypesSortedByLabel(TYPE type) {
    return type.getSubtypes().filter(x -> !x.isRoot()).sorted(Comparator.comparing(x -> x.getLabel().name()));
}
```

server/migrator/Schema.java
```diff
@@ -0,0 +1,181 @@
...
+                .append("when\n")
+                .append(getPatternString(wrapConjunction(rule.getWhenPreNormalised()), 1))
+                .append(COMMA_NEWLINE_INDENT)
+                .append("then\n")
+                .append(getPatternString(wrapConjunction(rule.getThenPreNormalised()), 1))
+                .append(SEMICOLON_NEWLINE_X2);
+    }
+
+    private String getValueTypeString(AttributeType.ValueType valueType) {
+        switch (valueType) {
```
alexjpwalker, line 134: I think it's fine to replace this method with
```java
return valueType.name().toLowerCase();
```
If you want to add a safety net you can add
`if (valueType == AttributeType.ValueType.OBJECT) throw GraknException.of(ILLEGAL_STATE);` but I think that's optional.

server/migrator/Schema.java
```diff
@@ -0,0 +1,181 @@
...
+            throw GraknException.of(ILLEGAL_STATE);
+        }
+    }
+
+    private Pattern wrapConjunction(Pattern pattern) {
+        return pattern.isConjunction() ? pattern : new Conjunction<>(list(pattern));
+    }
+
+    private static String indent(int indent) {
+        return IntStream.range(0, indent * 4).mapToObj(i -> " ").collect(Collectors.joining(""));
```
alexjpwalker, line 179: I think we can replace the last `collect` with `Collectors.joining()`?

server/util/ServerCommand.java
```diff
@@ -222,5 +230,36 @@ public ExportData asExportData() {
             return this;
         }
     }
+
+    @Command(name = "schema")
+    class GetSchema implements ServerCommand {
```
alexjpwalker, line 235: Should this maybe be `PrintSchema`? And the command could be `print schema`.

alexjpwalker (changes requested):

server/util/ServerCommand.java
lriuui0x0: It used to be called just `schema`

server/migrator/Schema.java
lriuui0x0: ok

server/migrator/Schema.java
lriuui0x0: Emm... I'd rather not to do that, I don't want to rely on the enum names. What happens if you rename `DATETIME` to `DATE_TIME`.

server/GraknServer.java
lriuui0x0: could do.

server/migrator/Schema.java
lriuui0x0: discussed in person, this is harder to do.

server/migrator/Schema.java
```diff
@@ -0,0 +1,201 @@
...
+
+    private void writePlays(StringBuilder builder, ThingType thingType) {
+        thingType.getPlays().sorted(Comparator.comparing(x -> x.getLabel().scopedName()))
+                .forEach(x -> builder
+                        .append(COMMA_NEWLINE_INDENT)
+                        .append(String.format("plays %s", x.getLabel().scopedName())));
+    }
+
+    private void writeRule(StringBuilder builder, Rule rule) {
+        builder.append(String.format("%s sub rule", rule.getLabel()))
```
flyingsilverfin, line 143: this is not correct rule syntax anymore, please use 2.0 syntax :)

server/migrator/proto/migrator.proto
```diff
@@ -40,6 +39,24 @@ message ExportData {
...
+message ImportData {
+  message Req {
+    string database = 1;
+    string filename = 2;
+    map<string, string> remap_labels = 3;
+  }
+}
+
+message GetSchema {
+  message Req {
```
flyingsilverfin, line 51: one question we never got to before: can you test this with a giant schema? just like auto generate one in python or something. We don't currently know how big of a schema export we can do (has to do with grpc message size limit). We know there's 1 or 2 users out there with giant schemas, but might be ok to break, but also would be good to know what the limit is at all

flyingsilverfin (changes requested):
see individual comments

flyingsilverfin:
@lriuui0x0  where are all the tests?? I'm pretty sure there was a cyclical "import file&data, export file&data, check that they match" or something??

lriuui0x0:
>  @lriuui0x0 where are all the tests?? I'm pretty sure there was a cyclical "import file&data, export file&data, check that they match" or something??

@flyingsilverfin Oh I didn't know there was a test .

server/util/ServerCommand.java
alexjpwalker: But the command for server start is called `grakn server`, right?

server/migrator/Schema.java
alexjpwalker: Fair enough

concept/type/AttributeType.java
```diff
@@ -38,6 +38,9 @@
     @Override
     Stream<? extends AttributeType> getSubtypes();
 
+    @Override
+    Stream<? extends AttributeType> getSubtypesDirect();
```
alexjpwalker: We were going to rename these methods to `getSubtypesExplicit`, etc, right?

concept/type/ThingType.java
```diff
@@ -51,17 +54,27 @@
...
 
+    Stream<? extends AttributeType> getOwnsDirect(boolean onlyKey);
+
     Stream<? extends AttributeType> getOwns(AttributeType.ValueType valueType);
 
+    Stream<? extends AttributeType> getOwnsDirect(AttributeType.ValueType valueType);
+
     Stream<? extends AttributeType> getOwns(AttributeType.ValueType valueType, boolean onlyKey);
 
+    Stream<? extends AttributeType> getOwnsDirect(AttributeType.ValueType valueType, boolean onlyKey);
```
alexjpwalker: Are all of these overloads used? I think it's enough to just implement the ones that we use for the exporter

server/migrator/Schema.java
```diff
@@ -59,20 +59,17 @@ public String getSchema() {
              final Grakn.Transaction tx = session.transaction(Arguments.Transaction.Type.READ)) {
             tx.concepts()
                     .getRootAttributeType()
-                    .getSubtypes()
-                    .filter(x -> !x.isRoot())
+                    .getSubtypesDirect()
```
alexjpwalker: I like that this method has resulted in the removal of `filter(x -> !x.isRoot())` 👍

alexjpwalker (changes requested):

concept/type/AttributeType.java
lriuui0x0: Yes already renamed, 2 seconds after you start reviewing.

server/migrator/Schema.java
lriuui0x0: Cool.

concept/type/ThingType.java
lriuui0x0: Emm...I think from an API point of view, it's better to make the complete. If we provide an API for `getOwns` then there should be an API for `getOwnsDirect`?

alexjpwalker: Yeah that's fine as well.

alexjpwalker (approved):

flyingsilverfin (approved):

PR #5957, opened 2020-12-16
Answer State Tracking

jmsfltchr:

We improve that way that the state of answers in reasoner is tracked throughout the reasoner. This new architecture aims to make clear which variables domain an answer state is in via it's namespace, and therefore which state it can move to next.

This needs to support differing methods for unified answers, mapped answers and answers for the root resolver. Most crucially that we can always expect a valid answer after a 1:1 variable mapping, but an answer may prove invalid when unified or un-unified.


reasoner/resolution/answer/AnswerState.java
```diff
@@ -0,0 +1,235 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+import static grakn.common.util.Objects.className;
+import static grakn.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
+import static grakn.core.common.exception.ErrorMessage.Pattern.INVALID_CASTING;
+
+public abstract class AnswerState {
```
flyingsilverfin, line 35: looking better! May i suggest the following naming change to see if anything falls out of it:
`UpstreamVars` -> `RequesterVars`
`DownstreamVars` -> `ResponderVars`

i think "upstream/downstream" (at least for me) is too closely tied to the shape of the actor execution, whereas "request/responder" is more generic? then it can read as `AnswerState.RequesterVars.Partial`, do you like that better?

should also replace occurrences of "Map<Ref, Set<Ref>>` with Unifier now i think

jmsfltchr: Totally open to re-hashing the names. Have gone with upstream/downstream as that's made things very clear in the implementation. Currently the usage goes: 
`fromUpstream.partial().aggregateWith(conceptMap).asMapped().toUpstreamVars()`
with this proposal it would be:
`fromUpstream.partial().aggregateWith(conceptMap).asMapped().toRequesterVars()`
what I like about the former is that it's easy to tell that it's the variables of the "parent" actor, whereas in the latter this actor is also a requester. DOn't know if this argument really holds or if I just like the upstream/downstream variable naming 🤔

jmsfltchr: Yes, can introduce `Unifier` as a data structure.

flyingsilverfin: didn't you arleady do that`?

flyingsilverfin: ok lets stick with your names for now :)

jmsfltchr: Here actually I call the transformer that applies the mapping a `Unifier` - so we should rename this to... a `Unificator`? 🤣 
But really, what shall we call these two things?

jmsfltchr: Let's get this merged and address the unifier naming when we implement the logic for unifier variable mapping and filtering.

reasoner/ReasonerProducer.java
```diff
@@ -22,11 +22,12 @@
 import grakn.core.concept.answer.ConceptMap;
```
flyingsilverfin, line 22: lets make the usage of `Mapping` and `Unifier` consistent everywhere

reasoner/resolution/answer/AnswerState.java
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(conceptMap);
+    }
+
+    public static class UpstreamVars {
+
+        public static class Partial extends AnswerState {
```
flyingsilverfin, line 60: last comment to make this go in: can we call this "Initial"? This would make the transition more clear i think: `Upstream.Initial` -> `Downstream.Mapped/Downstream.Unified` -> `Upstream.Derived`?

haikalpribadi (approved):

Issue #5955, opened 2020-12-16
Grakn falsely claims that an attribute type cannot be set to abstract as its subtypes are not abstract, when redefining existing schema


VladGan:
## Description

Grakn falsely claims that an attribute type cannot be set to abstract as its subtypes are not abstract, when redefining existing schema 

## Environment

1. OS (where Grakn server runs): { Mac OS 10 }
2. Grakn version (and platform): { Grakn Core 2.0a  }
3. Grakn client: { core }

## Reproducible Steps

```gherkin
When graql define
      """
      define name sub attribute, abstract, value string; location-name sub name;
      """
    Then transaction commits
    Then session opens transaction of type: write
    Then graql define
      """
      define name sub attribute, abstract, value string; location-name sub name;
      """
    Then the integrity is validated
```

## Expected Output

It should work.

## Actual Output

[TYW12] Invalid Type Write: The attribute type 'name' cannot be set to abstract as its subtypes are not abstract.

PR #5958, opened 2020-12-16
Fix logic of setting abstract on attribute types

VladGan:

Redefining existing schema follows different logic to creating a fresh schema. There was a bug where Grakn would
falsely claim that an attribute type couldn't be set to abstract as its subtypes were not abstract, which was
only triggered when redefining existing schema.

Also, unsetting abstract now checks that there are no subtypes.

close https://github.com/graknlabs/grakn/issues/5955

 - don't check subtypes when setting abstract on an attribute type;
 - check that there are no subtypes when unsetting abstract on attribute type.

### DIFF
--- a/dependencies/graknlabs/repositories.bzl
+++ b/dependencies/graknlabs/repositories.bzl
@@ -55,6 +55,6 @@ def graknlabs_grabl_tracing():
 def graknlabs_behaviour():
     git_repository(
         name = "graknlabs_behaviour",
-        remote = "https://github.com/VladGan/behaviour",
-        commit = "754afdade47a81652b487912eda5406fad8d8878", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_behaviour
+        remote = "https://github.com/graknlabs/behaviour",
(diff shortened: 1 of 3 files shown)


dependencies/graknlabs/repositories.bzl
```diff
@@ -55,6 +55,6 @@ def graknlabs_grabl_tracing():
 def graknlabs_behaviour():
     git_repository(
         name = "graknlabs_behaviour",
-        remote = "https://github.com/graknlabs/behaviour",
-        commit = "a5689e86fe816b10c6017281b4c2717bee47937e", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_behaviour
+        remote = "https://github.com/VladGan/behaviour",
+        commit = "754afdade47a81652b487912eda5406fad8d8878", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_behaviour
```
VladGan, line 59: https://github.com/graknlabs/behaviour/pull/122

haikalpribadi (approved):

PR #5959, opened 2020-12-16
Concludables and Retrievables are Resolvables

jmsfltchr:

We refactor the naming in reasoner to make way for splitting a Conjunction into a set of Resolvables. Resolvables are elements of the query that can be resolved either as a Retrievable using direct DB traversal, or as a Concludable that can be unified with one of the Conclusions from each rule.

- `HeadConcludable` -> `Rule.Conclusion`
- `ConjunctionConcludable` -> `Concludable extends Resolvable`
- Add `Retrievable extends Resolvable`
- Move static methods from `Concludable` to `tools`


logic/Rule.java
```diff
@@ -46,8 +57,8 @@
     private final RuleStructure structure;
     private final Conjunction when;
     private final Conjunction then;
-    private final Set<ThenConcludable<?, ?>> possibleThenConcludables;
-    private final Set<ConjunctionConcludable<?, ?>> requiredWhenConcludables;
+    private final Set<Conclusion<?>> possibleThenConcludables;
```
flyingsilverfin, line 60: var names need to be updated too

flyingsilverfin (approved):

jmsfltchr:
There is a potential integration issue where `TypeVariable.copyConstraints(TypeVariable toCopy)` could mean that the constraint back-links become duplicated and need to be removed.

PR #5960, opened 2020-12-16
Resolution Reiteration and rearchitect Resolvers

flyingsilverfin:
Implement resolution reiteration, which triggers clearing and resetting `Resolver` states before re-traversing the graph for more answers that may not have been discovered in the previous iterations. We also rearchitect the resolvers to allow for a more logical control flow, at the cost of re-introducing some more code duplication.

* Implement resolution reiteration, which is initiated and managed by the `ReasonerProducer`
* Iteration number is carried alongside requests, as this means we can continue to use the same requests and have clear request equalities, continuing to use the same key for a particular `ResponseProducer`
* `Resolver` abstract type does less work - specialised resolvers directly handle their `Map<Request, ResponseProducer>`. However, request routing is handled at the framework level still.
* Remove the wrapper methods that wrapped implementation methods. ie. no more `void executeReceiveRequest()` which calls `Either<Request, Response> receiveRequest()`, as it was not flexible enough. Now, we only have `void receiveRequest()`
* Remove middle `Conjunction` resolver that was shared by root and rule as it was overly restrictive
* Add traversal engine into `Resolver` classes
* Implement a test that _may_ require re-iteration


reasoner/Reasoner.java
```diff
@@ -46,12 +48,14 @@
...
     private final LogicManager logicMgr;
     private final ResolverRegistry resolverRegistry;
+    private final Actor<ResolutionRecorder> resolutionRecorder; // for explanations
 
     public Reasoner(ConceptManager conceptMgr, TraversalEngine traversalEng, LogicManager logicMgr) {
         this.conceptMgr = conceptMgr;
         this.traversalEng = traversalEng;
         this.logicMgr = logicMgr;
-        this.resolverRegistry = new ResolverRegistry(ExecutorService.eventLoopGroup());
+        this.resolutionRecorder = Actor.create(ExecutorService.eventLoopGroup(), ResolutionRecorder::new);
```
flyingsilverfin, line 57: hold onto the `ResolutionRecorder` for the future steps where we will need to access it to extract explanations.

reasoner/resolution/framework/Resolver.java
```diff
@@ -30,107 +29,59 @@
...
 
-    protected final String name;
-    private final Map<Request, ResponseProducer> responseProducers;
+    private final String name;
     private final Map<Request, Request> requestRouter;
-    private boolean isInitialised;
+    protected final ResolverRegistry registry;
 
-    public Resolver(Actor<T> self, String name) {
+    protected Resolver(Actor<T> self, String name, ResolverRegistry registry) {
```
flyingsilverfin, line 36: resolver registry is now passed in constructor, but should not be used until `initialiseDownstreamActors` is called

reasoner/resolution/framework/Resolver.java
```diff
@@ -30,107 +29,59 @@
...
 
-    protected abstract Either<Request, Response> receiveExhausted(Request fromUpstream, Response.Exhausted fromDownstream, ResponseProducer responseProducer);
+    protected abstract ResponseProducer responseProducerCreate(Request fromUpstream, int iteration);
 
-    /*
-     *
-     * Handlers for messages sent into the execution actor that are dispatched via the actor model.
-     *
-     */
-    public void executeReceiveRequest(Request fromUpstream, ResolverRegistry registry) {
```
flyingsilverfin, line 65: the "meat" of these methods, which is actually not that much, has been moved down into each actor. This was mostly to facilitate the `RootResolver`, which does not require maintaining a `Map<Request, ResponseProducer>`, nor does it have to return a `Response` to the parent (it does not have a parent!)

reasoner/resolution/framework/Resolver.java
```diff
@@ -30,107 +29,59 @@
...
+        return requestRouter.get(toDownstream);
     }
 
-    /*
-     *
-     * Helper method private to this class.
-     *
-     * */
-    private void requestFromDownstream(Request request, Request fromUpstream, ResolverRegistry registry) {
+    protected void requestFromDownstream(Request request, Request fromUpstream, int iteration) {
```
flyingsilverfin, line 66: this and the following method are called by subclasses to do the "framework" bit of dispatching requests/responses and recording routing

reasoner/resolution/framework/ResponseProducer.java
```diff
@@ -75,4 +81,16 @@ public void removeDownstreamProducer(Request request) {
...
+
+    public int iteration() {
+        return iteration;
+    }
+
+    /**
+     * Prepare a response producer for the another iteration from this one
+     * Notably maintains the set of produced answers for deduplication
+     */
+    public ResponseProducer newIteration(Iterator<ConceptMap> traversalProducer, int iteration) {
```
flyingsilverfin, line 93: create a new derived `ResponseProducer` from the current one, which saves the set of previously produced answers

jmsfltchr (approved):

lolski:
Tests have run successfully, just not being shown with status icons on github. Here's the test run: https://grabl.io/flyingsilverfin/grakn/354313342049fdb9c4bef5953d249c4bc24340d4

PR #5969, opened 2020-12-19
Propagate errors from our asynchronous producers to consumers

lolski:

We have fixed the issue where `Producer` implementations do not propagate errors to the consumers when they crash. Since the consumers aren't informed of the errors, they may stall forever waiting for results that will never arrive.

- Update the `Done` object to allow for error propagation:
  - construct it with `Done.error(error)` to signal an error
  - otherwise, construct it with `Done.success()`
- Adds exception propagation across every implementation of `Producer.Sink`
- `GraknException`  now takes in `Throwable` rather than `Exception`. This would allow us to propagate not only exceptions but also errors such as `AssertionError`s.

### DIFF
--- a/common/producer/BaseProducer.java
+++ b/common/producer/BaseProducer.java
@@ -32,13 +32,19 @@ public class BaseProducer<T> implements Producer<T> {
     @Override
     public void produce(Sink<T> sink, int count) {
         ExecutorService.forkJoinPool().submit(() -> {
-            for (int i = 0; i < count; i++) {
-                if (iterator.hasNext()) {
-                    sink.put(iterator.next());
-                } else {
-                    sink.done(this);
-                    break;
+            try {
+                for (int i = 0; i < count; i++) {
+                    if (iterator.hasNext()) {
+                        sink.put(iterator.next());
+                    } else {
+                        sink.done(this);
+                        break;
+                    }
                 }
+            }
+            catch (Throwable e) {
+                sink.done(this, e);
+                throw e;
             }
         });
     }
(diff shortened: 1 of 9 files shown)


common/producer/BaseProducer.java
```diff
@@ -32,14 +32,20 @@
...
+                    if (iterator.hasNext()) {
+                        sink.put(iterator.next());
+                    } else {
+                        sink.done(this);
+                        break;
+                    }
                 }
             }
+            catch (Throwable e) {
+                sink.done(this, e);
```
lolski, line 46: if anything throws, immediately get out of the loop and signal done.

traversal/producer/VertexProducer.java
```diff
@@ -45,7 +45,12 @@ public VertexProducer(GraphManager graphMgr, ProcedureVertex<?, ?> vertex, Trave
 
     @Override
     public void produce(Sink<VertexMap> sink, int count) {
-        if (future == null) future = runAsync(consume(count, sink), forkJoinPool());
+        if (future == null) {
+            future = runAsync(consume(count, sink), forkJoinPool()).exceptionally(e -> {
```
lolski, line 49: Whenever we have to spawn an async job, we need not forget to add exception handling. If we forget, the consumer may stall and stalling is hard to debug.

I am open to any ideas on making it less error-prone.

lolski:
Merging the PR after the tests have passed: https://grabl.io/lolski/grakn/7105c5204218b8acd19f0df8b7446f6ac671df69

For some reason, Grabl does not send Github notifications for the pipelines and @vmax is currently looking into it.

PR #5970, opened 2020-12-19
Fix check style

lriuui0x0:

Fix check style test that causes the build to fail.


haikalpribadi:
It has a conflict so I'll just add this myself, @lriuui0x0

PR #5973, opened 2020-12-20
Extract Retrievables from a Conjunction

jmsfltchr:

When finding answers to a conjunction is is necessary to break it down into Concludable parts, which can be resolved by a rule; and Retrievable parts, whose answers can be retrieved directly from the DB. Here we introduce a component to divide a conjunction into Retrievables given a set of Concludables that will be resolved by rules, and therefore should be omitted from the retrievable parts.

- An extractor class that lives inside `Retrievable` to do the extraction described above.
- Back-links for variables so that they don't only know what constraints constrain them, but also what constraints they are involved in as the constraining variable. This makes the algorithm for extracting Retrievables much more natural.
- `ResolvableResolver` as the parent resolver of `ConcludableResolver` and `RetrievableResolver`.


jmsfltchr:
Issue that needs fixing: 

We need to make _all_ core constraints of a Concludable visible, not just the one we consider the most important.

For example we need to know that a `has` concludable infers a `has`, `isa` and `value` if it does so. Presently we do not, and therefore this conjunction: 
```
{ $p isa person, has name $n; $n "Alice"; }
```
will yield 2 concludables, an `isa` for `$p isa person` and a `has` for the name.
With the current logic a retrievable of `$n "Alice" isa name;` will be created in addition, because this constraint is not listed as the constraint of either concludable. This will produce incorrect results, since the name Alice may not yet exist, and in this case retrieving it will fail, even though the rule infers its existence.

jmsfltchr:
There is a potential integration issue where TypeVariable.copyConstraints(TypeVariable toCopy) could mean that the constraint back-links become duplicated and need to be removed.

logic/resolvable/Concludable.java
```diff
@@ -53,16 +55,24 @@
 
     private final Map<Rule, Set<Unifier>> applicableRules;
     private final CONSTRAINT constraint;
+    private final Set<Constraint> coreConstraints;
 
-    private Concludable(CONSTRAINT constraint) {
+    private Concludable(CONSTRAINT constraint, Set<Constraint> otherCoreConstraints) {
```
flyingsilverfin, line 60: So this will get outdated quickly right? It will just take a Conjunction in the next iteration?

logic/resolvable/Concludable.java
```diff
@@ -182,8 +192,8 @@ public Conjunction conjunction() {
 
     public static class Relation extends Concludable<RelationConstraint> {
 
-        public Relation(final RelationConstraint constraint) {
-            super(ConstraintCopier.copyConstraint(constraint));
+        public Relation(final RelationConstraint constraint, @Nullable IsaConstraint isaConstraint) {
+            super(ConstraintCopier.copyConstraint(constraint), isaConstraint == null ? set() : set(isaConstraint));
```
flyingsilverfin, line 196: and these constructors will take a Conjunction of the correct shape already i think, then save pointers to the `isa` and `relation` constraints in that conjunction right?

logic/resolvable/Retrievable.java
```diff
@@ -35,4 +76,196 @@ public Retrievable asRetrievable() {
...
+            return connectedSubgraphs.stream()
+                    .filter(subGraph -> !(subGraph.size() == 1 && subGraph.iterator().next().constraints().size() == 0))
+                    .map(subGraph -> new Retrievable(new Conjunction(subGraph, set()))).collect(Collectors.toSet());
+        }
+
+        private void addToConnected(Variable copy) {
+            connectedSubgraph.add(copy);
+        }
+
+        private TypeVariable addTypeVariable(TypeVariable variable) {
```
flyingsilverfin, line 119: I believe this is the right time to use `register` - `registerTypeVariable()` and `registerThingVariable()`

logic/resolvable/Retrievable.java
```diff
@@ -35,4 +76,196 @@ public Retrievable asRetrievable() {
...
+            return addVariable(variable).asType();
+        }
+
+        private ThingVariable addThingVariable(ThingVariable variable) {
```
flyingsilverfin, line 123: hmmm okay this led me down a train of thought:
1. each retrievable should be its own subgraph
2. conceptually, if it were possible for `$x` to be present in two different subgraphs, we would have a bug because the `variablesMap` is not re-instantiated per subgraph. Even if the subgraphs can never share a variable, its a bit unclear that the `variablesMap` holds variables remapped from different subgraphs?

I think its a tad confusing/smelly because you have a stateful "subgraph" that you add to during a single recursion, and then re-instantiate.

Possibly a cleaner pattern is:
you have a `SubgraphRegister` that you instantiate. You pick your starting variable, then do `subgraphRegister.registerVariable(variable)`. This will do all the registering of variables _within_ the `SubgraphRegister` object, which solves the problem/conceptual oddity in point 2 above. Now, each subgraph has its own `VariablesMap` that is only uses internally.
After the recursion ends you add all the variables and constraints registered in this subgraph into a global `visitedVariables`/`visitedConstraints`.

Does that make sense, not sure if unclear in explaining lol

pattern/constraint/thing/RelationConstraint.java
```diff
@@ -53,6 +53,10 @@ public RelationConstraint(ThingVariable owner, List<RolePlayer> rolePlayers) {
         assert rolePlayers != null && !rolePlayers.isEmpty();
         this.rolePlayers = new ArrayList<>(rolePlayers);
         this.hash = Objects.hash(RelationConstraint.class, this.owner, this.rolePlayers);
+        for (RelationConstraint.RolePlayer rp : rolePlayers) {
+            rp.player().constraining(this);
+            rp.roleType().ifPresent(roleType -> roleType.constraining(this));
```
flyingsilverfin, line 58: out of curiosity, is there a functional difference between doing `roleType.constraining(rp)` instead of `this`? Is a `RelationConstraint.RolePlayer` a constraint itself?

flyingsilverfin (changes requested):

logic/resolvable/Concludable.java
jmsfltchr: Yes that's the plan!

logic/resolvable/Concludable.java
jmsfltchr: Also sounds correct 🤔

logic/resolvable/Retrievable.java
jmsfltchr: That does make sense, however to avoid infinite recursion we need to update the visited variables and constraints as we recurse, but I think these can be held locally to the subgraph, and then added to the global visited maps once that subgraph is complete.

pattern/constraint/thing/RelationConstraint.java
jmsfltchr: I hadn't been considering `Roleplayer` as a separate constraint, I don't thiiink that makes sense. What we need is to traverse the variables of the relation constraint, so if we used the RolePlayer that operation would become quite clumsy.

flyingsilverfin (approved):

haikalpribadi (approved):

Issue #5921, opened 2020-12-08
Reasoning: unification via query rewriting


flyingsilverfin:
## Problem to Solve
In various situations, it would be useful to "carry forward" certain constraints from a query that is being unified into a rule. In the Reasoner 2.0 execution model, we only carry concepts between resolution actors, and explicitly avoid rewriting queries as it can lead to recursively rewriting queries. We want to re-explore this option but try to re-implement it in a way that avoid recursive rewrites.

The scope of this should include carrying forward type hints to restrict search space downstream a particular resolution actor.

## Proposed Old vs new model

In the old model, we may have 
query:          `$a has $b;`
rule `then`:  `$p has weight 5;

To handle the anonymous variable `weight 5`, we convert it to a non-anomyous variable (`$/weight`) which we can unify with `$b`. The variable mapping we compute is then: 
```
$a -> $p
$b -> $/weight
```

In the new model, we could imagine that a `Unifier` consists of a set of `targetTransformations` which we use to modify the receiver instead of concept maps:
```
$p -> renameTo($a)
weight 5 -> deanonimise($b)
```
The result of processing the rule `then` will produce concept maps in the correct variable space: `$a` and `$b`.

We could extend this further to carry through type hints that we have on the query `$a` and `$b`, for instance we might imagine it as:
```
$p -> restrictTypes(renameTo($a), [Type_A, Type_Sub_A])
weight 5 -> restrictTypes(deanonimise($b), [Type_B])
```

We then need to avoid recursively rewriting the rule `when` after modifying the `then`.

We also need to consider further complex variable mapping cases such as:

query: `($x, $x)`
then:   `($x, $y)`
Note: this one is interesting because it asserts that although in the rule `$x` and `$y` are different, we only want them when they are the same. Our current approach means unifier must filter out when `$x != $y`, but query rewriting would allow us to carry forward the constraint that `$x is $y`, or alternatively just rewrite them both to the same variable.

and
query: `($x, $y)`
then:   `($x, $x)`

These cases were what prevented rule rewriting for unification before, and we should be especially cautious here.
An idea that might be relevant was come up with by max:
we can treat unifier as two mappings: one from source to a new middle variable space, and another from the target to the same middle variable space:
`$a -> $NEW, $b -> $NEW2`
and
`$p -> $NEW, weight 5 -> $NEW2`

PR #5977, opened 2020-12-22
Rearchitect Unifiers to include Requirements

flyingsilverfin:
We modify the architecture of `Unifier` to impose `Requirements` when doing reverse unification - in lieu of query rewriting, we must filter some materialised rule answers out when they do not match the user's exact query. 
We also redesign the general flow of computing `Unifier`s in `Concludable` to have a clearer flow, using a `Unifier.Builder` to avoid needless copying. This also leads to simplifying the states in `AnswerState`.

## Unifier "Requirements"
A major decision to this point was to avoid recursive query rewriting for the time being (https://github.com/graknlabs/grakn/issues/5921). To implement correct unification, a `Unifier` may now impose `Requirements` on a reverse unification that is being performed. We may require a Variable is one of a set of `isaExplicit` (exact type of an instance), or a type in a set of `types` (exact label of a type), or a `predicate`.

__isaExplicit__ and __predicate__ requirements stem from the following case:
query: `match $p has name "john"`
rule: `when {$x isa person; $y isa first-name;} then {$x has $y}`

Here, the rule my indicate that an instance has an attribute, but the user only requests concepts with `name` attributes, that are specifically `= "john"`. So in the Unifier, we have approximately: 
`Unifier.requirements().isaExplicit()` contains `set("name", "first-name", "last-name")` 
and
`Unifier.requirements().predicates()` contains `(attr) -> attr.value().equals("john")`

__type__ requirements stem from the following case:
query: `match $r isa friendship`
rule: `when {$x isa person;} then { (best-friend: $x) isa best-friendship;}`

Here, we unify `$r` with `$_0` (anonymous variable for relationship instance in the rule `when`), and unify `$_friendship` (a labeled variable) to `$_best-friendship` (corresponding labelled variable). To make a valid answer, `$_best-friendship` must be compatible with the requirements of `$_friendship` from the query - ie. that it be of type `friendship`, or any of its subtypes (a `best-friendship` would in fact satisfy the Requirements).

Note: this also requires that a Unifier now computes a mapping between `Identifier`, rather than `Reference.Name`, and must be the output of any rule `then` materialisation.

Note: `Requirements` can later be replaced by query rewriting, if we proceed carefully and deal with all the edge cases of rewriting.

* Unifiers now have a component called `Requirements` that can be used to filter out answers during reverse unification
* Materialisation and unification now operate over `Map<Identifier, Concept>`
* leads to: updating `AnswerState` to make the flow clearer: `AnswerState.UpstreamVars.Initial` -> `AnswerState.DownstreamVars.Mapped/Unified` -> `Answerstate.UpstreamVars.Derived`
* delete `//logic/transformer`: move `//logic/transformer/Unifier` -> `//logic/resolvable/Unifier` and `//logic/transformer/Mapping` -> `//reasoner/resolution/answer/Mapping`
* refactor and extend unification tests to more closely follow how unification should look in the execution flow of reasoning
* Minor bug fixes and architectural changes in `resolver`s


traversal/common/Predicate.java
```diff
@@ -351,7 +351,7 @@ public Value(Encoding.ValueType valueType) {
 
             public abstract boolean apply(ARG_VAL_OP operator, AttributeVertex<?> vertex, ARG_VAL_TYPE value);
 
-            private static int compareDoubles(double first, double second) {
+            public static int compareDoubles(double first, double second) {
```
flyingsilverfin, line 354: until we have query rewriting, we have a sort of re-implementation of `Predicate` evaluation in Reasoner (only to handle the case `match $x has name "john"` which may trigger a rule `then { $x has $a }` but is only valid when `$a` happens to be type of `name` with value `= "john"`.

pattern/constraint/thing/RelationConstraint.java
```diff
@@ -133,7 +133,6 @@ public AlphaEquivalence alphaEquals(RelationConstraint that) {
 
         private final TypeVariable roleType;
         private final ThingVariable player;
-        private final Set<Label> resolvedRoleTypes;
```
flyingsilverfin, line 136: we no longer require `resolvedRoleTypes`, as was discussed by haikal and max, so I've completely removed them (the state here was unused!)

test/integration/QueryTest.java
```diff
@@ -137,6 +137,12 @@ public void test_query_undefine() throws IOException {
                     GraqlUndefine query = Graql.parseQuery(queryString);
                     transaction.query().undefine(query);
 
+                    // undefine `performance-tracker-rule` first because it depends on an undefined role performance-tracker:tracker
+                    // else would throw -- TODO write a test to prevent undefining types used in rules
```
flyingsilverfin, line 141: this is an issue that will be addressed when we implement `Rule` type indexing

traversal/common/Predicate.java
flyingsilverfin: so we only evaluate `=` predicates in reasoner. we have it in our roadmap to revisit this requirement already.

logic/resolvable/Concludable.java
```diff
@@ -373,15 +503,41 @@ public Value(final ValueConstraint<?> constraint) {
...
+             */
+            if (constraint().isVariable()) {
+                assert constraint().value() instanceof ThingVariable;
+                ThingVariable value = (ThingVariable) constraint().value();
+                if (unificationSatisfiable(constraint().predicate(), valueConclusion.constraint().predicate())) {
+                    unifierBuilder.add(value.identifier(), valueConclusion.constraint().owner().identifier());
+                } else return Iterators.empty();
+                // } else {
+                // form: $x > 10 -> require $x to satisfy predicate > 10
+                // TODO after restructuring concludables, we whould revisit requirements on `Value` concludables
```
flyingsilverfin, line 529: this is commented with a note because the immediate follow up to be implemented is to rearchitect reasoner data structures to get rid of this `else`, among with some other inconsistencies/code smells

haikalpribadi (approved):
I only skimmed the PR, but approving it first. Left a note on hash codes, which could be applied to multiple places.

PR #5981, opened 2020-12-28
Partially clean-up TypeResolver and disable it (temporarily)

haikalpribadi:

As I was debugging `TypeResolver` to fix some bugs that it causes to break Match BDD, I found myself having to interpret some code that is more convoluted than necessary, so I started cleaning them up. I haven't fully cleaned up the entire source code file of TypeResolver - I only managed to clean up `SystemReference`, `TypeResolver.ResolverVariables`, and a bit of `ConstraintMapper`. However, there are already a few glaring questions marks that need to be resolved, which I cannot do with more context TypeResolver. So I've disabled `TypeResolver.resolveVariable.resolveVariablesExhaustive()` from reasoner temporarily, to proceed with fixing the rest of match query bugs.

### DIFF
--- a/logic/tool/TypeResolver.java
+++ b/logic/tool/TypeResolver.java
@@ -277,31 +277,35 @@ public class TypeResolver {
         }
 
         private TypeVariable convert(Variable variable) {
-            if (resolvers.contains(variable)) return resolvers.resolver(variable);
-
             if (variable.isType()) return convert(variable.asType());
             else if (variable.isThing()) return convert(variable.asThing());
             else throw GraknException.of(ILLEGAL_STATE);
         }
 
         private TypeVariable convert(TypeVariable variable) {
+            if (resolvers.contains(variable)) return resolvers.resolver(variable);
+
             TypeVariable resolver = resolvers.register(variable);
             copyNeighbours(variable, resolver);
+            // TODO: Why do we need `neighbours.putIfAbsent(resolver, new HashSet<>());` ?
+            //       Is this a patch for the fact that we forgot to use `to` variable in `copyNeighbours(from, to)` ?
             neighbours.putIfAbsent(resolver, new HashSet<>());
             return resolver;
         }
 
-        private TypeVariable convert(ThingVariable thingVariable) {
-            TypeVariable resolver = resolvers.register(thingVariable);
+        private TypeVariable convert(ThingVariable variable) {
+            if (resolvers.contains(variable)) return resolvers.resolver(variable);
+
+            TypeVariable resolver = resolvers.register(variable);
             neighbours.putIfAbsent(resolver, new HashSet<>());
 
-            if (thingVariable.constraints().isEmpty()) return resolver;
+            if (variable.constraints().isEmpty()) return resolver;
 
-            thingVariable.isa().ifPresent(constraint -> convertIsa(resolver, constraint));
-            thingVariable.is().forEach(constraint -> convertIs(resolver, constraint));
-            thingVariable.has().forEach(constraint -> convertHas(resolver, constraint));
-            thingVariable.value().forEach(constraint -> convertValue(resolver, constraint));
-            thingVariable.relation().forEach(constraint -> convertRelation(resolver, constraint));
+            variable.isa().ifPresent(constraint -> convertIsa(resolver, constraint));
+            variable.is().forEach(constraint -> convertIs(resolver, constraint));
+            variable.has().forEach(constraint -> convertHas(resolver, constraint));
+            variable.value().forEach(constraint -> convertValue(resolver, constraint));
+            variable.relation().forEach(constraint -> convertRelation(resolver, constraint));
             return resolver;
         }
 
@@ -395,7 +399,8 @@ public class TypeResolver {
                 else if (constraint.isPlays()) addNeighbours(from, constraint.asPlays().role());
                 else if (constraint.isRelates()) addNeighbours(from, constraint.asRelates().role());
                 else if (constraint.isIs()) addNeighbours(from, constraint.asIs().variable());
-                else throw GraknException.of(ILLEGAL_STATE);
+                // TODO: There are other constraints in a TypeVariable. Are we intentionally ignoring them?
+                //       Why do we not throw GraknException.of(ILLEGAL_STATE); ?
             }
         }
     }
--- a/logic/tool/TypeResolver.java
+++ b/logic/tool/TypeResolver.java
@@ -18,7 +18,6 @@
 
 package grakn.core.logic.tool;
 
-import grakn.core.common.exception.ErrorMessage;
 import grakn.core.common.exception.GraknException;
 import grakn.core.common.parameters.Label;
 import grakn.core.concept.ConceptManager;
@@ -71,18 +70,19 @@ public class TypeResolver {
 
     public Conjunction resolveVariablesExhaustive(Conjunction conjunction) {
         ConstraintMapper constraintMapper = new ConstraintMapper(conjunction);
-        ResolverVariables resolverVariables = constraintMapper.getResolveVars();
+        Resolvers resolverVariables = constraintMapper.resolvers();
         Map<Reference, Set<Label>> referenceResolversMapping =
-                retrieveResolveTypes(new HashSet<>(resolverVariables.getResolveVars()));
+                retrieveResolveTypes(new HashSet<>(resolverVariables.resolvers()));
         long numOfTypes = traversalEng.graph().schema().stats().thingTypeCount();
 
-        final Map<Identifier, TypeVariable> resolveeVars = resolverVariables.resolveeVars;
-
+        final Map<Identifier, TypeVariable> resolveeVars = resolverVariables.typeResolvers;
 
         for (Variable variable : conjunction.variables()) {
             if (variable.reference().isLabel()) continue;
             Set<Label> resolveLabels = referenceResolversMapping.get(resolveeVars.get(variable.identifier()).reference());
-            if (resolveLabels == null) throw GraknException.of(TYPE_NOT_RESOLVABLE, variable.toString());
+            if (resolveLabels == null) {
+                throw GraknException.of(TYPE_NOT_RESOLVABLE, variable.toString());
+            }
 
             if (variable.isThing()) {
                 if (resolveLabels.size() != numOfTypes) {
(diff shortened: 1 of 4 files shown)


logic/tool/TypeResolver.java
```diff
@@ -263,64 +261,75 @@ private TypeVariable createRootTypeVar(Label rootLabel) {
...
+        }
+
+        private TypeVariable convert(TypeVariable variable) {
+            if (resolvers.contains(variable)) return resolvers.resolver(variable);
+
+            TypeVariable resolver = resolvers.register(variable);
+            copyNeighbours(variable, resolver);
+            // TODO: Why do we need `neighbours.putIfAbsent(resolver, new HashSet<>());` ?
+            //       Is this a patch for the fact that we forgot to use `to` variable in `copyNeighbours(from, to)` ?
+            neighbours.putIfAbsent(resolver, new HashSet<>());
```
haikalpribadi, line 292: We can now see more clearly what the we do with `TypeVariable` by putting the logic into its own method. However, I'm still not sure why we need `neighbours.putIfAbsent(resolver, new HashSet<>());`. Is this a patch for the fact that we forgot to use `to` variable in `copyNeighbours(from, to)` ?

logic/tool/TypeResolver.java
```diff
@@ -373,88 +382,72 @@ private void convertValue(TypeVariable owner, ValueConstraint<?> constraint) {
...
-                } else if (constraint.isOwns()) {
-                    addNeighbours(fromCopy, constraint.asOwns().attribute());
-                } else if (constraint.isPlays()) {
-                    addNeighbours(fromCopy, constraint.asPlays().role());
-                } else if (constraint.isRelates()) {
-                    addNeighbours(fromCopy, constraint.asRelates().role());
-                } else if (constraint.isIs()) {
-                    addNeighbours(fromCopy, constraint.asIs().variable());
-                }
+        private void copyNeighbours(TypeVariable from, TypeVariable to) { // TODO: why is 'to' variable never used?
```
haikalpribadi, line 395: Why is 'to' variable never used? Flipping the usage of `from` to `to` in the method body does not work either. This feels like a bug.

logic/tool/TypeResolver.java
```diff
@@ -373,88 +382,72 @@ private void convertValue(TypeVariable owner, ValueConstraint<?> constraint) {
...
+            for (TypeConstraint constraint : from.constraints()) {
+                if (constraint.isSub()) addNeighbours(from, constraint.asSub().type());
+                else if (constraint.isOwns()) addNeighbours(from, constraint.asOwns().attribute());
+                else if (constraint.isPlays()) addNeighbours(from, constraint.asPlays().role());
+                else if (constraint.isRelates()) addNeighbours(from, constraint.asRelates().role());
+                else if (constraint.isIs()) addNeighbours(from, constraint.asIs().variable());
+                // TODO: There are other constraints in a TypeVariable. Are we intentionally ignoring them?
+                //       Why do we not throw GraknException.of(ILLEGAL_STATE); ?
```
haikalpribadi, line 403: There are other constraints in a TypeVariable. Are we intentionally ignoring them? Why do we not `throw GraknException.of(ILLEGAL_STATE);` ?

logic/tool/TypeResolver.java
```diff
@@ -373,88 +382,72 @@ private void convertValue(TypeVariable owner, ValueConstraint<?> constraint) {
...
-
     }
 
-    private static class ResolverVariables {
+    private static class Resolvers {
 
-        private final Map<Identifier, TypeVariable> resolveeVars;
-        private final Map<RelationConstraint.RolePlayer, TypeVariable> rolePlayerResolveVars;
+        private final Map<Identifier, TypeVariable> typeResolvers;
+        private final Map<RelationConstraint.RolePlayer, TypeVariable> roleTypeResolvers;
```
haikalpribadi, line 411: Why is there no checkers and getters for 'roleTypeResolvers'? Do we still need 'roleTypeResolvers' given that we no longer put resolveRoleTypes on RelationConstraints?

logic/tool/TypeResolver.java
```diff
@@ -373,88 +382,72 @@ private void convertValue(TypeVariable owner, ValueConstraint<?> constraint) {
...
 
-        public TypeVariable convert(RelationConstraint.RolePlayer key) {
-            if (!rolePlayerResolveVars.containsKey(key)) {
-                TypeVariable newTypeVar = new TypeVariable(Identifier.Variable.of(
-                        new SystemReference("temp" + addAndGetCounter())));
-                rolePlayerResolveVars.put(key, newTypeVar);
-            }
-            return rolePlayerResolveVars.get(key);
+        TypeVariable register(RelationConstraint.RolePlayer rolePlayer) {
+            return roleTypeResolvers.computeIfAbsent(rolePlayer, rp -> new TypeVariable(newSystemVariable()));
```
haikalpribadi, line 424: The entire logic in the method body was equivalent to a `computeIfAbsent()` - we should simplify code as much as possible - the shorter the better, so we do not burden the reader of the code with any unnecessary complexity in interpreting the code.

logic/tool/TypeResolver.java
```diff
@@ -373,88 +382,72 @@ private void convertValue(TypeVariable owner, ValueConstraint<?> constraint) {
...
-                    newTypeVar.copyConstraints(key.asType());
-                }
-                resolveeVars.put(key.identifier(), newTypeVar);
-            }
-            return resolveeVars.get(key.identifier());
+                if (variable.reference().isAnonymous()) newTypeVar = new TypeVariable(newSystemVariable());
+                else newTypeVar = new TypeVariable(variable.identifier());
+                if (variable.isType()) newTypeVar.copyConstraints(variable.asType());
+                return newTypeVar;
+            });
```
haikalpribadi, line 434: The entire logic in the method body was equivalent to a computeIfAbsent() - we should simplify code as much as possible - the shorter the better, so we do not burden the reader of the code with any unnecessary complexity in interpreting the code.

logic/tool/TypeResolver.java
```diff
@@ -373,88 +382,72 @@ private void convertValue(TypeVariable owner, ValueConstraint<?> constraint) {
...
         }
 
-        public boolean hasConversion(Variable key) {
-            return resolveeVars.containsKey(key.identifier());
+        private Identifier.Variable newSystemVariable() {
+            return Identifier.Variable.of(SystemReference.of(sysVarCounter++));
```
haikalpribadi, line 438: You can extract out auxiliary code like the construction of new "system variables" into its own method, so you can keep the high-level logic (in the method where it's being used) to remain simple.

logic/tool/TypeResolver.java
```diff
@@ -373,88 +382,72 @@ private void convertValue(TypeVariable owner, ValueConstraint<?> constraint) {
...
         }
 
-        public TypeVariable convert(Variable key) {
-            if (!resolveeVars.containsKey(key.identifier())) {
+        TypeVariable register(Variable variable) {
+            return typeResolvers.computeIfAbsent(variable.identifier(), id -> {
                 TypeVariable newTypeVar;
-                if (key.reference().isAnonymous()) {
-                    newTypeVar = new TypeVariable(Identifier.Variable.of(
-                            new SystemReference("temp" + addAndGetCounter())));
```
haikalpribadi, line 433: The fact that you need to prefix the argument of the `SystemReference` constructor with a `"temp"` string is a big code smell. Why do we need to this? I've pushed this logic down into the constructor method body (as it's always repeating), but this feels like something you should not be doing.

pattern/variable/SystemReference.java
```diff
@@ -18,17 +18,43 @@
...
+    public static String PREFIX = "$/";
+
+    private final int id;
+    private final int hash;
+
+    public SystemReference(int id) {
+        super("sys" + id); // TODO: why are we we doing this? Why does this class have to extend Reference.Name?
+        this.id = id;
+        this.hash = Objects.hash(SystemReference.class, id);
+    }
```
haikalpribadi, line 36: Why are we doing this? Why does this class have to extend Reference.Name? Does it ever need to be referred to as a "named variable"? I don't see that usage anywhere. If not then it should not subclass `Reference.Name`, and thus it should not need to prefix its `id` with a special string, such as "sys" (this was previously "temp").

pattern/variable/SystemReference.java
```diff
@@ -18,17 +18,43 @@
...
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        SystemReference that = (SystemReference) o;
+        return this.id == that.id;
+    }
+
+    @Override
+    public int hashCode() {
+        return hash;
```
haikalpribadi, line 58: This class did not have its own `equals()` and `hashCode()` function, which means it was borrowing the methods from its parent, `Reference.Name`, which had different logic that was invalid for this class.

haikalpribadi:
@maxbaxt and @flyingsilverfin I've left a bunch of comments to address in `TypeResolver`. I'm going to merge this PR for now, but please address each comment and fix the bugs in the TypeResolve so we can enable it again.

logic/tool/TypeResolver.java
```diff
@@ -227,32 +227,30 @@ private Type getType(Label label) {
...
 
         private TypeVariable createRootTypeVar(Label rootLabel) {
-
-            Optional<TypeVariable> rootType = conjunction.variables().stream().filter(Variable::isType).map(Variable::asType)
-                    .filter(variable -> variable.label().isPresent())
-                    .filter(variable -> variable.label().get().properLabel().equals(rootLabel))
-                    .findAny();
+            Optional<TypeVariable> rootType = iterate(conjunction.variables())
+                    .filter(Variable::isType).map(Variable::asType)
+                    .filter(v -> v.label().isPresent() && v.label().get().properLabel().equals(rootLabel)).first();
```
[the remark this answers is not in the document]
maxbaxt: Good point, I will use make sure I use them going forward.

logic/tool/TypeResolver.java
```diff
@@ -263,64 +261,71 @@ private TypeVariable createRootTypeVar(Label rootLabel) {
...
 
-        private TypeVariable convertVariable(Variable variable) {
-            if (resolverVariables.hasConversion(variable)) return resolverVariables.getConversion(variable);
-            if (variable.isType()) {
-                TypeVariable asTypeVar = resolverVariables.convert(variable);
-                addNeighboursOfTypeVariable(variable.asType(), asTypeVar);
-                neighbours.putIfAbsent(asTypeVar, new HashSet<>());
-            } else convertThingVariable(variable.asThing());
-            return resolverVariables.getConversion(variable);
+        private TypeVariable convert(Variable variable) {
```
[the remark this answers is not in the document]
maxbaxt: Makes sense, agreed.

logic/tool/TypeResolver.java
maxbaxt: `neighbours.putIfAbsent(resolver, new HashSet<>());` was necessary. But it's a moot point since the new `TypeResovler` will always be exhaustive, and therefore not need to know neighbourhood distance.

logic/tool/TypeResolver.java
maxbaxt: Yes it's because we're only concerned with `Constraints` that contain other variables. Fortunately we don't have to worry about neighbours now.

logic/tool/TypeResolver.java
maxbaxt: I've been able to avoid any kind of `roleTypeResolvers` in the new implementation.

logic/tool/TypeResolver.java
maxbaxt: We don't need to. I just did this for my own convenience to remind me how it was created.

logic/tool/TypeResolver.java
maxbaxt: Nice, that's been implemented in the re-write

pattern/variable/SystemReference.java
maxbaxt: I think there's a case to make `System` a 4th type along with `Name`, `Label` and `Anonymous`. I'll briefly discuss this with Josh since we wrote that class together, but I don't think it strictly needs to be `Name`. It does however need to be Referreable.

PR #5986, opened 2021-01-04
Suppress session timeout while transaction is open

alexjpwalker:

Grakn times out inactive sessions after a short period. Clients are supposed to send pulses at regular intervals to ensure sessions do not get timed out. But if a client is using all its available cores to do work in a transaction, it may be unable to send pulses. We now mitigate that by suppressing session timeout while any Transaction is open on that Session.

- Session timeout is now suppressed while any Transaction is open on that session


server/rpc/SessionRPC.java
```diff
@@ -97,11 +97,15 @@ private void setIdleTimeout() {
...
+        if (!transactionRPCs.isEmpty()) {
+            keepAlive();
+            return;
+        }
         close();
         LOG.warn("Session with ID " + session.uuid() + " timed out due to inactivity");
     }
 
-    void keepAlive() {
+    synchronized void keepAlive() {
```
alexjpwalker, line 108: Now that `triggerIdleTimeout` has a chance to call `keepAlive`, it's possible for `keepAlive` to be called concurrently by a (client-triggered) **session pulse** and a (server-triggered) **idle check**. This could lead to concurrent access of `setIdleTimeout` which in turn could lead to scheduling two tasks when we only need one.

haikalpribadi (approved):

PR #5987, opened 2021-01-04
Update grabl automation yaml syntax

lriuui0x0:

Recently we added the support for background `monitor` script. Therefore we change the foreground job script from `script` to `command`.


lriuui0x0:
This has already been updated. close the PR.

PR #5988, opened 2021-01-04
Fix Bugs on Type Resolver

maxbaxt:

Fixed bugs on TypeResolver, including: 
* TypeResolver no longer crashes if no types were possible for a query
* TypeResolver can infer relations that from roles that are not directly related (for example when they related to a sub of the given role in a query)
* when the answers of a value comparison include both a 'double' and a 'long', both value types are possible in the resolveTypes

* resolver no longer throws an error for finding no results, instead it sets to `isSatisfiable` flag to false
* queries like `(friend: $x) isa $r` now traverse for (`$r relates $sys; $sys sub friend` in order to capture indirect relation-role connexions.
* `double` and `long` value types are now ignored. This is a stop-gap fix, a full fix will occur when `TypeResolver` is redone on the Traversal Structure
* The `isa` of a Relation Variable is no longer restricted to be `relation`, (as it was realised this excluded `thing`)
* `copyConstraints` rebranded as `cloneVariable` and is now recurse to avoid contaminating the query conjunction with the typeResolver Conjunction.

### DIFF
--- a/logic/tool/TypeResolver.java
+++ b/logic/tool/TypeResolver.java
@@ -289,10 +289,14 @@ public class TypeResolver {
 
         private TypeVariable convert(TypeVariable variable) {
             if (resolvers.contains(variable)) return resolvers.resolver(variable);
-            TypeVariable resolver = resolvers.register(variable);
-            resolver = cloneVariable(variable);
 
-//            copyNeighbours(variable, resolver);
+            TypeVariable resolver = resolvers.register(variable);
+            neighbours.putIfAbsent(resolver, new HashSet<>());
+
+            if (variable.constraints().isEmpty()) return resolver;
+
+
+            resolver = cloneVariable(variable);
             // TODO: Why do we need `neighbours.putIfAbsent(resolver, new HashSet<>());` ?
             //       Is this a patch for the fact that we forgot to use `to` variable in `copyNeighbours(from, to)` ?
             //This is correct. There is no guarantee that cloning the variable
@@ -354,7 +358,7 @@ public class TypeResolver {
         }
 
         private void addRootTypeVar(TypeVariable variable, TypeVariable rootTypeVar) {
-            TypeVariable rootConverted = resolvers.register(rootTypeVar);
+            TypeVariable rootConverted = convert(rootTypeVar);
             variable.sub(rootConverted, false);
             addNeighbours(variable, rootConverted);
         }
@@ -388,10 +392,8 @@ public class TypeResolver {
             if (constraint.isBoolean()) owner.valueType(GraqlArg.ValueType.BOOLEAN);
             else if (constraint.isString()) owner.valueType(GraqlArg.ValueType.STRING);
             else if (constraint.isDateTime()) owner.valueType(GraqlArg.ValueType.DATETIME);
-            else if (constraint.isDouble()) owner.valueType(GraqlArg.ValueType.DOUBLE);
-            else if (constraint.isLong()) owner.valueType(GraqlArg.ValueType.LONG);
             else if (constraint.isVariable()) convert(constraint.asVariable().value());
-            else throw GraknException.of(ILLEGAL_STATE);
+            else if (!(constraint.isDouble() || constraint.isLong())) throw GraknException.of(ILLEGAL_STATE);
             if (hasNoInfo(owner)) addRootTypeVar(owner, rootAttributeType);
         }
 
--- a/logic/tool/TypeResolver.java
+++ b/logic/tool/TypeResolver.java
@@ -85,7 +85,8 @@ public class TypeResolver {
                 // TODO: Can the error message be improved by saying more explicitly that the provided type did not make sense somehow?
                 // TODO: The error message is not quite accurate when querying a relation using an entity type,
                 //       e.g.: match ($x) isa person;
-                throw GraknException.of(TYPE_NOT_RESOLVABLE, variable.toString());
+//                throw GraknException.of(TYPE_NOT_RESOLVABLE, variable.toString());
+                continue;
             }
 
             if (variable.isThing()) {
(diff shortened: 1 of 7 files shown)


test/integration/logic/test-schema.gql
```diff
@@ -0,0 +1,45 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+define
```
flyingsilverfin, line 18: this is a really vague name - can we call it `test-type-resolution.gql`?

reasoner/Reasoner.java
```diff
@@ -71,7 +71,8 @@ else return buffer(disjunction.conjunctions().stream()
 
     private List<Producer<ConceptMap>> producers(Conjunction conjunction) {
         conjunction = logicMgr.typeResolver().resolveLabels(conjunction);
-        // TODO enable: conjunction = logicMgr.typeResolver().resolveVariablesExhaustive(conjunction);
+//        TODO: re-enable
```
haikalpribadi, line 74: what do we need these `TODO`s for, @maxbaxt ?

reasoner/Reasoner.java
```diff
@@ -98,7 +99,8 @@ else return buffer(disjunction.conjunctions().stream()
 
     private ResourceIterator<ConceptMap> iterator(Conjunction conjunction) {
         conjunction = logicMgr.typeResolver().resolveLabels(conjunction);
-        // TODO enable: conjunction = logicMgr.typeResolver().resolveVariablesExhaustive(conjunction);
+//        TODO: re-enable
```
haikalpribadi, line 102: and this one too?

test/behaviour/debug/bugs/TypeResolverTest.java
```diff
@@ -27,7 +27,8 @@
         strict = true,
         plugin = "pretty",
         glue = "grakn.core.test.behaviour",
-        features = "test/behaviour/debug/bugs/type-resolver-bugs.feature"
+        features = "test/behaviour/debug/bugs/type-resolver-bugs.feature",
```
haikalpribadi, line 30: Do we still need this test class at all now? We've resolved all the bugs right?

test/behaviour/debug/bugs/type-resolver-bugs.feature
```diff
@@ -53,259 +53,288 @@ Feature: Graql Match Query
...
-      """
-    Then answer size is: 2
-    When get answers of graql query
-      """
-      match
-        $x isa attribute;
-        $x < 2.0;
-      """
-    Then answer size is: 1
+#    @Ignore
```
haikalpribadi, line 56: Do we need to leave this scenario while completely commented out like this?

test/behaviour/debug/bugs/type-resolver-bugs.feature
```diff
@@ -53,259 +53,288 @@ Feature: Graql Match Query
...
 
-  Scenario: when the answers of a value comparison include both a 'double' and a 'long', both answers are returned
-    Given graql define
-      """
-      define
-      length sub attribute, value double;
-      """
-    Given transaction commits
-    Given the integrity is validated
+  Scenario: 'iid' matches the instance with the specified internal iid
```
haikalpribadi, line 321: I this scenario still broken?

logic/tool/TypeResolver.java
```diff
@@ -397,17 +410,45 @@ private void addNeighbours(TypeVariable from, TypeVariable to) {
...
+                    addNeighbours(newVar, constraint.asPlays().role());
+                    constraint.asPlays().overridden().ifPresent(typeVariable -> addNeighbours(newVar, typeVariable));
+                } else if (constraint.isRelates()) {
+                    newVar.relates(convert(constraint.asRelates().role()),
+                                   constraint.asRelates().overridden().flatMap(var -> Optional.of(convert(var))).orElse(null));
+                    addNeighbours(newVar, constraint.asRelates().role());
+                    constraint.asRelates().overridden().ifPresent(typeVariable -> addNeighbours(newVar, typeVariable));
+                } else if (constraint.isIs()) {
+                    newVar.is(convert(constraint.asIs().variable()));
+                } else throw GraknException.of(ILLEGAL_STATE);
```
haikalpribadi, line 447: I think it would be neater if we extract each if-else block into its own method?

haikalpribadi (changes requested):
Can you help cleanup the code, on places I commented to begin with, and then leave some PR comments in this PR to explain the relevant blocks of code that you've changed?

logic/tool/TypeResolver.java
maxbaxt: Agreed, I'm going to do it with the Traversal re-write. This is also make the convert logic look similar for both `ThingVariable` and `TypeVariable`, which is a plus.

test/integration/logic/BUILD
```diff
@@ -64,7 +64,7 @@ host_compatible_java_test(
         "@graknlabs_graql//java:graql",
         "@graknlabs_common//:common",
     ],
-    data = [":basic-schema.gql"],
+    data = [":basic-schema.gql", ":test-type-resolution.gql"],
```
maxbaxt, line 67: I ad-hoc added another schema, but it's better to stick to one so I will fix this in the rewrite

test/behaviour/debug/bugs/type-resolver-bugs.feature
maxbaxt: No, it's fixed. It's in matchTest so it is passes Grabl it's fine.

test/integration/logic/TypeResolverTest.java
```diff
@@ -715,4 +735,142 @@ public void multiple_anonymous_vars() throws IOException {
...
+            put("$x", set());
+        }};
+        assertEquals(thingExpected, getHintMap(thingConjunction));
+        for (Variable variable : thingConjunction.variables()) {
+            assertTrue(variable.isSatisfiable());
+        }
+    }
+
+    @Test
+    public void impossible_queries_make_variables_unsatisfiable() throws IOException {
```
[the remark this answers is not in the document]
haikalpribadi: You mean the method `Variable.isSatisfiable()`? I don't know much about that method, actually. You wrote it says git-annotation, though.

haikalpribadi (approved):

logic/tool/TypeResolver.java
haikalpribadi: Okay, sounds good!

PR #5990, opened 2021-01-05
Close transaction on exception while iterating throw answers

lriuui0x0:

Currently there's a bug such that when type resolver tries to resolve types and throws an exception, it is not caught by the top level exception handler which closes the transaction.

This is because when we do `iterate(conjunctions).flatMap(this::iterator)`, since `flatMap` is lazy, `this::iterator` will not be called when we perform the match query, but only when we're consuming the list. At which point, there's not top level handler that handles closing transaction when exception happens anymore.

- Update behaviour test so that we properly reopen the transaction after query throws.
- Move type resolving as the first thing we do when we run match query, so that we will always have the top level handler that handles the exception.
- Implemented `ErrorHandledIterator` that will close transaction on exception while iterating through answers.


dependencies/graknlabs/repositories.bzl
```diff
@@ -56,5 +56,5 @@ def graknlabs_behaviour():
     git_repository(
         name = "graknlabs_behaviour",
         remote = "https://github.com/graknlabs/behaviour",
-        commit = "6764b9ad0bbde211c4511dd0bdb08e07873197af", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_behaviour
+        commit = "7c98d085c8cb93a19c99aefb4bff0b55bef0d836", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_behaviour
```
lriuui0x0, line 59: There's a problem in the test definition that we didn't reopen the transaction when transaction throws. We have updated behaviour to fix that in https://github.com/graknlabs/behaviour/pull/128

test/behaviour/debug/bugs/async-bugs.feature
```diff
@@ -111,91 +111,6 @@ Feature: Graql Match Query
       | key:ref:5 | label:worker                       |
       | key:ref:6 | label:worker                       |
 
-  Scenario: Relations can be queried with pairings of relation and role types that are not directly related to each other
```
lriuui0x0, line 114: This test has been fixed for both sync and async situation.

reasoner/Reasoner.java
```diff
@@ -63,15 +64,17 @@ public Reasoner(TraversalEngine traversalEng, ConceptManager conceptMgr, LogicMa
     }
 
     public ResourceIterator<ConceptMap> execute(Disjunction disjunction, boolean isParallel) {
-        if (!isParallel) return iterate(disjunction.conjunctions()).flatMap(this::iterator);
-        else return buffer(disjunction.conjunctions().stream()
+        Set<Conjunction> conjunctions = disjunction.conjunctions().stream()
```
lriuui0x0, line 67: Here we move type resolving calculation at top level so that it's not nested in the `ResourceIterator::flatMap`. This is because `flatMap` is lazy, and we need to make sure when type resolving happens some one at top level will catch its exception and properly close the transaction. Therefore it cannot be nested in `flatMap`.

common/iterator/ErrorHandledIterator.java
```diff
@@ -0,0 +1,43 @@
+package grakn.core.common.iterator;
+
+import grakn.core.common.exception.GraknException;
+
+import java.util.function.Function;
+
+public class ErrorHandledIterator<T> implements ResourceIterator<T> {
```
lriuui0x0, line 7: Added `ErrorHandledIterator` so that when exception happens even when we're iterating through the list, we will still be able to close the transaction.

concept/ConceptManager.java
```diff
@@ -187,8 +187,4 @@ public GraknException exception(ErrorMessage error) {
     public GraknException exception(Exception exception) {
         return graphMgr.exception(exception);
     }
-
-    public GraknException exception(GraknException exception) {
```
lriuui0x0, line 191: Removed `exception(GraknException)` method and only keeps `exception(Exception)` so that we only need to determine if we want to convert exception to `GraknException` at the lowest level.

reasoner/Reasoner.java
lriuui0x0: Moved type resolving to here so that we can deduplicate code for both sync and async case.

common/iterator/ErrorHandledIterator.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public void recycle() {
+        try {
+            this.iterator.recycle();
+        } catch (Exception e) {
+            throw this.exceptionFn.apply(e);
+        }
+    }
```
haikalpribadi, line 42: Do you think we need this, @lriuui0x0 ? `this.iterator.recycle()` cannot throw.

dependencies/graknlabs/repositories.bzl
haikalpribadi: Let's add a test case in BDD to test that transactions are closed when match throws, then update this again.

haikalpribadi (changes requested):
Looks good, let's just update BDD first.

haikalpribadi (approved):

PR #5992, opened 2021-01-06
Do not send super type in response when super type is null

lriuui0x0:

When we ask for the super type of `thing`, the concept API returns null correctly, yet we need to do a null check to make sure we don't add the null in response which throws null pointer exception.

- null check the super type that concept API returns.


server/rpc/concept/TypeHandler.java
```diff
@@ -207,9 +207,13 @@ private void isAbstract(Transaction.Req request, Type type) {
     }
 
     private void getSupertype(Transaction.Req request, Type type) {
+        ConceptProto.Type.GetSupertype.Res.Builder getSupertypeRes = ConceptProto.Type.GetSupertype.Res.newBuilder();
+        Type superType = type.getSupertype();
+        if (superType != null) {
```
lriuui0x0, line 212: `superType` can be `null` when `type` is `thing`. So we need to do a null check here before adding to response.

haikalpribadi (approved):

PR #5993, opened 2021-01-06
Type resolver implemented directly with traversals

maxbaxt:

`TypeResolver` now creates a `Traversal` directly. This has reduced the complexity of planning the `Traversal` execution, allowing us to no longer need a separate "simple type resolver" because the "exhaustive" type resolver is fast enough, even for query lookups. Therefore there is now only `resolveLabels()` and `resolve()` that the user of the class can use.

* `TypeResolver` rewritten. We now construct a `Traversal` object based on the input `Conjunction`.
* `ResolveLabels()` is now always called by `TypeResolver` before it executes `resolveVariables`
* There is now only a `resolveVariables`, which is an exhaustive search. This is because the speed ups mean the a simpler, faster `TypeResolver` is no longer needed.


common/test/Util.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static void assertThrowsWithMessage(Runnable function, String message) {
+        try {
+            function.run();
+            fail();
+        } catch (RuntimeException e) {
+            assert (e.toString().contains(message));
+        }
+    }
+
```
maxbaxt, line 43: These two methods were in `behaviour/test`. I think they will be helpful to any rest suites if particular throws are expected.

graph/SchemaGraph.java
```diff
@@ -501,6 +501,17 @@ public long abstractTypeCount() {
...
+            Supplier<Integer> fn = () ->
+                    toIntExact(thingTypes().stream().filter(typeVertex -> !typeVertex.isAbstract()).count());
+            if (isReadOnly) {
+                if (thingTypeCount == UNSET_COUNT) thingTypeCount = fn.get();
+                return thingTypeCount;
+            } else {
+                return fn.get();
+            }
+        }
+
```
maxbaxt, line 514: Pretty self-explanatory. Opposite to `abstractThingTypeCount()`.

logic/LogicCache.java
```diff
@@ -20,28 +20,28 @@
...
+import grakn.core.traversal.Traversal;
 import graql.lang.pattern.variable.Reference;
 
 import java.util.Map;
 import java.util.Set;
 
 public class LogicCache {
 
-    private CommonCache<Conjunction, Map<Reference, Set<Label>>> typeHinterCache;
+    private CommonCache<Traversal, Map<Reference, Set<Label>>> typeResolverCache;
```
maxbaxt, line 31: Name was still referring to "hints" instead of "resolver".

reasoner/Reasoner.java
```diff
@@ -65,8 +65,7 @@ public Reasoner(TraversalEngine traversalEng, ConceptManager conceptMgr, LogicMa
 
     public ResourceIterator<ConceptMap> execute(Disjunction disjunction, boolean isParallel) {
         Set<Conjunction> conjunctions = disjunction.conjunctions().stream()
-                .map(conjunction -> logicMgr.typeResolver().resolveLabels(conjunction))
-                .map(conjunction -> logicMgr.typeResolver().resolveVariablesExhaustive(conjunction))
+                .map(conjunction -> logicMgr.typeResolver().resolveVariables(conjunction))
```
maxbaxt, line 68: `resolveVariables` now always calls `resolveLabels`. So there is no need to do both. This is because you always have to call `resolveLabels` for `resolveVariables` to run, so it shouldn't be the responsibility if the caller to ensure this.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,174 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
-        if (supLabels.isEmpty()) return;
-
-        if (subLabels.isEmpty()) {
-            Set<Label> subResolvedOfSupLabels = supLabels.stream().flatMap(label -> getType(label).getSubtypes())
-                    .map(Type::getLabel).collect(Collectors.toSet());
-            subVar.addResolvedTypes(subResolvedOfSupLabels);
-            return;
-        }
+    public Conjunction resolveVariables(Conjunction conjunction) {
+        resolveLabels(conjunction);
```
maxbaxt, line 87: We always call `resolveLabels` because it's always necessary to do that before running `resolveVariables`.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,174 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
-        //TODO: use .getType(label) once ConceptManager can handle labels
-        resolveLabels.removeIf(label -> traversalEng.graph().schema().getType(label).isAbstract());
-        variable.addResolvedTypes(resolveLabels);
+    private void findVariablesWithNoHints(Conjunction conjunction) {
+        long numOfTypes = traversalEng.graph().schema().stats().thingTypeCount();
+        long numOfConcreteTypes = traversalEng.graph().schema().stats().concreteThingTypeCount();
+        iterate(conjunction.variables()).filter(variable -> (
+                variable.isType() && variable.resolvedTypes().size() == numOfTypes ||
+                        variable.isThing() && variable.resolvedTypes().size() == numOfConcreteTypes
+        )).forEachRemaining(Variable::clearResolvedTypes);
```
maxbaxt, line 104: Here we clear `resolvedTypes` for any variables that find every single type resolvable, and so the `resolvedTypes` bear no information. `Things` have had all abstract types removed, so we don't count those when considering the number of total types.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,174 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
+        return logicCache.resolver().get(traversalConstructor.resolverTraversal(), traversal -> {
             Map<Reference, Set<Label>> mapping = new HashMap<>();
-            traversalEng.iterator(conjunction.traversal()).forEachRemaining(
+            traversalEng.iterator(traversal).forEachRemaining(
                     result -> result.forEach((ref, vertex) -> {
                         mapping.putIfAbsent(ref, new HashSet<>());
-                        mapping.get(ref).add(Label.of(vertex.asType().label(), vertex.asType().scope()));
+                        assert vertex.isType();
+                        if (!(vertex.asType().isAbstract() && traversalConstructor.getVariable(ref).isThing()))
+                            mapping.get(ref).add(Label.of(vertex.asType().label(), vertex.asType().scope()));
```
maxbaxt, line 115: We could have put the results directly into the `conjunction`. But we put it into a Map and then place the labels from the Map to the `conjunction` so we can cache the Map for later use.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,174 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
-            this.rootAttributeType = retrieveTypeFromLabel(Label.of(ATTRIBUTE.toString()));
-            this.rootRelationType = retrieveTypeFromLabel(Label.of(RELATION.toString()));
-            this.rootRoleType = retrieveTypeFromLabel(Label.of(ROLE.toString(), RELATION.toString()));
+            this.conceptMgr = conceptMgr;
+            this.resolverTraversal = new Traversal();
+            this.variableRegister = new HashMap<>();
+            this.resolverRegister = new HashMap<>();
+            this.sysVarCounter = 0;
+            this.valueTypeRegister = new HashMap<>();
             conjunction.variables().forEach(this::convert);
```
maxbaxt, line 140: This is the main loop of the code. We take each variable and convert its constraints into edges on a `Traversal`.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,174 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
 
-    private Type getType(Label label) {
-        if (label.scope().isPresent()) {
-            assert conceptMgr.getRelationType(label.scope().get()) != null;
-            return conceptMgr.getRelationType(label.scope().get()).getRelates(label.name());
-        } else {
-            return conceptMgr.getThingType(label.name());
-        }
-    }
+    private static class TraversalConstructor {
```
maxbaxt, line 122: Essentially this class will take in a conjunction and create a `Traversal` that is the schematic representation of types induced by the conjunction.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,174 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
 
-    private static class ConstraintMapper {
-        private final Resolvers resolvers;
-        private final HashMap<TypeVariable, Set<TypeVariable>> neighbours;
-        private final TypeVariable rootAttributeType;
-        private final TypeVariable rootRelationType;
-        private final TypeVariable rootRoleType;
         private final Conjunction conjunction;
+        private final Map<Reference, Variable> variableRegister;
```
maxbaxt, line 125: As we meet variables in the conjunction we register the their references. This is so we can refer to the correct variables from references after the `TraversalEngine` has been run.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,174 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
+        private final Map<Identifier, TypeVariable> resolverRegister;
```
maxbaxt, line 126: For each `Variable`, we need a corresponding `TypeVariable` that will represent its type in the `resolverTraversal`. We call these corresponding `TypeVariables` "resolvers".

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,174 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
+        private final Traversal resolverTraversal;
+        private int sysVarCounter;
         private final ConceptManager conceptMgr;
+        private final Map<Identifier, Set<ValueType>> valueTypeRegister;
```
maxbaxt, line 130: This will be explained in `convertValue`

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,174 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
-
-            if (variable.constraints().isEmpty()) return resolver;
-
-            resolver = cloneVariable(variable);
-            neighbours.putIfAbsent(resolver, new HashSet<>());
-            return resolver;
+            if (resolverRegister.containsKey(variable.id())) return resolverRegister.get(variable.id());
+            resolverRegister.put(variable.id(), variable);
+            variableRegister.putIfAbsent(variable.reference(), variable);
+            variable.addTo(resolverTraversal);
```
maxbaxt, line 161: In fact, when a `Variable` is already a `TypeVariable`, we have only to add its to the traversal in the ordinary way. That's pretty convinient!

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,174 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
         private TypeVariable convert(ThingVariable variable) {
-            if (resolvers.contains(variable)) return resolvers.resolver(variable);
-
-            TypeVariable resolver = resolvers.register(variable);
-            neighbours.putIfAbsent(resolver, new HashSet<>());
+            if (resolverRegister.containsKey(variable.id())) return resolverRegister.get(variable.id());
 
-            if (variable.constraints().isEmpty()) return resolver;
+            TypeVariable resolver = new TypeVariable(variable.reference().isAnonymous() ?
+                                                             newSystemId() : variable.id());
```
maxbaxt, line 169: When a `Variable` is anonymous, we need the corresponding resolver to be referrable. This is because we want results from the traversal so we can add `resovledTypes` even to anonymous variables.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,174 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
+            resolverRegister.put(variable.id(), resolver);
+            variableRegister.putIfAbsent(resolver.reference(), variable);
+            valueTypeRegister.putIfAbsent(resolver.id(), set());
 
-            variable.iid().ifPresent(constraint -> convertIID(resolver, constraint));
+            //Note: order is important!
```
maxbaxt, line 174: Added this comment because it's not obvious why we can't just iterate over `variable.constraints()`. The reason is that we're calling conversion on other variables we meet along the way (ie this a depth-search-first algorithm). And we have to make an assumption on `convertValue` that certain processes have occurred after calling convert(). So we have to put that one near the top.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +84,175 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
+        private void convertRelation(TypeVariable owner, RelationConstraint constraint) {
+            for (RelationConstraint.RolePlayer rolePlayer : constraint.players()) {
+                TypeVariable playerResolver = convert(rolePlayer.player());
+                TypeVariable actingRoleResolver = convert(new TypeVariable(newSystemId()));
+                if (rolePlayer.roleType().isPresent()) {
+                    TypeVariable roleTypeResolver = convert(rolePlayer.roleType().get());
+                    resolverTraversal.sub(actingRoleResolver.id(), roleTypeResolver.id(), true);
+                }
+                resolverTraversal.relates(owner.id(), actingRoleResolver.id());
+                resolverTraversal.plays(playerResolver.id(), actingRoleResolver.id());
```
maxbaxt, line 217: When converting relation, we have to consider the possibility that the role type may be a super of what is actually being played by the role player. consider, so example:

```
define 
  partnership sub relation, relates partner;
  marriage sub partnership, relates spouse as partner;
  
  person sub entity, plays marriage:spouse
```

and

```
match ($role: $p) isa marriage;
```

clearly `marriage:spouse` is one possibility for foe `$role`, but so also is `parnternship:partner`. So we need one resolver for `$role` and another to link it to the particular role that the player (in this case `$p`) is playing.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +84,175 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
-
-        Resolvers() {
-            this.sysVarCounter = 0;
-            this.typeResolvers = new HashMap<>();
-            this.roleTypeResolvers = new HashMap<>();
-        }
-
-        TypeVariable register(RelationConstraint.RolePlayer rolePlayer) {
-            return roleTypeResolvers.computeIfAbsent(rolePlayer, rp -> new TypeVariable(newSystemVariable()));
+        private Set<ValueType> findComparableValueTypes(ValueConstraint<?> constraint) {
```
maxbaxt, line 232: Turns out convertValue is tricky. Consider

```
$x = 2;
$y = $x;
```

we'd like to convert this on the resolver traversal to be:
```
$x value long;
$y value $x
```

But this isn't possible, since `valueTypeConstraint` cannot have a `Variable` as `value` (unlike `ValueConstraint`, which can).

Therefore, we have to convert it directly to:
```
$x value long;
$y value;
```

We do this by converting any `variable` we see in `value`, and then returning that as the valueType. This is why we need the `valueTypeRegister`.

But there's another problem! `$x` also can be equal to 2.0. So we need to return the valueType *comparable* not *equal* to the value type.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +84,175 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
+            if (constraint.isBoolean()) return set(ValueType.BOOLEAN);
+            else if (constraint.isString()) return set(ValueType.STRING);
+            else if (constraint.isDateTime()) return set(ValueType.DATETIME);
+            else if (constraint.isLong() || constraint.isDouble()) return set(ValueType.LONG, ValueType.DOUBLE);
+            else if (constraint.isVariable()) {
+                TypeVariable comparableVar = convert(constraint.asVariable().value());
+                assert valueTypeRegister.containsKey(comparableVar.id());
+                subAttribute(comparableVar.id());
```
maxbaxt, line 240: At the very least, if it's equal to something it must be some kind of attribute.

dependencies/graknlabs/repositories.bzl
```diff
@@ -53,8 +53,19 @@ def graknlabs_grabl_tracing():
...
+        git_repository(
+            name = "graknlabs_behaviour",
+            remote = "https://github.com/maxbaxt/behaviour",
+            commit = "8b35a55da7fa23cd59173eb5deacf66da198fd67", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_behaviour
+        )
+
+#    native.local_repository(
+#        name = "graknlabs_behaviour",
+#        path = "//Users/max/programming/behaviour",
+#    )
```
maxbaxt, line 71: Reminder to myself to change this back when PR is approved.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,173 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
-            else if (isaConstraint.type().reference().isName()) owner.is(isaConstraint.type());
+            if (!isaConstraint.isExplicit())
+                resolverTraversal.sub(owner.id(), convert(isaConstraint.type()).id(), true);
+            else if (isaConstraint.type().reference().isName())
+                resolverTraversal.equalTypes(owner.id(), convert(isaConstraint.type()).id());
             else if (isaConstraint.type().label().isPresent())
-                owner.label(isaConstraint.type().label().get().properLabel());
-            else throw GraknException.of(ILLEGAL_STATE);
-            addNeighbours(owner, isaVar);
+                resolverTraversal.labels(owner.id(), isaConstraint.type().label().get().properLabel());
```
flyingsilverfin, line 196: is this if-else if sequence correct? What happens if the user writes a non-explicit labeled isa? will that still work here?

flyingsilverfin: can also add the `else ... ILLEGAL_STATE`

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,173 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
-        Resolvers() {
-            this.sysVarCounter = 0;
-            this.typeResolvers = new HashMap<>();
-            this.roleTypeResolvers = new HashMap<>();
-        }
-
-        TypeVariable register(RelationConstraint.RolePlayer rolePlayer) {
-            return roleTypeResolvers.computeIfAbsent(rolePlayer, rp -> new TypeVariable(newSystemVariable()));
+        private Set<ValueType> findComparableValueTypes(ValueConstraint<?> constraint) {
+            if (constraint.isBoolean()) return set(ValueType.BOOLEAN);
```
flyingsilverfin, line 232: I think you can replace lots of this by getting it from: `Encoding.ValueType.of(constraint).comparables()`!

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,173 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
 
-        TypeVariable register(Variable variable) {
-            return typeResolvers.computeIfAbsent(variable.id(), id -> {
-                TypeVariable newTypeVar;
-                if (variable.reference().isAnonymous()) newTypeVar = new TypeVariable(newSystemVariable());
-                else newTypeVar = new TypeVariable(variable.id());
-                return newTypeVar;
-            });
+        private void subAttribute(Identifier.Variable variable) {
+            Identifier.Variable attributeID = Identifier.Variable.of(Reference.label("attribute"));
```
flyingsilverfin, line 245: i think we have an `attribute` constant in Graql no? try not to hardcode well known constants if they already exist elsewhere

test/integration/logic/TypeResolverTest.java
```diff
@@ -150,21 +145,14 @@ public void is_inference() throws IOException {
                 "  $p is $q;" +
                 "  $q isa mammal;";
 
-        Conjunction exhaustiveConjunction = runExhaustiveHinter(typeResolver, queryString);
-        Conjunction simpleConjunction = runSimpleHinter(typeResolver, queryString);
+        Conjunction exhaustiveConjunction = runTraversalResolver(typeResolver, queryString);
```
flyingsilverfin, line 148: can we just call this method `conjunctionResolved()`? Also the variable name is not aligned anymore right?

test/integration/logic/TypeResolverTest.java
```diff
@@ -670,14 +666,14 @@ public void simple_always_infers_its_supers() {
...
         assertEquals(expected, getHintMap(exhaustiveConjunction));
-        assertEquals(expected, getHintMap(simpleConjunction));
     }
 
     // When a hint label exists, it can "skip" a generation, meaning a hint and the hint's descendent is possible, yet
     // none of the hint's direct children are possible.
     // We show this below on the hint labels of $t
     // We also show on $a that hints can be isolated form each other completely hierarchy-wise.
     @Test
+    //TODO: ignored as the gap fpesn't appea. Unclear if this is a resolver or traversal bug.
```
flyingsilverfin, line 676: stray TODO

flyingsilverfin (changes requested):
Ok so - some minor comments, some bigger issues I've identified.
The test coverage is looking great :)
The PR description and title should address why the change is being made/what the change brings about... we're not just refactoring for fun!

common/exception/ErrorMessage.java
```diff
@@ -456,6 +456,8 @@ public static void loadConstants() {
     public static class Reasoner extends ErrorMessage {
         public static final Reasoner REVERSE_UNIFICATION_MISSING_CONCEPT =
                 new Reasoner(1, "Reverse unification failed because a concept for identifier '%s' was not found in the provided map '%s'");
+        public static final Reasoner SCHEMATICALLY_UNSATISFIABLE_CONJUNCTION =
```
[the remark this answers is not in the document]
maxbaxt: I've moved the error into 'Pattern`. However, possibly it should go into its own class of errors?

I also changed the error message close to what you've suggested.

logic/tool/TypeResolver.java
```diff
@@ -1,18 +1,18 @@
 /*
  * Copyright (C) 2020 Grakn Labs
  *
- * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU Affero General Public License as
- * published by the Free Software Foundation, either version 3 of the
- * License, or (at your option) any later version.
+ *  This program is free software: you can redistribute it and/or modify
```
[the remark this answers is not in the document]
maxbaxt: Fixed this. I'll check my Inteliji settings to see what caused this.

logic/tool/TypeResolver.java
```diff
@@ -21,42 +21,38 @@
...
 import static grakn.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
+import static grakn.core.common.exception.ErrorMessage.Reasoner.SCHEMATICALLY_UNSATISFIABLE_CONJUNCTION;
 import static grakn.core.common.exception.ErrorMessage.TypeRead.ROLE_TYPE_NOT_FOUND;
 import static grakn.core.common.exception.ErrorMessage.TypeRead.TYPE_NOT_FOUND;
 import static grakn.core.common.iterator.Iterators.iterate;
-import static graql.lang.common.GraqlToken.Type.ATTRIBUTE;
-import static graql.lang.common.GraqlToken.Type.RELATION;
-import static graql.lang.common.GraqlToken.Type.ROLE;
 
+//TODO: here we remake Type Resolver, using a Traversal Structure instead of a Pattern to move on the graph and find out answers.
```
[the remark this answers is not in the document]
maxbaxt: Removed.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,173 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
-        Set<Label> subLabels = subVar.resolvedTypes();
-        if (supLabels.isEmpty()) return;
-
-        if (subLabels.isEmpty()) {
-            Set<Label> subResolvedOfSupLabels = supLabels.stream().flatMap(label -> getType(label).getSubtypes())
-                    .map(Type::getLabel).collect(Collectors.toSet());
-            subVar.addResolvedTypes(subResolvedOfSupLabels);
-            return;
-        }
+    public Conjunction resolveVariables(Conjunction conjunction) {
```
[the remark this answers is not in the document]
maxbaxt: This is correct. The new implementation will be much quicker, so there isn't a ned to have a "simple" implmentation.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,173 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
+        return conjunction;
     }
 
-    private void addInferredIsaLabels(ThingVariable variable, Set<Label> resolveLabels) {
-        //TODO: use .getType(label) once ConceptManager can handle labels
-        resolveLabels.removeIf(label -> traversalEng.graph().schema().getType(label).isAbstract());
-        variable.addResolvedTypes(resolveLabels);
+    private void findVariablesWithNoHints(Conjunction conjunction) {
+        long numOfTypes = traversalEng.graph().schema().stats().thingTypeCount();
+        long numOfConcreteTypes = traversalEng.graph().schema().stats().concreteThingTypeCount();
```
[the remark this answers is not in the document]
maxbaxt: I think it is better to keep, as otherwise the codebase will be littered with calculations that look cleaner expressed by a single method.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,173 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
-    private void addInferredSubLabels(TypeVariable variable, Set<Label> resolvedLabels) {
-        variable.addResolvedTypes(resolvedLabels);
+        return conjunction;
     }
 
-    private void addInferredIsaLabels(ThingVariable variable, Set<Label> resolveLabels) {
-        //TODO: use .getType(label) once ConceptManager can handle labels
-        resolveLabels.removeIf(label -> traversalEng.graph().schema().getType(label).isAbstract());
-        variable.addResolvedTypes(resolveLabels);
+    private void findVariablesWithNoHints(Conjunction conjunction) {
```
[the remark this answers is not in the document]
maxbaxt: I'm deliberately using "hints" to say that the resolved types give no additional information. So the method is shorthand for:

"find out which variables have Resolve Types which give no hint their possibilities".

However, since you've written this comment it must not have been clear, so I will think of something more appropriate.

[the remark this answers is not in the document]
maxbaxt: Changed to `clearResolvedTypesWhichProvideNoInfo`.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,173 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
+        return logicCache.resolver().get(traversalConstructor.resolverTraversal(), traversal -> {
             Map<Reference, Set<Label>> mapping = new HashMap<>();
-            traversalEng.iterator(conjunction.traversal()).forEachRemaining(
+            traversalEng.iterator(traversal).forEachRemaining(
                     result -> result.forEach((ref, vertex) -> {
                         mapping.putIfAbsent(ref, new HashSet<>());
-                        mapping.get(ref).add(Label.of(vertex.asType().label(), vertex.asType().scope()));
+                        assert vertex.isType();
+                        if (!(vertex.asType().isAbstract() && traversalConstructor.getVariable(ref).isThing()))
+                            mapping.get(ref).add(Label.of(vertex.asType().label(), vertex.asType().scope()));
```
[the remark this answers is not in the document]
maxbaxt: Good point, changed that.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,173 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
+                variable.isType() && variable.resolvedTypes().size() == numOfTypes ||
+                        variable.isThing() && variable.resolvedTypes().size() == numOfConcreteTypes
+        )).forEachRemaining(Variable::clearResolvedTypes);
     }
 
-    private Map<Reference, Set<Label>> retrieveResolveTypes(Set<Variable> resolveVars) {
-        Conjunction resolveVariableConjunction = new Conjunction(resolveVars, Collections.emptySet());
-        resolveVariableConjunction = resolveLabels(resolveVariableConjunction);
-        return logicCache.resolver().get(resolveVariableConjunction, conjunction -> {
+    private Map<Reference, Set<Label>> runTraversalEngine(TraversalConstructor traversalConstructor) {
```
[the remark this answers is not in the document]
maxbaxt: I'm happy with that name change.

logic/tool/TypeResolver.java
maxbaxt: This was a very valid with the previous iteration of `TypeResolver`. But we are no longer manipulating constraints, but instead a traversal. Calling `addTo()` on a `Variable` does not change the `Variable`'s state. (I just checked this). It only changes the `Traversal`'s state.

logic/tool/TypeResolver.java
maxbaxt: I though `resovleLabels` would sort this for us. However, there may be circumstances where more information is gained from the full `resolvedTypes`. This is will only occur when the `TypeVariable` is a `role`.

When this occurs, we want to reset the `resolvedTypes` on a `Variable` instead of just adding them. So I've added a `setResolvedTypes()` method onto `Variable`, and call that instead of `addResovledTypes()` after the resolve Traversal has been called.

logic/tool/TypeResolver.java
```diff
@@ -144,355 +83,173 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
-        private void convertHas(TypeVariable owner, HasConstraint hasConstraint) {
-            TypeVariable attributeTypeVar = convert(hasConstraint.attribute());
-            owner.owns(attributeTypeVar, null, false);
-            if (hasNoInfo(attributeTypeVar)) addRootTypeVar(attributeTypeVar, rootAttributeType);
-            addNeighbours(owner, attributeTypeVar);
-            assert attributeTypeVar.sub().isPresent();
-            addNeighbours(owner, attributeTypeVar.sub().get().type());
+        private void convertIID(TypeVariable owner, IIDConstraint iidConstraint) {
+            if (conceptMgr.getThing(iidConstraint.iid()) == null)
+                throw GraknException.of(SCHEMATICALLY_UNSATISFIABLE_CONJUNCTION, conjunction);
```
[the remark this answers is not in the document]
maxbaxt: In fact there is already a THING_NOT_FOUND error for when an iid doesn't return anything from the database.

logic/tool/TypeResolver.java
maxbaxt: This is correct. If a user writes a non-explicit labelled `isa`, for example
```
$p isa person;
```

Then it becomes `$p sub person`, which is handled on the traversal engine as:
```
$p --[SUB]-> $_person;
$_person {types:[peson]}
```

logic/tool/TypeResolver.java
maxbaxt: It's a nice idea, but it ends up ugly. You cannot call `Encoding.ValueType()` with a `constraint`. You need to use `constraint.value()`. But you have to exclude the possibility that it's a Variable Value Constraint, which leaves us with something like:

```
if (!constraint.isVariable()) {
                Set<Encoding.ValueType> comparables = Encoding.ValueType.of((ValueType) constraint.value()).comparables();
                return comparables.stream().map(Encoding.ValueType::graqlValueType).collect(Collectors.toSet());
            }
```

Which I think is far less clear than:
```
            if (constraint.isBoolean()) return set(ValueType.BOOLEAN);
            else if (constraint.isString()) return set(ValueType.STRING);
            else if (constraint.isDateTime()) return set(ValueType.DATETIME);
            else if (constraint.isLong() || constraint.isDouble()) return set(ValueType.LONG, 
```

However, I do think the logic of what's comparable to what should exist elsewhere, since that's something that will be universal to the codebase, and in the future there will be a lot more combinations of value types that are comparable... so we should return to this question later.

logic/tool/TypeResolver.java
maxbaxt: I've changed `"attribute"` to `ATTRIBUTE.toString()` if that's what you meant.

test/integration/logic/TypeResolverTest.java
maxbaxt: Not a stray this time. This is the query that I discoed led to a "gap" in the resolved types. Except the gap ins't appearing! I don't think this is a pressing issue current, so I have left it as a TODO.

test/integration/logic/TypeResolverTest.java
maxbaxt: Changed the method name to `resolveConjunction()` and the variable name to just `conjunction`.

logic/tool/TypeResolver.java
maxbaxt: Agreed it should end with `ILLEGAL_STATE` catch though.

common/exception/ErrorMessage.java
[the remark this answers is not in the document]
flyingsilverfin: pattern seems better than reasoner i think

logic/tool/TypeResolver.java
flyingsilverfin: I think we should not keep the code as you've written for precisely the duplication you've mentioned, the more times we duplicate whats "comparable", the more we can shoot ourselves in the foot. I think we should trade off some readability for a more standardised comparison set :)

test/integration/logic/TypeResolverTest.java
flyingsilverfin: Oh, why is the test enabled then?

maxbaxt: It isn't, there's "@ignore" right below that line.

logic/tool/TypeResolver.java
maxbaxt: I agree to avoid duplication, but I'm no sure that `Encoding.ValueType` is in a great format for dealing with that. 

I will use it as it is for now, but I suspect we will want to refactor it as it comes again.

pattern/variable/Variable.java
```diff
@@ -79,21 +79,25 @@ public ThingVariable asThing() {
         throw GraknException.of(INVALID_CASTING, className(this.getClass()), className(ThingVariable.class));
     }
 
-    public void addResolvedType(Label label) {
+    public void addSingleResolvedType(Label label) {
```
maxbaxt, line 82: renamed because before you have to distinguish `addResolvedType` and `addResolvedTypes`, which is a little inconvenient.

flyingsilverfin (approved):

graph/SchemaGraph.java
```diff
@@ -491,11 +492,22 @@ private Statistics() {
...
             Supplier<Integer> fn = () -> toIntExact(thingTypes().stream().filter(TypeVertex::isAbstract).count());
             if (isReadOnly) {
-                if (abstractTypeCount == UNSET_COUNT) abstractTypeCount = fn.get();
-                return abstractTypeCount;
+                if (abstractThingTypeCount == UNSET_COUNT) abstractThingTypeCount = fn.get();
+                return abstractThingTypeCount;
+            } else {
+                return fn.get();
+            }
+        }
```
haikalpribadi, line 503: I see there was an inconsistency between the method name `abstractTypeCount()` and the fact that it gets the stream from `thingTypes()`. However, should the fix be in the method name or in the stream? 🤔 

If I look at the user of this method, I think the method name seems correct, and looks like the stream should have been changed?

pattern/variable/Variable.java
haikalpribadi: the pattern to distinguish `singular` vs `plural` method names is actually very common across our codebase, @maxbaxt . We don't like things too verbose especially when a shorter name does the job non-ambiguously. So let's revert this one again.

reasoner/Reasoner.java
haikalpribadi: when then you should just rename the method `.resolveVariables()` to just `.resolve()`, no?

traversal/planner/PlannerVertex.java
```diff
@@ -260,7 +260,7 @@ void updateObjective(GraphManager graph) {
             if (!props().labels().isEmpty()) {
                 setObjectiveCoefficient(props().labels().size());
             } else if (props().isAbstract()) {
-                setObjectiveCoefficient(graph.schema().stats().abstractTypeCount());
+                setObjectiveCoefficient(graph.schema().stats().abstractThingTypeCount());
```
haikalpribadi, line 263: What's the reason we want to restrict the cost to just `abstract` `thingtypes` and not `types`, here?

haikalpribadi (changes requested):
Reviewed everything except for `TypeResolver` class itself. Left some change requests in previous comments.

traversal/planner/PlannerVertex.java
maxbaxt: This was just a changing of names. the implementation remains the same.

graph/SchemaGraph.java
maxbaxt: Good question

Since it was the implementation rather than the name that was wrong, I will fix the implementation.

pattern/variable/Variable.java
haikalpribadi: Can you address this @maxbaxt ?

test/behaviour/debug/debug.feature
```diff
@@ -23,4 +23,4 @@ Feature: Debugging Space
     Given connection does not have any database
 
   # Paste any scenarios below for debugging.
-  # Do not commit any changes to this file.
+  # Do not commit any changes to this file.
```
haikalpribadi, line 26: can you put back the line here @maxbaxt

haikalpribadi (approved):

test/behaviour/debug/debug.feature
maxbaxt: Yup, fixed.

pattern/variable/Variable.java
maxbaxt: Changed it back to `addResolvedType()`

PR #5996, opened 2021-01-06
Rule materialisation

flyingsilverfin:
Rule conclusions can be materialised, using the `TraversalEngine` to perform lookups if a satisfactory conclusion already exists in the DB, otherwise we use a `ConceptManager` to insert the relevant fact.

* Implement `Rule.Conclusion.putConclusion()`
* Add tests for materialisation
* Graph `ThingEdge` now has a `isInferred` flag, similarly to `ThingVertex`


graph/DataGraph.java
```diff
@@ -170,7 +170,7 @@ public ThingVertex create(TypeVertex typeVertex, boolean isInferred) {
         final ThingVertex vertex = new ThingVertexImpl.Buffered(this, iid, isInferred);
         thingsByIID.put(iid, vertex);
         thingsByTypeIID.computeIfAbsent(typeVertex.iid(), t -> new HashSet<>()).add(vertex);
-        statistics.vertexCreated(typeVertex.iid());
+        if (!isInferred) statistics.vertexCreated(typeVertex.iid());
```
flyingsilverfin, line 173: we should only additionally count non-inferred vertices? @lriuui0x0  what do you think of this

graph/DataGraph.java
```diff
@@ -269,11 +269,15 @@ ATT_VERTEX getOrReadFromStorage(Map<ATT_IID, ATT_VERTEX> map, ATT_IID attIID, Fu
...
+                    if (!isInferred) statistics.attributeVertexCreated(v.iid());
                     return v;
                 }
         );
-        if (!isInferred && vertex.isInferred()) vertex.isInferred(false);
-        statistics.attributeVertexCreated(vertex.iid());
+        if (!isInferred && vertex.isInferred()) {
+            // promote inferred attribute to non-inferred attribute
+            vertex.isInferred(false);
+            statistics.attributeVertexCreated(vertex.iid());
```
flyingsilverfin, line 279: here, when an inferred attribute is inserted explicitly, we then count it

graph/DataGraph.java
```diff
@@ -376,7 +396,7 @@ public void delete(ThingVertex vertex) {
             if (thingsByTypeIID.containsKey(vertex.type().iid())) {
                 thingsByTypeIID.get(vertex.type().iid()).remove(vertex);
             }
-            statistics.vertexDeleted(vertex.type().iid());
+            if (!vertex.isInferred()) statistics.vertexDeleted(vertex.type().iid());
```
flyingsilverfin, line 399: similarly for decrement @lriuui0x0

graph/adjacency/impl/ThingAdjacencyImpl.java
```diff
@@ -129,47 +135,55 @@ public ThingEdge edge(Encoding.Edge.Thing encoding, ThingVertex adjacent) {
         return edge;
     }
 
-    private ThingEdgeImpl put(Encoding.Edge.Thing encoding, ThingEdgeImpl edge, IID[] infixes, boolean isModified, boolean recurse) {
+    private ThingEdgeImpl put(Encoding.Edge.Thing encoding, ThingEdgeImpl edge, IID[] infixes, boolean isModified, boolean symmetricPut) {
```
flyingsilverfin, line 138: renamed "recurse" to "symmetricPut" as "recursive" makes the user think it can go arbitrarily deep!

graph/adjacency/impl/ThingAdjacencyImpl.java
```diff
@@ -48,7 +51,7 @@
     final ThingVertex owner;
     final Encoding.Direction.Adjacency direction;
     final ConcurrentMap<InfixIID.Thing, Set<InfixIID.Thing>> infixes;
-    final ConcurrentMap<InfixIID.Thing, Set<ThingEdge>> edges;
+    final ConcurrentMap<InfixIID.Thing, Map<EdgeIID.Thing, ThingEdge>> edges;
```
flyingsilverfin, line 54: @haikalpribadi  this is a major change - I'm adding an inner index by IID, so that we can look up edges and set the `isInferred` again later - a set doesn't let us retrieve the item in the set! Note that i think in terms of storage, i think this is free because a set is a map with a `boolean` as the value - now we're actually making use of it!

graph/adjacency/impl/ThingAdjacencyImpl.java
flyingsilverfin: name this to `putReflexive`

haikalpribadi (approved):

jmsfltchr (approved):

PR #5998, opened 2021-01-07
Conclusion refactor

jmsfltchr:

Refactor a Rule to have only one conclusion, one of: relation, variable has, explicit has. This is in response to additional challenges in unification which required a re-think of the approach, with architectural ramifications.

- Rule only returns one `Conclusion`
- Change possible Conclusions to `Rule.Conclusion.Has.Explicit`, `Rule.Conclusion.Has.Variable`, `Rule.Conclusion.Relation`
- `Rule.Conclusion.Relation` can be unified with `Concludable.Relation`
- `Rule.Conclusion.Has` can be unified with `Concludable.Has`
- `Rule.Conclusion.Has.Explicit` and `Rule.Conclusion.Relation` implement `Rule.Conclusion.Isa`, so they can be unified with `Concludable.Isa`
- `Rule.Conclusion.Has.Explicit` implements `Rule.Conclusion.Value`, so it can be unified with `Concludable.Attribute`

### DIFF
--- a/logic/Rule.java
+++ b/logic/Rule.java
@@ -222,10 +222,6 @@ public class Rule {
             return false;
         }
 
-        public boolean isHas() {
-            return false;
-        }
-
         public boolean isIsa() {
             return false;
         }
@@ -235,19 +231,15 @@ public class Rule {
         }
 
         public boolean isExplicitHas() {
-            return false; // TODO Implement in subclass
+            return false;
         }
 
         public boolean isVariableHas() {
-            return false; // TODO Implement in subclass
+            return false;
         }
 
         public Relation asRelation() {
             throw GraknException.of(INVALID_CASTING, className(this.getClass()), className(Relation.class));
-        }
-
-        public Has asHas() {
-            throw GraknException.of(INVALID_CASTING, className(this.getClass()), className(Has.class));
         }
 
         public Isa asIsa() {
@@ -256,6 +248,14 @@ public class Rule {
 
         public Value asValue() {
             throw GraknException.of(INVALID_CASTING, className(this.getClass()), className(Value.class));
+        }
+
+        public Has.Variable asVariableHas() {
+            throw GraknException.of(INVALID_CASTING, className(this.getClass()), className(Has.Variable.class));
+        }
+
+        public Has.Explicit asExplicitHas() {
+            throw GraknException.of(INVALID_CASTING, className(this.getClass()), className(Has.Explicit.class));
         }
 
         public static class Relation extends Conclusion<RelationConstraint> {
@@ -312,16 +312,6 @@ public class Rule {
                 return null;
             }
 
-            @Override
-            public boolean isHas() {
-                return true;
-            }
-
-            @Override
-            public Has asHas() {
-                return this;
-            }
-
             public static class Explicit extends Has {
 
                 private final IsaConstraint isaConstraint;
@@ -336,6 +326,11 @@ public class Rule {
                 public boolean isExplicitHas() {
                     return true;
                 }
+
+                @Override
+                public Has.Explicit asExplicitHas() {
+                    return this;
+                }
             }
 
             public static class Variable extends Has {
@@ -346,6 +341,11 @@ public class Rule {
                 @Override
                 public boolean isVariableHas() {
                     return true;
+                }
+
+                @Override
+                public Variable asVariableHas() {
+                    return this;
                 }
             }
 
--- a/logic/Rule.java
+++ b/logic/Rule.java
@@ -198,20 +198,18 @@ public class Rule {
 
     public static abstract class Conclusion<CONSTRAINT extends Constraint> {
 
-        public CONSTRAINT constraint() {
-            return null; // TODO
-        }
+        public abstract CONSTRAINT constraint();
 
         public static Conclusion<?> create(Conjunction then) {
             return Iterators.iterate(then.variables()).filter(Variable::isThing).map(Variable::asThing)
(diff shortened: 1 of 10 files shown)


logic/Rule.java
```diff
@@ -203,25 +194,24 @@ private Conjunction thenPattern(ThingVariable<?> thenVariable) {
...
-            else if (constraint.isIsa()) return Isa.create(constraint.asIsa(), whenContext);
-            else if (constraint.isValue()) return Value.create(constraint.asValue(), whenContext);
-            else throw GraknException.of(ILLEGAL_STATE);
+    public static abstract class Conclusion {
+
+        public static Conclusion create(Conjunction then) {
+            return Iterators.iterate(then.variables()).filter(Variable::isThing).map(Variable::asThing)
+                    .flatMap(variable -> Iterators.iterate(variable.constraints()).map(constraint -> {
+                        if (constraint.isRelation()) {
+                            return new Relation(constraint.asRelation(), variable.isa().get());
```
flyingsilverfin, line 203: probably good to do an assert inside each if for the right constraints existing:
```
assert variable.isa().isPresent()
```

logic/Rule.java
```diff
@@ -203,25 +194,24 @@ private Conjunction thenPattern(ThingVariable<?> thenVariable) {
...
+                        } else if (constraint.isHas()) {
+                            if (constraint.asHas().attribute().isa().isPresent()) {
+                                assert constraint.asHas().attribute().value().size() == 1;
```
flyingsilverfin, line 206: just like here! And with the `isa` too

logic/Rule.java
```diff
@@ -258,27 +256,30 @@ public Value asValue() {
...
+        public interface Value {
+            ValueConstraint<?> value();
+        }
+
+        public static class Relation extends Conclusion implements Isa {
+
+            private final RelationConstraint relation;
+            private final IsaConstraint isa;
+
+            public Relation(RelationConstraint relation, IsaConstraint isa) {
```
flyingsilverfin, line 280: actually on second thought its clearer if the constructor takes the entire conjunction, and extracts the constraints in here. This way the assertions can also sit inside the constructor, which feels like a better place?

logic/resolvable/Concludable.java
```diff
@@ -244,21 +242,21 @@ public Relation(final RelationConstraint constraint, @Nullable IsaConstraint isa
...
-                if (unificationSatisfiable(constraint().owner(), relationConclusion.constraint().owner())) {
-                    unifierBuilder.add(constraint().owner().id(), relationConclusion.constraint().owner().id());
+                if (unificationSatisfiable(constraint().owner(), relationConclusion.relation().owner())) {
+                    unifierBuilder.add(constraint().owner().id(), relationConclusion.relation().owner().id());
                 } else return Iterators.empty();
             }
 
             if (constraint().owner().isa().isPresent()) {
-                assert relationConclusion.constraint().owner().isa().isPresent(); // due to known shapes of rule conclusions
+                assert relationConclusion.relation().owner().isa().isPresent(); // due to known shapes of rule conclusions
```
flyingsilverfin, line 257: i think we can get rid of this now, as the assertion on the "shape" of each conclusion will live in the conclusion constructors

pattern/Conjunction.java
```diff
@@ -125,6 +125,7 @@ public boolean equals(Object obj) {
         if (this == obj) return true;
         if (obj == null || obj.getClass() != getClass()) return false;
         final Conjunction that = (Conjunction) obj;
+        // TODO This doesn't work! It doesn't compare constraints
```
flyingsilverfin, line 128: maybe bring this up with haikal?

test/integration/logic/RuleTest.java
```diff
@@ -53,23 +54,7 @@ private long relationConcludablesCount(Set<Concludable<?>> concludables) {
     }
 
     private long valueConcludablesCount(Set<Concludable<?>> concludables) {
```
flyingsilverfin, line 56: name mismatched

test/integration/logic/resolvable/UnifyAttributeConcludableTest.java
```diff
@@ -121,28 +122,25 @@ private Conjunction parseConjunction(String query) {
...
 
-        // rule: when { $x isa person; } then { $x has first-name "john"; }
-        Conjunction whenHasName = parseConjunction("{$x isa person;}");
-        Conjunction thenHasNameJohn = parseConjunction("{ $x has first-name 'john'; }");
-        IsaConstraint thenHasNameIsa = findIsaConstraint(thenHasNameJohn);
-        Rule.Conclusion.Isa hasIsaConclusion = Rule.Conclusion.Isa.create(thenHasNameIsa, whenHasName.variables()); // TODO this will become a Has conclusion
+        Rule rule = createRule("isa-rule", "{ $x isa person; }", "$x has first-name \"john\"");
 
-        List<Unifier> unifiers = queryConcludable.unify(hasIsaConclusion, conceptMgr).toList();
+        List<Unifier> unifiers = queryConcludable.unify(rule.conclusion().asIsa(), conceptMgr).toList();
```
flyingsilverfin, line 143: yessss

flyingsilverfin (approved):

logic/resolvable/Concludable.java
jmsfltchr: Ah yes, thought this was gone.

logic/Rule.java
jmsfltchr: Resolved in conversation: now using static constructors that return optionals to move the validation logic down into the Conclusion classes.

PR #6001, opened 2021-01-07
Create namespaced Encoding.Statistics and Encoding.Index

flyingsilverfin:
We namespace `Encoding` to contain`Statistics`, and `Index`. This makes it clear which byte definitions are for `Statistics`, which assign their own non-overlapping enums with the primary byte enums; simiarly for `Index`. This enables more advanced `Index` encodings, required for indexing rules.

* create `Encoding.Statistics`, and `Encoding.Index`
* Delete `Encoding.Graph.InfixType`, as it was unused and added confusion over what its purpose really was
* Move `Encoding.Index` to `Encoding.Index.Prefix`, in order to prepare for next steps where we will create `Encoding.Index.Infix`, used for more advanced indexing patterns.
* Refactor naming in `Encoding.Statistics` to be more in line with our style


graph/iid/VertexIID.java
```diff
@@ -59,9 +59,9 @@
 
     public static VertexIID of(byte[] bytes) {
         switch (Encoding.Prefix.of(bytes[0]).type()) {
-            case TYPE:
+            case GRAPH_TYPE:
                 return VertexIID.Type.of(bytes);
-            case THING:
+            case GRAPH_THING:
```
haikalpribadi, line 64: I think you forgot to revert this back to without `GRAPH_`

graph/util/Encoding.java
```diff
@@ -109,16 +125,17 @@ public boolean isBackward() {
...
     public enum PrefixType {
         INDEX(0),
-        TYPE(1),
-        THING(2),
-        RULE(3),
-        STATISTICS(4);
+        STATISTICS(1),
+        GRAPH_TYPE(2),
+        GRAPH_THING(3),
+        GRAPH_RULE(4);
```
haikalpribadi, line 131: this should be reverted to without `GRAPH_`

graph/util/Encoding.java
```diff
@@ -285,49 +217,46 @@ public byte key() {
      *
      * The size of a prefix is 1 byte; i.e. min-value = 0 and max-value = 255.
      */
-    // TODO I think we should compress these more
+    // TODO we should compress these more
     // TODO for example, group properties 1-50, type edges 40-60, instance edges 60-80, reserved for future use 80-127
```
haikalpribadi, line 221: Do you think we should act on this now? while we can still re-release the first alpha?

graph/util/Encoding.java
```diff
@@ -425,13 +311,12 @@ public Infix infix() {
...
         STRING(40, String.class, true, true, GraqlArg.ValueType.STRING),
         DATETIME(50, LocalDateTime.class, true, true, GraqlArg.ValueType.DATETIME);
-
         public static final ZoneId TIME_ZONE_ID = ZoneId.of("Z");
         public static final Charset STRING_ENCODING = UTF_8;
         public static final int STRING_SIZE_ENCODING = Bytes.SHORT_SIZE;
         public static final int STRING_MAX_SIZE = Bytes.SHORT_UNSIGNED_MAX_VALUE;
-        public static final double DOUBLE_PRECISION = 0.0000000000000001;
 
+        public static final double DOUBLE_PRECISION = 0.0000000000000001;
```
haikalpribadi, line 319: why the movement of this line?

pattern/variable/SystemReference.java
```diff
@@ -24,7 +24,7 @@
 
 public class SystemReference extends Reference.Name {
 
-    public static String PREFIX = "$/";
+    public static String PREFIX = "$/"; // TODO replace with GraqlToken.Char.$SYS
```
haikalpribadi, line 27: Can you just replace it here anyways?

haikalpribadi (changes requested):

graph/util/Encoding.java
flyingsilverfin: fair point, will do

pattern/variable/SystemReference.java
flyingsilverfin: turns out that this has been deleted from Graql, correctly, and should only live here - i was looking at an outdated Graql repo!

pattern/variable/SystemReference.java
```diff
@@ -18,6 +18,7 @@
 
 package grakn.core.pattern.variable;
 
+import graql.lang.common.GraqlToken;
```
haikalpribadi, line 21: You add GraqlToken, but not actually use it, @flyingsilverfin ?

haikalpribadi (changes requested):

pattern/variable/SystemReference.java
flyingsilverfin: oof

haikalpribadi (approved):

PR #6003, opened 2021-01-08
Rule constructor throws Unimplemented Exception

flyingsilverfin:
To prevent users from using Rules before Reasoning is in a working state, we throw an exception. This will prevent users from having to migrate data between incompatible DB versions.

* Throw exception in `RuleStructureImpl` constructor


flyingsilverfin:
Closed as too much of a pain to balance keeping tests running/green with this requirement.

PR #6015, opened 2021-01-12
Refactor Concludables to achieve proper unification

jmsfltchr:

We refactor the structure of a _concludable_ to hold one of four specific structures. This makes the task of unification with Rule Conclusions more succinct and straightforward. This new structure also properly accommodates the unification of value comparison cases, some of which were difficult to accommodate.

- Reasoner requires the ability to copy sub-sections of a query that are to be retrieved directly from the DB, or concluded by rules. `Conjunction.Cloner` takes a set of constraints and copies them and their variables. This is used for the creation of Concludables and Retrievables.
- Refactor `Concludable` to take in prescribed constraints, recognising them in full, rather than having one "main" constraint as previously.
- Update `unify()` methods to conform to the new model of unification. 
- We now include exact equality (`=`) value constraints as requirements in the unifiers. 
- All other value constraints are not considered by unifiers and so are needed in the retrievable. We have elected that they should also be duplicated in the concludables for improved efficiency when traversing the concludable.
- Update the unification tests to not cast rule conclusions such that dispatching the unification is tested.
- Un-ignore tests that were dependent upon the completion of this refactor.


pattern/constraint/thing/IIDConstraint.java
```diff
@@ -87,4 +88,9 @@ public String toString() {
         syntax.append(Bytes.bytesToHexString(iid));
         return syntax.toString();
     }
+
+    @Override
+    protected IIDConstraint clone(ConstraintCloner cloner) {
+        return cloner.cloneVariable(owner).iid(iid);
+    }
```
jmsfltchr, line 95: This may not be favourable in terms of naming/architecture, as we already have the style below for using VariableCloner. I didn't spot that we had this for a while though as it's not called `clone`. Feels like `clone` is more to the point to me. Can't decide if a static method is better. Thoughts?
```
static HasConstraint of(ThingVariable owner, HasConstraint clone, VariableCloner cloner) {
        return new HasConstraint(owner, cloner.clone(clone.attribute()));
    }
```

logic/resolvable/Resolvable.java
```diff
@@ -38,7 +38,7 @@ public Retrievable asRetrievable() {
         throw GraknException.of(INVALID_CASTING);
     }
 
-    public Concludable<?> asConcludable() {
+    public Concludable asConcludable() {
         throw GraknException.of(INVALID_CASTING);
```
flyingsilverfin, line 42: errr I'm pretty sure `INVALID_CASTING` takes two arguments hmm

logic/resolvable/Retrievable.java
```diff
@@ -79,81 +61,56 @@ public boolean isRetrievable() {
 
     public static class Extractor {
```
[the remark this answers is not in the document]
flyingsilverfin: yaas

pattern/constraint/ConstraintCloner.java
```diff
@@ -0,0 +1,74 @@
...
+import grakn.core.pattern.variable.Variable;
+import grakn.core.traversal.common.Identifier;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import static grakn.common.collection.Collections.set;
+
+public class ConstraintCloner {
```
flyingsilverfin, line 31: looks nice!

pattern/constraint/thing/IIDConstraint.java
flyingsilverfin: It's an interesting deliniation - I think I prefer having a `clone()` method, over an overloaded `of`. However, it now needs to be made super clear this isn't a simple 'cut/copy' operation that is disconnected from any conjunction or graph. As such, I'd probably end up renaming the `ConstraintCloner` to `ConjunctionCloner`, and the the `clone` method is then really `cloneInto`?

To follow that line of reasoning, then the `ConjunctionCloner` is actually a `ConjunctionCloneBuilder`! Further, it seems to imply that `Cloner` should be an inner class of `Conjunction`? 

The API is then:
`Conjunction.Cloner.cloneExactly(Set<Constraint> subgraph)` - which aligns really well with the terminology you have in the `Retrievable` constructor. The equivalent of the `.build()` method that would normally exist on a Builder, here would return a new `conjunction` -- eg. exactly as you use in `Retrievable`!

The names and stuff aren't super there yet in the above paragraphs, but I think you get the gist of what I'm trying to convey

pattern/constraint/thing/RelationConstraint.java
```diff
@@ -149,8 +156,8 @@ public AlphaEquivalence alphaEquals(RelationConstraint that) {
         private final int hash;
 
         public RolePlayer(@Nullable TypeVariable roleType, ThingVariable player, int repetition) {
-            assert roleType == null || roleType.reference().isName() ||
-                    (roleType.label().isPresent() && roleType.label().get().scope().isPresent());
+//            assert roleType == null || roleType.reference().isName() ||
+//                    (roleType.label().isPresent() && roleType.label().get().scope().isPresent()); // TODO We have a problem here if we reach this before we've added a labelconstraint to the role type variable
```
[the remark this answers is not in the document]
flyingsilverfin: umm hmmmm hmmmm hmmm So I think it would actually be quite dangerous to remove this check, because we can rely on it to make sure we've made a _valid_ clone. However this pushes us towards option 2... in terms of tradeoff having something safer vs something somewhat ugly I'd rather take the safer option? It may well end up that we have to define a total order of the constraint cloning (eg. labels first, then value, the has, etc. etc.)

pattern/variable/ThingVariable.java
```diff
@@ -241,9 +241,10 @@ public String toString() {
         StringBuilder tail = new StringBuilder();
 
         if (reference().isName()) head.append(reference());
-        tail.append(Stream.of(relationConstraints, set(isaConstraint), hasConstraints, valueConstraints, isConstraints)
-                            .flatMap(Collection::stream).filter(Objects::nonNull).map(ThingConstraint::toString)
-                            .collect(Collectors.joining("" + COMMA + SPACE)));
+        tail.append(Stream.of(relationConstraints, isaConstraint == null ? new HashSet<IsaConstraint>() : set(isaConstraint),
+                              hasConstraints, valueConstraints, isConstraints)
+                              .flatMap(Collection::stream).filter(Objects::nonNull).map(ThingConstraint::toString)
+                              .collect(Collectors.joining("" + COMMA + SPACE)));
```
flyingsilverfin, line 247: if we're not sure we should put a TODO and mention the TODOs in an issue about toString

pattern/variable/TypeVariable.java
```diff
@@ -96,8 +96,8 @@ TypeVariable constrainType(List<graql.lang.pattern.constraint.TypeConstraint> co
         return this;
     }
 
-    void constrainClone(TypeVariable clone, VariableCloner cloner) {
-        clone.constraints().forEach(constraint -> this.constrain(TypeConstraint.of(this, constraint, cloner)));
+    void constrainClone(TypeVariable toClone, VariableCloner cloner) {
+        toClone.constraints().forEach(constraint -> this.constrain(TypeConstraint.of(this, constraint, cloner)));
```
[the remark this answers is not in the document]
flyingsilverfin: wait what is `constraintClone()` and `VariableCloner` even doing anymore?? Shouldn't it all go through the `ConstraintCloner` now? And the constraints clone themselves?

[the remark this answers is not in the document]
flyingsilverfin: (same in `TypeVariable`?)

logic/resolvable/Retrievable.java
```diff
@@ -79,81 +61,56 @@ public boolean isRetrievable() {
...
                     extractedVariables.addAll(subgraph.registeredVariables());
-                    extractedConstraints.addAll(subgraph.registeredConstraints());
+                    extractedConstraints.addAll(subgraph.registeredConstraints);
                 }
             });
-            return subgraphs.stream()
-                    .filter(subGraph -> !(subGraph.variables().size() == 1 && subGraph.variables().iterator().next().constraints().size() == 0))
-                    .map(subGraph -> new Retrievable(new Conjunction(subGraph.variables(), set()))).collect(Collectors.toSet());
+            return Iterators.iterate(subgraphs).filter(SubgraphRegistry::isValid).map(subgraph -> {
+                ConstraintCloner cloner = ConstraintCloner.cloneFromConstraints(subgraph.registeredConstraints);
```
flyingsilverfin, line 90: yaas

flyingsilverfin (changes requested):

flyingsilverfin:
overall nice and clean PR, don't forget to milestone and label!

logic/resolvable/ConcludableTest.java
```diff
@@ -193,63 +194,63 @@ public void test_conjunction_creates_multiple_value_constraints_for_same_attribu
...
-                .get().asRelation().constraint();
+                .get().asRelation().relation();
         IsaConstraint relationIsaConstraint = relationConstraint.owner().isa().get();
         assertEquals("$_diagnosis", relationIsaConstraint.type().reference().asLabel().syntax());
         assertEquals("diagnosis", relationIsaConstraint.type().label().get().label());
-        assertEquals("person", list(relationConstraint.players()).get(0).player().isa().get().type().label().get().label());
+        assertFalse(list(relationConstraint.players()).get(0).player().isa().isPresent());
     }
 
+    //TODO This test only worked in debug
```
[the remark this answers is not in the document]
jmsfltchr: Mistakenly left in - the only test that would produce the bug with the assertion on the RelationConstraint structure

logic/resolvable/Concludable.java
```diff
@@ -440,30 +475,44 @@ public Has asHas() {
...
-            if (isaConstraint != null) c.add(isaConstraint);
-            return c;
-        }
     }
 
-    public static class Isa extends Concludable<IsaConstraint> {
+    public static class Isa extends Concludable {
+
+        private final IsaConstraint isa;
+        private final Set<ValueConstraint<?>> values;
```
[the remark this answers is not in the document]
jmsfltchr: In our call in December we hashed it all out and we have in our sheet that IsaConcludables will accept a set of values.

Alright I suggest we have a call to discuss this.

To your last point, yes we agreed that `$x > $y` should be in a Retrievable. In which case, it seems like we would like to split `ValueConstraint` into `ValueConstraint.Constant` and `ValueConstraint.Variable`.

logic/resolvable/Concludable.java
```diff
@@ -489,21 +538,37 @@ public Isa asIsa() {
 
         @Override
         AlphaEquivalence alphaEquals(Isa that) {
-            return constraint().alphaEquals(that.constraint());
+            return isa().alphaEquals(that.isa());
         }
     }
 
-    public static class Attribute extends Concludable<ValueConstraint<?>> {
+    public static class Attribute extends Concludable {
```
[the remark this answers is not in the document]
jmsfltchr: Actually I believe this handles any case where the user doesn't use an `isa` and refers to an attribute value in isolation e.g. `$x > 5;`. Yes I agree, comments will be valuable here.

logic/resolvable/Resolvable.java
jmsfltchr: Oof, caught me

jmsfltchr:
Conclusions from meeting with @flyingsilverfin : 

We confirmed that `Concludable.Isa` should accept a `Set<ValueConstraint<?>>` as it does now, we need(/may need) to update the unification to handle this.

Isa and Attribute Concludables could receive ValueConstraints such as `$a > 5;`. We need the Concludables to hold these constraints since when we directly traverse the Concludable having these constraints will be much more efficient. Conversely, we don't want to implement logic in the unifier to filter results based on value comparisons (@flyingsilverfin looked at this, and it's complex work that duplicates work done in the traversal engine). The best solution is rule rewriting, which we defer until later. In the meantime these constraints need to be concluded in both the Concludable and in a Retrievable.

We need to expose certain variables from concludables as those that it "generates". These variables will be necessary to determine the dependency ordering of Concludables and Retrievables. Good terminology is to discriminate between retrieved vs generated variables. Note that not all variables in a Concludable are generated. In particular we considered the case where there is a Retrievable of  `{ $x is $y; }`. This needs to be dependent upon both `$x` and `$y` being generated elsewhere, and therefore needs to be executed last in the reasoner's plan.

`Concludable.Attribute` should accept a `Set<ValueConstraint<?>>`.
We could get rid of `Concludable.Attribute` by making `Concludable.Isa` have a Nullable IsaConstraint.

We need to update the tests to give the correct number of concludables we now expect, and therefore also the method that determines which Concludables are built.

logic/resolvable/Unifier.java
```diff
@@ -196,17 +198,17 @@ Note that in the future we may treat these (and variable equality constraints en
     public static class Requirements {
         Map<Identifier, Set<Label>> types;
         Map<Identifier, Set<Label>> isaExplicit;
-        Map<Identifier, Function<Attribute, Boolean>> predicates;
+        Pair<Identifier, Function<Attribute, Boolean>> predicate;
```
jmsfltchr, line 201: Went with a Pair so that both the id and the function are Nullable together

logic/tool/ConstraintCopier.java
```diff
@@ -1,129 +0,0 @@
...
-import grakn.core.pattern.variable.TypeVariable;
-
-import java.util.LinkedHashSet;
-import java.util.Set;
-import java.util.stream.Collectors;
-
-import static grakn.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
-import static grakn.core.common.iterator.Iterators.iterate;
-
-public class ConstraintCopier {
```
jmsfltchr, line 36: Moved to inside Conjunction

pattern/constraint/thing/HasConstraint.java
```diff
@@ -93,9 +94,11 @@ public String toString() {
         if (attribute.reference().isName()) {
             syntax.append(attribute.reference().toString());
         } else {
-            syntax.append(attribute.isa().get().type().label().get().label());
+            if (attribute.isa().isPresent() && attribute.isa().get().type().label().isPresent()) {
+                syntax.append(attribute.isa().get().type().label().get().label());
+            }
             syntax.append(SPACE);
-            syntax.append(attribute.value().iterator().next());
+            attribute.value().forEach(value -> syntax.append(value.toString()));
```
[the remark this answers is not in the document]
jmsfltchr: Sure

pattern/variable/TypeVariable.java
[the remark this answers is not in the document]
jmsfltchr: For the record here, these are for the VariableCloner, which will no longer need when we switch Conjunctions to be constraint-centric.

test/integration/logic/resolvable/UnifyAttributeConcludableTest.java
```diff
@@ -129,16 +130,15 @@ private Rule createRule(String label, String whenConjunctionPattern, String then
         }
     }
 
-    @Ignore // TODO enable when we have the final structure of conclusions and concludable in place
     @Test
     public void literal_predicates_unify_and_filter_answers() {
-        String conjunction = "{ $a > 'b'; $a < 'y'; }";
-        Set<Concludable<?>> concludables = Concludable.create(parseConjunction(conjunction));
+        String conjunction = "{ $a = 'john'; }";
```
jmsfltchr, line 135: Scrutinise this! Changed given our conclusion on how we should handle ValueConstraints other than `EQ`

test/integration/logic/resolvable/UnifyAttributeConcludableTest.java
```diff
@@ -150,65 +150,53 @@ public void literal_predicates_unify_and_filter_answers() {
         // test requirements
         assertEquals(0, unifier.requirements().types().size());
         assertEquals(0, unifier.requirements().isaExplicit().size());
-        assertEquals(2, unifier.requirements().predicates().size());
+        // TODO Assert that there are 2 predicates in the unifier requirements;
```
jmsfltchr, line 153: This assertion has to stay in as a TODO, there's no valid way to write it now.

test/integration/logic/resolvable/UnifyAttributeConcludableTest.java
```diff
@@ -150,65 +150,53 @@ public void literal_predicates_unify_and_filter_answers() {
...
         Concludable.Attribute queryConcludable = concludables.iterator().next().asAttribute();
-
-        /*
-        TODO build attribute concludables (2) out of the conjunction
-        Each should successfully unify with rules that can conclude new concepts
-        and nothing else
-        */
+        // Get the name of the attribute variable from whichever concludable is chosen
+        String var = Iterators.iterate(concludables).map(Concludable::conjunction).flatMap(
+                c -> Iterators.iterate(c.variables())).filter(Variable::isThing).map(v -> v.reference().syntax()).first().get();
```
jmsfltchr, line 179: Wanted to make sure this test doesn't turn flakey if we happen to get the concludable that concerns `$x` rather than `$y`. Lmk if you have better ideas here.

logic/resolvable/Concludable.java
```diff
@@ -110,31 +102,17 @@ private void computeApplicableRules(ConceptManager conceptMgr, LogicManager logi
...
-        else throw GraknException.of(ILLEGAL_STATE);
-    }
-
-    ResourceIterator<Unifier> unify(Rule.Conclusion.Relation relationConclusion, ConceptManager conceptMgr) { return Iterators.empty(); }
-
-    ResourceIterator<Unifier> unify(Rule.Conclusion.Has hasConclusion, ConceptManager conceptMgr) { return Iterators.empty(); }
-
-    ResourceIterator<Unifier> unify(Rule.Conclusion.Isa isaConclusion, ConceptManager conceptMgr) { return Iterators.empty(); }
-
-    ResourceIterator<Unifier> unify(Rule.Conclusion.Value valueConclusion, ConceptManager conceptMgr) { return Iterators.empty(); }
```
jmsfltchr, line 127: Look at this disappearing into the children! :)

logic/resolvable/Concludable.java
```diff
@@ -233,27 +206,127 @@ boolean unificationSatisfiable(ThingVariable concludableThingVar, ThingVariable
         }
     }
 
-    public static class Relation extends Concludable<RelationConstraint> {
+    private static Set<ValueConstraint<?>> equalsConstraints(Set<ValueConstraint<?>> values) {
+        return Iterators.iterate(values).filter(v -> v.predicate().equals(EQ)).toSet();
+    }
+
+    private static Function<grakn.core.concept.thing.Attribute, Boolean> valueEqualsFunction(ValueConstraint<?> value) {
```
jmsfltchr, line 213: This handles equal value equality checking, which we now need not only in Has but also in Isa and Attribute, so it's moved up to sit in the parent class `Concludable`

logic/resolvable/Concludable.java
```diff
@@ -233,27 +206,127 @@ boolean unificationSatisfiable(ThingVariable concludableThingVar, ThingVariable
...
+        @Override
+        public Set<Constraint> concludableConstraints() {
+            return isa == null ? set(new HashSet<>(labels), relation) : set(new HashSet<>(labels), relation, isa);
         }
 
         @Override
+        ResourceIterator<Unifier> unify(Rule.Conclusion conclusion, ConceptManager conceptMgr) {
+            if (conclusion.isRelation()) return unify(conclusion.asRelation(), conceptMgr);
+            return Iterators.empty();
+        }
```
jmsfltchr, line 314: This is now the entry point for unification. We could even skip this and go directly to the method below and do the casting there - thoughts? I quite like how obvious this check is right now though.

logic/resolvable/Concludable.java
```diff
@@ -334,24 +407,73 @@ public Relation asRelation() {
...
+            Set<Constraint> constraints = new HashSet<>();
+            constraints.add(has);
+            if (isa != null) constraints.add(isa);
+            constraints.addAll(equalsConstraints(values));
+            return constraints;
+        }
+
+        @Override
+        ResourceIterator<Unifier> unify(Rule.Conclusion conclusion, ConceptManager conceptMgr) {
+            if (conclusion.isHas()) return unify(conclusion.asHas(), conceptMgr);
```
jmsfltchr, line 466: If it's an explicit has or variable has, both are a has, so cast to that and unify.

logic/resolvable/Concludable.java
```diff
@@ -433,30 +509,61 @@ public Has asHas() {
...
+            Set<Constraint> constraints = new HashSet<>();
+            constraints.add(isa);
+            constraints.addAll(equalsConstraints(values));
+            return constraints;
+        }
+
+        @Override
+        ResourceIterator<Unifier> unify(Rule.Conclusion conclusion, ConceptManager conceptMgr) {
+            if (conclusion.isIsa()) return unify(conclusion.asIsa(), conceptMgr);
+            return Iterators.empty();
```
jmsfltchr, line 556: Any rule conclusion that can infer an `isa` implements the `isa` interface: `Relation`, `ExplicitHas`, `Isa`, so here it's this simple.

flyingsilverfin (changes requested):

logic/resolvable/Concludable.java
```diff
@@ -334,24 +407,73 @@ public Relation asRelation() {
...
+     * `{ $x has age 30; }`,
+     * `{ $x has $a; $a isa age; }`,
+     * `{ $x has $a; $a 30 isa age; }`,
+     * `{ $x has $a; $a 30; }`
+     */
+    public static class Has extends Concludable {
+
+        private final HasConstraint has;
+        private final IsaConstraint isa;
+        private final Set<ValueConstraint<?>> values;
```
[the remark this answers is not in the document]
jmsfltchr: I've added notes, but it's actually even more general than that because of the `like` and `contains` constraints.

logic/resolvable/Concludable.java
```diff
@@ -433,30 +509,61 @@ public Has asHas() {
 
         @Override
         AlphaEquivalence alphaEquals(Has that) {
-            return constraint().alphaEquals(that.constraint());
+            return has().alphaEquals(that.has()).validIfAlphaEqual(isa, that.isa).validIfAlphaEqual(values, that.values);
```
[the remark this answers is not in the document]
jmsfltchr: Ah you're right, but maybe we should do it this way now that the structure of concludables is explicit.

logic/resolvable/Concludable.java
```diff
@@ -433,30 +509,61 @@ public Has asHas() {
...
+        }
+
+        @Override
+        ResourceIterator<Unifier> unify(Rule.Conclusion conclusion, ConceptManager conceptMgr) {
+            if (conclusion.isIsa()) return unify(conclusion.asIsa(), conceptMgr);
+            return Iterators.empty();
+        }
+
         ResourceIterator<Unifier> unify(Rule.Conclusion.Isa isaConclusion, ConceptManager conceptMgr) {
+            // TODO This needs to handle the ValueConstraint unification
```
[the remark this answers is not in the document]
jmsfltchr: Yes, just a trailing TODO, removing

pattern/Conjunction.java
```diff
@@ -134,4 +142,68 @@ public boolean equals(Object obj) {
...
+            return cloneExactly(ordered);
+        }
+
+        public static Cloner cloneExactly(Constraint constraint) {
+            final LinkedHashSet<Constraint> orderedSet = new LinkedHashSet<>();
+            orderedSet.add(constraint);
+            return cloneExactly(orderedSet);
+        }
+
+        private static Cloner cloneExactly(LinkedHashSet<? extends Constraint> constraints) {
```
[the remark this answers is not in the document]
jmsfltchr: So that we can guarantee that `LabelConstraint`s are cloned first, since these determine the validity of a `RelationConstraint`.

pattern/constraint/type/AbstractConstraint.java
```diff
@@ -70,4 +71,9 @@ public int hashCode() {
 
     @Override
     public String toString() { return ABSTRACT.toString(); }
+
+    @Override
+    public AbstractConstraint clone(Conjunction.Cloner cloner) {
+        return cloner.cloneVariable(owner).setAbstract();
```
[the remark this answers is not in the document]
jmsfltchr: We would have that but it clashes with Java's `abstract`!

flyingsilverfin (approved):

PR #6017, opened 2021-01-12
Fix stalling when closing transactions in parallel

alexjpwalker:

Intermittently, clients would stall when attempting to close many transactions in parallel. Transactions were stored in a non-threadsafe collection (a `HashSet`) on the server, which was being modified concurrently by multiple threads. Sometimes, this caused iteration over the collection to not return all the elements, which meant the missed Transactions would remain open indefinitely, causing the client to stall.

We changed the server implementation to use a `ConcurrentHashMap` to remedy the issue.

- Fix stalling when closing transactions in parallel

### DIFF
--- a/server/rpc/SessionRPC.java
+++ b/server/rpc/SessionRPC.java
@@ -24,8 +24,7 @@ import grakn.protocol.TransactionProto;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.HashSet;
-import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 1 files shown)


server/rpc/SessionRPC.java
```diff
@@ -40,23 +39,23 @@
 
     private final Grakn.Session session;
     private final GraknRPCService graknRPCService;
-    private final Set<TransactionRPC> transactionRPCs;
+    private final ConcurrentHashMap<Integer, TransactionRPC> transactionRPCs;
```
alexjpwalker, line 42: Intermittently, clients would stall when attempting to close many transactions in parallel. Transactions were stored in a non-threadsafe collection (a `HashSet`) on the server, which was being modified concurrently by multiple threads. Sometimes, this caused iteration over the collection to not return all the elements, which meant the missed Transactions would remain open indefinitely, causing the client to stall.

We changed the server implementation to use a `ConcurrentHashMap` to remedy the issue.

haikalpribadi (approved):

Issue #6008, opened 2021-01-10
StopIteration Exception when iterating over a concept map 


tomassabat:
## Description

I am printing all the instances of a type in Python. After a while Grakn hangs and it throws up an `StopIteration` Exception.

## Environment

1. OS: Mac OS 10.14
2. Grakn version: Grakn Core 2.0.0-alpha
3. Grakn client-python version: client-python 2.0.0-alpha-4
4. Python version: 3.9

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Run this grakn distribution: https://www.dropbox.com/s/qtoy4igwsw5of3p/grakn-2.0-alpha-tomas-20210110.zip?dl=0
2. In a terminal run python, and run: 
```
from grakn.client import GraknClient, SessionType, TransactionType
client = GraknClient("localhost:1729")
session = client.session("biograkn_covid_19", SessionType.DATA)
tx = session.transaction(TransactionType.READ)
answer_iterator = tx.query().match("match $x isa gene, has entrez-id $gs; get $gs;")
n = 0
for answer in answer_iterator:
	print(answer.get("gs"))
	n = n + 1
```
4. After a while Grakn while will throw this error: 
```
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/grakn/rpc/transaction.py", line 136, in _fetch
    response = next(self._response_iterator)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/grpc/_channel.py", line 416, in __next__
    return self._next()
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/grpc/_channel.py", line 801, in _next
    raise StopIteration()
StopIteration

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/grakn/rpc/stream.py", line 57, in __next__
    return next(self)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/grakn/rpc/stream.py", line 50, in __next__
    res = self._transaction._fetch(self._request_id)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/grakn/rpc/transaction.py", line 145, in _fetch
[...]

PR #6019, opened 2021-01-12
Do not start new async producer job when not necessary

lriuui0x0:

We appended a new async producer job even when the iterator has already exhausted, which caused a long tail of producer jobs that does nothing. We fixed this by not shutting down the current producer job whenever we know we need to produce more answers from the current iterator.

This PR fixes https://github.com/graknlabs/grakn/issues/6008

- Added a `jobs` map that represents the number of answers needed to be produced for an iterator.
- Keep processing `jobs` without shutting down the async producer thread, until `jobs` become 0.


traversal/producer/GraphProducer.java
```diff
@@ -55,40 +57,48 @@ public GraphProducer(GraphManager graphMgr, GraphProcedure procedure, Traversal.
...
+    private void produceAsync(Queue<VertexMap> queue, ResourceIterator<VertexMap> iterator) {
         try {
-            int i = 0;
-            for (; i < count && iterator.hasNext(); i++) queue.put(iterator.next());
-            if (!iterator.hasNext()) {
-                finish(iterator);
-                int toCompensate = count - i;
+            int toConsume = 0;
+            int consumed = 0;
+            while (iterator.hasNext()) {
```
lriuui0x0, line 93: Here we have an outer loop that keeps checking if there're more works that needs to be done for the current iterator. If there is, then we continue processing instead of shutting down the thread.

traversal/producer/GraphProducer.java
```diff
@@ -55,40 +57,48 @@ public GraphProducer(GraphManager graphMgr, GraphProcedure procedure, Traversal.
...
+                toConsume = take(iterator);
+                for (consumed = 0; consumed < toConsume && iterator.hasNext(); consumed++)
+                    queue.put(iterator.next());
+            }
+            synchronized (this) {
+                futures.remove(iterator);
+                int toCompensate = toConsume - consumed + take(iterator);
+                if (!iterator.hasNext()) jobs.remove(iterator);
                 if (toCompensate > 0) produce(queue, toCompensate);
```
lriuui0x0: It's possible that we need to compensate even when the iterator hasn't finished. Since other thread might add more jobs after the current thread thinks there're no more jobs.

traversal/producer/GraphProducer.java
```diff
@@ -55,41 +57,55 @@ public GraphProducer(GraphManager graphMgr, GraphProcedure procedure, Traversal.
...
+            int consumed = 0;
+            while (iterator.hasNext()) {
+                toConsume = take(iterator);
+                for (consumed = 0; consumed < toConsume && iterator.hasNext(); consumed++)
+                    queue.put(iterator.next());
+            }
+            synchronized (this) {
+                futures.remove(iterator);
+                if (iterator.hasNext()) {
+                    int toCompensate = take(iterator);
```
lriuui0x0, line 101: It's still possible that some other thread put jobs for this iterator after the iterator thinks there's no more job hence decides to stop. Therefore we need to restart this iterator.

traversal/producer/GraphProducer.java
```diff
@@ -55,41 +57,55 @@ public GraphProducer(GraphManager graphMgr, GraphProcedure procedure, Traversal.
...
-            for (ResourceIterator<VertexMap> iterator : futures.keySet()) {
-                futures.computeIfPresent(iterator, (k, v) -> v.thenRunAsync(
-                        () -> produceAsync(queue, k, splitCount), forkJoinPool()
-                ));
+        if (jobs.size() == 0) done(queue);
+        else {
+            int splitCount = (int) Math.ceil((double) count / jobs.size());
+            for (ResourceIterator<VertexMap> iterator : jobs.keySet()) {
+                jobs.computeIfPresent(iterator, (k, v) -> v + splitCount);
+                if (!futures.containsKey(iterator))
```
lriuui0x0, line 77: Start a thread for the job if the job isn't started yet.

traversal/producer/GraphProducer.java
```diff
@@ -44,6 +45,7 @@
     private final ConcurrentHashMap.KeySetView<VertexMap, Boolean> produced;
     private final AtomicBoolean isDone;
     private final Map<ResourceIterator<VertexMap>, CompletableFuture<Void>> futures;
+    private final Map<ResourceIterator<VertexMap>, Integer> jobs;
```
lriuui0x0, line 48: `jobs` keeps the number of answers that the iterator needs to return.

haikalpribadi (approved):

PR #6024, opened 2021-01-13
In BDD, split Match step into Aggregate, Group etc

alexjpwalker:

The `get answers of graql query` step was problematic for Clients Python and NodeJS, because they don't have Graql parsing libraries - meaning they can't reliably figure out which query type the raw String corresponds to (from the feature file) To mitigate the issue, we split the step up into multiple steps: `get answers of graql match aggregate`, `get answers of graql match group`... and so on, and implemented the new steps in Grakn Core.

- Split 'get answers of graql query' step into match, aggregate, group, group aggregate


haikalpribadi (approved):
Looks good. Just need to update the dependency to upstream repo.

PR #6025, opened 2021-01-13
Refactor of Concludable Relation Unifier

maxbaxt:

* Reimplement the `Unify()` code on `Relations` now that all `RolePlayers` in a `Relation` can be distinguished using the `repetition` field. The previous implementation was reliant on indexing the `RolePlayers` because `RolePlayers` with the same `RoleType` and `Player` were equal.

* Refactoring of `Unify()` algorithm. It no longer needs to use indexing as uniqueness of `RolePlayers` is now guaranteed.
* Some minor cleanup elsewhere


logic/resolvable/Concludable.java
```diff
@@ -268,67 +267,61 @@ public Relation(final RelationConstraint constraint, @Nullable IsaConstraint isa
...
+
+            return unifierBuilder.build();
         }
 
         private boolean unificationSatisfiable(RolePlayer concludableRolePlayer, RolePlayer conclusionRolePlayer, ConceptManager conceptMgr) {
             assert conclusionRolePlayer.roleType().isPresent();
             boolean satisfiable = true;
             if (concludableRolePlayer.roleType().isPresent()) {
-                satisfiable &= unificationSatisfiable(concludableRolePlayer.roleType().get(), conclusionRolePlayer.roleType().get(), conceptMgr);
+                satisfiable = unificationSatisfiable(concludableRolePlayer.roleType().get(), conclusionRolePlayer.roleType().get(), conceptMgr);
```
maxbaxt, line 319: The `&=` wasn't needed since `satifisable` was already true at this point

logic/resolvable/Concludable.java
```diff
@@ -209,7 +208,7 @@ boolean unificationSatisfiable(TypeVariable concludableTypeVar, TypeVariable con
     boolean unificationSatisfiable(ThingVariable concludableThingVar, ThingVariable conclusionThingVar) {
         boolean satisfiable = true;
         if (!concludableThingVar.resolvedTypes().isEmpty() && !conclusionThingVar.resolvedTypes().isEmpty()) {
-            satisfiable &= Collections.disjoint(concludableThingVar.resolvedTypes(), conclusionThingVar.resolvedTypes());
+            satisfiable = Collections.disjoint(concludableThingVar.resolvedTypes(), conclusionThingVar.resolvedTypes());
```
maxbaxt, line 211: The &= wasn't needed since satifisable was already true at this point

logic/resolvable/Concludable.java
```diff
@@ -268,67 +267,61 @@ public Relation(final RelationConstraint constraint, @Nullable IsaConstraint isa
                 } else return Iterators.empty();
             }
 
-            // TODO this will work for now, but we should rewrite using role player `repetition`
             List<RolePlayer> conjRolePlayers = list(constraint().players());
```
maxbaxt, line 270: Since we are recursing by "peeling" off the Conjunction Role Players, we still them in a `List` structure.

logic/resolvable/Concludable.java
```diff
@@ -268,67 +267,61 @@ public Relation(final RelationConstraint constraint, @Nullable IsaConstraint isa
...
-                        clone.get(conjRP).add(thenIdx);
+                        clone.get(conjRP).add(thenRP);
                         return clone;
-                    }).flatMap(newMapping -> matchRolePlayerIndices(conjRolePlayers.subList(1, conjRolePlayers.size()),
-                                                                    thenRolePlayers, newMapping, conceptMgr));
+                    }).flatMap(newMapping -> matchRolePlayers(conjRolePLayers.subList(1, conjRolePLayers.size()),
+                                                              thenRolePlayers, newMapping, conceptMgr));
+        }
+
+        private Unifier convertRPMappingToUnifier(
```
maxbaxt, line 294: We still need to create a mapping from `RolePlayer` to `RolePlayer`, and then turn that into a unification. I looked into doing this all in a single loop, but there is no way to know if a Then RolePlayer has been seen before. I would have to change the `Unification.Builder` object to do that. The current solution therefore seems more intuitive.

logic/resolvable/Concludable.java
```diff
@@ -268,67 +267,61 @@ public Relation(final RelationConstraint constraint, @Nullable IsaConstraint isa
...
+            mapping.forEach((conjRP, thenRPs) -> thenRPs.forEach(thenRP -> {
+                unifierBuilder.add(conjRP.player().id(), thenRP.player().id());
+                if (conjRP.roleType().isPresent()) {
+                    assert thenRP.roleType().isPresent();
+                    TypeVariable conjRoleType = conjRP.roleType().get();
+                    TypeVariable thenRoleType = thenRP.roleType().get();
+                    unifierBuilder.add(conjRoleType.id(), thenRoleType.id());
+                    if (conjRoleType.reference().isLabel()) {
+                        Set<Label> allowedTypes = conjRoleType.resolvedTypes().stream()
+                                .flatMap(roleLabel -> subtypeLabels(roleLabel, conceptMgr))
```
maxbaxt, line 305: Kept as `.stream()` because subtypeRoleLabels returns a `Stream`. I suspect this all needs to be refactored to use Grakn `Iterator`, but for now this I'll keep using `.stream()`.

flyingsilverfin (approved):

flyingsilverfin:
I think we can do a little bit better with the PR title and goal:
`Unification Refactor` could be _anything_ - unification is a pretty broad domain/code
mentioning `repetition` field is good in the goal, but you're missing the why - what was not right with the previous approach?

PR #6035, opened 2021-01-15
Rule index: fast rule lookup and prevent deleting types in rules

flyingsilverfin:
Create a Rule Index, which indexes Types to Rules they are either 1) mentioned in, or 2) they they may be concluded by.  1) allows us to prevent undefining types that are utilised in rules. 2) allows us to find rules that are relevant to a reasoning query, without relying on scanning _every_ rule.

* Introduce `SchemaGraph.Rules`, which contains `SchemaGraph.Rules.Concludes` and `SchemaGraph.Rules.References`
* Introduce `Encoding.Index.Infix` to store indexed lists (eg. Type -> [Rule1, Rule2] gets stored as Type_Infix_Rule1; Type_Infix_Rule2)
* Allow `Concludable` to reach the rule index via `LogicManager`
* Enable `TypeResolver` in rules, as we depend on it to correctly index concluded types -- this led to finding rules in our tests that were unsatisfiable, all of which have been fixed
* Introduce tests to ensure we are not allowed to undefine a type used in rules, and that this index is maintained through a define-undefine cycle
* Introduce tests to ensure we create, update, and delete rule conclusion indices
* Include a safe `byte signedByte(int)` and `byte unsignedByte(int)` in `Bytes` to check and possibly throw before casting an integer to byte


concept/type/impl/RelationTypeImpl.java
```diff
@@ -225,11 +225,7 @@ public RoleTypeImpl getRelatesExplicit(String roleLabel) {
 
     @Override
     public void delete() {
-        if (getSubtypes().anyMatch(s -> !s.equals(this)))
-            throw exception(GraknException.of(TYPE_HAS_SUBTYPES, getLabel()));
-        else if (getInstances().findFirst().isPresent())
-            throw exception(GraknException.of(TYPE_HAS_INSTANCES, getLabel()));
-
+        validateDelete();
```
flyingsilverfin, line 228: see next comments in other concept types

graph/structure/RuleStructure.java
```diff
@@ -51,5 +52,13 @@
...
      */
     void commit();
+
+    void createConcludingIsaIndex(Label label);
+
+    void clearConcludingIsaIndex(Label label);
+
+    void createConcludingHasAttributeIndex(Label type);
+
+    void clearConcludesHasAttributeIndex(Label type);
```
flyingsilverfin, line 62: This is where we "inject" higher-level inferences down into the graph, via the rule structure.

graph/structure/impl/RuleStructureImpl.java
```diff
@@ -107,6 +182,12 @@ public Buffered(SchemaGraph graph, StructureIID.Rule iid, String label, Conjunct
             super(graph, iid, label, when, then);
             this.isCommitted = new AtomicBoolean(false);
             setModified();
+            indexLabels();
+        }
+
+        private void indexLabels() {
```
flyingsilverfin, line 188: a buffered rule structure will index the type labels that are mentioned in the rule explicitly to prevent types from being undefined while the label is in use in a rule

graph/util/Encoding.java
```diff
@@ -164,11 +166,12 @@ public boolean isBackward() {
         VERTEX_ROLE(180, PrefixType.THING),
         STRUCTURE_RULE(190, PrefixType.RULE);
 
+
         private final byte key;
         private final PrefixType type;
 
         Prefix(int key, PrefixType type) {
-            this.key = (byte) (key & 0xff);
+            this.key = checkedCastUnsigned(key);
```
flyingsilverfin, line 174: with these new checked cast helpers, we can just specify if we want to use the range -128 to 127 or 0-255 :D

graph/iid/IndexIID.java
```diff
@@ -49,33 +49,123 @@
...
+            }
+
+            /**
+             * @return a byte array representing the the index scan prefix of a given type concluded in rules
+             */
+            public static RuleIndex concludedIsaScan(VertexIID.Type typeIID) {
+                return new RuleIndex(join(Encoding.Index.Prefix.TYPE.bytes(), typeIID.bytes(),
+                                          Encoding.Index.Infix.CONCLUDED_ISA.bytes()),
+                                     true);
+            }
```
flyingsilverfin, line 112: this is quite useful imo and i wonder if tis applicable in other places, introducing a `scan` iid option

graph/structure/impl/RuleStructureImpl.java
```diff
@@ -91,6 +102,26 @@ public boolean isDeleted() {
...
+    }
+
+    @Override
+    public void createConcludingHasAttributeIndex(Label type) {
+        graph.ruleIndex().concludingHasAttribute(this, type);
+    }
+
+    @Override
+    public void clearConcludesHasAttributeIndex(Label type) {
+        graph.ruleIndex().deleteConcludingHasAttribute(this, type);
```
flyingsilverfin, line 122: injecting data into the rule index from the higher level reasoning computation - we pass through the `RuleStructure` to the `SchemaGraph.RuleIndex`

logic/resolvable/Concludable.java
```diff
@@ -84,23 +84,15 @@ private Concludable(Conjunction conjunction) {
     }
 
     public ResourceIterator<Rule> getApplicableRules(ConceptManager conceptMgr, LogicManager logicMgr) {
-        if (applicableRules == null) computeApplicableRules(conceptMgr, logicMgr);
+        synchronized (this) {
+            if (applicableRules == null) applicableRules = applicableRules(conceptMgr, logicMgr);
+        }
```
flyingsilverfin, line 89: this class is shared across threads so the computation of applicable rules needs to be thread safe

test/integration/migrator/schema.gql
```diff
@@ -647,29 +647,6 @@ rule repo-collaborator-permission-write-owns-read-rule:
...
-    when
-    {
-        (repo: $x, team: $y) isa repo-team;
-        (team: $y, member: $z) isa team-member;
-    }
-    then
-    {
-        (repo: $x, collaborator: $z) isa repo-collaborator;
-    };
-
```
flyingsilverfin, line 672: deleted since these rules were unsatisfiable in this schema

test/integration/schema.gql
```diff
@@ -37,7 +37,8 @@ reference-type sub attribute, value string;
 format sub attribute, value string;
```
flyingsilverfin, line 37: updated to current Grabl schema

logic/Rule.java
```diff
@@ -61,56 +61,48 @@
 import static grakn.common.util.Objects.className;
 import static grakn.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
 import static grakn.core.common.exception.ErrorMessage.Pattern.INVALID_CASTING;
-import static grakn.core.logic.LogicManager.validateRuleStructureLabels;
 
 
 public class Rule {
 
-    private final LogicManager logicManager;
+    // note: as `Rule` is cached between transactions, we cannot hold any transaction-bound objects such as Managers
```
jmsfltchr, line 68: Why is `LogicManager` ok though? LogicManager only changes in a schema session?

logic/resolvable/Concludable.java
jmsfltchr: Now we can remove the `TODO Implement` in the constructor :)

logic/resolvable/Concludable.java
```diff
@@ -384,6 +376,29 @@ private boolean unificationSatisfiable(RolePlayer concludableRolePlayer, RolePla
...
+            Map<Rule, Set<Unifier>> applicableRules = new HashMap<>();
+            relationTypes.forEach(type -> logicMgr.rulesConcludingIsa(type)
+                    .forEachRemaining(rule -> unify(rule.conclusion(), conceptMgr)
+                            .forEachRemaining(unifier -> {
+                                applicableRules.putIfAbsent(rule, new HashSet<>());
+                                applicableRules.get(rule).add(unifier);
+                            })));
+
+            return applicableRules;
+        }
```
jmsfltchr, line 400: Nice

jmsfltchr (changes requested):

haikalpribadi (changes requested):

graph/SchemaGraph.java
```diff
@@ -451,20 +360,425 @@ public void commit() {
...
+            public void isOutdated(boolean isOutdated) {
+                this.outdated = isOutdated;
+            }
+
+            public ResourceIterator<RuleStructure> concludesVertex(TypeVertex type) {
+                assert !outdated;
+                return link(persisted.concludesVertex(type), buffered.concludesVertex(type));
+            }
+
+            public ResourceIterator<RuleStructure> concludesEdgeTo(TypeVertex type) {
```
haikalpribadi, line 539: does it really need the `to` suffix?

logic/Rule.java
```diff
@@ -215,14 +213,19 @@ private void pruneThenResolvedTypes() {
...
-    private Conjunction thenPattern(ThingVariable<?> thenVariable) {
-        return logicManager.typeResolver().resolveLabels(
-                new Conjunction(VariableRegistry.createFromThings(list(thenVariable)).variables(), set()));
+    private Conjunction thenPattern(ThingVariable<?> thenVariable, LogicManager logicMgr) {
+        // TODO when applying the type resolver, we should be using _insert semantics_ during the type resolution!!!
+        Conjunction conj = new Conjunction(VariableRegistry.createFromThings(list(thenVariable)).variables(), set());
+        return logicMgr.typeResolver().resolve(conj);
+    }
+
+    public void reIndex() {
```
haikalpribadi, line 226: "reindex" should be one word, like "unindex", right?

haikalpribadi (approved):
Only left minor comments on naming, but you can do that in follow up PRs. Well done, @flyingsilverfin !! You've come a long way in this PR :)

flyingsilverfin:
Thank you! Feels good!
I've fixed it already since i had to rebase again :)

PR #6036, opened 2021-01-15
Enable Graql Define tests in CI

alexjpwalker:

Our Graql Define tests were not being run in CI because some of them were failing. Given it was a relatively small number (8), we decided to ignore the failing tests for now, allowing us to add the passing ones to CI. This helps to guard against regressions.

- Enable Graql Define tests in CI

### DIFF
--- a/.grabl/automation.yml
+++ b/.grabl/automation.yml
@@ -76,7 +76,8 @@ build:
         bazel test //test/behaviour/concept/... --test_output=streamed
         bazel test //test/behaviour/graql/language/match/... --test_output=streamed
         bazel test //test/behaviour/graql/language/get/... --test_output=streamed
-      # bazel test //test/behaviour/graql/language/define/... --test_output=streamed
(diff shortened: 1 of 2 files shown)


.grabl/automation.yml
```diff
@@ -76,7 +76,8 @@ build:
         bazel test //test/behaviour/concept/... --test_output=streamed
         bazel test //test/behaviour/graql/language/match/... --test_output=streamed
         bazel test //test/behaviour/graql/language/get/... --test_output=streamed
-      # bazel test //test/behaviour/graql/language/define/... --test_output=streamed
+        bazel test //test/behaviour/graql/language/define/... --test_output=streamed
```
alexjpwalker, line 79: Our Graql Define tests were not being run in CI because some of them were failing. Given it was a relatively small number (8), we decided to ignore the failing tests for now, allowing us to add the passing ones to CI. This helps to guard against regressions.

haikalpribadi (approved):

PR #6048, opened 2021-01-19
Validate rules may only infer insertable type combinations

maxbaxt:

We use `TypeResolver` to validate rules before insertion. First, we check that removing the `Then`'s types that cannot appear in the `when` does not make the `then` unsatisfiable. Next, we check to see if the rule's `when` has type permutations that the `then` cannot satisfy. If there is, we invalidate the rule and throw an error.

* Adjusted `TypeResolver` to have an optional flag to check specify if the `conjunction` it is running on also needs to be insertable.
* Added new validation logic in `Rule` class to check if the rule's `then` and `when` resolved types are compatible.
* Added `isSystemRefernce()` helper method to `SystemReference`.


logic/tool/TypeResolver.java
```diff
@@ -118,7 +146,8 @@ public Conjunction resolve(Conjunction conjunction) {
                         mapping.putIfAbsent(ref, new HashSet<>());
                         assert vertex.isType();
                         // TODO: This filter should not be needed if we enforce traversal only to return non-abstract
-                        if (!(vertex.asType().isAbstract() && traversalConstructor.getVariable(ref).isThing()))
+                        if (!traversalConstructor.getVariable(ref).isPresent()) return;
+                        if (!(vertex.asType().isAbstract() && traversalConstructor.getVariable(ref).get().isThing()))
```
maxbaxt, line 150: This might not be the neatest way, but this will be refactored in an upcoming PR anyway.

logic/tool/TypeResolver.java
```diff
@@ -86,10 +102,10 @@ public Conjunction resolveLabels(Conjunction conjunction) {
         return conjunction;
     }
 
-    public Conjunction resolve(Conjunction conjunction) {
+    public Conjunction resolve(Conjunction conjunction, boolean insertable) {
         resolveLabels(conjunction);
-        TraversalBuilder traversalConstructor = new TraversalBuilder(conjunction, conceptMgr);
-        Map<Reference, Set<Label>> resolvedLabels = executeResolverTraversals(traversalConstructor);
+        TraversalBuilder traversalBuilder = new TraversalBuilder(conjunction, conceptMgr, insertable);
```
maxbaxt, line 107: Changed to `traversalBuilder` to reflect the class now being `TraversalBuilder`

logic/Rule.java
```diff
@@ -168,25 +168,23 @@ public final int hashCode() {
...
+        ResourceIterator<Map<Reference.Name, Label>> possibleWhenPerms = logicManager.typeResolver().retrievePossibleTypeCombos(when, false);
+        Set<Map<Reference.Name, Label>> possibleThenSet = logicManager.typeResolver().retrievePossibleTypeCombos(then, true).toSet();
 
-        To do this you may want to
-        1. utilise the streaming mode of `TypeResolver` (can re-run it, thats ok) to get all combinations
-        2. utilise the `Rule.Conclusion` classes before (there's exactly 1 of the 3 per rule) to validate a combination is compatible with the conclusion
-           using these Rule.Conclusion data structures indicates that it is a "high level" logical rule validation, which is true
-         */
+        possibleWhenPerms.forEachRemaining(nameLabelMap -> {
+            if (possibleThenSet.stream().noneMatch(thenMap -> nameLabelMap.entrySet().containsAll(thenMap.entrySet())))
```
maxbaxt, line 185: All named vars in the `then` are bound in the `when`, but not the other way around. hence use `containsAll` not `equals`.

logic/tool/TypeResolver.java
```diff
@@ -86,10 +101,10 @@ public Conjunction resolveLabels(Conjunction conjunction) {
...
+        Map<Reference, Set<Label>> resolvedLabels = executeResolverTraversals(traversalBuilder);
         if (resolvedLabels.isEmpty()) {
             conjunction.setSatisfiable(false);
```
flyingsilverfin, line 109: since we set conjunctions as satisfiable now, instead of variables do we: 
1. at all need the isSatisfiable flag in Variable?
2. in rule validation, do we not have to check the conjunction's `isSatisfiable` flag instead of each variable's?

logic/tool/TypeResolver.java
```diff
@@ -68,6 +70,19 @@ public TypeResolver(ConceptManager conceptMgr, TraversalEngine traversalEng, Log
         this.logicCache = logicCache;
     }
 
+    public ResourceIterator<Map<Reference.Name, Label>> retrievePossibleTypeCombos(Conjunction conjunction, boolean insertable) {
```
flyingsilverfin, line 73: i think we can just call this `combinations(Conjunction, boolean)`

logic/tool/TypeResolver.java
flyingsilverfin: v good, consistency is great

logic/tool/TypeResolver.java
```diff
@@ -220,7 +253,7 @@ private void registerHas(TypeVariable owner, HasConstraint hasConstraint) {
         private void registerRelation(TypeVariable owner, RelationConstraint constraint) {
             for (RelationConstraint.RolePlayer rolePlayer : constraint.players()) {
                 TypeVariable playerResolver = register(rolePlayer.player());
-                TypeVariable actingRoleResolver = register(new TypeVariable(newSystemId()));
+                TypeVariable actingRoleResolver = new TypeVariable(newSystemId());
```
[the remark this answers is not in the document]
flyingsilverfin: I'm not quite following - do we do this for all new system-id generated TypeVariables? How does this not break the fact that we do want to put resolved labels on anonymous variables in when doing normal type resolution?

test/integration/logic/RuleTest.java
```diff
@@ -381,4 +385,56 @@ public void rule_has_explicit_materialises_when_missing() throws IOException {
...
+                    dog.setSupertype(animal);
+
+                    assertThrows(() -> txn.logic().putRule(
+                            "animals-are-named-fido",
+                            Graql.parsePattern("{$x isa animal;}").asConjunction(),
+                            Graql.parseVariable("$x has name 'fido'").asThing()));
+                }
+            }
+        }
+    }
```
flyingsilverfin, line 439: can we also add a 1-4 tests for `insert` in `TypeResolver` please? Maybe if you can think of some slightly non-obvious cases and a few obvious ones

flyingsilverfin (changes requested):

logic/tool/TypeResolver.java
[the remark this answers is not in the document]
maxbaxt: Because this is not an anonymous variable. When a `roleType` is missing, it simply is missing, there is no anonymous variable to add `resolvedTypes` to.

logic/tool/TypeResolver.java
maxbaxt: Changed.

logic/tool/TypeResolver.java
maxbaxt: Discussed with Josh, we will park this issue as it needs more architectural thought.

test/integration/logic/RuleTest.java
maxbaxt: Added rules to test this.

test/integration/logic/TypeResolverTest.java
```diff
@@ -1015,4 +1015,78 @@ public void roles_can_handle_type_constraints() throws IOException {
...
+        resolveConjunction(typeResolver, queryString);
+
+        assertThrows(() -> transaction.logic().putRule(
+                "marriage-rule",
+                Graql.parsePattern("{$x isa person;}").asConjunction(),
+                Graql.parseVariable("(wife: $x) isa partnership").asThing()));
+    }
+
+    @Test
+    public void cannot_insert_if_role_is_too_general() {
```
flyingsilverfin, line 1065: nice

test/integration/logic/resolvable/UnifyHasConcludableTest.java
```diff
@@ -436,6 +436,8 @@ public void has_many_to_one_unifier() {
     }
 
     @Test
+    @Ignore
+    //TODO: ignored because this rule is no longer valid
```
flyingsilverfin, line 440: if its not updatedable to a valid test, we should remove it. Can we fix it here already?

maxbaxt: I think the test is has to be rewritten. The point is to make a one-to-many unifier, but it's the way it goes about it isn't possible anymore. Therefore I will write a new test to create a one-to-many unifier situation.

flyingsilverfin (approved):
this is so cool! haha

PR #6050, opened 2021-01-19
Permit override of played roles in Plays constraints

Phillammon:

Specifying the scope of an overridden role is not allowed syntactically when specifying a Plays constraint in graql, and previously, it was not possible to infer a label appropriately for overridden roles. We added, in combination with some changes in graql, the ability to correctly infer the scope of the overridden type, allowing overridden types in definitions.

Where an overridden type is supplied in a define query's plays clause, this is caught in definePlays and the overriding role's supertypes are scanned for an appropriately named roleType. If one is found, this roletype's relation is used as a scope for the overridden type, allowing it to be correctly added to the Grakn graph.

### DIFF
--- a/dependencies/graknlabs/repositories.bzl
+++ b/dependencies/graknlabs/repositories.bzl
@@ -36,7 +36,7 @@ def graknlabs_graql():
     git_repository(
         name = "graknlabs_graql",
         remote = "https://github.com/Phillammon/graql",
-        commit = "4415abe06af9e358051c0f98585b7276841726a7", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_graql
+        commit = "13d201e7e056aea7a664f1a4dd639ab4e648bd48", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_graql
     )
 
 def graknlabs_protocol():
--- a/dependencies/graknlabs/repositories.bzl
+++ b/dependencies/graknlabs/repositories.bzl
@@ -36,7 +36,7 @@ def graknlabs_graql():
     git_repository(
         name = "graknlabs_graql",
         remote = "https://github.com/Phillammon/graql",
-        commit = "05746da0ebdd56e633c14f48cf29580fa1b57d3e", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_graql
+        commit = "4415abe06af9e358051c0f98585b7276841726a7", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_graql
     )
 
 def graknlabs_protocol():
--- a/dependencies/graknlabs/repositories.bzl
+++ b/dependencies/graknlabs/repositories.bzl
@@ -35,8 +35,8 @@ def graknlabs_common():
 def graknlabs_graql():
     git_repository(
         name = "graknlabs_graql",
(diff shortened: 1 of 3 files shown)


dependencies/graknlabs/repositories.bzl
```diff
@@ -56,6 +56,6 @@ def graknlabs_grabl_tracing():
 def graknlabs_behaviour():
     git_repository(
         name = "graknlabs_behaviour",
-        remote = "https://github.com/graknlabs/behaviour",
-        commit = "e1db016c66d24df58df3165b457079f78db29507", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_behaviour
+        remote = "https://github.com/Phillammon/behaviour",
+        commit = "e49acf7f8af427a7b1eee05f832a6df576b18552", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_behaviour
```
flyingsilverfin, line 60: don't forget to revert!

flyingsilverfin (approved):

flyingsilverfin:
why does the title mention that this is specific to `Relation` types? surely its a problem for any `plays...as` no?

flyingsilverfin:
Ah come to think - please check out this edge case:
What if the thing you're trying to override with `as`, eg `plays R:A as C`, but `C` is itself shadowed?
for instance in the following schema;
```
define 
  R1 sub relation, relates C;
  R2 sub R1, relates B as C;
  R3 sub R2, relates A as B;
 
  X1 sub entity, plays R1:C;
  X2 sub X1, plays R2:B as C;
  X3 sub X2, plays R3:A as B
```
this shouldn't be allowed anymore, as `B` isn't a type that X3 inherits right?
So intuitively, the check should be: override `B` if `X3` has a role its currently allowed to play via inheritance, which also happens to be a supertype of `R3:A`

am i making sense?

flyingsilverfin:
don't forget labels, project, and milestone :)

dependencies/graknlabs/repositories.bzl
Phillammon: Will be reverting once they're in live, this is mostly for making sure CI works for the time being

query/Definer.java
```diff
@@ -245,10 +247,18 @@ private void definePlays(ThingType thingType, Set<PlaysConstraint> playsConstrai
...
+                final RoleType roleType = getRoleType(roleTypeLabel).asRoleType();
                 if (plays.overridden().isPresent()) {
-                    final RoleType overriddenType = getRoleType(plays.overridden().get().label().get()).asRoleType();
-                    thingType.setPlays(roleType, overriddenType);
+                    String overriddenLabelName = plays.overridden().get().label().get().properLabel().name();
+                    Optional<? extends RoleType> overriddenType = roleType.getSupertypes()
+                            .filter(rt -> rt.getLabel().name().equals(overriddenLabelName)).findFirst();
+                    if (overriddenType.isPresent()) {
+                        thingType.setPlays(roleType, overriddenType.get());
+
```
haikalpribadi, line 258: not a big deal, we can clean this up later, but don't forget to clean up the formatting of the PR, @Phillammon

haikalpribadi (approved):

haikalpribadi:
Congrats on your first PR to Grakn Core, @Phillammon ! :)

PR #6052, opened 2021-01-20
Rewritten Producers to be thread-safe with a clean algorithm

### LINKED ISSUE #6042: Issues with ConcurrentModificationException and Iterable producer
## Description

To reproduce this issue:

1) Install client python by running 
`pip3 install --extra-index-url https://repo.grakn.ai/repository/pypi-snapshot/simple grakn-client==0.0.0-17a1a94e1efba0f34234f8f1d0ecab692e2c4ba6`

2) run migrator.py from biograkn repo using (sha `caeb3454f0557162a22f5324d2146ff512f705dd`)
`python3 migrator.py`

3) this may run for a while, I stopped it after ~15 minutes. No need to upload the whole data, but wait for at least 15 - 20 minutes.

4) Run grakn server with this command from the root of the grakn project
`bazel run //server:server-bin-mac`

5) Run console with this command from the root of the console project

6) in console run 
`transaction biograkn_covid_19 data read`
to enter `biograkn_covid_19` `read data` transaction from previously migrated data.

7) run this query
`match ($x, $y); ($y,$z); ($x, $z);`

this ^^ query usually breaks and kills everything (both console and grakn). In case it didn't kill grakn, wait a bit (the results will be showing up slowly).

8) stop the query from the console with Ctrl + C

9) Notice that in the `htop` the usage of cores did not drop, although the query is stopped (this doesn't happen with match `$x isa thing;` for example).

10) run
`match ($x, $y); ($y,$z); ($x, $z);`

again

11) at this point if grakn and console still alive, you can stop grakn server with Ctrl + C. This will most likely throw the issue with SIGSEGV and rocksdb, or some other similar issue.

12) after that, restart grakn server and console and do the same queries. For me, after I got the 'big' issue once, all the following queries in console failed immediately with errors in grakn, so even

   `match ($x, $y); ($y,$z);`
 
wouldn't work, although it works if you start grakn 'fresh' after restarting the computer. 

## Bugs discovered in `IterableProducer` and `GraphProducer`: 

1) `.put`  method in `.iterableProducer.Queue` does 2 things:

     `calls blockingQueue.put`
     `calls pending.decrement`

     there are 3 key time points: 
         `blockingQueue.put` started
         `blockingQueue.put` ended
         `pending.decrement` started and ended

     `mayProduce` might be called in between those time points. If it is called after `blockingQueue.put` is ended but before `pending.decrement` is called, it thinks that there is an element in the queue and an element pending to get in a queue (which is the same element). 

    Because of that `mayProduce` thinks that there is a negative amount of slots available. For example: let's say the maximum buffer is 1 slot, then there is 1 element in a queue, 1 element pending to get in a queue (which is the same element that is already in a queue). This means we have -1 available slots. This might be not critical and actually work, but it is not what is expected from this function. It can be solved by making `.put` synchronized (and maybe the rest of the function other than `.take`). Also, the queue should anticipate getting more than it asked for.

2) There is an issue that in case of exception in graphProducer, it catches it and calls done on a queue, which doesn't stop already existing iterators to continue to put stuff in the queue (because it is not the same done call as the one when all the iterators exhausted). But when we called done on a queue, it set its pending to 0, so all the following puts after the queue was called done will push pending below 0. It can be fixed by ending all the iterators once we caught an exception.

3) There is an issue in the math
`int splitCount = (int) Math.ceil((double) count / iteratorRequested.size())`

    A lot of iterators being called to compensate with 1 or 2 items left (`toCompensate`). 
If you do the math and divide 1 / 8 (there can be 8 iterators running) with rounding up, it will be equal 1, and this 1 (`splitCount`) will be added to `requested` of every iterator (8), they will produce more answers, pending will go below 0 and it will cause math to go wrong in `mayProduce`. It can be fixed by requesting a precise number of answers.

4) `mayProduce` calls 

      ```
          ExecutorService.forkJoinPool().submit(() -> {            
               assert !producers.isEmpty();
               producers.peek().produce(queue, available);
          });
      ```
      
       Inside `.produce` there is the following code 

      ```
          for (ResourceIterator<VertexMap> iterator : iteratorRequested.keySet()) {
                assert iteratorRequested.containsKey(iterator);
                iteratorRequested.computeIfPresent(iterator, (k, v) -> v + splitCount);
                if (!iteratorJobs.containsKey(iterator)) {
                    iteratorJobs.put(iterator, runAsync(() -> produceAsync(queue, iterator), forkJoinPool()));
                }
           }
      ```
      
         So we have two functions that call each other in a loop and each of them calls each other in parallel, which is hard to debug and predict it's behavior, given that they use `iterator` which should not be used by 2 different threads at the same time

5) `.produce` calls in parallel for `produceAsync` in a loop

      ```
           for (ResourceIterator<VertexMap> iterator : iteratorRequested.keySet()) {
                ...
                    iteratorJobs.put(iterator, runAsync(() -> produceAsync(queue, iterator), forkJoinPool()));
                ...
           }
      ```
      
       each of the `produceAsync` calls will eventually call 
      ```
          compensate(queue, iterator, requested - produced);
      ``` 
      
      which will call 
      
      ```
      .produce(queue, toCompensate);
      ```
      
      Although compensateis called only when the iterator is exauhsted, it will call produce, which will iterate through ALL the iterators, so same iterator can be used by multiple threads.
      
      Can be fixed by locking the iterators `.has` and `.hasNext`

6) Ongoing invesigation - `done` is not being called on a queue once all the iterators are exhausted.

haikalpribadi:

The first draft of the Async Producer, that came out with 2.0.0-alpha release, was not yet fully thread-safe, which cause plenty of bugs that were all caused by various race conditions. It also caused "thread leaks" that overburdened the CPU to the point of unusable over time. We have now fixed this problem by rewriting the Async Producer to be thread-safe from the start, with an algorithm that is much easier to follow, which allows us to easily improve it in the future.

Fixes #6042

### DIFF
--- a/common/concurrent/ExecutorService.java
+++ b/common/concurrent/ExecutorService.java
@@ -22,10 +22,9 @@ import grakn.core.common.concurrent.actor.EventLoopGroup;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.concurrent.Executors;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
-
-import static java.util.concurrent.ForkJoinPool.defaultForkJoinWorkerThreadFactory;
 
 public class ExecutorService {
 
@@ -39,8 +38,7 @@ public class ExecutorService {
     private final ScheduledThreadPoolExecutor scheduledThreadPool;
 
     private ExecutorService(int parallelisation) {
-        forkJoinPool = new ForkJoinPool(parallelisation, defaultForkJoinWorkerThreadFactory,
-                                        (t, e) -> LOG.error(e.getMessage(), e), true);
+        forkJoinPool = (ForkJoinPool) Executors.newWorkStealingPool(parallelisation);
         eventLoopGroup = new EventLoopGroup(parallelisation, "grakn-elg");
         scheduledThreadPool = new ScheduledThreadPoolExecutor(1);
         scheduledThreadPool.setRemoveOnCancelPolicy(true);
--- a/common/concurrent/ExecutorService.java
+++ b/common/concurrent/ExecutorService.java
@@ -19,12 +19,17 @@
 package grakn.core.common.concurrent;
 
 import grakn.core.common.concurrent.actor.EventLoopGroup;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import java.util.concurrent.Executors;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 
+import static java.util.concurrent.ForkJoinPool.defaultForkJoinWorkerThreadFactory;
+
 public class ExecutorService {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ExecutorService.class);
 
     public static int PARALLELISATION_FACTOR = -1;
     private static ExecutorService singleton = null;
@@ -33,9 +38,10 @@ public class ExecutorService {
     private final EventLoopGroup eventLoopGroup;
     private final ScheduledThreadPoolExecutor scheduledThreadPool;
 
-    private ExecutorService(int parallelisationFactor) {
-        forkJoinPool = (ForkJoinPool) Executors.newWorkStealingPool(parallelisationFactor);
-        eventLoopGroup = new EventLoopGroup(parallelisationFactor, "grakn-elg");
+    private ExecutorService(int parallelisation) {
+        forkJoinPool = new ForkJoinPool(parallelisation, defaultForkJoinWorkerThreadFactory,
+                                        (t, e) -> LOG.error(e.getMessage(), e), true);
+        eventLoopGroup = new EventLoopGroup(parallelisation, "grakn-elg");
         scheduledThreadPool = new ScheduledThreadPoolExecutor(1);
         scheduledThreadPool.setRemoveOnCancelPolicy(true);
     }
--- a/common/producer/BaseProducer.java
+++ b/common/producer/BaseProducer.java
@@ -20,42 +20,53 @@ package grakn.core.common.producer;
 
 import grakn.core.common.iterator.ResourceIterator;
 
+import javax.annotation.concurrent.ThreadSafe;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import static grakn.core.common.concurrent.ExecutorService.forkJoinPool;
 
+@ThreadSafe
 public class BaseProducer<T> implements Producer<T> {
 
     private final ResourceIterator<T> iterator;
+    private final AtomicBoolean isDone;
     private CompletableFuture<Void> future;
 
     BaseProducer(ResourceIterator<T> iterator) {
         this.iterator = iterator;
+        this.isDone = new AtomicBoolean(false);
         this.future = CompletableFuture.completedFuture(null);
     }
 
     @Override
     public synchronized void produce(Queue<T> queue, int request) {
-        future = future.thenRunAsync(() -> produceAsync(queue, request), forkJoinPool());
+        if (isDone.get()) return;
+        future = future.thenRunAsync(() -> {
+            try {
+                int i = 0;
+                for (; i < request && iterator.hasNext() && !isDone.get(); i++) queue.put(iterator.next());
+                if (i < request && !isDone.get()) done(queue);
+            } catch (Throwable e) {
+                queue.done(e);
+            }
+        }, forkJoinPool());
     }
 
-    private void produceAsync(Queue<T> queue, int count) {
-        try {
-            for (int i = 0; i < count; i++) {
-                if (iterator.hasNext()) {
-                    queue.put(iterator.next());
-                } else {
-                    queue.done();
-                    break;
-                }
-            }
-        } catch (Throwable e) {
+    private void done(Queue<T> queue) {
+        if (isDone.compareAndSet(false, true)) {
+            queue.done();
+        }
+    }
+
+    private void done(Queue<T> queue, Throwable e) {
+        if (isDone.compareAndSet(false, true)) {
             queue.done(e);
         }
     }
 
     @Override
-    public void recycle() {
+    public synchronized void recycle() {
         iterator.recycle();
     }
 }
--- a/common/producer/BaseProducer.java
+++ b/common/producer/BaseProducer.java
@@ -35,8 +35,8 @@ public class BaseProducer<T> implements Producer<T> {
     }
 
     @Override
-    public synchronized void produce(Queue<T> queue, int count) {
-        future = future.thenRunAsync(() -> produceAsync(queue, count), forkJoinPool());
+    public synchronized void produce(Queue<T> queue, int request) {
+        future = future.thenRunAsync(() -> produceAsync(queue, request), forkJoinPool());
     }
 
     private void produceAsync(Queue<T> queue, int count) {
@@ -45,12 +45,12 @@ public class BaseProducer<T> implements Producer<T> {
                 if (iterator.hasNext()) {
                     queue.put(iterator.next());
                 } else {
-                    queue.done(this);
+                    queue.done();
                     break;
                 }
             }
         } catch (Throwable e) {
-            queue.done(this, e);
+            queue.done(e);
         }
     }
 
--- a/common/producer/BaseProducer.java
+++ b/common/producer/BaseProducer.java
@@ -44,9 +44,9 @@ public class BaseProducer<T> implements Producer<T> {
         if (isDone.get()) return;
         future = future.thenRunAsync(() -> {
             try {
-                int i = 0;
-                for (; i < request && iterator.hasNext() && !isDone.get(); i++) queue.put(iterator.next());
-                if (i < request && !isDone.get()) done(queue);
+                int unfulfilled = request;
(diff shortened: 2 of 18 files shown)


VladGan:
one bug I found - `runningJobs.size()` can grow way over 8

common/producer/FilteredProducer.java
```diff
@@ -31,15 +33,16 @@
...
+    public void produce(Producer.Queue<T> queue, int request) {
+        baseProducer.produce(new Queue(queue), request);
     }
 
     @Override
     public void recycle() {
         baseProducer.recycle();
     }
 
+    @ThreadSafe
```
lriuui0x0, line 45: Does this do anything?

traversal/producer/GraphProducer.java
```diff
@@ -46,91 +47,102 @@
...
     @Override
-    public synchronized void produce(Queue<VertexMap> queue, int count) {
-        if (iteratorRequested.size() < parallelisation) {
-            for (int i = iteratorRequested.size(); i < parallelisation && start.hasNext(); i++) {
-                ResourceIterator<VertexMap> iterator =
-                        new GraphIterator(graphMgr, start.next(), procedure, params, filter).distinct(produced);
-                iteratorRequested.put(iterator, 0);
-            }
+    public synchronized void produce(Producer.Queue<VertexMap> queue, int request) {
+        if (isDone.get()) return;
```
lriuui0x0, line 71: Is it possible that `isDone` is set to `true` after this check? Is this OK?

traversal/producer/GraphProducer.java
```diff
@@ -46,91 +47,102 @@
...
         }
     }
 
-    private synchronized int take(ResourceIterator<VertexMap> iterator) {
-        int count = iteratorRequested.get(iterator);
-        iteratorRequested.put(iterator, 0);
-        return count;
+    private synchronized void transition(Queue<VertexMap> queue, ResourceIterator<VertexMap> iterator, int unfulfilled) {
+        if (iterator.hasNext()) return;
+        if (runningJobs.containsKey(iterator) && start.hasNext()) replace(queue, iterator, unfulfilled);
```
lriuui0x0, line 102: Is `runningJobs.containsKey(iterator)` condition always true?

traversal/producer/GraphProducer.java
```diff
@@ -46,91 +47,102 @@
...
-            int requested;
-            int produced;
-            do {
-                requested = take(iterator);
-                produced = 0;
-                for (; produced < requested && iterator.hasNext(); produced++)
+            int i = 0;
+            if (runningJobs.containsKey(iterator)) {
+                for (; i < request && iterator.hasNext() && !isDone.get(); i++) {
                     queue.put(iterator.next());
```
lriuui0x0, line 122: Is it possible that `done()` is called after `isDone.get()` check?

traversal/producer/GraphProducer.java
```diff
@@ -46,91 +47,102 @@
...
 
-    private synchronized void compensate(Queue<VertexMap> queue, ResourceIterator<VertexMap> iterator, int requested) {
-        iteratorJobs.remove(iterator);
-        int toCompensate = requested + take(iterator);
-        if (!iterator.hasNext()) iteratorRequested.remove(iterator);
-        if (toCompensate > 0) produce(queue, toCompensate);
+    private synchronized void replace(Queue<VertexMap> queue, ResourceIterator<VertexMap> iterator, int unfulfilled) {
+        runningJobs.remove(iterator);
+        ResourceIterator<VertexMap> newIter =
+                new GraphIterator(graphMgr, start.next(), procedure, params, filter).distinct(produced);
```
lriuui0x0, line 110: Maybe assert `start.hasNext()` here. It's not clear what the preconditions of this function are.

traversal/producer/GraphProducer.java
```diff
@@ -46,91 +47,102 @@
...
     }
 
-    private synchronized int take(ResourceIterator<VertexMap> iterator) {
-        int count = iteratorRequested.get(iterator);
-        iteratorRequested.put(iterator, 0);
-        return count;
+    private synchronized void transition(Queue<VertexMap> queue, ResourceIterator<VertexMap> iterator, int unfulfilled) {
+        if (iterator.hasNext()) return;
+        if (runningJobs.containsKey(iterator) && start.hasNext()) replace(queue, iterator, unfulfilled);
+        else if (!runningJobs.isEmpty() && unfulfilled > 0) distribute(queue, unfulfilled);
```
lriuui0x0, line 103: Would changing this logic to this make the branching a bit clearer?

```
if (!runningJobs.isEmpty()) {
  if (unfulfilled > 0) distribute(queue, unfulfilled);
} else {
  done(queue);
}
```

common/producer/FilteredProducer.java
haikalpribadi: Snnotations are generally there to serve as a "reminder" of the code's behaviour, @lriuui0x0

traversal/producer/GraphProducer.java
haikalpribadi: It is possible. But I think it's okay?

traversal/producer/GraphProducer.java
haikalpribadi: not always, that was the bug that @VladGan found earlier.

traversal/producer/GraphProducer.java
haikalpribadi: it is possible, but only when there has been an error. I don't think it's possible if the producer was ended correctly?

common/concurrent/ExecutorService.java
```diff
@@ -19,23 +19,27 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.concurrent.Executors;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 
 public class ExecutorService {
 
+    private static final Logger LOG = LoggerFactory.getLogger(ExecutorService.class);
```
VladGan, line 31: not used

VladGan, line 31: not used

common/producer/BaseProducer.java
```diff
@@ -20,42 +20,53 @@
...
         this.iterator = iterator;
+        this.isDone = new AtomicBoolean(false);
         this.future = CompletableFuture.completedFuture(null);
     }
 
     @Override
-    public synchronized void produce(Queue<T> queue, int count) {
-        future = future.thenRunAsync(() -> produceAsync(queue, count), forkJoinPool());
+    public synchronized void produce(Queue<T> queue, int request) {
+        if (isDone.get()) return;
```
VladGan, line 44: does it need to be atomic given that `produce` is `synchronized`?

common/producer/BaseProducer.java
```diff
@@ -20,42 +20,53 @@
...
+        future = future.thenRunAsync(() -> {
+            try {
+                int i = 0;
+                for (; i < request && iterator.hasNext() && !isDone.get(); i++) queue.put(iterator.next());
+                if (i < request && !isDone.get()) done(queue);
```
VladGan, line 49: I think we can do

```
for (; request > 0 && iterator.hasNext() && !isDone.get(); request--) queue.put(iterator.next());
if (request > 0 && !isDone.get()) done(queue);
```

common/producer/ProducerIterator.java
```diff
@@ -0,0 +1,196 @@
...
+            } catch (InterruptedException e) {
+                throw GraknException.of(e);
+            }
+        }
+
+        private int size() {
+            return blockingQueue.size();
+        }
+    }
+}
```
VladGan, line 196: not reviewing it since we tested it a lot

traversal/producer/GraphProducer.java
```diff
@@ -46,91 +47,106 @@
...
         }
     }
 
     @Override
     public synchronized void recycle() {
         start.recycle();
-        iteratorJobs.keySet().forEach(ResourceIterator::recycle);
+        runningJobs.keySet().forEach(ResourceIterator::recycle);
     }
 }
```
VladGan, line 152: not reviewing it

VladGan (changes requested):

traversal/producer/GraphProducer.java
haikalpribadi: I've removed `iterator` from the input completely now

traversal/producer/GraphProducer.java
haikalpribadi: I've clarified this method differently now

common/concurrent/ExecutorService.java
haikalpribadi: we're keeping them in every class for now, cause we intend to add improved logging soon

common/producer/BaseProducer.java
haikalpribadi: `done(queue)` and `done(queue, error)` are not synchronised

common/producer/BaseProducer.java
haikalpribadi: I've taken inspiration from this and do this instead:
```
int unfulfilled = request;
if (runningJobs.containsKey(iterator)) {
    for (; unfulfilled > 0 && iterator.hasNext() && !isDone.get(); unfulfilled--) {
        queue.put(iterator.next());
    }
}
if (!isDone.get()) transition(queue, iterator, unfulfilled);
```

NOTE: we should not modify input arguments if we can afford to. We should treat them as final, and create a new variable in which we can modify.

common/producer/FilteredProducer.java
haikalpribadi: Similar to `@Override`, @lriuui0x0

PR #6053, opened 2021-01-20
Keep the protocol version in sync with Grakn Cluster

lolski:

The protocol version included in Grakn Core has been kept in sync with what's needed by Grakn Cluster.

- Update protocol version to the one that has client failover functionalities


haikalpribadi:
I'm not sure I understand the title of this PR, @lolski. Do you mean "Synchronise the protocol with Grakn Cluster"?

PR #6054, opened 2021-01-20
Reasoner integration

jmsfltchr:

Provide the final integration between components (along with the necessary fixes), such that the reasoner runs end-to-end for basic reasoning queries.

- Implement real traversals, removing the last of the mocking in reasoner, and thereby running the first working reasoner queries successfully! This required functionality to implement bound traversals - for use when supplying a set of known ids to add to a traversal. Critical since we execute conjunctions knowing a partial answer in all of reasoner's resolvers.
- Various bug fixes necessary to get the reasoner up and running:
  - Update `iterationInferredAnswer` so that reiteration terminates.
  - AnswerState merging of answers was duplicated and done incorrectly. Now `aggregate` methods are removed, in favour of `withInitial()`.
  - Removed some overly-aggressive assertions.
  - Behaviour of RuleResolver and RootResolver when they received exhausted messages.
  - Add return statements to while loops so that they terminate.
  - Missing variable from materialisation when materialising rules.
  - Unification of inequality value constraints.
- Add basic tests of reasoner queries, `ReasonerTest`.


lolski (approved):

pattern/Conjunction.java
```diff
@@ -123,6 +123,12 @@ public Traversal traversal(List<Identifier.Variable.Name> filter) {
         return traversal;
     }
 
+    public Traversal traversal() {
+        Traversal traversal = new Traversal();
+        variableSet.forEach(variable -> variable.addTo(traversal));
+        return traversal;
+    }
```
flyingsilverfin, line 130: how is this different from the above method with an empty set of filters? if you want both, then one should call the other and add the filters inside

logic/resolvable/Concludable.java
```diff
@@ -249,6 +249,12 @@ else if (a.getType().isDouble())
         return predicateFn;
     }
 
+    protected void addValueEqualsRequirements(Unifier.Builder unifierBuilder, Set<ValueConstraint<?>> values) {
+        for (ValueConstraint<?> value : equalsConstraints(values)) {
```
flyingsilverfin, line 253: maybe able to inline `equalsConstraints` if not used anywhere else

reasoner/Reasoner.java
```diff
@@ -82,7 +82,7 @@ private boolean isInfer() {
                                                 Context.Query context) {
         ResourceIterator<Conjunction> conjunctions = iterate(disjunction.conjunctions());
         if (!context.options().parallel()) return conjunctions.flatMap(conj -> iterator(conj, filter));
-        else return produce(conjunctions.flatMap(conj -> producers(conj, filter)).toList());
+        else return produce(conjunctions.flatMap(conj -> producers(conj, filter)).toList()).distinct();
```
flyingsilverfin, line 85: we wanted to remove this no?

reasoner/resolution/answer/AnswerState.java
```diff
@@ -58,6 +58,13 @@ public int hashCode() {
         return Objects.hash(conceptMap);
     }
 
+    @Override
+    public String toString() {
+        return "AnswerState{" +
+                "conceptMap=" + conceptMap +
+                '}';
+    }
```
flyingsilverfin, line 66: can actually be deleted, impl by each subclass

test/integration/reasoner/ReasonerTest.java
```diff
@@ -0,0 +1,184 @@
...
+        grakn.databases().create(database);
+    }
+
+    @After
+    public void tearDown() {
+        grakn.close();
+    }
+
+    @Test
+    public void test_basic() {
```
flyingsilverfin, line 71: basic is a bad name imo, `test_no_rules()` better?

test/integration/reasoner/ReasonerTest.java
```diff
@@ -0,0 +1,184 @@
...
+                    assertEquals("person", a.get("p1").asThing().getType().getLabel().scopedName());
+                    assertEquals("person", a.get("p2").asThing().getType().getLabel().scopedName());
+                    assertEquals("name", a.get("na").asAttribute().getType().getLabel().scopedName());
+                });
+
+                assertEquals(2, ans.size());
+            }
+        }
+    }
+}
```
flyingsilverfin, line 184: eheheheehehehehehehehehehehe

flyingsilverfin (changes requested):

reasoner/Reasoner.java
jmsfltchr: Yeah but then the basic test fails! So I put it back.

pattern/Conjunction.java
jmsfltchr: Fair point!

logic/resolvable/Concludable.java
jmsfltchr: It's used in 3 places which is why I pulled it out: in Has, Value and Attribute.

flyingsilverfin (approved):

PR #6055, opened 2021-01-20
Fix undefiner bugs from BDD

Phillammon:

Repairing the undefine behavioural tests revealed a number of flaws in the undefine process for the core server, which we proceeded to fix. Among these are changes to the behaviour of attempting to unset plays or owns on types that do not play or own anything, which previously was a no-op but now throws, as well as fixes to permit the undefinition of relationships that have roles set on them, and steps for testing rules.

- Added exceptions to cover undefining a nonexistent owns, ownskey, plays or relate edge (either due to a connection being inherited or just simply being unconnected
- Throw these exceptions appropriately.
- When undefining role types as part of undefining a relationship, we match the role's scope to the relation, not the role's scope to the relation's scope.
- Rules BDD steps added


common/exception/ErrorMessage.java
```diff
@@ -414,12 +414,24 @@ public static void loadConstants() {
                 new TypeWrite(36, "The ability of type '%s' to own attribute type '%s' cannot be undefined because it is currently owned by existing instances.");
         public static final TypeWrite INVALID_UNDEFINE_PLAYS_HAS_INSTANCES =
                 new TypeWrite(37, "The ability of type '%s' to play role type '%s' cannot be undefined because it is currently played by existing instances.");
+        public static final TypeWrite INVALID_UNDEFINE_INHERITED_RELATES =
+                new TypeWrite(38, "The ability of type '%s' to relate role type '%s' cannot be undefined because the ability is inherited from a supertype.");
+        public static final TypeWrite INVALID_UNDEFINE_INHERITED_OWNS =
+                new TypeWrite(39, "The ability of type '%s' to own attribute type '%s' cannot be undefined because the ability is inherited from a supertype.");
+        public static final TypeWrite INVALID_UNDEFINE_INHERITED_PLAYS =
+                new TypeWrite(40, "The ability of type '%s' to play role type '%s' cannot be undefined because the ability is inherited from a supertype.");
```
Phillammon, line 422: Major change of this request is making attempting to undefine nonexistent graph edges throw instead of no-op. This necessitated these new exceptions for attempting to undefine inherited relate/own/play edges...

common/exception/ErrorMessage.java
```diff
@@ -414,12 +414,24 @@ public static void loadConstants() {
...
+        public static final TypeWrite INVALID_UNDEFINE_NONEXISTENT_RELATES =
+                new TypeWrite(41, "The ability of type '%s' to relate role type '%s' cannot be undefined because because it does not have that ability.");
+        public static final TypeWrite INVALID_UNDEFINE_NONEXISTENT_OWNS =
+                new TypeWrite(42, "The ability of type '%s' to own attribute type '%s' cannot be undefined because it does not have that ability.");
+        public static final TypeWrite INVALID_UNDEFINE_NONEXISTENT_PLAYS =
+                new TypeWrite(43, "The ability of type '%s' to play role type '%s' cannot be undefined because it does not have that ability.");
```
Phillammon, line 428: ...and these three for when the two nodes are wholly unconnected.

concept/type/impl/ThingTypeImpl.java
```diff
@@ -153,8 +157,13 @@ public void unsetOwns(AttributeType attributeType) {
...
             throw exception(GraknException.of(INVALID_UNDEFINE_OWNS_HAS_INSTANCES, vertex.label(), attVertex.label()));
         }
-        if ((edge = vertex.outs().edge(OWNS, attVertex)) != null) edge.delete();
         if ((edge = vertex.outs().edge(OWNS_KEY, attVertex)) != null) edge.delete();
+        else if ((edge = vertex.outs().edge(OWNS, attVertex)) != null) edge.delete();
+        else if (this.getOwns().anyMatch(attr -> attr.equals(attributeType)))
+            throw exception(GraknException.of(INVALID_UNDEFINE_INHERITED_OWNS,
+                    this.getLabel().toString(), attributeType.getLabel().toString()));
+        else throw exception(GraknException.of(INVALID_UNDEFINE_NONEXISTENT_OWNS,
+                    this.getLabel().toString(), attributeType.getLabel().toString()));
```
Phillammon, line 166: instead of taking no action when the edge is not found, we throw one of the two previously defined exceptions- inherited if inherited, nonexistent if unconnected.

concept/type/impl/ThingTypeImpl.java
```diff
@@ -321,7 +330,13 @@ public void setPlays(RoleType roleType, RoleType overriddenType) {
...
     public void unsetPlays(RoleType roleType) {
         final TypeEdge edge = vertex.outs().edge(Encoding.Edge.Type.PLAYS, ((RoleTypeImpl) roleType).vertex);
-        if (edge == null) return;
+        if (edge == null) {
+            if (this.getPlays().anyMatch(attr -> attr.equals(roleType)))
+                throw exception(GraknException.of(INVALID_UNDEFINE_INHERITED_PLAYS,
+                        this.getLabel().toString(), roleType.getLabel().toString()));
+            else throw exception(GraknException.of(INVALID_UNDEFINE_NONEXISTENT_PLAYS,
+                    this.getLabel().toString(), roleType.getLabel().toString()));
+        }
```
Phillammon, line 339: Likewise, when undefining plays that aren't present, we find out which situation we are in and throw as appropriate.

query/Undefiner.java
```diff
@@ -167,14 +167,13 @@ private void undefineSub(ThingType thingType, SubConstraint subConstraint) {
...
-            } else if (thingType.getSupertypes().noneMatch(t -> t.equals(supertype))
-                    && !(supertype instanceof ThingTypeImpl.Root)) {
+            } else if (thingType.getSupertypes().noneMatch(t -> t.equals(supertype))) {
                 throw GraknException.of(INVALID_UNDEFINE_SUB, thingType.getLabel(), supertype.getLabel());
             }
             if (thingType instanceof RelationType) {
                 variables.stream().filter(
                         v -> v.label().isPresent() && v.label().get().scope().isPresent() &&
-                                v.label().get().scope().get().equals(thingType.getLabel().scope().get())
+                                v.label().get().scope().get().equals(thingType.getLabel().toString())
```
Phillammon, line 176: This was causing a rocksdb error- we are asking if the scope of the roletype is the same as the type of the relation we are looking at- that is, if the role is scoped to this relation. Relations do not themselves have scope, hence this change.

query/Undefiner.java
```diff
@@ -167,14 +167,13 @@ private void undefineSub(ThingType thingType, SubConstraint subConstraint) {
             final ThingType supertype = getThingType(subConstraint.type().label().get());
             if (supertype == null) {
                 throw GraknException.of(TYPE_NOT_FOUND, subConstraint.type().label().get());
-            } else if (thingType.getSupertypes().noneMatch(t -> t.equals(supertype))
-                    && !(supertype instanceof ThingTypeImpl.Root)) {
+            } else if (thingType.getSupertypes().noneMatch(t -> t.equals(supertype))) {
```
Phillammon, line 170: This line was leftover from when getSupertypes did not return the root thing type. It now does, so the callout is no longer needed.

query/Undefiner.java
```diff
@@ -225,8 +224,8 @@ private void undefineOwns(ThingType thingType, Set<OwnsConstraint> ownsConstrain
                                             owns.attribute().label().get());
                 } else if (owns.isKey()) {
                     throw GraknException.of(INVALID_UNDEFINE_OWNS_KEY,
-                                            owns.attribute().label().get(),
-                                            owns.attribute().label().get());
+                            owns.attribute().label().get(),
+                            owns.attribute().label().get());
```
Phillammon, line 228: When autoformatting this file, found out these were too far out.

test/behaviour/graql/language/undefine/BUILD
```diff
@@ -34,6 +34,7 @@ java_test(
         "//test/behaviour/connection:steps",
         "//test/behaviour/connection/session:steps",
         "//test/behaviour/config:parameters",
+        "//test/behaviour/logic/rule:steps",
```
Phillammon, line 37: Specific rule steps have now been introduced, and undefineTest needs them.

test/behaviour/logic/rule/BUILD
```diff
@@ -0,0 +1,49 @@
...
+
+        # External Maven Dependencies
+        "@maven//:junit_junit",
+        "@maven//:io_cucumber_cucumber_java",
+    ],
+    native_libraries_deps = [
+        "//concept:concept",
+        "//logic:logic",
+    ]
+)
```
Phillammon, line 49: Standard BUILD boilerplate for steps

test/behaviour/logic/rule/RuleSteps.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Then("rules do not contain")
+    public void rules_do_not_contain(List<String> ruleLabels) {
+        Set<String> actuals = tx().logic().rules().map(Rule::getLabel).toSet();
+        for (String ruleLabel : ruleLabels) {
+            assertFalse(actuals.contains(ruleLabel));
+        }
+    }
+}
+
```
Phillammon, line 57: And the actual rules steps. Again, should be self-explanatory.

concept/type/impl/ThingTypeImpl.java
haikalpribadi: if it's multiline, we prefer to use the curly brackets, @Phillammon

haikalpribadi: and are you sure the exception constructor don't fit one line (up to 120 char width) ?

concept/type/impl/ThingTypeImpl.java
haikalpribadi: Same as above: formatting of brackets and line breaks

query/Undefiner.java
haikalpribadi: why call `toString()` here? that's a bit weird. `toString()` doesn't describe the property of the object, but rather a string representation of the object. This should return `.name()` ?

haikalpribadi (changes requested):
almost there

haikalpribadi (changes requested):
Also need to enable Undefine BDD test in CI

haikalpribadi (approved):

PR #6064, opened 2021-01-22
Improve performance of data transactions

lriuui0x0:

We improved two aspects of data transaction behaviour. First, we removed unnecessary locking when the schema graph is in "read-only" mode. Second, we improved the performance of commit validation by reducing unnecessary stream parallelisation.

- Cached schema graph doesn't need to be locked for read-write access, so we remove unnecessary locks.
- Remove parallel stream that makes commit validation slow.

### DIFF
--- a/graph/SchemaGraph.java
+++ b/graph/SchemaGraph.java
@@ -48,7 +48,7 @@ import java.util.stream.Stream;
 import static grakn.common.collection.Collections.list;
 import static grakn.common.collection.Collections.pair;
 import static grakn.core.common.exception.ErrorMessage.SchemaGraph.INVALID_SCHEMA_WRITE;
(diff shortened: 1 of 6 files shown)


graph/SchemaGraph.java
```diff
@@ -286,16 +289,20 @@ public TypeVertex getType(String label, @Nullable String scope) {
...
+            }
         }
     }
 
     public RuleStructure getRule(String label) {
         assert storage.isOpen();
         try {
-            multiLabelLock.lockRead();
-            singleLabelLocks.computeIfAbsent(label, x -> new ManagedReadWriteLock()).lockRead();
+            if (!isReadOnly) {
```
lriuui0x0, line 302: No need to lock when we're in readonly mode.

Issue #6066, opened 2021-01-22
Producer sometimes crashes on hasNext()


maxbaxt:
## Description

Grakn reaches an illegal state when calling `hasNext()` on producer

## Environment

Grakn 2.0.2-alpha

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. checkout this branch: https://github.com/maxbaxt/grakn/tree/non-determistic-hasNext-exception
2. Go to `TypeResolverTest`
3. run the last test **in debug mode:**: `type_resolver_on_big_schema`

## Expected Output

runs the test without issue

## Actual Output

Throws an error because it reaches an illegal state when it calls `hasNext()` on `Producer`.

Issue #6061, opened 2021-01-21
GraphProducer: Unfulfilled request was not zero when job iterator still has next


haikalpribadi:
## Description

This seems to be a flakey bug that is hard to reproduce. You can find the occurrence in these Grabl jobs: 

https://grabl.io/graknlabs/grakn/f14b5fe6302402aea53f2adeef2006247d6ae8ae/build/1/correctness/1/test-behaviour/1
https://grabl.io/graknlabs/grakn/267a213e0acd8b07c57527e458b89a3ee88f2245/build/1/correctness/1/test-behaviour/1

## Environment

1. OS (where Grakn server runs): Grabl `graknlabs-ubuntu-20.04`
2. Grakn version (and platform): Grakn Core https://github.com/graknlabs/grakn/commit/f14b5fe6302402aea53f2adeef2006247d6ae8ae https://github.com/graknlabs/grakn/commit/267a213e0acd8b07c57527e458b89a3ee88f2245
3. Grakn client: embedded Grakn through Cucumber
4. Other environment details: NA

## Reproducible Steps

https://grabl.io/graknlabs/grakn/f14b5fe6302402aea53f2adeef2006247d6ae8ae/build/1/correctness/1/test-behaviour/1

```
1) the maximum value for a particular variable within each answer group can be retrieved using a group 'max'(Graql Get Clause)
grakn.core.common.exception.GraknException: java.lang.AssertionError
at grakn.core.common.exception.GraknException.of(GraknException.java:51)
at grakn.core.common.producer.ProducerIterator.hasNext(ProducerIterator.java:93)
at java.base/java.util.Iterator.forEachRemaining(Iterator.java:132)
at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
at grakn.core.query.Matcher$Group$Aggregator.execute(Matcher.java:594)
at grakn.core.query.QueryManager.match(QueryManager.java:143)
[...]

PR #6067, opened 2021-01-22
Fix a concurrency bug in async producer which may result in illegal state exception

lriuui0x0:

Fix a concurrency bug in async producer which may result in illegal state exception.

Currently there's check in the async job on if `runningJobs` contains the iterator. It's possible that this job is started before we put the future into the job. Therefore we always add `CompletableFuture` to the `runningJobs` before we start the job.

Fix https://github.com/graknlabs/grakn/issues/6066

Fix https://github.com/graknlabs/grakn/issues/6061

### DIFF
--- a/traversal/producer/GraphProducer.java
+++ b/traversal/producer/GraphProducer.java
@@ -113,10 +113,11 @@ public class GraphProducer implements Producer<VertexMap> {
     private synchronized void compensate(Queue<VertexMap> queue, int unfulfilled) {
         ResourceIterator<VertexMap> newIter =
                 new GraphIterator(graphMgr, start.next(), procedure, params, filter).distinct(produced);
-        CompletableFuture<Void> asyncJob = unfulfilled > 0
-                ? CompletableFuture.runAsync(() -> job(queue, newIter, unfulfilled), forkJoinPool())
-                : CompletableFuture.completedFuture(null);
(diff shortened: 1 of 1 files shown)


traversal/producer/GraphProducer.java
```diff
@@ -113,10 +113,11 @@ private synchronized void transition(Queue<VertexMap> queue, ResourceIterator<Ve
     private synchronized void compensate(Queue<VertexMap> queue, int unfulfilled) {
         ResourceIterator<VertexMap> newIter =
                 new GraphIterator(graphMgr, start.next(), procedure, params, filter).distinct(produced);
-        CompletableFuture<Void> asyncJob = unfulfilled > 0
-                ? CompletableFuture.runAsync(() -> job(queue, newIter, unfulfilled), forkJoinPool())
-                : CompletableFuture.completedFuture(null);
-        runningJobs.put(newIter, asyncJob);
+        runningJobs.put(newIter, CompletableFuture.completedFuture(null));
```
lriuui0x0, line 116: Currently there's check in the async job on if `runningJobs` contains the iterator. It's possible that this job is started before we put the future into the job. Therefore we always add `CompletableFuture` to the `runningJobs` before we start the job.

haikalpribadi (approved):

PR #6068, opened 2021-01-22
Run reasoner in read transactions when enabled in query options

flyingsilverfin:
To enable reasoner from certain tests, un-ignore the reasoning code paths, and add a boolean to enable or disable them 
instead. The default, for now, is to disable reasoner, as specified in query options.

We also disable reasoning in data write transactions; support for this will be addressed by the following issue in a future release of Grakn: github.com/graknlabs/grakn/issues/6076.

* Add boolean `reasoner` flag in `Options` methods to enable or disable executing traversals in the reasoning engine.
* Log warning and disable reasoner in any write transactions
* Remove `Options.Transaction` from Transaction interface, and expose `Context.Transaction` instead

### DIFF
--- a/query/Matcher.java
+++ b/query/Matcher.java
@@ -98,8 +98,8 @@ public class Matcher {
         return new Group.Aggregator(group, query);
     }
 
-    public ResourceIterator<ConceptMap> execute(boolean isParallel, boolean reasoning) {
-        ResourceIterator<ConceptMap> answers = reasoner.execute(disjunction, filter, isParallel, reasoning);
+    public ResourceIterator<ConceptMap> execute() {
+        ResourceIterator<ConceptMap> answers = reasoner.execute(disjunction, filter, options);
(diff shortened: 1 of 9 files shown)


query/Matcher.java
```diff
@@ -98,8 +98,8 @@ public static Matcher create(Reasoner reasoner, GraqlMatch query, Options.Query
         return new Group.Aggregator(group, query);
     }
 
-    public ResourceIterator<ConceptMap> execute(boolean isParallel) {
-        ResourceIterator<ConceptMap> answers = reasoner.execute(disjunction, filter, isParallel);
+    public ResourceIterator<ConceptMap> execute(boolean isParallel, boolean reasoning) {
```
haikalpribadi, line 101: perhaps this is better called `isReasoning` ?

haikalpribadi (approved):

PR #6073, opened 2021-01-22
Fixes to insert behaviour

Phillammon:

In upgrading BDD tests to cover insert behaviour as it should appear in grakn 2.0, a number of inconsistencies were revealed. We resolved these inconsistencies.

- Added matched concepts to the return value of insert queries
- Enabled extending existing relations with more roleplayers

### DIFF
--- a/query/Inserter.java
+++ b/query/Inserter.java
@@ -129,6 +129,7 @@ public class Inserter {
             else throw GraknException.of(THING_ISA_MISSING, ref);
             if (ref.isName()) inserted.put(ref.asName(), thing);
             if (!var.has().isEmpty()) insertHas(thing, var.has());
+            if (!var.relation().isEmpty()) extendRelation((Relation) thing, var);
             return thing;
         }
     }
@@ -154,12 +155,16 @@ public class Inserter {
     }
 
     private RoleType getRoleType(TypeVariable var) {
+        return getRoleType(var, null);
+    }
+
+    private RoleType getRoleType(TypeVariable var, RelationType relation) {
         try (ThreadTrace ignored = traceOnThread(TRACE_PREFIX + "get_role_type")) {
             assert var.reference().isLabel() && var.label().isPresent();
-            final RelationType relationType;
-            final RoleType roleType;
-            if ((relationType = conceptMgr.getRelationType(var.label().get().scope().get())) != null &&
-                    (roleType = relationType.getRelates(var.label().get().label())) != null) {
+                final RelationType relationType = relation != null ?
+                        relation : conceptMgr.getRelationType(var.label().get().scope().get());
+                final RoleType roleType;
+                if ((relationType != null && (roleType = relationType.getRelates(var.label().get().label())) != null)) {
                 return roleType;
             } else {
                 throw GraknException.of(TYPE_NOT_FOUND, var.label().get().scopedLabel());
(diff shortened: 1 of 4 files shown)


haikalpribadi:
Unfortunately, the insert query now breaks Match Query BDD tests, as they also use insert queries, @Phillammon. So we'll have to address those first.

query/Inserter.java
```diff
@@ -222,17 +229,27 @@ private Attribute insertAttribute(AttributeType attributeType, ThingVariable var
...
                 var.relation().iterator().next().players().forEach(rolePlayer -> {
                     final RoleType roleType;
                     final Thing player = insert(rolePlayer.player());
                     final Set<RoleType> inferred;
                     if (rolePlayer.roleType().isPresent()) {
-                        roleType = getRoleType(rolePlayer.roleType().get());
+                        roleType = getRoleType(rolePlayer.roleType().get(), relation.getType());
                     } else if ((inferred = player.getType().getPlays()
-                            .filter(rt -> rt.getRelationType().equals(relationType))
+                            .filter(rt -> rt.getRelationType().equals(relation.getType()))
```
haikalpribadi, line 252: What's the difference between `relationType` and `relation.getType()` ? 🤔

query/Inserter.java
```diff
@@ -222,17 +229,27 @@ private Attribute insertAttribute(AttributeType attributeType, ThingVariable var
     }
 
     private Relation insertRelation(RelationType relationType, ThingVariable var) {
+        Relation relation = relationType.create();
+        return extendRelation(relation, var);
```
haikalpribadi, line 233: Are you sure we should call `extendRelation()` here and then call it again at the end of the main `insert()` method? This feels like a bug to me.

query/Inserter.java
```diff
@@ -87,48 +87,51 @@ public static Inserter create(ConceptManager conceptMgr, List<graql.lang.pattern
...
+            if (matchedContains(var)) {
+                thing = matchedGet(var);
                 if (var.isa().isPresent() && !thing.getType().equals(getThingType(var.isa().get().type()))) {
                     throw GraknException.of(THING_ISA_REINSERTION, ref, var.isa().get().type());
                 }
             } else if (var.isa().isPresent()) thing = insertIsa(var.isa().get(), var);
             else throw GraknException.of(THING_ISA_MISSING, ref);
             if (ref.isName()) inserted.put(ref.asName(), thing);
             if (!var.has().isEmpty()) insertHas(thing, var.has());
+            if (!var.relation().isEmpty()) extendRelation((Relation) thing, var);
```
haikalpribadi, line 134: We're calling this twice now. Once already in the `insertRelation()`

query/Inserter.java
```diff
@@ -87,48 +87,51 @@ public static Inserter create(ConceptManager conceptMgr, List<graql.lang.pattern
...
-        return existing.get(var.reference().asName()).asThing();
+    public Thing matchedGet(ThingVariable var) {
+        return matched.get(var.reference().asName()).asThing();
     }
 
     private Thing insert(ThingVariable var) {
         try (ThreadTrace ignored = traceOnThread(TRACE_PREFIX + "insert")) {
-            Thing thing; Reference ref = var.reference();
+            Thing thing;
+            Reference ref = var.reference();
```
haikalpribadi, line 119: This could remain one line.

haikalpribadi (changes requested):
I think I see a few potential bugs, but also this PR breaks Match Query BDD now. So we'll have to address them first.

haikalpribadi (approved):

Issue #6072, opened 2021-01-22
RPC prefetch is too aggressive for Insert queries


flyingsilverfin:
## Problem to Solve
1. When doing insert queries, or match-insert queries, we do not want to initiate precomputation (the number of prefetch we compute and send, based on the batch size and latency)
2. For all queries, we should not use the time delta we currently use (latency = time from user sending request, to first answer), as it can include either query planning time or the time to `match` all answers, then execute `insert` or `delete` on each of them. Please double check that our understanding is correct on this @alexjpwalker.

## Proposed Solution
1. for insert queries, we should initiate sending batches of answers (especially because it costs us nothing to actually compute them - they are already computed!) when the user requests the first batch, so in total there will be two round trips to initiate this. We can then continue with latency compensation and batching

2. For latency compensation, we should compensate the normal latency, not including query-planning, and/or the time to perform match-insert, or match-delete.

In essence, we want to not overly prefetch answer to the user, as apparently sending too many grpc messages has significant CPU time overhead. A clear win is to not pre-send any answers for insert queries, as outlined in 1. above.

flyingsilverfin:
Here's the hack i used to test this with Grami:
```
        <T> void beginIteration(TransactionProto.Transaction.Req request, Iterator<T> iterator, int batchSize, Function<List<T>, TransactionProto.Transaction.Res> responseBuilderFn) {
            QueryProto.Query.Req req = request.getQueryReq();
            final String requestId = request.getId();
            BatchingIterator<T> batchingIterator;
            if (req.getReqCase() == INSERT_REQ) {
                batchingIterator = new BatchingIterator<>(requestId, iterator, responseBuilderFn, 1, 0);
            } else {
                final int latencyMillis = request.getLatencyMillis();
                batchingIterator = new BatchingIterator<>(requestId, iterator, responseBuilderFn, batchSize, latencyMillis);
            }
            iterators.compute(requestId, (key, oldValue) -> {
                if (oldValue == null) return batchingIterator;
                else throw GraknException.of(DUPLICATE_REQUEST, requestId);
            });
            batchingIterator.iterateBatch();
        }

```

probably worth reproducing we do make significant performance gains for scenario 1.

alexjpwalker:
[...]

Issue #6075, opened 2021-01-25
Iterators and the BatchingIterator logic are hard to read


alexjpwalker:
## Problem to Solve

The Iterators and BatchingIterator logic are getting bloated and hard to read.

## Proposed Solution

Refactor the BatchingIterator methods. Split them up into submethods where applicable. Iterators may not even be the best name anymore.

alexjpwalker:
The server architecture was refactored, and this was cleaned up by the creation of `TransactionService.ResponseStream<T>`.

PR #6077, opened 2021-01-25
Add prefetch option for queries, which defaults to false for inserts

alexjpwalker:

On executing a Graql query, Grakn automatically streams the first batch of responses back to the client. This ensures they get their answers to a `match` query as fast as possible. But, for an `insert` query, they usually don't need those answers. And the CPU cost of compiling these unwanted response messages was inflicting a significant performance penalty. To remedy this, we made `prefetch` a configurable per-query option, which defaults to true, except for Insert queries where it defaults to false.

- Don't prefetch responses to Insert queries unless specified with a configurable flag in QueryOptions (fixes https://github.com/graknlabs/grakn/issues/6072)

### DIFF
--- a/common/parameters/Options.java
+++ b/common/parameters/Options.java
@@ -19,6 +19,8 @@ package grakn.core.common.parameters;
 
 import grakn.core.common.exception.GraknException;
 
+import java.util.Optional;
+
 import static grakn.core.common.exception.ErrorMessage.Internal.ILLEGAL_ARGUMENT;
 
 public abstract class Options<PARENT extends Options<?, ?>, SELF extends Options<?, ?>> {
@@ -33,6 +35,7 @@ public abstract class Options<PARENT extends Options<?, ?>, SELF extends Options
     private Boolean infer = null;
     private Boolean explain = null;
     private Integer batchSize = null;
+    private Optional<Boolean> prefetch = null;
     private Integer sessionIdlTimeoutMillis = null;
     private Integer schemaLockAcquireTimeoutMillis = null;
 
@@ -73,6 +76,15 @@ public abstract class Options<PARENT extends Options<?, ?>, SELF extends Options
 
     public SELF batchSize(int batchSize) {
         this.batchSize = batchSize;
+        return getThis();
+    }
+
+    public Optional<Boolean> prefetch() {
+        return prefetch;
+    }
+
+    public SELF prefetch(boolean prefetch) {
+        this.prefetch = Optional.of(prefetch);
         return getThis();
     }
 
--- a/common/parameters/Options.java
+++ b/common/parameters/Options.java
@@ -35,7 +35,7 @@ public abstract class Options<PARENT extends Options<?, ?>, SELF extends Options
     private Boolean infer = null;
     private Boolean explain = null;
     private Integer batchSize = null;
-    private Optional<Boolean> prefetch = null;
+    private Boolean prefetch = null;
     private Integer sessionIdlTimeoutMillis = null;
     private Integer schemaLockAcquireTimeoutMillis = null;
 
@@ -79,12 +79,12 @@ public abstract class Options<PARENT extends Options<?, ?>, SELF extends Options
         return getThis();
     }
 
-    public Optional<Boolean> prefetch() {
+    public Boolean prefetch() {
         return prefetch;
     }
 
     public SELF prefetch(boolean prefetch) {
-        this.prefetch = Optional.of(prefetch);
+        this.prefetch = prefetch;
         return getThis();
     }
 
--- a/dependencies/graknlabs/repositories.bzl
+++ b/dependencies/graknlabs/repositories.bzl
@@ -42,8 +42,8 @@ def graknlabs_graql():
 def graknlabs_protocol():
     git_repository(
         name = "graknlabs_protocol",
-        remote = "https://github.com/graknlabs/protocol",
-        tag = "2.0.0-alpha-6", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_protocol
+        remote = "https://github.com/alexjpwalker/protocol",
+        commit = "e5f1a1046c2696d7d7301f5688d9e33e7a38283e", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_protocol
     )
 
 def graknlabs_grabl_tracing():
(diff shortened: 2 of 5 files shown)


common/parameters/Options.java
```diff
@@ -33,6 +35,7 @@
     private Boolean infer = null;
     private Boolean explain = null;
     private Integer batchSize = null;
+    private Optional<Boolean> prefetch = null;
```
alexjpwalker, line 38: Unlike our other `Options`, `null` is considered a valid value for `prefetch`, and it means "whatever the default for this Graql query type is."

dependencies/graknlabs/repositories.bzl
```diff
@@ -42,8 +42,8 @@ def graknlabs_graql():
 def graknlabs_protocol():
     git_repository(
         name = "graknlabs_protocol",
-        remote = "https://github.com/graknlabs/protocol",
-        tag = "2.0.0-alpha-6", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_protocol
+        remote = "https://github.com/alexjpwalker/protocol",
+        commit = "e5f1a1046c2696d7d7301f5688d9e33e7a38283e", # sync-marker: do not remove this comment, this is used for sync-dependencies by @graknlabs_protocol
```
alexjpwalker, line 46: (https://github.com/graknlabs/protocol/pull/112) Update to graknlabs before merging.

server/rpc/TransactionRPC.java
```diff
@@ -185,25 +186,35 @@ void closeWithError(Throwable error) {
...
             final String requestId = request.getId();
             final int latencyMillis = request.getLatencyMillis();
             final BatchingIterator<T> batchingIterator = new BatchingIterator<>(requestId, iterator, responseBuilderFn, batchSize, latencyMillis);
             iterators.compute(requestId, (key, oldValue) -> {
                 if (oldValue == null) return batchingIterator;
                 else throw GraknException.of(DUPLICATE_REQUEST, requestId);
             });
-            batchingIterator.iterateBatch();
+            if (prefetch) batchingIterator.iterateBatch();
+            else respond(continueRes(requestId));
```
alexjpwalker, line 213: `continueRes` is a message that conceptually just reads "CONTINUE". It signifies that if the client wants any (more) answers, they need to ask for them.

common/parameters/Options.java
flyingsilverfin: the only comment: this is why we have `Boolean __ = null`, we don't store optionals :)

haikalpribadi: Isn't that the semantics of every other field too? 🤔

server/rpc/TransactionRPC.java
```diff
@@ -185,25 +186,35 @@ void closeWithError(Throwable error) {
         private final ConcurrentMap<String, BatchingIterator<?>> iterators = new ConcurrentHashMap<>();
 
         /**
-         * Spin up an iterator and begin batch iterating.
+         * Spin up a new iterator and begin streaming responses immediately.
          */
-        <T> void beginIteration(TransactionProto.Transaction.Req request, Iterator<T> iterator, Function<List<T>, TransactionProto.Transaction.Res> responseBuilderFn) {
-            beginIteration(request, iterator, transaction.options().batchSize(), responseBuilderFn);
+        <T> void createNewIterator(TransactionProto.Transaction.Req request, Iterator<T> iterator, Function<List<T>, TransactionProto.Transaction.Res> responseBuilderFn) {
```
haikalpribadi, line 191: `createNewIterator()` implies the construction of a new object and thus returning it. This method does quite the opposite.

server/rpc/query/QueryHandler.java
```diff
@@ -90,8 +90,9 @@ public void handleRequest(Transaction.Req request) {
     }
 
     private void match(Transaction.Req request, QueryProto.Query.Match.Req req, Options.Query options) {
-        GraqlMatch query = Graql.parseQuery(req.getQuery()).asMatch();
-        ResourceIterator<ConceptMap> answers = queryManager.match(query, options);
+        if (!options.prefetch().isPresent()) options.prefetch(true);
+        final GraqlMatch query = Graql.parseQuery(req.getQuery()).asMatch();
+        final ResourceIterator<ConceptMap> answers = queryManager.match(query, options);
```
haikalpribadi, line 95: Can we not bring back `final` for local variables?

common/parameters/Options.java
alexjpwalker: What I mean is it behaves differently to the others. The others all have DEFAULT values specified in this file. `prefetch` doesn't have a default - it defaults to whatever the default is for the Graql query that's being executed.

server/rpc/query/QueryHandler.java
```diff
@@ -90,6 +90,7 @@ public void handleRequest(Transaction.Req request) {
     }
 
     private void match(Transaction.Req request, QueryProto.Query.Match.Req req, Options.Query options) {
+        if (!options.prefetch().isPresent()) options.prefetch(true);
```
haikalpribadi, line 93: I feel like these default values should not be defined "scattered" across this class. I feel like they should be aggregated somewhere as a set of enums or constants of some sorts, which sits in a class that holds the relevant domain knowledge.

common/parameters/Options.java
haikalpribadi: I don't think this should be captured with optional

haikalpribadi (changes requested):
Left revision in separate comments

server/rpc/TransactionRPC.java
alexjpwalker: +1. In fact I think the purpose of these Iterators has shifted quite considerably from when they were first created. I've raised an issue here: https://github.com/graknlabs/grakn/issues/6075.

I can't think of a better name really. `beginIteration` was misleading because the 2nd overload _doesn't_ begin iteration if it's an `insert` query.

server/rpc/query/QueryHandler.java
alexjpwalker: updated

server/rpc/TransactionRPC.java
haikalpribadi: `iterate(...)` would be an appropriate enough name here.

alexjpwalker: Updated

PR #6078, opened 2021-01-25
Add allow secondary replica option for Grakn Cluster

lolski:

We have added a new option: allow secondary replica which is used by Grakn Cluster. Additionally, we have opened up `RocksSession` for subclassing.

1. Add `allowSecondaryReplica` option in the `Options` class.
2. Allow subclassing `RocksSession.Schema` and `RocksSession.Data`:
    - Make the classes non-final
    - Open up a few fields to be accessible by subclasses

### DIFF
--- a/common/parameters/Options.java
+++ b/common/parameters/Options.java
@@ -33,6 +33,7 @@ public abstract class Options<PARENT extends Options<?, ?>, SELF extends Options
     public static final boolean DEFAULT_PARALLEL = true;
     public static final boolean DEFAULT_QUERY_READ_PREFETCH = true;
     public static final boolean DEFAULT_QUERY_WRITE_PREFETCH = false;
+    public static final boolean DEFAULT_PRIMARY_REPLICA = true;
 
     private PARENT parent;
     private Boolean infer = null;
@@ -40,6 +41,7 @@ public abstract class Options<PARENT extends Options<?, ?>, SELF extends Options
     private Integer batchSize = null;
     private Integer sessionIdlTimeoutMillis = null;
     private Integer schemaLockAcquireTimeoutMillis = null;
+    private Boolean primaryReplica = null;
 
     protected Boolean prefetch = null;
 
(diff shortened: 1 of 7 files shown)


rocks/Factory.java
```diff
@@ -48,6 +48,8 @@
 
     interface TransactionSchema {
 
+        RocksTransaction.Schema transactionSchemaInit(RocksSession.Schema session);
```
lolski, line 51: We need transactionSchemaInit in order to provide the hook for cluster to supply an additional logic with regards to schema initialisation.

common/parameters/Options.java
```diff
@@ -99,6 +101,17 @@ public SELF schemaLockAcquireTimeoutMillis(int acquireSchemaLockTimeoutMillis) {
         return getThis();
     }
 
+    public boolean primaryReplica() {
```
lolski, line 104: I'm adding the replica options here since I cannot subclass the `Options` class in Grakn Cluster. This isn't ideal though and I'm open to ideas on how to solve the issue.

rocks/RocksFactory.java
```diff
@@ -80,8 +80,17 @@ public RocksDatabase databaseLoadAndOpen(RocksGrakn grakn, String name) {
 
     private synchronized Factory.TransactionSchema transactionSchemaFactory() {
         if (transactionSchemaFactory == null) {
-            transactionSchemaFactory = (session, type, options) ->
-                    new RocksTransaction.Schema(session, type, options, storageFactory());
+            transactionSchemaFactory = new TransactionSchema() {
+                @Override
+                public RocksTransaction.Schema transactionSchemaInit(RocksSession.Schema session) {
+                    throw new UnsupportedOperationException();
```
lolski, line 86: TODO: tidy up this particular loose end

rocks/RocksTransaction.java
```diff
@@ -52,7 +52,7 @@
     protected ConceptManager conceptMgr;
     protected AtomicBoolean isOpen;
     TraversalEngine traversalEng;
-    LogicManager logicMgr;
```
haikalpribadi, line 55: You should have re-ordered this @lolski

PR #6081, opened 2021-01-25
Type resolution correctly handles explicitly labelled roles

maxbaxt:

`TypeResolver` can now handle `TypeVariable` within a query. We no longer use the `Variable` method `addTo()`. This will fix a bug where explicitly labelled `roleTypes` were not dealt with properly. The test have also been improved, so that it checks which `resolvedTypes` are added to Anonymous and Labels Variables as well as Named Variables.

* `TypeResolver` now adds to the `Traversal` directly
* This ensures the correct `resolvers` are added to the `traversal`, which is important when the `resolvers` are `SystemReference`
* Changed Tests so that all variables are checked, not just named ones
* Added test to ensure resolved Types on labelled roles are further pruned by the full `resolve()` function.

### DIFF
--- a/logic/tool/TypeResolver.java
+++ b/logic/tool/TypeResolver.java
@@ -38,6 +38,7 @@ import grakn.core.pattern.constraint.type.PlaysConstraint;
 import grakn.core.pattern.constraint.type.RegexConstraint;
 import grakn.core.pattern.constraint.type.RelatesConstraint;
 import grakn.core.pattern.constraint.type.SubConstraint;
+import grakn.core.pattern.constraint.type.TypeConstraint;
 import grakn.core.pattern.constraint.type.ValueTypeConstraint;
 import grakn.core.pattern.variable.SystemReference;
 import grakn.core.pattern.variable.ThingVariable;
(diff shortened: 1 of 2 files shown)


logic/tool/TypeResolver.java
```diff
@@ -186,51 +236,51 @@ private TypeVariable register(ThingVariable var) {
...
-            var.is().forEach(constraint -> registerIs(resolver, constraint));
+            var.is().forEach(constraint -> registerIsThing(resolver, constraint));
             var.has().forEach(constraint -> registerHas(resolver, constraint));
             var.relation().forEach(constraint -> registerRelation(resolver, constraint));
             var.iid().ifPresent(constraint -> registerIID(resolver, constraint));
             return resolver;
         }
 
-        private void registerIID(TypeVariable owner, IIDConstraint iidConstraint) {
+        private void registerIID(TypeVariable resolver, IIDConstraint iidConstraint) {
```
maxbaxt, line 246: Decided to make `resolver` used more consistently.

logic/tool/TypeResolver.java
```diff
@@ -170,10 +178,52 @@ private TypeVariable register(TypeVariable var) {
             }
             resolverRegister.put(var.id(), resolver);
             variableRegister.putIfAbsent(resolver.reference(), var);
+            if (!var.resolvedTypes().isEmpty()) traversal.labels(resolver.id(), var.resolvedTypes());
```
maxbaxt, line 181: `ResolveLabels` already dealt with `LabelConstraints`, so we just need to pass over the results.

logic/tool/TypeResolver.java
```diff
@@ -170,10 +178,50 @@ private TypeVariable register(TypeVariable var) {
...
-            resolver.addTo(traversal);
+            if (!var.resolvedTypes().isEmpty()) traversal.labels(resolver.id(), var.resolvedTypes());
+            resolver.abstractConstraint().ifPresent(constraint -> registerAbstract(resolver));
+            resolver.is().forEach(constraint -> registerIsType(resolver, constraint));
+            resolver.owns().forEach(constraint -> registerOwns(resolver, constraint));
+            resolver.plays().forEach(constraint -> registerPlays(resolver, constraint));
+            resolver.regex().ifPresent(constraint -> registerRegex(resolver, constraint));
+            resolver.relates().forEach(constraint -> registerRelates(resolver, constraint));
+            resolver.sub().ifPresent(constraint -> registerSub(resolver, constraint));
+            resolver.valueType().ifPresent(constraint -> registerValueType(resolver, constraint));
```
flyingsilverfin, line 189: two and a half things:
1. shouldn't this be operating over the constraints of `var`, not `resolver`?
1a. if so, do you have any tests that would have caught this mistake?
2. if the order is not important here, I think its less error prone to do
```var.constraints().forEach(constraint -> {
if (constraint.isLabel()) ...
else if (constraint.isAbstract()) ...
```
because this would mean at the end you can throw an exception in the `else`, which means you know if you've accidentally forgotten to handle a constraint, or if we add a new one?

flyingsilverfin: 3. do we have any tests that show why this catches a bug?

logic/tool/TypeResolver.java
flyingsilverfin: great :)

flyingsilverfin (changes requested):

logic/tool/TypeResolver.java
maxbaxt: 1. Yes good point. And good point there isn't a test for that...
2. Could do. Doesn't feel as pretty somehow. But I see your point about safely catching all the constraints.

maxbaxt: So it turns out the tests weren't checking the anonymous variables. I've fixed that, and now also put a test in to ensure that the `resolve()` function can further prune the `resolvedLabels` on a `role`.

test/integration/logic/TypeResolverTest.java
```diff
@@ -925,9 +970,33 @@ public void infer_key_attributes() {
...
+
+        Set<String> expectedLabels = set("partnership:partner", "business:partner");
+
+        assertEquals(expectedLabels, getHintMap(conjunction).get("$_relation:partner"));
+
+        typeResolver.resolve(conjunction);
+        Set<String> expectedResolvedTypes = set("partnership:partner");
+
+        assertEquals(expectedResolvedTypes, getHintMap(conjunction).get("$_relation:partner"));
+    }
```
flyingsilverfin, line 1001: @maxbaxt  this is for reduction of the set of roles. However I think there should also be a test that uses `match $role type relation:friend; ($role: $x) isa $y;` or the like.

flyingsilverfin:
@maxbaxt  I've modified your title to remove technical terms (class name).
Also, please add one more tests like the one I mentioned in the above comment :)
I'm already approving, but plz do that, then wait for tests and we good

flyingsilverfin (approved):

PR #6095, opened 2021-01-27
Added timeout exception when opening session or transaction

### LINKED ISSUE #6065: Attempting to open a transaction should time out after 10 seconds, not stall indefinitely
## Problem to Solve

Currently, if a user (say, a database admin) keeps a schema session open for a few minutes to perform some maintenance, any other client attempting to connect to the database and write some data will just stall indefinitely.

## Current Workaround

Sit there looking confused.

## Proposed Solution

When attempting to open a transaction, we have a blocking call that waits for a lock to become available. It currently blocks indefinitely. Instead, it should time out after 10 seconds.

Phillammon:

Previously, attempting to open a data transaction while a schema write transaction was open would cause the data transaction to hang indefinitely with no indication of why. We added a timeout to this process, so that if the timeout is met, the transaction opening throws in a way that is propogated to the client, with slightly more indication of what's going on.

Resolves #6065 

- Added a timeout, taken from the transaction options, to attempting to acquire a read lock when opening a schema transaction.
- Added a new Session exception


rocks/RocksSession.java
```diff
@@ -26,12 +26,12 @@
 import grakn.core.concurrent.lock.ManagedLock;
 
 import java.util.UUID;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.*;
```
haikalpribadi, line 29: Fix this

rocks/RocksSession.java
```diff
@@ -179,7 +179,12 @@ boolean isData() {
         @Override
         public RocksTransaction.Data transaction(Arguments.Transaction.Type type, Options.Transaction options) {
             if (!isOpen.get()) throw GraknException.of(SESSION_CLOSED);
-            long lock = database().dataWriteSchemaLock().readLock();
+            long lock;
+            try {
+                lock = database().dataWriteSchemaLock().tryReadLock(options.schemaLockAcquireTimeoutMillis(), TimeUnit.MILLISECONDS);
```
haikalpribadi, line 184: make sure less than 120 char width

common/exception/ErrorMessage.java
```diff
@@ -131,6 +131,8 @@ public static void loadConstants() {
                 new Session(1, "Session with UUID '%s' does not exist.");
         public static final Session SESSION_CLOSED =
                 new Session(2, "Attempted to open a transaction from closed session.");
+        public static final Session ACQUIRE_LOCK_TIMEOUT =
```
haikalpribadi, line 134: this should be in Transaction group

haikalpribadi (changes requested):

flyingsilverfin (approved):

haikalpribadi (approved):

PR #6099, opened 2021-01-27
Optimised key ownership validation at commit time

flyingsilverfin:
We make key validation much cheaper, when no keys can be present. In our previous implementation of concept validation, we checked whether required key attributes are owned - however in practice we accidentally scanned ALL attributes owned if _no_ attributes are owned as keys! This PR changes the implementation to only check for keys if there are key-attributes.

* Only scan owned attributes by keyed type, if the attribute is required to own any keys.


haikalpribadi (approved):
Great catch, @flyingsilverfin 👍

PR #6100, opened 2021-01-27
Migrate and update reasoning behaviour tests from 1.8

flyingsilverfin:
We implement the required steps to run reasoning BDD (called "resolution" features for now). We also bring in and update all the test and BUILD files required to execute reasoning BDD scenarios.

* Migrate the `//test/behaviour/resolution` package from core 1.8
* Update `ResolutionSteps` to the 2.0 behaviour syntax for opening and closing transactions, sessions, etc.
* Fix alpha-equivalence bug in `Concludable`, and the alpha-equivalence mapping ordering in `ResolverRegistry`
* Fix materialisation bug where an identifier could be assigned the same IID twice


logic/Rule.java
```diff
@@ -399,10 +399,14 @@ public Relation asRelation() {
...
                 traversal.isa(relationId, Identifier.Variable.label(relationType.getLabel().name()), false);
+                Set<Identifier.Variable> playersWithIds = new HashSet<>();
                 players.forEach(rp -> {
                     // note: NON-transitive role player types - we require an exact role being played
                     traversal.rolePlayer(relationId, rp.playerIdentifier, set(rp.roleType.getLabel()), rp.repetition);
-                    traversal.iid(rp.playerIdentifier, rp.player.getIID());
+                    if (!playersWithIds.contains(rp.playerIdentifier)) {
+                        traversal.iid(rp.playerIdentifier, rp.player.getIID());
+                        playersWithIds.add(rp.playerIdentifier);
+                    }
```
flyingsilverfin, line 409: this fixes a bug where if we had duplicate role player `(actor: $x, director: $x)`, we tried to assign an IID for `$x` twice, which is not allowed. we now we only add an ID for any role player once

jmsfltchr (approved):

PR #6101, opened 2021-01-27
Reasoner resolvers handle old exhausted messages correctly

flyingsilverfin:
To prevent a class of bugs to do with reiteration (which are difficult to test), we make sure we don't let previous iterations' `Exhausted` event loop messages pollute more advanced iteration states. One visible consequence is when a reasoning query reports it is finished, and we close resources, we no longer see exceptions due to ongoing resolver work.

* The ReasonerProducer only retries requests from previous iterations that failed in new iterations, if it hasn't already completely terminated
* Each resolver receiving an out of date response will immediately forward it upstream, and not allow it to affect new iteration state

### DIFF
--- a/reasoner/ReasonerProducer.java
+++ b/reasoner/ReasonerProducer.java
@@ -69,25 +69,25 @@ public class ReasonerProducer implements Producer<ConceptMap> {
     }
 
     private void requestFailed(int iteration) {
-        LOG.trace("New iteration {}", iteration);
-        assert this.iteration == iteration || this.iteration == iteration + 1;
+        LOG.trace("Failed to find answer to request in iteration: " + iteration);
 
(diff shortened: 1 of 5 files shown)


jmsfltchr (approved):
Woot!

reasoner/ReasonerProducer.java
```diff
@@ -69,25 +70,25 @@ private void requestAnswered(ResolutionAnswer answer) {
     }
 
     private void requestFailed(int iteration) {
```
jmsfltchr, line 72: Actually, I would prefer that this method be called `requestExhausted`. Otherwise we suddenly switch terminology and I don't see any reason to!

jmsfltchr (changes requested):

PR #6104, opened 2021-01-28
Run reasoning exclusively when rules are present

flyingsilverfin:
To avoid having to deduplicate answers from traversal engine and reasoner, we choose to either only execute a query in the traversal engine, or only in reasoner. The switch for this is whether there are any rules present. Further, the root reasoning resolver only creates child resolvers if any rules are applicable to the query it contains.

* Either return a producer from the traversal, or a producer from reasoner when executing a query, depending on whether there are any rules present
* `RootResolver` only creates child resolvers if there are any applicable rules to any of the built concludables
* Throw an exception when trying to set inference on/off per-query (must be at transaction level, because we materialise answers in the graph. If the user runs a reasoning query, then disables reasoner, they will still see inferred answers! This is unexpected UX.)
* Rules no longer contain concludables directly, however, `RuleResolver` builds concludables for the rule's `when` conjunction lazily
* Migrate some `RuleTest` tests to `ConcludableTest`

### DIFF
--- a/reasoner/Reasoner.java
+++ b/reasoner/Reasoner.java
@@ -38,11 +38,9 @@ import grakn.core.reasoner.resolution.ResolutionRecorder;
 import grakn.core.reasoner.resolution.ResolverRegistry;
 import grakn.core.traversal.TraversalEngine;
 import grakn.core.traversal.common.Identifier;
-import grakn.core.traversal.common.VertexMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
 import java.util.List;
 
 import static grakn.common.collection.Collections.list;
@@ -51,7 +49,6 @@ import static grakn.core.common.exception.ErrorMessage.Pattern.UNSATISFIABLE_CON
 import static grakn.core.common.exception.ErrorMessage.ThingRead.CONTRADICTORY_BOUND_VARIABLE;
 import static grakn.core.common.iterator.Iterators.iterate;
 import static grakn.core.common.iterator.Iterators.link;
-import static grakn.core.common.iterator.Iterators.single;
 import static grakn.core.common.parameters.Arguments.Query.Producer.EXHAUSTIVE;
 import static grakn.core.concurrent.common.ExecutorService.PARALLELISATION_FACTOR;
 import static grakn.core.concurrent.producer.Producers.produce;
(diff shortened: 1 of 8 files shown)


reasoner/Reasoner.java
```diff
@@ -80,51 +77,49 @@ public Reasoner(ConceptManager conceptMgr, LogicManager logicMgr,
...
+            else {
+                producer = traversalEng.producer(conj.traversal(filter), context.producer(), PARALLELISATION_FACTOR)
+                        .map(conceptMgr::conceptMap);
+            }
         } else if (!filter.isEmpty() && iterate(filter).anyMatch(id -> conj.variable(id).isThing()) ||
                 iterate(conjunction.variables()).anyMatch(Variable::isThing)) {
             throw GraknException.of(UNSATISFIABLE_CONJUNCTION, conjunction);
         } else {
-            return single(Producers.empty());
+            return Producers.empty();
```
jmsfltchr, line 110: The fact that this is `Producers`, plural, says again to me we need a better name 🤔

reasoner/Reasoner.java
```diff
@@ -133,13 +128,13 @@ else return iterate(producers).map(prod -> prod.filter(ans -> !iterate(conjuncti
...
-        if (!isInfer(context)) LOG.warn("Reasoning is disabled in write transactions");
+        if (!isInfer(context)) LOG.info("Reasoning will not run");
 
         ResourceIterator<ConceptMap> answers;
         Conjunction conj = logicMgr.typeResolver().resolve(conjunction);
         if (conj.isSatisfiable()) {
-            answers = traversalEng.iterator(conjunction.traversal(filter)).map(conceptMgr::conceptMap);
-            if (isInfer(context)) answers = link(answers, produce(resolve(conj), context.producer()));
+            if (!isInfer(context)) answers = produce(resolve(conj), context.producer());
+            else answers = traversalEng.iterator(conjunction.traversal(filter)).map(conceptMgr::conceptMap);
```
jmsfltchr, line 137: ✅ 😄

reasoner/resolution/resolver/RootResolver.java
```diff
@@ -169,16 +170,16 @@ protected void receiveExhausted(Response.Exhausted fromDownstream, int iteration
...
+        if (!concludables.isEmpty()) {
+            Set<Retrievable> retrievables = Retrievable.extractFrom(conjunction, concludables);
+            Set<Resolvable> resolvables = new HashSet<>();
+            resolvables.addAll(concludables);
+            resolvables.addAll(retrievables);
+            plan.addAll(planner.plan(resolvables));
+            iterate(plan).forEachRemaining(resolvable -> {
+                downstreamResolvers.put(resolvable, registry.registerResolvable(resolvable));
+            });
+        }
```
jmsfltchr, line 182: We should use the same strategy in `RuleResolver`!

flyingsilverfin: done!

jmsfltchr (approved):

reasoner/Reasoner.java
```diff
@@ -133,13 +128,13 @@ else return iterate(producers).map(prod -> prod.filter(ans -> !iterate(conjuncti
 
     private ResourceIterator<ConceptMap> iterator(Conjunction conjunction, List<Identifier.Variable.Name> filter,
                                                   Context.Query context) {
-        if (!isInfer(context)) LOG.warn("Reasoning is disabled in write transactions");
+        if (!isInfer(context)) LOG.info("Reasoning will not run");
```
haikalpribadi, line 131: Does this really need to be `info`? 🤔 I really don't think so...

haikalpribadi (approved):

PR #6106, opened 2021-01-29
Correct traversal backtracking through relation scopes

flyingsilverfin:
We encountered a query plans where answers go missing, when doing optimised backtracking from a scoped role back to the relation that scopes the role instance or role player edge. This PR corrects backtracking optimisations to avoid missing anwers that are valid.

* Implement a priority queue associated with each relation that can be a scoper of role instances
* when backtracking from a scoped role, we only backtrack to the last role `branch` edge, rather than the edge that generates the relation itself. This means we "free" up roles to be re-visited by later iterators, rather than missing them completely

### DIFF
--- a/traversal/iterator/GraphIterator.java
+++ b/traversal/iterator/GraphIterator.java
@@ -163,12 +163,13 @@ public class GraphIterator extends AbstractResourceIterator<VertexMap> {
             if (edge.onlyStartsFromRelation()) {
                 seekStack.addSeeks(iterate(scopedEdgeOrders.get(edge.from().id().asVariable())).toSet());
                 seekStack.addSeeks(edge.to().dependedEdgeOrders());
-            }
-            if (edge.onlyEndsAtRelation()) {
+            } else if (edge.onlyEndsAtRelation()) {
                 seekStack.addSeeks(edge.from().dependedEdgeOrders());
                 seekStack.addSeeks(iterate(scopedEdgeOrders.get(edge.to().id().asVariable())).toSet());
+            } else {
+                seekStack.addSeeks(edge.to().dependedEdgeOrders());
+                seekStack.addSeeks(edge.from().dependedEdgeOrders());
             }
-
             return false;
         }
     }
--- a/traversal/iterator/GraphIterator.java
+++ b/traversal/iterator/GraphIterator.java
@@ -54,7 +54,7 @@ public class GraphIterator extends AbstractResourceIterator<VertexMap> {
     private final List<Identifier.Variable.Name> filter;
     private final Map<Identifier, ResourceIterator<? extends Vertex<?, ?>>> iterators;
     private final Map<Identifier, Vertex<?, ?>> answer;
-    private final Map<Identifier.Variable, Set<ThingVertex>> scoped;
+    private final Map<Identifier.Variable, Set<ThingVertex>> scopedRoles;
     private final Map<Identifier.Variable, PriorityQueue<Integer>> scopedEdgeOrders;
     private final Map<Identifier, ThingVertex> roles;
     private final SeekStack seekStack;
@@ -73,7 +73,7 @@ public class GraphIterator extends AbstractResourceIterator<VertexMap> {
         this.filter = filter;
         this.edgeCount = procedure.edgesCount();
         this.iterators = new HashMap<>();
-        this.scoped = new HashMap<>();
+        this.scopedRoles = new HashMap<>();
         this.scopedEdgeOrders = new HashMap<>();
         this.roles = new HashMap<>();
         this.answer = new HashMap<>();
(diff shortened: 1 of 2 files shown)


traversal/iterator/GraphIterator.java
```diff
@@ -122,7 +128,12 @@ private boolean computeFirstBranch(int pos) {
                     else {
                         backTrackCleanUp(pos);
                         answer.remove(toID);
-                        seekStack.addSeeks(edge.from().dependedEdgeOrders());
+                        if (edge.onlyStartsFromRelation()) {
+                            assert fromId.isVariable();
+                            seekStack.addSeeks(iterate(scopedEdgeOrders.get(fromId.asVariable())).toSet());
```
haikalpribadi, line 133: change SeekStack.addSeeks() to take in collection and we don't need to double iterate

traversal/iterator/GraphIterator.java
```diff
@@ -145,8 +161,18 @@ private boolean computeFirstClosure(int pos) {
...
+                seekStack.addSeeks(iterate(scopedEdgeOrders.get(edge.from().id().asVariable())).toSet());
+                seekStack.addSeeks(edge.to().dependedEdgeOrders());
+            } else if (edge.onlyEndsAtRelation()) {
+                assert edge.from().id().isVariable();
+                seekStack.addSeeks(edge.from().dependedEdgeOrders());
+                seekStack.addSeeks(iterate(scopedEdgeOrders.get(edge.to().id().asVariable())).toSet());
+            } else {
+                seekStack.addSeeks(edge.to().dependedEdgeOrders());
+                seekStack.addSeeks(edge.from().dependedEdgeOrders());
+            }
```
haikalpribadi, line 175: This should go into a method

traversal/iterator/GraphIterator.java
```diff
@@ -207,8 +233,12 @@ private boolean computeNextBranch(int pos) {
...
                 newIter = branch(fromVertex, edge);
                 if (!newIter.hasNext()) {
-                    assert !edge.from().ins().isEmpty();
-                    computeNextSeekPos = edge.from().branchEdge().order();
+                    if (edge.onlyStartsFromRelation() && !scopedEdgeOrders.get(edge.from().id().asVariable()).isEmpty()) {
+                        Integer lastRoleGenerator = scopedEdgeOrders.get(edge.from().id().asVariable()).peek();
+                        computeNextSeekPos = lastRoleGenerator;
+                    } else if (!edge.from().ins().isEmpty()) {
+                        computeNextSeekPos = edge.from().branchEdge().order();
+                    }
```
haikalpribadi, line 241: else should be set to 0

haikalpribadi: and assert that it's actually the start vertex

traversal/iterator/GraphIterator.java
```diff
@@ -221,8 +251,14 @@ private boolean computeNextBranch(int pos) {
 
     private boolean isClosure(ProcedureEdge<?, ?> edge, Vertex<?, ?> fromVertex, Vertex<?, ?> toVertex) {
         if (edge.isRolePlayer()) {
-            Set<ThingVertex> withinScope = scoped.computeIfAbsent(edge.asRolePlayer().scope(), id -> new HashSet<>());
-            return edge.asRolePlayer().isClosure(graphMgr, fromVertex, toVertex, params, withinScope);
+            Set<ThingVertex> withinScope = scopedRoles.computeIfAbsent(edge.asRolePlayer().scope(), id -> new HashSet<>());
+            PriorityQueue<Integer> scopedEdgeOrder = scopedEdgeOrders.computeIfAbsent(edge.asRolePlayer().scope(), id -> new PriorityQueue<>(Comparator.reverseOrder()));
+            boolean closed = edge.asRolePlayer().isClosure(graphMgr, fromVertex, toVertex, params, withinScope);
```
haikalpribadi, line 256: var rename to isClosure

traversal/iterator/GraphIterator.java
```diff
@@ -52,7 +55,8 @@
     private final List<Identifier.Variable.Name> filter;
     private final Map<Identifier, ResourceIterator<? extends Vertex<?, ?>>> iterators;
     private final Map<Identifier, Vertex<?, ?>> answer;
-    private final Map<Identifier.Variable, Set<ThingVertex>> scoped;
+    private final Map<Identifier.Variable, Set<ThingVertex>> scopedRoles;
+    private final Map<Identifier.Variable, PriorityQueue<Integer>> scopedEdgeOrders;
```
haikalpribadi, line 59: I think this should be wrapped in a dedicated class now, which provides one API that manages both datasets synchronously under-the-hood.

traversal/iterator/GraphIterator.java
```diff
@@ -231,23 +267,31 @@ private boolean isClosure(ProcedureEdge<?, ?> edge, Vertex<?, ?> fromVertex, Ver
...
+            Set<ThingVertex> withinScope = scopedRoles.computeIfAbsent(scope, id -> new HashSet<>());
+            PriorityQueue<Integer> scopedEdgeOrder = scopedEdgeOrders.computeIfAbsent(scope, id -> new PriorityQueue<>(Comparator.reverseOrder()));
             toIter = edge.asRolePlayer().branchEdge(graphMgr, fromVertex, params).filter(e -> {
                 if (withinScope.contains(e.optimised().get())) return false;
                 else {
-                    removePreviousScopedRole(edge.asRolePlayer().scope(), edge.to().id());
+                    removePreviousScopedRole(scope, edge.to().id(), edge);
                     withinScope.add(e.optimised().get());
+                    assert !scopedEdgeOrder.contains(edge.order());
+                    scopedEdgeOrder.add(edge.order());
```
haikalpribadi, line 294: I dont understand why we're removing and adding the same edge

haikalpribadi (changes requested):

haikalpribadi (approved):

PR #6107, opened 2021-01-29
Resolution and reiteration test

jmsfltchr:

Reasoner resolution relies on a message-passing actor model internally. In order to test this we have integration tests that check that the messages sent and received are as expected. We upgrade these tests (which were mocked) to be compatible with the now functional reasoner. We also create a thorough test case for reasoner reiteration, which is required in order to run Nth passes over materialised answers in order to find the full and correct set of answers for a query.

- Upgrade `ResolutionTest`.
- Split our reiteration test and upgrade.
- The ReiterationTest is ignored until we fix a bug.


reasoner/resolution/framework/ResponseProducer.java
```diff
@@ -20,14 +20,16 @@
...
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Set;
 
 public class ResponseProducer {
     private final Set<ConceptMap> produced;
     private final Iterator<ConceptMap> traversalProducer;
-    private final Set<Request> downstreamProducer;
+    private final List<Request> downstreamProducer;
```
flyingsilverfin, line 32: put an note: list helps preserve determinism

reasoner/resolution/resolver/ConcludableResolver.java
```diff
@@ -50,7 +51,7 @@
 public class ConcludableResolver extends ResolvableResolver<ConcludableResolver> {
     private static final Logger LOG = LoggerFactory.getLogger(ConcludableResolver.class);
 
-    private final Map<Actor<RuleResolver>, Set<Unifier>> applicableRules;
+    private final LinkedHashMap<Actor<RuleResolver>, Set<Unifier>> applicableRules;
```
flyingsilverfin, line 54: put an note: linked map helps preserve determinism

flyingsilverfin (approved):

flyingsilverfin:
good stuff!

haikalpribadi (approved):

PR #6111, opened 2021-01-29
Throw exception correctly on transaction lock timeout

Phillammon:

When a transaction fails to acquire a write lock, presently, it throws `ERROR: Error: 13 INTERNAL: null Please check server logs for the stack trace.` when a commit is attempted, which is incorrect. We made it instead throw  `ERROR: Error: 13 INTERNAL: [TXN13] Invalid Transaction Operation: Could not acquire lock for data transaction. A schema session may have been left open. Please check server logs for the stack trace.` immediately, which is much more informative.

Check the return value from trying to acquire a lock and correctly throw an exception if the result is 0, meaning the attempt timed out.

### DIFF
--- a/rocks/RocksDatabase.java
+++ b/rocks/RocksDatabase.java
@@ -151,7 +151,7 @@ public class RocksDatabase implements Grakn.Database {
                 lock = dataWriteSchemaLock().tryWriteLock(options.schemaLockAcquireTimeoutMillis(), TimeUnit.MILLISECONDS);
                 if (lock == 0) throw GraknException.of(SCHEMA_ACQUIRE_LOCK_TIMEOUT);
             } catch (InterruptedException e) {
-                throw GraknException.of(SCHEMA_ACQUIRE_LOCK_TIMEOUT);
+                throw GraknException.of(e);
             }
             session = sessionFactory.sessionSchema(this, options);
         } else if (type.isData()) {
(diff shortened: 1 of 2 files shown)


rocks/RocksDatabase.java
```diff
@@ -149,6 +149,7 @@ RocksSession createAndOpenSession(Arguments.Session.Type type, Options.Session o
         if (type.isSchema()) {
             try {
                 lock = dataWriteSchemaLock().tryWriteLock(options.schemaLockAcquireTimeoutMillis(), TimeUnit.MILLISECONDS);
+                if (lock == 0) throw GraknException.of(SCHEMA_ACQUIRE_LOCK_TIMEOUT);
             } catch (InterruptedException e) {
                 throw GraknException.of(SCHEMA_ACQUIRE_LOCK_TIMEOUT);
```
haikalpribadi, line 154: This should be replaced with `throw GraknException.of(e)`

rocks/RocksSession.java
```diff
@@ -185,6 +185,7 @@ boolean isData() {
             try {
                 int timeout = options.schemaLockAcquireTimeoutMillis();
                 lock = database().dataWriteSchemaLock().tryReadLock(timeout, TimeUnit.MILLISECONDS);
+                if (lock == 0) throw GraknException.of(DATA_ACQUIRE_LOCK_TIMEOUT);
             } catch (InterruptedException err) {
                 throw GraknException.of(DATA_ACQUIRE_LOCK_TIMEOUT);
```
haikalpribadi, line 190: This should be replaced with `throw GraknException.of(e)`

haikalpribadi (changes requested):

haikalpribadi (approved):

Issue #6115, opened 2021-02-01
Conjunctions should own constraints, which own variables


flyingsilverfin:
## Problem to Solve
Currently, the ownership structure looks like:
`Conjunction`, owns `Variable`, which owns sets of `Constraint`.

However, the equality of `Conjunction`, depends on the equality of `Variable`. To prevent recursive equality checks however, `Variable` equality is strictly dependent on the identifier on the variable, rather than the `Constraint`s that make it up; `Constraint` equality depends on all the `Variable`s inside the constraint being equal.

This breaks in unexpected places, such as putting different Conjunctions into a set, can eliminate some of them if they have the same `Variable` identifiers, but different constraints!

## Proposed Solution
We flip the ownership structure:
`Conjunction` owns `Constraint`s which own `Variable`. This will make the equality functions work naturally.

Further, in reasoner, we will be able to create specific types of `Conjunction` eg. a `Concludable.Relation`, which _only_ permits certain types of constraints, allowing us to create specific shapes of conjunctions with relative ease.

haikalpribadi:
So we need this for production, @flyingsilverfin ? You placed it in `production` project board, btw.

flyingsilverfin:
I think it would be good, since it avoids common mistakes with the `equals()` in conjunctions... see my open PR https://github.com/graknlabs/grakn/pull/6116

flyingsilverfin:
Implementing this should allow Caching Type Inference outcomes from the conjunction directly, without having to build a traversal first... if we want this because we'll lose the many-to-one caching behaviour (multiple queries end up with the same type inference query)

PR #6116, opened 2021-02-01
Prevent losing disjunctions' conjunctions

flyingsilverfin:
We currently do an unintuitive equality check for Conjunctions, depending only on the identifier of the variables - regardles of constraints, which can lead to losing "equal" conjunctions. To prevent deduplicating conjunctions we create in a disjunction, we store them in a list rather than a set. This can be correctly implemented after is https://github.com/graknlabs/grakn/issues/6115 is complete.

* Store conjunctions in a list, within a disjunction
* Bump `@graknlabs_behaviour`, which contains a test that fails for this case without this fix.

### DIFF
--- a/dependencies/graknlabs/repositories.bzl
+++ b/dependencies/graknlabs/repositories.bzl
@@ -56,5 +56,5 @@ def graknlabs_behaviour():
     git_repository(
         name = "graknlabs_behaviour",
         remote = "https://github.com/graknlabs/behaviour",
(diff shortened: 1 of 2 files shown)


haikalpribadi:
This feels like a dangerous hack that could lead to other bugs, doesn't it, @flyingsilverfin ? 🤔 Should we just take on #6115 straight on?

haikalpribadi (approved):

PR #6121, opened 2021-02-02
Simplify and correct traversal scoping and unscoping

flyingsilverfin:

Diamond-shaped relation queries led to incorrectly scoped role players, which in turn led to missing answers. To correct this, we implement a new stack-like data structure to handle role player scoping within the graph traversal, replacing a more complicated map-based structure.

* replace `roles` map, which failed in diamond-shaped queries, and change the `Scopes` data structures to behave more like a stack
* using the stack-like datastructures within each scope (eg. each relation), we can simply pop or push onto the `Scoped` structure when we recurse forward or backtrack, respectively


traversal/iterator/GraphIterator.java
```diff
@@ -268,19 +265,24 @@ private boolean isClosure(ProcedureEdge<?, ?> edge, Vertex<?, ?> fromVertex, Ver
...
+                    if (scoped.orderVisited(edge.order())) scoped.replace(edge.order(), role.asThing());
+                    else scoped.push(role.asThing(), edge.order());
                     return true;
                 }
             });
+        } else if (edge.from().id().isScoped()) {
+            Identifier.Variable scope = edge.from().id().asScoped().scope();
+            Scopes.Scoped scoped = scopes.getOrInitialise(scope);
+            scoped.push(fromVertex.asThing(), edge.order());
+            toIter = edge.branch(graphMgr, fromVertex, params);
```
flyingsilverfin, line 277: we were actually forgetting to add the `from` vertex to a `scope`! This is a bug, when the start vertex is scoped to a relation that is the endpoint of an edge coming out of the start vertex.

traversal/iterator/GraphIterator.java
```diff
@@ -296,30 +298,22 @@ private boolean isClosure(ProcedureEdge<?, ?> edge, Vertex<?, ?> fromVertex, Ver
...
+        if (edge.to().id().isScoped()) {
+            Identifier.Variable scope = edge.to().id().asScoped().scope();
+            if (scopes.get(scope).orderVisited(pos)) scopes.get(scope).popLast();
+        } else if (edge.from().id().isScoped()) {
+            Identifier.Variable scope = edge.from().id().asScoped().scope();
+            assert scopes.get(scope).orderVisited(pos);
+            scopes.get(scope).popLast();
+        } else if (edge.isRolePlayer()) {
+            Identifier.Variable scope = edge.asRolePlayer().scope();
+            if (scopes.get(scope).orderVisited(pos)) scopes.get(scope).popLast();
```
flyingsilverfin, line 316: backtracking mirrors the `branch()` flow, where we scope roles depending on whether it is the `to()`, `from()` or shortcut edge that is scoped.

traversal/iterator/GraphIterator.java
```diff
@@ -360,48 +354,55 @@ public Scoped get(Identifier.Variable scope) {
 
         public static class Scoped {
 
-            private final Set<ThingVertex> roleInstances;
-            private final PriorityQueue<Integer> edgeOrders;
+            Set<ThingVertex> roles;
+            TreeMap<Integer, ThingVertex> visited;
```
flyingsilverfin, line 358: a `TreeMap` can tell us the _last_ element in the collection. Since the `Integer` is the edge order, and we always walk the edges in order, getting the last element is equivalent to getting the last added element!

Deleting the last element it corresponds to popping from a stack, and adding to the treemap should only be allowed when the new element being added is greater than the last greater one, under the use case we have in the graph iterator

flyingsilverfin: the `set` is to allow fast lookups for what is contained in the scope already

traversal/iterator/GraphIterator.java
```diff
@@ -360,48 +354,55 @@ public Scoped get(Identifier.Variable scope) {
...
+                assert (isEmpty() || order > lastEdgeOrder()) && !roles.contains(role);
+                visited.put(order, role);
+                roles.add(role);
             }
 
-            public void remove(ThingVertex previousRole, int order) {
-                assert roleInstances.contains(previousRole) && edgeOrders.contains(order);
-                roleInstances.remove(previousRole);
-                edgeOrders.remove(order);
+            public void replace(int order, ThingVertex newRole) {
```
flyingsilverfin, line 395: i can't think of a fast way to "swap" out a role for the same order, all operations are essentially remove an element from a map and replace it. However, since we also have to retrieve the old value to delete it from the fast lookup set, we might as well just delete it

traversal/iterator/GraphIterator.java
```diff
@@ -73,7 +71,6 @@ public GraphIterator(GraphManager graphMgr, Vertex<?, ?> start, GraphProcedure p
         this.filter = filter;
         this.edgeCount = procedure.edgesCount();
         this.iterators = new HashMap<>();
-        this.roles = new HashMap<>();
```
flyingsilverfin, line 76: `roles` has been completely absorbed!

haikalpribadi (approved):

PR #6131, opened 2021-02-03
Fix concurrency bug causing null pointer exception while iterating edges

lriuui0x0:

Fix concurrency bug causing null pointer exception while iterating edges.

- Added a null check so that we don't falsely assume edges already exist while it's still being loaded. In case of the edges are null, returning an empty iterator is safe since it will be merged together with persisted edges that contain every edge.
- Fix edges map not updated atomically. 
- Fix https://github.com/graknlabs/grakn/issues/6118

### DIFF
--- a/graph/adjacency/impl/ThingAdjacencyImpl.java
+++ b/graph/adjacency/impl/ThingAdjacencyImpl.java
@@ -141,8 +141,8 @@ public abstract class ThingAdjacencyImpl implements ThingAdjacency {
 
         edges.compute(infixIID, (iid, edgesByOutIID) -> {
             if (edgesByOutIID == null) edgesByOutIID = new ConcurrentHashMap<>();
-            if (edgesByOutIID.containsKey(edge.outIID())) {
-                ThingEdge thingEdge = edgesByOutIID.get(edge.outIID());
+            ThingEdge thingEdge = edgesByOutIID.get(edge.outIID());
+            if (thingEdge != null) {
                 if (thingEdge.isInferred() && !edge.isInferred()) thingEdge.isInferred(false);
             } else {
                 edgesByOutIID.put(edge.outIID(), edge);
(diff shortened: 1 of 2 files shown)


graph/adjacency/impl/ThingAdjacencyImpl.java
```diff
@@ -95,7 +95,7 @@
             iids = newIIDs;
         }
 
-        return iterate(iids).flatMap(iid -> iterate(edges.get(iid).values()));
+        return iterate(iids).flatMap(iid -> iterate(edges.get(iid) != null ? edges.get(iid).values().iterator() : emptyIterator()));
```
lriuui0x0, line 98: Added a null check so that we don't falsely assume edges already exist while it's still being loaded. In case of the edges are null, returning an empty iterator is safe since it will be merged together with persisted edges that contain every edge.

graph/adjacency/impl/ThingAdjacencyImpl.java
```diff
@@ -139,13 +139,16 @@ private ThingEdgeImpl put(Encoding.Edge.Thing encoding, ThingEdgeImpl edge, IID[
...
         }
 
-        Map<EdgeIID.Thing, ThingEdge> edgesByOutIID = edges.computeIfAbsent(infixIID, iid -> new ConcurrentHashMap<>());
-        if (edgesByOutIID.containsKey(edge.outIID())) {
-            ThingEdge thingEdge = edgesByOutIID.get(edge.outIID());
-            if (thingEdge.isInferred() && !edge.isInferred()) thingEdge.isInferred(false);
-        } else {
-            edgesByOutIID.put(edge.outIID(), edge);
-        }
+        edges.compute(infixIID, (iid, edgesByOutIID) -> {
```
lriuui0x0, line 142: Fix edges map not updated atomically.

traversal/procedure/GraphProcedure.java
```diff
@@ -263,6 +264,11 @@ public GraphProcedure build() {
             return type;
         }
 
+        public ProcedureVertex.Thing setTypes(ProcedureVertex.Thing thing, Set<String> types) {
```
lriuui0x0, line 267: This procedure builder method can be used later.

haikalpribadi (approved):

PR #6137, opened 2021-02-04
Allow creation of "schema graph initialisation" transactions

alexjpwalker:

In Grakn Cluster, all transactions (without the `readAnyReplica` flag set) are verified against the replica state, and throw an error if the replica is not primary. This was causing an issue where a Cluster node would fail to restart correctly after crashing, because it would have to perform schema-graph initialisation again (in a transaction) and that would throw. We fix it by introducing a package-private method `initialisationTransaction` into `RocksTransaction.Schema`. It does essentially the same thing as `transaction`, but uses a different method from `TransactionSchemaFactory` - that distinction allows the two to behave differently in Cluster.

Add `initialisationTransaction` to `RocksTransaction.Schema`.


haikalpribadi (approved):

lolski (approved):

haikalpribadi (changes requested):
I see what you changed since the last time we discussed this. I like it. Simpler and makes sense. Change the name of `transactionGraphInit()` to `initialisationTransaction()` (for 2 methods) an we're good to go, @lolski @alexjpwalker !

haikalpribadi (approved):

PR #6149, opened 2021-02-08
Make insert deleter updater respect parallelisation options

lriuui0x0:

Fix a bug which caused insert, deleter and updater to always run in parallel mode even when we set the parallelisation context to false.

- Seperate out `executeParallel` and `executeSerial` method for inserter, deleter and updater based on the parallelisation context.


haikalpribadi (approved):
👍

PR #6158, opened 2021-02-10
Do not acquire schema lock from data read transaction

lriuui0x0:

Currently, we erroneously acquired schema lock for a data read transaction, but only unlock for data write transaction. This causes sometimes schema lock is acquired and never unlocked, hence causing the database to unable to open transactions.

- Acquire schema lock only when data transaction type is WRITE.

### DIFF
--- a/rocks/RocksSession.java
+++ b/rocks/RocksSession.java
@@ -203,13 +203,15 @@ public abstract class RocksSession implements Grakn.Session {
         @Override
         public RocksTransaction.Data transaction(Arguments.Transaction.Type type, Options.Transaction options) {
             if (!isOpen.get()) throw GraknException.of(SESSION_CLOSED);
-            long lock;
-            try {
-                int timeout = options.schemaLockTimeoutMillis();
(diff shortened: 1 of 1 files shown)


rocks/RocksSession.java
```diff
@@ -203,13 +203,15 @@ boolean isData() {
...
-            } catch (InterruptedException e) {
-                throw GraknException.of(e);
+            long lock = 0;
+            if (type == Arguments.Transaction.Type.WRITE) {
+                try {
+                    int timeout = options.schemaLockTimeoutMillis();
+                    lock = database().schemaLock().tryReadLock(timeout, MILLISECONDS);
+                    if (lock == 0) throw GraknException.of(DATA_ACQUIRE_LOCK_TIMEOUT);
+                } catch (InterruptedException e) {
+                    throw GraknException.of(e);
```
haikalpribadi, line 213: This should be using `ErrorMessage.Internal.UNEXPECTED_INTERRUPTION`.

haikalpribadi (approved):
Looks good. Just need a minor change, but approving first.

PR #6168, opened 2021-02-14
Traversals and reasoner return concepts for anonymous variables too

flyingsilverfin:
To enable reasoner to bind concepts to anonymous variables in conjunctions, we make a sweeping change that traversal engine and reasoner both return anonymous variables as well. Consequently, in Grakn now ConceptMaps contain `Identifer` rather than `References`, and can identify anonymous variables.

* Traversal and reasoner can bind to and return anonymous variable answers
* Create a new type `Identifier.Variable.Retrievable` which is extended by `Identifier.Variable.Named` and `Identifier.Variable.Anonymous`
* Propagate usage of `Retrievable` to unifications, mappings, concept maps, vertex maps as tightly as possible
* As a new optimisation, we check `Unifier.Requirements.Constraint` _forwards_ when unifiying a previous partial answer as well as backward when we find the complete answer. This required building unified and non-unified requirements up together to deal with non-retrieved identifiers' requirements nicely
* Rename the `type` requirement in Unifier.Requirements to `roleType`, which is the only place it should be used. Labeled types for `isa friendship` or `has name $x` become part of the `isaExplicit` requirements now
* fix bug with infinitely reiteration - `Root` resolvers do not clear their deduplication sets when reiterating
* unignore some tests and fix unification tests

### DIFF
--- a/concept/ConceptManager.java
+++ b/concept/ConceptManager.java
@@ -39,7 +39,7 @@ import grakn.core.graph.iid.VertexIID;
 import grakn.core.graph.vertex.ThingVertex;
 import grakn.core.graph.vertex.TypeVertex;
 import grakn.core.graph.vertex.Vertex;
-import grakn.core.traversal.common.Identifier.Variable.Retrieved;
+import grakn.core.traversal.common.Identifier.Variable.Retrievable;
 import grakn.core.traversal.common.VertexMap;
 
 import java.util.ArrayList;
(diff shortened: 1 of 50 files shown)


flyingsilverfin:
TODO: clean up `Unifier.Requirements` if we still need them - may still need to impose Instance requirements for named type vars -unifiedwith-> labeled vars?

haikalpribadi (approved):

PR #6171, opened 2021-02-15
Refactor partial answer flow to fix missing answers

jmsfltchr:

We fix several misbehaviours with an architectural cleanup of how we handle answer states. There were three issues all leading to missing answers. These are caused by incorrect termination criteria for reasoning, leading to premature termination.

- Reiteration was not triggered when queries contained more than one resolvable part, due to not propagating the state of whether an answer has been inferred between sibling actors in a conjunction.
- We should not reuse answers from a previous reasoning iteration we need to restart all answer searches or we will miss answers.
- Tracking the path that led to an answer was not enough to determine whether to terminate reasoning, instead the full history of the answer is needed, so we now use `AnswerState` to track this, each keeping a reference to its parent.


reasoner/resolution/framework/Request.java
```diff
@@ -107,78 +92,15 @@ public int hashCode() {
...
+                "sender=" + sender +
+                ", receiver=" + receiver +
+                ", partial=" + partialAnswer +
+                ", partialDerivation=" + partialAnswer.derivation() +
                 '}';
     }
 
-    public ResolutionAnswer.Derivation partialResolutions() {
-        return partialDerivation;
+    public AnswerState.Derivation partialResolutions() {
```
flyingsilverfin, line 102: rename to `partialDerivation()`

reasoner/resolution/resolver/ConcludableResolver.java
```diff
@@ -104,32 +100,15 @@ protected void receiveAnswer(Answer fromDownstream, int iteration) {
...
         } else {
             if (explanations()) {
-                ResolutionAnswer.Derivation derivation = new ResolutionAnswer.Derivation(map(pair(fromDownstream.sourceRequest().receiver(),
-                                                                                                  fromDownstream.answer())));
-                ResolutionAnswer deduplicated = new ResolutionAnswer(fromDownstream.answer().derived(), concludable.toString(),
-                                                                     derivation, self(), fromDownstream.answer().isInferred());
-                LOG.trace("{}: Recording deduplicated answer derivation: {}", name(), deduplicated);
-                resolutionRecorder.tell(actor -> actor.record(deduplicated));
+                LOG.trace("{}: Recording deduplicated answer derivation: {}", name(), upstreamAnswer);
+                resolutionRecorder.tell(actor -> actor.record(upstreamAnswer));
```
flyingsilverfin, line 111: noiceee

reasoner/resolution/resolver/ConjunctionResolver.java
```diff
@@ -228,18 +212,13 @@ private ResponseProducer mayUpdateAndGetResponseProducer(Request fromUpstream, i
     @Override
     protected ResponseProducer responseProducerCreate(Request fromUpstream, int iteration) {
         LOG.debug("{}: Creating a new ResponseProducer for request: {}", name(), fromUpstream);
-
-//        ResourceIterator<AnswerState.UpstreamVars.Derived> upstreamAnswers = toUpstreamAnswers(
-//                fromUpstream, compatibleBoundAnswers(conceptMgr, conjunction, fromUpstream.partialAnswer().conceptMap()));
-//
         Plans.Plan plan = plans.getOrCreate(fromUpstream.partialAnswer().conceptMap().concepts().keySet(), resolvables, negateds);
```
flyingsilverfin, line 215: while we're here:
arguments for doing traversals in conjunctions if there are no concludables or negations:
1. saves doing messages to the child `Retrievable`

2. saves saving all the deduplicated answers in both the `Retrievable` and the `Root.X` (root must have it anyway, I suppose we could disable it if there's no negation or concludable?). NOTE: don't need deduplication set in retrievable at all - traversal engine already guarantees distinctness!

flyingsilverfin (changes requested):

reasoner/resolution/resolver/ConjunctionResolver.java
```diff
@@ -313,30 +287,27 @@ public Nested(Actor<Nested> self, Conjunction conjunction, Actor<ResolutionRecor
...
         }
 
         @Override
-        protected ResourceIterator<AnswerState.UpstreamVars.Derived> toUpstreamAnswers(Request fromUpstream,
-                                                                                       ResourceIterator<ConceptMap> downstreamConceptMaps) {
-            return downstreamConceptMaps.map(conceptMap -> fromUpstream.partialAnswer().asIdentity()
-                    .aggregateToUpstream(conceptMap, null));
+        protected ResourceIterator<Partial<?>> toUpstreamAnswers(Request fromUpstream,
+                                                                 ResourceIterator<ConceptMap> downstreamConceptMaps) {
+            return downstreamConceptMaps.map(conceptMap -> fromUpstream.partialAnswer().asFiltered().toUpstream(self()));
```
flyingsilverfin, line 305: since conjunctions don't do traversals anymore, I dont think we need this nor its abstract parent at all :D

reasoner/resolution/answer/AnswerState.java
```diff
@@ -256,30 +466,76 @@ public String toString() {
             @Override
             public Unified asUnified() { return this; }
 
+            @Override
+            Unified with(ConceptMap conceptMap, Actor<? extends Resolver<?>> resolver, boolean requiresReiteration,
+                         @Nullable Derivation derivation) {
+                return new Unified(conceptMap, parent(), unifier, instanceRequirements, resolver, root(), requiresReiteration,
```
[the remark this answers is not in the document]
jmsfltchr: This is correct - it's a question of either doing the merging in the `.with()` calls or in the `toUpstream` family of methods. The latter is preferable as it's where we also do conceptMap merging.

flyingsilverfin (approved):

haikalpribadi (approved):

PR #6177, opened 2021-02-17
Make RocksDB (Debug) library the default

lolski:

We have made the `//rocks` package depends on RocksDB (Debug) library in order to increase debuggability during development.

However, there are two caveats:

1. In Bazel: RocksDB (Debug) is disabled for Linux until we fix the assertion tests, and for Windows until we deploy RocksDB (Debug) for Windows platform. Therefore, it is only active for Mac.
2. In the actual distribution, the `--debug` flag is active for Mac and Linux. For Windows, it will effectively be ignored until we deploy RocksDB (Debug) for the Windows platform. 

- Make `//rocks` package depend on RocksDB (Debug)
- Make the actual distribution contain both RocksDB (Debug) which is placed under `server/lib/dev` and RocksDB (Prod) which is placed under `server/lib/prod`

### DIFF
--- a/rocks/BUILD
+++ b/rocks/BUILD
@@ -58,7 +58,7 @@ native_java_libraries(
     linux_deps = [
         "@maven//:com_google_ortools_ortools_linux_x86_64",
         "@maven//:com_google_ortools_ortools_linux_x86_64_java",
-        "@maven//:org_rocksdb_rocksdbjni_dev_linux",
+        "@maven//:org_rocksdb_rocksdbjni",
     ],
     windows_deps = [
         "@maven//:com_google_ortools_ortools_win32_x86_64",
--- a/rocks/BUILD
+++ b/rocks/BUILD
@@ -40,8 +40,6 @@ native_java_libraries(
 
         # External dependencies from Maven
         "@maven//:com_google_code_findbugs_jsr305",
-        # "@maven//:org_rocksdb_rocksdbjni_dev", # Use this JAR for debugging RocksDB on Mac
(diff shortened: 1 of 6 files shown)


rocks/BUILD
```diff
@@ -55,14 +53,17 @@ native_java_libraries(
...
         "@maven//:com_google_ortools_ortools_linux_x86_64",
-        "@maven//:com_google_ortools_ortools_linux_x86_64_java"
+        "@maven//:com_google_ortools_ortools_linux_x86_64_java",
+        "@maven//:org_rocksdb_rocksdbjni_dev_linux",
     ],
     windows_deps = [
         "@maven//:com_google_ortools_ortools_win32_x86_64",
-        "@maven//:com_google_ortools_ortools_win32_x86_64_java"
+        "@maven//:com_google_ortools_ortools_win32_x86_64_java",
+        "@maven//:org_rocksdb_rocksdbjni",
```
lolski, line 66: We don't yet have RocksDB (Debug) for Windows. Therefore we substitute it with RocksDB (Prod) for the time being.

server/BUILD
```diff
@@ -161,37 +155,23 @@ java_deps(
...
     },
     visibility = ["//:__pkg__"],
     maven_name = True,
 )
 
 java_deps(
-    name = "server-deps-dev",
-    target = "@maven//:org_rocksdb_rocksdbjni_dev",
-    java_deps_root = "server/lib/dev/",
+    name = "server-deps-prod",
```
lolski, line 166: Add `server-deps-prod` to be added into the real distribution

PR #6185, opened 2021-02-18
Prevent open iterators from crashing RocksDB when tx closes

flyingsilverfin:

It's currently possible to crash RocksDB by closing open iterators that a class is currently using. Guarding against this requires synchronisation between state changes in `RocksIterator`, and guarding between closing storage and opening new iterators in RocksStorage.

* Guard state changes in RocksIterator with `synchronized` and remove need for `isOpen` atomic boolean
* Re-check that the storage is still open after obtaining an iterator, throwing an exception if not open anymore. This method means we don't need to use `synchronized` for this guard, which would be very expensive and contended.
* Lazily compute the next answer in `RocksIterator`, instead of right after consuming the last answer
* Because a transaction closing while iteration is till occuring is very common, we add a special case for the error logging in `GraphIterator` when it gets a `TRANSACTION_CLOSED` exception

### DIFF
--- a/rocks/RocksIterator.java
+++ b/rocks/RocksIterator.java
@@ -43,24 +43,6 @@ public final class RocksIterator<T> extends AbstractResourceIterator<T> implemen
         state = State.INIT;
     }
 
-    private synchronized boolean initialise() {
-        if (state != State.COMPLETED) {
-            this.internalRocksIterator = storage.getInternalRocksIterator();
-            this.internalRocksIterator.seek(prefix);
-            state = State.EMPTY;
-            byte[] key;
-            if (!internalRocksIterator.isValid() || !bytesHavePrefix(key = internalRocksIterator.key(), prefix)) {
-                recycle();
-                return false;
-            }
-            next = constructor.apply(key, internalRocksIterator.value());
-            state = State.FETCHED;
-            return true;
-        } else {
-            return false;
-        }
-    }
-
     public final T peek() {
         if (!hasNext()) throw new NoSuchElementException();
         return next;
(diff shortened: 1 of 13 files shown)


rocks/RocksIterator.java
```diff
@@ -42,35 +40,39 @@
...
+            state = State.EMPTY;
+            byte[] key;
+            if (!internalRocksIterator.isValid() || !bytesHavePrefix(key = internalRocksIterator.key(), prefix)) {
+                recycle();
+                return false;
+            }
+            next = constructor.apply(key, internalRocksIterator.value());
+            state = State.FETCHED;
+            return true;
+        } else {
```
haikalpribadi, line 59: Why are duplicating the code from `fetchAndCheck()` into `initialise()`, @flyingsilverfin ? I don't think this is a good idea.

rocks/RocksIterator.java
```diff
@@ -81,18 +83,26 @@ public final boolean hasNext() {
             case EMPTY:
                 return fetchAndCheck();
             case INIT:
-                initalise();
-                return fetchAndCheck();
+                return initialise();
```
haikalpribadi, line 86: I don't think we should merge `initialise()` and `fetchAndCheck()` together.

rocks/RocksIterator.java
haikalpribadi: `initialise()` does more than just "initialising" now.

rocks/RocksIterator.java
```diff
@@ -81,18 +83,26 @@ public final boolean hasNext() {
...
                 return false;
         }
     }
 
-    @Override
-    public final T next() {
-        if (!hasNext()) throw new NoSuchElementException();
-        state = State.EMPTY;
-        return next;
+    private synchronized boolean fetchAndCheck() {
```
haikalpribadi, line 92: Do we need to reorder this method in this class? It's harder to review the correctness of this PR now as more things change than necessary.

traversal/iterator/GraphIterator.java
```diff
@@ -102,8 +103,14 @@ else if (state == State.INIT) {
...
-            LOG.error("GraphProcedure: " + procedure.toString());
+            // note: catching runtime exception until we can gracefully interrupt running queries on tx close
+            if (e instanceof GraknException && ((GraknException) e).code().isPresent()
+                    && ((GraknException) e).code().get().equals(TRANSACTION_CLOSED.code())) {
+                LOG.debug("Transaction was closed during graph iteration");
+            } else {
+                LOG.error("Parameters: " + params.toString());
+                LOG.error("GraphProcedure: " + procedure.toString());
+            }
             throw e;
```
haikalpribadi, line 114: Do you still want to throw exception if transaction was closed? Looks like from your code that's not the case? You only `LOG.error()` in the `else` block.

haikalpribadi (changes requested):

traversal/iterator/GraphIterator.java
flyingsilverfin: it needs to propagate up to however's owning this to make sure they know the transaction is closed - might be their only signal. It just that this is so common it gets really slow and ugly to LOG.error the entire procedure each time it happens

rocks/RocksIterator.java
flyingsilverfin: By merging them together, we don't have to `synchronize` the `hasNext()` method because the inner methods that change the `state` are synchronized blocks. `synchronized` on `hasNext()` does have some performance implications (i tested) because it has to obtain a lock even when the iterator is just going to return `completed` or `fetched` I can have one call the other though and rename it

rocks/RocksIterator.java
flyingsilverfin: renamed and refactored

flyingsilverfin:
Updated @haikalpribadi have another look please :)

rocks/RocksIterator.java
```diff
@@ -81,18 +65,41 @@ public final boolean hasNext() {
...
                 return false;
         }
     }
 
-    @Override
-    public final T next() {
-        if (!hasNext()) throw new NoSuchElementException();
-        state = State.EMPTY;
-        return next;
+    private synchronized boolean initialiseAndFetch() {
```
haikalpribadi, line 74: This should be `initialiseAndCheck()`

rocks/RocksIterator.java
```diff
@@ -81,18 +65,41 @@ public final boolean hasNext() {
...
+    private synchronized boolean fetchAndCheck() {
+        if (state != State.COMPLETED) {
+            internalRocksIterator.next();
+            return mayConstructNext();
+        } else {
+            return false;
+        }
+    }
+
+    private synchronized boolean mayConstructNext() {
```
haikalpribadi, line 94: This should be called `hasValidNext()`

haikalpribadi (approved):
Still got naming changes to be made, but approving already

PR #6203, opened 2021-02-22
Fix possible segfault when closing session

alexjpwalker:

We fixed a possible segfault when closing a session while a db operation is running.

- Verify that the Storage is open before attempting to iterate over the schema graph

## Additional information

See https://github.com/graknlabs/grakn/pull/6185 for the related fix to the data graph.

### DIFF
--- a/rocks/RocksStorage.java
+++ b/rocks/RocksStorage.java
@@ -181,9 +181,11 @@ public abstract class RocksStorage implements Storage {
 
         @Override
         public <G> ResourceIterator<G> iterate(byte[] key, BiFunction<byte[], byte[], G> constructor) {
-            assert isOpen();
+//            assert isOpen();
+            if (!isOpen()) throw GraknException.of(TRANSACTION_CLOSED);
             RocksIterator<G> iterator = new RocksIterator<>(this, key, constructor);
             iterators.add(iterator);
+            if (!isOpen()) throw GraknException.of(TRANSACTION_CLOSED); //guard against close() race conditions
             return iterator.onFinalise(iterator::close);
         }
     }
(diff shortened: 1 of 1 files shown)


rocks/RocksStorage.java
```diff
@@ -181,9 +181,11 @@ public Cache(OptimisticTransactionDB rocksDB) {
 
         @Override
         public <G> ResourceIterator<G> iterate(byte[] key, BiFunction<byte[], byte[], G> constructor) {
-            assert isOpen();
+//            assert isOpen();
+            if (!isOpen()) throw GraknException.of(TRANSACTION_CLOSED);
             RocksIterator<G> iterator = new RocksIterator<>(this, key, constructor);
             iterators.add(iterator);
+            if (!isOpen()) throw GraknException.of(TRANSACTION_CLOSED); //guard against close() race conditions
```
alexjpwalker, line 188: Another possible solution is to make `iterate` `synchronized`, however this is deemed unacceptable for performance reasons.

rocks/RocksStorage.java
```diff
@@ -181,9 +181,11 @@ public Cache(OptimisticTransactionDB rocksDB) {
 
         @Override
         public <G> ResourceIterator<G> iterate(byte[] key, BiFunction<byte[], byte[], G> constructor) {
-            assert isOpen();
```
flyingsilverfin, line 184: one thing here @alexjpwalker  is that we should ask @haikalpribadi  why this was an assert not a proper exceptino before :) there might be solid reasoning it shouldn't be possible to ever reach this case (barring any new race conditions we uncovered!)

alexjpwalker: I will provide background here for @haikalpribadi :

The bug fixed by this PR was reproduced by running, from a client, a `match; delete; insert` query, and closing the session before the query had a chance to run to completion.

On investigating the segfault stack trace me and Joshua discovered it was in the schema graph. He noticed that the "is storage closed" checks (that he added for data graph operations in https://github.com/graknlabs/grakn/pull/6185) were not present in the schema operations, so we added them for schema.

PR #6204, opened 2021-02-22
In-use iterators that are closed can throw a closed message

flyingsilverfin:
When dealing with race conditions between using and closing RocksIterators, we can end up in a state where the user of an iterator calls `hasNext()` and get a `true` back, but by the time they call `next()` the iterator is closed. To avoid throwing a `NoSuchElementException`, we introduce a `Resource closed` exception

* `RocksIterator` throws a new message`RESOURCE_CLOSED` when it is closed but a user calls `next()` on it

### DIFF
--- a/traversal/iterator/GraphIterator.java
+++ b/traversal/iterator/GraphIterator.java
@@ -42,6 +42,7 @@ import java.util.Set;
 import java.util.TreeMap;
 
 import static grakn.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
+import static grakn.core.common.exception.ErrorMessage.Internal.RESOURCE_CLOSED;
 import static grakn.core.common.exception.ErrorMessage.Transaction.TRANSACTION_CLOSED;
 import static java.util.stream.Collectors.toMap;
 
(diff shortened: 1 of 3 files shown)


traversal/iterator/GraphIterator.java
```diff
@@ -104,9 +105,11 @@ else if (state == State.INIT) {
...
         } catch (Throwable e) {
             // note: catching runtime exception until we can gracefully interrupt running queries on tx close
-            if (e instanceof GraknException && ((GraknException) e).code().isPresent()
-                    && ((GraknException) e).code().get().equals(TRANSACTION_CLOSED.code())) {
-                LOG.debug("Transaction was closed during graph iteration");
+            if (e instanceof GraknException && ((GraknException) e).code().isPresent()) {
+                String code = ((GraknException) e).code().get();
+                if (code.equals(TRANSACTION_CLOSED.code()) || code.equals(RESOURCE_CLOSED.code())) {
+                    LOG.debug("Transaction was closed during graph iteration");
+                }
```
flyingsilverfin, line 112: this helps reduce the frequency at which we print graph procedures in "error" mode, which look like bad exceptions in the logs and are therefore bad UX

PR #6205, opened 2021-02-22
Revert to synchronous retrievables in reasoner

flyingsilverfin:
We revert to non-asynchronous producers in Retrievable until the CI stalling is fixed. We also refactor AnswerState to have more correct equality checks and a better flow when moving between states

* Revert `RetrievableResolver` to not use asynchronous producers for traversals, and instead use standard iterators
* Fix `AnswerState` data flow when moving up and down states
* Equality functions and hashes are updated


flyingsilverfin:
DO NOT merge - turns out we still stall reasoner in CI with this PR!!! However we will merge the AnswerState refactor soon

flyingsilverfin:
Answer state work was merged separately in https://github.com/graknlabs/grakn/pull/6208

PR #6208, opened 2021-02-23
Resolution answer state equality depends on applicable resolver

flyingsilverfin:
To ensure that equality functions in `AnswerState` can distinguish between two answers that come from content-identical paths but with different resolvers (eg. two resolvers that produce the same answer, but have different queries or resolver instances), we update the equality functions to include the correct resolver that extends any answer state.

* Update AnswerState equality functions
* Inject resolver that will receive an answer state, into the answer state
* The parent answer state that is being extended using `.with()` is responsible for updating derivations and concepts maps, instead of the child performing the operation and passing it in.


lolski:
Merging based on the request from Joshua.

PR #6209, opened 2021-02-23
Enable reasoner with synchronous iterators

flyingsilverfin:

Enable reasoning by default, and add the set of passing reasoning tests to CI. To do this, we revert reasoning to use synchronous (iterator-based) traversals. This PR signals that reasoner is now ready to start being tested!

* Enable reasoner by default in options
* Revert `RetrievableResolver` to use a iterator instead of `producer`


lolski (approved):
Approving `automation.yml` only

jmsfltchr (approved):

PR #6215, opened 2021-02-25
Resolver track per-request state with non-generic data structures

flyingsilverfin:
To clean up the resolution `Resolver` architecture, we need to get rid of the overly-generic data structure that was used by all resolvers to track responses per request. To allow resolvers to share some structure, we parametrise Compound resolvers by the type of response data structure they will use.

* Create abstract `CompoundResolver`, the parent of abstract `Conjunction` and abstract `Disjunction` resolvers
* Parametrise `CompoundResolver` by the type of `Responses` they will use
* Remove hacks to handle offsets by introducing abstract methods to accept answers `tryAcceptUpstreamAnswer` which is implemented differently in `Root` resolvers
* Every resolver now has the minimal required `Responses` object, rather than only using parts of a larger type of `Responses`


reasoner/resolution/resolver/CompoundResolver.java
```diff
@@ -0,0 +1,148 @@
...
+            }
+        }
+        return responses.get(fromUpstream);
+    }
+
+    abstract R responsesCreate(Request fromUpstream, int iteration);
+
+    abstract R responsesReiterate(Request fromUpstream, R priorResponses, int iteration);
+
+    static class Responses {
```
jmsfltchr, line 109: Is there a better name than `Responses`? i don't feel like it captures what this class does - it's not just a set of responses.

reasoner/resolution/resolver/DisjunctionResolver.java
```diff
@@ -146,64 +89,72 @@ protected void initialiseDownstreamResolvers() {
...
+    protected Responses responsesCreate(Request fromUpstream, int iteration) {
+        LOG.debug("{}: Creating a new Responses for request: {}", name(), fromUpstream);
         assert fromUpstream.partialAnswer().isFiltered() || fromUpstream.partialAnswer().isIdentity();
-        ResponseProducer responseProducer = new ResponseProducer(Iterators.empty(), iteration);
-        assert !downstreamResolvers.isEmpty();
+        Responses responses = new Responses(iteration);
+        assert !this.responses.isEmpty();
         for (Actor<ConjunctionResolver.Nested> conjunctionResolver : downstreamResolvers) {
-            Filtered downstream = fromUpstream.partialAnswer()
+            AnswerState.Partial.Filtered downstream = fromUpstream.partialAnswer()
```
jmsfltchr, line 98: Use static import to abbreviate.

jmsfltchr (changes requested):

jmsfltchr (approved):
Nice :)

PR #6218, opened 2021-02-25
Remove offset and limit from reasoner

flyingsilverfin:
It's incorrect to apply offset and limit inside reasoner, as we cannot sort natively. To do this properly, we keep offset and limit only at the top level iterator, and the caller is responsible for making sure that reasoner is not going to "over-compute" when a limit is present.

* Remove `offset` and `limit` from reasoner
* Clean up naming and imports in Resolvers


flyingsilverfin:
merging for @jmsfltchr to review post-merge

PR #6219, opened 2021-02-26
Rule materialisation lookups should use relation type

flyingsilverfin:
We incorrectly fail to use the relation type when performing a traversal to lookup whether the rule conclusion is already satisfied with a given set of concepts. This PR includes the relation type correctly in the lookup, fixing several reasoning BDD scenarios (notably, recursion tests).

* Rule materialisations lookup with the correct (always known) relation type.
* log the location of the resolution tracing log files in DEBUG mode
* avoid creating an extra unnecessary concept map in AnswerState
* Fix a reasoning test that tried to check a exception cause rather than the exception itself

### DIFF
--- a/logic/Rule.java
+++ b/logic/Rule.java
@@ -187,12 +187,6 @@ public class Rule {
         });
     }
 
-    @Override
-    public String toString() {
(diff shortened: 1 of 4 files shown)


flyingsilverfin:
Squashing to be reviewed by @jmsfltchr later

Issue #6228, opened 2021-03-02
Triggering an undefined rule throws NullPointerException


jmsfltchr:
## Description

Having Undefined any rule, making a query that would trigger the rule then throws a NullPointerException.

## Environment

1. OS (where Grakn server runs): Mac OS 10
2. Grakn version (and platform): Grakn Core 2.0 alpha 8
3. Grakn client: console

## Reproducible steps

1. Load the [docs schema](https://github.com/graknlabs/docs/blob/development/files/social-network/schema.gql)
2. We'll target this rule (any rule will do):
```
rule events-overlap:
		when {
			$e1 isa periodic-event;
			$e1 has start-date $sd1, has end-date $ed1;
			$e2 isa periodic-event;
			$e2 has start-date $sd2, has end-date $ed2;
			$sd2 > $sd1;
			$sd2 < $ed1;
			$e1 != $e2;
		} then {
			(overlapping: $e1, overlapping: $e2) isa event-overlap;
		};
```
3. Undefine the rule, commit, and try to trigger the rule
```
> transaction social_network schema write
social_network::schema::write> undefine rule events-overlap;

Concepts have been undefined
social_network::schema::write> commit
Transaction changes committed
> transaction social_network data read
social_network::data::read> match (overlapping: $e1, overlapping: $e2) isa event-overlap;

java.lang.NullPointerException Please check server logs for the stack trace.
>
```

flyingsilverfin:
Closed with #6230

PR #6230, opened 2021-03-03
Fully remove rule indexes when undefining a rule

flyingsilverfin:

As outlined in #6228, undefining a rule and committing appears to not completely clear the rule indices, causing subsequent null pointer exceptions. This PR actually deletes persisted rule indexes when undefining rules.

* Delete index keys from storage when undefining a rule
* On delete, only check if storage is open as opposed to ensuring that Transaction and storage are open (as the transaction is closed when the committing begins), causing deletes to throw a RESOURCE_CLOSED exception

### DIFF
--- a/reasoner/Reasoner.java
+++ b/reasoner/Reasoner.java
@@ -95,7 +95,6 @@ public class Reasoner {
         for (Conjunction conj : disjunction.conjunctions()) {
             Set<Variable> vars = conj.variables();
             Set<Negation> negs = conj.negations();
-            if (iterate(vars).anyMatch(v -> v.resolvedTypes().isEmpty())) return true;
             if (iterate(vars).flatMap(v -> iterate(v.resolvedTypes())).distinct().anyMatch(this::hasRule)) return true;
             if (!negs.isEmpty() && iterate(negs).anyMatch(n -> mayReason(n.disjunction()))) return true;
         }
--- a/rocks/RocksStorage.java
+++ b/rocks/RocksStorage.java
@@ -230,7 +230,7 @@ public abstract class RocksStorage implements Storage {
 
         @Override
         public void delete(byte[] key) {
-            if (!isOpen() || !transaction.isOpen()) throw GraknException.of(RESOURCE_CLOSED);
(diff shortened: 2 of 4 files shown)


reasoner/Reasoner.java
```diff
@@ -95,7 +95,6 @@ private boolean mayReason(Disjunction disjunction) {
         for (Conjunction conj : disjunction.conjunctions()) {
             Set<Variable> vars = conj.variables();
             Set<Negation> negs = conj.negations();
-            if (iterate(vars).anyMatch(v -> v.resolvedTypes().isEmpty())) return true;
```
flyingsilverfin, line 98: type resolver no longer produces empty type hints, unless the query is inCoherent or a schema query

test/integration/QueryTest.java
```diff
@@ -202,6 +202,9 @@ public void test_query_undefine() throws IOException {
 
                     // check total count
                     assertEquals(15 - 5, tx.logic().rules().toList().size());
+
+                    // a query that used to trigger a rule should not cause an error
+                    List<ConceptMap> answers = tx.query().match(Graql.parseQuery("match $x isa repo-fork;").asMatch()).toList();
```
haikalpribadi, line 207: Add a BDD scenario

rocks/RocksStorage.java
```diff
@@ -230,7 +230,7 @@ public Cache(OptimisticTransactionDB rocksDB) {
 
         @Override
         public void delete(byte[] key) {
-            if (!isOpen() || !transaction.isOpen()) throw GraknException.of(RESOURCE_CLOSED);
+            if (!isOpen()) throw GraknException.of(RESOURCE_CLOSED);
```
haikalpribadi, line 233: keep tx.isOpen, but check that tx is for session type data

haikalpribadi (approved):

test/integration/QueryTest.java
flyingsilverfin: test existed in BDD, but was ignored because reasoning was not enabled.  https://github.com/graknlabs/behaviour/pull/183 enables these and other tests

PR #6231, opened 2021-03-03
Fix rule satisfiability validation

### LINKED ISSUE #6227: Rule validation doesn't catch != between non-attributes
## Description

Rule validation fails to catch invalid rules where `!=` is used between entities.

## Environment

2. Grakn version (and platform): 2.0 alpha 8
3. Grakn client: console

## Reproducible Steps

Loading the [docs schema](https://github.com/graknlabs/docs/blob/development/files/social-network/schema.gql) completes successfully, despite the fact that this rule is present:
```
rule friends-excluded-permission:
		when {
			(content: $sc, by: $sb, hidden-from: $hf) isa friends-with-exclusion-sharing;
			(friend: $sb, $f) isa friendship;
			$f != $hf;
		} then {
			(content: $sc, grantee: $f) isa content-permission;
		};
```
This lead to the fact that nothing is inferred when querying `match (content: $sc, grantee: $f) isa content-permission;`. No error is thrown in the server logs.

## Expected Output

TypeResolver should have thrown an error when trying to load the schema.

flyingsilverfin:
We previously checked that a rule is coherent/satisfiability by checking each variable in the `when` and the `then`. However, we no longer set the satisfiability of a variable at all, instead setting only `isCoherent` at the conjunction. This PR updates Rule validation to reflect this change and removes `isSatisfiable` from `Variable`

Fixes #6227 

* Fix rule validation
* Add a new test case to test `when` validation, extending the current `then` validation
* Remove `isSatisfiable` getters and setters from `Variable`

### DIFF
--- a/logic/Rule.java
+++ b/logic/Rule.java
@@ -198,17 +198,11 @@ public class Rule {
     private void pruneThenResolvedTypes() {
         // TODO: name is inconsistent with elsewhere
         then.variables().stream().filter(variable -> variable.id().isName())
-                .forEach(thenVar ->
-                                 when.variables().stream()
-                                         .filter(whenVar -> whenVar.id().equals(thenVar.id()))
-                                         .filter(whenVar -> !(whenVar.isSatisfiable() && whenVar.resolvedTypes().isEmpty()))
-                                         .findFirst().ifPresent(whenVar -> {
-                                     if (thenVar.resolvedTypes().isEmpty() && thenVar.isSatisfiable()) {
-                                         thenVar.addResolvedTypes(whenVar.resolvedTypes());
-                                     } else thenVar.retainResolvedTypes(whenVar.resolvedTypes());
-                                     if (thenVar.resolvedTypes().isEmpty()) then.setCoherent(false);
-                                 })
-                );
+                .forEach(thenVar -> {
+                    Variable whenVar = when.variable(thenVar.id());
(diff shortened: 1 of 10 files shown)


logic/Rule.java
```diff
@@ -198,19 +196,12 @@ void validateCycles() {
...
-                                         .findFirst().ifPresent(whenVar -> {
-                                     if (thenVar.resolvedTypes().isEmpty() && thenVar.isSatisfiable()) {
-                                         thenVar.addResolvedTypes(whenVar.resolvedTypes());
-                                     } else thenVar.retainResolvedTypes(whenVar.resolvedTypes());
-                                     if (thenVar.resolvedTypes().isEmpty()) thenVar.setSatisfiable(false);
-                                 })
-                );
+                .forEach(thenVar -> {
+                    Variable whenVar = when.variable(thenVar.id());
+                    thenVar.retainResolvedTypes(whenVar.resolvedTypes());
```
jmsfltchr, line 202: What happens if the when variable's resolvedTypes is empty? (In the case it could be any Thing)

jmsfltchr (approved):

logic/Rule.java
jmsfltchr: This is now impossible as an empty set of types indicates failure, and need indicates "all types".

PR #6233, opened 2021-03-03
Prevent rare segfaults with storage read-write lock

flyingsilverfin:

To prevent segfaults that can occur when a transaction has asynchronous work calling `put()` or `get()` on storage, after another thread has called `close()` on storage. To prevent this, we insert a read-write lock that needs to be obtained before each storage operation. Benchmark simulations show this isn't a major performance bottleneck, but we should still move towards more efficient implementations if possible (starting with investigating the use of Optimistic locking for low-contention locks)

* Add read-write lock to every storage object
* any data `get()` and `put()` must obtain the read lock to prevent race conditions with `close()`
* any data delete must obtain write lock to prevent race conditions with `close()` and `put()` and `get()`

### DIFF
--- a/rocks/RocksStorage.java
+++ b/rocks/RocksStorage.java
@@ -257,13 +257,23 @@ public abstract class RocksStorage implements Storage {
 
         @Override
         public GraknException exception(ErrorMessage errorMessage) {
-            transaction.close();
+            try {
+                readWriteLock.writeLock().lock();
+                transaction.close();
+            } finally {
+                readWriteLock.writeLock().unlock();
+            }
             return super.exception(errorMessage);
         }
 
         @Override
         public GraknException exception(Exception exception) {
-            transaction.close();
+            try {
+                readWriteLock.writeLock().lock();
+                transaction.close();
+            } finally {
+                readWriteLock.writeLock().unlock();
+            }
             return super.exception(exception);
         }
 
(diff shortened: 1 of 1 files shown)


rocks/RocksStorage.java
```diff
@@ -297,25 +311,27 @@ public Schema(RocksDatabase database, RocksTransaction transaction) {
...
-                if (transaction.isOpen()) readWriteLock.writeLock().unlock();
+                readWriteLock.writeLock().unlock();
             }
         }
 
         @Override
         public void putUntracked(byte[] key, byte[] value) {
             assert isOpen() && !isReadOnly;
             try {
-                if (transaction.isOpen()) readWriteLock.writeLock().lock();
```
flyingsilverfin, line 313: why did we hav this check?

rocks/RocksStorage.java
```diff
@@ -339,29 +355,41 @@ public Data(RocksDatabase database, RocksTransaction transaction) {
...
+            } finally {
+                readWriteLock.readLock().unlock();
             }
         }
 
         @Override
         public void mergeUntracked(byte[] key, byte[] value) {
             assert isOpen() && !isReadOnly;
             try {
+                readWriteLock.readLock().lock();
```
flyingsilverfin, line 386: this only uses a read lock to guarantee race conditions against `close()`

rocks/RocksStorage.java
```diff
@@ -339,29 +355,41 @@ public Data(RocksDatabase database, RocksTransaction transaction) {
         public void put(byte[] key, byte[] value) {
             assert isOpen() && !isReadOnly;
             try {
+                readWriteLock.readLock().lock();
```
haikalpribadi, line 358: rename the lock to `deleteOrCloseLock`

rocks/RocksStorage.java
haikalpribadi: We should use a local boolean to remember to unlock, not remove it altogether

haikalpribadi (approved):

PR #6238, opened 2021-03-04
Implement rule negation validation to prevent contradictions

flyingsilverfin:
To properly support negation in rules, we must validate that rules cannot contradict themselves and reach a stable final solution, we require that rules containing negations do not recurse through the negation - this is called stratified negation.
This PR implements the commit time cycle detection required to ban cyclical rules through negations.

* At commit time, we reindex rules if required, then validate there are no cycles through rules with negation
* we remove the restriction that rules cannot have negations
* we enable a set of reasoning BDD files that require negations: `negation`, `recursion`, `schema-queries`
* introduce `Rule.Condition`, which caches concludables and their applicable rules to avoid recomputing them many times
* `ConditionResolver` uses the new `Rule.Condition`
* fix a bug in type resolution to do with including abstract types

### DIFF
--- a/logic/tool/TypeResolver.java
+++ b/logic/tool/TypeResolver.java
@@ -25,6 +25,7 @@ import grakn.core.concept.ConceptManager;
 import grakn.core.concept.thing.Thing;
 import grakn.core.graph.common.Encoding;
 import grakn.core.graph.vertex.TypeVertex;
+import grakn.core.graph.vertex.Vertex;
 import grakn.core.logic.LogicCache;
 import grakn.core.pattern.Conjunction;
 import grakn.core.pattern.Disjunction;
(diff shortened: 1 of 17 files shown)


kasper-piskorski (approved):
LGTM, I added extra test cases in behaviour. Also maybe `RuleRecursion`and `recursiveRules` are not the best names as the rules aren't necessarily recursive, simple RuleDependency and dependentRules or something similar might be clearer.

logic/tool/TypeResolver.java
```diff
@@ -182,28 +184,42 @@ private TraversalBuilder builder(Traversal traversal, Conjunction conjunction, L
     private Map<Identifier.Variable.Retrievable, Set<Label>> executeResolverTraversals(TraversalBuilder traversalBuilder) {
```
jmsfltchr, line 184: Rename to `resolveLabels` or `resolveVariableLabels`

jmsfltchr (changes requested):

PR #6242, opened 2021-03-08
Traversal must clear scoped roles when backtracking closures

flyingsilverfin:

A nondeterministic traversal execution bug was discovered by enabling ignored tests in https://github.com/graknlabs/behaviour/pull/183. This PR replicates the failure as a Traversal integration test and correctly clears scopes when a closure succeeds, but the next branch fails.

* pop scopes when a closure succeeds but then the next recursive branch/closure fails, in `computeFirst`
* replicate and keep the traversal test for this case and add to CI
* bump behaviour repository to the latest master

### DIFF
--- a/test/integration/BasicTest.java
+++ b/test/integration/BasicTest.java
@@ -818,12 +818,12 @@ public class BasicTest {
                         tx.close();
                         session.close();
                     }).start();
-                    tx.query().match(Graql.parseQuery("match $x sub thing;").asMatch());
+                    tx.query().match(Graql.parseQuery("match $x isa thing;").asMatch());
                 }).start();
             }
             Grakn.Session session = grakn.session(database, Arguments.Session.Type.DATA);
             Grakn.Transaction tx = session.transaction(Arguments.Transaction.Type.WRITE);
-            tx.query().match(Graql.parseQuery("match $x sub thing;").asMatch());
+            tx.query().match(Graql.parseQuery("match $x isa thing;").asMatch());
         }
     }
 }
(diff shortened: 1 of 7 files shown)


test/integration/BasicTest.java
```diff
@@ -818,12 +818,12 @@ public void test_query_cancelled_asynchronously() throws IOException {
...
                         session.close();
                     }).start();
-                    tx.query().match(Graql.parseQuery("match $x isa thing;").asMatch());
+                    tx.query().match(Graql.parseQuery("match $x sub thing;").asMatch());
                 }).start();
             }
             Grakn.Session session = grakn.session(database, Arguments.Session.Type.DATA);
             Grakn.Transaction tx = session.transaction(Arguments.Transaction.Type.WRITE);
-            tx.query().match(Graql.parseQuery("match $x isa thing;").asMatch());
+            tx.query().match(Graql.parseQuery("match $x sub thing;").asMatch());
```
haikalpribadi, line 826: Why do we need to change this, @flyingsilverfin ?

haikalpribadi (commented):
All looks good, just got one comment below

test/integration/BasicTest.java
flyingsilverfin: errr I can move this to my other PR, however in the other PR type resolver complains this query is unsatisfiable - which is correct, there are no concrete subtypes of `thing`.

In the other PR i've made the "no abstract types are allowed on Thing variables" smarter than before, which is why it picks this up now. However probably makes sense to move this change over there!

flyingsilverfin: removed from this PR

haikalpribadi: yes it does make sense to keep that change scoped in the right PR, @flyingsilverfin

haikalpribadi (approved):

PR #6243, opened 2021-03-08
Optimise resolution of fully bound concludables

flyingsilverfin:
In some cases, a concludable that is being resolved may have all of its retrievable variables (named and anonymous) fully bound. This means that only a single answer is required to prove the concludable holds, rather than all possible ways. We add short circuiting to prevent exploring further rules in a concludable when a single answer has been found.

* Concludables `RequestState` can record whether a single answer is required
* When producing an answer from a `ConcludableResolver`, we may short circuit and remove all possible downstream answer producers if only 1 answer is required

## ADR: why only short circuit in fully bound concludables
We could short circuit when any of our resolvers end up in a fully bound state. However, it's only useful in concludables:
* A fully bound `Retrievable` will only find exactly 1 answer, so short circuiting would not speed anything up
* `Negation` resolvers only ever look for 1 answer
* all `Conjunction` resolvers will very rarely be fully bound. Even if they are, if retrievables and concludables only return 1 answer when they are fully bound, then the conjunction too will only return 1 answer - so it doesn't add anything here

We could implement this optimisation for disjunctions, however, it is not worth our time at this moment as a fully bound disjunction is uncommon.

### DIFF
--- a/reasoner/resolution/resolver/ConcludableResolver.java
+++ b/reasoner/resolution/resolver/ConcludableResolver.java
@@ -216,13 +216,12 @@ public class ConcludableResolver extends Resolver<ConcludableResolver> {
             iterationState.nextIteration(iteration);
         }
 
-        boolean singleAnswerRequired = fromUpstream.partialAnswer().conceptMap().concepts().keySet().containsAll(missingBounds());
-
         assert fromUpstream.partialAnswer().isMapped();
         FunctionalIterator<Partial<?>> upstreamAnswers =
                 traversalIterator(concludable.pattern(), fromUpstream.partialAnswer().conceptMap())
                         .map(conceptMap -> fromUpstream.partialAnswer().asMapped().aggregateToUpstream(conceptMap));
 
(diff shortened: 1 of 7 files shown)


reasoner/resolution/resolver/ConcludableResolver.java
```diff
@@ -69,6 +75,7 @@ public ConcludableResolver(Driver<ConcludableResolver> driver, Concludable concl
         this.applicableRules = new LinkedHashMap<>();
         this.recursionStates = new HashMap<>();
         this.requestStates = new HashMap<>();
+        this.missingBounds = missingBounds(concludable.pattern());
```
jmsfltchr, line 78: I'm also thinking that this could be better named `unboundVars`, as they aren't _missing_ per se

jmsfltchr (changes requested):

flyingsilverfin:
merging with verbal approval from @jmsfltchr  after addressing the outstanding comment

PR #6244, opened 2021-03-09
All conjunction resolvers deduplicate concept maps

flyingsilverfin:
As an optimisation we have accidentally removed, we re-introduce the ability for each Conjunction resolver to deduplicte answers. This should reduce work where (notably in Condition resolvers), more variables are present than in the upstream resolver (eg. compare conclusion resolvers to condition resolvers) - not filtering out duplicates produced by varying variables the upstream doesn't care about can lead to doing too much work.

* All conjunction resolvers use the same `RequestState`, which is used to deduplicte answers

### DIFF
--- a/reasoner/resolution/resolver/ConjunctionResolver.java
+++ b/reasoner/resolution/resolver/ConjunctionResolver.java
@@ -21,6 +21,7 @@ package grakn.core.reasoner.resolution.resolver;
 import grakn.core.common.exception.GraknException;
 import grakn.core.common.iterator.Iterators;
 import grakn.core.concept.ConceptManager;
+import grakn.core.concept.answer.ConceptMap;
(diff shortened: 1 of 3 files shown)


reasoner/resolution/resolver/ConjunctionResolver.java
```diff
@@ -206,6 +218,33 @@ protected REQ_STATE requestStateReiterate(Request fromUpstream, REQ_STATE reques
         }
     }
 
+
+    public static class RequestState extends CompoundResolver.RequestState {
```
flyingsilverfin, line 222: this behaviour is now in all conjunction resolvers, so we only have 1 type of Conjunction request state

jmsfltchr (approved):

PR #6245, opened 2021-03-09
Upgraded TransactionService to receive batched requests to increase throughput

haikalpribadi:

When sending large amounts of requests from the client to server, it is possible to reach GRPC's maximum throughput of messages per second, on large scale systems. For example, a client with 32 cores, can have 1000 concurrent transactions sending requests to the server. This will easily go beyond what GRPC can handle, and thus becomes the bottleneck in the system. That client will not be sending as many requests it wants per second, and the server may still be able to handle more requests to handle. We have now overcome this problem, by allowing the transaction stream (the predominant GRPC service) to _"batch"_ the transaction requests into one GRPC message to send to the server, within a given time window. In the client-side implementations, we have chosen the time window to be 1 millisecond - small enough to not cause any noticeable performance degradation on individual requests, but big enough to _"batch"_ all requests within that time window into one message and multiply the throughput.


.grabl/automation.yml
```diff
@@ -132,8 +132,8 @@ build:
     deploy-artifact-snapshot:
       image: graknlabs-ubuntu-20.04
       filter:
-        owner: graknlabs
-        branch: master
+        owner: [graknlabs, haikalpribadi]
+        branch: [master, optimise-grpc-server]
```
alexjpwalker, line 136: Is this still supposed to be here?

concurrent/common/Executors.java
```diff
@@ -58,7 +59,7 @@ private Executors(int parallelisation) {
         if (parallelisation <= 0) throw GraknException.of(ILLEGAL_ARGUMENT);
         // TODO: this is temporarily not used until we enable TransactionExecutor
         // serviceExecutorService = newFixedThreadPool(serviceThreadCount(parallelisation), threadFactory(GRAKN_CORE_SERVICE_THREAD_NAME));
```
alexjpwalker, line 61: We should delete these comments.

server/GraknService.java
```diff
@@ -55,13 +56,13 @@
...
-    private final TransactionExecutor executor;
+    private final AsyncTransactionExecutor executor;
 
     public GraknService(Grakn grakn) {
         this.grakn = grakn;
         sessionServices = new ConcurrentHashMap<>();
-        // TODO: this is temporarily not used until we enable TransactionExecutor
-        //executor = new TransactionExecutor(PARALLELISATION_FACTOR, TRANSACTION_EXECUTOR_QUEUE_FACTOR);
+        // TODO: Enable this along with AsyncTransactionExecutor
+        //executor = new AsyncTransactionExecutor(PARALLELISATION_FACTOR, TRANSACTION_EXECUTOR_QUEUE_FACTOR);
```
alexjpwalker, line 65: Is this related to the reintroduction of `executorServiceExecutor` above in the PR?

server/GraknService.java
```diff
@@ -201,7 +206,8 @@ public void remove(SessionService sessionSrv) {
     }
 
     public void close() {
-//         executor.close();
+        // TODO: Enable this along with AsyncTransactionExecutor
+        // executor.close();
```
alexjpwalker, line 210: Same comment as above

server/transaction/TransactionService.java
```diff
@@ -62,11 +61,13 @@
     private final AtomicBoolean isOpen;
     private final AtomicBoolean commitRequested;
     private final AtomicInteger ongoingRequests;
+    private final int latencyMillis;
     private volatile String commitRequestID;
 
     public TransactionService(SessionService sessionSrv, TransactionStream stream, TransactionProto.Transaction.Open.Req request) {
         this.sessionSrv = sessionSrv;
         this.stream = stream;
+        this.latencyMillis = request.getLatencyMillis();
```
alexjpwalker, line 70: Do you think it's worth renaming this to `networkLatencyMillis` for clarity, since we use the term _latency_ for multiple things? Or is it clear enough?

.grabl/automation.yml
haikalpribadi: yes this will be removed before merging

concurrent/common/Executors.java
haikalpribadi: I think it's okay to keep `TODO` comments, but in this specific instance, the comment is no longer correct, so I'll remove it.

server/BUILD
```diff
@@ -162,7 +162,7 @@ assemble_deps_common = [
 ]
 
 assemble_files = {
-    "//server/conf:logback": "server/conf/logback.xml",
+    "//server/conf:logback": "server/conf/logback-test.xml",
```
lolski, line 165: Let's revert back to `logback.xml`

server/BUILD
```diff
@@ -171,7 +171,7 @@ assemble_files = {
 
 permissions = {
     "server/conf/grakn.properties": "0755",
-    "server/conf/logback.xml": "0755",
+    "server/conf/logback-test.xml": "0755",
```
lolski, line 174: Let's revert back to `logback.xml`

lolski (approved):

server/GraknService.java
haikalpribadi: I've decided to just remove the comments for now. We'll bring them back properly when we get to implement `matchAsync` and `insertAsync`.

server/transaction/TransactionService.java
haikalpribadi: I think `networkLatencyMillis` would be clearer here, you're right. But if we do that, we have to rename `latency_millis` to `network_latency_millis` in the protocol definitions too, right?

alexjpwalker: Yeah I think that would be worth doing too.

server/TransactionService.java
```diff
@@ -0,0 +1,346 @@
...
+    private volatile int networkLatencyMillis;
+
+    public TransactionService(GraknService graknSrv, StreamObserver<TransactionProto.Transaction.Res> responder) {
+        this.responder = SynchronizedStreamObserver.of(responder);
+        this.streams = new ConcurrentHashMap<>();
+        this.graknSrv = graknSrv;
+        this.isOpen = new AtomicBoolean(false);
+    }
+
+    private class Services {
```
lolski, line 93: should we group it at the bottom with the other inner classes?

lolski (approved):

PR #6249, opened 2021-03-11
Implement Explanations

flyingsilverfin:

We extend reasoner to introduce a new `explain()` API, which explains all reasons a fact is inferred, iteratively retruning all explanations. To access this API, the user must obtain an explainable concept map, which is returned with `explain=True` in the query/transaction options.

* Introduce `explain()` API
* Introduce `ExplanationProducer`, which can instigate reasoning to find all reasons a particular inferred answer (eg. a `ConceptMap.Explainable` + the concept map itself as bounds) is true
* `explain()` flag returns explainable concept maps, which ignore `get` filters and returns even anonymous variables
* Create `Explanation`, which can explain a single layer of inference: from a conjunction, through a rule to a rule's `when` conjunction (the next layer of explanations can then be asked for recursively
* Largely refactor `AnswerState` to transparently create `Explain` alternatives for each state in the state transitions, which record information required to generate `Explanation`s - we also extract out `AnswerState` to encapsulate much of the API and casting operations separately from `AnswerStateImpl` - this makes everything less bloated
* Update `Conjunction.toString()` to not have labeled variables print themselves as `$_friend type friend`, and more cleanly integrate labels into the toString representations


concept/BUILD
```diff
@@ -50,6 +50,7 @@ native_java_libraries(
         "@maven//:com_google_code_findbugs_jsr305",
     ],
     native_libraries_deps = [
+        "//pattern:pattern", # TODO review this is ok
```
flyingsilverfin, line 53: todo review this haikal

haikalpribadi (changes requested):
Partial review

flyingsilverfin:
closing in order to split into two PRs,  https://github.com/graknlabs/grakn/pull/6265 and an uncoming one implementing protocol + rpc

PR #6257, opened 2021-03-16
Add encoding prefix for Cluster's "last applied index" field

alexjpwalker:

Previously, Grakn Cluster stored its "last applied index" metadata field with RocksDB key of (-1). This could have led to trouble in the future if that key became incidentally used by anything else. To avoid this scenario, we now have a dedicated encoding prefix named `CLUSTER_LAST_APPLIED`.

Add CLUSTER_LAST_APPLIED encoding prefix

### DIFF
--- a/graph/common/Encoding.java
+++ b/graph/common/Encoding.java
@@ -134,7 +134,8 @@ public class Encoding {
         STATISTICS(1),
         TYPE(2),
         THING(3),
-        RULE(4);
+        RULE(4),
+        CLUSTER(5);
 
         private final int key;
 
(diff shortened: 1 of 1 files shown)


alexjpwalker:
We decided to NOT merge this PR because Grakn Core's codebase should not contain information that is only relevant to Cluster at such a deep level.

We raised https://github.com/graknlabs/grakn/issues/6258 to ensure that we compress the encoding prefixes soon, as they are currently very spread out which has become constricting to our ability to cleanly roll out new Grakn features.

PR #6259, opened 2021-03-16
Refactor GraknServer to be open for extension

lolski:

We have extended `GraknServer` and `ServerCommand.Start` to be open for extension benefitting Grakn Cluster.

1.  Changes to `GraknServer`:
    - Updated the constructor to allow a subclass to supply its own `Factory` implementation.
    - Allow a subclass to override the RPC server. In Grakn Cluster, we would provide an RPC server that has additional RPC endpoints related to clustering. 
    - Allow overriding the product name. By default, it will be`Grakn Core Server` but in Grakn Cluster it will be `Grakn Cluster Server`. It's mostly useful for printing logs.
    - Allow a subclass to access key fields: `command`, `grakn`, `server`, and `factory`.

2. Changes to `ServerCommand.Start`
    - Allow a subclass to access `data` and `logs` fields. This class would be extended by Grakn Cluster to provide more startup options related to clustering.

3. Additionally, we've refactored the `GraknServer` class.
    - Rearranged the ordering of methods to be more sensible.
    - Extracted static methods related to printing logo and parsing command line options into `Util`.


server/GraknServer.java
```diff
@@ -101,35 +94,14 @@ private GraknServer(ServerCommand.Start command) throws IOException {
...
-        return command.dataDir();
-    }
-
-    private void configureAndVerifyDataDir() throws IOException {
+    private void configureAndVerifyDataDir() {
         if (!Files.isDirectory(this.command.dataDir())) {
             if (this.command.dataDir().equals(ServerDefaults.DATA_DIR)) {
-                Files.createDirectory(this.command.dataDir());
+                try {
+                    Files.createDirectory(this.command.dataDir());
```
lolski, line 101: Making the `configureAndVerifyDataDir` throw a runtime exception, since failing to create the data directory is not a recoverable condition.

server/GraknServer.java
```diff
@@ -26,70 +26,66 @@
...
 public class GraknServer implements AutoCloseable {
 
     private static final Logger LOG = LoggerFactory.getLogger(GraknServer.class);
 
-    private final Grakn grakn;
-    private final Server server;
-    private final ServerCommand.Start command;
-    private final GraknService graknService;
-    private final MigratorService migratorRPCService;
+    protected final ServerCommand.Start command;
```
lolski, line 65: May want to discuss how to order the fields here. I think `server, `graknService`, and `migratorService` should be grouped together. Additionally, `factory` and `grakn` should also be close together.

server/GraknServer.java
```diff
@@ -153,74 +127,79 @@ private void configureTracing() {
...
     }
 
-    private static void printASCIILogo() throws IOException {
-        if (ASCII_LOGO_FILE.exists()) {
-            System.out.println("\n" + new String(Files.readAllBytes(ASCII_LOGO_FILE.toPath()), StandardCharsets.UTF_8));
-        }
+    protected Server rpcServer() {
+        assert Executors.isInitialised();
+
+        graknService = new GraknService(grakn);
```
lolski, line 133: The service classes are now instantiated here rather than the constructor, for consistency with how we construct it in Grakn Cluster. In Grakn Cluster, we will instantiate an additional service for clustering.

server/GraknServer.java
```diff
@@ -279,49 +241,24 @@ private static void startGraknServer(ServerCommand.Start command) throws IOExcep
...
-            server.shutdown();
-            server.awaitTermination();
-            grakn.close();
-            System.runFinalization();
-            LOG.info("Grakn Core Server has been shutdown");
-        } catch (InterruptedException e) {
-            LOG.error(FAILED_AT_STOPPING.message(), e);
-            Thread.currentThread().interrupt();
-        }
+    protected static void printSchema(ServerCommand.PrintSchema printSchemaCommand) {
```
lolski, line 247: I don't think it makes sense for us to extract the migrator-related method since they're actually quite small. What do you think Haikal?

haikalpribadi (approved):

PR #6263, opened 2021-03-22
Always use the little endian to represent merge operator numbers

lriuui0x0:

RocksDB uses little endian to represent numbers for merge operators always, instead of depending on the system architecture dynamically.

- Uses `LITTLE_ENDIAN` always to represent merge operator numbers.


haikalpribadi (changes requested):
Can we also shorten the code with using `ByteBuffer.wrap()` and compact it to one line if possible, @lriuui0x0 ?

haikalpribadi (approved):

PR #6286, opened 2021-04-07
In gRPC, encode Request IDs as bytes, and move 'inferred' to Thing

alexjpwalker:

Previously the ID of a transaction request was stored as a string, which is an inefficient representation of a UUID. We've changed it to bytes.

Also, we deleted `IsInferred` and added the `inferred` flag to `Thing`, allowing clients to check if a Thing is inferred without an additional network roundtrip.

Store request IDs as bytes, not strings
Add `inferred` flag to `Thing` in protobuf


alexjpwalker:
Do not merge until Clients are updated.

haikalpribadi (approved):

PR #6289, opened 2021-04-08
Remove most internal uses of java Streams

flyingsilverfin:
We almost entirely remove usages of java's `Stream` class from our internal code, replacing it with our own `FunctionalIterator`. 

We still use streams occasionaly when requiring:
* sorting
* collecting to an array (instead of a list)
* `Streams.compareSize`

* Remove `Stream` from `//concept`, and propagate the changes upwards. This happens to remove almost all usage of Stream


concept/ConceptManager.java
```diff
@@ -248,7 +247,7 @@ static void writeAttributeType(StringBuilder builder, AttributeType attributeTyp
             writeOwns(builder, attributeType);
             writePlays(builder, attributeType);
             builder.append(StringBuilders.SEMICOLON_NEWLINE_X2);
-            attributeType.getSubtypesExplicit()
+            attributeType.getSubtypesExplicit().stream()
```
flyingsilverfin, line 250: these have stayed as streams because we use `sorted()`, we could re-implement a bit later if we wanted. I was aiming to remove the use of `Stream` from method signatures

concept/type/impl/AttributeTypeImpl.java
```diff
@@ -777,11 +776,13 @@ private String(GraphManager graphMgr, TypeVertex vertex) {
         @Override
         public void setRegex(Pattern regex) {
             if (regex != null) {
-                getInstances().parallel().forEach(attribute -> {
+                // TODO can we do this in parallel as it was before?
```
flyingsilverfin, line 779: question here: do we need to retain `parallel()` here? if so, we should examine how we can make our functional iterators have a parallel() method -- possibly using Producers + `ProducerIterator`

common/iterator/Iterators.java
```diff
@@ -50,6 +50,11 @@
         return new BaseIterator<>(Either.second(collection.iterator()));
     }
 
+    @SafeVarargs
+    public static <T> FunctionalIterator<T> iterate(T... elements) {
+        return new BaseIterator<T>(Either.second(list(elements).iterator()));
```
haikalpribadi, line 55: This should call the pre-existing method above it. And move this method above it.

concept/type/impl/RelationTypeImpl.java
```diff
@@ -141,7 +141,7 @@ public void setRelates(String roleLabel, String overriddenLabel) {
         Optional<RoleTypeImpl> inherited;
         assert getSupertype() != null;
         if (declaredRoles().anyMatch(r -> r.getLabel().name().equals(overriddenLabel)) ||
-                !(inherited = getSupertype().asRelationType().getRelates().filter(role -> role.getLabel().name().equals(overriddenLabel)).findFirst()).isPresent()) {
+                !(inherited = getSupertype().asRelationType().getRelates().filter(role -> role.getLabel().name().equals(overriddenLabel)).first()).isPresent()) {
```
haikalpribadi, line 144: formatting

concept/type/impl/ThingTypeImpl.java
```diff
@@ -38,7 +39,6 @@
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static grakn.core.common.collection.Streams.compareSize;
```
haikalpribadi, line 44: You can convert this helper to `Iterators` now

haikalpribadi (approved):
Approved with some comments for revision

PR #6294, opened 2021-04-12
At most a single relation is allowed on a thing variable

flyingsilverfin:
We address the fact that a user can write semantically undefined multiple relation constraints on a single relation variable. In this PR we disallow writing multiple relation constraints on the same relation variable. 

## Semantic Explanation

When addressing a relation in a query, we constraint a variable with a relation constraint:

```
$r (role-a: $x);
```

This statements requires that the variable `r` identifies a relation with one role player `$x` playing the role `role-a`. However, if we add another, separate relation constraint:
```
$r (role-b: $y);
```
we now also constraint the relation `r` to be idenfied with a player `$y` playing a role `role-b`. Since we view relations as hyper-relations defined as a set of role players, it doesn't semantically make sense to try to define `$r` by two different sets of role players (or alternatively, it's identified by their intersection). The next-best semantically sound thing to execute would be to do the following:
```
$r (role-a: $x, role-b: $y);
```

If we wanted to query for other parts of the same relation, we should define a new relation variable and set its identity to be equivalent to the first:
```
$r1 (role-a: $x);
$r2 is $r1;
$r2 (role-b: $y);
```
This allows us to satisfy the semantic requirements of definition of relations, and explore other role players on the next relation variable `$r2`

* Disallow writing multiple relation constraints on a relation variable
* `ThingVariable` now contains a single nullable `relationConstraint` instead of a possibly empty set

### DIFF
--- a/common/exception/ErrorMessage.java
+++ b/common/exception/ErrorMessage.java
@@ -204,18 +204,20 @@ public abstract class ErrorMessage extends grakn.common.exception.ErrorMessage {
                 new Pattern(8, "The thing variable '%s' has multiple 'iid' constraints.");
         public static final Pattern MULTIPLE_THING_CONSTRAINT_ISA =
                 new Pattern(9, "The thing variable '%s' has multiple 'isa' constraints, '%s' and '%s'.");
+        public static final Pattern MULTIPLE_THING_CONSTRAINT_RELATION =
+                new Pattern(10, "The thing variable '%s' has multiple 'relation' constraints");
         public static final Pattern ILLEGAL_DERIVED_THING_CONSTRAINT_ISA =
-                new Pattern(10, "The thing variable '%s' has a derived 'isa' constraint, in a query that does not allow it.");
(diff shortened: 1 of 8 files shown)


common/exception/ErrorMessage.java
```diff
@@ -204,18 +204,20 @@ public static void loadConstants() {
                 new Pattern(8, "The thing variable '%s' has multiple 'iid' constraints.");
         public static final Pattern MULTIPLE_THING_CONSTRAINT_ISA =
                 new Pattern(9, "The thing variable '%s' has multiple 'isa' constraints, '%s' and '%s'.");
+        public static final Pattern MULTIPLE_THING_CONSTRAINT_RELATION =
+                new Pattern(10, "The thing variable '%s' has multiple 'relation' constraints");
```
haikalpribadi, line 208: You can say `"The relation variable ..."` here to be more helpful

haikalpribadi (approved):
Don't forget the comment on error message

PR #6296, opened 2021-04-13
Validate rule value type compatibility and fix attribute materialisation 

flyingsilverfin:

We fix a bug in which rules attempted to do an invalid value conversion, causing an exception. We also refactor rule validation to include a validation step to make sure attributes to be inserted are of the right type.

For example, the following now throws a meaningful error message about converting a Double to a Long:
```
define rule bad-double-as-long: when {
  $x isa person;
} then {
  $x has long-age 2.5;
}
``` 

* Implement correct value conversion when materialising inferences
* Add a validation step to validate that Explicit Has conclusions are able to insert the written attribute value as the provided type
* Refactor rule constructor to keep only 1 constructor
* bump `graknlabs_behaviour` containing new rule validation tests


haikalpribadi (changes requested):
Replace the exception-driven logic with value type "assignable" comparisons

haikalpribadi (approved):

PR #6300, opened 2021-04-20
Fix validation for owns to owns key schema migration

flyingsilverfin:
When migrating `owns` to `owns @key` in the schema, we were incorrectly checking every attribute is owned once -- we should check that every attribute that is owned, is only owned once. We now correctly validate that
1. all instances that should have a key have exactly 1 key attribute
2. every attribute keying an instance of a type is only keying exactly 1 instance of that type

* Fix key-conversion validation
* bump behaviour with an extended test that would've caught this issue earlier


concept/type/impl/ThingTypeImpl.java
```diff
@@ -199,18 +200,21 @@ private void ownsKey(AttributeTypeImpl attributeType) {
 
         if ((ownsEdge = vertex.outs().edge(OWNS, attVertex)) != null) {
             // TODO: These ownership and uniqueness checks should be parallelised to scale better
-            if (getInstances().anyMatch(thing -> compareSize(thing.getHas(attributeType), 1) != 0)) {
-                throw exception(TypeDBException.of(OWNS_KEY_PRECONDITION_OWNERSHIP, vertex.label(), attVertex.label()));
-            } else if (attributeType.getInstances().anyMatch(att -> compareSize(att.getOwners(this), 1) != 0)) {
-                throw exception(TypeDBException.of(OWNS_KEY_PRECONDITION_UNIQUENESS, attVertex.label(), vertex.label()));
-            }
+            getInstances().forEachRemaining(thing -> {
+                List<? extends Attribute> attributes = thing.getHas(attributeType).limit(2).toList();
```
haikalpribadi, line 204: Use iterator and check for all 3 cases

haikalpribadi (changes requested):

haikalpribadi (approved):

PR #6317, opened 2021-05-07
Made fields and methods RocksDatabaseManager accessible to subclasses

lolski:

We've made several fields in `RocksDatabaseManager` accessible to subclasses such as the one in TypeDB Cluster.

Additionally,  we've introduced `close` method so that subclasses can provide their own implementation.

- Change the visibility of `typedb`, `databases`, `databaseFactory`, and `loadAll()` to `protected` for use by the subclasses
- Introduced `close` method with `protected` visibility.


rocks/RocksDatabaseManager.java
```diff
@@ -80,4 +80,8 @@ public RocksDatabase get(String name) {
     void remove(RocksDatabase database) {
         databases.remove(database.name());
     }
+
+    protected void close() {
```
lolski, line 84: Create a `close()` method so that it can be overridden in TypeDB Cluster.

PR #6320, opened 2021-05-11
Concludable and Retrievable Subsumption Caching

jmsfltchr:

We introduce caching in the reasoner avoid repeated work. Adding this caching has also allowed us to tighten the criteria upon which we trigger an outer reiteration of the reasoner for a given query. Such an iteration is necessary to search for inferred facts that can only be found based on the inferences made by prior iterations.

The cache ensures that work can be re-used for the same exact partial answer when revisited. We also implement cache subsumption: any partial answer can utilise the cached answers for any more general (subsuming) partial answer which has a complete cache.

- Subsumptive caching for ConcludableResolver
- Subsumptive caching for RetrievableResolver
- Abstracts RequestState across the non-compound resolvers. The compound resolvers need to be brought in line with this approach in a follow-up PR.
- Adds Poller classes to implement polling the cache for new answers
- Account for the case of `$x isa $type` in rules (previously handled by always triggering at least one reiteration) by generating answers over the cartesian product of all possible types for the answer found. This takes place during un-unification.

Authored jointly with @flyingsilverfin

### DIFF
--- a/reasoner/resolution/Planner.java
+++ b/reasoner/resolution/Planner.java
@@ -18,8 +18,6 @@
 package com.vaticle.typedb.core.reasoner.resolution;
 
 import com.vaticle.typedb.core.common.exception.TypeDBException;
-import com.vaticle.typedb.core.concept.ConceptManager;
-import com.vaticle.typedb.core.logic.LogicManager;
 import com.vaticle.typedb.core.logic.resolvable.Concludable;
 import com.vaticle.typedb.core.logic.resolvable.Resolvable;
 import com.vaticle.typedb.core.pattern.variable.ThingVariable;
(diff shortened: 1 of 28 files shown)


reasoner/resolution/Planner.java
```diff
@@ -82,31 +77,30 @@ private void computePlan() {
                 Optional<com.vaticle.typedb.core.logic.resolvable.Retrievable> retrievable;
 
                 // Retrievable where:
+                // it is connected
                 // all of it's dependencies are already satisfied,
-                // which will answer the most variables
-                retrievable = mostAnsweredVars(dependenciesSatisfied(hasAnsweredVar(remaining.stream().filter(Resolvable::isRetrievable))))
+                retrievable = sortedByStatistics(dependenciesSatisfied(hasAnsweredVar(remaining.stream().filter(Resolvable::isRetrievable))))
```
flyingsilverfin, line 82: `sortByVisitCount`

jmsfltchr:
Note that we have one flakey test: `ancestor-friend test` in `reasoner/recursion.feature` which is the next priority to fix.

haikalpribadi (approved):

flyingsilverfin (approved):

jmsfltchr:
The test `given a directed graph, all pairs of vertices (x,y) such that y is reachable from x can be found` is a regression, so holding off the merge until that's been investigated.

jmsfltchr:
> The test `given a directed graph, all pairs of vertices (x,y) such that y is reachable from x can be found` is a regression, so holding off the merge until that's been investigated.

Has been investigated and determined that reiteration flags are not being propagated to the root. Previously we thought this scenario couldn't occur, but this test proved otherwise. This will be fixed in a follow-up PR.

PR #6330, opened 2021-05-18
Statistics background thread catches correct conflict codes

flyingsilverfin:

As of #6306 we have enabled ACID transactionality, but the background statistics counter did not catch the correct error codes that represent conflict operations and simply died. We now catch the correct error codes and ensure statistics are still updated in the background thread.

* catch the correct error codes in statistics counting thread

### DIFF
--- a/rocks/RocksDatabase.java
+++ b/rocks/RocksDatabase.java
@@ -33,7 +33,6 @@ import com.vaticle.typedb.core.logic.LogicCache;
 import com.vaticle.typedb.core.traversal.TraversalCache;
 import org.rocksdb.OptimisticTransactionDB;
 import org.rocksdb.RocksDBException;
-import org.rocksdb.Status;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


rocks/RocksDatabase.java
```diff
@@ -614,15 +614,12 @@ private void countFn() {
                 } catch (TypeDBException e) {
                     if (e.code().isPresent() && e.code().get().equals(DATABASE_CLOSED.code())) {
                         break;
+                    } else if (e.code().isPresent() && (e.code().get().equals(TRANSACTION_CONSISTENCY_MODIFY_DELETE_VIOLATION.code()) ||
```
lriuui0x0, line 617: Do we wrap rocksdb exception with our exception now?

flyingsilverfin: rocks doesn't throw this exception anymore, we do all the conflict detection ourselves (#6306 )

haikalpribadi (approved):

PR #6340, opened 2021-05-26
Use a poller to store answer sources for the reasoner cache

jmsfltchr:

Within the cache we had a source of new answers as an iterator. We want to poll this store of answers in case more have been added. For this we introduce pollers that can be linked together as new answer sources are found.

- Add `BasePoller` which polls an iterator.
- Add `LinkedPoller` which polls over both linked Pollers.
- Refactor `FlatMapPoller` to take in a function that yields a poller, for consistency with `LinkedPoller`.
- `Pollers` class for static poller convenience methods.

### DIFF
--- a/common/poller/AbstractPoller.java
+++ b/common/poller/AbstractPoller.java
@@ -18,12 +18,10 @@
 
 package com.vaticle.typedb.core.common.poller;
 
-import com.vaticle.typedb.core.common.iterator.FunctionalIterator;
-
 import java.util.Optional;
 import java.util.function.Function;
 
-public class AbstractPoller<T> implements Poller<T> {
+public abstract class AbstractPoller<T> implements Poller<T> {
 
     @Override
     public Optional<T> poll() {
@@ -31,8 +29,16 @@ public class AbstractPoller<T> implements Poller<T> {
     }
 
     @Override
-    public <U> Poller<U> flatMap(Function<T, FunctionalIterator<U>> flatMappingFn) {
+    public <U> Poller<U> flatMap(Function<T, Poller<U>> flatMappingFn) {
         return new FlatMappedPoller<>(this, flatMappingFn);
     }
+
+    @Override
+    public Poller<T> link(Poller<T> toLink) {
+        return new LinkedPoller<>(this, toLink);
+    }
+
+    @Override
+    public abstract void recycle();
 
 }
--- a/common/poller/AbstractPoller.java
+++ b/common/poller/AbstractPoller.java
@@ -21,6 +21,8 @@ package com.vaticle.typedb.core.common.poller;
 import java.util.Optional;
 import java.util.function.Function;
 
+import static com.vaticle.typedb.common.collection.Collections.list;
+
 public abstract class AbstractPoller<T> implements Poller<T> {
 
     @Override
@@ -34,11 +36,8 @@ public abstract class AbstractPoller<T> implements Poller<T> {
     }
 
     @Override
-    public Poller<T> link(Poller<T> toLink) {
-        return new LinkedPoller<>(this, toLink);
+    public Poller<T> link(Poller<T> poller) {
+        return new LinkedPollers<>(list(this, poller));
     }
-
-    @Override
-    public abstract void recycle();
 
 }
--- a/common/poller/BasePoller.java
+++ b/common/poller/BasePoller.java
@@ -1,0 +1,26 @@
+package com.vaticle.typedb.core.common.poller;
+
+import com.vaticle.typedb.core.common.iterator.FunctionalIterator;
+
+import java.util.Optional;
+
+public class BasePoller<T> extends AbstractPoller<T> {
+
+    private final FunctionalIterator<T> source;
+
+    public BasePoller(FunctionalIterator<T> iterator) {
+        source = iterator;
+    }
+
+    @Override
+    public Optional<T> poll() {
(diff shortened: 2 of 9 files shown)


common/poller/AbstractPoller.java
```diff
@@ -18,21 +18,27 @@
...
     }
 
     @Override
-    public <U> Poller<U> flatMap(Function<T, FunctionalIterator<U>> flatMappingFn) {
+    public <U> Poller<U> flatMap(Function<T, Poller<U>> flatMappingFn) {
         return new FlatMappedPoller<>(this, flatMappingFn);
     }
 
+    @Override
+    public Poller<T> link(Poller<T> toLink) {
```
flyingsilverfin, line 37: you should match the naming conventions of the normal `Iterators` - the name here would then just be `poller` for the arg

common/poller/AbstractPoller.java
```diff
@@ -18,21 +18,27 @@
...
+        return new LinkedPoller<>(this, toLink);
+    }
+
+    @Override
+    public abstract void recycle();
```
flyingsilverfin, line 42: we don't actually have to put this here if it's in the Poller interface right?

common/poller/BasePoller.java
```diff
@@ -0,0 +1,44 @@
...
+    private final FunctionalIterator<T> source;
+
+    public BasePoller(FunctionalIterator<T> iterator) {
+        source = iterator;
+    }
+
+    @Override
+    public Optional<T> poll() {
+        if (source.hasNext()) return Optional.of(source.next());
+        source.recycle(); // TODO: Can we call hasNext() after recycle()?
```
flyingsilverfin, line 36: not sure we want to call `recycle()` eagerly here yeah, it should only be recycled when the BasePoller itself is recycled

flyingsilverfin: I also think it's more in our style to write this kinda thing as explicit `if-else` blocks even though both return (haikal's argument is it makes the flow easier to read)

common/poller/FlatMappedPoller.java
```diff
@@ -18,36 +18,43 @@
...
 import java.util.Optional;
 import java.util.function.Function;
 
 public class FlatMappedPoller<T, U> extends AbstractPoller<U> {
 
     private final Poller<T> source;
-    private final Function<T, FunctionalIterator<U>> flatMappingFn;
-    private FunctionalIterator<U> currentIterator;
+    private final Function<T, Poller<U>> flatMappingFn;
+    private final List<Poller<U>> flatMappingPollers;
```
flyingsilverfin, line 30: i believe these are the `flatMappedPollers` right rather than `Mapping` pollers, they are already the outcome of the map operation?

common/poller/FlatMappedPoller.java
```diff
@@ -18,36 +18,43 @@
...
-        else return Optional.empty();
+        for (Poller<U> poller : flatMappingPollers) {
+            Optional<U> next = poller.poll();
+            if (next.isPresent()) return next;
+        }
+        Optional<T> fromSource = source.poll();
+        if (fromSource.isPresent()) {
+            Poller<U> newPoller = flatMappingFn.apply(fromSource.get());
+            flatMappingPollers.add(newPoller);
+            return newPoller.poll();
```
flyingsilverfin, line 48: errr is that enough? if the new poller doesn't have anything you need to loop over the `source` until you've exhausted it, and only if that's all empty and all flatmapped pollers are empty then can you return empty?

common/poller/LinkedPoller.java
```diff
@@ -0,0 +1,43 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package com.vaticle.typedb.core.common.poller;
+
+import java.util.Optional;
+
+public class LinkedPoller<T> extends AbstractPoller<T> {
```
flyingsilverfin, line 23: should again follow the style of `LinkedIterator` if it makes sense I think!
That one actually has an arbitrarily long list of linked iterators :o we should be able to implement it the same here i think

common/poller/Poller.java
```diff
@@ -18,15 +18,17 @@
...
 import java.util.function.Function;
 
 public interface Poller<T> {
 
     Optional<T> poll();
 
-    <U> Poller<U> flatMap(Function<T, FunctionalIterator<U>> flatMappingFn);
+    <U> Poller<U> flatMap(Function<T, Poller<U>> flatMappingFn);
+
+    Poller<T> link(Poller<T> toLink);
```
flyingsilverfin, line 30: same with naming here

flyingsilverfin (changes requested):

common/poller/FlatMappedPoller.java
jmsfltchr: Actually I think this implementation is enough as you first loop over all previously found `flatMappedPollers` (each derived from an answer from the source). Then we also can't "exhaust" the source as it's a poller.

flyingsilverfin (approved):

common/poller/FlatMappedPoller.java
jmsfltchr: Change made, @flyingsilverfin was right!

flyingsilverfin (approved):

PR #6341, opened 2021-05-27
Hugely reduce transaction memory usage

flyingsilverfin:

To enable transactions to traverse and operate over millions of vertices and edges, we cannot cache every vertex and edge read from the storage layer. This PR disables caching vertex (and implicitly edge) caching, allowing traversals of millions of vertices without OOM. Concepts automatically convert their inner `read` vertex to a `write` vertex (which is cached)  whenever a write is submitted onto a concept.

* Create `ThingVertex.Read` and `ThingVertex.Write` split
* `ThingGraph` may _only_ cache `ThingVertex.Write` vertices, ie. ones that have been modified
* by default, we always use a non-cached `read` vertex. Concepts that are performing a write operation convert their vertex to a `ThingVertex.Write` on the fly and transparently
* we eliminate `distinct()` calls which can also cause OOM errors on large answer sets and counts, where possible
* we do not need `distinct` set to be shared between async graph iterators as they all produce from different starting points
* we also no longer track read transactions in the `ConsistencyManager` as they can never generate conflicts, allowing releasing memorying more frequently


flyingsilverfin:
TODO:
in a write transaction, we want to prevent OOM's too in `match-insert` or `match-delete`
We can do this by caching only the outputs of the traversal (eg. answers) and not any intermediate vertices!

We probably also have to cache all retrieved concept API vertices.
Finally, we have decided to keep read and write transactions split - read transactions will never cache anything unless forced to by reasoner, which will enabling retrieving much larger answer sets than even a write transaction can handle and in general reduce memory pressure

rocks/RocksStorage.java
```diff
@@ -339,7 +339,7 @@ public Data(RocksDatabase database, RocksTransaction transaction) {
             this.deletedKeys = new ConcurrentSkipListSet<>();
             this.exclusiveInsertKeys = new ConcurrentSkipListSet<>();
             this.snapshotEnd = null;
-            this.database.consistencyMgr().register(this);
+            if (transaction.type().isWrite()) this.database.consistencyMgr().register(this);
```
flyingsilverfin, line 342: we should only do conflict checking for write transactions!

concept/ConceptManager.java
```diff
@@ -79,14 +78,17 @@ public ConceptManager(GraphManager graphMgr) {
...
     }
 
-    public FunctionalIterator<ConceptMap> conceptMaps(FunctionalIterator<VertexMap> vertexMap) {
-        return vertexMap.map(this::conceptMap);
+    public ConceptMap conceptMap(VertexMap vertexMap) {
+        return conceptMap(vertexMap, true);
     }
 
-    public ConceptMap conceptMap(VertexMap vertexMap) {
+    public ConceptMap conceptMap(VertexMap vertexMap, boolean useCachedVertices) {
```
flyingsilverfin, line 85: when we convert a concept map we can refresh the concepts to be the cached (ie canonical?) ones

concept/type/impl/TypeImpl.java
```diff
@@ -114,7 +114,7 @@ public boolean isAbstract() {
     public abstract FunctionalIterator<? extends Type> getSubtypesExplicit();
 
     <THING> FunctionalIterator<THING> instances(Function<ThingVertex, THING> thingConstructor) {
-        return getSubtypes().flatMap(t -> graphMgr.data().get(t.vertex)).map(thingConstructor);
+        return getSubtypes().flatMap(t -> graphMgr.data().getAll(t.vertex )).map(thingConstructor);
```
flyingsilverfin, line 117: we had to rename this from `get` to `getAll` due to signatures clashing with the new optional booleans

graph/ThingGraph.java
```diff
@@ -117,49 +117,80 @@ public TypeGraph type() {
     }
 
     public ThingVertex get(VertexIID.Thing iid) {
+        return get(iid, true);
+    }
+
+    public ThingVertex get(VertexIID.Thing iid, boolean cacheVertex) {
```
flyingsilverfin, line 123: all getters now have a `cacheVertex` parameter

graph/adjacency/impl/ThingAdjacencyImpl.java
```diff
@@ -66,13 +66,13 @@
     IID[] infixTails(ThingEdge edge) {
         if (edge.encoding().isOptimisation()) {
             if (direction.isOut()) {
-                return new IID[]{edge.outIID().infix().asRolePlayer().tail(), edge.to().iid().prefix(), edge.to().iid().type()};
+                return new IID[]{edge.outIID().infix().asRolePlayer().tail(), edge.toIID().prefix(), edge.toIID().type()};
```
flyingsilverfin, line 69: to avoid infinite circular calls to `to()` when NOT caching (which we don't need to here), we switch from using `to().iid()` to `toIID()`, which doesn't even have to ask the `ThingGraph` to do any conversion of iid to vertex

graph/adjacency/ThingAdjacency.java
```diff
@@ -71,71 +74,84 @@
...
+        FunctionalIterator<ThingEdge> get();
 
-    void deleteAll();
+    }
 
-    ThingEdge cache(ThingEdge edge);
+    interface Read extends ThingAdjacency {
 
-    void remove(ThingEdge edge);
+        // TODO this feels weird, should we just use ThingAdjacency as Read?
```
flyingsilverfin, line 89: do we want this?

graph/iid/IID.java
```diff
@@ -24,7 +24,6 @@
 
     String readableString; // for debugging
     final ByteArray bytes;
-    private int hash = 0;
```
flyingsilverfin, line 27: the hashcode is already cached in the `ByteArray`, so we can save some memory by not double caching it here

graph/vertex/AttributeVertex.java
```diff
@@ -47,6 +46,12 @@
      */
     VALUE value();
 
+    AttributeVertex.Write<VALUE> writable(); // TODO do we want both of these?
+
+    AttributeVertex.Write<VALUE> asWrite();
```
flyingsilverfin, line 51: feels a bit odd to have both of these - but one is a cast and one is a conversion

graph/vertex/impl/AttributeVertexImpl.java
```diff
@@ -27,282 +27,522 @@
...
+                case DATETIME:
+                    return new AttributeVertexImpl.Read.DateTime(graph, iid.asDateTime());
+                default:
+                    assert false;
+                    return null;
+            }
+        }
+
+        @Override
+        public VertexIID.Attribute<VALUE> iid() {
```
flyingsilverfin, line 70: from here downward is a fair amount of duplication, but also omits some things that are in the `AttributeVertexImpl.Write` (eg. `isInferred`)

graph/vertex/impl/ThingVertexImpl.java
```diff
@@ -126,114 +77,254 @@ public boolean isDeleted() {
...
 
+        // TODO do we need both of these methods?
         @Override
-        public void delete() {
-            if (isDeleted.compareAndSet(false, true)) {
-                deleteEdges();
-                deleteVertexFromStorage();
-                deleteVertexFromGraph();
-            }
+        public ThingVertex.Write asWrite() {
```
flyingsilverfin, line 233: same question here with `asWrite` and `writable()`

graph/vertex/impl/ThingVertexImpl.java
```diff
@@ -126,114 +77,254 @@ public boolean isDeleted() {
...
         }
 
         @Override
-        public void setModified() {
-            if (!isModified) {
-                isModified = true;
-                graph.setModified(iid);
+        public ThingVertex.Write writable() {
+            return this;
+        }
```
flyingsilverfin, line 240: same question with `asWrite()` and `writable()`

graph/vertex/impl/VertexImpl.java
```diff
@@ -51,7 +46,7 @@ public String toString() {
     @Override
     public boolean equals(Object object) {
         if (this == object) return true;
-        if (object == null || getClass() != object.getClass()) return false;
+        if (!(object instanceof VertexImpl)) return false; // instanceof includes null check
```
flyingsilverfin, line 49: now that a `ThingVertex` can either a `Read` or `Write`, we need the equality check to treat them as equal! This bug sent me around the houses for a while!

traversal/procedure/VertexProcedure.java
```diff
@@ -106,7 +106,7 @@ public String toString() {
             iterator = iterator.filter(v -> e.isClosure(graphMgr, v, v, params));
         }
 
-        return iterator.map(v -> VertexMap.of(map(pair(vertex.id().asVariable().asRetrievable(), v)))).distinct();
+        return iterator.map(v -> VertexMap.of(map(pair(vertex.id().asVariable().asRetrievable(), v))));
```
flyingsilverfin, line 109: don't need distinct() for vertex procedures, already guaranteed to be unique

concept/thing/impl/AttributeImpl.java
```diff
@@ -67,28 +68,29 @@ private AttributeImpl(AttributeVertex<VALUE> vertex) {
 
     public abstract VALUE getValue();
 
+    @Override
+    protected AttributeVertex.Write<VALUE> vertexWritable() {
```
haikalpribadi, line 71: Method name should be `writableVertex()`

concept/thing/impl/ThingImpl.java
```diff
@@ -78,24 +78,33 @@ public static ThingImpl of(ThingVertex vertex) {
         }
     }
 
+    protected ThingVertex vertex() {
+        return vertex;
+    }
+
+    protected ThingVertex.Write vertexWritable() {
```
haikalpribadi, line 85: Method name should be `writableVertex()`

concept/type/impl/TypeImpl.java
```diff
@@ -81,6 +81,10 @@ public static TypeImpl of(GraphManager graphMgr, TypeVertex vertex) {
         }
     }
 
+    public TypeVertex vertex() {
+        return vertex;
+    }
```
haikalpribadi: We don't need this getter for `TypeImpl` (sub)classes

haikalpribadi (changes requested):

graph/vertex/impl/VertexImpl.java
```diff
@@ -39,10 +38,6 @@ public void iid(VERTEX_IID iid) {
         this.iid = iid;
     }
 
-    public boolean isModified() {
-        return isModified;
-    }
-
```
haikalpribadi, line 45: I think this should stay at `VertexImpl`, and we just override it once at `ThingVertexImpl.Read` to always return `false`.

rocks/RocksTypeDB.java
```diff
@@ -83,7 +83,7 @@ protected RocksTypeDB(Options.Database options, Factory.DatabaseManager database
 
     private BlockBasedTableConfig initRocksDBTableOptions() {
         BlockBasedTableConfig rocksDBTableOptions = new BlockBasedTableConfig();
-        long blockSize = Math.round(Runtime.getRuntime().totalMemory() * 0.4); // TODO: generalise through typedb.properties
+        long blockSize = Math.round(Runtime.getRuntime().freeMemory() * 0.4); // TODO: generalise through typedb.properties
```
haikalpribadi, line 86: This should be `maxMemory()`

haikalpribadi (changes requested):

haikalpribadi (approved):

Issue #6248, opened 2021-03-11
Update error message from internal to user facing: Invalid concept conversion from 'EntityTypeImpl' to 'AttributeType'


daniel-crowe:
## Description

Error message given by Grakn when defining the below schema should be updated to be more user friendly. 
 
## Environment
1. OS (where Grakn server runs): Mac OS 10
2. Grakn version (and platform): Grakn Core 2.0-alpha-9
3. Grakn client: console

## Reproducible Steps
Attempting to define the following schema: 
`define
organisation sub entity,
owns country;
country sub entity;
`
## Expected Output

###Return error message: 
#### Generic: 
`Invalid Type Read: Type <my-schema-type> isa <schema-type> and cannot be owned by a concept type.`

#### Specific: 
`Invalid Type Read: Type 'country' isa entity and cannot be owned by a concept type.`

## Actual Output

Returns error message: 
`Invalid Type Read: The type 'dept_name' does not exist.`

haikalpribadi:
Thanks for the bug report @daniel-crowe, don't forget to remove they curly brackets in the issue template.

With regards to the "Actual Output" above, does it actually say "dept_name"? That term does not exist in the short schema you wrote in the reproducible steps above.

alexjpwalker:
@daniel-crowe Which of the error messages do we get - the message in the issue title (Invalid concept conversion), or the message in the issue body (Invalid type read)?

daniel-crowe:
@flyingsilverfin @alexjpwalker here's the error message in the latest TypeDB version: `[TYR01] Invalid Type Read: Invalid concept conversion from 'EntityTypeImpl' to 'AttributeType'. Please check server logs for the stack trace.`

flyingsilverfin:
Closed with #6347

PR #6347, opened 2021-06-02
Type casting errors include type label

flyingsilverfin:

As identified in #6248 and in several community messages, we clean up UX that is exposed when a user mis-uses a type as a different meta type (eg. uses an entity in place of an attribute). We now include the type label in the error message to allow tracking down the error much more easily.

* Type casting errors include the type label of the type that failed to cast

### DIFF
--- a/concept/type/impl/TypeImpl.java
+++ b/concept/type/impl/TypeImpl.java
@@ -22,6 +22,10 @@ import com.vaticle.typedb.core.common.exception.TypeDBException;
 import com.vaticle.typedb.core.common.iterator.FunctionalIterator;
 import com.vaticle.typedb.core.common.parameters.Label;
 import com.vaticle.typedb.core.concept.ConceptImpl;
+import com.vaticle.typedb.core.concept.type.AttributeType;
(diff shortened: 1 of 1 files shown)


concept/type/impl/TypeImpl.java
```diff
@@ -22,6 +22,10 @@
 import com.vaticle.typedb.core.common.iterator.FunctionalIterator;
 import com.vaticle.typedb.core.common.parameters.Label;
 import com.vaticle.typedb.core.concept.ConceptImpl;
+import com.vaticle.typedb.core.concept.type.AttributeType;
```
flyingsilverfin, line 25: the only awkward think about this implementation is that now an `Impl` depends on a higher-level iterface that is one level higher in the package structure. However, they are still within the same bazel package

haikalpribadi: It's actually fine because the implementation classes do tend to depend on their interfaces already, and not the other way around. In this case, we're just adding more subtypes of the interface that `TypeImpl` depends on, which is `Type.java`.

haikalpribadi (approved):

PR #6351, opened 2021-06-03
Make RocksDatabaseManager not usable if it's already closed

lolski:

We make the `RocksDatabaseManager` stricter - it should reject operations if it's already closed. Additionally, we make `RocksDatabase` more extensible by the subclasses by making the `delete` method `protected.

- Add an `isOpen` flag to indicate the state of the database manager
- Add checks in the methods such that operations are rejected, when `isOpen` is set to `false`
- Make `RocksDatabase::delete` protected so that it can be extended by the subclasses.


flyingsilverfin (approved):
Looks good! fix de bug

PR #6363, opened 2021-06-08
Ensure that the RPC prefetch size is at least 1

flyingsilverfin:

Ensure that the minimum prefetch size is at least 1, so that the user cannot end up in a situation that streams back 0 answers per `res_part` request.

* add a minimum of 1 to the RPC messages that are sent back per request

### DIFF
--- a/server/TransactionService.java
+++ b/server/TransactionService.java
@@ -312,7 +312,7 @@ public class TransactionService implements StreamObserver<TransactionProto.Trans
         }
 
         private void streamResParts() {
-            streamResPartsWhile(i -> i < prefetchSize && iterator.hasNext());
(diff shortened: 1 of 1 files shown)


server/TransactionService.java
```diff
@@ -312,7 +312,7 @@ private boolean isClientCancelled(Throwable error) {
         }
 
         private void streamResParts() {
-            streamResPartsWhile(i -> i < prefetchSize && iterator.hasNext());
+            streamResPartsWhile(i -> i < Math.max(prefetchSize, 1) && iterator.hasNext());
```
alexjpwalker, line 315: Maybe this Math.max should be moved into the constructor?

alexjpwalker (approved):

PR #6369, opened 2021-06-10
Delete and Insert queries are validated once rather than per answer

flyingsilverfin:

We refactor `Deleter`, `Inserter` and `Updater` in order to validate the queries once, before execution, rather than per-answer.

* extract Validation for `Insert` and `Delete` operations into a static validation that runs before each query is executed


haikalpribadi (approved):
Very nice! Clean, clear, and consistent! :)

PR #6373, opened 2021-06-11
Streamline statistics counting and fix ACID with no RocksDB writes

flyingsilverfin:

As of #6306, Attribute PUT operations do a `read` before doing a `write` to rocksDB. We can harness this to only add a attribute PUT count job if the attribute is actually new, dramatically reducing the number of background count operations the statistics thread has to perform. 

This leads to an ACID bug where doing true attribute PUT leads to no actual writes on RocksDB, and therefore no storage version range against which to check for concurrent commits. We fix this by always writing at least one key to RocksDB per write transaction.

* use the `isPersisted()` flag on an attribute vertex to only add a count job when putting an attribute that is actually brand new
* Always write a dummy key when committing a storage transaction. This guarantees that the RocksDB sequence number will increment and lead to a non-zero range that we can use for doing key-conflict detection and guaranteeing ACID safety.

Example of the problematic ACID case when statistics is writing fewer keys and no longer hiding this bug:
1. `name "Alice"` already exists in the storage, which is at sequence number 10
2. open write tx, put `name "Alice"`, commit. After commit, we are still at storage sequence number 10
3. open a write tx, delete `name "Alice"`. commit. This leads to a a conflict, because we opened a transaction at version 10 and a storage committed after or equal to version 10. However, it is not expected to fail due to using serialised transactions!

This becomes a non-issue if the sequence number always increments by at least one, because the second transaction opens at version 11. Then we can use the logic of checking only against storages that are committed greater than a version 10 (rather than equal to as well)

* To implement the dummy write we create a new `Encoding` prefix `System` and namespace `System` for future system-level operations
* Reserve a range of prefixes for use externally (eg. cluster).

### DIFF
--- a/graph/common/Encoding.java
+++ b/graph/common/Encoding.java
@@ -149,7 +149,7 @@ public class Encoding {
     /**
      * The values in this class will be used as 'prefixes' within an IID in the
      * of every object database, and must not overlap with each other.
-     * <p>
+     *
      * The size of a prefix is 1 unsigned byte; i.e. min-value = 0 and max-value = 255.
      */
     public enum Prefix {
@@ -252,7 +252,7 @@ public class Encoding {
     /**
      * The values in this class will be used as 'infixes' between two IIDs of
      * two objects in the database, and must not overlap with each other.
-     * <p>
+     *
      * The size of a prefix is 1 signed byte; i.e. min-value = -128 and max-value = 127.
      */
     public enum Infix {
--- a/graph/common/Encoding.java
+++ b/graph/common/Encoding.java
@@ -131,6 +131,7 @@ public class Encoding {
     }
 
     public enum PrefixType {
+        SYSTEM(-1),
         INDEX(0),
         STATISTICS(1),
         TYPE(2),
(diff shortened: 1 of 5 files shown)


graph/common/Encoding.java
```diff
@@ -131,6 +131,7 @@ public boolean isBackward() {
     }
 
     public enum PrefixType {
+        SYSTEM(-1),
```
haikalpribadi, line 134: This `PrefixType` enum does not need `int key`. We can remove them.

graph/common/Encoding.java
```diff
@@ -191,12 +194,12 @@ public boolean isBackward() {
                 pair(STRUCTURE_RULE.key, STRUCTURE_RULE)
         );
 
-
         private final byte key;
         private final PrefixType type;
         private final ByteArray bytes;
 
         Prefix(int key, PrefixType type) {
+            assert key < 200 : "Core prefix encoding should not use reserved range";
```
haikalpribadi, line 202: Let's replace our usage of "Core" with "TypeDB" going forward. Also, we can make the error message more informative:

> The encoding range of >= 200 is reserved or TypeDB Cluster

graph/common/Encoding.java
```diff
@@ -960,6 +973,29 @@ public ByteArray bytes() {
         }
     }
 
+    public interface System {
+
+        ByteArray bytes();
+
+        enum Core implements System {
```
haikalpribadi, line 980: This can just be `System` enum

haikalpribadi (changes requested):

haikalpribadi (approved):

Issue #6362, opened 2021-06-08
BDD forward-chaining oracle tests


jmsfltchr:
## Description

Bring back the "oracle" tests - the tests where we create a separate, highly simplified, non-performant implementation of a reasoning engine using forward chaining and test against the answers TypeDB gives.

Requires #6361

PR #6374, opened 2021-06-11
Reasoner Correctness Verification

jmsfltchr:

We introduce an internal framework for verifying the soundness and completeness of reasoning by comparing with a naive forward-chaining reasoner.

### Initialisation

Use a simplistic forward-chaining method to create a reference transaction with all inferable facts materialised. While materialising record all of the facts that could be inferred all of the ways they could have been materialised.

### Soundness Verification

Check that all of the lines of reasoning taken are sound.

Run a query in a live reasoning transaction. For each answer explore the explanations depth-first. (A) Look up in the materialisation records of the materialised Tx the combination of when bindings plus the rule used to see that there's exactly one `then` answer. If there is no such lineage then the explanation is unsound and a rule was triggered incorrectly. If there is more than one materialisation then assert something in the framework is wrong.

Take the inferred `then` variable binding from the materialisation information. The inferred concepts in the two transactions can be mapped such that when verifying the next explanation in the tree the `when` is fully bound. Repeat from (A).

### Completeness Verification

From the query under test, split into concludables and find the relevant materialisations for each concludable. Do this in a depth first manner through the rules also. This finds the lines of reasoning needed to arrive at the specific query's answers. 

Starting from the leaf materialisations, check that the rule conclusion that should be reached has a single matching answer when queried in the live reasoning tx (since a rule does a put, and the `when` is fully bound there must be one answer). Next check that the concludable used to reach that rule has one or more answers in the live reasoning tx (since a concludable could be answered by a more specific pattern there can be more than one answer).

This approach has a limitation: we cannot verify concludables or conclusions that contain more than one inferred variable. This includes the cases that an inferred relation owning an inferred attribute, or an inferred relation relates an inferred concept. This is because we cannot also verify the additional inferred concepts without an onerous computational expense.

### BDD Resolution Steps

We also notice that the BDD scenarios all follow a distinct pattern for resolution tests, so we overhaul the steps to simplify them.

### Related Issues and PRs

Solves #6362 and #6384.
Changes to tests are made in https://github.com/vaticle/typedb-behaviour/pull/206

Failing verification steps are present, there's an issue for that here: https://github.com/vaticle/typedb/issues/6396


test/behaviour/resolution/correctness/SoundnessVerifier.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private void verifyAnswer(ConceptMap answer, Transaction tx) {
+        answer.explainables().iterator().forEachRemaining(explainable -> {
+            tx.query().explain(explainable.id()).forEachRemaining(explanation -> {
+                // This check is valid given that there is no mechanism for recursion termination given by the UX of
+                // explanations, so we do it ourselves
+                if (verified.contains(explanation)) return;
+                else verified.add(explanation);
+                verifyAnswer(explanation.conditionAnswer(), tx);
+                verifyExplanation(explanation);
```
flyingsilverfin, line 72: hmm yeah I suppose it reads ok here, but less well in the `Completeness` checker i think!

test/behaviour/resolution/correctness/test/MaterialiserTest.java
```diff
@@ -0,0 +1,127 @@
...
+                        var("continent").isa("continent").has("name", "Europe"),
+                        var().rel("superior", "continent").rel("subordinate", "country").isa("location-hierarchy"),
+                        var().rel("superior", "country").rel("subordinate", "city").isa("location-hierarchy"),
+                        var().rel("superior", "city").rel("subordinate", "area").isa("location-hierarchy")
+                )).asInsert());
+                tx.commit();
+            }
+        }
+    }
+}
```
flyingsilverfin, line 127: ooopaaa

test/behaviour/resolution/ResolutionSteps.java
```diff
@@ -22,145 +22,192 @@
...
+        correctnessVerifier.verifyCompleteness(typeQLQuery);
     }
 
-    private TypeDB.Transaction materialisedDbTxn() {
-        return getTransaction(materialisedSession());
-    }
+    private static void resetDirectory() throws IOException {
+        if (Files.exists(dataDir)) {
+            System.out.println("Database directory exists!");
+            Files.walk(dataDir).sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
```
flyingsilverfin, line 205: out of curiosity why do we do it sorted?

flyingsilverfin (changes requested):

pattern/Conjunction.java
```diff
@@ -153,6 +154,15 @@ public Variable variable(Identifier.Variable identifier) {
         return variableMap.keySet();
     }
 
+    public Set<Identifier.Variable.Retrievable> retrieves() {
```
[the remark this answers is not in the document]
jmsfltchr: Nice

test/behaviour/resolution/correctness/BoundPattern.java
```diff
@@ -0,0 +1,168 @@
...
+
+        public ConceptMap bounds() {
+            return bounds;
+        }
+    }
+
+    static class BoundConcludable {
+        private final BoundConjunction fullyBound;
+        final Concludable concludable;
+        final ConceptMap bounds;
```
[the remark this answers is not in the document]
jmsfltchr: Good spot

test/behaviour/resolution/ResolutionSteps.java
jmsfltchr: No idea, this is copied over from the preexisting steps and I didn't want to mess with it!

flyingsilverfin (approved):

flyingsilverfin:
yuuuuuge. Great work!

PR #6388, opened 2021-06-29
Include attribute equality properties in planner costs

flyingsilverfin:

We identify and correct a mismatch between the execution of a procedure and the planner cost functions - notably the presence of an attribute's value equality predicate, which significantly constraints the search space but was not included in the cost function of traversing `has` edges in either direction.

* generify the numerator or divisor of the `has` cost functions to include the potential presence of attribute equality constraints


flyingsilverfin:
NOTE
-----
We decided to park this and revisit it as part of a general update/cleanup of the Query Planning coefficient computation.

As part of that, we should incorporate better costs to do with estimating the number of times we have to touch the DB (eg. open iterator, length of an iterator, or do a exact `get` lookup). We should also clean up assumptions in the `Thing` edges to do with not having resolved types available - we now operate under the assumption of always having them!

haikalpribadi:
Testing a comment on a PR with checklist:

- [ ] item 1
- [ ] item 2
- [ ] item 3

flyingsilverfin:
Closed - outdated

Issue #6310, opened 2021-05-04
Error during import of data from exported db


nicolamassarenti:
Hi,

This issue is because of an error we are encountering with import/export from Linux Ubuntu 20.04 to Windows 10.

## Environment

1. OS: Linux Ubuntu 20.04, Windows 10
2. Grakn version: `Grakn 2.0.2`

## Details
When exporting, the server printed the following log:
`Exported 1726 entities, 1584 attributes, 8099 relations (20098 roles), 6810 ownerships`

The schema is:
```
define

lemma sub attribute,
    value string;

name sub attribute,
    value string;

relation-type sub attribute,
    value string;

sentiment sub attribute,
    value string;

text sub attribute,
    value string;

token-type sub attribute,
    value string;

category sub relation,
    relates sub-category,
    relates super-category;

class sub relation,
    relates sub-class,
    relates super-class;

containing sub relation,
    relates contained,
    relates container;

element sub relation,
    relates sub-element,
    relates super-element;

instance sub relation,
    relates sub-instance,
    relates super-instance;

mined-relation sub relation,
    owns relation-type,
    relates child,
    relates contained,
    relates head;

property sub relation,
    relates sub-property,
    relates super-property;

synthesis sub relation,
    relates express,
    relates synthesize;

chapter sub entity,
    owns name,
    plays category:super-category,
    plays class:sub-class;

characteristic sub entity,
    owns name,
    plays property:sub-property;

concept sub entity,
    owns name,
    plays element:super-element,
    plays instance:sub-instance;

document sub entity,
    owns name,
    plays class:super-class;

object sub entity,
    owns name,
    plays element:sub-element,
    plays property:super-property,
    plays synthesis:synthesize;

paragraph sub entity,
    owns name,
    plays category:sub-category,
    plays containing:container,
    plays instance:super-instance;

sentence sub entity,
    owns sentiment,
    owns text,
    plays containing:contained,
    plays containing:container,
    plays mined-relation:contained,
    plays synthesis:express;

token sub entity,
    owns lemma,
    owns token-type,
    plays containing:contained,
    plays mined-relation:child,
    plays mined-relation:head;

```

The schema is imported and exported with no problems.

I exported (from Ubuntu) the data command `grakn server export <DB_NAME> data.grakn`.

My colleague imported the data with command `grakn server import <DB_NAME> data.grakn` and got some errors. We repeted the import and got different errors. They are shown in the pictures below:

[...]

PR #6397, opened 2021-07-09
Implement scalable import and correct export functionality

flyingsilverfin:

We fix some stability issues in the import and export process that were indicated by #6310. Imports are now also multi-threaded, multi-transactional, and can scales to huge imports without OOM.

* correctly clean up session/transaction resources at the end of an import
* rewrite `Import` to be multi-threaded and write ID mappings to a rocksDB instance to avoid running out of memory
* since on commit new IIDs are rewritten, we expose the mapping from buffered to persisted IIDs as a `RocksTransaction` API
* add missing header to exported data file
* use `toHexString()` for IID encoding, instead of `encodeString()` -- this is because UTF_8 encoding converts invalid bytes sequences (which we have in our IID byte arrays) with standard sequences. This can cause corruption because two concepts can end up with the same serialised string ID
*  make progress logging more granular with individual types tracked
* replace the `//test/integration/migration/data.typedb` file with a correct export containing both header and non-clashing IDs
* update and remove most rules from the test schema to speed up the test execution (rule validation can take around 20 seconds)


flyingsilverfin:
TODO: we can still end up an error state if more than one BATCH_SIZE are imported in the transaction, because of the ID rewrites we do

migrator/DataImporter.java
```diff
@@ -34,286 +35,539 @@
...
-                    attribute = attributeType.asLong().put(valueMsg.getLong());
-                    break;
-                case DOUBLE:
-                    attribute = attributeType.asDouble().put(valueMsg.getDouble());
-                    break;
-                case DATETIME:
-                    attribute = attributeType.asDateTime().put(
-                            Instant.ofEpochMilli(valueMsg.getDatetime()).atZone(ZoneId.of("Z")).toLocalDateTime());
-                    break;
+    private class InitAndAttributes extends ImportWorker {
```
flyingsilverfin: in the first pass of the file we load attributes, and search for the checksum at the end of the file

test/integration/migrator/MigratorTest.java
```diff
@@ -48,7 +48,7 @@
     private static final Path logDir = dataDir.resolve("logs");
     private static final Database options = new Database().dataDir(dataDir).logsDir(logDir);
     private static final String database = "typedb";
-    private static final Path schemaPath = Paths.get("test/integration/migrator/schema.gql");
+    private static final Path schemaPath = Paths.get("test/integration/migrator/schema.tql");
     private final Path dataPath = Paths.get("test/integration/migrator/data.typedb");
```
lolski, line 52: we're adding this file `test/integration/migrator/data.typedb 
`, of size 6.25MB, intentionally, right?

flyingsilverfin: yes, wanted something a bit more complete

migrator/BUILD
```diff
@@ -63,7 +70,17 @@ native_java_libraries(
     native_libraries_deps = [
         "//:typedb",
         "//concept:concept",
-        "//logic:logic"
+        "//logic:logic",
+        "//rocks:rocks",
```
flyingsilverfin, line 74: awkward to depend on rocks - should we be putting the APIs on the Transaction interface that we require?

Issue #6356, opened 2021-06-07
Reasoner Architectural Refactor


jmsfltchr:
## Problem to Solve

We want to achieve a cleaner architecture for the actors of the reasoner. Improving this cleanliness makes it faster to understand this complex domain and increases the ease with which we can add features.

Presently we find that the storyline is unclear and there are code smells, for example actors carrying 

Create a better storyline by better organising the logic of the actors of the reasoner. We should find that the whole architecture is far easier to understand and to work with. We want: a clearer separation between the processing of match and explain queries, 

## Proposed Solution

- [x] We realise that there is too much state held in many of resolvers. This state can be divided by the fixed variable bounds that are being processed. Over the course of these PRs we introduce a "bound" resolver as a complement to the retrievable, concludable and conclusion actors. These original actors become coordinators. In doing so we remove subsumption, which would otherwise cross actor boundaries. Subsumption was previously handled by the cache, which overcomplicated the cache significantly.
  - #6403
  - #6408
  - #6410
  - #6416
  - #6427
- [x] In order to find all facts, the reasoner was using an outer _reiteration_ to re-search all facts under certain conditions in case more answers could be derived. This process could often result in a linear increase in query latency, we believe a typical 2x slowdown for queries with rule cycles. Besides the performance impact, this method causes significant headaches in the code. It's an extra complexity that we would like to alleviate. In #6429 we introduce a new replacement feature, which detects when an infinite cycle has been blocked, and re-explores accordingly to find new facts that have become inferable. This PR addressed:
  - #6357 
  - #6278 
- [x] While working on #6429 we needed to add a significant feature to the `ResolutionTracer` to allow tracing in a fully parallel actor setting. This required all messages sent between actors to be traced. 
- [x] The model of the request tracing introduced was unsatisfactory in terms of traced requests vs. request factories. Pulling on this string we found that `RequestState` should not be tied to a request, but instead tracks the resolution of a particular state, now called `ResolutionState`. Not only this, but the `BoundConcludableResolver` in particular changes a lot, and becomes two actors, `Blocked` and `Exploring`. Both are used if the concludable is part of a cycle: #6448.
- [ ] #6453 partly addressed by https://github.com/vaticle/typedb/pull/6456
- [ ] #6451

We believe that we have found a better message passing model for the reasoner. If this model is successful then this work will be delayed due to being unnecessary or unhelpful:
- [ ] Create cache actors to address https://github.com/vaticle/typedb/issues/6450
[...]

PR #6403, opened 2021-07-19
Split Retrievable Resolver into more actors

jmsfltchr:

We start to devolve bloated actors in the reasoner to offload work to more actors. We start with actors that resolve retrievables, splitting the work for each set of bounds into a new actor, with one actor to coordinate them.

- We split the content of RetrievableResolver so that what were RequestStates are now separate actors.
- We keep subsumption active by creating specialised request types to indicate that subsumption should be used. The RetrievableResolver actor coordinates subsumption, and directs the appropriate requests to the BoundRetrievableResolver actors.

Part of #6356

### DIFF
--- a/common/poller/FilteredPoller.java
+++ b/common/poller/FilteredPoller.java
@@ -19,19 +19,29 @@
 package com.vaticle.typedb.core.common.poller;
 
 import java.util.Optional;
-import java.util.function.Function;
 import java.util.function.Predicate;
 
-public interface Poller<T> {
+public class FilteredPoller<T> extends AbstractPoller<T> {
 
-    Optional<T> poll();
+    private final Poller<T> poller;
+    private final Predicate<T> predicate;
 
-    <U> Poller<U> flatMap(Function<T, Poller<U>> mappingFn);
+    public FilteredPoller(Poller<T> poller, Predicate<T> predicate) {
+        this.poller = poller;
+        this.predicate = predicate;
+    }
 
-    Poller<T> link(Poller<T> poller);
+    @Override
+    public Optional<T> poll() {
+        Optional<T> fromSource;
+        while ((fromSource = poller.poll()).isPresent()) {
+            if (predicate.test(fromSource.get())) return fromSource;
+        }
+        return Optional.empty();
+    }
 
-    Poller<T> filter(Predicate<T> predicate);
-
-    void recycle();
-
+    @Override
+    public void recycle() {
+        poller.recycle();
+    }
 }
--- a/reasoner/resolution/ResolverRegistry.java
+++ b/reasoner/resolution/ResolverRegistry.java
@@ -194,7 +194,7 @@ public class ResolverRegistry {
     public Actor.Driver<BoundRetrievableResolver> registerBoundRetrievable(Retrievable retrievable, ConceptMap bounds) {
         LOG.debug("Register BoundRetrievableResolver, pattern: {} bounds: {}", retrievable.pattern(), bounds);
         Actor.Driver<BoundRetrievableResolver> resolver = Actor.driver(driver -> new BoundRetrievableResolver(
-                retrievable, bounds, driver, this, traversalEngine, conceptMgr, resolutionTracing
+                driver, retrievable, bounds, this, traversalEngine, conceptMgr, resolutionTracing
         ), executorService);
         resolvers.add(resolver);
         if (terminated.get()) throw TypeDBException.of(RESOLUTION_TERMINATED); // guard races without synchronized
--- a/reasoner/resolution/ResolverRegistry.java
+++ b/reasoner/resolution/ResolverRegistry.java
@@ -21,6 +21,7 @@ package com.vaticle.typedb.core.reasoner.resolution;
 import com.vaticle.typedb.common.collection.ConcurrentSet;
 import com.vaticle.typedb.core.common.exception.TypeDBException;
 import com.vaticle.typedb.core.concept.ConceptManager;
+import com.vaticle.typedb.core.concept.answer.ConceptMap;
 import com.vaticle.typedb.core.concurrent.actor.Actor;
 import com.vaticle.typedb.core.concurrent.actor.ActorExecutorGroup;
 import com.vaticle.typedb.core.logic.LogicManager;
@@ -28,12 +29,14 @@ import com.vaticle.typedb.core.logic.Rule;
 import com.vaticle.typedb.core.logic.resolvable.Concludable;
 import com.vaticle.typedb.core.logic.resolvable.Negated;
 import com.vaticle.typedb.core.logic.resolvable.Resolvable;
+import com.vaticle.typedb.core.logic.resolvable.Retrievable;
 import com.vaticle.typedb.core.pattern.Conjunction;
 import com.vaticle.typedb.core.pattern.Disjunction;
 import com.vaticle.typedb.core.pattern.equivalence.AlphaEquivalence;
 import com.vaticle.typedb.core.reasoner.resolution.answer.AnswerState.Top.Explain;
 import com.vaticle.typedb.core.reasoner.resolution.answer.AnswerState.Top.Match;
 import com.vaticle.typedb.core.reasoner.resolution.framework.Resolver;
+import com.vaticle.typedb.core.reasoner.resolution.resolver.BoundRetrievableResolver;
 import com.vaticle.typedb.core.reasoner.resolution.resolver.ConcludableResolver;
 import com.vaticle.typedb.core.reasoner.resolution.resolver.ConclusionResolver;
 import com.vaticle.typedb.core.reasoner.resolution.resolver.ConditionResolver;
@@ -186,6 +189,16 @@ public class ResolverRegistry {
         resolvers.add(resolver);
         if (terminated.get()) throw TypeDBException.of(RESOLUTION_TERMINATED); // guard races without synchronized
         return ResolverView.retrievable(resolver, retrievable.retrieves());
+    }
+
+    public Actor.Driver<BoundRetrievableResolver> registerBoundRetrievable(Retrievable retrievable, ConceptMap bounds) {
+        LOG.debug("Register BoundRetrievableResolver, pattern: {} bounds: {}", retrievable.pattern(), bounds);
+        Actor.Driver<BoundRetrievableResolver> resolver = Actor.driver(driver -> new BoundRetrievableResolver(
+                retrievable, bounds, driver, this, traversalEngine, conceptMgr, resolutionTracing
+        ), executorService);
+        resolvers.add(resolver);
+        if (terminated.get()) throw TypeDBException.of(RESOLUTION_TERMINATED); // guard races without synchronized
+        return resolver;
     }
 
     // note: must be thread safe. We could move to a ConcurrentHashMap if we create an alpha-equivalence wrapper
--- a/reasoner/resolution/answer/AnswerStateImpl.java
+++ b/reasoner/resolution/answer/AnswerStateImpl.java
@@ -1004,6 +1004,11 @@ public abstract class AnswerStateImpl implements AnswerState {
             }
 
             @Override
+            public Retrievable<P> with(ConceptMap extension) {
+                return new RetrievableImpl<>(filter, parent(), extendAnswer(extension), root());
+            }
+
+            @Override
             public boolean equals(Object o) {
                 if (this == o) return true;
                 if (o == null || getClass() != o.getClass()) return false;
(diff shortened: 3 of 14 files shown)


common/poller/FilteredPoller.java
```diff
@@ -0,0 +1,47 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package com.vaticle.typedb.core.common.poller;
+
+import java.util.Optional;
+import java.util.function.Predicate;
+
+public class FilteredPoller<T> extends AbstractPoller<T> {
```
jmsfltchr, line 24: In fact we don't need the `FilteredPoller` in the end - think we should keep it anyway?

reasoner/resolution/answer/AnswerStateImpl.java
```diff
@@ -1003,6 +1003,11 @@ public P aggregateToUpstream(ConceptMap concepts) {
                 return parent().with(concepts);
             }
 
+            @Override
+            public Retrievable<P> with(ConceptMap extension) {
+                return new RetrievableImpl<>(filter, parent(), extendAnswer(extension), root());
+            }
+
```
jmsfltchr, line 1010: Needed so that we can pass from a subsumer to the resolver it subsumes without the answerstate following the normal `upstream` pattern.

reasoner/resolution/framework/AnswerCache.java
```diff
@@ -37,42 +37,51 @@
...
     protected boolean complete;
     protected final Map<SUBSUMES, ? extends AnswerCache<?, SUBSUMES>> cacheRegister;
     protected final ConceptMap state;
+    private final Supplier<FunctionalIterator<ANSWER>> answerSourceSupplier;
+    private boolean sourceCleared;
+    private boolean sourceExhausted;
 
-    protected AnswerCache(Map<SUBSUMES, ? extends AnswerCache<?, SUBSUMES>> cacheRegister, ConceptMap state) {
+    protected AnswerCache(Map<SUBSUMES, ? extends AnswerCache<?, SUBSUMES>> cacheRegister, ConceptMap state,
+                          Supplier<FunctionalIterator<ANSWER>> answerSourceSupplier) {
```
jmsfltchr, line 62: We only ever use one source, and it's always an iterator, so that lets us simplify. Mainly doing this because we need it. It means that we are able to kill the traversal when we do subsumption, where previously we got the cache to set itself as complete because it was aware of the other caches.

reasoner/resolution/framework/Request.java
```diff
@@ -95,4 +98,86 @@ public String toString() {
...
+
+    public boolean isToSubsumer() {
+        return false;
+    }
+
+    public ToSubsumer asToSubsumer() {
+        throw TypeDBException.of(ILLEGAL_STATE);
+    }
+
+    public static class ToSubsumed extends Request {
```
jmsfltchr, line 117: These are the extra request types that can tell a resolver where to go next for subsumption answers

reasoner/resolution/ResolverRegistry.java
```diff
@@ -188,6 +191,16 @@ public ResolverView registerResolvable(Resolvable<?> resolvable) {
         return ResolverView.retrievable(resolver, retrievable.retrieves());
     }
 
+    public Actor.Driver<BoundRetrievableResolver> registerBoundRetrievable(Retrievable retrievable, ConceptMap bounds) {
```
jmsfltchr, line 194: We need an elegant way to terminate these

reasoner/resolution/resolver/ConcludableResolver.java
```diff
@@ -255,7 +258,7 @@ private void nextAnswer(Request fromUpstream, CachingRequestState<?, ConceptMap>
                     ConcludableAnswerCache answerCache = new ConcludableAnswerCache(cacheRegister, answerFromUpstream);
                     cacheRegister.put(answerFromUpstream, answerCache);
                     requestState = new Explain(fromUpstream, answerCache, iteration);
-                    registerRules(fromUpstream, requestState.asExploration());
+                    requestState.asExploration().downstreamManager().addDownstreams(ruleDownstreams(fromUpstream));
```
jmsfltchr, line 261: Just some tidying in this class

reasoner/resolution/resolver/RetrievableResolver.java
```diff
@@ -69,89 +64,74 @@ public void receiveRequest(Request fromUpstream, int iteration) {
...
+            Driver<BoundRetrievableResolver> boundRetrievable = getOrReplaceBoundRetrievable(root, bounds);
+            // TODO: Re-enable subsumption when async bug is fixed
+            // Optional<ConceptMap> subsumer = subsumptionTrackers.computeIfAbsent(
+            //         root, r -> new SubsumptionTracker()).getSubsumer(bounds);
+            // // If there is a finished subsumer, let the BoundRetrievable know that it can go there for answers
+            // Request request = subsumer
+            //         .map(conceptMap -> Request.ToSubsumed.create(
+            //                 driver(), boundRetrievable, boundRetrievablesByRoot.get(root).get(conceptMap),
+            //                 fromUpstream.partialAnswer()))
+            //         .orElseGet(() -> Request.create(driver(), boundRetrievable, fromUpstream.partialAnswer()));
```
jmsfltchr, line 84: How subsumption works when enabled

reasoner/resolution/resolver/SubsumptionTracker.java
```diff
@@ -0,0 +1,88 @@
...
+import com.vaticle.typedb.core.concept.answer.ConceptMap;
+import com.vaticle.typedb.core.traversal.common.Identifier;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+class SubsumptionTracker {
```
jmsfltchr, line 30: If we do manage to create an abstract `Coordinator` class, perhaps with a subclass which is a `SubsumptiveCoordinator` then `SubsumptionTracker` can be an inner class there. For now it's a bit random where it sits, happy enough for that while we change things.

reasoner/resolution/resolver/BoundRetrievableResolver.java
```diff
@@ -0,0 +1,174 @@
...
+    @Override
+    public void receiveRequest(Request fromUpstream, int iteration) {
+        if (fromUpstream.isToSubsumed()) {
+            assert fromUpstream.partialAnswer().conceptMap().equals(bounds);
+            receiveSubsumedRequest(fromUpstream.asToSubsumed(), iteration);
+        } else if (fromUpstream.isToSubsumer()) {
+            receiveSubsumerRequest(fromUpstream.asToSubsumer(), iteration);
+        } else {
+            assert fromUpstream.partialAnswer().conceptMap().equals(bounds);
+            receiveDirectRequest(fromUpstream, iteration);
```
jmsfltchr, line 70: This is the (not impressively named) path for non-subsumptive requests

reasoner/resolution/resolver/BoundRetrievableResolver.java
```diff
@@ -0,0 +1,174 @@
...
+                fromUpstream, request -> new SubsumerRequestState(request, cache, iteration)));
+    }
+
+    private void receiveDirectRequest(Request fromUpstream, int iteration) {
+        sendAnswerOrFail(fromUpstream, iteration, requestStates.computeIfAbsent(
+                fromUpstream, request -> new BoundRequestState(request, cache, iteration)));
+    }
+
+    @Override
+    protected void receiveAnswer(Response.Answer fromDownstream, int iteration) {
```
jmsfltchr, line 102: Only for subsumption

reasoner/resolution/resolver/BoundRetrievableResolver.java
```diff
@@ -0,0 +1,174 @@
...
+            if (upstreamAnswer.isPresent()) {
+                answerToUpstream(upstreamAnswer.get(), fromUpstream, iteration);
+            } else {
+                requestFromSubsumer(fromUpstream, iteration);
+            }
+        }
+    }
+
+    @Override
+    protected void receiveFail(Response.Fail fromDownstream, int iteration) {
```
jmsfltchr, line 117: Only for subsumption

reasoner/resolution/framework/AnswerCache.java
```diff
@@ -37,42 +37,51 @@
...
+    protected FunctionalIterator<ANSWER> answerSource;
     protected boolean complete;
     protected final Map<SUBSUMES, ? extends AnswerCache<?, SUBSUMES>> cacheRegister;
     protected final ConceptMap state;
+    private final Supplier<FunctionalIterator<ANSWER>> answerSourceSupplier;
+    private boolean sourceCleared;
+    private boolean sourceExhausted;
 
-    protected AnswerCache(Map<SUBSUMES, ? extends AnswerCache<?, SUBSUMES>> cacheRegister, ConceptMap state) {
+    protected AnswerCache(Map<SUBSUMES, ? extends AnswerCache<?, SUBSUMES>> cacheRegister, ConceptMap state,
```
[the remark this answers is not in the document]
flyingsilverfin: we decided that we will pass over this for now and see how it evolves over the refactor

reasoner/resolution/resolver/RetrievableResolver.java
```diff
@@ -69,89 +64,74 @@ public void receiveRequest(Request fromUpstream, int iteration) {
...
 
-    @Override
-    protected void receiveAnswer(Answer fromDownstream, int iteration) {
-        throw TypeDBException.of(ILLEGAL_STATE);
-    }
-
-    @Override
-    protected void receiveFail(Response.Fail fromDownstream, int iteration) {
-        throw TypeDBException.of(ILLEGAL_STATE);
+    private Driver<BoundRetrievableResolver> getOrReplaceBoundRetrievable(Driver<? extends Resolver<?>> root, ConceptMap bounds) {
```
flyingsilverfin, line 98: replace?

reasoner/resolution/resolver/SubsumptionTracker.java
```diff
@@ -0,0 +1,88 @@
...
+        this.finishedStates = new HashSet<>();
+        this.subsumersMap = new HashMap<>();
+        this.finishedMapping = new HashMap<>();
+    }
+
+    public void addFinished(ConceptMap conceptMap) {
+        this.finishedStates.add(conceptMap);
+    }
+
+    public Optional<ConceptMap> getSubsumer(ConceptMap unfinished) {
```
flyingsilverfin, line 46: `getSubsumer` doesn't sound like it has any notion of finish/unfinished. 
This class from outside appears to have only to do with concept maps being subsets of other ones, right?

Maybe you already want to turn this into a `SubsumptiveCoordinator` which possibly has a `SubsumptionTracker` in it? The `Tracker` only knows about relationships between concept maps. 
The coordinator keeps track of finished/unfinished states and is roughly aware of "resolvers" answering queries and being complete or incomplete.

In other words the story/abstraction still feels a bit rough. How would you explain it in words?

flyingsilverfin (changes requested):
I'm a bit conflicted about how to proceed - i can't quite keep up with your reviews, especially since I'm remote right now and it probably warrants discussion. I'd probably say it's better to discuss the model you're implementing and then just do it, and we can do a pair review at certain checkpoints?

reasoner/resolution/ResolverRegistry.java
jmsfltchr: Ah yes, currently we don't terminate the bound resolvers when we're actually done with them, because they are bound to a root and iteration.

reasoner/resolution/resolver/RetrievableResolver.java
jmsfltchr: Yes I think so...

reasoner/resolution/resolver/RetrievableResolver.java
```diff
@@ -69,89 +64,74 @@ public void receiveRequest(Request fromUpstream, int iteration) {
...
+            requestMapByRoot.computeIfAbsent(root, r -> new HashMap<>()).put(request, fromUpstream);
+            requestFromDownstream(request, fromUpstream, iteration);
         }
     }
 
     private void prepareNextIteration(Driver<? extends Resolver<?>> root, int iteration) {
         iterationByRoot.put(root, iteration);
-        cacheRegistersByRoot.get(root).clear();
+        boundRetrievablesByRoot.remove(root);
+        subsumptionTrackers.remove(root);
```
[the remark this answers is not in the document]
jmsfltchr: So we've decided we're keeping this for as long as it isn't getting in the way, since it nearly works.

reasoner/resolution/resolver/SubsumptionTracker.java
jmsfltchr: The `SubsumptionTracker` class isn't doing much if it isn't tracking which subsumers are finished and which finished subsumers we've already found in the past. I've moved it to be an inner class and tweaked `getSubsumer` to `getFinishedSubsumer` for clarity. Unsure of the next move.

PR #6408, opened 2021-07-22
Refactor Reasoner Concludable Actor

jmsfltchr:

We break down the actor that handles concludable parts of a query. It becomes one actor for each partial answer, plus one coordinating actor.

- Split `ConcludableResolver` into `ConcludableResolver` and `BoundConcludableResolver`. 
- This required moving the reiteration messaging down into the BoundCOncludableResolvers.
- `ConcludableResolver` is almost identical to `RetrievableResolver` now so should be abstracted.

### DIFF
--- a/reasoner/ReasonerProducer.java
+++ b/reasoner/ReasonerProducer.java
@@ -140,7 +140,7 @@ public class ReasonerProducer implements Producer<ConceptMap> {
     private void requestFailed(int iteration) {
         LOG.trace("Failed to find answer to request in iteration: " + iteration);
         if (options.traceInference()) ResolutionTracer.get().finish();
-        if (resolverRegistry.boundConcludableResolvers().size() == 0) {
+        if (resolverRegistry.reiterationQueryRespondents(rootResolver, iteration).size() == 0) {
             finish();
         } else if (!sentReiterationRequests && iteration == this.iteration) {
             sendReiterationRequests();
@@ -148,17 +148,27 @@ public class ReasonerProducer implements Producer<ConceptMap> {
     }
 
     private void finish() {
-        // query is completely terminated
-        done = true;
-        queue.done();
-        required.set(0);
+        if (!done) {
+            // query is completely terminated
+            done = true;
+            queue.done();
+            required.set(0);
+        }
+    }
+
+    private void exception(Throwable e) {
+        if (!done) {
+            done = true;
+            required.set(0);
+            queue.done(e);
+        }
     }
 
     private void sendReiterationRequests() {
         assert reiterationQueryRespondents == null || reiterationQueryRespondents.isEmpty();
         sentReiterationRequests = true;
-        reiterationQueryRespondents = new HashSet<>(resolverRegistry.boundConcludableResolvers());
-        resolverRegistry.boundConcludableResolvers()
+        reiterationQueryRespondents = new HashSet<>(resolverRegistry.reiterationQueryRespondents(rootResolver, iteration));
+        resolverRegistry.reiterationQueryRespondents(rootResolver, iteration)
                 .forEach(res -> res.execute(actor -> actor.receiveReiterationQuery(reiterationRequest)));
     }
 
@@ -174,14 +184,6 @@ public class ReasonerProducer implements Producer<ConceptMap> {
             } else {
                 finish();
             }
-        }
-    }
-
-    private void exception(Throwable e) {
-        if (!done) {
-            done = true;
-            required.set(0);
-            queue.done(e);
         }
     }
 
--- a/reasoner/ReasonerProducer.java
+++ b/reasoner/ReasonerProducer.java
@@ -33,14 +33,11 @@ import com.vaticle.typedb.core.reasoner.resolution.framework.Request;
 import com.vaticle.typedb.core.reasoner.resolution.framework.ResolutionTracer;
 import com.vaticle.typedb.core.reasoner.resolution.framework.Resolver;
 import com.vaticle.typedb.core.traversal.common.Identifier;
-import com.vaticle.typeql.lang.pattern.variable.UnboundVariable;
-import com.vaticle.typeql.lang.query.TypeQLMatch;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.concurrent.ThreadSafe;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -125,8 +122,7 @@ public class ReasonerProducer implements Producer<ConceptMap> {
     }
 
     @Override
-    public void recycle() {
-    }
+    public void recycle() {}
 
     // note: root resolver calls this single-threaded, so is thread safe
     private void requestAnswered(Finished answer) {
@@ -144,7 +140,7 @@ public class ReasonerProducer implements Producer<ConceptMap> {
     private void requestFailed(int iteration) {
         LOG.trace("Failed to find answer to request in iteration: " + iteration);
         if (options.traceInference()) ResolutionTracer.get().finish();
-        if (resolverRegistry.concludableResolvers().size() == 0) {
+        if (resolverRegistry.boundConcludableResolvers().size() == 0) {
             finish();
         } else if (!sentReiterationRequests && iteration == this.iteration) {
             sendReiterationRequests();
@@ -161,8 +157,8 @@ public class ReasonerProducer implements Producer<ConceptMap> {
     private void sendReiterationRequests() {
         assert reiterationQueryRespondents == null || reiterationQueryRespondents.isEmpty();
         sentReiterationRequests = true;
-        reiterationQueryRespondents = new HashSet<>(resolverRegistry.concludableResolvers());
-        resolverRegistry.concludableResolvers()
+        reiterationQueryRespondents = new HashSet<>(resolverRegistry.boundConcludableResolvers());
+        resolverRegistry.boundConcludableResolvers()
                 .forEach(res -> res.execute(actor -> actor.receiveReiterationQuery(reiterationRequest)));
     }
 
--- a/reasoner/resolution/resolver/BoundConcludableResolver.java
+++ b/reasoner/resolution/resolver/BoundConcludableResolver.java
@@ -1,0 +1,398 @@
+package com.vaticle.typedb.core.reasoner.resolution.resolver;
+
+import com.vaticle.typedb.core.common.exception.TypeDBException;
+import com.vaticle.typedb.core.common.iterator.FunctionalIterator;
+import com.vaticle.typedb.core.common.iterator.Iterators;
+import com.vaticle.typedb.core.concept.ConceptManager;
+import com.vaticle.typedb.core.concept.answer.ConceptMap;
+import com.vaticle.typedb.core.logic.Rule;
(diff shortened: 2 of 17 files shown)


jmsfltchr:
Needs a rebase before reviewing!

reasoner/resolution/resolver/BoundConcludableResolver.java
```diff
@@ -0,0 +1,415 @@
...
+    private final Set<Identifier.Variable.Retrievable> unboundVars;
+    private final ConceptMap bounds;
+    private final Map<Driver<ConclusionResolver>, Rule> resolverRules;
+    private final LinkedHashMap<Driver<ConclusionResolver>, Set<Unifier>> applicableRules;
+    private final Map<Request, RequestState.CachingRequestState<?, ConceptMap>> requestStates;
+    private RequestState exploringRequestState;
+
+    public BoundConcludableResolver(Driver<BoundConcludableResolver> driver, Concludable concludable, ConceptMap bounds,
+                                    Map<Driver<ConclusionResolver>, Rule> resolverRules,
+                                    LinkedHashMap<Driver<ConclusionResolver>, Set<Unifier>> applicableRules,
```
flyingsilverfin, line 70: any reason we are passing in applicable rules if we have it in the `Concludable`?

flyingsilverfin: the set of data structures is odd. The `concludable` can you get you applicable rules with unifiers. The `registry` can you `ConclusionResolvers` -- why do you need either of these maps (`resolverRules` and `applicableRules`?)

reasoner/resolution/resolver/BoundConcludableResolver.java
```diff
@@ -0,0 +1,415 @@
...
+                    requestState.answerCache().setComplete(); // TODO: The cache should not be set as complete during recursion
+                    failToUpstream(fromUpstream, iteration);
+                }
+            } else {
+                failToUpstream(fromUpstream, iteration);
+            }
+        }
+    }
+
+
```
flyingsilverfin, line 192: extra newline

reasoner/resolution/resolver/BoundConcludableResolver.java
```diff
@@ -0,0 +1,415 @@
...
+    private void requestFromSubsumer(Request.ToSubsumed fromUpstream, int iteration) {
+        Request toSubsumer = Request.ToSubsumer.create(driver(), fromUpstream.subsumer(),
+                                                       fromUpstream, fromUpstream.partialAnswer());
+        requestFromDownstream(toSubsumer, fromUpstream, iteration);
+    }
+
+    @Override
+    protected void initialiseDownstreamResolvers() {
+        throw TypeDBException.of(ILLEGAL_STATE);
+    }
```
flyingsilverfin, line 202: this is a code smell now. I'd expect by the end of the refactor this kind of thing will disappear naturally though?

reasoner/resolution/resolver/BoundConcludableResolver.java
```diff
@@ -0,0 +1,415 @@
...
+
+    public void receiveReiterationQuery(ReiterationQuery.Request request) {
+        driver().execute(actor -> actor.respondToReiterationQuery(request, cache.requiresReiteration()));
+    }
+
+    private void respondToReiterationQuery(ReiterationQuery.Request request, boolean reiterate) {
+        request.onResponse().accept(ReiterationQuery.Response.create(driver(), reiterate));
+    }
```
flyingsilverfin, line 210: what's the point of the extra method here? we can just respond immediately in `receiveReiterationQuery` no?

reasoner/resolution/resolver/BoundConcludableResolver.java
```diff
@@ -0,0 +1,415 @@
...
+import java.util.Optional;
+import java.util.Set;
+
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+
+public class BoundConcludableResolver extends Resolver<BoundConcludableResolver>  {
+
+    private static final Logger LOG = LoggerFactory.getLogger(BoundConcludableResolver.class);
+    private AnswerCache<?, ConceptMap> cache;
```
flyingsilverfin, line 59: non-final should come after finals for ease of reading

reasoner/resolution/resolver/BoundConcludableResolver.java
```diff
@@ -0,0 +1,415 @@
...
+        List<Request> downstreams = new ArrayList<>();
+        AnswerState.Partial.Concludable<?> partialAnswer = fromUpstream.partialAnswer().asConcludable();
+        for (Map.Entry<Driver<ConclusionResolver>, Set<Unifier>> entry : applicableRules.entrySet()) {
+            Driver<ConclusionResolver> conclusionResolver = entry.getKey();
+            for (Unifier unifier : entry.getValue()) {
+                Optional<? extends AnswerState.Partial.Conclusion<?, ?>> unified = partialAnswer.toDownstream(
+                        unifier, resolverRules.get(conclusionResolver));
+                unified.ifPresent(
+                        conclusion -> downstreams.add(Request.create(driver(), conclusionResolver, conclusion)));
+            }
```
flyingsilverfin, line 298: this will get simpler without `applicableRules` and `resolveRules`

also, is the comment above still relevant? I pulled your branch and it looks like the loop termination now happens when we create `Following` request states right?

reasoner/resolution/resolver/BoundConcludableResolver.java
```diff
@@ -0,0 +1,415 @@
...
+            this.singleAnswerRequired = singleAnswerRequired;
+        }
+
+        @Override
+        public boolean isExploration() {
+            return true;
+        }
+
+        @Override
+        public Exploration asExploration() {
```
flyingsilverfin, line 335: fyi  i think it looked like `Exploration` is only used in `BoundConcludable` -- do we want to move it here?

reasoner/resolution/resolver/BoundRetrievableResolver.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
+
+public class BoundRetrievableResolver extends Resolver<BoundRetrievableResolver> {
+    private final ConceptMapCache cache;
+    private final Map<Request, RequestState> requestStates;
+    private ConceptMap bounds;
```
flyingsilverfin, line 45: final

reasoner/resolution/resolver/ConcludableResolver.java
```diff
@@ -92,101 +81,73 @@ public void receiveRequest(Request fromUpstream, int iteration) {
...
     private void prepareNextIteration(Driver<? extends Resolver<?>> root, int iteration) {
         iterationByRoot.put(root, iteration);
-        cacheRegistersByRoot.get(root).clear();
+        cacheRegistersByRoot.remove(root);
+        requestMapByRoot.remove(root);
+        subsumptionTrackers.remove(root);
+        boundConcludableResolversByRoot.remove(root);
+    }
+
+    private Driver<BoundConcludableResolver> getOrReplaceBoundConcludable(Driver<? extends Resolver<?>> root, ConceptMap bounds) {
```
flyingsilverfin, line 116: again, does this replace?

reasoner/resolution/resolver/ConcludableResolver.java
```diff
@@ -92,101 +81,73 @@ public void receiveRequest(Request fromUpstream, int iteration) {
...
-    }
-
-    @Override
-    public void terminate(Throwable cause) {
-        super.terminate(cause);
-        requestStates.clear();
-    }
-
     @Override
     protected void initialiseDownstreamResolvers() {
```
flyingsilverfin, line 152: note: if we don't need `resolverRules` in `BoundConcludable` make sure you get rid of it here too, only used for that purpose

flyingsilverfin (changes requested):
Ok I think i'm happy enough with checking out the code and reviewing it for now. It's definitely a transition state right now so hopefully a lot of things shake out to be cleaner soon!

reasoner/ReasonerProducer.java
```diff
@@ -139,25 +138,35 @@ private void requestAnswered(Finished answer) {
     private void requestFailed(int iteration) {
         LOG.trace("Failed to find answer to request in iteration: " + iteration);
         if (options.traceInference()) ResolutionTracer.get().finish();
-        if (resolverRegistry.concludableResolvers().size() == 0) {
+        if (resolverRegistry.reiterationQueryRespondents(rootResolver, iteration).size() == 0) {
```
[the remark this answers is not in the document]
jmsfltchr: Yeah fair

reasoner/resolution/resolver/BoundConcludableResolver.java
jmsfltchr: This is a matter of whether we want to do `.initialiseDownstreamResolvers()` in the `ConcludableResolver` or `BoundConcludableResolver`. We can do it in the `BoundConcludableResolver` and rely on the fact that the `registry` does a put operation to create so that we don't have duplicate actors. For this decision I'm more inclined to think that it seems like the job of the coordinators to do this - if we move it to the bound resolvers then the coordinators become very thin classes.

Regarding  the `resolverRules` we can ask the registry for this if we store it there, or we can ask the resolver itself (not ideal) but if we're moving to have the coordinators spawn their workers then I'm not convinced. We don't want to be reading from the registry in that case right?

reasoner/resolution/resolver/BoundConcludableResolver.java
jmsfltchr: We can, I thought this would be clearer to show where the response is. If you think not then I'll just remove it.

reasoner/resolution/resolver/BoundConcludableResolver.java
jmsfltchr: You're right the comment is outdated

reasoner/resolution/resolver/BoundConcludableResolver.java
jmsfltchr: If we did that then surely the casting won't work?

reasoner/resolution/resolver/ConcludableResolver.java
jmsfltchr: You're right it's a put

reasoner/resolution/resolver/BoundConcludableResolver.java
flyingsilverfin: i think what we'll end up with is an `initialiseDownstreamCoordinators` in each coordinator, which can be done with the registry. The registry will become a `coordinator` registry. 

This say we can use the registry and concludables to access the coordinators from the `Bound` resolvers. And we don't need all this state anymore :) 

Seem ok?

reasoner/resolution/resolver/BoundConcludableResolver.java
jmsfltchr: Yes that's definitely my intention!

reasoner/resolution/resolver/BoundConcludableResolver.java
jmsfltchr: Yes that sounds great!

PR #6410, opened 2021-07-23
Create clear structure of resolver types

jmsfltchr:

We see that the resolvers share common patterns which can be abstracted. We also see that within the Concludable resolver there are distinct code paths depending upon whether the user's query is a match or explain query. It's now possible, based on prior refactoring, to split match and explain code paths into two separate classes.

- Create a `Coordinator` class which encapsulates the main operations of `ConcludableResolver` and `RetrievableResolver`.
- Create `MatchBoundConcludableResolver` and `ExplainBoundConcludableResolver` since they have separate behaviours. This means explaning and matching now no longer share a cache.

Part of #6356


reasoner/resolution/ResolverRegistry.java
```diff
@@ -208,18 +210,25 @@ public ResolverView registerResolvable(Resolvable<?> resolvable) {
...
             Concludable concludable, ConceptMap bounds, Map<Actor.Driver<ConclusionResolver>, Rule> resolverRules,
             LinkedHashMap<Actor.Driver<ConclusionResolver>, Set<Unifier>> applicableRules,
-            Actor.Driver<? extends Resolver<?>> root, int iteration) {
+            Actor.Driver<? extends Resolver<?>> root, int iteration, boolean explain) {
         LOG.debug("Register BoundConcludableResolver, pattern: {} bounds: {}", concludable.pattern(), bounds);
-        Actor.Driver<BoundConcludableResolver> resolver = Actor.driver(driver -> new BoundConcludableResolver(
-                driver, concludable, bounds, resolverRules, applicableRules, this, traversalEngine, conceptMgr, resolutionTracing
-        ), executorService);
+        Actor.Driver<BoundConcludableResolver> resolver;
+        if (explain) {
```
jmsfltchr, line 216: I think the end state we want is a single `ConcludableResolver` class, with all its sub-actors as nested classes. At that point I think we should question constructing all actors here in the registry and instead have the Coordinator (which is probably the `ConcludableResolver` class itself) do the construction. This `explain` flag way looks weird to me, but I didn't want yet another method in the registry for the two types.

flyingsilverfin: I'm inclined to agree. The reason for the registry remains to create the shared coordinators (I think) for rule and concludable. The coordinator classes, when terminated from the registry, will then terminate their children. We'll have to flesh out the termination into a bit less of a hack into a proper set of Requests/Responses. Goes to show we'll need multiple non-answer messages: materialise and terminate (that i can see now)

flyingsilverfin (approved):
Still not 100% on the names `Coordinator` and `Worker`, although at least they sort of form a pair...  maybe we can think of some alternatives that work nicely.

reasoner/resolution/ResolverRegistry.java
jmsfltchr: Good point, that makes sense

Issue #6404, opened 2021-07-20
Remove all usages of `.iterator.next()` to get one element


flyingsilverfin:
## Problem to Solve

Throughout the core codebase we have usages where we open an iterator and only retrieve the first element. This can end up being inefficient because we do not release the underlying Rocks iterators (if it backed by a rocks iterator). 

## Proposed Solution

We should use `.first()` if we only want to retrieve the first or some element from an iterator. This method will automatically call `.recycle()` on the iterator

haikalpribadi:
Solved by https://github.com/vaticle/typedb/pull/6413

PR #6413, opened 2021-07-26
Replace next() with first() to ensure resources are closed

flyingsilverfin:

As mentioned in #6404 we occasionally use `iterator.next()` instead of `.first()`. This can cause iterators to remain open and un-recycled until much later when finalised. We should use terminating iterator operations such as `first()` when only retriving one element from an iterator.

* replace usages of a single `.next()` call with `.first().get()`


haikalpribadi:
Should this PR mark that it will `close #6404` @flyingsilverfin ?

haikalpribadi (approved):

PR #6416, opened 2021-07-30
Refactor Conclusion Resolver

jmsfltchr:

As per the refactors to the actors for Concludables and Retrievables, we break down rule conclusion actors into one actor per set of bounds being evaluated. This requires a single actor to perform materialisation such that the put operations of materialisations don't suffer race conditions. This approach should better distribute the work done and the cleaner architecture will accelerate development.

- `ConclusionResolver` spawns `BoundConclusionResolver`s
- `Materialiser` actor to synchronise put operations.

Part of #6356

### DIFF
--- a/concurrent/actor/Actor.java
+++ b/concurrent/actor/Actor.java
@@ -17,6 +17,9 @@
 
 package com.vaticle.typedb.core.concurrent.actor;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import javax.annotation.concurrent.ThreadSafe;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
@@ -25,9 +28,11 @@ import java.util.function.Function;
 @ThreadSafe
 public abstract class Actor<ACTOR extends Actor<ACTOR>> {
 
+    private static final Logger LOG = LoggerFactory.getLogger(Actor.class);
     private static final String ERROR_ACTOR_DRIVER_IS_NULL = "driver() must not be null.";
     private final Driver<ACTOR> driver;
     private final String name;
+    private boolean terminated;
 
     public static <A extends Actor<A>> Driver<A> driver(Function<Driver<A>, A> actorFn, ActorExecutorGroup service) {
         return new Driver<>(actorFn, service);
@@ -38,6 +43,7 @@ public abstract class Actor<ACTOR extends Actor<ACTOR>> {
     protected Actor(Driver<ACTOR> driver, String name) {
         this.driver = driver;
         this.name = name;
+        this.terminated = false;
     }
 
     protected Driver<ACTOR> driver() {
@@ -48,6 +54,13 @@ public abstract class Actor<ACTOR extends Actor<ACTOR>> {
     public String name() {
         return name;
     }
+
+    public void terminate(Throwable cause) {
+        LOG.debug("Actor terminated. ", cause);
+        this.terminated = true;
+    }
+
+    public boolean isTerminated() { return terminated; }
 
     public static class Driver<ACTOR extends Actor<ACTOR>> {
 
--- a/logic/Rule.java
+++ b/logic/Rule.java
@@ -20,7 +20,6 @@ package com.vaticle.typedb.core.logic;
 
 import com.vaticle.typedb.core.common.exception.TypeDBException;
 import com.vaticle.typedb.core.common.iterator.FunctionalIterator;
-import com.vaticle.typedb.core.common.iterator.Iterators;
 import com.vaticle.typedb.core.common.parameters.Label;
 import com.vaticle.typedb.core.concept.Concept;
 import com.vaticle.typedb.core.concept.ConceptManager;
@@ -331,8 +330,8 @@ public class Rule {
          * @param conceptMgr   - used to insert the conclusion if it doesn't already exist
          * @return - all possible conclusions: there may be multiple preexisting satisfactory conclusions, we return all
          */
-        public abstract FunctionalIterator<Map<Identifier.Variable, Concept>> materialise(ConceptMap whenConcepts, TraversalEngine traversalEng,
-                                                                                          ConceptManager conceptMgr);
+        public abstract Optional<Map<Identifier.Variable, Concept>> materialise(ConceptMap whenConcepts, TraversalEngine traversalEng,
+                                                                                ConceptManager conceptMgr);
 
         public Rule rule() {
             return rule;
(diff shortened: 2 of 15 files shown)


logic/Rule.java
```diff
@@ -472,30 +472,91 @@ public Relation(RelationConstraint relation, IsaConstraint isa, Rule rule) {
...
+                });
+                Map<String, Map<Concept, Integer>> insertedMap = new HashMap<>();
+                inserted.getPlayersByRoleType().forEach((role, players) -> {
+                    players.forEach(player -> {
+                        Map<Concept, Integer> playerEntry = insertedMap.computeIfAbsent(role.getLabel().name(), r -> new HashMap<>());
+                        playerEntry.computeIfPresent(player, (p, count) -> count + 1);
+                        playerEntry.putIfAbsent(player, 1);
                     });
+                });
+                return relationMap.equals(insertedMap);
```
jmsfltchr, line 520: To do the comparison in an easy-to-understand way, we just create a map for each and compare them. Seemed less error prone than trying to iterate over one and compare with the other while making sure both have the same counts.

concurrent/actor/Actor.java
```diff
@@ -25,9 +28,11 @@
 @ThreadSafe
 public abstract class Actor<ACTOR extends Actor<ACTOR>> {
 
+    private static final Logger LOG = LoggerFactory.getLogger(Actor.class);
     private static final String ERROR_ACTOR_DRIVER_IS_NULL = "driver() must not be null.";
     private final Driver<ACTOR> driver;
     private final String name;
+    private boolean terminated;
```
jmsfltchr, line 35: This is abstracted in the next PR and as you say should be a message type of it's own

logic/Rule.java
```diff
@@ -472,42 +472,118 @@ public Relation(RelationConstraint relation, IsaConstraint isa, Rule rule) {
...
+                if (rp.roleType().get().reference().isName()) {
+                    return whenConcepts.get(rp.roleType().get().reference().asName()).asRoleType();
+                } else {
+                    assert rp.roleType().get().reference().isLabel();
+                    return scope.getRelates(rp.roleType().get().label().get().properLabel().name());
                 }
             }
 
-            private FunctionalIterator<Map<Identifier.Variable, Concept>> insert(RelationType relationType, ConceptMap whenConcepts, ConceptManager conceptMgr) {
+            private Map<Identifier.Variable, Concept> insert(RelationType relationType, ConceptMap whenConcepts) {
```
jmsfltchr, line 559: `insert` and `thenConcept` are very similar, but calling `thenConcepts` after `insert` would incur real cost as the role types would be retrieved from the graph twice.

reasoner/resolution/ResolverRegistry.java
```diff
@@ -208,16 +216,44 @@ public ResolverView registerResolvable(Resolvable<?> resolvable) {
...
+        if (explain) {
+            // TODO: Use explain resolver
+            resolver = Actor.driver(driver -> new BoundConclusionResolver(
+                    driver, conclusion, bounds, conditionResolver, materialiser, this, traversalEngine, conceptMgr,
+                    resolutionTracing), executorService);
+        } else {
+            // TODO: Use match resolver
+            resolver = Actor.driver(driver -> new BoundConclusionResolver(
+                    driver, conclusion, bounds, conditionResolver, materialiser, this, traversalEngine, conceptMgr,
+                    resolutionTracing), executorService);
```
jmsfltchr, line 251: We could discuss this - it's ready to split the Match and Explain, but actually for conclusion rather than concludable the code paths don't justify a whole new class to manage each possibility. They're basically one method different.

reasoner/resolution/resolver/BoundConclusionResolver.java
```diff
@@ -0,0 +1,377 @@
...
+        requestState.downstreamManager().removeDownstream(fromDownstream.sourceRequest());
+        sendAnswerOrSearchDownstreamOrFail(fromUpstream, requestState, iteration);
+    }
+
+    @Override
+    protected void initialiseDownstreamResolvers() {
+        throw TypeDBException.of(ILLEGAL_STATE);
+    }
+
+    private ConclusionRequestState<?> createRequestState(Request fromUpstream, int iteration) {
```
jmsfltchr, line 212: for conclusion rather than concludable the code paths don't justify a whole new class to manage each possibility. They're basically one method different.

reasoner/resolution/resolver/BoundConclusionResolver.java
```diff
@@ -0,0 +1,377 @@
...
+        } else if (fromUpstream.partialAnswer().asConclusion().isMatch()) {
+            requestState = new ConclusionRequestState.Match(fromUpstream, iteration, conditionDownstreams(fromUpstream));
+        } else {
+            throw TypeDBException.of(ILLEGAL_STATE);
+        }
+
+        return requestState;
+    }
+
+    private List<Request> conditionDownstreams(Request fromUpstream) {
```
flyingsilverfin, line 227: it's one downstream but multiple possible "completions" given a concept that is meant to be generated! These completions are all for the same `Rule.Condition` (since this conclusion belongs to just 1 rule), but all may be different in terms of bounds

reasoner/resolution/resolver/BoundConclusionResolver.java
```diff
@@ -0,0 +1,377 @@
...
+        Set<Identifier.Variable.Retrievable> named = iterate(conclusion.retrievableIds()).filter(Identifier::isName).toSet();
+        return answers.map(ans -> partialAnswer.extend(ans).toDownstream(named));
+    }
+
+    private static abstract class ConclusionRequestState<CONCLUDABLE extends Concludable<?>> extends RequestState {
+
+        private final DownstreamManager downstreamManager;
+        private final WaitedMaterialisations waitedMaterialisations;
+        private boolean complete;
+        protected FunctionalIterator<CONCLUDABLE> materialisations;
```
flyingsilverfin, line 258: should this be a `Poller`? since you extend and query it multiple times...

reasoner/resolution/resolver/ConcludableResolver.java
```diff
@@ -66,88 +56,18 @@ public ConcludableResolver(Driver<ConcludableResolver> driver, Concludable concl
...
-        cacheRegistersByRoot.remove(root);
-        requestMapByRoot.remove(root);
-        subsumptionTrackers.remove(root);
-        boundConcludableResolversByRoot.remove(root);
-    }
-
-    private Driver<BoundConcludableResolver> getOrReplaceBoundConcludable(Driver<? extends Resolver<?>> root, ConceptMap bounds) {
-        return boundConcludableResolversByRoot.computeIfAbsent(root, r -> new HashMap<>()).computeIfAbsent(bounds, b -> {
-            LOG.debug("{}: Creating a new BoundConcludableResolver for bounds: {}", name(), bounds);
+    Driver<BoundConcludableResolver> getOrReplaceWorker(Driver<? extends Resolver<?>> root, AnswerState.Partial<?> partial) {
```
flyingsilverfin, line 62: as per previous comments, don't think we can `replace`

test/behaviour/reasoner/verification/Materialiser.java
```diff
@@ -109,7 +109,6 @@ void close() {
 
     private class Rule {
         private final com.vaticle.typedb.core.logic.Rule rule;
-        private final Concludable thenConcludable;
         private final Map<ConceptMap, Materialisation> conditionAnsMaterialisations;
         private boolean requiresReiteration;
```
flyingsilverfin, line 114: fyi our `typeDBRule` already has a name, since it lives in the `logic` package it should be `logicRule`

flyingsilverfin (changes requested):

concurrent/actor/Actor.java
jmsfltchr: I have not, we'll table that for discussion

logic/Rule.java
```diff
@@ -472,42 +472,118 @@ public Relation(RelationConstraint relation, IsaConstraint isa, Rule rule) {
...
+                    while (existingRelations.hasNext()) {
+                        com.vaticle.typedb.core.concept.thing.Relation preexisting = existingRelations.next();
+                        if (!preexisting.isInferred()) return Optional.empty();
+                        else {
+                            if (insertable(preexisting, whenConcepts)) {
+                                return Optional.of(thenConcepts(preexisting, whenConcepts));
+                            }
+                        }
+                    }
+                    return Optional.empty();
```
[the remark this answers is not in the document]
jmsfltchr: I thought the same at first but there's a subtlety in the logic in line 485: the moment we find a non-inferred concept we return to avoid doing any extra work. Otherwise, if we ever find an inferred fact that was insertable then we need to return that materialisation. The fact that we need to break from the iterator prematurely for efficiency stops us from using `.filter().map().first()`. My assertion there is that it's helping exactly the fast relation lookup issue you have been fixing and so we don't want to sacrifice performance for code elegance in this instance.

logic/Rule.java
jmsfltchr: I've been able to get to the 16 lines you describe, but only without incorrect equality. Equality for a map of lists isn't what we want here. Really, the smell is that `getPlayersByRoleType` gives a map of lists rather than a map of multisets imo, then working with it and doing the comparison would be easy.

logic/Rule.java
```diff
@@ -472,42 +472,118 @@ public Relation(RelationConstraint relation, IsaConstraint isa, Rule rule) {
...
+                    com.vaticle.typedb.core.concept.thing.Relation inserted, ConceptMap whenConcepts) {
+                if (isa.type().label().isPresent()) {
+                    if (!inserted.getType().getLabel().equals(isa.type().label().get().properLabel())) return false;
+                } else if (!inserted.getType().getLabel().equals(whenConcepts.get(isa.type().id().asRetrievable()).asType().getLabel())) return false;
+                Map<String, Map<Concept, Integer>> relationMap = new HashMap<>();
+                relation.players().forEach(player -> {
+                    assert player.roleType().isPresent(); // Must be present to be insertable
+                    TypeVariable roleType = player.roleType().get();
+                    String descopedRoleTypeLabel = roleType.label().map(LabelConstraint::properLabel)
+                            .orElseGet(() -> whenConcepts.get(roleType.id().asRetrievable()).asType().getLabel()).name();
```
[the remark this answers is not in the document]
jmsfltchr: Good idea

reasoner/resolution/resolver/BoundConclusionResolver.java
```diff
@@ -0,0 +1,377 @@
...
+        super(driver, initName(conclusion, bounds), registry, traversalEngine, conceptMgr, resolutionTracing);
+        this.bounds = bounds;
+        this.conditionResolver = conditionResolver;
+        this.conclusion = conclusion;
+        this.materialiser = materialiser;
+        this.requestStates = new HashMap<>();
+        this.materialiserRequestRouter = new HashMap<>();
+    }
+
+    private static String initName(Rule.Conclusion conclusion, ConceptMap bounds) {
```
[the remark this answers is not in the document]
jmsfltchr: Yeah I've got rid of this everywhere

reasoner/resolution/resolver/BoundConclusionResolver.java
```diff
@@ -0,0 +1,377 @@
...
+    private final Driver<ConditionResolver> conditionResolver;
+    private final Rule.Conclusion conclusion;
+    private final Driver<Materialiser> materialiser;
+    private final Map<Request, ConclusionRequestState<? extends Concludable<?>>> requestStates;
+    private final Map<Materialiser.Request, Pair<Request, Integer>> materialiserRequestRouter;
+
+    public BoundConclusionResolver(Driver<BoundConclusionResolver> driver, Rule.Conclusion conclusion, ConceptMap bounds,
+                                   Driver<ConditionResolver> conditionResolver,
+                                   Driver<Materialiser> materialiser, ResolverRegistry registry, TraversalEngine traversalEngine,
+                                   ConceptManager conceptMgr, boolean resolutionTracing) {
```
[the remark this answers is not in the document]
jmsfltchr: Ok true that approach makes sense. I have been cautious because I don't want concurrent modification of the registry, but as long as that's avoided we're ok.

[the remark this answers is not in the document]
jmsfltchr: In fact I now wonder this: if the Bound resolvers are to be orchestrated by "coordinators", then effectively those coordinators are registries in the same way as the `ResolverRegistry`. In which case they should all become the same class of thing, and the `ResolverRegistry` should be an actor too.

reasoner/resolution/resolver/BoundConclusionResolver.java
jmsfltchr: Yes! Is this a suggestion for a change?

reasoner/resolution/resolver/BoundConclusionResolver.java
```diff
@@ -0,0 +1,377 @@
...
+    protected void receiveMaterialisation(Materialiser.Response response) {
+        if (isTerminated()) return;
+        Materialiser.Request toDownstream = response.sourceRequest();
+        Pair<Request, Integer> fromUpstream = fromUpstream(toDownstream);
+        ConclusionRequestState<? extends Concludable<?>> requestState = this.requestStates.get(fromUpstream.first());
+        LOG.trace("{}: received materialisation response: {}", name(), response);
+        Optional<Map<Identifier.Variable, Concept>> materialisation = response.materialisation();
+        materialisation.ifPresent(m -> requestState.newMaterialisation(response.partialAnswer(), m));
+        sendAnswerOrSearchDownstreamOrFail(fromUpstream.first(), requestState, fromUpstream.second());
+        ConclusionRequestState.WaitedMaterialisations state = requestState.waitedMaterialisations();
```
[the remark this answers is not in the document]
jmsfltchr: No, it's an encapsulation of the state of whether a `RequestState` needs to wait for a materialisation to be received before it can respond upstream.

reasoner/resolution/resolver/BoundConclusionResolver.java
jmsfltchr: The implementation has already been revamped using a `Deque`

concurrent/actor/Actor.java
jmsfltchr: Ah in fact I misread this - it's a great catch as I thought I was only changing `reasoner/resolution/framework`. The reason for this change is to abstract termination behaviour between the resolvers and other actors, which for now is only the `Materialiser`. I've now made a `ReasonerActor` class to encapsulate common reasoner actor behaviours, anticipating that there are more actor types to come in this refactor.

PR #6417, opened 2021-07-30
Refactor Review

jmsfltchr:

We address any concerns that we have raised during a fast-moving refactor.

TBD


flyingsilverfin:
Feel free to lump the new comments from the last PR and merge this after, but we'll do a comprehensive review because I feel like up to this point the model might've progressed but the implementation is still a long way off. If the implementation isn't able to be done neatly it's a smell the model isn't there yet either !

Issue #6336, opened 2021-05-24
identical query only returns some answers if they have already been returned by another query 


JonThom:
## Description

After upgrading to TypeDB 2.1.0, I ran a test suite which uses a toy [tenancy dataset](https://github.com/JonThom/tybedb-database-tenancy). 

One test failed because a query that should return two answers now only returned one. 

I ran the query again in console, and confirmed that the query only returned one answer.

In order to confirm that no data was missing from the database, I  ran a more inclusive query to return the expected concepts. They were duly returned.

I tried to run the initial query a second time. And now the unchanged query *did* return both of the expected answers!

## Environment

1. OS (where TypeDB server runs): Mac OS 10
2. TypeDB version (and platform):TypeDB 2.1.0
3. TypeDB client: client-python

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Set up the [tenancy dataset](https://github.com/JonThom/tybedb-database-tenancy). 
2. open the console with `typedb console`
3. run a query that should return two answers 
`match $relation ($node_role:$node, $neighbour_role:$neighbour) isa $reltype; {$reltype type subtenancy; } or {$reltype type birth; } or {$reltype type tenancy; } or {$reltype type tenancy-dispute; } or {$reltype type neighbourship; }; $neighbour has name contains "Mufti"; $reltype relates $node_role; $reltype relates $neighbour_role; $node isa! person; $node has name contains "Bentsen"; get $node, $node_role, $relation, $neighbour, $neighbour_role;`

but only returns one answer:

> { $node_role type tenancy-dispute:disputing-tenant sub dispute:party; $node iid 0x966e80048000000000000003 isa person; $neighbour_role type tenancy-dispute:disputing-landlord sub dispute:party; $neighbour iid 0x966e80048000000000000005 isa person; $relation iid 0xaa8280018000000000000000 (tenancy-dispute:disputing-tenant: iid 0x966e80048000000000000003, tenancy-dispute:disputing-landlord: iid 0x966e80048000000000000005, tenancy-dispute:disputed-tenancy: iid 0xaa8280068000000000000001) isa tenancy-dispute; }

4. Run the more inclusive query  which does return the concepts that should have been included in a second answer in 3.

```
[...]

PR #6418, opened 2021-08-02
Reproduce missing traversal answer

flyingsilverfin:

We create a failing traversal integration test for the issue outlined in #6336 . We decide not to fix it right now, since that would invalidate an optimisation in the traversal engine and leave it to be addressed and unignored in future work.

* Add small traversal test for failing test which shows how the seeking backtrack skips an answers


haikalpribadi (approved):

lolski:
I'm adding labels, milestone and also assigns the PR to the `TypeDB Launch` project @haikalpribadi @flyingsilverfin .

The labels were done according to [the issue](https://github.com/vaticle/typedb/issues/6336) this PR is for.

PR #6429, opened 2021-08-26
Cycle blocking and cycle re-exploration

jmsfltchr:

Cycles within reasoning need to be terminated to prevent infinite computation. These cycles need to be revisited when new answers are found at the start of the cycle. We introduce a new actor-centric mechanism for detecting these cycles and determining when they require revisiting. This replaces the preexisting method, which was to reiterate all reasoning actions to reveal new facts from the cycles.

- Exact cycle detection.
- A new `blocked` response type to indicate that a cycle was encountered.
- A new `revisit` request type to trigger downstream unblocking.
- Downstream paths can be set to blocked by a given set of cycles, unblocked when new answers are encountered or a relevant revisit request is received.
- A `Template` Request from which `Visit` or `Revisit` requests can be built with a supplied `Trace`.
- Remove much of the subsumption code since this wasn't in a working state.
- Add a replay buffer to `BoundConcludabeResolver` so that when waiting for materialisations is finished messages are replayed in the expected order exactly.
- Upgrade the tracer to handle parallel requests
- Request traces are disabled when `traceInference` is off, for better performance.


flyingsilverfin:
I've made done some general cleanup that could have been done when cleaning up your own PR as well (i just did cursory, there's probably more):
- using private instead of protected in one place
- use consistent naming between the type and the variable name
- renaming: `getOrCreateWorker` -> `getOrCreateBoundResolver`, "worker" is too generic
- remove state that isn't required

on the first three points, I think they could have been avoided by re-examining the area in which you make a change again to see what can be changed further based on those changes. I still need to work on this too, but often deleting one thing means three other things can also be simplified/changed etc.

on the last point, I my thought process was "this looks like a lot of state and I don't really want to undersatnd it all, what can we delete". The more general version is keep deleting stuff unless it is absolutely necessary.

flyingsilverfin:
Higher level thoughts for when you're back:

- I don't like the `Template` thing we ended up with in the end! I realise i suggested it, but when I read the code it doesn't seem to fit well!
  * the code smell for me is that we're in MANY places un-instantiating the template and also instantiating it (for example, `Request.template()` is called in many places)
  * another code smell: `Template` looks a hell of a lot like a Factory, but if we call it Factory it feels very strange!

If we follow why we need to instantiate and un-instantiate the template many places as we go up and down the request chains, sending downstream or upstream, it's because we lose the unique request with the original Trace after we use the Template to look up the request state that stores all the state (it's a many-to-one mapping from Request to RequestState if trace is enabled).

I think I now see another way we could try to refactor this, let me know if that is something you've tried already:
1.  Make `Trace` nullable in `Request`, and keep including it in the equals/hashcode
2. Use the requestRouter to keep track of Requests with Traces. I think the Child -> Parent relationship we currently re-create by instantiating a Template with the Trace we can also do with the requestRouter's state
3. The requestState object will actually contain a many-to-one mapping! i.e. don't use the Template as the key, always use the request with the Trace inside as the key. But they all share the same RequestState
  - consequence 1: we only have to find the correct RequestState when ignoring the Trace once - when we do getOrCreateRequestState ... this is much less frequently than creating new Templates all the time. 
  - consequence 2: we no longer need the Template as the key of any maps
  - consequence 3: the `downstreams` will still be a Request.Template to allow us to instantiate the requests with a particular trace. BUT now it actually makes sense we call it `Request.Factory`! I suspect then we can also reduce the API of `Request.Factory` to be two/four methods: `createVisit()/createVisit(trace)` and `createRevisit()/creatRevisit(trace)`.


Let's do this review the way Haikal and I do it - one major refactor and then take another look. I think it'll be easier for me to follow once this is changed too :)

jmsfltchr:
> * there's a TODO in `registerBoundConcludable` that looks like it could be addressed?

Not addressed in order to not expand the scope of this PR even further.

> * what's the line of reasoning behind making `identity()` static in `ResolverRegistry`?

This was a helpful suggestion from IntelliJ since this methods doesn't access any of the class's fields. I saw no reason not to implement it.

PR #6437, opened 2021-09-06
Implement naiive multi-variable sorting

flyingsilverfin:

We implement the feature outlined in https://github.com/vaticle/typedb/issues/6434, and refactor the comparators used to implementing the sorting.

* Implement multi-variable sorting using `Comparator`

### DIFF
--- a/concept/thing/impl/AttributeImpl.java
+++ b/concept/thing/impl/AttributeImpl.java
@@ -28,6 +28,8 @@ import com.vaticle.typedb.core.graph.iid.PrefixIID;
 import com.vaticle.typedb.core.graph.vertex.AttributeVertex;
 
 import java.time.LocalDateTime;
+import java.util.Comparator;
+import java.util.Optional;
 
 import static com.vaticle.typedb.common.util.Objects.className;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.ThingRead.ATTRIBUTES_NOT_COMPARABLE;
@@ -68,14 +70,6 @@ public abstract class AttributeImpl<VALUE> extends ThingImpl implements Attribut
     }
 
     public abstract VALUE getValue();
-
-    @Override
-    public int compareTo(Attribute o) {
-        AttributeImpl<?> other = (AttributeImpl<?>) o;
-        if (getType().getValueType().comparables().contains(other.getType().getValueType())) {
-            return getType().getValueType().comparator(other.getType().getValueType()).compare(getValue(), other.getValue());
-        } else throw TypeDBException.of(ATTRIBUTES_NOT_COMPARABLE, this, other);
-    }
 
     @Override
     protected AttributeVertex.Write<VALUE> writableVertex() {
--- a/concept/thing/impl/AttributeImpl.java
+++ b/concept/thing/impl/AttributeImpl.java
@@ -28,11 +28,8 @@ import com.vaticle.typedb.core.graph.iid.PrefixIID;
 import com.vaticle.typedb.core.graph.vertex.AttributeVertex;
 
 import java.time.LocalDateTime;
-import java.util.Comparator;
-import java.util.Optional;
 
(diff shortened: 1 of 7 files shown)


concept/thing/impl/AttributeImpl.java
```diff
@@ -41,6 +42,7 @@
 public abstract class AttributeImpl<VALUE> extends ThingImpl implements Attribute {
 
     AttributeVertex<VALUE> attributeVertex;
+    AttributeTypeImpl attributeType;
```
flyingsilverfin, line 45: we introduce a cache of attribute type to make calls to things like `getValueType()` fast, which would occur many times when sorting

concept/thing/impl/AttributeImpl.java
```diff
@@ -67,6 +69,14 @@ private AttributeImpl(AttributeVertex<VALUE> vertex) {
 
     public abstract VALUE getValue();
 
+    @Override
+    public int compareTo(Attribute o) {
+        AttributeImpl<?> other = (AttributeImpl<?>) o;
+        if (getType().getValueType().comparables().contains(other.getType().getValueType())) {
+            return getType().getValueType().comparator(other.getType().getValueType()).compare(getValue(), other.getValue());
+        } else throw TypeDBException.of(ATTRIBUTES_NOT_COMPARABLE, this, other);
+    }
```
flyingsilverfin, line 78: we move the comparison by attribute value to `Attribute` concept rather than implementing it in the `Matcher`. We also define the Comparators used for particular value types exactly once: in the `Encoding`

concept/type/AttributeType.java
```diff
@@ -80,6 +81,8 @@
         DATETIME(Encoding.ValueType.DATETIME);
 
         private final Encoding.ValueType encoding;
+        private Set<ValueType> comparables;
+        private Set<ValueType> assignables;
```
flyingsilverfin, line 85: cache the comparables and assignables as they would be computed over and over again - this is good because inside an enum, these sets will be computed once and used everywhere

graph/common/Encoding.java
```diff
@@ -400,13 +401,27 @@ public Infix infix() {
...
+                pair(DOUBLE, new Compare(map(
+                        pair(LONG, Comparator.comparing(d -> (double) d, Double::compareTo)),
+                        pair(DOUBLE, Comparator.comparing(d -> (double) d, Double::compareTo))
+                ))),
+                pair(DATETIME, new Compare(map(
+                        pair(DATETIME, Comparator.comparing(d -> (LocalDateTime) d, LocalDateTime::compareTo))
+                ))),
+                pair(STRING, new Compare(map(
+                        pair(STRING, Comparator.comparing(s -> (String) s, String::compareToIgnoreCase))
+                )))
```
flyingsilverfin, line 424: we now define the comparables AND the comparator to use when comparing to a particular type in one place, rather than scattering the implementation around

query/Matcher.java
```diff
@@ -132,43 +135,37 @@ public static Matcher create(Reasoner reasoner, TypeQLMatch query, Context.Query
...
-                return att1.asLong().getValue().compareTo(att2.asLong().getValue());
-            } else if (att1.isDouble() || att2.isDouble()) {
-                Double double1 = att1.isLong() ? att1.asLong().getValue() : att1.asDouble().getValue();
-                Double double2 = att2.isLong() ? att2.asLong().getValue() : att2.asDouble().getValue();
-                return double1.compareTo(double2);
-            } else if (att1.isDateTime()) {
-                return (att1.asDateTime().getValue()).compareTo(att2.asDateTime().getValue());
-            } else {
-                throw TypeDBException.of(ILLEGAL_STATE);
-            }
```
flyingsilverfin, line 166: this has been moved into the `Attribute` class, which in turn gets comparators to use from the `ValueType` and `Encoding` classes

haikalpribadi (approved):

PR #6438, opened 2021-09-09
Update Meetup Badge

daniel-crowe:
Update meetup badge [meetup|typedb].


haikalpribadi:
I would presume so, @flyingsilverfin, but this PR is from 9th of September so I don't think it's valid anymore.

PR #6456, opened 2021-10-18
Refactor alpha equivalence

jmsfltchr:

It is possible that there is more than one alpha equivalent mapping between two concludable elements. We need to know all mappings in order to improve the efficiency of the reasoner when this scenario is encountered.

- Alpha equivalence returns an iterator to accommodate scenarios such as symmetric roleplayers.
- The notion of valid/invalid alpha equivalence is dropped in favour of an empty iterator for invalid cases.
- Reduce the scope of alpha equivalence to only be useful for the comparison of concludables. As a consequence:
  - The tests require a running DB for the purpose of type inference.
  - The test is now an integration test.
  - There is no need in general to compare for alpha equivalence across sets, only one specific case for `Set<ValueConstraint>`, which is easy to solve given the exact nature of it's `alphaEquals`.
- Refactor tests and add coverage for `isa` and `attribute` concludables.

Required to solve https://github.com/vaticle/typedb/issues/6453

### DIFF
--- a/test/integration/logic/resolvable/AlphaEquivalenceTest.java
+++ b/test/integration/logic/resolvable/AlphaEquivalenceTest.java
@@ -277,6 +277,76 @@ public class AlphaEquivalenceTest {
     }
 
     @Test
+    public void test_symmetric_relation_has_multiple_reflexive_results() {
+        schema("define siblingship sub relation, relates sibling; person sub entity, plays siblingship:sibling;");
+        Concludable r = relation(concludable("$r(sibling: $p, sibling: $c) isa siblingship", "$p isa person"));
+        Set<Map<String, String>> alphaMaps = r.alphaEquals(r).map(AlphaEquivalenceTest::alphaMapToStringMap).toSet();
+        Map<String, String> map1 = new HashMap<String, String>() {{
+            put("$r", "$r");
+            put("$p", "$p");
+            put("$c", "$c");
+            put("$_siblingship:sibling", "$_siblingship:sibling");
+            put("$_siblingship", "$_siblingship");
+        }};
+        Map<String, String> map2 = new HashMap<String, String>() {{
+            put("$r", "$r");
+            put("$p", "$c");
+            put("$c", "$p");
+            put("$_siblingship:sibling", "$_siblingship:sibling");
+            put("$_siblingship", "$_siblingship");
+        }};
+        assertEquals(set(map1, map2), alphaMaps);
+    }
+
+    @Test
+    public void test_symmetric_relation_by_roles_has_multiple_reflexive_results() {
+        schema("define employment sub relation, relates employee, relates employer; person sub entity, plays " +
+                       "employment:employee, plays employment:employer;");
+        Concludable r = relation(concludable("$r($role1: $p, $role2: $p) isa employment", "$p isa person"));
+        Set<Map<String, String>> alphaMaps = r.alphaEquals(r).map(AlphaEquivalenceTest::alphaMapToStringMap).toSet();
+        Map<String, String> map1 = new HashMap<String, String>() {{
+            put("$r", "$r");
+            put("$p", "$p");
+            put("$role1", "$role1");
+            put("$role2", "$role2");
+            put("$_employment", "$_employment");
+        }};
+        Map<String, String> map2 = new HashMap<String, String>() {{
+            put("$r", "$r");
+            put("$p", "$p");
+            put("$role1", "$role2");
+            put("$role2", "$role1");
+            put("$_employment", "$_employment");
+        }};
+        assertEquals(set(map1, map2), alphaMaps);
+    }
+
+    @Test
+    public void test_symmetric_relation_by_roles_and_roleplayers_has_multiple_reflexive_results() {
+        schema("define employment sub relation, relates employee, relates employer; person sub entity, plays " +
+                       "employment:employee, plays employment:employer;");
+        Concludable r = relation(concludable("$r($role1: $p1, $role2: $p2) isa employment", "$p isa person"));
+        Set<Map<String, String>> alphaMaps = r.alphaEquals(r).map(AlphaEquivalenceTest::alphaMapToStringMap).toSet();
+        Map<String, String> map1 = new HashMap<String, String>() {{
+            put("$r", "$r");
+            put("$p1", "$p1");
+            put("$p2", "$p2");
+            put("$role1", "$role1");
+            put("$role2", "$role2");
+            put("$_employment", "$_employment");
+        }};
+        Map<String, String> map2 = new HashMap<String, String>() {{
+            put("$r", "$r");
+            put("$p1", "$p2");
+            put("$p2", "$p1");
+            put("$role1", "$role2");
+            put("$role2", "$role1");
+            put("$_employment", "$_employment");
+        }};
+        assertEquals(set(map1, map2), alphaMaps);
+    }
+
+    @Test
     public void test_relation_with_duplicate_roleplayers_equivalent() {
         schema("define friendship sub relation, relates friend; " +
                        "person sub entity, plays friendship:friend;");


logic/resolvable/Concludable.java
```diff
@@ -315,6 +326,10 @@ public RelationConstraint relation() {
             return relation;
         }
 
+        public IsaConstraint isa() {
+            return isa;
+        }
```
flyingsilverfin, line 331: This should return an Optional

pattern/constraint/thing/HasConstraint.java
```diff
@@ -92,8 +93,8 @@ public String toString() {
     }
 
     @Override
-    public AlphaEquivalence alphaEquals(HasConstraint that) {
-        return AlphaEquivalence.valid().validIfAlphaEqual(attribute, that.attribute);
+    public FunctionalIterator<AlphaEquivalence> alphaEquals(HasConstraint that) {
+        return attribute.alphaEquals(that.attribute);
```
flyingsilverfin, line 97: Given that we want to move to a constraint-to-constraint comparison for alpha-equivalence, we should include all variables including `owner` in the alpha equals check in each constraint

So we should aim that:
Variable knows how to do alpha-equals to another Variable
Constraint<T> can do alpha-equals to only other Constraint<T>

the combination of alpha-equals in general we do at a higher level - for now in Concludable, in the future maybe in Conjunction :)

When reflecting this change we can simplify the concludable `alphaEquals` methods to not check `owner` explicitly

pattern/constraint/thing/RelationConstraint.java
```diff
@@ -136,19 +137,24 @@ public int hashCode() {
...
+    public FunctionalIterator<AlphaEquivalence> alphaEquals(RelationConstraint that) {
+        return AlphaEquivalence.empty()
+                .alphaEqualIf(players().size() == that.players().size())
+                .flatMap(a -> roleplayerEquivalences(that).flatMap(a::extendIfCompatible));
+    }
+
+    private FunctionalIterator<AlphaEquivalence> roleplayerEquivalences(RelationConstraint that) {
+        return Iterators.permutation(players()).flatMap(playersPermutation -> {
+            Iterator<RolePlayer> thisRolePlayersIt = playersPermutation.iterator();
+            Iterator<RolePlayer> thatRolePlayersIt = that.players().iterator();
```
flyingsilverfin, line 149: not our style to us `It` to denote the type, since we generally work in an IDE that can tell us the type if we have forgotten it :)

pattern/equivalence/AlphaEquivalence.java
```diff
@@ -17,262 +17,115 @@
...
-        public AlphaEquivalence addOrInvalidate(AlphaEquivalence alphaMap) {
-            if (!alphaMap.isValid()) return AlphaEquivalence.invalid();
-            Optional<Map<Variable, Variable>> m = mergeVariableMapping(variableMapping(), alphaMap.asValid().variableMapping());
-            Optional<Map<Variable, Variable>> r = mergeVariableMapping(reverseVariableMapping(), alphaMap.asValid().reverseVariableMapping());
-            if (m.isPresent() && r.isPresent()) {
-                map = m.get();
-                reverseMap = r.get();
-                if (map.size() != reverseMap.size()) return AlphaEquivalence.invalid();
-                return this;
+    private static Optional<Map<Variable, Variable>> mergeVariableMapping(Map<Variable, Variable> existing,
```
flyingsilverfin, line 67: modifying an input is sometimes the right style to use (generally we don't unless we have to). When we do, we should then not return that modified object back because the API looks like it has created a new object.

This function is particularly bad because it can modify the `existing` variable for a while, then fail and return an empty Optional. The API reads as if the operation failed and nothing happened, but we've implicitly updated the mapping as well.

flyingsilverfin: There's another smell to chase here which is that `variableMapping()` always returns a new copy of the `map` that this class holds, which is why this wasn't broken before.

It's probably cleaner to have separate APIs for `mapping()` and `mappingCopy()` and use them as appropriate - best to check the places `variableMapping()` is called in this class once over.

pattern/equivalence/AlphaEquivalence.java
```diff
@@ -17,262 +17,115 @@
...
-        @Override
-        public Valid asValid() {
-            throw new ClassCastException("Cannot cast " + this.getClass().getSimpleName() + " to " + Valid.class.getSimpleName());
-        }
-
-        @Override
-        public <T extends AlphaEquivalent<T>> AlphaEquivalence validIfAlphaEqual(Set<T> set1, Set<T> set2) {
-            return this;
-        }
+    public AlphaEquivalence extend(Variable from, Variable to) {
```
flyingsilverfin, line 80: i actually feel like in terms of importance this method should sit right next to `empty()` because it's the only way to actually add a pair of variables to the mapping, and the assertion is encoding the information that alpha equivalence only works on named variables!

pattern/equivalence/AlphaEquivalence.java
```diff
@@ -17,262 +17,115 @@
...
-                }
-            } else {
-                while (it.hasNext()) {
-                    AlphaEquivalence alphaMap = it.next().alphaEquals(o);
-                    if (alphaMap.isValid()) return Valid.create(alphaMap.asValid().variableMapping());
-                }
-            }
-            return AlphaEquivalence.invalid();
-        }
+    public Map<Retrievable, Retrievable> retrievableMapping() {
```
flyingsilverfin, line 104: If this class can only contain Variables with named identifiers, the mapping of identifiers returned here can be tighter: there's a subtype called `Identifier.Retrievable.Name`

pattern/equivalence/AlphaEquivalence.java
```diff
@@ -17,262 +17,115 @@
...
-        public Valid asValid() {
-            throw new ClassCastException("Cannot cast " + this.getClass().getSimpleName() + " to " + Valid.class.getSimpleName());
-        }
-
-        @Override
-        public <T extends AlphaEquivalent<T>> AlphaEquivalence validIfAlphaEqual(Set<T> set1, Set<T> set2) {
-            return this;
-        }
+    public AlphaEquivalence extend(Variable from, Variable to) {
+        assert from.reference().isName() == to.reference().isName();
```
flyingsilverfin, line 81: ah we've not been good about this in reasoner but i think we should be using `from.id()` instead of `from.reference()` -- actually I think we should remove `reference()` from the API at all

pattern/variable/ThingVariable.java
```diff
@@ -249,14 +250,10 @@ public ThingVariable asThing() {
...
-                .validIfAlphaEqual(this.isaConstraint, that.isaConstraint)
-                .validIfAlphaEqual(this.relationConstraint, that.relationConstraint)
-                .validIfAlphaEqual(this.hasConstraints, that.hasConstraints)
-                .validIfAlphaEqual(this.valueConstraints, that.valueConstraints)
-                .addMapping(this, that);
+    public FunctionalIterator<AlphaEquivalence> alphaEquals(ThingVariable that) {
+        return AlphaEquivalence.empty()
+                .alphaEqualIf(id().isName() == that.id().isName())
+                .flatMap(a -> a.alphaEqualIf(this.inferredTypes().equals(that.inferredTypes())))
+                .map(a -> a.extend(this, that));
```
flyingsilverfin, line 257: here we encode that assumption that alpha equivalence only works over named variables -- we can have that encoded here, but the `extend()` method maybe should actually be able to take any type of variable and handle it?

ie. maybe we should get rid of the assertion that variables can only have named IDs when extending the alpha equivalence

pattern/variable/TypeVariable.java
```diff
@@ -254,11 +255,11 @@ public void addTo(GraphTraversal.Thing traversal) {
...
-                .validIf(id().isName() == that.id().isName())
-                .validIfAlphaEqual(labelConstraint, that.labelConstraint)
-                .validIfAlphaEqual(valueTypeConstraint, that.valueTypeConstraint)
-                .addMapping(this, that);
+    public FunctionalIterator<AlphaEquivalence> alphaEquals(TypeVariable that) {
+        assert !inferredTypes().isEmpty();
+        return AlphaEquivalence.empty()
+                .alphaEqualIf(id().isName() == that.id().isName())
+                .flatMap(a -> a.alphaEqualIf(inferredTypes().equals(that.inferredTypes())))
+                .map(a -> a.extend(this, that));
```
flyingsilverfin, line 263: Given our new idea of what Alpha Equivalence does for Variable/Constraint/Concludable, at this level we can I think generalise to handle named and labeled variables - as long as they are the same type of identifier, it's ok.

test/integration/logic/resolvable/AlphaEquivalenceTest.java
```diff
@@ -0,0 +1,551 @@
...
+    private void schema(String schema) {
+        try (TypeDB.Session session = typeDB.session(AlphaEquivalenceTest.database, Arguments.Session.Type.SCHEMA)) {
+            try (TypeDB.Transaction tx = session.transaction(Arguments.Transaction.Type.WRITE)) {
+                tx.query().define(TypeQL.parseQuery(String.join("\n", schema)).asDefine());
+                tx.commit();
+            }
+        }
+    }
+
+    private void typeInference(Conjunction conjunction) {
```
flyingsilverfin, line 84: `inferTypes(conjunction)`

logic/resolvable/Concludable.java
```diff
@@ -265,6 +265,17 @@ else if (a.getType().isDouble())
         return predicateFn;
     }
 
+    private static FunctionalIterator<AlphaEquivalence> alphaEqualValueConstraints(Set<ValueConstraint<?>> set1,
+                                                                                   Set<ValueConstraint<?>> set2) {
+        if (set1.size() != set2.size()) return Iterators.empty();
+        else {
+            for (ValueConstraint<?> s1 : set1) {
+                if (!Iterators.iterate(set2).flatMap(s1::alphaEquals).first().isPresent()) return Iterators.empty();
```
flyingsilverfin, line 273: fyi `iterate()` and `single()` are already statically imported.

Is it possible for two different value constraints to be alpha-equivalent to the same target? Eg. can many map to one? This line seems like there could be a hole where we don't exclude the constraints from `set2` that have already been mapped to when we check the next element in `set1` - is that a concern?

flyingsilverfin (commented):
Last comment: I think the title can be more specific than "refactor" alpha equivalence - perhaps "Compute all alpha equivalences"? or "Alpha equivalence returns iterators"? or something like that

flyingsilverfin (commented):
Last comment: I think the title can be more specific than "refactor" alpha equivalence - perhaps "Compute all alpha equivalences"? or "Alpha equivalence returns iterators"? or something like that.

jmsfltchr:
> Last comment: I think the title can be more specific than "refactor" alpha equivalence - perhaps "Compute all alpha equivalences"? or "Alpha equivalence returns iterators"? or something like that

It's already merged 😢 but that would have been better

pattern/equivalence/AlphaEquivalence.java
jmsfltchr: > There's another smell to chase here which is that `variableMapping()` always returns a new copy of the `map` that this class holds, which is why this wasn't broken before.

Ah yes I do this where I deem appropriate because it's in the Essential Java book which advocates it to guard against exactly the kind of bug I wrote here. I realise we aren't doing this throughout the codebase though, so better to stick to the style we have. I've checked and we can have just one getter and make a copy where needed.

pattern/equivalence/AlphaEquivalence.java
jmsfltchr: This isn't the case! When we were reviewing I think you said you picked this up from `assert from.reference().isName() == to.reference().isName();` but this assertion says that both variables must either both be named or both anonymous. In general for alpha equivalence the mappings can contain a mixture of both named and anonymous variables.

pattern/variable/ThingVariable.java
jmsfltchr: Again, same as above, this check makes sure the variables are either both named or both anonymous, so there is no change to make.

logic/resolvable/Concludable.java
jmsfltchr: You're right, and I was aware of this but failed to actually account for it. The hole is fixed now.

pattern/variable/TypeVariable.java
jmsfltchr: I think this is the same problem as above, we're already doing it!

PR #6457, opened 2021-10-26
Optimise reasoner for symmetric patterns

jmsfltchr:

When the reasoner encounters a symmetric pattern, for example `($x, $y);` or `$a has $x; $a has $y;` it will do duplicate work if it evaluates `$x` with a particular concept and also `$y` with the same concept. We aim to optimise this, which will reduce work by twofold in these examples, and threefold or more if there are three symmetries in the same pattern (etc).

- Generate mappings that are used to detect when we can re-use an existing bound resolver
- This approach has been architected in the abstract Coordinator, but is currently only used in the `ConcludableResolver`. There we can perform alpha equivalence between a concludable and itself to determine symmetry. This is possible for `conclusion` and `retrievable` and could represent big optimisations, however these both require comparing a conjunction with itself to identify correct symmetric mappings. (For conclusions the `then` doesn't matter in isolation, the `when` must also be considered.)
- Remap the mapping in a `Concludable` partialAnswer
- One cache per explaining ResolutionState

### DIFF
--- a/reasoner/resolution/resolver/BoundConcludableResolver.java
+++ b/reasoner/resolution/resolver/BoundConcludableResolver.java
@@ -55,7 +55,6 @@ public abstract class BoundConcludableResolver<RESOLVER extends BoundConcludable
     protected final ConceptMap bounds;
     protected final BoundConcludableContext context;
     protected final AnswerCache<ConceptMap> matchCache;
-    private final AnswerCache<Partial.Concludable<?>> explainCache;
 
     protected BoundConcludableResolver(Driver<RESOLVER> driver, String name, BoundConcludableContext context,
                                        ConceptMap bounds, ResolverRegistry registry) {
@@ -63,7 +62,6 @@ public abstract class BoundConcludableResolver<RESOLVER extends BoundConcludable
         this.context = context;
         this.bounds = bounds;
         this.matchCache = new AnswerCache<>(() -> traversalIterator(context.concludable().pattern(), bounds));
-        this.explainCache = new AnswerCache<>(Iterators::empty);  // TODO: This does nothing useful, delete it when answers are used directly as soon as they are found
     }
 
     @Override
@@ -100,7 +98,7 @@ public abstract class BoundConcludableResolver<RESOLVER extends BoundConcludable
 
     protected BoundConcludableResolutionState<?> createResolutionState(Partial<?> fromUpstream) {
         if (fromUpstream.asConcludable().isExplain()) {
-            return new ExplainResolutionState(fromUpstream, explainCache);
+            return new ExplainResolutionState(fromUpstream, new AnswerCache<>(Iterators::empty));  // TODO: This cache is only useful when the same exact explain request is made more than once. Delete it when answers are used directly as soon as they are found.
         } else {
             boolean singleAnswerRequired = bounds.concepts().keySet().containsAll(unboundVars());
             return new MatchResolutionState(fromUpstream, matchCache, singleAnswerRequired);


reasoner/resolution/resolver/RetrievableResolver.java
```diff
@@ -35,19 +39,25 @@
...
         this.retrievable = retrievable;
         this.boundResolvers = new HashMap<>();
     }
 
     @Override
-    Driver<BoundRetrievableResolver> getOrCreateBoundResolver(AnswerState.Partial<?> partial) {
-        return boundResolvers.computeIfAbsent(partial.conceptMap(), p -> {
-            LOG.debug("{}: Creating a new BoundRetrievableResolver for bounds: {}", name(), partial);
-            return registry.registerBoundRetrievable(retrievable, partial.conceptMap());
+    protected Driver<BoundRetrievableResolver> getOrCreateBoundResolver(AnswerState.Partial<?> partial, ConceptMap mapped) {
```
flyingsilverfin, line 49: we should pass the mapping in, which couples the two arguments together. The instinct is that the `mapped` you pass in should be called `bounds`, but then it's unclear whether to use the bounds from the partial or the mapped. Also, `mapped` is domain knowledge from above ie. outside this function. We fix all those by passing in `mapping`

reasoner/resolution/resolver/SubsumptiveCoordinator.java
```diff
@@ -63,14 +73,28 @@ public void receiveRevisit(Request.Revisit fromUpstream) {
...
+        Driver<? extends Resolver<?>> worker = getOrCreateBoundResolver(fromUpstream.visit().partialAnswer(),
+                                                                        mapping.transform(conceptMap));
+        Request.Factory requestFactory = getOrCreateRequestFactory(fromUpstream.visit().partialAnswer(), worker, mapping);
         Request.Revisit revisit = requestFactory.createRevisit(fromUpstream.trace(), fromUpstream.cycles());
         revisitDownstream(revisit, fromUpstream);
     }
 
-    private Request.Factory getOrCreateRequestFactory(AnswerState.Partial<?> partial, Driver<? extends Resolver<?>> receiver) {
-        return requestFactories.computeIfAbsent(partial, p -> Request.Factory.create(driver(), receiver, p));
+    protected Mapping computeReflexiveMappings(ConceptMap conceptMap) {
```
flyingsilverfin, line 86: not plural

reasoner/resolution/resolver/SubsumptiveCoordinator.java
```diff
@@ -63,14 +73,28 @@ public void receiveRevisit(Request.Revisit fromUpstream) {
...
+        equivalentMappings.forEach(m -> reflexiveMappings.put(m.unTransform(conceptMap), m));
+        return reflexiveMappings.get(conceptMap);
+    }
+
+    private Request.Factory getOrCreateRequestFactory(AnswerState.Partial<?> partial, Driver<? extends Resolver<?>> receiver, Mapping mapping) {
+        return requestFactories.computeIfAbsent(partial, p -> Request.Factory.create(driver(), receiver, applyRemapping(p, mapping)));
+    }
```
flyingsilverfin, line 93: note to future iterations: this is using a lot of memory - trade off here is a model strangeness of creating a factory each time vs saving the factory. Hopefully this goes away in the new model since there is no clean/good solution here.

reasoner/resolution/resolver/SubsumptiveCoordinator.java
```diff
@@ -39,21 +40,30 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(SubsumptiveCoordinator.class);
     private final Map<AnswerState.Partial<?>, Request.Factory> requestFactories;
+    protected final Set<Mapping> equivalentMappings;
+    protected final Map<ConceptMap, Mapping> reflexiveMappings;
```
flyingsilverfin, line 44: `symmetricMappings`
`boundsRedirectionMappings`

reasoner/resolution/resolver/SubsumptiveCoordinator.java
```diff
@@ -39,21 +40,30 @@
...
     @Override
     public void receiveVisit(Request.Visit fromUpstream) {
         LOG.trace("{}: received Visit: {}", name(), fromUpstream);
         if (!isInitialised) initialiseDownstreamResolvers();
         if (isTerminated()) return;
-        Driver<? extends Resolver<?>> worker = getOrCreateBoundResolver(fromUpstream.partialAnswer());
-        Request.Factory requestFactory = getOrCreateRequestFactory(fromUpstream.partialAnswer(), worker);
+        ConceptMap conceptMap = fromUpstream.partialAnswer().conceptMap();
+        Mapping mapping = reflexiveMappings.get(conceptMap);
+        if (mapping == null) mapping = computeReflexiveMappings(conceptMap);
```
flyingsilverfin, line 63: the story abstraction level is:
`ConceptMap` -> `Mapping` -> Worker
the implementation detail of how to compute the mapping can be pushed one layer down - in the future we could do a more or less efficient implementation, but the story stays the same.

Siblings should stay at the same level of abstraction when reading the high level story.

reasoner/resolution/resolver/ConcludableResolver.java
```diff
@@ -77,8 +85,11 @@ public Concludable concludable() {
         });
     }
 
-    private boolean isCycle(AnswerState.Partial<?> partialAnswer) {
-        ConceptMap bounds = partialAnswer.conceptMap();
+    protected AnswerState.Partial<?> applyRemapping(AnswerState.Partial<?> partial, Mapping mapping) {
+        return partial.asConcludable().remap(mapping);
+    }
```
flyingsilverfin, line 90: missing override?

flyingsilverfin (commented):
Title looks good on this one!

reasoner/resolution/answer/Mapping.java
```diff
@@ -54,7 +62,7 @@ public ConceptMap transform(ConceptMap conceptMap) {
     }
 
     public ConceptMap unTransform(ConceptMap conceptMap) {
-        assert reverseMapping.size() == conceptMap.concepts().size();
+        assert reverseMapping.size() >= conceptMap.concepts().size();
```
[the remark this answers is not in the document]
jmsfltchr: Thus far we have only ever used `unTransform` once we have a fully-bound answer, which means there will be a concept for every element in the mapping. Now we're also using `Mapping` for ConceptMaps that aren't fully bound when we create bound resolvers. We could enforce this again, but allowing a mapping of `$x -> $y` to `untransform` an empty ConceptMap (to another empty conceptMap) seems natural.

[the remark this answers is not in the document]
flyingsilverfin: ok

PR #6462, opened 2021-10-29
Simplify cycle answer count tracking

jmsfltchr:

The tracking of the number of answers seen when a cycle was blocked was held inside a cycle. This caused issues in comparing cycles to determine which are outdated and can be discarded, thereby reducing the amount of work done. We move tracking the number of answer to be passed alongside cycles.

- Move `numAnswersSeen` outside of `Cycle` and pass them upstream as the keys and values of a map. When passing `Revisit` requests downstream only the cycles are needed and not the `numAnswersSeen`.


jmsfltchr:
Closing in favour of #6474

Issue #6372, opened 2021-06-10
Enable usage of external config file in default locations


haskie-lambda:
## Problem to Solve

Installing TypeDB in a readonly location produces a number of problems as explained in the linked issues.
Especially configuration is difficult to achieve and calls for wrapper scripts etc.

Akin to config files in other programs and services like vim,neovim,ssh,recoll,zsh & co 
parsing the config file from default locations like

- `$HOME/.typedb.properties`
- `$HOME/_typedb.properties` (windows)

would be a nice way to configure typedb when installed in a readonly container.

## Current Workaround

Using a wrapper script with commandline parameters; However there is issue #6370 which limits
this greatly

## Proposed Solution

parsing the config file from default locations like

- `$HOME/.typedb.properties`
- `$HOME/_typedb.properties` (windows)

if it is not found in the install location, with the following precedence order:
- commandline arguments 
- config file of the install location
- config file in `$HOME`

## Additional Information

#6370
#6371

flyingsilverfin:
The mentioned designs are finished and should take care of any related issues

Issue #6370, opened 2021-06-10
Customizing `server.logs` to a location outside the typedb folder has no effect


haskie-lambda:
## Description

Using 
```
typedb server --data=/path/to/some/directory/other/than/the/install/location \
                      --logs=/path/to/some/directory/other/than/the/install/location
```
results in the following log:
```
15:06:43,618 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Could NOT find resource [logback-test.xml]
15:06:43,619 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Could NOT find resource [logback.groovy]
15:06:43,620 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Found resource [logback.xml] at [file:/nix/store/0fp1967g51vp872034kz504r7bxgpmg6-typedb-2.1.1/server/conf/logback.xml]
15:06:43,698 |-INFO in ch.qos.logback.classic.joran.action.ConfigurationAction - debug attribute not set
15:06:43,703 |-INFO in ch.qos.logback.classic.joran.action.ConfigurationAction - Will scan for changes in [file:/nix/store/0fp1967g51vp872034kz504r7bxgpmg6-typedb-2.1.1/server/conf/logback.xml] 
15:06:43,703 |-INFO in ch.qos.logback.classic.joran.action.ConfigurationAction - Setting ReconfigureOnChangeTask scanning period to 30 seconds
15:06:43,709 |-INFO in ch.qos.logback.core.joran.action.AppenderAction - About to instantiate appender of type [ch.qos.logback.core.ConsoleAppender]
15:06:43,713 |-INFO in ch.qos.logback.core.joran.action.AppenderAction - Naming appender as [STDOUT]
15:06:43,721 |-INFO in ch.qos.logback.core.joran.action.NestedComplexPropertyIA - Assuming default type [ch.qos.logback.classic.encoder.PatternLayoutEncoder] for [encoder] property
15:06:43,788 |-INFO in ch.qos.logback.core.joran.action.AppenderAction - About to instantiate appender of type [ch.qos.logback.core.rolling.RollingFileAppender]
15:06:43,791 |-INFO in ch.qos.logback.core.joran.action.AppenderAction - Naming appender as [LOGFILE]
15:06:43,792 |-INFO in ch.qos.logback.core.joran.action.NestedComplexPropertyIA - Assuming default type [ch.qos.logback.classic.encoder.PatternLayoutEncoder] for [encoder] property
[...]

PR #6473, opened 2021-11-05
YAML Configuration

flyingsilverfin:

In order to facilitate our long term goals of having a single configuration system for the entire database, including logging, we convert the preexisting `typedb.properties` and `logback.xml` into a Yaml single configuration file. This will also allow us to replicate configurations across an entire cluster in the future. In the process we also fix the bug where we were unable to configure log directories to be outside of the server distribution (#6370, #6372).

The top-level API is almost the same as before, please use the help command to get an overview:
```
./typedb server --help
```

Note: all YAML configurations can be overriden or extended with command line args, with `.` for nesting:
```
./typedb server --log.output.file.directory=/var/log/typedb
```
will override the persisted log directory set in the configuration.

The medium term goal is to eliminate the user-configurable `typedb.yml` file that comes with the server and ask the user to update configurations directly via console (https://github.com/vaticle/typedb/issues/6461).

* Introduce the idea of YAML-shaped configurations
* to set a configuration, either override it from the command line with `--<option>=<value>`, or provide a custom, complete YAML configuration file from an external place and set it via `--configFile=<path-to-config-file>`, or update the configuration file that ships with the distribution
* Logging is now configured at bootup by the server, accoring to the configuration
* we replace PicoCLI and implement our own parsing to use a YAML configuration file and allow named/dynamic keys to be provided
* implement our log configuration specification, along with Help menu printing, etc.
* Update the entry point of the `TypeDBServer` to consume a `Configuration`, and remove the old class `RunOptions`


server/common/EntryParser.java
```diff
@@ -0,0 +1,279 @@
...
+
+import static com.vaticle.typedb.common.util.Objects.className;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_CAST;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Server.CONFIG_OPTION_ENUM_UNEXPECTED_VALUE;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Server.CONFIG_OPTION_UNEXPECTED_VALUE_TYPE;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Server.MISSING_CONFIGURATION_OPTION;
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+import static com.vaticle.typedb.core.server.common.Util.scopeOption;
+
+public abstract class EntryParser implements Option.CliHelp {
```
flyingsilverfin, line 39: Each entry in the configuration file is specified with an `EntryParser` - these can be `Named.Unrestricted`, `Named.Enum`, or `Unnamed`. Each entry is also paired with a `ValueParser`, which parses the value in a key-value pair. This can be a simple Leaf value, or a Nested value. All the `Configuration` Parsers after the root parser are `ValueParser`'s, which take the value of a key-value section of the YAML configuration, and convert it into an object.

server/common/EntryParser.java
```diff
@@ -0,0 +1,279 @@
...
+                return true;
+            }
+
+            @Override
+            Nested<T> asNested() {
+                return this;
+            }
+        }
+
+        static class Leaf<T> extends ValueParser<T> {
```
flyingsilverfin, line 86: we pre-define some types of leaves/primitive value parsers. We could also do things like define a `Path` type here!

server/common/EntryParser.java
```diff
@@ -0,0 +1,279 @@
...
+                return this;
+            }
+
+            public String help() {
+                return help;
+            }
+        }
+    }
+
+    static class Unnamed<TYPE> extends EntryParser {
```
flyingsilverfin, line 159: This is the thing PicoCLI didn't support: we want to be able to have the user use a dynamic name for a _key_ in our yaml config - that's what this type of entry does for us.

server/conf/defaults.yml
```diff
@@ -0,0 +1,48 @@
+#
```
flyingsilverfin, line 1: this is now the file we set all our defaults in! No more code-defined defaults.

server/conf/typedb.yml
```diff
@@ -0,0 +1,49 @@
...
+#
+
+
+data: server/data
+port: 1729
+log:
+  output:
+    stdout:
+      type: stdout
+    dir:
```
lolski, line 25: Based on our verbal discussion, we should go with

```
    file:
      type: file
      directory: server/logs
      maxFileMB: 50
      maxFilesCount: 60
      maxFilesGB: 1
```

server/conf/typedb.yml
```diff
@@ -0,0 +1,49 @@
...
+      type: directory
+      path: server/logs
+      maxFileMB: 50
```
lolski, line 28:
```
file-max-size: 50MB
file-max-count: 60
```

lolski:
We're adding a lot of code. Let's make sure to run 'optimise imports' for the whole project.

lolski (changes requested):

lolski (approved):

PR #6474, opened 2021-11-05
Refactor review

jmsfltchr:

Address the comments raised on prior pull requests on a fast-moving branch. Changes relate to giving a clearer storyline and consistency in the code for better maintainability.

- Refactor AlphaEquivalence to always check the owner variable of a resolvable
- Make a nullable field optional in Concludable
- Remove excessive @Nullable decorators
- Move static cycle and blocking-related methods to the base Resolver class
- Improve method naming throughout

### DIFF
--- a/pattern/constraint/type/LabelConstraint.java
+++ b/pattern/constraint/type/LabelConstraint.java
@@ -104,8 +104,8 @@ public class LabelConstraint extends TypeConstraint implements AlphaEquivalent<L
 
     @Override
     public FunctionalIterator<AlphaEquivalence> alphaEquals(LabelConstraint that) {
-        return AlphaEquivalence.empty().alphaEqualIf(label().equals(that.label()))
-                .flatMap(a -> owner.alphaEquals(that.owner).flatMap(a::extendIfCompatible));
+        return owner.alphaEquals(that.owner)
+                .flatMap(a -> a.alphaEqualIf(label().equals(that.label())).flatMap(a::extendIfCompatible));
     }
 
     @Override
--- a/pattern/constraint/type/LabelConstraint.java
+++ b/pattern/constraint/type/LabelConstraint.java
@@ -105,7 +105,7 @@ public class LabelConstraint extends TypeConstraint implements AlphaEquivalent<L
     @Override
     public FunctionalIterator<AlphaEquivalence> alphaEquals(LabelConstraint that) {
         return owner.alphaEquals(that.owner)
-                .flatMap(a -> a.alphaEqualIf(label().equals(that.label())).flatMap(a::extendIfCompatible));
+                .flatMap(a -> a.alphaEqualIf(label().equals(that.label())));
     }
 
     @Override
--- a/pattern/constraint/type/LabelConstraint.java
+++ b/pattern/constraint/type/LabelConstraint.java
@@ -104,7 +104,8 @@ public class LabelConstraint extends TypeConstraint implements AlphaEquivalent<L
 
     @Override
     public FunctionalIterator<AlphaEquivalence> alphaEquals(LabelConstraint that) {
-        return AlphaEquivalence.empty().alphaEqualIf(label().equals(that.label()));
+        return AlphaEquivalence.empty().alphaEqualIf(label().equals(that.label()))
+                .flatMap(a -> owner.alphaEquals(that.owner).flatMap(a::extendIfCompatible));
     }
 
     @Override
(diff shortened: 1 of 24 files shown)


reasoner/resolution/framework/Materialiser.java
```diff
@@ -149,7 +149,7 @@ private Response(Request request, @Nullable Map<Identifier.Variable, Concept> ma
             return sourceRequest().sender();
         }
 
-        public @Nullable Trace trace() {
+        public Trace trace() {
```
flyingsilverfin, line 152: if these can be null and it isn't too verbose to do this, we should use `Optional` around nullable fields we return

reasoner/resolution/framework/Resolver.java
```diff
@@ -313,32 +307,81 @@ public void clear() {
...
+                }
+            });
+        }
+
+        private Optional<Cycle> findEquivalent(Set<Cycle> cycles, Cycle cycle) {
+            return iterate(cycles).filter(c -> c.origin().equals(cycle.origin())).first();
         }
 
-        public void unblock(Set<Cycle> cycles) {
+        public Set<Cycle> cyclesToRevisit(Pair<Concludable, ConceptMap> origin, int numAnswers) {
```
flyingsilverfin, line 349: reading this function alone, without context `cyclesToRevisit` given `origin` doesn't somehow make sense. The fact we're using the output to `revisit` later is not part of the definition of this function, it just does some filtering operation. What do we think a better name is? I think it will make `origin` also become a better name/be renamed too.

reasoner/resolution/resolver/BoundConcludableResolver.java
```diff
@@ -155,7 +161,7 @@ public ConceptMap answerFromPartial(Partial<?> partial) {
 
         @Override
         public void resetCacheSource() {
-            answerCache.setSource(() -> traversalIterator(context.concludable().pattern(), bounds));
+            answerCache.setSource(BoundConcludableResolver.this::createTraversal);
```
flyingsilverfin, line 164: good catch

reasoner/resolution/framework/Materialiser.java
flyingsilverfin: applies to other types of messages too

pattern/constraint/type/ValueTypeConstraint.java
```diff
@@ -91,7 +91,7 @@ public String toString() {
 
     @Override
     public FunctionalIterator<AlphaEquivalence> alphaEquals(ValueTypeConstraint that) {
-        return AlphaEquivalence.empty().alphaEqualIf(valueType().equals(that.valueType()));
+        return owner.alphaEquals(that.owner).flatMap(a -> a.alphaEqualIf(valueType().equals(that.valueType())));
```
flyingsilverfin, line 94: this looks nice now!

pattern/constraint/type/LabelConstraint.java
```diff
@@ -104,7 +104,8 @@ public String toString() {
 
     @Override
     public FunctionalIterator<AlphaEquivalence> alphaEquals(LabelConstraint that) {
-        return AlphaEquivalence.empty().alphaEqualIf(label().equals(that.label()));
+        return AlphaEquivalence.empty().alphaEqualIf(label().equals(that.label()))
```
flyingsilverfin, line 107: tbh i think it would be nice and more consistent if you always check the owner first

reasoner/resolution/framework/Resolver.java
```diff
@@ -255,8 +248,8 @@ protected CachingResolutionState(Partial<?> fromUpstream, AnswerCache<ANSWER> an
 
     public static class ExplorationManager {
```
flyingsilverfin, line 249: I've not really reviewed this because we pair programmed much of it (correct me if I'm wrong)

flyingsilverfin (changes requested):

pattern/constraint/type/LabelConstraint.java
jmsfltchr: Nice, I've switched it.

reasoner/resolution/framework/Resolver.java
jmsfltchr: Yes that's true

reasoner/resolution/framework/Materialiser.java
jmsfltchr: Noted, given the big changes coming I'll omit making changes

reasoner/resolution/framework/Resolver.java
jmsfltchr: We're definitely going to replace this, so skipping it

flyingsilverfin (approved):

PR #6476, opened 2021-11-12
Configure RocksDB to pin L0 index to reduce memory

flyingsilverfin:

We can empirically reduce 10-20% of memory usage during large data loads by making RocksDB pin the L0 index and bloom structures to the block cache.

* set `setPinL0FilterAndIndexBlocksInCache` to `true`

Explanation:
As far as I can tell, the bloom filters and index structure that are their own RocksDB blocks (?) can either live inside or outside the BlockCache, which is capped. So if the index/blooms live inside the BlockCache, we restrict their ability to take up space that we haven't accounted for. Blooms/Indexes otherwise live outside the block cache, forever. This options forces only the L0 block index/blooms to live in the cache.

A second, second option is `setCacheIndexAndFilterBlocks` -- this will force all index/blooms to live in the block cache for all to abide a memory cap. However, this is tested and has a massive (10x) performance penalty during heavy writes. The only restriction on the amount of memory being used by the blocks for index/blooms is the `num_open_files`, which will eventually hit the OS limits (eg. max-open-files on Ubuntu is 1024).


common/collection/Bytes.java
```diff
@@ -45,4 +52,38 @@ public static byte unsignedByte(int value) {
         return (byte) (value & 0xff);
     }
 
+    /**
+     * Derived from logback FileSize implementation
+     */
+    public static class Parser {
```
flyingsilverfin, line 58: Parser for strings such as `"10 mb"`, borrowed and cut from Logback

common/exception/ErrorMessage.java
```diff
@@ -125,20 +125,22 @@ public static void loadConstants() {
                 new Internal(6, "Attempted to utilise a closed resource.");
         public static final Internal UNRECOGNISED_VALUE =
                 new Internal(7, "Unrecognised encoding value!");
+        public static final Internal STORAGE_PROPERTY_EXCEPTION =
+                new Internal(8, "Failed to get storage property.");
```
flyingsilverfin, line 129: It's possible RocksDB throws when we try to read a property about the SST files or the current cache status.

common/parameters/Options.java
```diff
@@ -167,6 +169,18 @@ public Path reasonerDebuggerDir() {
...
+        else if (parent != null) return parent.storageDataCacheSize();
+        else throw TypeDBException.of(ILLEGAL_STATE);
+    }
+
+    public long storageIndexCacheSize() {
+        if (storageIndexCacheSize != null) return storageIndexCacheSize;
+        else if (parent != null) return parent.storageIndexCacheSize();
+        else throw TypeDBException.of(ILLEGAL_STATE);
+    }
+
```
flyingsilverfin, line 183: new configurations the user can set: storage cache sizes for index and data blocks. These come from the Configuration

graph/iid/InfixIID.java
```diff
@@ -26,7 +26,7 @@
 
 public abstract class InfixIID<EDGE_ENCODING extends Encoding.Edge> extends IID {
 
-    static final int LENGTH = 1;
+    public static final int LENGTH = 1;
```
flyingsilverfin, line 29: We will use the infix length to compute a good prefix extractor when defining bloom filters, so we make it public

rocks/RocksDatabase.java
```diff
@@ -103,8 +128,18 @@ protected RocksDatabase(RocksTypeDB typedb, String name, Factory.Session session
...
+            rocksDataOptions = rocksDataOptions();
+            if (LOG.isDebugEnabled()) {
+                rocksDataOptions.setStatistics(new Statistics()).setStatsDumpPeriodSec(ROCKS_LOG_PERIOD);
+                rocksPropertiesLogger = Executors.newScheduledThreadPool(1);
+                rocksData = OptimisticTransactionDB.open(rocksDataOptions, dataDirPath);
+                rocksPropertiesLogger.scheduleAtFixedRate(new RocksProperties.Logger(rocksData, name), 0, ROCKS_LOG_PERIOD, SECONDS);
+            } else {
+                rocksData = OptimisticTransactionDB.open(rocksDataOptions, dataDirPath);
+                rocksPropertiesLogger = null;
+            }
```
flyingsilverfin, line 142: we construct the `Options` in each Database object separately now. This is because if we share one `Options` globally, they actually end up sharing the block cache, which can have detrimental performance effects (one cache per database is best)

rocks/RocksDatabase.java
```diff
@@ -302,10 +337,21 @@ void remove(RocksSession session) {
...
+    private void shutdownRocksPropertiesLogger() {
+        try {
+            rocksPropertiesLogger.shutdown();
+            boolean terminated = rocksPropertiesLogger.awaitTermination(5, SECONDS);
+            if (!terminated) throw TypeDBException.of(ROCKS_LOGGER_SHUTDOWN_FAILED);
+        } catch (InterruptedException e) {
+            throw TypeDBException.of(e);
+        }
+    }
+
```
flyingsilverfin, line 354: if the user tells the `rocks` package to `debug`, then we periodically log properties we read from RocksDB. Really useful for debugging open snapshots, and getting estimated data size/number of keys, etc.

rocks/RocksDatabase.java
```diff
@@ -329,6 +377,90 @@ public void delete() {
...
+    private BlockBasedTableConfig rocksSchemaTableOptions() {
+        BlockBasedTableConfig rocksDBTableOptions = new BlockBasedTableConfig();
+        LRUCache uncompressedCache = new LRUCache(64 * MB);
+        rocksDBTableOptions.setBlockSize(16 * KB);
+        rocksDBTableOptions.setFormatVersion(5);
+        rocksDBTableOptions.setIndexBlockRestartInterval(16);
+        rocksDBTableOptions.setEnableIndexCompression(false);
+        rocksDBTableOptions.setBlockCache(uncompressedCache);
+        return rocksDBTableOptions;
+    }
```
flyingsilverfin, line 400: we should configure the Schema database differently to the Data database - high level, we don't need any bloom filters, much smaller cache and no compression

rocks/RocksDatabase.java
```diff
@@ -329,6 +377,90 @@ public void delete() {
...
+        // L0 index/filter should always be in memory, and is small
+        rocksDBTableOptions.setPinL0FilterAndIndexBlocksInCache(true);
+        // to cap memory usage, we must pin filter blocks and indexes in memory
+        rocksDBTableOptions.setCacheIndexAndFilterBlocks(true);
+        // ensure that the bloom filter partitioning index, plus the data index live in the cache always
+        rocksDBTableOptions.setPinTopLevelIndexAndFilter(true);
+        // use reserved section of block cache for blooms/index - cause massive thrashing if not using the high priority cache section
+        rocksDBTableOptions.setCacheIndexAndFilterBlocksWithHighPriority(true);
+        return rocksDBTableOptions;
+    }
```
flyingsilverfin, line 462: hard won knowldge of RocksDB configuration!!

rocks/RocksProperties.java
```diff
@@ -0,0 +1,146 @@
...
+
+import static com.vaticle.typedb.core.common.collection.Bytes.MB;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.STORAGE_PROPERTY_EXCEPTION;
+
+/**
+ * A list of interesting properties to retrieve from RocksDB.
+ * All properties are visible here:
+ * https://github.com/facebook/rocksdb/blob/20357988345b02efcef303bc274089111507e160/include/rocksdb/db.h#L750
+ */
+public enum RocksProperties {
```
flyingsilverfin, line 37: this class defines interesting properties we may want to Log periodically from RocksDB -
there are two types:
- runtime properties (rocksDB terminology), which are runtime properties like cache usage and compaction bytes to process
- data properties, which I define as an aggregated summary of the data that is persisted in SST on disk, such as total size of bloom filters, indexes, data etc.

rocks/RocksProperties.java
```diff
@@ -0,0 +1,146 @@
...
+    RocksProperties(String property, Function<String, String> formatter) {
+        this.property = property;
+        this.formatter = formatter;
+    }
+
+    private String getFormatted(OptimisticTransactionDB rocksDB) throws RocksDBException {
+        return formatter.apply(rocksDB.getProperty(property));
+    }
+
+    static class Logger implements Runnable {
```
flyingsilverfin, line 79: we bundle a logger for Properties that does things like formatting and aggregations that can be run once or periodically

rocks/RocksTypeDB.java
```diff
@@ -50,44 +44,17 @@
...
     protected RocksTypeDB(Options.Database options, Factory.DatabaseManager databaseMgrFactory) {
         if (!Executors.isInitialised()) Executors.initialise(MAX_THREADS);
         this.typeDBOptions = options;
-        this.rocksDBOptions = initRocksDBOptions();
         this.databaseMgr = databaseMgrFactory.databaseManager(this);
         this.databaseMgr.loadAll();
         this.isOpen = new AtomicBoolean(true);
     }
 
-    private org.rocksdb.Options initRocksDBOptions() {
```
flyingsilverfin, line 66: we no longer share Options between databases, and move their instantiation down to the `RocksDatabase` level

server/TypeDBServer.java
```diff
@@ -89,7 +90,9 @@ protected TypeDBServer(Configuration configuration, boolean debug, Factory facto
 
         Options.Database options = new Options.Database()
                 .typeDBDir(getTypedbDir())
-                .dataDir(configuration.dataDir())
+                .dataDir(configuration.storage().dataDir())
+                .storageDataCacheSize(configuration.storage().dbCache().dataSize())
+                .storageIndexCacheSize(configuration.storage().dbCache().indexSize())
```
flyingsilverfin, line 95: using the updated YAML configuration that now has a `Storage` sub-section for configuring storage options

server/TypeDBServer.java
```diff
@@ -156,7 +159,7 @@ private void configureTracing() {
         typeDBService = new TypeDBService(typedb);
         MigratorService migratorService = new MigratorService(typedb, Version.VERSION);
 
-        return NettyServerBuilder.forPort(configuration.port())
+        return NettyServerBuilder.forAddress(configuration.server().address())
```
flyingsilverfin, line 162: we now newly support listening to all addresses (using `0.0.0.0`) as well as just local address (`localhost`)

server/common/CommandLine.java
```diff
@@ -74,9 +74,13 @@ public CommandLine command(Command.Parser<?> parser) {
             String arg = args[i];
             if (arg.startsWith(Option.PREFIX)) {
                 int index = arg.indexOf("=");
-                if (index != -1) options.add(new Option(arg.substring(2, index), arg.substring(index + 1)));
-                else if (args[i + 1].startsWith(Option.PREFIX)) options.add(new Option(arg.substring(2)));
-                else options.add(new Option(arg.substring(2), args[++i]));
+                if (index != -1) {
+                    options.add(new Option(arg.substring(2, index), arg.substring(index + 1)));
+                } else if (i + 1 == args.length || args[i + 1].startsWith(Option.PREFIX)) {
```
flyingsilverfin, line 79: fix a bug where we can get an array out of bounds exception if  `i + 1` does not exist in the array

server/common/ConfigKVParser.java
```diff
@@ -102,14 +105,14 @@ public String key() {
 
         abstract TYPE parse(Yaml.Map yaml, String scope);
 
-        static class AnyValue<TYPE> extends EntryParser<TYPE> {
+        static class TypedValue<TYPE> extends EntryParser<TYPE> {
```
flyingsilverfin, line 108: renaming

server/common/ConfigKVParser.java
```diff
@@ -226,7 +229,25 @@ boolean isNested() {
...
+                        if (!yaml.isString()) return false;
+                        // use URI to parse IPV4, IVP4 and host names - however, we must add a dummy scheme to use it
+                        URI uri = URI.create("scheme://" + yaml.asString().value());
+                        return uri.getHost() != null && uri.getPort() != -1;
+                    },
+                    (yaml) -> {
+                        URI uri = URI.create("scheme://" + yaml.asString().value());
+                        return new InetSocketAddress(uri.getHost(), uri.getPort());
+                    },
+                    "<address:port>"
```
flyingsilverfin, line 250: we define two new types of configuration entries:
1. a `Size`, which uses the Bytes parser to read a user size entry such as "5 mb" into bytes
2. a `Address`, which parses things like `123.4.5.1:1123` or `web.blabla.com:1823`

server/common/Configuration.java
```diff
@@ -61,25 +63,25 @@
 
 public class Configuration {
 
-    private final Path dataDir;
-    private final int port;
+    private final Server server;
+    private final Storage storage;
```
flyingsilverfin, line 67: we support new sub-section `Server` and `Storage` for network related configurations and storage-layer options

server/common/Configuration.java
```diff
@@ -140,6 +142,126 @@ public VaticleFactory vaticleFactory() {
...
+
+        private final InetSocketAddress address;
+
+        private Server(InetSocketAddress address) {
+            this.address = address;
+        }
+
+        static class Parser extends ValueParser.Nested<Server> {
+
+            private static final EntryParser<InetSocketAddress> addressParser = TypedValue.create("address", "Address to listen for GRPC clients on.", INET_SOCKET_ADDRESS);
```
flyingsilverfin, line 158: the `Server` in Core only requires `address` as an entry, which must be an `InetSocketAddress`

server/common/Configuration.java
```diff
@@ -140,6 +142,126 @@ public VaticleFactory vaticleFactory() {
...
+
+        private Storage(Path dataDir, DbCache dbCache) {
+            this.dataDir = dataDir;
+            this.dbCache = dbCache;
+        }
+
+        static class Parser extends ValueParser.Nested<Storage> {
+
+            private static final EntryParser<Path> dataParser = TypedValue.create("data", "Directory in which user databases will be stored.", PATH);
+            private static final EntryParser<DbCache> dbCacheParser = TypedValue.create(DbCache.name, DbCache.description, new DbCache.Parser());
```
flyingsilverfin, line 196: `storage` has `data` and `db-cache` configurations, reflecting the YAML

server/common/Configuration.java
```diff
@@ -140,6 +142,126 @@ public VaticleFactory vaticleFactory() {
...
+
+            private DbCache(long dataSize, long indexSize) {
+                this.dataSize = dataSize;
+                this.indexSize = indexSize;
+            }
+
+            static class Parser extends ValueParser.Nested<DbCache> {
+
+                private static final EntryParser<Long> dataParser = TypedValue.create("data", "Size of storage-layer cache for data.", BYTES_SIZE);
+                private static final EntryParser<Long> indexParser = TypedValue.create("index", "Size of storage-layer cache for indexes.", BYTES_SIZE);
```
flyingsilverfin, line 238: we allow the data and index cache sizes to be configured by the user

server/common/Configuration.java
```diff
@@ -331,9 +453,9 @@ private File(String type, Path path, String fileSizeCap, String archivesSizeCap)
...
-                        private static final EntryParser<String> fileSizeCapParser = AnyValue.create("file-size-cap", "Log file size cap before creating new file (eg. 50mb).", STRING);
-                        private static final EntryParser<String> archivesSizeCapParser = AnyValue.create("archives-size-cap", "Total size cap of all archived log files in directory (eg. 1gb).", STRING); // TODO reasoner needs to respect this
+                        private static final EntryParser<Path> pathParser = TypedValue.create("directory", "Directory to write to. Relative paths are relative to distribution path.", PATH);
+                        private static final EntryParser<Long> fileSizeCapParser = TypedValue.create("file-size-cap", "Log file size cap before creating new file (eg. 50mb).", BYTES_SIZE);
+                        private static final EntryParser<Long> archivesSizeCapParser = TypedValue.create("archives-size-cap", "Total size cap of all archived log files in directory (eg. 1gb).", BYTES_SIZE); // TODO reasoner needs to respect this
```
flyingsilverfin, line 458: refactor old code to use new BYTES_SIZE parser rather than a generic String

server/common/Configuration.java
```diff
@@ -608,35 +730,38 @@ Reasoner parse(Yaml yaml, String scope) {
...
         }
 
         static class Parser extends ValueParser.Nested<VaticleFactory> {
 
-            private static final EntryParser<Boolean> traceParser = AnyValue.create("trace", "Enable Vaticle Factory tracing.", BOOLEAN);
-            private static final EntryParser<String> uriParser = AnyValue.create("uri", "URI of Vaticle Factory server.", STRING);
-            private static final EntryParser<String> usernameParser = AnyValue.create("username", "Username for Vaticle Factory server.", STRING);
-            private static final EntryParser<String> tokenParser = AnyValue.create("token", "Authentication token for Vaticle Factory server.", STRING);
-            private static final Set<EntryParser<?>> entryParsers = set(traceParser, uriParser, usernameParser, tokenParser);
+            private static final EntryParser<Boolean> enableParser = TypedValue.create("enable", "Enable Vaticle Factory tracing.", BOOLEAN);
```
flyingsilverfin, line 755: rename

server/common/Util.java
```diff
@@ -184,9 +182,8 @@ private static void configureRootLogger(Unfiltered defaultLogger, LoggerContext
         SizeAndTimeBasedRollingPolicy<?> policy = new SizeAndTimeBasedRollingPolicy<>();
         policy.setContext(context);
         policy.setFileNamePattern(logPath + TYPEDB_LOG_FILE_ARCHIVE_SUFFIX);
-        FileSize fileSize = FileSize.valueOf(outputType.fileSizeCap());
-        long directorySize = fileSize.getSize() + FileSize.valueOf(outputType.archivesSizeCap()).getSize();
-        policy.setMaxFileSize(fileSize);
+        long directorySize = outputType.fileSizeCap() + outputType.archivesSizeCap();
+        policy.setMaxFileSize(new FileSize(outputType.fileSizeCap()));
```
flyingsilverfin, line 186: refactor to use our own bytes size computation at parse time, simplifies this code and removes reliance on Logback!

server/conf/typedb.yml
```diff
@@ -16,8 +16,17 @@
...
+server:
+  address: 0.0.0.0:1729
+
+storage:
+  data: server/data
+  db-cache:
+    # configure data and index per database in the storage layer
+    # for large datasets, it is more important to have a large index cache than a data cache
+    data: 500mb
+    index: 500mb
```
flyingsilverfin, line 28: refactored configurations - `address` contains both a listen address and a port
`storage` contains the `data` directory and the per-database caches

server/test/ConfigurationTest.java
```diff
@@ -46,14 +48,16 @@
     @Test
     public void config_file_is_read() {
         Configuration configuration = (new Configuration.Parser()).getConfig();
```
flyingsilverfin, line 50: tests have been updated to reflect new Configuration structure and contents

lolski (approved):

server/test/ConfigurationTest.java
```diff
@@ -137,7 +143,7 @@ public void config_file_wrong_path_type_throws() {
         } catch (TypeDBException e) {
             assert e.code().isPresent();
             assertEquals(CONFIG_UNEXPECTED_VALUE_TYPE.code(), e.code().get());
-            assertEquals(CONFIG_UNEXPECTED_VALUE_TYPE.message("data", "1729[int]", ConfigKVParser.ValueParser.Leaf.PATH.help()), e.getMessage());
+            assertEquals(CONFIG_UNEXPECTED_VALUE_TYPE.message("storage.data", "123456[int]", ConfigKVParser.ValueParser.Leaf.PATH.help()), e.getMessage());
```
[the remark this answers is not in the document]
flyingsilverfin: the `<type>` is for help menus, and `[type]` is for printing an actual value annotated with a type. You think we should unify them? Maybe we should use `[type]` for help menu too

server/test/config-wrong-path-type.yml
```diff
@@ -15,8 +15,16 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
-data: 1729
```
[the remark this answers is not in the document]
flyingsilverfin: good idea, will be easier to read for sure

Issue #6471, opened 2021-11-01
Multiple IID negations return concepts with negated ids


flyingsilverfin:
## Description

Including two negations of IIDs will return answers that includes one of those IIDs, even though they should have been excluded by the negation.

## Environment

Any platform TypeDB 2.5.0

## Reproducible Steps

Schema:
```
define
  person sub entity, owns name;
  name sub attribute, value string;
```

insert data:
```
insert
  $x isa person, has name "bob";
  $y isa person, has name "alice";
  $z isa person, has name "catherine";
```

try the following queries:
```
> match $x isa person; not { $x iid <bob's iid> }; not { $x iid <alice's iid>; };
... returns all people excluding BOB
```

The same happens when going through reasoning.

## Expected Output

Alice and Bob should have been excluded in the answer, only returning the person with name Catherine.

## Actual Output

Returns all people only excluding Bob.
 
## Additional information

1. This occurs in both reasoning and non-reasoning
2. it may extend to non-IID negations
3. querying the following:
```
match $x isa person; not { {$x iid <bob's iid>; } or { $x iid <alice's iid>;};};
```
succeeds as expected!

FrankUrbach:
Another query which leads to unexpected  behavior is:
```
match 
$x sub thing;
not {$x type thing;};
not {$x type entity;};
not {$x type attribute;};
not {$x type relation;};
offset 0;
limit 1000;
```
Here thing, entity, attribute and relation shouldn't be visible.

flyingsilverfin:
Fixed with #6483

PR #6483, opened 2021-11-24
Correctly apply multiple negations with the same variables

flyingsilverfin:

The parsing of a TypeQL query into server-side Disjunctions, Conjunctions, and Negations, we relied on the equality function of negations, which is still ill-defined (consequence of #6115). As a result, negations were lost when negations contain the same variables and were put in a set. Identified in #6471.

* Don't rely on sets to contain negation patterns, and instead put them into a list, as we do with Variables already

### DIFF
--- a/pattern/Conjunction.java
+++ b/pattern/Conjunction.java
@@ -220,8 +220,10 @@ public class Conjunction implements Pattern, Cloneable {
         if (obj == null || obj.getClass() != getClass()) return false;
         Conjunction that = (Conjunction) obj;
         // TODO: This doesn't work! It doesn't compare constraints
+        // TODO: negations should be a set not list
+        // TODO: both are corrected with https://github.com/vaticle/typedb/issues/6115
         return (this.variableSet.equals(that.variables()) &&
-                this.negations.equals(that.negations())); // TODO negations should be a set not list
+                this.negations.equals(that.negations()));
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


pattern/Conjunction.java
```diff
@@ -61,26 +62,27 @@
...
-import static java.util.stream.Collectors.toSet;
+import static java.util.stream.Collectors.toList;
 
 public class Conjunction implements Pattern, Cloneable {
 
     private static final String TRACE_PREFIX = "conjunction.";
     private final Map<Identifier.Variable, Variable> variableMap;
     private final Set<Variable> variableSet;
-    private final Set<Negation> negations;
+    private final List<Negation> negations;
```
flyingsilverfin, line 75: key change: store negations in a `List` instead of a `Set`, which can use negations due to ill-defined equality functions.

pattern/Disjunction.java
```diff
@@ -87,6 +87,7 @@ public boolean equals(Object o) {
         if (o == null || getClass() != o.getClass()) return false;
 
         Disjunction that = (Disjunction) o;
+        // TODO should use a set-comparison
```
flyingsilverfin, line 90: how do we think we should link these TODOs to an open issue? This is possible with #6115

haikalpribadi (approved):

Issue #6376, opened 2021-06-15
Introduce ENCODING_VERSION in storage to catch breaking changes


flyingsilverfin:
## Problem to Solve

In the future, we will end up modifying the storage data format, which is a non-backward compatible change. To prevent users from migrating blindly to a version that is incompatible, we should store the data version number in storage and check it on bootup.

## Current Workaround

Don't migrate by referencing the data folder if there is a known incompatibility.

## Proposed Solution

Add to the SYSTEM encoding enum a ENCODING_VERSION. This is incremented when a breaking change occurs, and validated on startup to check that the persisted ENCODING_VERSION is the same as what is supported by the source code.

haikalpribadi:
"data" is to generic. We should call it `encoding_version`.

flyingsilverfin:
Implemented with #6484

PR #6484, opened 2021-11-25
Validate encoding version and compress encoding ranges

flyingsilverfin:

On the rare occasion we change the data format or break backward compatibility, we should prevent users from  copy-pasting their databases across incompatible server versions. We now write a data encoding version into the schema database on creation, and validate it against the server's data encoding version on open. Users will receive an exception and the server will refuse to start when given an incompatible database.

**non backward-compatible change**

* Introduce `Encoding.ENCODING_VERSION`, which is written on database creation, and validated on database open
* The key used to store the encoding version is `Encoding.System.ENCODING_VERSION_KEY`
* We also compress the `Encoding.Prefix` range to allow more room to grow in the future.
  * TypeDB Cluster now has prefix range 180+

Implements #6376

### DIFF
--- a/rocks/RocksDatabase.java
+++ b/rocks/RocksDatabase.java
@@ -55,6 +55,7 @@ import java.util.stream.Stream;
 
 import static com.vaticle.typedb.common.collection.Collections.hasIntersection;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Database.DATABASE_CLOSED;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Database.DATABASE_INCOMPATIBLE_ENCODING;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.DIRTY_INITIALISATION;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.UNEXPECTED_INTERRUPTION;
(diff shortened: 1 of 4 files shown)


rocks/RocksDatabase.java
```diff
@@ -144,12 +148,22 @@ protected void initialise() {
     protected void load() {
         try (RocksSession.Schema session = createAndOpenSession(SCHEMA, new Options.Session()).asSchema()) {
             try (RocksTransaction.Schema txn = session.initialisationTransaction()) {
+                validateEncodingVersion(txn);
```
flyingsilverfin, line 151: when we load a database, we validate the encoding version

rocks/RocksDatabase.java
```diff
@@ -144,12 +148,22 @@ protected void initialise() {
...
         }
     }
 
+    private void validateEncodingVersion(RocksTransaction.Schema txn) {
+        ByteArray encoding = txn.schemaStorage().get(ENCODING_VERSION_KEY.bytes());
+        if (encoding == null || encoding.isEmpty()) { // no or missing encoding version is version 1
+            throw TypeDBException.of(DATABASE_INCOMPATIBLE_ENCODING, name(), 1, ENCODING_VERSION);
+        } else if (encoding.decodeInt() < ENCODING_VERSION || encoding.decodeInt() > ENCODING_VERSION) {
+            throw TypeDBException.of(DATABASE_INCOMPATIBLE_ENCODING, name(), encoding.decodeInt(), ENCODING_VERSION);
+        }
```
flyingsilverfin, line 164: throw useful error messages when the encoding version mismatches

haikalpribadi (approved):

PR #6489, opened 2021-12-08
Partition storage using RocksDB column families

flyingsilverfin:

We can improve the efficiency of our storage layer by taking advantage of partitioned RocksDB column families. Using these, we can introduce different caching, bloom filter, and memtable configurations for each part of the data. This reduces overall memory requirements and helps TypeDB scale to huge amounts of data. At scale, this change improves random lookup performance, plus memory requirements. We also delete redundant data structures on disk, reducing disk storage requirements by at least 10%, up to much higher propertions in attribute-heavy, sparely connected datasets.

* introduce 5 column families: 
  - `DEFAULT`: used by all schema database keys, and data database vertices
  - `VARIABLE_START_EDGE`: for edges starting with variable size start vertices (eg. attributes)
  - `FIXED_START_EDGE`: for edges starting with a fixed with vertex, plus a simple infix and a vertex type prefix (17 bytes)
  - `OPTIMISATION_EDGE`: for edges with a prefix length of 21, eg. a fixed with vertex, plus an infix and role type, plus a vertex type prefix
  - `STATISTICS`: for all things statistics
* create custom RocksDB configurations for each, primarily around bloom filters, caches, and write buffer sizes
* introduce `RocksPartitionManager` to allow us to translate a PARTITION enum into a column family handle for schema and data databases
* introduce `Storage.Key`, which is used for almost all `Storage` APIs, which is a `ByteArray` + a Partition -- all accesses to the storage layer must now carry the partition they live in. Most IIDs  implement `Storage.Key` and define which partitions they are found in
* Introduce `Storage.Key.Prefix`, which carries a prefix, a partition, and a builder for the full key object
* Create a couple of new IID types that implement `Storage.Key`, which avoids using raw byte arrays to perform storage operations.
* Delete code for attribute indexing, since they were never being written or used!
* delete ISA edges, since we can now find all instances of a type just by iterating by vertex prefix in the default column family


graph/ThingGraph.java
```diff
@@ -214,15 +209,15 @@ public ThingVertex convertToReadable(VertexIID.Thing iid) {
...
             Map<IID, ? extends VERTEX> map, IID attIID, Function<IID, VERTEX> vertexConstructor) {
         VERTEX vertex = map.get(attIID);
-        if (vertex == null && storage.get(attIID.bytes()) != null) return vertexConstructor.apply(attIID);
+        if (vertex == null && storage.get(attIID) != null) return vertexConstructor.apply(attIID);
         else return vertex;
     }
 
     public FunctionalIterator.Sorted<ThingVertex> getReadable(TypeVertex typeVertex) {
-        ByteArray.Base prefix = join(typeVertex.iid().bytes(), Encoding.Edge.ISA.in().bytes());
+        Key.Prefix<EdgeIID.InwardsISA> prefix = EdgeIID.InwardsISA.prefix(typeVertex.iid());
```
flyingsilverfin, line 217: Each type of IID that we can look up in a partition can produce a `Key.Prefix` which can be iterated with. The prefix contains a builder for creating the full object

flyingsilverfin:
Review summary from in person review with Haikal:

There were three major changes requested during the review, each at the core of the major work done in this PR. Firstly, we refactored the `Key.Prefix` class, then we corrected the ownership model of vertex/edge, and lastly we changed the entry point for `RocksCFManager` (renamed to `RocksPartitionManager`).

1. Implicit and incorrect factory behaviour
We created a `Storage.Key` class to contain a byte array + a partition that the byte array should be looked up in/lives in.
A Prefix is also a Key, but I wanted them to be able to generate full Key instances, making them look a lot like factories at first glance.

I had implemented Prefix abstract type that had factory behaviour, eg. `Prefix<K>` had a method `K complete(ByteArray key)`, which returned a full `Key` instances. However, a Prefix is just a data structure, NOT a factory. It  should however carry a builder for a complete Key, plus the prefix to look up and the partition to look up in. Thus, refactored the Prefix to be exactly that: a container for a prefix array, a partition, and a `Builder<K>` which is just implemented by functional interfaces like a constructor. This cleaned up all the Prefixes in the classes we had, because instead of each Key `K` having their own concrete `Prefix` inner class, we just have a static method `prefix(...)` which return a reusuable `Key.Prefix<K>` class. This dovetails nicely with the other static constructors we already, have such as `VertexIID.of()` or `VertexIID.extract()` -- we add `VertexIID.Prefix`. The domain logic of what is required to build a prefix is still in the same code area as when it was an inner class, but in this case the inner class was not the right solution (quite a bit of redundant code, plus each prefix inner class was actually a factory when it should not have been).


2. Ownership between `VertexIID` and `EdgeIID`
During the first pass of this change, we had two types of VertexIID: VertexIID and PartitionedVertexIID. This came from a mental model where a Vertex is logically composed of the vertex identifier, plus edges that contain that vertex. Thus a VertexIID itself could be thought of as being partitioned by having edges in different partitions. However, this implicitly flips the ownership model we have everywhere, in which edges own vertices, not the other way around. Now, EdgeIID just owns VertexIID, and a VertexIID is always partitioned (there is no more "unpartitioned" vertexIID)

3. Constructing `RocksPartitionManager` by injecting components
The `RocksPartitionManager` holds on to two parallel lists that need to be passed to native components (Rocks) to initialise column families. 

In the first pass, these two lists were constructed in the `RocksPartitionManager`, which then meant much of the RocksDB initialisation had to happen inside the partition manager. This led to ALL the RocksDB initialisation, including database creation, to be done inside the `RocksPartitionManager`. This was clearly wrong - a partition manager should not construct a database! 

To correct this, we create the Rocks database, and the two parallel lists outside of the `RocksPartitionManager`, and then inject these two lists into the constructor of the `RocksPartitionManager`. This ended up making the partiton manager less stateful and simpler and correct the domain model.

flyingsilverfin:
Merged with approval from Haikal

PR #6490, opened 2021-12-10
Implement transaction timeouts

flyingsilverfin:

We implement transaction timeouts (outined in https://github.com/vaticle/typedb/issues/6487), to prevent users from accidentally leaving transactions open indefinitely and leaking memory as a result. The transaction timeout is configurable via the usual transaction/session options.

* Add transaction timeout that starts from the second the transaction is opened
* automatically close (with error) the transaction when the timeout occurs

### DIFF
--- a/common/exception/ErrorMessage.java
+++ b/common/exception/ErrorMessage.java
@@ -89,18 +89,20 @@ public abstract class ErrorMessage extends com.vaticle.typedb.common.exception.E
                 new Server(26, "Concept does not exist.");
         public static final Server BAD_VALUE_TYPE =
                 new Server(27, "The value type '%s' was not recognised.");
+        public static final Server TRANSACTION_EXCEEDED_MAX_SECONDS =
+                new Server(28, "Transaction exceeded maximum configured duration of '%s' seconds.");
         public static final Server EMPTY_TRANSACTION_REQUEST =
-                new Server(28, "Empty transaction request.");
+                new Server(29, "Empty transaction request.");
         public static final Server UNKNOWN_REQUEST_TYPE =
-                new Server(29, "The request message was not recognised.");
+                new Server(30, "The request message was not recognised.");
         public static final Server ITERATION_WITH_UNKNOWN_ID =
-                new Server(30, "Iteration was requested for ID '%s', but this ID does not correspond to an existing query iterator.");
+                new Server(31, "Iteration was requested for ID '%s', but this ID does not correspond to an existing query iterator.");
         public static final Server DUPLICATE_REQUEST =
-                new Server(31, "The request with ID '%s' is a duplicate.");
+                new Server(32, "The request with ID '%s' is a duplicate.");
         public static final Server ALREADY_RUNNING =
-                new Server(32, "Another instance of TypeDB server is already running at this port: '%s'.");
+                new Server(33, "Another instance of TypeDB server is already running at this port: '%s'.");
         public static final Server INCOMPATIBLE_JAVA_RUNTIME =
-                new Server(33, "Incompatible Java runtime version: '%s'. Please use Java 11 or above.");
+                new Server(34, "Incompatible Java runtime version: '%s'. Please use Java 11 or above.");
 
         private static final String codePrefix = "SRV";
         private static final String messagePrefix = "Invalid Server Operation";
--- a/common/exception/ErrorMessage.java
+++ b/common/exception/ErrorMessage.java
@@ -186,6 +186,8 @@ public abstract class ErrorMessage extends com.vaticle.typedb.common.exception.E
                 new Session(2, "Attempted to open a transaction from closed session.");
         public static final Session SCHEMA_ACQUIRE_LOCK_TIMEOUT =
                 new Session(3, "Could not acquire lock for schema session. Another schema session may have been left open.");
+        public static final Session SESSION_IDLE_TIMEOUT_NOT_CONFIGURABLE =
+                new Session(4, "The session idle timeout is not configurable at the '%s' level.");
 
         private static final String codePrefix = "SSN";
         private static final String messagePrefix = "Invalid Session Operation";
@@ -230,6 +232,8 @@ public abstract class ErrorMessage extends com.vaticle.typedb.common.exception.E
                 new Transaction(16, "Could not acquire lock for data transaction. A schema session may have been left open.");
         public static final Transaction RPC_PREFETCH_SIZE_TOO_SMALL =
                 new Transaction(17, "RPC answer streaming prefetch size must be at least 1, is set to: %d.");
+        public static final Transaction TRANSACTION_TIMEOUT_NOT_CONFIGURABLE =
+                new Transaction(18, "Transaction timeout cannot be configured at the '%s' level.");
 
         private static final String codePrefix = "TXN";
         private static final String messagePrefix = "Invalid Transaction Operation";
--- a/common/parameters/Options.java
+++ b/common/parameters/Options.java
@@ -29,16 +29,15 @@ import static com.vaticle.typedb.core.common.exception.ErrorMessage.Reasoner.REA
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Reasoner.REASONING_CANNOT_BE_TOGGLED_PER_QUERY;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Session.SESSION_IDLE_TIMEOUT_NOT_CONFIGURABLE;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Transaction.TRANSACTION_TIMEOUT_NOT_CONFIGURABLE;
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static java.util.concurrent.TimeUnit.SECONDS;
 
 public abstract class Options<PARENT extends Options<?, ?>, SELF extends Options<?, ?>> {
 
-    public static final int SECOND_MILLIS = 1000;
-    public static final int MINUTE_SECONDS = 60;
-
     public static final int DEFAULT_PREFETCH_SIZE = 50;
-    public static final int DEFAULT_SESSION_IDLE_TIMEOUT_MILLIS = 30 * SECOND_MILLIS;
-    public static final int DEFAULT_TRANSACTION_TIMEOUT_MILLIS = 5 * MINUTE_SECONDS * SECOND_MILLIS;
-    public static final int DEFAULT_SCHEMA_LOCK_ACQUIRE_TIMEOUT_MILLIS = 10_000;
+    public static final long DEFAULT_SESSION_IDLE_TIMEOUT_MILLIS = SECONDS.toMillis(30);
+    public static final long DEFAULT_TRANSACTION_TIMEOUT_MILLIS = MINUTES.toMillis(5);
+    public static final long DEFAULT_SCHEMA_LOCK_ACQUIRE_TIMEOUT_MILLIS = SECONDS.toMillis(10);
     public static final boolean DEFAULT_INFER = false;
     public static final boolean DEFAULT_TRACE_INFERENCE = false;
     public static final boolean DEFAULT_EXPLAIN = false;
@@ -53,9 +52,9 @@ public abstract class Options<PARENT extends Options<?, ?>, SELF extends Options
     private Boolean explain = null;
     private Boolean parallel = null;
     private Integer prefetchSize = null;
-    private Integer sessionIdlTimeoutMillis = null;
+    private Long sessionIdlTimeoutMillis = null;
     private Long transactionTimeoutMillis = null;
-    private Integer schemaLockAcquireTimeoutMillis = null;
+    private Long schemaLockAcquireTimeoutMillis = null;
     private Boolean readAnyReplica = null;
     protected Boolean prefetch = null;
     protected Path typeDBDir = null;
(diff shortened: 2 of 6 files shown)


common/exception/ErrorMessage.java
```diff
@@ -89,18 +89,20 @@ public static void loadConstants() {
                 new Server(26, "Concept does not exist.");
         public static final Server BAD_VALUE_TYPE =
                 new Server(27, "The value type '%s' was not recognised.");
+        public static final Server TRANSACTION_EXCEEDED_MAX_SECONDS =
+                new Server(28, "Transaction exceeded maximum configured duration of '%s' seconds.");
```
flyingsilverfin, line 93: we kill transactions after N seconds (default: 5 minutes = 300 seconds)

common/exception/ErrorMessage.java
```diff
@@ -184,6 +186,8 @@ public static void loadConstants() {
                 new Session(2, "Attempted to open a transaction from closed session.");
         public static final Session SCHEMA_ACQUIRE_LOCK_TIMEOUT =
                 new Session(3, "Could not acquire lock for schema session. Another schema session may have been left open.");
+        public static final Session SESSION_IDLE_TIMEOUT_NOT_CONFIGURABLE =
+                new Session(4, "The session idle timeout is not configurable at the '%s' level.");
```
flyingsilverfin, line 190: added a new error if someone tries to set a session idle timeout at the tx/query level, which is nonsense

common/exception/ErrorMessage.java
```diff
@@ -228,6 +232,8 @@ public static void loadConstants() {
                 new Transaction(16, "Could not acquire lock for data transaction. A schema session may have been left open.");
         public static final Transaction RPC_PREFETCH_SIZE_TOO_SMALL =
                 new Transaction(17, "RPC answer streaming prefetch size must be at least 1, is set to: %d.");
+        public static final Transaction TRANSACTION_TIMEOUT_NOT_CONFIGURABLE =
+                new Transaction(18, "Transaction timeout cannot be configured at the '%s' level.");
```
flyingsilverfin, line 236: added a new error if someone tries to set a transaction time at below-transaction level, which is nonsense

common/parameters/Options.java
```diff
@@ -22,15 +22,22 @@
...
 
 public abstract class Options<PARENT extends Options<?, ?>, SELF extends Options<?, ?>> {
 
+    public static final int SECOND_MILLIS = 1000;
+    public static final int MINUTE_SECONDS = 60;
+
     public static final int DEFAULT_PREFETCH_SIZE = 50;
-    public static final int DEFAULT_SESSION_IDLE_TIMEOUT_MILLIS = 30_000;
+    public static final int DEFAULT_SESSION_IDLE_TIMEOUT_MILLIS = 30 * SECOND_MILLIS;
+    public static final int DEFAULT_TRANSACTION_TIMEOUT_MILLIS = 5 * MINUTE_SECONDS * SECOND_MILLIS;
```
flyingsilverfin: default transaction timeout is 5 minutes (in milliseconds)

common/parameters/Options.java
```diff
@@ -129,6 +137,17 @@ public SELF sessionIdleTimeoutMillis(int idleTimeoutMillis) {
...
+    public SELF transactionTimeoutMillis(long timeoutMillis) {
+        this.transactionTimeoutMillis = timeoutMillis;
+        return getThis();
+    }
+
+    public long transactionTimeoutMillis() {
+        if (transactionTimeoutMillis != null) return transactionTimeoutMillis;
+        else if (parent != null) return parent.transactionTimeoutMillis();
+        else return DEFAULT_TRANSACTION_TIMEOUT_MILLIS;
+    }
```
flyingsilverfin, line 148: new option for transaction timeout

common/parameters/Options.java
```diff
@@ -268,10 +292,14 @@ public Query traceInference(boolean traceInference) {
             throw TypeDBException.of(REASONER_TRACING_CANNOT_BE_TOGGLED_PER_QUERY);
         }
 
+        @Override
+        public Query transactionTimeoutMillis(long timeoutMillis) {
+            throw TypeDBException.of(TRANSACTION_TIMEOUT_NOT_CONFIGURABLE, className(getClass()));
+        }
```
flyingsilverfin, line 297: not configurable at Query level

server/TransactionService.java
```diff
@@ -189,10 +196,13 @@ private void open(TransactionProto.Transaction.Req request) {
         networkLatencyMillis = Math.min(openReq.getNetworkLatencyMillis(), MAX_NETWORK_LATENCY_MILLIS);
         sessionSvc = sessionService(typeDBSvc, openReq);
         sessionSvc.register(this);
-        transaction = transaction(sessionSvc, openReq);
+        options = new Options.Transaction().parent(sessionSvc.options());
+        applyDefaultOptions(options, openReq.getOptions());
+        transaction = transaction(sessionSvc, openReq, options);
```
flyingsilverfin, line 200: refactored to pass options into the `transaction()` function, so we have access to them a to also set the transaction timeout with

server/common/RequestReader.java
```diff
@@ -66,6 +67,9 @@ public static UUID byteStringAsUUID(ByteString byteString) {
         if (request.getSessionIdleTimeoutOptCase().equals(SESSION_IDLE_TIMEOUT_MILLIS)) {
             options.sessionIdleTimeoutMillis(request.getSessionIdleTimeoutMillis());
         }
+        if (request.getTransactionTimeoutOptCase().equals(TRANSACTION_TIMEOUT_MILLIS)) {
+            options.transactionTimeoutMillis(request.getTransactionTimeoutMillis());
+        }
```
flyingsilverfin, line 72: read protocol option in from RPC

common/parameters/Options.java
```diff
@@ -22,16 +22,22 @@
...
+import static java.util.concurrent.TimeUnit.SECONDS;
 
 public abstract class Options<PARENT extends Options<?, ?>, SELF extends Options<?, ?>> {
 
     public static final int DEFAULT_PREFETCH_SIZE = 50;
-    public static final int DEFAULT_SESSION_IDLE_TIMEOUT_MILLIS = 30_000;
-    public static final int DEFAULT_SCHEMA_LOCK_ACQUIRE_TIMEOUT_MILLIS = 10_000;
+    public static final long DEFAULT_SESSION_IDLE_TIMEOUT_MILLIS = SECONDS.toMillis(30);
+    public static final long DEFAULT_TRANSACTION_TIMEOUT_MILLIS = MINUTES.toMillis(5);
+    public static final long DEFAULT_SCHEMA_LOCK_ACQUIRE_TIMEOUT_MILLIS = SECONDS.toMillis(10);
```
flyingsilverfin, line 40: we should start using java's `TimeUnit` builtins I think!

common/parameters/Options.java
```diff
@@ -118,24 +125,35 @@ public SELF parallel(boolean parallel) {
...
-    public SELF sessionIdleTimeoutMillis(int idleTimeoutMillis) {
+    public SELF sessionIdleTimeoutMillis(long idleTimeoutMillis) {
         this.sessionIdlTimeoutMillis = idleTimeoutMillis;
         return getThis();
     }
 
-    public int schemaLockTimeoutMillis() {
+    public SELF transactionTimeoutMillis(long timeoutMillis) {
+        this.transactionTimeoutMillis = timeoutMillis;
+        return getThis();
```
flyingsilverfin, line 141: Java TimeUnit uses longs, so I propagage that here

dependencies/vaticle/repositories.bzl
```diff
@@ -42,7 +42,7 @@ def vaticle_typedb_protocol():
     git_repository(
         name = "vaticle_typedb_protocol",
         remote = "https://github.com/vaticle/typedb-protocol",
-        tag = "2.5.0", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_protocol
+        commit = "4ce4e1339995c8982914e9fd3177273b693609b7", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_protocol
```
flyingsilverfin, line 45: new protocol with transaction timeout option

server/TransactionService.java
```diff
@@ -189,10 +195,13 @@ private void open(TransactionProto.Transaction.Req request) {
...
         services = new Services();
         respond(ResponseBuilder.Transaction.open(byteStringAsUUID(request.getReqId())));
         isTransactionOpen.set(true);
+        scheduledTimeout = scheduled().schedule(this::timeout, options.transactionTimeoutMillis(), MILLISECONDS);
```
flyingsilverfin, line 204: on timeout, call `timeout()`

server/TransactionService.java
```diff
@@ -271,13 +279,18 @@ private void mayCloseTrace(@Nullable FactoryTracingThreadStatic.ThreadTrace trac
         if (trace != null) trace.close();
     }
 
+    private void timeout() {
+        close(TypeDBException.of(TRANSACTION_EXCEEDED_MAX_SECONDS, MILLISECONDS.toSeconds(options.transactionTimeoutMillis())));
+    }
```
flyingsilverfin, line 284: close with an error that indicates a timeout happened

server/TransactionService.java
haikalpribadi: This is awkward. Feels weird that you have to pull out the parsing of the `Options` out to a higher level, right? And it feels weird that you have manage the `scheduledTimeout` of the transaction from outside of itself, right?

Given that the transaction is capable of managing its own lifetime, wouldn't it make more sense to get the transaction to manage its own lifetime? It already has access to the timeout limit it should adhere to, and it already has the ability to close itself upon exception. Which means we can encapsulate all the `scheduledTimout` logic inside `RocksTransaction`, right? You would be creating more _cohesion_ without increasing _coupling_, as everything you need to implement the logic is already available in the class. Always take the opportunity to _increase cohesion_ and _reduce coupling_.

server/TransactionService.java
```diff
@@ -271,13 +279,18 @@ private void mayCloseTrace(@Nullable FactoryTracingThreadStatic.ThreadTrace trac
...
+
     @Override
     public synchronized void close() {
         if (isRPCAlive.compareAndSet(true, false)) {
             if (isTransactionOpen.compareAndSet(true, false)) {
                 transaction.close();
                 sessionSvc.remove(this);
             }
+            if (scheduledTimeout != null) scheduledTimeout.cancel(false);
```
haikalpribadi, line 293: this is awkward, see explanation above

server/TransactionService.java
```diff
@@ -288,6 +301,7 @@ public synchronized void close(Throwable error) {
...
```
haikalpribadi, line 304: this is awkward, see explanation above

haikalpribadi (changes requested):

common/parameters/Options.java
```diff
@@ -232,6 +250,11 @@ Session getThis() {
         Transaction getThis() {
             return this;
         }
+
+        @Override
+        public Transaction sessionIdleTimeoutMillis(long idleTimeoutMillis) {
+            throw TypeDBException.of(SESSION_IDLE_TIMEOUT_NOT_CONFIGURABLE, className(getClass()));
+        }
```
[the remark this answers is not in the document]
flyingsilverfin: yep

server/TransactionService.java
```diff
@@ -88,6 +92,8 @@
     private volatile TypeDB.Transaction transaction;
     private volatile Services services;
     private volatile int networkLatencyMillis;
+    private volatile ScheduledFuture<?> scheduledTimeout;
+    private volatile Options.Transaction options;
```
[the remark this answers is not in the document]
flyingsilverfin: what's your process for deciding the ordering of fields? I always end up with two or three reasonable looking variations

server/TransactionService.java
flyingsilverfin: So I didn't see that before, but I don't think this is very awkward either.
The first part of creating the options first, then passing them into the transaction read fine to me:

1. get session service
2. set up transaction options
3. create the transaction using the session and options

I see your point with the second part, however the intent is actually more that the transaction lifetime is a feature aimed at the RPC, similar to the session idle timeout. This is actually made clearer in the extended version of this feature, outlined in https://github.com/vaticle/typedb/issues/6492. 

Secondly, we currently don't have the architecture for a Transaction (inner component) to pass a sensible error message to its parent (TransactionService) when it times out. We would end up in a (imo) less ideal state that simply logs the transaction timeout error on the server, and throws a "Transaction closed" exception to the client when it next tries to use the transaction.

server/TransactionService.java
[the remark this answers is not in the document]
haikalpribadi: I meant sort both of the lines above `private volatile int`. I generally try to group things that similar closer to each other first. In this case `int` is the only primitive data structure, and thus come last.

server/TransactionService.java
haikalpribadi: If you think about what RPC means -- Remote Procedure Call, the lifetime of the object really should have nothing to do with the fact that it can be called "remotely".  Which means the session lifetime is probably something that should have been managed by Session itself? I can't be too sure without looking the code directly. However, at this point, I think the ability to send a helpful error message is probably more important. And since we don't want to spend on refactoring our error message architecture, let's move forward with this implementation for now, and refactor it in the next version. But hopefully, the lesson of cohesiveness and coupling can still be taken away.

haikalpribadi (approved):

server/TransactionService.java
flyingsilverfin: Yep sounds good.
One note on viewing it as an "RPC"  feature, i agree with the definition of RPC you pointed out. Our implementation of "RPC"-based sessions/transactions are referred to as `SessionService` and `TransactionService`, so from those names it wouldn't be strange to have the services manage the lifetimes of the objects they serve either?

flyingsilverfin:
Review summary: main point is the comment from https://github.com/vaticle/typedb/pull/6490#discussion_r772667615 , where we discuss the cohesion of transaction behaviour and lifetime. The discussion is not concluded on whether the management of lifetime handled by services is better handled by the component itself (eg. the session idle timeout and the transaction timeout).

PR #6503, opened 2022-01-05
Allow setting a type as abstract given only indirect instances

flyingsilverfin:

We allow setting a type as abstract when a type only has indirect instances (a child types' instances, rather than its own). This was raised in https://github.com/vaticle/typedb/issues/6494.

* we introduce `getInstancesExplicit()` to get the exact instances of a type, ie. excluding the child types' instances
* setting abstractness is now only disallowed when there are explicit instances

### DIFF
--- a/concept/type/AttributeType.java
+++ b/concept/type/AttributeType.java
@@ -41,6 +41,9 @@ public interface AttributeType extends ThingType {
     @Override
     FunctionalIterator<? extends Attribute> getInstances();
 
+    @Override
+    FunctionalIterator<? extends Attribute> getInstancesExplicit();
+
     void setSupertype(AttributeType superType);
 
     boolean isKeyable();
@@ -149,6 +152,9 @@ public interface AttributeType extends ThingType {
         @Override
         FunctionalIterator<? extends Attribute.Boolean> getInstances();
 
+        @Override
+        FunctionalIterator<? extends Attribute.Boolean> getInstancesExplicit();
+
         Attribute.Boolean put(boolean value);
 
         Attribute.Boolean put(boolean value, boolean isInferred);
@@ -166,6 +172,9 @@ public interface AttributeType extends ThingType {
 
         @Override
         FunctionalIterator<? extends Attribute.Long> getInstances();
+
+        @Override
+        FunctionalIterator<? extends Attribute.Long> getInstancesExplicit();
 
         Attribute.Long put(long value);
 
@@ -185,6 +194,9 @@ public interface AttributeType extends ThingType {
         @Override
         FunctionalIterator<? extends Attribute.Double> getInstances();
 
+        @Override
+        FunctionalIterator<? extends Attribute.Double> getInstancesExplicit();
+
         Attribute.Double put(double value);
 
         Attribute.Double put(double value, boolean isInferred);
@@ -202,6 +214,9 @@ public interface AttributeType extends ThingType {
 
         @Override
         FunctionalIterator<? extends Attribute.String> getInstances();
+
+        @Override
+        FunctionalIterator<? extends Attribute.String> getInstancesExplicit();
 
         void setRegex(Pattern regex);
 
@@ -226,6 +241,9 @@ public interface AttributeType extends ThingType {
 
         @Override
         FunctionalIterator<? extends Attribute.DateTime> getInstances();
+
+        @Override
+        FunctionalIterator<? extends Attribute.DateTime> getInstancesExplicit();
 
         Attribute.DateTime put(LocalDateTime value);
 
--- a/concept/type/EntityType.java
+++ b/concept/type/EntityType.java
@@ -32,6 +32,9 @@ public interface EntityType extends ThingType {
     @Override
     FunctionalIterator<? extends Entity> getInstances();
 
+    @Override
+    FunctionalIterator<? extends Entity> getInstancesExplicit();
+
     void setSupertype(EntityType superType);
 
     Entity create();
--- a/concept/type/RelationType.java
+++ b/concept/type/RelationType.java
@@ -32,6 +32,9 @@ public interface RelationType extends ThingType {
     @Override
     FunctionalIterator<? extends Relation> getInstances();
 
+    @Override
+    FunctionalIterator<? extends Relation> getInstancesExplicit();
+
     void setSupertype(RelationType superType);
 
     void setRelates(String roleLabel);
--- a/concept/type/ThingType.java
+++ b/concept/type/ThingType.java
@@ -37,6 +37,8 @@ public interface ThingType extends Type {
 
     FunctionalIterator<? extends Thing> getInstances();
 
+    FunctionalIterator<? extends Thing> getInstancesExplicit();
+
     void setAbstract();
 
     void unsetAbstract();
--- a/concept/type/impl/AttributeTypeImpl.java
+++ b/concept/type/impl/AttributeTypeImpl.java
@@ -99,14 +99,6 @@ public abstract class AttributeTypeImpl extends ThingTypeImpl implements Attribu
     }
 
     @Override
-    public void setAbstract() {
-        if (getInstancesExplicit().first().isPresent()) {
-            throw exception(TypeDBException.of(TYPE_HAS_INSTANCES_SET_ABSTRACT, getLabel()));
-        }
-        vertex.isAbstract(true);
-    }
-
-    @Override
     public void unsetAbstract() {
         if (getSubtypes().anyMatch(sub -> !sub.equals(this))) {
             throw exception(TypeDBException.of(ATTRIBUTE_UNSET_ABSTRACT_HAS_SUBTYPES, getLabel()));
--- a/concept/type/impl/AttributeTypeImpl.java
+++ b/concept/type/impl/AttributeTypeImpl.java
@@ -22,6 +22,7 @@ import com.vaticle.typedb.core.common.exception.TypeDBException;
 import com.vaticle.typedb.core.common.iterator.FunctionalIterator;
 import com.vaticle.typedb.core.common.iterator.Iterators;
 import com.vaticle.typedb.core.concept.thing.Attribute;
+import com.vaticle.typedb.core.concept.thing.Thing;
 import com.vaticle.typedb.core.concept.thing.impl.AttributeImpl;
 import com.vaticle.typedb.core.concept.type.AttributeType;
 import com.vaticle.typedb.core.concept.type.RoleType;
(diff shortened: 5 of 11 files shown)


concept/type/AttributeType.java
```diff
@@ -41,6 +41,9 @@
     @Override
     FunctionalIterator<? extends Attribute> getInstances();
 
+    @Override
+    FunctionalIterator<? extends Attribute> getInstancesExplicit();
```
flyingsilverfin, line 45: we implement this new API following the style of our other "exact" methods like `getOwnsExplicit()`. This method retrieves instances of this type only, and excludes instances of subtypes.

concept/type/EntityType.java
```diff
@@ -32,6 +32,9 @@
     @Override
     FunctionalIterator<? extends Entity> getInstances();
 
+    @Override
+    FunctionalIterator<? extends Entity> getInstancesExplicit();
```
flyingsilverfin, line 36: new API implementation, following API of `getInstances()`

concept/type/RelationType.java
```diff
@@ -32,6 +32,9 @@
     @Override
     FunctionalIterator<? extends Relation> getInstances();
 
+    @Override
+    FunctionalIterator<? extends Relation> getInstancesExplicit();
```
flyingsilverfin, line 36: new API implementation, following API of `getInstances()`

concept/type/ThingType.java
```diff
@@ -37,6 +37,8 @@
 
     FunctionalIterator<? extends Thing> getInstances();
 
+    FunctionalIterator<? extends Thing> getInstancesExplicit();
```
flyingsilverfin, line 40: new API implementation, following API of `getInstances()`

concept/type/impl/AttributeTypeImpl.java
```diff
@@ -96,14 +98,6 @@ public static AttributeTypeImpl of(GraphManager graphMgr, TypeVertex vertex) {
...
     }
 
-    @Override
-    public void setAbstract() {
-        if (getInstances().first().isPresent()) {
-            throw exception(TypeDBException.of(TYPE_HAS_INSTANCES_SET_ABSTRACT, getLabel()));
-        }
-        vertex.isAbstract(true);
-    }
-
```
flyingsilverfin, line 106: the abstract setter for AttributeTypes turns out to be the exact same as for `ThingType` (the generic implementation), so we can remove this. 

It was here from two iterations ago:
1. we originally used to check that when setting attribute type as abstract, it did not have any subtypes
2. we then moved this validation to when we add set the supertype of a new Attribute type: a new attribute type _must_ have an abstract parent

After this second step, this method ends up being the same as the parent implementation but was not removed before

concept/type/impl/AttributeTypeImpl.java
```diff
@@ -361,6 +355,11 @@ public void setSupertype(AttributeType superType) {
             return instances(v -> AttributeImpl.of(v.asAttribute()));
         }
 
+        @Override
+        public FunctionalIterator<AttributeImpl<?>> getInstancesExplicit() {
+            return empty();
+        }
```
flyingsilverfin, line 361: root types have no instances

concept/type/impl/AttributeTypeImpl.java
```diff
@@ -423,6 +422,11 @@ private Boolean(GraphManager graphMgr, TypeVertex vertex) {
             return instances(v -> new AttributeImpl.Boolean(v.asAttribute().asBoolean()));
         }
 
+        @Override
+        public FunctionalIterator<AttributeImpl.Boolean> getInstancesExplicit() {
+            return instancesExplicit(v -> new AttributeImpl.Boolean(v.asAttribute().asBoolean()));
+        }
```
flyingsilverfin, line 428: boolean attr explicit instances

concept/type/impl/AttributeTypeImpl.java
```diff
@@ -556,6 +560,11 @@ private Long(GraphManager graphMgr, TypeVertex vertex) {
             return instances(v -> new AttributeImpl.Long(v.asAttribute().asLong()));
         }
 
+        @Override
+        public FunctionalIterator<AttributeImpl.Long> getInstancesExplicit() {
+            return instancesExplicit(v -> new AttributeImpl.Long(v.asAttribute().asLong()));
```
flyingsilverfin, line 565: long instances

concept/type/impl/AttributeTypeImpl.java
```diff
@@ -689,6 +698,11 @@ private Double(GraphManager graphMgr, TypeVertex vertex) {
             return instances(v -> new AttributeImpl.Double(v.asAttribute().asDouble()));
         }
 
+        @Override
+        public FunctionalIterator<AttributeImpl.Double> getInstancesExplicit() {
+            return instancesExplicit(v -> new AttributeImpl.Double(v.asAttribute().asDouble()));
```
flyingsilverfin, line 703: double instances

concept/type/impl/AttributeTypeImpl.java
```diff
@@ -822,6 +836,11 @@ private String(GraphManager graphMgr, TypeVertex vertex) {
             return instances(v -> new AttributeImpl.String(v.asAttribute().asString()));
         }
 
+        @Override
+        public FunctionalIterator<AttributeImpl.String> getInstancesExplicit() {
+            return instancesExplicit(v -> new AttributeImpl.String(v.asAttribute().asString()));
```
flyingsilverfin, line 841: string instances

concept/type/impl/AttributeTypeImpl.java
```diff
@@ -994,6 +1013,11 @@ private DateTime(GraphManager graphMgr, TypeVertex vertex) {
             return instances(v -> new AttributeImpl.DateTime(v.asAttribute().asDateTime()));
         }
 
+        @Override
+        public FunctionalIterator<AttributeImpl.DateTime> getInstancesExplicit() {
+            return instancesExplicit(v -> new AttributeImpl.DateTime(v.asAttribute().asDateTime()));
```
flyingsilverfin, line 1018: date instances

concept/type/impl/EntityTypeImpl.java
```diff
@@ -82,6 +82,11 @@ public void setSupertype(EntityType superType) {
         return instances(EntityImpl::of);
     }
 
+    @Override
+    public FunctionalIterator<EntityImpl> getInstancesExplicit() {
+        return instancesExplicit(EntityImpl::of);
```
flyingsilverfin, line 87: entity instances

concept/type/impl/RelationTypeImpl.java
```diff
@@ -81,9 +81,13 @@ public void setLabel(String label) {
 
     @Override
     public void setAbstract() {
-        if (getInstances().first().isPresent()) throw exception(TypeDBException.of(TYPE_HAS_INSTANCES_SET_ABSTRACT, getLabel()));
-        vertex.isAbstract(true);
-        declaredRoles().forEachRemaining(RoleTypeImpl::setAbstract);
+        if (!isAbstract()) {
```
flyingsilverfin, line 84: this is just for cleanliness - setting abstract should do nothing (eg. be idempotent) if already abstract. We technically don't need this check, but it leads to fewer cases to think about in the future if we do -- the validation we run is only scoped to when it is not abstract, we don't have to consider it if it is. It also feels more correct to do the validation for instances when not abstract.

concept/type/impl/RelationTypeImpl.java
```diff
@@ -81,9 +81,13 @@ public void setLabel(String label) {
...
+            if (getInstancesExplicit().first().isPresent()) {
```
flyingsilverfin, line 85: this is the main change: we use `getInstancesExplicit()` instead of `getInstances()` - this allows us to toggle abstractness even when a child has instances, but the type itself does not

concept/type/impl/RelationTypeImpl.java
```diff
@@ -113,6 +117,11 @@ public void setSupertype(RelationType superType) {
         return instances(RelationImpl::of);
     }
 
+    @Override
+    public FunctionalIterator<RelationImpl> getInstancesExplicit() {
+        return instancesExplicit(RelationImpl::of);
```
flyingsilverfin, line 122: relation instances

concept/type/impl/ThingTypeImpl.java
```diff
@@ -101,7 +103,7 @@ public static ThingTypeImpl of(GraphManager graphMgr, TypeVertex vertex) {
     public void setAbstract() {
         if (!isAbstract()) {
             validateIsNotDeleted();
-            if (getInstances().first().isPresent()) {
+            if (getInstancesExplicit().first().isPresent()) {
```
flyingsilverfin, line 106: The main change - use `getInstancesExplicit()` instead of `getInstances()` to allow setting abstractness even when a child has instances but the type does not

concept/type/impl/ThingTypeImpl.java
```diff
@@ -517,6 +519,11 @@ public Root(GraphManager graphMgr, TypeVertex vertex) {
             });
         }
 
+        @Override
+        public FunctionalIterator<ThingImpl> getInstancesExplicit() {
+            return empty();
```
flyingsilverfin, line 524: root has no instances

concept/type/impl/TypeImpl.java
```diff
@@ -122,6 +122,10 @@ public boolean isAbstract() {
                 .flatMap(t -> graphMgr.data().getReadable(t.vertex)).map(thingConstructor);
     }
 
+    <THING> FunctionalIterator<THING> instancesExplicit(Function<ThingVertex, THING> thingConstructor) {
+        return graphMgr.data().getReadable(vertex).map(thingConstructor);
+    }
```
flyingsilverfin, line 127: helper method used everywhere to get explicit instances (see method above to see implementation of explicit + child instances)

dependencies/vaticle/repositories.bzl
```diff
@@ -48,8 +48,8 @@ def vaticle_typedb_protocol():
 def vaticle_typedb_behaviour():
     git_repository(
         name = "vaticle_typedb_behaviour",
-        remote = "https://github.com/vaticle/typedb-behaviour",
-        commit = "05063b44ee914ad1c71a580c8ecc6454e33c302a", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
+        remote = "https://github.com/flyingsilverfin/typedb-behaviour",
```
flyingsilverfin, line 51: will bump back after this is merged to master -- we updated behaviour cases to test we are allowed to set abstractness when there are indirect but not new explicit instances

test/behaviour/concept/type/thingtype/ThingTypeSteps.java
```diff
@@ -133,6 +133,12 @@ public void thing_type_set_abstract(RootLabel rootLabel, String typeLabel, boole
         else get_thing_type(rootLabel, typeLabel).unsetAbstract();
     }
 
+    @When("{root_label}\\( ?{type_label} ?) set abstract: {bool}; throws exception")
+    public void thing_type_set_abstract_throws(RootLabel rootLabel, String typeLabel, boolean isAbstract) {
```
flyingsilverfin, line 137: a new step to used in the new scenarios to verify setting abstractness throws in some cases

concept/type/impl/ThingTypeImpl.java
haikalpribadi: same as above, perhaps flip the `if` and return to break out?

haikalpribadi (approved):
Looks good! Left just a few formatting comments.

concept/type/impl/ThingTypeImpl.java
flyingsilverfin: I actually prefer the nested style because you can visually "see" the logical dependencies/where the code flows, but happy to change it

PR #6504, opened 2022-01-05
Utilise IID type information in type inference

flyingsilverfin:

We can use some of the information in IIDs by extracting the contained type IID and using that for type resolution, rather than ignoring IIDs completely. We had previous removed the code path within type inference that examined the IIDs within the query, following the logic that a static type checker should not validate the data specified within the query, only the structure & typing. This changed is outlined in https://github.com/vaticle/typedb/issues/6477.

* extract type information from IIDs and include them in type inference
* add test to the suite of integration tests for type inference

### DIFF
--- a/logic/tool/TypeInference.java
+++ b/logic/tool/TypeInference.java
@@ -426,7 +426,7 @@ public class TypeInference {
         private void registerIID(TypeVariable resolver, IIDConstraint constraint) {
             TypeVertex type = graphMgr.schema().convert(VertexIID.Thing.of(constraint.iid()).type());
             if (type == null) throw TypeDBException.of(UNSATISFIABLE_PATTERN, conjunction, constraint);
-            traversal.labels(resolver.id(), type.properLabel());
+            restrict(resolver.id(), iterate(type));
         }
 
         private void registerHas(TypeVariable resolver, HasConstraint hasConstraint) {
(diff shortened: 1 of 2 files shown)


logic/tool/TypeInference.java
```diff
@@ -416,6 +419,16 @@ private void registerIs(TypeVariable resolver, IsConstraint isConstraint) {
...
 
+        /**
+         * We only extract the type of the IID, and include this information in the type inference
+         * We only look at the type prefix because the static type checker (this class) only read schema, not data
+         */
+        private void registerIID(TypeVariable resolver, IIDConstraint constraint) {
+            TypeVertex type = graphMgr.schema().convert(VertexIID.Thing.of(constraint.iid()).type());
+            if (type == null) throw TypeDBException.of(UNSATISFIABLE_PATTERN, conjunction, constraint);
+            restrict(resolver.id(), iterate(type));
+        }
```
flyingsilverfin, line 430: restrict the allowed types of a variable if we have an IID by extracting the type of the IID

test/integration/logic/TypeInferenceTest.java
```diff
@@ -188,6 +191,30 @@ public void is_inference() throws IOException {
...
+        // using a person IID, the attribute can only be a name or email, but not a dog label
+        String queryString = "match $p iid " + person.getIID().toHexString() + "; $p has $a;";
+        Disjunction disjunction = createDisjunction(queryString);
+        typeInference.infer(disjunction);
+
+        Map<String, Set<String>> expected = new HashMap<>() {{
+            put("$p", set("person"));
+            put("$a", set("name", "email"));
+        }};
+
```
flyingsilverfin, line 214: test that the IID type information restricts both the `$p` and by connection the attributes we search for

jmsfltchr (approved):

PR #6511, opened 2022-01-20
Improve rule validation with filterable type inference

flyingsilverfin:

We make rule validation more efficient by allowing type inference to infer type permutations for fewer variables.

* We pass the variables from the `then` of the rule into the type inference operation when getting `when` type permutations that must be insertable in the `then`. This should allow the traversal engine (if it is clever enough!) to do less work by observing the filter variables that must be returned


flyingsilverfin:
Integrated into work on traversal 3.0

PR #6512, opened 2022-01-21
Refactor config framework

lolski:
**For the title of this PR:** please follow the grammatical rules of a usual publication title, without capitalisation (except for the first letter). Thus, the title should NOT CONTAIN CODE: no dots, no parentheses, no backticks, no brackets, etc. It needs to be distinctive (not detailed) and succinct (not lengthy). Details of this PR will go in the description. **For the description of this PR:** please replace every line in curly brackets ( { like this } ) with an appropriate description following the guidance. Finally, **please remove this paragraph**.

  - its responsibility, 


server/parameters/util/ArgsParser.java
```diff
@@ -0,0 +1,85 @@
...
+package com.vaticle.typedb.core.server.parameters.util;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public class ArgsParser<RESULT> {
```
lolski, line 28: We refactored the `CommandLine` class in the following ways:

- It has been renamed to `ArgsParser` to clearly describe its responsibility: to parse `args` array in the `main` file.
- The `Option` class has been extracted into its own class.
- The `CliHelp` interface has been deleted since it is superfluous.
- The `Help`  class has been extracted into a standalone class since it is shared by several classes: `SubcommandParser`, `OptionParser`, and `YamlParser`

server/parameters/util/ArgsParser.java
```diff
@@ -0,0 +1,85 @@
...
+                String[] tokens = Arrays.copyOfRange(args, 0, commandTokens.length);
+                if (Arrays.equals(commandTokens, tokens)) {
+                    return Optional.of(parser);
+                }
+            }
+        }
+        return Optional.empty();
+    }
+
+    private List<SubcommandParser<? extends RESULT>> sortBySpecificity(ArrayList<SubcommandParser<? extends RESULT>> parsers) {
```
lolski, line 62: It used to be called `specificToGeneralCommands`, which isn't a good name. The new name should clearly describe its responsibility better, which is to sort subcommand parsers based on their specificity.

server/parameters/util/SubcommandParser.java
```diff
@@ -0,0 +1,85 @@
...
+
+import com.vaticle.typedb.core.common.exception.TypeDBException;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Server.CLI_OPTION_MISSING_PREFIX;
+
+public abstract class SubcommandParser<RESULT> {
```
lolski, line 29: The class is responsible for parsing the given subcommand according to the specified `tokens`. It is extracted from the `ServerSubcommand` class (which was called `Command`, prior to this PR).

server/parameters/util/OptionParser.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Optional;
+import java.util.Set;
+
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Server.CLI_FLAG_OPTION_HAS_VALUE;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Server.CLI_OPTION_REQUIRES_TYPED_VALUE;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Server.DUPLICATE_CLI_OPTION;
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+import static com.vaticle.typedb.core.server.parameters.util.YamlParser.concatenate;
+
+public abstract class OptionParser {
```
lolski, line 33: This class is responsible for parsing "CLI option" such as `--help` and `--version`. It has been extracted from `ArgsParser` (which was called `CommandLine`, prior to this PR).

server/parameters/util/Option.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ */
+
+package com.vaticle.typedb.core.server.parameters.util;
+
+import javax.annotation.Nullable;
+import java.util.Objects;
+import java.util.Optional;
+
+public class Option {
```
lolski, line 25: This class represents a "CLI option". It has been extracted from `ArgsParser` (which used to be called `CommandLine`, prior to this PR).

server/parameters/ServerSubcommand.java
```diff
@@ -0,0 +1,165 @@
...
+package com.vaticle.typedb.core.server.parameters;
+
+import com.vaticle.typedb.core.common.exception.TypeDBException;
+
+import java.nio.file.Path;
+
+import static com.vaticle.typedb.common.util.Objects.className;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_CAST;
+
+public abstract class ServerSubcommand {
```
lolski, line 28: Represents the parsed value for the "server" subcommand.

server/parameters/ConfigFactory.java
```diff
@@ -0,0 +1,120 @@
...
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Server.CLI_OPTION_REQUIRES_VALUE;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Server.CONFIG_FILE_NOT_FOUND;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Server.CONFIG_YAML_MUST_BE_MAP;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Server.ENV_VAR_NOT_FOUND;
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+import static com.vaticle.typedb.core.server.common.Constants.CONFIG_PATH;
+
+public class ConfigFactory {
+
+    public static Config create(ConfigParser parser) {
```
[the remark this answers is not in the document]
flyingsilverfin: The name `create` looks ok here but I would still call them `.config` because of the convention we have where `.create()` could be static constructor for the factory itself

server/parameters/ConfigFactory.java
```diff
@@ -0,0 +1,120 @@
...
+            yaml.put(key, Yaml.load(get(options, key)));
+        }
+        return new Yaml.Map(yaml);
+    }
+
+    private static String get(Set<Option> source, String key) {
+        Set<String> values = iterate(source).filter(opt -> opt.name().equals(key))
+                .map(opt -> opt.stringValue().get()).toSet();
+        if (values.size() == 1) return values.iterator().next();
+        else return "[" + String.join(", ", values) + "]";
```
flyingsilverfin, line 91: oh cool i forgot we have support for merging multiple repeated cli args into a list in yaml!

server/parameters/ConfigFactory.java
```diff
@@ -0,0 +1,120 @@
...
+                    String envVarName = valueStr.substring(1);
+                    if (System.getenv(envVarName) == null) throw TypeDBException.of(ENV_VAR_NOT_FOUND, valueStr);
+                    else yaml.put(key, Yaml.load(System.getenv(envVarName)));
+                }
+            } else if (value.isMap()) {
+                substituteEnvVars(value.asMap());
+            }
+        }
+    }
+}
```
flyingsilverfin, line 120: yeah this looks like the right place to put all these methods. Nice find

server/parameters/ConfigParser.java
```diff
@@ -0,0 +1,472 @@
...
+        private static final String description = "Server and networking configuration.";
+
+        private static final Predefined<InetSocketAddress> address =
+                predefined("address", "Address to listen for GRPC clients on.", INET_SOCKET_ADDRESS);
+        private static final Set<Predefined<?>> parsers = set(address);
+
+        @Override
+        public Config.Server parse(Yaml yaml, String path) {
+            if (yaml.isMap()) {
+                validatedRecognisedParsers(parsers, yaml.asMap().keys(), path);
```
flyingsilverfin, line 92: i think this should be called `validatePredefinedParsers` now

server/parameters/ConfigParser.java
```diff
@@ -0,0 +1,472 @@
...
+    private static void validatedRecognisedParsers(Set<Predefined<?>> parsers, Set<String> keys, String path) {
+        Set<String> unrecognisedKeys = new HashSet<>(keys);
+        parsers.forEach(parser -> unrecognisedKeys.remove(parser.key()));
+        if (!unrecognisedKeys.isEmpty()) {
+            Set<String> childPaths = iterate(unrecognisedKeys).map(key -> YamlParser.concatenate(path, key)).toSet();
+            throw TypeDBException.of(UNRECOGNISED_CONFIGURATION_OPTIONS, childPaths);
+        }
+    }
+
+}
```
flyingsilverfin, line 472: class looks good! Splitting out data structure and parser was definitely the right call :)

server/parameters/ServerSubcommandParser.java
```diff
@@ -0,0 +1,154 @@
...
+
+import static com.vaticle.typedb.common.collection.Collections.list;
+import static com.vaticle.typedb.common.collection.Collections.set;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Server.CLI_OPTION_REQUIRED;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Server.CLI_OPTION_UNRECOGNISED;
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+
+public class ServerSubcommandParser {
+
+    private static void validateRequiredArgs(Set<OptionParser> requiredParsers, Set<Option> options) {
```
flyingsilverfin, line 39: argument naming here hints this should be called `validateRequiredOptions`, and below `validateUnrecognisedOptions`

server/parameters/ServerSubcommandParser.java
```diff
@@ -0,0 +1,154 @@
...
+            Optional<Path> configPath = configPathParser.parse(auxOptions);
+            Config config = configPath
+                    .map(path -> ConfigFactory.create(path, configOptions, configParser))
+                    .orElse(ConfigFactory.create(configOptions, configParser));
+            return new ServerSubcommand.Server(debugParser.parse(auxOptions), helpParser.parse(auxOptions),
+                    versionParser.parse(auxOptions), config);
+        }
+
+        private Set<Option> findAuxiliaryOptions(Set<Option> options) {
+            return iterate(options).filter(opt1 -> iterate(auxiliaryParsers).anyMatch(opt2 -> opt2.name.equals(opt1.name()))).toSet();
```
flyingsilverfin, line 85: inner names could be `opt` and then `auxOpt` rather than `opt1`/`opt2`

server/parameters/ServerSubcommandParser.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        private Set<Option> excludeOptions(Set<Option> options, Set<Option> exclude) {
+            return iterate(options).filter(opt -> !exclude.contains(opt)).toSet();
+        }
+
+        @Override
+        public List<HelpEntry> helpEntries() {
+            List<HelpEntry> aux = list(helpParser.helpEntry(""), versionParser.helpEntry(""), debugParser.helpEntry(""), configPathParser.helpEntry(""));
```
flyingsilverfin, line 94: got confused on first reading, I think this could be cleared up by calling this `auxEntries`, so that the line below looks like it's concatenating two lists

server/parameters/util/OptionParser.java
```diff
@@ -0,0 +1,139 @@
...
+
+    public java.lang.String name() {
+        return name;
+    }
+
+    java.lang.String description() {
+        return description;
+    }
+
+    public HelpEntry helpEntry(java.lang.String optionScope) {
```
flyingsilverfin, line 53: I think since this is no longer implementing the same interface, it doesn't make sense to have an `optionScope` here - all options are the root scope of the arg parser right?

server/parameters/ServerSubcommandParser.java
```diff
@@ -0,0 +1,154 @@
...
+        protected ServerSubcommand.Export parse(Set<Option> options) {
+            validateRequiredArgs(parsers, options);
+            validateUnrecognisedArgs(parsers, options);
+            return new ServerSubcommand.Export(databaseParser.parse(options).get(), filePathParser.parse(options).get(),
+                    portParser.parse(options).get());
+        }
+
+        @Override
+        public List<HelpEntry> helpEntries() {
+            return list(databaseParser.helpEntry(""), filePathParser.helpEntry(""), portParser.helpEntry(""));
```
flyingsilverfin, line 151: left a comment related to this on the `OptionParser.helpEntry()`

server/parameters/util/ArgsParser.java
```diff
@@ -0,0 +1,85 @@
...
+    public String help() {
+        StringBuilder help = new StringBuilder("Usage: \n");
+        parsers.forEach(parser -> {
+            help.append(parser.help());
+            help.append("\n");
+        });
+        return help.toString();
+    }
+
+}
```
flyingsilverfin, line 85: looks good!

server/parameters/util/HelpEntry.java
```diff
@@ -0,0 +1,90 @@
...
+            }
+
+            private boolean anyLeafContents() {
+                return iterate(entries).anyMatch(content -> content instanceof Simple);
+            }
+        }
+
+    }
+
+    public static class Simple extends HelpEntry {
```
flyingsilverfin, line 71: don't like this name... can we choose one of the existing ones that isn't related to `Yaml` specifically?
there are some floating around already:
- `KeyValue`
- `Option`
- `Parameter`
what do you think?

flyingsilverfin: (I think `Option` is the closest to the domain, it's used for `ArgParser` option helps right?)

server/parameters/util/OptionParser.java
```diff
@@ -0,0 +1,139 @@
...
+public abstract class OptionParser {
+
+    public final java.lang.String name;
+    private final java.lang.String description;
+    private final java.lang.String typeDescription;
+
+    public OptionParser(java.lang.String name, java.lang.String description, java.lang.String typeDescription) {
+        this.name = name;
+        this.description = description;
+        this.typeDescription = typeDescription;
```
flyingsilverfin, line 42: i recon this is `valueDescription`, or even more precisely `valueTypeDescription`, but I think the first option is clear enough

server/parameters/util/HelpEntry.java
```diff
@@ -0,0 +1,90 @@
...
+        this.description = description;
+    }
+
+    public static abstract class Yaml extends HelpEntry {
+
+        private Yaml(String scopedName, String description) {
+            super(scopedName, description);
+        }
+
+        public static class Grouped extends Yaml {
```
flyingsilverfin, line 42: ok on reading down in the usages more it feels strange to have a single subtype of `Yaml` for `Grouped`. Either we should create a second "simple" or "leaf"  HelpEntry for Yaml, that is used in the `YamlParser`, or flatten the `Grouped` into the `Yaml` with a decent name. 

For example, the content of `Yaml.Grouped` actually itself has nothing yaml-specific, but only that is holding a group of entries.

server/parameters/util/YamlParser.java
```diff
@@ -0,0 +1,335 @@
...
+        public static class Dynamic<TYPE> extends KeyValue {
+
+            private final Value<TYPE> valueParser;
+
+            private Dynamic(String description, Value<TYPE> valueParser) {
+                super(description);
+                this.valueParser = valueParser;
+            }
+
+            public Map<String, TYPE> parseFrom(Yaml.Map yaml, String path, Predefined<?>... exclude) {
```
flyingsilverfin, line 117: ohhh this signature looks nice now with `exclude` being of type `Predefined`!

server/parameters/util/YamlParser.java
```diff
@@ -0,0 +1,335 @@
...
+
+        boolean isCompound() {
+            return false;
+        }
+
+        Compound<TYPE> asCompound() {
+            throw TypeDBException.of(ILLEGAL_CAST, className(getClass()), className(Compound.class));
+        }
+
+        public static abstract class Compound<T> extends Value<T> {
```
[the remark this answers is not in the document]
flyingsilverfin: we actually use the term `Compound` in reasoner to represent a non-atomic piece of reasoning work (eg. composed of other parts) so I think I like it haha

server/TypeDBServer.java
```diff
@@ -108,7 +112,7 @@ protected TypeDBServer(Configuration configuration, boolean debug, Factory facto
     }
 
     private void configureLogging() {
-        configureLogback((LoggerContext) LoggerFactory.getILoggerFactory(), configuration);
+        configure((LoggerContext) LoggerFactory.getILoggerFactory(), config);
```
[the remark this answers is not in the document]
lolski: Fixed in af4066fbf48d3bb73f54fb0aa14a56fe9fd1eb0c.

server/TypeDBServer.java
```diff
@@ -122,32 +126,33 @@ private void configureAndVerifyJavaVersion() {
     }
 
     private void configureAndVerifyDataDir() {
-        if (!Files.isDirectory(configuration.storage().dataDir())) {
-            if (configuration.storage().dataDir().equals((new Configuration.Parser()).getConfig().storage().dataDir())) {
+        if (!Files.isDirectory(config.storage().dataDir())) {
+            Config defaultConfig = ConfigFactory.create(configParser);
```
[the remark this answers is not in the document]
lolski: Good point. Fixed in af4066fbf48d3bb73f54fb0aa14a56fe9fd1eb0c.

server/parameters/Config.java
```diff
@@ -0,0 +1,373 @@
...
+
+                public boolean isEnabled() {
+                    return enable;
+                }
+
+                public Output.Type.File output() {
+                    assert output != null;
+                    return output;
+                }
+
```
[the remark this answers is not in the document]
lolski: I see! Fixed in bdd4c0e6974cb134e4eb3dab781fe1cb87b89347.

server/parameters/Config.java
```diff
@@ -0,0 +1,373 @@
...
+        public Optional<String> username() {
+            return Optional.ofNullable(username);
+        }
+
+        public Optional<String> token() {
+            return Optional.ofNullable(token);
+        }
+    }
+
```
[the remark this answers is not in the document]
lolski: Fixed in bdd4c0e6974cb134e4eb3dab781fe1cb87b89347 too.

There are also a bunch of other places that I've fixed as well in that commit.

server/parameters/ServerSubcommandParser.java
lolski: Fixed in 443186cb5e02eba13fd86437dd8e39d299a5e64c.

server/parameters/ServerSubcommandParser.java
lolski: Oops! Fixed in 443186cb5e02eba13fd86437dd8e39d299a5e64c.

server/parameters/ConfigParser.java
lolski: Ah much better. Fixed in 443186cb5e02eba13fd86437dd8e39d299a5e64c.

server/parameters/util/OptionParser.java
lolski: Good idea. Fixed in 0dcd92aeacf469ecb0920deddb96c0d03c908e5b.

server/parameters/util/HelpEntry.java
lolski: I don't think we should because then we're naming it based on where it's used.

It is shared equally by `OptionParser`, `YamlParser.KeyValue.Dynamic`, and `YamlParser.KeyValue.Predefined` and therefore the name shouldn't be associated with any of them.

I think `HelpEntry.Simple` is a good name because it represents what it is - a simple entry that has no nesting / grouping.

What do you think?

server/parameters/util/HelpEntry.java
lolski: > For example, the content of Yaml.Grouped actually itself has nothing yaml-specific, but only that is holding a group of entries.

Actually, I deliberately put them under Yaml because it is Yaml specific - take a look at how we render the String with `.` separating the children.

server/parameters/util/Help.java
```diff
@@ -0,0 +1,25 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+package com.vaticle.typedb.core.server.parameters.util;
+
+public interface Help {
+
+    String name();
```
flyingsilverfin, line 23: missing newlines in between :P

flyingsilverfin (approved):

flyingsilverfin:
Joshua's key takeaways from this version of the configuration parsing, compared to the prior implemenation I created.
Many of the changes Ganesh made could have been derived from these principles.

**Consistent application of the same abstraction**
We had multiple places existing abstractions should have been utilised further, and pushed into all corners of the code rather than where they were just strictly necessary.
Example 1: We have `Value` parsers for YAML key-value pairs. The entire YAML blob itself should have been parsed with a  a `Value` parser, just like any of its nested children. Instead, the entry parser was a free-form class. By not requiring the top-level parser to be its own subtype of `Value` parser, we allowed other tasks to leak in that did not fit into the domain of Value parsers.
Example 2: We had two usages of `help()`, in different places -- for `KeyValue` and for `Value` parsers. The existing interface requiring `help()` to be implemented however was only implemented by `KeyValue`, and not `Value`, which was confusing to the user of the APIs

One advise to implement this point would be to **keep iterating on the high-level design first**, before pushing the implementation everywhere it should go. This would identify the correct abstractions earlier and end up saving refactoring later.

**Mixing code with different domains and responsibilities closely**
Throughout the code base we make extensive use of nested classes. However, they are almost always of the same domain and responsibility. In this previous iteration, we had one `Configuration` class, which contained two nested trees of structures: one for the resulting value classes that were parsed, and one for the parsers themselves.
The clue for this should have been that the `Configuration` class did not have one responsibility (was both parsing and holding the parsed data). Additionally the sibling classes within `Configuration` were both the Parser for configuration, as well as the Configuration's child data classes. Siblings in general should be of the same domain and complexity.

**Closely follow user's expected separated concerns**
Given that this is a user-facing API, the programmatic API should follow what the user sees. In general we did this well with making the YAML structure reflected in Parsers. However, where the previous version did not do we well was to mix together Configuration and CLI options together without clear separation. In this version, it is clear that the set of configuration and cli options together are considered "args" to the server. "args" are split into Configuration and auxiliary options, which more closely follows the user's instinctive model. 

Additionally, the new version separates out generic parsing logic into a new package called `parser`, which is clear in its naming and intent. The subtypes of the parsers are the implementation of how to process server parameters.

**A bunch of utils are probably a missing abstraction**
There were two missing class-level abstractions that were missing, both derived from classes in `//server/common:util.java`. The first is the `ConfigFactory`, which is an abstraction that merges together override options, as well as actually instantiating Configuration objects. This class absorbed a lot of the methods in the util class into the more correct place. 

The second example is the `Logback` configuration, which should have been recognised as a common set of configuration methods and become its own class outside of `//server/common`

PR #6513, opened 2022-01-26
Guarantee 'match' query answer uniqueness

flyingsilverfin:

TypeDB guarantees uniqueness of `match` query answers in all cases. Previously, there were cases where violations of answer uniqueness occurred, such as when the server made particular query optimisations. 

* use a `distinct()` set to guarantee concept map uniqueness
  * where the distinct set is applied depends on whether the start variable is  part of the variables returned to the user
  * if the start variable is not returned, we have to use a more expensive larger distinct set to guarantee uniqueness - ther is room for optimisation in the future.

## Example of bug:

If this these two answers exist, trying to return just `$x`:
```
{$_0: 0x10, $x: 0x2}
{$_0: 0x10, $x: 0x2}
```

If the start variable is `$_0`, then the old code would only use a distinct set for the remainder of the answer, so we end up with two answers of `{$x: 0x2}`.
If the start variable was `$x`, we would end up with just one answer of `{$x: 0x2}`.

### DIFF
--- a/traversal/procedure/GraphProcedure.java
+++ b/traversal/procedure/GraphProcedure.java
@@ -203,8 +203,8 @@ public class GraphProcedure implements PermutationProcedure {
                     sv -> new GraphIterator(graphMgr, sv, this, params, filter).distinct()
             );
         } else {
+            // TODO we can reduce the size of the distinct() set if the traversal engine doesn't overgenerate as much
             return startVertex().iterator(graphMgr, params).flatMap(
-                    // TODO we can reduce the size of the distinct() set if the traversal engine doesn't overgenerate as much
                     sv -> new GraphIterator(graphMgr, sv, this, params, filter)
             ).distinct();
         }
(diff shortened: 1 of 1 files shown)


traversal/procedure/GraphProcedure.java
```diff
@@ -176,10 +176,17 @@ private void assertWithinFilterBounds(Set<Identifier.Variable.Retrievable> filte
...
-        return async(startVertex().iterator(graphMgr, params).map(
-                // TODO we can reduce the size of the distinct() set if the traversal engine doesn't overgenerate as much
-                v -> new GraphIterator(graphMgr, v, this, params, filter).distinct()
-        ), parallelisation);
+        if (startVertex().id().isRetrievable() && filter.contains(startVertex().id().asVariable().asRetrievable())) {
+            return async(startVertex().iterator(graphMgr, params).map(
+                    // TODO we can reduce the size of the distinct() set if the traversal engine doesn't overgenerate as much
+                    v -> new GraphIterator(graphMgr, v, this, params, filter).distinct()
+            ), parallelisation);
+        } else {
```
flyingsilverfin, line 184: if the start vertex is NOT part of the filtered/returned set of variables, we need to put the `distinct()` set at the outer layer, or we can have answers from remainder of the variables

we do this for both the async producer and non-parallel iterator methods

haikalpribadi (approved):

PR #6515, opened 2022-02-01
Fix iterators to recycle sub-iterators correctly

flyingsilverfin:

All TypeDB iterators (`FunctionalIterators`) correctly release resources when asked to `recycle()`, allowing things like RocksDB iterators to be reused and preventing a memory leak. We discover and fix two remaining places where iterators were not releasing resources correctly: `FlatMapped` and `GraphIterator`.

* Discover that it's possible to cause a "memory leak" / growing amount of memory in a long-lived transaction. This is caused by both `FlatMapped` iterators and `GraphIterator` (eg. the traversal executor) not releasing sub-iterators correctly.
* Fix both `FlatMapped` and `GraphIterator` to release their iterators when completing

### DIFF
--- a/common/iterator/FlatMappedIterator.java
+++ b/common/iterator/FlatMappedIterator.java
@@ -73,5 +73,6 @@ class FlatMappedIterator<T, U> extends AbstractFunctionalIterator<U> {
     @Override
     public void recycle() {
         sourceIterator.recycle();
+        if (currentIterator != null) currentIterator.recycle();
     }
 }
--- a/traversal/scanner/GraphIterator.java
+++ b/traversal/scanner/GraphIterator.java
@@ -86,6 +86,11 @@ public class GraphIterator extends AbstractFunctionalIterator<VertexMap> {
         }
     }
 
+    private void setCompleted() {
+        state = State.COMPLETED;
+        recycle();
+    }
+
     @Override
     public boolean hasNext() {
         try {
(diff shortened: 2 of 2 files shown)


common/iterator/FlatMappedIterator.java
```diff
@@ -73,5 +73,6 @@ public U next() {
     @Override
     public void recycle() {
         sourceIterator.recycle();
+        if (currentIterator != null) currentIterator.recycle();
```
flyingsilverfin, line 76: the active iterator we spawn from the `sourceIterator` must also be recycled, if there is one

traversal/scanner/GraphIterator.java
```diff
@@ -45,6 +46,7 @@
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.RESOURCE_CLOSED;
 import static java.util.stream.Collectors.toMap;
 
+@NotThreadSafe
```
flyingsilverfin, line 49: Just wanted to make sure this is explicitly noted as not thread safe. Previously, it was only implicit knowledge that it should be used from a single thread.

traversal/scanner/GraphIterator.java
```diff
@@ -115,6 +117,11 @@ else if (state == State.INIT) {
         }
     }
 
+    private void setCompleted() {
+        state = State.COMPLETED;
+        recycle();
```
flyingsilverfin, line 122: it's possible to fail half way through a `GraphIterator` because something in the middle of the plan does not have an answer. However, this means we have to recycle all the previous iterators that took us to this point!

traversal/scanner/GraphIterator.java
```diff
@@ -344,6 +351,8 @@ private VertexMap toVertexMap(Map<Identifier, Vertex<?, ?>> answer) {
 
     @Override
     public void recycle() {
+        iterators.values().forEach(FunctionalIterator::recycle);
+        iterators.clear();
```
flyingsilverfin: which is what we do here - all iterators we opened so far need to be recycled

traversal/scanner/GraphIterator.java
```diff
@@ -344,6 +351,7 @@ private VertexMap toVertexMap(Map<Identifier, Vertex<?, ?>> answer) {
 
     @Override
     public void recycle() {
+        iterators.values().forEach(FunctionalIterator::recycle);
```
flyingsilverfin, line 354: which is what we do here - all opened iterators so far need to be recycled when recycle() is called.

Can't believe we didn't catch this before!

haikalpribadi (approved):

PR #6520, opened 2022-02-15
Fix bazel server binary by preparing required directories

flyingsilverfin:

We address the requirement of pre-creating directory structures before running the server binary via bazel, which was previously broken. The new requirements were caused by #6512 in which we expect the configuration file and data directory to exist at server bootup. 

* Use a bazel `genrule` to pre-create the data directory, and copy the configuration file into the expected location before executing the server binary
* Fix a bug in loading of the configuration using the `ConfigFactory`, which should only load the default configuration file lazily if there is no configuration path given otherwise

### DIFF
--- a/server/BUILD
+++ b/server/BUILD
@@ -95,8 +95,9 @@ native_java_libraries(
 )
 
 genrule(
-    name = "create-config-and-data-dir",
+    name = "prepare-binary-directories",
     srcs = [":config"],
+    # TODO: can we find a way to create a directory without including an empty file in the outs?
     cmd = "cp $(location :config) $(@D)/conf/ && touch $(@D)/data/empty",
     outs = ["conf/config.yml", "data/empty"]
 )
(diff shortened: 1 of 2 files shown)


server/BUILD
```diff
@@ -94,12 +94,20 @@ native_java_libraries(
...
 )
 
+genrule(
+    name = "prepare-server-directories",
+    srcs = [":config"],
+    # TODO: can we find a way to create a directory without including an empty file in the outs?
+    cmd = "cp $(location :config) $(@D)/conf/ && touch $(@D)/data/empty",
+    outs = ["conf/config.yml", "data/empty"]
+)
+
```
flyingsilverfin, line 104: this is the new rule that prepares the directory structure expected by the server binary - copies the `config.yml` and create the `data` directory

server/parameters/ServerSubcommandParser.java
```diff
@@ -77,7 +77,7 @@ public Server(ConfigParser configParser) {
             Optional<Path> configPath = Server.configPath.parse(auxOptions);
             Config config = configPath
                     .map(path -> ConfigFactory.config(path, configOptions, configParser))
-                    .orElse(ConfigFactory.config(configOptions, configParser));
+                    .orElseGet(() -> ConfigFactory.config(configOptions, configParser));
```
flyingsilverfin, line 80: This was not lazy before, and could fail if the default config path did not exist!

lolski (approved):

PR #6521, opened 2022-02-15
Significantly faster rule conclusion insertability validation

flyingsilverfin:

Rule validation that checks each permutation of types from the `when` of the rule is insertable in the `then` of the rule is made much more performant. This allows very widely applicable rules to be validated and committed in a reasonable amount of time. Previously, when a rule could insert many permutations of types (>100k) the validation could appear to stall the server.

* we pass a filter into type inference, so it returns fewer type permutations
  * when the query planning gets smarter, we can make this even faster by planning the variables we want to return earlier in the plan. This will allow us to switch from overgenerate and filter, to generate fewer permutations in the first place
* we use a set `contains()` instead of `O(n)` check to see whether a permutation is valid. This hugely improves performance when there are many valid permutations, eg. > 10k that have to be checked for each of >10k when type permutations


logic/Rule.java
```diff
@@ -402,12 +402,16 @@ public void validate(LogicManager logicMgr, ConceptManager conceptMgr) {
         }
 
         private void validateInsertable(LogicManager logicMgr) {
-            FunctionalIterator<Map<Identifier.Variable.Name, Label>> whenCombinations = logicMgr.typeInference().typePermutations(rule.when, false);
-            Set<Map<Identifier.Variable.Name, Label>> allowedThenCombinations = logicMgr.typeInference().typePermutations(rule.then, true).toSet();
+            Set<Identifier.Variable.Retrievable> sharedIDs = iterate(rule.then.identifiers())
+                    .filter(thenID -> thenID.isName() && rule.when.identifiers().contains(thenID))
+                    .map(Identifier.Variable::asRetrievable).toSet();
```
flyingsilverfin, line 407: identify variables that are shared between `when` and `then`

logic/Rule.java
```diff
@@ -402,12 +402,16 @@ public void validate(LogicManager logicMgr, ConceptManager conceptMgr) {
...
+            FunctionalIterator<Map<Identifier.Variable.Name, Label>> whenCombinations = logicMgr.typeInference().typePermutations(rule.when, false, sharedIDs);
+            Set<Map<Identifier.Variable.Name, Label>> insertableThenCombinations = logicMgr.typeInference().typePermutations(rule.then, true, sharedIDs).toSet();
```
flyingsilverfin, line 409: permutations for when and then only return permutations of the shared variables - a significantly smaller number for each!

logic/Rule.java
```diff
@@ -402,12 +402,16 @@ public void validate(LogicManager logicMgr, ConceptManager conceptMgr) {
...
 
             whenCombinations.forEachRemaining(nameLabelMap -> {
-                if (allowedThenCombinations.stream().noneMatch(thenMap -> nameLabelMap.entrySet().containsAll(thenMap.entrySet())))
+                if (!insertableThenCombinations.contains(nameLabelMap)) {
```
flyingsilverfin, line 412: using a `contains` instead of a `stream()` converts a O(n) operation into an O(1) operation!

logic/tool/TypeInference.java
```diff
@@ -234,10 +235,14 @@ private TraversalBuilder(Conjunction conjunction, GraphTraversal.Type initialTra
...
         }
 
-        public Set<Identifier.Variable.Retrievable> retrievedResolvers() {
+        public Set<Identifier.Variable.Retrievable> retrievableInferenceIDs() {
             return iterate(resolverToOriginal.keySet()).filter(Identifier::isRetrievable).map(Identifier.Variable::asRetrievable).toSet();
         }
 
+        public Set<Identifier.Variable.Retrievable> toInferenceIDs(Set<Identifier.Variable.Retrievable> originalIDs) {
+            return iterate(originalIDs).map(id -> originalToResolver.get(id).id().asRetrievable()).toSet();
+        }
```
flyingsilverfin, line 244: new methods which enable passing a filter into the type inference

logic/Rule.java
jmsfltchr: Also, why can't we combine the when and then into a single conjunction and do type inference over that?

jmsfltchr (changes requested):

flyingsilverfin:
Thoughts on creating a single query for the validation:
The problem statement is: "Check that there does not exist any when permutation that is in not in the then permutation"
Which is the same as: Make sure there are none of "Any when permutation that is not in the then permutation"

This is equivalent to check the set of type permutations of the _when_ clause is a subset of the type permutations of the _then_.

We could implement this as a negation, which is pretty much what we're doing with our two-step validation of generating the set of `then` first, then validating each `when` against it. We would expect the efficiency to be similar to the current approach.

jmsfltchr (approved):

PR #6522, opened 2022-02-17
Make partition manager class extendable

lolski:

The partition manager is open for extension from TypeDB Cluster. Additionally, the partitions are declared with the appropriate encodings.

1. Make partition manager class extendable:
    - Relax visibility modifiers of the column family handles, allowing subclasses to override them
    - Rename the class from `RocksPartitionManager` to `CorePartitionManager`
2. Improve the extendability of the database class by adding partition manager factory methods that can be overridden by a subclass.
3. Declare partitions with proper encodings rather than enum names.

### DIFF
--- a/database/CoreDatabase.java
+++ b/database/CoreDatabase.java
@@ -102,8 +102,8 @@ public class CoreDatabase implements TypeDB.Database {
     private final Factory.Session sessionFactory;
     protected OptimisticTransactionDB rocksSchema;
     protected OptimisticTransactionDB rocksData;
-    protected RocksPartitionManager.Schema rocksSchemaPartitionMgr;
-    protected RocksPartitionManager.Data rocksDataPartitionMgr;
+    protected CorePartitionManager.Schema rocksSchemaPartitionMgr;
+    protected CorePartitionManager.Data rocksDataPartitionMgr;
     protected CoreSession.Data statisticsBackgroundCounterSession;
     protected StatisticsBackgroundCounter statisticsBackgroundCounter;
     protected ScheduledExecutorService scheduledPropertiesLogger;
@@ -160,7 +160,7 @@ public class CoreDatabase implements TypeDB.Database {
 
     private void openSchema() {
         try {
-            List<ColumnFamilyDescriptor> schemaDescriptors = RocksPartitionManager.Schema.descriptors(rocksConfiguration.schema());
+            List<ColumnFamilyDescriptor> schemaDescriptors = CorePartitionManager.Schema.descriptors(rocksConfiguration.schema());
             List<ColumnFamilyHandle> schemaHandles = new ArrayList<>();
             rocksSchema = OptimisticTransactionDB.open(
                     rocksConfiguration.schema().dbOptions(),
@@ -168,15 +168,19 @@ public class CoreDatabase implements TypeDB.Database {
                     schemaDescriptors,
                     schemaHandles
             );
-            rocksSchemaPartitionMgr = new RocksPartitionManager.Schema(schemaDescriptors, schemaHandles);
+            rocksSchemaPartitionMgr = partitionManagerSchema(schemaDescriptors, schemaHandles);
         } catch (RocksDBException e) {
             throw TypeDBException.of(e);
         }
     }
 
+    protected CorePartitionManager.Schema partitionManagerSchema(List<ColumnFamilyDescriptor> schemaDescriptors, List<ColumnFamilyHandle> schemaHandles) {
+        return new CorePartitionManager.Schema(schemaDescriptors, schemaHandles);
+    }
+
     private void openAndInitialiseData() {
         try {
-            List<ColumnFamilyDescriptor> dataDescriptors = RocksPartitionManager.Data.descriptors(rocksConfiguration.data());
+            List<ColumnFamilyDescriptor> dataDescriptors = CorePartitionManager.Data.descriptors(rocksConfiguration.data());
             List<ColumnFamilyHandle> dataHandles = new ArrayList<>();
             rocksData = OptimisticTransactionDB.open(
                     rocksConfiguration.data().dbOptions(),
@@ -186,11 +190,15 @@ public class CoreDatabase implements TypeDB.Database {
             );
             assert dataHandles.size() == 1;
             dataHandles.addAll(rocksData.createColumnFamilies(dataDescriptors.subList(1, dataDescriptors.size())));
-            rocksDataPartitionMgr = new RocksPartitionManager.Data(dataDescriptors, dataHandles);
+            rocksDataPartitionMgr = partitionManagerData(dataDescriptors, dataHandles);
         } catch (RocksDBException e) {
             throw TypeDBException.of(e);
         }
         mayInitRocksDataLogger();
+    }
+
+    protected CorePartitionManager.Data partitionManagerData(List<ColumnFamilyDescriptor> dataDescriptors, List<ColumnFamilyHandle> dataHandles) {
+        return new CorePartitionManager.Data(dataDescriptors, dataHandles);
     }
 
     protected void load() {
@@ -208,7 +216,7 @@ public class CoreDatabase implements TypeDB.Database {
 
     private void openData() {
         try {
-            List<ColumnFamilyDescriptor> dataDescriptors = RocksPartitionManager.Data.descriptors(rocksConfiguration.data());
+            List<ColumnFamilyDescriptor> dataDescriptors = CorePartitionManager.Data.descriptors(rocksConfiguration.data());
             List<ColumnFamilyHandle> dataHandles = new ArrayList<>();
             rocksData = OptimisticTransactionDB.open(
                     rocksConfiguration.data().dbOptions(),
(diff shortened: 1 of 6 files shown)


graph/common/Storage.java
```diff
@@ -109,6 +113,15 @@ default Schema asSchema() {
...
             private final Integer fixedStartBytes;
 
+            public static Partition fromName(String name) {
+                if (name.equals(DEFAULT.name())) return DEFAULT;
+                else if (name.equals(VARIABLE_START_EDGE.name())) return VARIABLE_START_EDGE;
+                else if (name.equals(FIXED_START_EDGE.name())) return FIXED_START_EDGE;
+                else if (name.equals(OPTIMISATION_EDGE.name())) return OPTIMISATION_EDGE;
+                else if (name.equals(STATISTICS.name())) return STATISTICS;
+                else throw TypeDBException.of(UNRECOGNISED_VALUE);
+            }
```
flyingsilverfin, line 123: Use a int or even better a short to save space in the raft log and network

graph/common/Storage.java
```diff
@@ -27,11 +27,15 @@
 import com.vaticle.typedb.core.graph.iid.VertexIID;
 
 import javax.annotation.Nullable;
+import java.util.Arrays;
```
flyingsilverfin, line 30: unused import

flyingsilverfin (approved):
Approved with comments!

database/CorePartitionManager.java
```diff
@@ -42,12 +45,12 @@
 import static com.vaticle.typedb.core.graph.common.Storage.Key.Partition.STATISTICS;
 import static com.vaticle.typedb.core.graph.common.Storage.Key.Partition.VARIABLE_START_EDGE;
 
-abstract class RocksPartitionManager {
+public abstract class CorePartitionManager {
```
lolski, line 48: Relax the access modifier of the class, inner classes, methods, and fields. This is to allow classes from other packages to subclass and / or access them.

database/CoreDatabase.java
```diff
@@ -160,23 +160,27 @@ protected void initialise() {
...
                     schemaHandles
             );
-            rocksSchemaPartitionMgr = new RocksPartitionManager.Schema(schemaDescriptors, schemaHandles);
+            rocksSchemaPartitionMgr = partitionManagerSchema(schemaDescriptors, schemaHandles);
         } catch (RocksDBException e) {
             throw TypeDBException.of(e);
         }
     }
 
+    protected CorePartitionManager.Schema partitionManagerSchema(List<ColumnFamilyDescriptor> schemaDescriptors, List<ColumnFamilyHandle> schemaHandles) {
```
lolski, line 177: Create a factory method that can be overridden by the subclass, providing their own schema partition manager implementation.

graph/common/Storage.java
```diff
@@ -101,18 +102,41 @@ default Schema asSchema() {
...
+            VARIABLE_START_EDGE(Encoding.Partition.VARIABLE_START_EDGE, null),
+            FIXED_START_EDGE(Encoding.Partition.FIXED_START_EDGE, VertexIID.Thing.DEFAULT_LENGTH + InfixIID.Thing.DEFAULT_LENGTH + VertexIID.Thing.PREFIX_W_TYPE_LENGTH),
+            OPTIMISATION_EDGE(Encoding.Partition.OPTIMISATION_EDGE, VertexIID.Thing.DEFAULT_LENGTH + InfixIID.Thing.RolePlayer.LENGTH + VertexIID.Thing.PREFIX_W_TYPE_LENGTH),
+            STATISTICS(Encoding.Partition.STATISTICS, null);
 
+            private final short ID;
             private final Integer fixedStartBytes;
 
-            Partition(@Nullable Integer fixedStartBytes) {
+            public static Partition fromId(short ID) {
```
lolski, line 114: Return the appropriate partition enum, given a partition ID.

graph/common/Storage.java
```diff
@@ -101,18 +102,41 @@ default Schema asSchema() {
     interface Key extends Comparable<Key> {
 
         enum Partition {
-            DEFAULT(null),
-            VARIABLE_START_EDGE(null),
-            FIXED_START_EDGE(VertexIID.Thing.DEFAULT_LENGTH + InfixIID.Thing.DEFAULT_LENGTH + VertexIID.Thing.PREFIX_W_TYPE_LENGTH),
-            OPTIMISATION_EDGE(VertexIID.Thing.DEFAULT_LENGTH + InfixIID.Thing.RolePlayer.LENGTH + VertexIID.Thing.PREFIX_W_TYPE_LENGTH),
-            STATISTICS(null);
+            DEFAULT(Encoding.Partition.DEFAULT, null),
```
lolski, line 105: Construct partition enum with the appropriate ID.

graph/common/Encoding.java
```diff
@@ -52,6 +52,14 @@
     public static final String ROCKS_SCHEMA = "schema";
     public static final int ENCODING_VERSION = 1;
 
+    public interface Partition {
+        short DEFAULT = 0;
+        short VARIABLE_START_EDGE = 1;
+        short FIXED_START_EDGE = 2;
+        short OPTIMISATION_EDGE = 3;
+        short STATISTICS = 4;
+    }
```
flyingsilverfin, line 61: I think looking at this we should just put the Partition Enum here, and encode three things: `name`, `shortID`, and `fixedLengthStart`.
We should then make a TODO that says the `shortID` should eventually replace the `name` if/when we break backward compatibility again

flyingsilverfin (approved):

database/CorePartitionManager.java
```diff
@@ -127,26 +126,26 @@ ColumnFamilyHandle get(Key.Partition partition) {
                     configuration.defaultCFOptions()
             );
             descriptors[VARIABLE_START_EDGE_HANDLE_INDEX] = new ColumnFamilyDescriptor(
-                    ByteArray.encodeString(VARIABLE_START_EDGE.name(), STRING_ENCODING).getBytes(),
+                    VARIABLE_START_EDGE.encoding().partitionName().get().getBytes(),
```
lolski, line 129: Declare column families with partition name defined in the `Encoding` class instead using the `name()` built-in function. Using the latter is unsafe - if we refactor the enum declaration, the value returned by `name()` will change accordingly.

graph/common/Storage.java
```diff
@@ -101,18 +102,40 @@ default Schema asSchema() {
...
+            VARIABLE_START_EDGE(Encoding.Partition.VARIABLE_START_EDGE, null),
+            FIXED_START_EDGE(Encoding.Partition.FIXED_START_EDGE, VertexIID.Thing.DEFAULT_LENGTH + InfixIID.Thing.DEFAULT_LENGTH + VertexIID.Thing.PREFIX_W_TYPE_LENGTH),
+            OPTIMISATION_EDGE(Encoding.Partition.OPTIMISATION_EDGE, VertexIID.Thing.DEFAULT_LENGTH + InfixIID.Thing.RolePlayer.LENGTH + VertexIID.Thing.PREFIX_W_TYPE_LENGTH),
+            STATISTICS(Encoding.Partition.STATISTICS, null);
 
+            private final Encoding.Partition encoding;
             private final Integer fixedStartBytes;
 
-            Partition(@Nullable Integer fixedStartBytes) {
+            public static Partition fromID(short ID) {
```
lolski, line 114: Return the appropriate enum given the enum ID.

graph/common/Encoding.java
```diff
@@ -52,6 +54,30 @@
     public static final String ROCKS_SCHEMA = "schema";
     public static final int ENCODING_VERSION = 1;
 
+    public enum Partition {
```
lolski, line 57: Define partition encodings.

graph/common/Encoding.java
```diff
@@ -52,6 +54,30 @@
...
+        DEFAULT((short) 0, null),
```
lolski, line 58: The encoding for the "default" partition does not need a partition name.

graph/common/Encoding.java
```diff
@@ -52,6 +54,30 @@
...
+        VARIABLE_START_EDGE((short) 1, ByteArray.encodeString("VARIABLE_START_EDGE", STRING_ENCODING)),
```
lolski, line 59: Define the encoding for the other partitions, with the appropriate ID and partition name. A partition name is a String that is converted into `ByteArray` using `STRING_ENCODING`.

graph/common/Encoding.java
lolski: Me and @flyingsilverfin had a verbal discussion and agreed to keep the "partition key" enum in the `Storage` class and the "partition encoding" enum in the `Encoding` class. They will be unified when we refactor the `Encoding` class in the future.

graph/common/Storage.java
lolski: This has been fixed as per the last commit.

graph/common/Storage.java
lolski: This has been solved as per the last commit.

graph/common/Encoding.java
lolski: The TODO has been added as per the last commit.

dependencies/vaticle/repositories.bzl
```diff
@@ -35,7 +35,7 @@ def vaticle_typedb_common():
     git_repository(
         name = "vaticle_typedb_common",
         remote = "https://github.com/vaticle/typedb-common",
-        commit = "aa549ae3a3337e6b6eba7fc00142ea646faef143", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_common
+        commit = "035ea2f8b0342ec322e0472afa19be5279557f75", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_common
```
lolski, line 38: Update `typedb-common` to the latest version

PR #6524, opened 2022-02-21
Traversal engine 3.0 groundwork

flyingsilverfin:

Implement the foreseeable groundwork for the next-generation traversal engine. This work is mostly centered on pushing `Forwardable` iterators throughout the code base, and especially in the graph and graph traversal engine. Through this process we uncover some optimisations and simplifications in especially the lower level graph and storage components of the system. The impact of the optimisations appears to be a 10-20% improvement for write speeds (reads not measured).

`//common`
* Improve error messages for rule validation and type inference
* Split `Sorted` iterators into their own package `//common/iterator/sorted`
* Sorted iterators now carry a generic for Ascending or Descending order. In general, we only use Ascending, but in the upcoming Traversal engine we will use Descending when the user does a descending sort on a starting vertex!
* Tweak `forward()` method on all `Forwardable` iterators to be fully lazy, deferring computing `hasNext()` until `hasNext()` is actually called (previously, calling `forward()` could sometimes call `hasNext()` eagerly)
* Created missing `Sorted`/`Forwardable` versions of `FunctionalIterators` we required such as `Limited`, `Finalise` etc.
* Fix a race condition in `AsyncProducer` between `recycle()` and `job()` using a read-write lock. This class is meant to be thread-safe.

`//concept`
* `Thing` and `Type` concepts are `Comparable` to themselves. This enables returning `Forwardable` iterators from the Concept API calls, which in turns enables a few optimisations using `forward()`, such as during data validation on insert
* Upgrade `Type` and `Thing` concepts to utilise `Forwardable` iterators internall where possible, enabling some optimisations. This constitutes a large part of the change in the `//concept `package
* Upgrade `Type` concepts to utilise the Type graph caches to compute things like attribute ownerships, roles played etc. This makes data validation at insert time much faster. Appears to **improve write speed by 10-20%**

`//graph`
* Apply an intended rename from `In/Out` to `Forward/Backward` in the low-level Edge IID encodings. Now, we have `Forward` and `Backward` infixes for edges. a `Adjacency` however is still either an `In` or `Out` adjacency, because they only contains edges pointing In or Out
* Make directed `Adjacency` subtypes: we now have `ThingAdjacency.In/Out` and `TypeAdjacency.In/Out`. These are split in order to allow having the `from()` vertices return a `Forwardable` iterator in the `In` case, and the `to()` vertices return a `Forwardable` Iterator in the `Out` case
* `In` and `Out` adjacencies return `InEdgeIterator` or `OutEdgeIterator`, which contain logic for mapping a `Forwardable` edge iterator to a `Forwardable` iterator of `from()` or `to()` vertices, or tuples of `from/to + optimised` vertices
* Introduce `ThingEdge.View` and `TypeEdge.View`, which view a logical edge as either the forward or backward directed edge. To match this we rename `EdgeIID` to `EdgeViewIID`
* Extend the `Type` graph caches to cache not just attribute ownerships, but also roles played, owners of attributes, subtypes etc. This cache is used from three places now: 1) `Type` concepts 2) Query planning 3) Procedure execution. It's both more efficient to re-use the same cache for all 3, plus more likely to make errors appear. Before, the cache was only used in the query planning, which is hard to debug to see if the sets of attribute ownership were actually correct!

`//pattern`
* `Relation.RolePlayer` constraints now carry a set of `inferredTypes`, which represent the inferred types of the role _instances_. This will be populated regardless of whether there is an explicitly written role type or not
* Rename `Conjunction.Cloner` to `Conjunction.ConstraintCloner` since it is only used to clone parts of conjunctions rather than entire conjunctions

`//logic`
* improved rule insertability validation and error messages
* Extensively refactored Type inference to have a simpler logical flow and entry point, and incorporate type inference for role instances that are set into the new `Relation.RolePlayer.inferredTypes` field

`//traversal`
* Massively simplify `Planner` classes because we can now assume type inference has executed and provided types for any `Thing` vertices or edges
* `Procedure` edges and vertices now almost exclusively use `Forwadable` iterators. This enabled some optimisations (such as closures being computed with a `forward()` and check). However, this is fundamentally the main change that will allow us to do fast intersections in the next-generation traversal engine!
* `Procedure` edges/verticesWe also leverage the `Type` graph cache to traversal the schema graph much more heavily.
* `Procedure` edges and vertices are simplified a bit because can assume type inference is always executed
* `GraphIterator` holds onto `Forwardable` iterators but doesn't leverage them yet

`//database`
* `RocksIterator` has both `Ascending or `Descending` subtypes


graph/vertex/TypeVertex.java
```diff
@@ -27,7 +27,7 @@
 
 import java.util.regex.Pattern;
 
-public interface TypeVertex extends Vertex<VertexIID.Type, Encoding.Vertex.Type> {
+public interface TypeVertex extends Vertex<VertexIID.Type, Encoding.Vertex.Type>, Comparable<TypeVertex> {
```
flyingsilverfin, line 30: `TypeVertex` is now Comparable to other Type vertices!

graph/vertex/TypeVertex.java
```diff
@@ -53,9 +53,9 @@
 
     void scope(String scope);
 
-    TypeAdjacency outs();
+    TypeAdjacency.Out outs();
 
-    TypeAdjacency ins();
+    TypeAdjacency.In ins();
```
flyingsilverfin, line 58: more strongly type the Adjacencies since different directions have different methods exposed

PR #6527, opened 2022-02-28
Faster and more space-efficient database statistics

flyingsilverfin:

We implement new statistics tracking by always counting attributes and has edges the same way as entities/relations/roles. Doing so may cause some miscounts when committing concurrently, since an attribute is store uniquely - therefore we correct any miscounts in a lightweight background job. In practice, identical concurrent attribute/has writes are generally uncommon, so the new mechanism rarely does any work. 

This **reduces disk space requirements by 30% and increases write speed by 10-20% while correcting statistics drift significantly more quickly**.

**NOTE: this change breaks backward compatibility. Use the export/import feature to migrate easily.**

**High level implementation**
* `ThingGraph` only records concept counts as if there were no concurrency or miscounting due to attribute deduplication
* At commit time, the `CoreTransaction` utilises the `IsolationManager` to obtain concurrently committing transactions.  These are used by the `StatisticsCorrector` to compute potentially over/under-counted statistics by the transaction. This information is recorded as "Metadata" about the data in a new column family, and used to correct the miscounts in a background task.
* The `StatisticsCorrector` is notified after each transaction commit to correct overcounts and undercounts in the attribute and has edge counts using a single threaded background executor

**Example**:
1. Transaction `A` commits `name: John`, and transaction `B` also commits `name: John`  concurrently. The database count for `name` will temporarily be `2` after both commit, even though only 1 `name: John` exists.
2. At commit time, our transactions are quickly validated serially against concurrent transactions. Let's say `A` proceeds first, at which time `B` is not yet committing. Therefore, `A` records no commit-time metadata about miscounts for `name`. However, when `B` enters the serial block, we detect that `A` has concurrently started committing, and calculate that both `A` and `B` will both have counted `name: John`. Thus, `B` will record as part of its commit metadata saying `name: John` may have been overcounted by transaction `B` if transaction `A` successfully commits.
3. Post-commit of `A` and `B`, the count for `name` is `2`. The background `StatisticsCorrector` can spin up a task to confirm that `A` committed, which means that the overcount of `name: John` by `B` is confirmed, so we need to decrement `name` by 1. 

**Low level changes**

Replacing the `Statistics` column family with `Metadata` column family
* Since we don't need a very large column family to store statistics counting data, we can generalise the old `Statistics` column family to be a new `Metadata` column family that is more general-purpose.
* `Statistics` information, such as counts and count correction metadata, is stored in the `Metadata` column family
* `Metadata` colum family will also be used by Cluster to store non user-inserted data

Refactors caused by introducing correction mechanism into the commit time operations:
* Rename `ConsistencyManager` to `IsolationManager`, since enforcing non-conflicting isolation between transactions is just one part of guaranteeing consistency
* Since the transaction commits now need to get the other concurrently commiting transactions, the `IsolationManager` now deals with `CoreTransaction.Data` instead of `RocksStorage.Data`
* Rewrite `IsolationManager` to be simpler and more efficient, hopefully eliminating any potential dangling isolation bugs
* We retain `CoreTransaction.Data` after it is `closed` or `committed` - the last operation to call is `delete()`. This is managed by the `IsolationManager`, which knows when it is a safe to completely delete the contents of a transaction (this is when no further transactions may need to check for conflicts against that transaction)

Other refactors:
* Introduce a new pattern of calling `closed(this)` or `committed(this)` on higher-level objects that need to know when their children are closing or comitting, rather than using `remove(this)` - `remove()` was too restrictive in terms of the semantics what the method could do
* Completely `clear()` the `ThingGraph`, which results in a more stable and lower memory profile
* Simplify data structures to do with key conflict detection and inefficient parts of the TypeInference traversal builder
* Correctly shutdown the import job executor when completed

Other new features:
* Implement an `intersection` iterator for `Forwardable` iterators. This was ultimately used not used in this PR, but we keep it for the implementation of the next-generation traversal engine


graph/edge/impl/ThingEdgeImpl.java
```diff
@@ -251,13 +251,11 @@ public void delete() {
...
                 to.ins().remove(this);
-                if (!(from.status().equals(BUFFERED)) && !(to.status().equals(BUFFERED))) {
+                if (from.status().equals(PERSISTED) && to.status().equals(PERSISTED)) {
                     graph.storage().deleteTracked(forward.iid());
                     graph.storage().deleteUntracked(backward.iid());
                 }
-                if (encoding == Encoding.Edge.Thing.Base.HAS && !isInferred) {
-                    graph.stats().hasEdgeDeleted(from.iid(), to.iid().asAttribute());
-                }
+                graph.edgeDeleted(this);
```
flyingsilverfin, line 258: notify edge deleted paradigm

graph/edge/impl/ThingEdgeImpl.java
```diff
@@ -525,9 +523,7 @@ public void delete() {
                 graph.convertToWritable(toIID).setModified();
                 graph.storage().deleteTracked(forward.iid());
                 graph.storage().deleteUntracked(backward.iid());
-                if (encoding == Encoding.Edge.Thing.Base.HAS && !isInferred) {
-                    graph.stats().hasEdgeDeleted(fromIID, toIID.asAttribute());
-                }
+                graph.edgeDeleted(this);
```
flyingsilverfin, line 526: notify edge deleted paradigm

database/CoreDatabase.java
```diff
@@ -437,7 +444,15 @@ private void shutdownRocksPropertiesLogger() {
...
-        statisticsBgCounterStop();
+        statisticsCorrector.close();
+        Optional<CompletableFuture<Void>> correction = statisticsCorrector.runningCorrection();
+        if (correction.isPresent()) {
+            try {
+                correction.get().get(Executors.SHUTDOWN_TIMEOUT_MS, MILLISECONDS);
+            } catch (InterruptedException | ExecutionException | TimeoutException e) {
+                LOG.warn("Statistics corrector was not fully shutdown");
+            }
+        }
```
flyingsilverfin, line 455: we try to safely shut down the corrector before killing RocksDB otherwise we can cause assertion errors in RocksDB

traversal/scanner/CombinationFinder.java
```diff
@@ -72,7 +72,6 @@ public CombinationFinder(GraphManager graphMgr, Set<CombinationProcedure> proced
 
         for (CombinationProcedure procedure : procedures) {
             start(procedure);
-            if (procedure.vertices().size() == 1) continue;
```
flyingsilverfin, line 75: this is a bug that we discovered by simplifying the `TypeInference` class and was then revelead by a `TypeInferenceTest` which included `$a has $a`

database/CoreDatabase.java
```diff
@@ -456,203 +471,308 @@ public void delete() {
...
+
+        private void recordMiscounts(CoreTransaction.Data txn, Set<CoreTransaction.Data> concurrentTxn) {
+            Map<AttributeVertex.Write<?>, Set<Long>> attrOvercountDependencies = new HashMap<>();
+            Map<AttributeVertex.Write<?>, Set<Long>> attrUndercountDependencies = new HashMap<>();
+            Map<Pair<ThingVertex.Write, AttributeVertex.Write<?>>, Set<Long>> hasOvercountDependencies = new HashMap<>();
+            Map<Pair<ThingVertex.Write, AttributeVertex.Write<?>>, Set<Long>> hasUndercountDependencies = new HashMap<>();
+            for (CoreTransaction.Data concurrent : concurrentTxn) {
+                // note: fail-fast if checks are much faster than using empty iterators (probably due to concurrent data structures)
+                if (!txn.graphMgr.data().attributesCreated().isEmpty() && !concurrent.graphMgr.data().attributesCreated().isEmpty()) {
```
flyingsilverfin, line 731: we can extract c-style functions for pairs of these

graph/common/Encoding.java
```diff
@@ -935,97 +928,57 @@ public ByteArray bytes() {
...
-                throw TypeDBException.of(UNRECOGNISED_VALUE);
-            }
+            enum Prefix {
 
-            public byte key() {
-                return key;
-            }
+                SNAPSHOT(0),
+                VERTEX_COUNT(1),
+                HAS_TYPE_EDGE_COUNT(2),
```
flyingsilverfin, line 939: name `HAS_EDGE` only
and move counts to 10,11 and MISCOUNTABLE to 50 -> 60

migrator/data/DataImporter.java
```diff
@@ -226,6 +233,8 @@ void run() {
                 commitBatch();
             } catch (InterruptedException e) {
                 e.printStackTrace();
+            } catch (Exception e) {
+                e.printStackTrace();
```
flyingsilverfin, line 237: use a single catch for the same conditions

graph/vertex/impl/AttributeVertexImpl.java
```diff
@@ -74,7 +76,12 @@
 
         @Override
         public Encoding.Status status() {
-            return Encoding.Status.IMMUTABLE;
+            return isPersisted() ? PERSISTED : BUFFERED;
+        }
```
flyingsilverfin, line 80: interesting new change: because we do a true PUT on attributes, we actually always know whether an attribute is persisted or buffered - this aligns them with all other concept instances, which are also either buffered or persisted!

common/iterator/sorted/SortedIterators.java
```diff
@@ -116,6 +116,15 @@
             return new MergeMappedIterator.Forwardable<>(iterators, e -> e, order);
         }
 
+        @SafeVarargs
```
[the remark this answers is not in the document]
flyingsilverfin: SafeVarargs is used to indicate to the Java compiler that a signature like this is okay:
```
<T> T[] toArray(T... elements) { }
```
this is a a warning from Java because arrays cannot be generified.

common/iterator/AbstractFunctionalIterator.java
```diff
@@ -220,6 +221,17 @@ public long count() {
         return count;
     }
 
+    @Override
+    public <U> Optional<U> reduce(U initialAccumulator, BiFunction<T, U, U> accumulate) {
```
flyingsilverfin, line 225: we introduce `reduce()` which can be used to implement list aggregations or summations generically

lolski (approved):

flyingsilverfin:
PR Review Summary:
1. Optimise PR Goals/Changes/Comments for the reviewer
We should aim to make the PR description maximally user- and reviewer-friendly. This includes splitting logical separations of the goals. In a bigger PR, it is also useful to create a story (with examples if required) of the individual changes to tie them all together from high to low level. 

In comments, we should try to comment each file with the high level changes to the file, plus the important individual changes. This helps the reader understand how different files are related to the goal of the PR.

2. More diligent self-reviews
In long PRs it is more important than in short ones to do thorough self-reviews before submitting to the final review. This should certainly catch most simple changes, such as TODOs, printlines, missing test cases, high level story, etc.

3. Naming and structure within the `StatisticsCorrector`
It took Ganesh and I some time to get to the final naming around the `StatisticsCorrector`. We originally had `StatisticsCompensator`, which had overcounts and undercounts. This however meant that we had `compensate()` as the main method, but it led to an unclear storyline when correcting a single miscount versus many. 

We also managed to simplify the high-level story by modularising the correction mechanism, which used to all run in the main `compensate()` method. Now, we split out the correction mechanism into a better high-level flow by modularising the single miscount correction away from the correction loop, and choosing names that are symmetric throughout the entry method.

4. Transaction lifetime and interaction with IsolationManager and StatisticsCorrector

We had to introduce another transaction state after close, in which the transaction is deleted. This means transactions are opened, committed/closed, and deleted (the last state is not exposed to the users). 

Before discussion with Ganesh the cleanup of the transaction's state within the `IsolationManager` and the `StatisticsCorrector` happened without the Transaction's knowledge. However, this requires the `IsolationManager` to communicate directly with other the `StatisticsCorrector` to notify it to clean up resources to do with the transactions it is also cleaning up. In this model, the interactions are very complex: `Transaction <-> IsolationManager. Transaction <-> StatisticsCorrector. StatisticsCorrector <-> IsolationManager.`

Now, we remove the need for the `IsolationManager` to know about the `StatisticsCorrector`, and centralise control inside of the Transaction. Ideally, we want the transaction to behave as if it is the only transaction running in the system.

The transaction at commit time it interacts with the `IsolationManager`, validating its commit start. Using information from the `IsolationManager` the transaction commit manages the interaction with the `StatisticsCorrector`. The goal is to centralise all transaction lifetime and interaction with longer-lived components __inside__ of the transaction. 

This control flow dictated by the Transaction itself is currently broken in one place: transactions do not delete themselves, they are notified by the `IsolationManager` to delete themselves. We then resume the expected Transaction-oriented control where the transaction reports to relevant other components (eg. `StatisticsCorrector`) that it is being deleted.

_Alternative_
One separate model could be that the `IsolationManager` becomes a `ConcurrentTransactionManager` (or even `TransactionManager`, and we delete sessions(?)), and encomposses isolation and statistics correction -- both of these require information about transactions that live beyond the user's transaction `close()`.

Given this, one design could be that this manager dictates all aspects of the Transaction's lifetime (so control flow is purely from the manager, to the transaction). This was not explored as it is too large of a change for now.

Another alternative is that a transaction's `close()` is the end of that object, and it copies out required state into the the `ConcurrentTransactionManager`, which handles all post-commit information flow including to do with isolation and statistics correction. This model was explored without the `ConcurrentTransactionManager` but required too much transaction state to be moved from the closed transaction to another object to be considered elegant.

flyingsilverfin:

We implement new statistics tracking by always counting attributes and has edges the same way as entities/relations/roles. Doing so may cause some miscounts when committing concurrently, since an attribute is store uniquely - therefore we correct any miscounts in a lightweight background job. In practice, identical concurrent attribute/has writes are generally uncommon, so the new mechanism rarely does any work. 

This **reduces disk space requirements by 30% and increases write speed by 10-20% while correcting statistics drift significantly more quickly**.

**NOTE: this change breaks backward compatibility. Use the export/import feature to migrate easily.**

**High level implementation**
* `ThingGraph` only records concept counts as if there were no concurrency or miscounting due to attribute deduplication
* At commit time, the `CoreTransaction` utilises the `IsolationManager` to obtain concurrently committing transactions.  These are used by the `StatisticsCorrector` to compute potentially over/under-counted statistics by the transaction. This information is recorded as "Metadata" about the data in a new column family, and used to correct the miscounts in a background task.
* The `StatisticsCorrector` is notified after each transaction commit to correct overcounts and undercounts in the attribute and has edge counts using a single threaded background executor

**Example**:
1. Transaction `A` commits `name: John`, and transaction `B` also commits `name: John`  concurrently. The database count for `name` will temporarily be `2` after both commit, even though only 1 `name: John` exists.
2. At commit time, our transactions are quickly validated serially against concurrent transactions. Let's say `A` proceeds first, at which time `B` is not yet committing. Therefore, `A` records no commit-time metadata about miscounts for `name`. However, when `B` enters the serial block, we detect that `A` has concurrently started committing, and calculate that both `A` and `B` will both have counted `name: John`. Thus, `B` will record as part of its commit metadata saying `name: John` may have been overcounted by transaction `B` if transaction `A` successfully commits.
3. Post-commit of `A` and `B`, the count for `name` is `2`. The background `StatisticsCorrector` can spin up a task to confirm that `A` committed, which means that the overcount of `name: John` by `B` is confirmed, so we need to decrement `name` by 1. 


**Low level changes**

Replacing the `Statistics` column family with `Metadata` column family
* Since we don't need a very large column family to store statistics counting data, we can generalise the old `Statistics` column family to be a new `Metadata` column family that is more general-purpose.
* `Statistics` information, such as counts and count correction metadata, is stored in the `Metadata` column family
* `Metadata` colum family will also be used by Cluster to store non user-inserted data

Refactors caused by introducing correction mechanism into the commit time operations:
* Rename `ConsistencyManager` to `IsolationManager`, since enforcing non-conflicting isolation between transactions is just one part of guaranteeing consistency
* Since the transaction commits now need to get the other concurrently commiting transactions, the `IsolationManager` now deals with `CoreTransaction.Data` instead of `RocksStorage.Data`
* Rewrite `IsolationManager` to be simpler and more efficient, hopefully eliminating any potential dangling isolation bugs
* We retain `CoreTransaction.Data` after it is `closed` or `committed` - the last operation to call is `delete()`. This is managed by the `IsolationManager`, which knows when it is a safe to completely delete the contents of a transaction (this is when no further transactions may need to check for conflicts against that transaction)

Other refactors:
* Introduce a new pattern of calling `closed(this)` or `committed(this)` on higher-level objects that need to know when their children are closing or comitting, rather than using `remove(this)` - `remove()` was too restrictive in terms of the semantics what the method could do
* Completely `clear()` the `ThingGraph`, which results in a more stable and lower memory profile
* Simplify data structures to do with key conflict detection and inefficient parts of the TypeInference traversal builder
* Correctly shutdown the import job executor when completed

Other new features:
* Implement an `intersection` iterator for `Forwardable` iterators. This was ultimately used not used in this PR, but we keep it for the implementation of the next-generation traversal engine

PR #6528, opened 2022-03-11
Data importer can handle an arbitary number of relations

flyingsilverfin:

We make the built in TypeDB Importer (`typedb server import`) scale for a large number of relations by removing in-memory data structures used to track imported concepts. Instead, the data is persisted into the preexisting temporary RocksDB instance, ensuring the import will scale to a huge number of concepts to import. 

* Fix import/export logging verbosity (!netty!)
* Fix import/export error propagation to the client side, which allows the user to see a useful error when the import fails for any reason
* No longer use in-memory data structures during relation import to track partial relations. Instead, we rely on the temporary RocksDB instance to mark partial relations. We then iteratively load the partial relations until they are completed.
* Introduce a very small encoding scheme used in the temporary rocksdb storage to be able to store completeness information along with ID mappings

### DIFF
--- a/common/collection/ByteArray.java
+++ b/common/collection/ByteArray.java
@@ -53,7 +53,7 @@ public abstract class ByteArray implements Comparable<ByteArray> {
         this.array = array;
     }
 
-    public static ByteArray.Base of(byte[] array) {
+    public static ByteArray of(byte[] array) {
         return new ByteArray.Base(array);
     }
 
@@ -128,13 +128,13 @@ public abstract class ByteArray implements Comparable<ByteArray> {
         return Base64.getEncoder().encodeToString(getBytes());
     }
 
-    public static ByteArray.Base encodeString(String string, Charset encoding) {
+    public static ByteArray encodeString(String string, Charset encoding) {
         return of(string.getBytes(encoding));
     }
 
     public abstract String decodeString(Charset encoding);
 
-    public static ByteArray.Base encodeUnsignedShort(int num) {
+    public static ByteArray encodeUnsignedShort(int num) {
         byte[] bytes = new byte[SHORT_SIZE];
         bytes[1] = (byte) (num);
         bytes[0] = (byte) (num >> 8);
@@ -155,7 +155,7 @@ public abstract class ByteArray implements Comparable<ByteArray> {
 
     public abstract int decodeInt();
 
-    public static ByteArray.Base encodeUUID(UUID uuid) {
+    public static ByteArray encodeUUID(UUID uuid) {
         ByteBuffer buffer = ByteBuffer.wrap(new byte[16]);
         buffer.putLong(uuid.getMostSignificantBits());
         buffer.putLong(uuid.getLeastSignificantBits());
--- a/concept/thing/Relation.java
+++ b/concept/thing/Relation.java
@@ -19,6 +19,9 @@
 package com.vaticle.typedb.core.concept.thing;
 
 import com.vaticle.typedb.core.common.iterator.FunctionalIterator;
+import com.vaticle.typedb.core.common.iterator.sorted.SortedIterator;
+import com.vaticle.typedb.core.common.iterator.sorted.SortedIterator.Forwardable;
+import com.vaticle.typedb.core.common.iterator.sorted.SortedIterator.Order;
 import com.vaticle.typedb.core.concept.type.RelationType;
 import com.vaticle.typedb.core.concept.type.RoleType;
 
@@ -36,12 +39,18 @@ public interface Relation extends Thing {
 
     void removePlayer(RoleType roleType, Thing player);
 
-    FunctionalIterator<? extends Thing> getPlayers(String roleType, String... roleTypes);
+    FunctionalIterator<Thing> getPlayers();
 
-    FunctionalIterator<? extends Thing> getPlayers(RoleType... roleTypes);
+    FunctionalIterator<Thing> getPlayers(String[] roleTypes);
+
+    FunctionalIterator<Thing> getPlayers(RoleType[] roleTypes);
+
+    Forwardable<Thing, Order.Asc> getPlayers(String roleType, String... roleTypes);
+
+    Forwardable<Thing, Order.Asc> getPlayers(RoleType roleType, RoleType... roleTypes);
 
     // TODO: This method should just return FunctionalIterator<Pair<RoleType, Thing>>
-    Map<? extends RoleType, ? extends List<? extends Thing>> getPlayersByRoleType();
+    Map<? extends RoleType, List<Thing>> getPlayersByRoleType();
 
     FunctionalIterator<? extends RoleType> getRelating();
 }
--- a/concept/thing/Relation.java
+++ b/concept/thing/Relation.java
@@ -38,9 +38,7 @@ public interface Relation extends Thing {
 
     void removePlayer(RoleType roleType, Thing player);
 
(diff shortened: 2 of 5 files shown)


common/collection/ByteArray.java
```diff
@@ -155,7 +155,7 @@ public static ByteArray encodeInt(int num) {
 
     public abstract int decodeInt();
 
-    public static ByteArray.Base encodeUUID(UUID uuid) {
```
flyingsilverfin, line 158: the static methods should not be returning the sub (ie. implementation-specific) type, only the more generic `ByteArray`

concept/thing/Relation.java
```diff
@@ -36,12 +39,18 @@
...
+    FunctionalIterator<Thing> getPlayers();
 
-    FunctionalIterator<? extends Thing> getPlayers(RoleType... roleTypes);
+    FunctionalIterator<Thing> getPlayers(String[] roleTypes);
+
+    FunctionalIterator<Thing> getPlayers(RoleType[] roleTypes);
+
+    Forwardable<Thing, Order.Asc> getPlayers(String roleType, String... roleTypes);
+
+    Forwardable<Thing, Order.Asc> getPlayers(RoleType roleType, RoleType... roleTypes);
```
flyingsilverfin, line 50: we make the `Relation` API more expressive, splitting `getPlayers` into variants that return `forwardable` and regular iterators.

Ideally, we would have two variants that look like this:
```
Functionaliterator<> getPlayers(String roleType...)
```
and
```
Forwardable<> getPlayers(String roleType, String... roleTypes);
```

which effectively switch the return type when using 1 or more role types to a forwardable one. However, this is ambiguous and not allowed by java.

To allow passing a variable number of role types instead, we use an `String[]`, which is as useful as `String...` (they are interchangable, but the caller has to construct an array now).

migrator/data/DataImporter.java
```diff
@@ -205,23 +203,27 @@ void execute() {
     private abstract class ImportWorker {
 
         private final BlockingQueue<DataProto.Item> items;
-        private final Map<ByteArray, String> bufferedToOriginalIds;
-        private final Map<String, ByteArray> originalToBufferedIds;
+        private final Map<ByteArray, String> bufferedToOriginalIDs;
+        private final Map<String, ByteArray> originalToBufferedIDs;
+        private final Set<String> incompleteIDs;
+        private final Set<String> completedIDs;
```
flyingsilverfin, line 209: track the incomplete and completed IDs in the batch being imported

migrator/data/DataImporter.java
```diff
@@ -239,50 +241,64 @@ void run() {
...
-            originalToBufferedIds.clear();
+            transaction.committedIIDs().forEachRemaining(pair -> {
+                String originalID = bufferedToOriginalIDs.get(pair.first());
+                conceptTracker.recordMapping(originalID, pair.second());
+                if (incompleteIDs.contains(originalID)) conceptTracker.recordIncomplete(originalID);
+            });
+            bufferedToOriginalIDs.clear();
+            originalToBufferedIDs.clear();
+            completedIDs.forEach(conceptTracker::deleteIncomplete);
+        }
```
flyingsilverfin, line 254: a commit  batch has to track completed and incomplete IDs, as well as ID mappings

migrator/data/DataImporter.java
```diff
@@ -361,97 +377,160 @@ private void insertEntity(TypeDB.Transaction transaction, DataProto.Item.Entity
...
             if (type == null) throw TypeDBException.of(TYPE_NOT_FOUND, msg.getLabel());
             Entity entity = type.create();
-            recordCreated(entity.getIID(), msg.getId());
+            recordMapping(entity.getIID(), msg.getId());
             status.entityCount.incrementAndGet();
         }
     }
 
-    private class CompleteRelations extends ImportWorker {
+    private class Relations extends ImportWorker {
```
flyingsilverfin, line 385: Relations always import incrementially - if there are any role players imported, we immediately create the relation, then mark it is incomplete if we are missing any role players. The relations will be completed iteratively

migrator/data/DataImporter.java
```diff
@@ -361,97 +377,160 @@ private void insertEntity(TypeDB.Transaction transaction, DataProto.Item.Entity
...
+                } else if (isIncompleteID(item.getRelation().getId())) {
+                    return tryExtend(importedRelation.asRelation(), item.getRelation());
                 }
-            } else return 0;
+            }
+            return 0;
         }
 
-        private Optional<Integer> tryInsertCompleteRelation(DataProto.Item.Relation relationMsg) {
+        private int tryImport(DataProto.Item.Relation relationMsg) {
```
flyingsilverfin, line 407: try to import a new relation with at least one role player

migrator/data/DataImporter.java
```diff
@@ -361,97 +377,160 @@ private void insertEntity(TypeDB.Transaction transaction, DataProto.Item.Entity
...
+                    }
+                    relation.addPlayer(roleType, player);
+                    rolesCreated++;
                 }
             }
-            return Optional.of(players);
+            return rolesCreated;
         }
+
+        private int tryExtend(Relation relation, DataProto.Item.Relation relationMsg) {
```
flyingsilverfin, line 443: try to extend an existing relation with the missing role players

migrator/data/DataImporter.java
```diff
@@ -361,97 +377,160 @@ private void insertEntity(TypeDB.Transaction transaction, DataProto.Item.Entity
...
+                    }
+                }
+            }
+            if (!stillIncomplete) recordCompletedID(relationMsg.getId());
+            status.roleCount.addAndGet(rolesCreated);
+            return rolesCreated;
+        }
+    }
+
+    private void importRelations() {
```
flyingsilverfin, line 463: top-level logic for loading relations

migrator/data/DataImporter.java
```diff
@@ -361,97 +377,160 @@ private void insertEntity(TypeDB.Transaction transaction, DataProto.Item.Entity
...
+        if (relationsUnfinished()) loadCyclicalRelations();
+        assert !conceptTracker.containsIncomplete();
     }
 
-    private void importSkippedRelations() {
+    private boolean relationsUnfinished() {
+        return skippedRelations.get() || conceptTracker.containsIncomplete();
+    }
+
+    private void loadCyclicalRelations() {
```
flyingsilverfin, line 480: we still have to handle the potential for relations that are pure cycles - eg. `r1 ---> r2` and `r2 ---> r1` at the same time - this requires we create the relation instances without any role players first, then attach the relation role players within that same transaction.

migrator/data/DataImporter.java
```diff
@@ -520,8 +600,57 @@ public ByteArray get(String originalID) {
...
+            } catch (RocksDBException e) {
+                throw TypeDBException.of(e);
+            }
+        }
+
+        private ByteArray encodeOriginalID(String originalID) {
+            return ByteArray.encodeString(originalID, STRING_ENCODING);
+        }
+
+        enum Prefix {
```
flyingsilverfin, line 646: we have two keys we need to store into our temporary RocksDB storage. We had to introduce this to replace all-in memory data structures, otherwise we could OOM on large imports!

flyingsilverfin:
Merging without review. Discussed with @lolski -- this code is likely to be rewritten anyway and has never been properly reviewed, and will need a rearchitect properly at some point in the future anyway.

Issue #6526, opened 2022-02-23
Declare database partitions with partition IDs


lolski:
## Problem to Solve

A database partition, backed by a RocksDB column family, uses "partition name" as its identifier.

The name is big in size and therefore a shorter identifier "partition ID" has been added, in order to keep TypeDB Cluster data replication efficient.

However, we can't yet remove the "partition name" in order to stay backwards compatible.

## Proposed Solution

Remove the legacy "partition name" and fully utilise "partition ID". But, do it with appropriate timing since it will require users to reload their data.

flyingsilverfin:
Implemented with #6529

PR #6529, opened 2022-03-18
Replace partition name with partition ID

flyingsilverfin:

Introduce a canonical identifier for each Partition (RocksDB Column Families), replacing the separate IDs used by Cluster (a `short`) and Core (a `string`). 

Implements #6526.

* Introduce a `byte` partition/colum family identifier, which replaces both previous identifiers of String and short type. This ID will be used in both Core and Cluster

### DIFF
--- a/database/CorePartitionManager.java
+++ b/database/CorePartitionManager.java
@@ -125,19 +125,19 @@ public abstract class CorePartitionManager {
                     configuration.defaultCFOptions()
             );
             descriptors[VARIABLE_START_EDGE_HANDLE_INDEX] = new ColumnFamilyDescriptor(
-                    VARIABLE_START_EDGE.encoding().partitionName().get().getBytes(),
+                    new byte[]{VARIABLE_START_EDGE.encoding().ID()},
                     configuration.variableStartEdgeCFOptions()
             );
             descriptors[FIXED_START_EDGE_HANDLE_INDEX] = new ColumnFamilyDescriptor(
-                    FIXED_START_EDGE.encoding().partitionName().get().getBytes(),
+                    new byte[]{FIXED_START_EDGE.encoding().ID()},
                     configuration.fixedStartEdgeCFOptions()
             );
             descriptors[OPTIMISATION_EDGE_HANDLE_INDEX] = new ColumnFamilyDescriptor(
-                    OPTIMISATION_EDGE.encoding().partitionName().get().getBytes(),
+                    new byte[]{OPTIMISATION_EDGE.encoding().ID()},
                     configuration.optimisationEdgeCFOptions()
             );
             descriptors[METADATA_HANDLE_INDEX] = new ColumnFamilyDescriptor(
-                    METADATA.encoding().partitionName().get().getBytes(),
+                    new byte[]{METADATA.encoding().ID()},
                     configuration.metadataCFOptions()
             );
             return Arrays.asList(descriptors);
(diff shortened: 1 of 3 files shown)


database/CorePartitionManager.java
```diff
@@ -125,19 +125,19 @@ protected Data(List<ColumnFamilyDescriptor> descriptors, List<ColumnFamilyHandle
...
                     configuration.fixedStartEdgeCFOptions()
             );
             descriptors[OPTIMISATION_EDGE_HANDLE_INDEX] = new ColumnFamilyDescriptor(
-                    OPTIMISATION_EDGE.encoding().partitionName().get().getBytes(),
+                    new byte[]{OPTIMISATION_EDGE.encoding().ID()},
                     configuration.optimisationEdgeCFOptions()
             );
             descriptors[METADATA_HANDLE_INDEX] = new ColumnFamilyDescriptor(
-                    METADATA.encoding().partitionName().get().getBytes(),
+                    new byte[]{METADATA.encoding().ID()},
```
flyingsilverfin, line 140: RocksDB expects a byte array for the column family name, so we create it here only

graph/common/Encoding.java
```diff
@@ -55,28 +56,21 @@
     public static final int ENCODING_VERSION = 2;
 
     public enum Partition {
-        DEFAULT((short) 0, null),
-        VARIABLE_START_EDGE((short) 1, ByteArray.encodeString("VARIABLE_START_EDGE", STRING_ENCODING)),
-        FIXED_START_EDGE((short) 2, ByteArray.encodeString("FIXED_START_EDGE", STRING_ENCODING)),
-        OPTIMISATION_EDGE((short) 3, ByteArray.encodeString("OPTIMISATION_EDGE", STRING_ENCODING)),
-        METADATA((short) 4, ByteArray.encodeString("METADATA", STRING_ENCODING));
```
flyingsilverfin, line 62: we don't need a string name at all!

graph/common/Storage.java
```diff
@@ -110,7 +110,7 @@ default Schema asSchema() {
             private final Encoding.Partition encoding;
             private final Integer fixedStartBytes;
 
-            public static Partition fromID(short ID) {
+            public static Partition fromID(byte ID) {
```
flyingsilverfin, line 113: this will have to be updated in cluster too

database/CorePartitionManager.java
```diff
@@ -125,19 +125,19 @@ protected Data(List<ColumnFamilyDescriptor> descriptors, List<ColumnFamilyHandle
                     configuration.defaultCFOptions()
```
flyingsilverfin, line 125: Note: the `DEFAULT` column family won't use the ID 0 when creating the column families, even though we'd like it to, because RocksDB requires it be called their built in default name

graph/common/Encoding.java
```diff
@@ -55,28 +56,21 @@
...
+        DEFAULT(0),
+        VARIABLE_START_EDGE(1),
+        FIXED_START_EDGE(2),
+        OPTIMISATION_EDGE(3),
+        METADATA(4);
 
-        private final short ID;
-        // TODO: Remove partition name (See issue #6526)
-        private final ByteArray partitionName;
+        private final byte ID;
```
flyingsilverfin, line 65: A `byte` is actually shorter than a `short`, which will save even more space in TypeDB Cluster.
Note that we never expect to have more than 255 column families, so this should suffice for all future needs

lolski (approved):

PR #6530, opened 2022-03-18
Client transactions are closable when executing queries

flyingsilverfin:

We allow transactions initiated by external RPC clients to be closed immediately, without waiting for the current query to finish running. This also allows server shutdown to proceed without blocking on running queries. Previously, a query that was taking too long could not be terminated at all: not by closing the trasaction from the client-side nor by shutting down the server!

* Do not synchronize `TransactionService.close()` against `TransactionService.execute()`
  * This was preventing the server shutdown (since we first close all open transactions, before stopping the server). 
  * It was also preventing the clients from killing server transactions by closing sessions or transactions on the client side

### DIFF
--- a/server/TransactionService.java
+++ b/server/TransactionService.java
@@ -284,7 +284,7 @@ public class TransactionService implements StreamObserver<TransactionProto.Trans
     }
 
     @Override
-    public synchronized void close() {
+    public void close() {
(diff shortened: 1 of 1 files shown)


flyingsilverfin:
merged with verbal approval from Haikal

PR #6531, opened 2022-03-21
Remove partial support for write transaction inference

flyingsilverfin:

Simplify and remove internal APIs that were designed to support reasoning in write transactions. Reasoning in write transaction is not going to be supported for the foreseeable future, so continuing to support this logic incurs minor but avoidable overheads.

* Remove setters for `isInferred` for both `ThingEdge` and `ThingVertex`
* Remove users of these setters, which were trying to implement "upgrading" an inferred vertex or edge to a user-inserted one in the middle of a write transaction

### DIFF
--- a/graph/ThingGraph.java
+++ b/graph/ThingGraph.java
@@ -309,7 +309,7 @@ public class ThingGraph {
                     return v;
                 }
         );
-        if (!isInferred && vertex.isInferred()) vertex.isInferred(false);
+        assert isInferred == vertex.isInferred();
         return vertex;
     }
 
@@ -327,7 +327,7 @@ public class ThingGraph {
                     return v;
                 }
         );
-        if (!isInferred && vertex.isInferred()) vertex.isInferred(false);
+        assert isInferred == vertex.isInferred();
         return vertex;
     }
 
@@ -345,7 +345,7 @@ public class ThingGraph {
                     return v;
                 }
         );
-        if (!isInferred && vertex.isInferred()) vertex.isInferred(false);
+        assert isInferred == vertex.isInferred();
         return vertex;
     }
 
@@ -374,7 +374,7 @@ public class ThingGraph {
                     return v;
                 }
         );
-        if (!isInferred && vertex.isInferred()) vertex.isInferred(false);
+        assert isInferred == vertex.isInferred();
         return vertex;
     }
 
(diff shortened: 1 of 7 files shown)


graph/ThingGraph.java
```diff
@@ -309,7 +309,7 @@ public ThingVertex convertToReadable(VertexIID.Thing iid) {
                     return v;
                 }
         );
-        if (!isInferred && vertex.isInferred()) vertex.isInferred(false);
+        assert isInferred == vertex.isInferred();
```
flyingsilverfin, line 312: since we no longer support changing the vertex's inferred flag, we should assert that the state that is expected matches the vertex's actual status

graph/adjacency/impl/ThingAdjacencyImpl.java
```diff
@@ -304,10 +304,10 @@ private void put(Encoding.Edge.Thing encoding, ThingEdgeImpl.Buffered edge, IID[
                     if (existingEdge == null) {
                         if (isOut()) owner.graph().edgeCreated(edge); // only record creation in one direction
                         return edge;
-                    } else if (existingEdge.isInferred() && !edge.isInferred()) {
-                        existingEdge.isInferred(true);
+                    } else {
+                        assert existingEdge.isInferred() == edge.isInferred();
```
flyingsilverfin, line 308: for edges, we also check that the existing edge has the expected inferred status

graph/edge/ThingEdge.java
```diff
@@ -40,8 +40,6 @@
 
     Optional<ThingVertex> optimised();
 
-    void isInferred(boolean isInferred);
```
flyingsilverfin, line 43: not required anymore!

graph/edge/impl/ThingEdgeImpl.java
```diff
@@ -235,11 +235,6 @@ public Buffered(Encoding.Edge.Thing encoding, ThingVertex.Write from, ThingVerte
             }
         }
 
-        @Override
-        public void isInferred(boolean isInferred) {
-            this.isInferred = isInferred;
-        }
```
flyingsilverfin, line 241: not required!

graph/vertex/impl/ThingVertexImpl.java
```diff
@@ -246,7 +241,7 @@ public boolean isWrite() {
 
         public static class Buffered extends ThingVertexImpl.Write {
 
-            protected boolean isInferred;
+            private final boolean isInferred;
```
flyingsilverfin, line 244: `protected` was a mistake before, should have been private!
Additionally, we can now use `final`, since the inferred status is not required to change at all

PR #6546, opened 2022-05-09
Rewrite query engine to leverage data sortedness

flyingsilverfin:

We implement a new, non-recursive traversal algorithm that takes advantage of TypeDB's natively sorted data layout to (potentially dramatically) improve query performance. This is especially useful in queries with multiple meeting points which can make use of intersections over the sorted data streams.

The new architecture is also ready to leverage multiple starting points in queries should the query planner provide it.

We have fundamentally changed the the traversal engine in several ways:
1. We utilise a different type of "join" for traversing vertices, which uses sortedness. As part of this we change from an edge-by-edge traversal to a vertex-by-vertex traversal.
2. We use a non-recursive backtracking algorithm instead of a recursive backtracking algorithm, which allows us to short circuit backtracking in tree-shaped queries more easily and without patchy data structures
3. We cannot filter out duplicate roles in the traversal during edge iteration, because we no longer traverse edge-by-edge, but instead filter them out when we visit each vertex as a separate step.

### Algorithmic changes

**Intersection-based vertex exploration**

We can consider this algorithm as roughly equivalent to implementing a sophisticated `merge-join` (https://en.wikipedia.org/wiki/Sort-merge_join).

By using sorted `Forwardable` data iterators, which can seek through the data to a particular point, we can treat all query graph meeting points as intersections over sorted iterators. This is combined with the previous backtracking algorithm, to explore the next query vertex using intersections over iterators from the previously explored and held fixed data vertices.

This implementation flips the query graph exploration from being a backtracking algorithm that explores edge-by-edge to a model that goes vertex-by-vertex. Any number of edges that lead into a new vertex being explored are intersected at the same time. 

Note: In the future, we may choose not to use only intersections when one of the iterators is small, since the underlying RocksDB `seek()` is slower than just doing a `get()`.

The logic of the actual traversal, and checking the validity of intersections, recording where to backtrack to and revisit, is performed at the top level of the `GraphIterator`. The vertex iterators are held by `GraphIterator.VertexTraverser` - these objects manage iterator cleanup and cache the last known intersection for use when re-trying.

**Non-recursive query graph exploration**

We also convert the previous recursive algorithm into a nonrecursive one. It operates in a "traverse" or "backtrack" mode,and in forward exploration always traverses the next lowest of the ordered vertices in a "traverse" set (vertices that must be visited next to find an answer). Or, at a dead end, it consumes the last vertex in the "revisit" set (vertices that have caused a dead end and need to be retried).

Utilising these sets naturally allows the algorithm to short circuit through the plan in tree-shaped queries where part of the exploration are isolated from each other. For example:
```
0: $a [name]
1: $x [person]
  $a -has-> $x
2: $y [dog]
  $a -has-> $y
```
If we fail when exploring `$y`, we can skip back to `$a` without going to `$x` in between. Similarly, in a tree multiple layers deep, we can fast-forward to the children of changed vertices without exploring all the interleaved ones. Abstract example:
```
     0
  /      \
1        2 
|         |
3       4
```
In this case, if `4` fails to find an answer, we can immediately generate the next `2` without changing `3` first. This does moderately change the output of the query plan, however we know that in general DFS in tree-shaped queries is cheaper than a layered BFS approach, which opens more iterators overall. Additionally, this approach is the fastest way to answer a query with no answers (in this example, if there are no 2's with a 4 at all, following the plan to find other 3's has no effect at all!)
This makes us slightly more immune to query planning irregularities at no cost.

Compare this to the recursive algorithm, which implemented the backtracking short circuiting by maintaining complex state from a dead end, in order to skip through unrelated vertices to get to previous vertices that could have caused the dead end. In the new model, we can do both forward and backward skipping by nature of the implementation without specialised state.

**Traversing roles**

We no longer check and filter duplicate roles in a relation's scope at generate time (ie. within the edge iterators).
Instead, we go over the edges regardless of the role, record the role, and then validate no duplicates exist separately at each vertex. To see why this had to be changed, please read: https://github.com/vaticle/typedb/pull/6546/files#r894024814

Note: this is the mechanism that allows us to query `($x, $y)` and get all permutations of `$x` and `$y` without them being the same role players. 

Additionally, because we no longer do edge-by-edge forward and back-tracking, we cannot answer queries like `$r ($x, $x)` by doing an intersection over two identical iterators on `$r` - we would intersect correctly, but would also record identical roles in the process. To solve this, we now de-optimise one of the duplicate role players queries by traversing one `$x --RolePlayer--> $r` optimised edge, and visit one role vertex as well: `$x --Playing--> $_role:$r` and `$r --Relating--> $_role:$r`.  This means we can backtrack to a vertex and get the required answer permutations.

### Smaller changes

* Separate `loop` edges in Traversal objects (including in all of `Traversal` subtypes: `Structure`, `Planner`, and `Procedure`) from `ins` and `outs`.
  * in most places, we were filtering out/specially handling loop edges already. Splitting them into a separate structure makes the special cases more obvious and introduces fewer `if` checks throughout the traversal package
  * NOTE: remaining code smell -`Planner` edges currently MUST be `Directional`, which `loop` edges don't have to be.
* `GraphProcedure` has been changed to be constructed via the `Builder` inner class. Previously, we just used a static factory method to create the `GraphProcedure` from a `GraphPlanner`. 
  * This allows the `GraphProcedure` just to accept an ordered array of `ProcedureVertex`, and the builder takes care of registering and ordering vertices before the `GraphProcedure` is created. This will be simplified after the query planner orders vertices natively.
* We used to restrict, by typing, that only `ThingVertex` was comparable to `ThingVertex`, and `TypeVertex` was comparable to `TypeVertex` for sortedness. However, we now just allow any `Vertex` to be comparable to any `Vertex`, simplifying some generics
  * this was required because the traversal engine only operates over intersections of generic `Vertex` iterators
* Cache hash codes in `Procedure` objects - these hash codes end up incurring a real cost because they are used so often
* Fix various bugs in laziness and evaluation of `Forwardable` iterators


traversal/procedure/GraphProcedure.java
```diff
@@ -54,117 +54,54 @@ public class GraphProcedure implements PermutationProcedure {
...
-    private GraphProcedure(int edgeSize) {
-        vertices = new HashMap<>();
-        edges = new ProcedureEdge<?, ?>[edgeSize];
+    private GraphProcedure(ProcedureVertex<?, ?>[] vertices) {
+        this.vertices = vertices;
     }
 
     public static GraphProcedure create(GraphPlanner planner) {
-        GraphProcedure procedure = new GraphProcedure(planner.edges().size());
+        GraphProcedure.Builder builder = new Builder();
```
flyingsilverfin, line 66: we use the `Builder` to register all the vertices against. the `Builder` then contains the transformation logic for converting edges that have an order to vertices that have an order. This helps us not expose a partially correct `GraphProcedure` at any point, and directly produce one from the Builder when the building is finished.

traversal/procedure/GraphProcedure.java
```diff
@@ -54,117 +54,54 @@ public class GraphProcedure implements PermutationProcedure {
...
-        planner.vertices().forEach(vertex -> procedure.registerVertex(vertex, registeredVertices, registeredEdges));
-        return procedure;
+        planner.vertices().forEach(vertex -> builder.registerVertex(vertex, registeredVertices, registeredEdges));
+        return builder.orderAndbuild();
     }
 
-    public static GraphProcedure.Builder builder(int edgeSize) {
-        GraphProcedure procedure = new GraphProcedure(edgeSize);
-        return procedure.new Builder();
-    }
```
flyingsilverfin, line 77: builder is now static and directly created

traversal/procedure/GraphProcedure.java
```diff
@@ -54,117 +54,54 @@ public class GraphProcedure implements PermutationProcedure {
...
+        return endVertices;
     }
 
-    private ProcedureVertex.Thing thingVertex(Identifier identifier, boolean isStart) {
-        return vertices.computeIfAbsent(
-                identifier, id -> new ProcedureVertex.Thing(id, isStart)
-        ).asThing();
+    public ProcedureVertex<?, ?> vertex(int pos) {
+        assert 0 <= pos && pos < vertices.length;
+        return vertices[pos];
```
flyingsilverfin, line 94: look up vertices by position in the plan

traversal/procedure/GraphProcedure.java
```diff
@@ -214,74 +150,160 @@ public boolean equals(Object o) {
...
-        }
-        str.append("\n\tedges:");
-        for (ProcedureEdge<?, ?> e : procedureEdges) {
-            str.append("\n\t\t").append(e);
+        for (int i = 0; i < vertexCount(); i++) {
+            ProcedureVertex<?, ?> vertex = vertices[i];
+            str.append("\n\t").append(vertex);
+            for (ProcedureEdge<?, ?> edge : vertex.ins()) {
+                str.append("\n\t\t\t").append(edge);
+            }
```
flyingsilverfin, line 170: change the `toString()` to print out the vertex ordering, and group edges below each vertex that lead into that vertex as a group.

traversal/procedure/GraphProcedure.java
```diff
@@ -54,117 +54,54 @@ public class GraphProcedure implements PermutationProcedure {
...
-
-    public ProcedureEdge<?, ?> edge(int pos) {
-        return edges[pos - 1];
-    }
-
-    public int edgesCount() {
-        return edges.length;
-    }
-
-    private void registerVertex(PlannerVertex<?> plannerVertex, Set<PlannerVertex<?>> registeredVertices,
```
flyingsilverfin, line 103: the `register` methods have moved into the `GraphProcedure.Builder`

traversal/procedure/GraphProcedure.java
```diff
@@ -214,74 +150,160 @@ public boolean equals(Object o) {
...
+            for (ProcedureEdge<?, ?> edge : vertex.ins()) {
+                str.append("\n\t\t\t").append(edge);
+            }
         }
         str.append("\n}");
         return str.toString();
     }
 
-    public class Builder {
+    public static class Builder {
```
flyingsilverfin, line 176: now static since we don't want to allow using an incomplete `GraphProcedure` at all

traversal/procedure/GraphProcedure.java
```diff
@@ -214,74 +150,160 @@ public boolean equals(Object o) {
...
+
+        Map<Identifier, ProcedureVertex<?, ?>> vertices;
+
+        public Builder() {
+            this.vertices = new HashMap<>();
+        }
+
+        // TODO: remove when the query planner orders vertices natively
```
flyingsilverfin, line 184: as noted, we need to remove this when the query planner natively orders vertices

cc @dmitrii-ubskii

traversal/procedure/GraphProcedure.java
```diff
@@ -214,74 +150,160 @@ public boolean equals(Object o) {
...
 
         public GraphProcedure build() {
-            assert iterate(edges).noneMatch(Objects::isNull);
-            return GraphProcedure.this;
+            return new GraphProcedure(vertices.values().stream().sorted(Comparator.comparing(ProcedureVertex::order))
+                    .toArray(ProcedureVertex[]::new));
+        }
+
+        /**
+         * TODO: this is a compatibility layer to translate edge-ordered procedures into vertex-ordered procedures
```
flyingsilverfin, line 195: as noted, we need to remove this when the query planner natively orders vertices

cc @dmitrii-ubskii

traversal/procedure/GraphProcedure.java
```diff
@@ -214,74 +150,160 @@ public boolean equals(Object o) {
...
+            List<ProcedureVertex<?, ?>> vertexList = iterate(vertices.values()).toList();
+            vertexList.sort(Comparator.comparing(v -> v.lastInEdge() == null ? 0 : v.lastInEdge().order()));
+            for (int i = 0; i < vertexList.size(); i++) {
+                orderedVertices[i] = vertexList.get(i);
+                orderedVertices[i].setOrder(i);
+            }
+            return orderedVertices;
+        }
+
+        private void registerVertex(PlannerVertex<?> plannerVertex, Set<PlannerVertex<?>> registeredVertices,
```
flyingsilverfin, line 209: cut and pasted from `GraphProcedure`

traversal/procedure/GraphProcedure.java
```diff
@@ -214,74 +150,160 @@ public boolean equals(Object o) {
...
-            return namedType(name, false);
+        private ProcedureVertex.Type typeVertex(Identifier identifier) {
+            return vertices.computeIfAbsent(identifier, ProcedureVertex.Type::new).asType();
         }
 
-        public ProcedureVertex.Type namedType(String name, boolean isStart) {
-            return typeVertex(Identifier.Variable.of(Reference.name(name)), isStart);
+        // ---- manual builder methods ----
+
+        public ProcedureVertex.Type labelledType(int order, String label) {
```
flyingsilverfin, line 279: note: a separate set of builder methods exist for manually building procedures. Here the vertices directly get the order number.

traversal/procedure/GraphProcedure.java
```diff
@@ -300,157 +322,169 @@ public ProcedureVertex.Thing setPredicate(ProcedureVertex.Thing thing, Predicate
         }
 
         public ProcedureEdge.Native.Type.Sub.Forward forwardSub(
-                int order, ProcedureVertex.Type child, ProcedureVertex.Type parent, boolean isTransitive) {
+                ProcedureVertex.Type child, ProcedureVertex.Type parent, boolean isTransitive) {
             ProcedureEdge.Native.Type.Sub.Forward edge =
-                    new ProcedureEdge.Native.Type.Sub.Forward(child, parent, order, isTransitive);
+                    new ProcedureEdge.Native.Type.Sub.Forward(child, parent, -1, isTransitive);
```
flyingsilverfin, line 327: edges don't need an order if the vertices already have one natively!

traversal/procedure/ProcedureVertex.java
```diff
@@ -106,18 +102,54 @@ public ProcedureVertex.Type asType() {
...
+    public Set<Identifier.Variable> scopesVisited() {
+        if (scopesVisited == null) scopesVisited = computeScopesVisited();
+        return scopesVisited;
+    }
+
+    Set<Identifier.Variable> computeScopesVisited() {
+        return set();
+    }
+
+    void setOrder(int order) {
```
flyingsilverfin, line 137: This setter needs to be removed this when the query planner natively orders vertices

cc @dmitrii-ubskii

traversal/procedure/ProcedureEdge.java
```diff
@@ -131,22 +133,6 @@ public Encoding.Direction.Edge direction() {
...
-        return false;
-    }
-
-    public boolean onlyStartsFromRelation() {
-        return false;
-    }
-
-    public boolean onlyEndsAtRelation() {
-        return false;
-    }
```
flyingsilverfin, line 148: un-needed methods in the new model

traversal/procedure/ProcedureEdge.java
```diff
@@ -81,11 +81,13 @@
 
     private final int order;
     private final Encoding.Direction.Edge direction;
+    private final int hash;
 
     private ProcedureEdge(VERTEX_FROM from, VERTEX_TO to, int order, Encoding.Direction.Edge direction, String symbol) {
         super(from, to, symbol);
         this.order = order;
```
flyingsilverfin, line 88: This state needs to be removed once the planner natively spits out an ordering for vertices instead of edges.

cc @dmitrii-ubskii

traversal/procedure/ProcedureEdge.java
```diff
@@ -1186,16 +1178,21 @@ public boolean isClosure(GraphManager graphMgr, Vertex<?, ?> fromVertex, Vertex<
                                                 .edge(ROLEPLAYER, rt, player.iid().prefix(), player.iid().type())
                                                 .toAndOptimised(),
                                         ASC
-                                ).filter(kv -> kv.key().equals(player) && !scoped.contains(kv.value()));
+                                ).filter(kv -> kv.key().equals(player));
```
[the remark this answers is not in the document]
flyingsilverfin: TODO: try to re-instante this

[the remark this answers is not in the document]
flyingsilverfin: So I attemped to revert back to using the role scope filtering during generation, rather than at vertex verification time.

The main issue is that we are no longer moving forward and backtracking based on edges, but instead we are doing vertex-by-vertex.

So now imagine a plan:
```
0: $r [friendship]
1: $x [dog, person]
  $r --RolePlayer--> $x
2: $y [dog]
  $r --RolePlayer--> $y
3: $a [dog-name]
  $y --Has--> $a
```

In this case, traversing `$r` will forward verify the edges generating `$y` and `$x`. However, `$x` is checked first and occupies the only role leading to a `dog`. We then verify that `$y` has an instance. In the old "filtering" model, we will immediately filter out the only role leading to a valid `dog`, since that is occupied by the traversal for `$x`. However, since we don't want to have multiple types of backtracking, we can't re-visit `$x` to get the next person and free up the `dog` role for `$y` to use.

Even if we did implement edge backtracking during the edge lookahead, the cases for backtracking later are difficult. Imagine the first intersections succeed, and we find a `$x` and `$y`. However, the _next_ `$y` may not be found because of `$x`, so we need to know that `$y` could have failed because of `$x` - tracking this causality with the filtering becomes really hard.

traversal/procedure/ProcedureEdge.java
```diff
@@ -1186,16 +1177,21 @@ public boolean isClosure(GraphManager graphMgr, Vertex<?, ?> fromVertex, Vertex<
...
                         closures.forward(KeyValue.of(player, null));
                         Optional<KeyValue<ThingVertex, ThingVertex>> next = closures.first();
                         if (next.isPresent() && next.get().key().equals(player)) {
-                            scoped.push(next.get().value(), order());
+                            scoped.record(this, next.get().value());
```
flyingsilverfin, line 1184: rename

traversal/procedure/ProcedureVertex.java
```diff
@@ -37,65 +37,60 @@
...
         PROPERTIES extends TraversalVertex.Properties
         > extends TraversalVertex<ProcedureEdge<?, ?>, PROPERTIES> {
 
-    private final boolean isStartingVertex;
-    private final AtomicReference<Set<Integer>> dependedEdgeOrders;
-    private ProcedureEdge<?, ?> branchEdge;
+    private int order;
+    private ProcedureEdge<?, ?> lastInEdge;
+    private List<ProcedureEdge<?, ?>> orderedOuts;
+    private Set<ProcedureEdge<?, ?>> transitiveOuts;
```
flyingsilverfin, line 68: cache various useful sets

traversal/procedure/ProcedureVertex.java
```diff
@@ -37,65 +37,60 @@
...
-        this.dependedEdgeOrders = new AtomicReference<>(null);
     }
 
     public abstract Forwardable<? extends VERTEX, Order.Asc> iterator(GraphManager graphMgr, Traversal.Parameters parameters);
 
     @Override
     public void in(ProcedureEdge<?, ?> edge) {
         super.in(edge);
-        if (branchEdge == null || edge.order() < branchEdge.order()) branchEdge = edge;
+        if (lastInEdge == null || edge.order() > lastInEdge.order()) lastInEdge = edge;
```
flyingsilverfin, line 80: `lastInEdge` and its consequences can be deleted when the query planner orders vertices natively

cc @dmitrii-ubskii

traversal/procedure/ProcedureVertex.java
```diff
@@ -106,18 +101,47 @@ public ProcedureVertex.Type asType() {
...
+
+    void setOrder(int order) {
+        this.order = order;
+    }
+
     @Override
     public String toString() {
-        String str = super.toString();
-        if (isStartingVertex) str += " (start)";
+        String str = order() + ": " + super.toString();
```
flyingsilverfin, line 133: `ProcedureVertex` prints its own order number - useful for debugging as well when inspecting variables!

traversal/procedure/ProcedureVertex.java
```diff
@@ -135,6 +159,13 @@ public Thing asThing() {
             return this;
         }
 
+        public Set<ProcedureEdge<?, ?>> inRolePlayers() {
```
flyingsilverfin, line 162: used repeatedly during traversal, so we cache it

PR #6559, opened 2022-05-31
Refactor Reactives to remove Operators abstraction

jmsfltchr:

We had introduced an abstraction such that each Reactive contained an Operation. However, we found this abstraction redundant and want to remove it in favour of inheritance.

- Operations are removed in favour of sub-typing one of the base Reactive implementations.


reasoner/processor/reactive/Source.java
```diff
@@ -20,44 +20,50 @@
...
-                                                 Operator.Source<OUTPUT> operator) {
-        return new Source<>(processor, operator);
+    public PACKET next() {
+        assert !isExhausted();
+        return iterator().next();
     }
 
-    private Operator.Source<PACKET> operator() {
-        return sourceOperator;
+    private boolean isExhausted() {
```
flyingsilverfin, line 49: why not just mirror iterator terminology and call this `hasNext()`?

reasoner/processor/reactive/TransformationStream.java
```diff
@@ -121,7 +98,56 @@ public Stream<OUTPUT, OUTPUT> buffer() {
 
     @Override
     public String toString() {
-        return getClass().getSimpleName() + ":" + operator().getClass().getSimpleName();
+        return getClass().getSimpleName();
+    }
+
+    public static class MapStream<INPUT, OUTPUT> extends TransformationStream<INPUT, OUTPUT> {
```
flyingsilverfin, line 104: so I didn't delve too deep into this PR but fundamentally this is what we discussed changing and it looks ok to me!

flyingsilverfin (approved):
When I'm back we'll continue our review end to end I think!

PR #6565, opened 2022-06-08
Add retries to SoundnessVerifier for reasoner soundness tests

krishnangovindraj:
Behaviour tests for the soundness of the reasoner fail when they involve cyclic explanations. This was due to the incorrect mapping of inferred concepts between the reasoner's transaction and the oracle's transaction.

The SoundnessVerifier can only correctly map an inferred concept 'C' when it processes an explanation where
1. All the concepts in the condition of the rule are mapped
2. C occurs in the conclusion of the rule3. 

This repeatedly processes the explanation set, processing only explanations which satisfy the above condition. 
Each iteration verifies some explanations, enabling others to be verified in the following iteration, until all explanations have been processed. If progress is not made, an exception is thrown.

### DIFF
--- a/test/behaviour/reasoner/verification/SoundnessVerifier.java
+++ b/test/behaviour/reasoner/verification/SoundnessVerifier.java
@@ -44,13 +44,13 @@ class SoundnessVerifier {
     private final ForwardChainingMaterialiser materialiser;
     private final TypeDB.Session session;
     private final Map<Concept, Concept> inferredConceptMapping;
-    private final Set<Explanation> verifiedExplanations;
+    private final Set<Explanation> collectedExplanations;
 
     private SoundnessVerifier(ForwardChainingMaterialiser materialiser, TypeDB.Session session) {
         this.materialiser = materialiser;
         this.session = session;
         this.inferredConceptMapping = new HashMap<>();
-        this.verifiedExplanations = new HashSet<>();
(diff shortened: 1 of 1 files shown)


jmsfltchr (approved):

jmsfltchr:
Approved subject to adding the dependency on `behaviour` and seeing the tests green!

test/behaviour/reasoner/verification/SoundnessVerifier.java
```diff
@@ -59,21 +59,54 @@ static SoundnessVerifier create(ForwardChainingMaterialiser materialiser, TypeDB
...
+            // Too slow? Create a dependency graph, retry only those from which a newly verified explanation is reachable
+            Set<Explanation> unverifiedExplanations = collectedExplanations;
+            Set<Explanation> dependentUnverifiedExplanations = new HashSet<>();
+            boolean madeProgress = true;
+            while (!unverifiedExplanations.isEmpty() && madeProgress) {
+                madeProgress = false;
+                for (Explanation e : unverifiedExplanations) {
+                    if (canExplanationBeVerified(e)) {
+                        verifyExplanationAndMapConcepts(e);
+                        madeProgress = true;
```
jmsfltchr, line 78: Nice, I like this much better

PR #6590, opened 2022-06-16
Fix intermittent incomplete type matches

dmitrii-ubskii:

Fix a bug that caused an `isa` query where the type is a variable to occasionally match fewer types than expected.

A typo in ProcedureEdge caused the traversal to only consider direct types of a thing when traversing the `isa` edge when coming from the type. It appears that this is a rarely chosen direction of an `isa` edge in the planning (and, therefore, in the behaviour test suite), so an explicit test is added.

### DIFF
--- a/traversal/procedure/ProcedureEdge.java
+++ b/traversal/procedure/ProcedureEdge.java
@@ -362,7 +362,7 @@ public abstract class ProcedureEdge<
                     if (to.props().hasIID()) {
                         return to.iterateAndFilterFromIID(graphMgr, params).filter(vertex -> isaTypes.contains(vertex.type()));
                     } else {
-                        FunctionalIterator<TypeVertex> toTypes = iterate(isaTypes).filter(v -> to.props().types().contains(type.properLabel()));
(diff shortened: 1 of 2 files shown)


traversal/procedure/ProcedureEdge.java
```diff
@@ -362,7 +362,7 @@ static class Backward extends Isa<ProcedureVertex.Type, ProcedureVertex.Thing> {
                     if (to.props().hasIID()) {
                         return to.iterateAndFilterFromIID(graphMgr, params).filter(vertex -> isaTypes.contains(vertex.type()));
                     } else {
-                        FunctionalIterator<TypeVertex> toTypes = iterate(isaTypes).filter(v -> to.props().types().contains(type.properLabel()));
+                        FunctionalIterator<TypeVertex> toTypes = iterate(isaTypes).filter(v -> to.props().types().contains(v.properLabel()));
```
dmitrii-ubskii, line 365: In the test, the inferred types of `$x` only contain `person`. When `type` is `thing` or `entity`, the lambda body always returned false, even though `person` is in the set `isaTypes`.

haikalpribadi (approved):
Oh boy! Great catch, @dmitrii-ubskii.

Issue #6550, opened 2022-05-19
Query slow-down on adding attribute comparison


krishnangovindraj:
## Description
I have a molecular model dataset with a relation `bond` which forms cycles and is symmetric. 
```
    chemical_entity sub entity,
        plays constitutes:constituent,
        plays bond:constituent_1,
        plays bond:constituent_2;

    atom sub entity,
        owns atom_id @key,
        owns charge;

    atom_c sub atom;
    (...)

    bond sub relation,
        relates compound,
        relates constituent_1,
        relates constituent_2;

    single_bond sub bond;
    double_bond sub bond;
    aromatic_bond sub bond;

```   

The following are queries to identify benzene rings (6 `aromatic_bond` relations starting and ending with the same `atom_c` entity). The latter de-duplicates by enforcing an order.
```
# fast: 84 results in 132 ms
match
    (constituent_1:$c1, constituent_2:$c2) isa aromatic_bond;
    (constituent_1:$c2, constituent_2:$c3) isa aromatic_bond;
    (constituent_1:$c3, constituent_2:$c4) isa aromatic_bond;
    (constituent_1:$c4, constituent_2:$c5) isa aromatic_bond;
    (constituent_1:$c5, constituent_2:$c6) isa aromatic_bond;
    (constituent_1:$c6, constituent_2:$c1) isa aromatic_bond;
    $c1 isa atom_c, has atom_id $e1;
    $c2 isa atom_c, has atom_id $e2;
    $c3 isa atom_c, has atom_id $e3;
    $c4 isa atom_c, has atom_id $e4;
    $c5 isa atom_c, has atom_id $e5;
    $c6 isa atom_c, has atom_id $e6;
get
    $e1, $e2, $e3, $e4, $e5, $e6;

# slow: 14 results in 252487 ms 
match
    (constituent_1:$c1, constituent_2:$c2) isa aromatic_bond;
    (constituent_1:$c2, constituent_2:$c3) isa aromatic_bond;
    (constituent_1:$c3, constituent_2:$c4) isa aromatic_bond;
    (constituent_1:$c4, constituent_2:$c5) isa aromatic_bond;
    (constituent_1:$c5, constituent_2:$c6) isa aromatic_bond;
    (constituent_1:$c6, constituent_2:$c1) isa aromatic_bond;
    $c1 isa atom_c, has atom_id $e1;
    $c2 isa atom_c, has atom_id $e2;
    $c3 isa atom_c, has atom_id $e3;
[...]

PR #6592, opened 2022-06-20
Update query planner to allow multiple starting points

dmitrii-ubskii:

We implement a new formulation of the optimisation problem at the heart of query planning that allows the new traversal algorithm to start from multiple points in the query graph. The new query planner also takes into account the specifics of the traversal engine to directly optimise the expected computational costs of traversal.

### Summary

The new vertex-ordering query planner places no restriction on the number of starting vertices to allow the new traversal engine to fully leverage the intersection iterators. The bulk of the meaningful change is contained in GraphPlanner and PlannerEdge/Vertex. This change leads to minor tweaks in GraphProcedure and GraphIterator that allow us to maintain compatibility between the classes.

The linear solver was also switched from the mixed-integer (that is, mixing integer and floating-point variables) SCIP mode to an integer SAT mode, which is optimised around all the variables and coefficients being integer valued, and as such allows the solver to arrive at a better or optimal solution faster.

### Query Planner

The new query planner attempts to directly estimate the number of database operations performed during traversal for a given query plan and minimizes that function. 

The traversal is performed as follows. For each value in the traversal procedure, the number of values we expect to hit every time we enter it is estimated to be either the minimum among the traversal cost of its incoming edges, or, if it has no incoming edges, the estimated size of the vertex itself. As the traversal is performed in an ordered fashion, the number of times we enter a vertex—and, therefore, the number of times we open an iterator into it and iterate through its values—depends straightforwardly on all the vertices we have visited before.

As an illustration, consider a diamond query graph with a given ordering.

```
   0
 ╱   ╲
1  —  2
 ╲   ╱
   3
```

Its traversal, in pseudocode, is performed like so:
```
i0 := iterate(0, {})
for v0 in i0:
    i1 := iterate(1, {v0})
    for v1 in i1:
        i2 := iterate(2, {v0, v1})
        for v2 in i2:
            i3 := iterate(3, {v1, v2})
            for v3 in i3:
                emit(v0, v1, v2, v3)
```

Let α(_v_) be the size of the vertex _v_, and β(_v_, _w_) be the number of values in _w_ that we expect to match for each value of the vertex _v_. Here is that same snippet, with each loop now annotated with the expected number of iterations:

```
i0 := iterate(0, {})
for v0 in i0:                           # α(0) iterations
    i1 := iterate(1, {v0})
    for v1 in i1:                       # β(0, 1) iterations
        i2 := iterate(2, {v0, v1})
        for v2 in i2:                   # min(β(0, 2), β(1, 2)) iterations
            i3 := iterate(3, {v1, v2})
            for v3 in i3:               # min(β(1, 3), β(2, 3)) iterations
                emit(v0, v1, v2, v3)
```

Note: we take minimum of βs, because the resulting set of values in a vertex is an intersection of the sets produced by all incoming edges, which has cardinality of at most min(β,..).

We will omit the parentheses in the α and β notation from this point on to improve legibility in the formulae.

It is easy to see that the number of iterators we create (or reset) in the previous snippet is

`seeks  = 1 + α0(1 + β01(1 + min(β02, β12)))`,

and the number of times we advance an iterator is 

`nexts  = α0(1 + β01(1 + min(β02, β12)(1 + min(β13, β23))))`.

Opening the brackets reveals that the difference between the two values is `α0 β01 min(β02, β12) min(β13, β23) - 1`. While seeks are significantly (5x to 10x) more expensive than nexts, optimising just them loses information on the cost of traversing the end vertices which will lead to the planner favouring expensive bipartite plans with as many ending vertices as possible. A simple solution therefore is to optimise the number of times the storage iterators are advanced. A more complex set of constraints taking that difference into account can be implemented in the future if needed.

To recap, so far the function we'd like to optimise looks like `α0(1 + β01(1 + min(β02, β12)(1 + min(β13, β23))))`. This is, however, not easily representable as a linear combination of some variables. The route we take is to instead minimize the function `(1 + α0) (1 + β01) (1 + min(β02, β12)) (1 + min(β13, β23))` (note the placement of parentheses). This formula strictly overestimates the number of operations performed.

This function generalizes to 

Π{_v_ ∈ S} (1 + α(_v_)) * Π{_v_ ∈ V}(1 + min({β(_x_, _v_): edge(_x_, _v_) is selected}))

where S is the set of starting vertices, V is the set of all vertices, and minimum of an empty set is defined to be 0.

Taking the natural logarithm of this expressions transforms the product into a sum of logarithms, which can be represented as a linear combination of boolean variables:

Σ{_v_ ∈ V} (log(1 + α(_v_)) * _v_:isStartingVertex) + Σ{_e_ ∈ E} (log(1 + β(_e_)) * _e_:isMinimal)

where:
* V is the set of all vertices,
* E is the set of all edges,
* β(_e_) is defined analogously to β(_v_, _w_) above (this accounts for edge multiplicity as well),
* _v_:isStartingVertex is true iff _v_ has no incoming edges,
* _e_:isMinimal is true iff _e_ is selected by the plan, and its cost is the lowest among all the incoming edges of its destination vertex.

This is the objective function of the new planner. The constraints serve to infer the values of these boolean variables from an ordering of the vertices.

### Performance measurements

The two commits we compare in this section are
* ec3b63054, which is exactly equivalent to master, but for the collection of the RocksDB usage statistics,
* a155d32b7, the latest commit in this PR at the time of measurement.

The main criterion we consider is the number of RocksDB operations, which we call here seeks and nexts. Timings are given as an intuitive reference only.

#### typedb-benchmark

The benchmark was run only once for each build. This was not intended as a true performance comparison, but as a smoke test, to make sure that the new planner does not degrage the server's performance.

With the build on current master, the simulation ran for 15 minutes and 45 seconds, performing 44 million seeks and 44 million nexts. 

With the new planner, the simulation ran for just over 12 minutes, performing 58 million seeks and 40 million nexts.

#### Complex queries (see #6550)

Each of the two queries—with and without attribute comparison—was run several times. After each query the statistics were collected and the server was restarted in order to purge all caches.

Both versions performed comparably on the query without attribute comparisons, performing around 3000 seeks and 2700 nexts and finishing within 600 to 700 ms.

On the query _with_ attribute comparisons, the version on master performed erratically, taking from 1 to 5 seconds to complete. It performed anywhere from 80'000 seeks, 300'000 nexts to 800'000 seeks, 600'000 nexts depending on the run. In contrast, the version in this PR performed around 3200 seeks and 10'000 next run after run, performance comparable to that of the query with no attribute comparisons.

### Smaller changes

* Improvements to `Optimiser`:
  * QoL improvements: native retrieval of the initial value, optimiser agnostic `value` retrieval;
  * allow to change constraints between optimiser passes;
  * bugfix: don't reset the model hint after first optimiser pass.
* Streamlined `GraphPlanner`:
  * easier to follow cost computation path:
    * costs computed separately from setting the objective function coefficients;
    * no back and forth calls between GraphPlanner and PlannerEdge/Vertex;
  * simplified value initialisation to a method;
  * methods reordered for readability.
* Stricter equality comparisons between `Identifier`s. 
* Stricter equality comparisons between `ProcedureEdge`s now that we can't rely on `order` for disambiguation.

### DIFF
--- a/common/optimiser/Optimiser.java
+++ b/common/optimiser/Optimiser.java
@@ -46,6 +46,7 @@ public class Optimiser {
     private MPSolverParameters parameters;
     private Status status;
     private boolean hasSolver;
+    private double objectiveValueCache;
 
     public Optimiser() {
         variables = new ArrayList<>();
@@ -56,6 +57,10 @@ public class Optimiser {
     }
 
     public synchronized Status optimise(long timeLimitMillis) {
+        if (hasSolver && constraintsChanged()) {
+            constraints.forEach(constraint -> constraint.initialise(solver));
+            status = Status.FEASIBLE;
+        }
         if (isOptimal()) return status;
         else if (!hasSolver) initialiseSolver();
         solver.setTimeLimit(timeLimitMillis);
@@ -78,8 +83,12 @@ public class Optimiser {
         return status == Status.ERROR;
     }
 
+    private boolean constraintsChanged() {
+        return constraints.stream().anyMatch(c -> !c.isInitialised);
+    }
+
     private void initialiseSolver() {
-        solver = MPSolver.createSolver("SCIP");
+        solver = MPSolver.createSolver("SAT");
         solver.objective().setMinimization();
         parameters = new MPSolverParameters();
         parameters.setIntegerParam(PRESOLVE, PRESOLVE_ON.swigValue());
@@ -92,6 +101,7 @@ public class Optimiser {
     }
 
     private void releaseSolver() {
+        objectiveValueCache = solver.objective().value();
         constraints.forEach(OptimiserConstraint::release);
         variables.forEach(OptimiserVariable::release);
         parameters.delete();
@@ -103,13 +113,18 @@ public class Optimiser {
         objectiveCoefficients.forEach((var, coeff) -> solver.objective().setCoefficient(var.mpVariable(), coeff));
     }
 
+    public double objectiveValue() {
+        if (hasSolver) return solver.objective().value();
+        else return objectiveValueCache;
+    }
+
     private void applyInitialisation() {
         assert iterate(variables).allMatch(OptimiserVariable::hasInitial);
         MPVariable[] mpVariables = new MPVariable[variables.size()];
         double[] initialisations = new double[variables.size()];
         for (int i = 0; i < variables.size(); i++) {
             mpVariables[i] = variables.get(i).mpVariable();
-            initialisations[i] = variables.get(i).getInitial();
+            initialisations[i] = variables.get(i).hint();
         }
         solver.setHint(mpVariables, initialisations);
     }
--- a/common/optimiser/Optimiser.java
+++ b/common/optimiser/Optimiser.java
@@ -58,7 +58,7 @@ public class Optimiser {
 
     public synchronized Status optimise(long timeLimitMillis) {
         if (hasSolver && constraintsChanged()) {
-            constraints.forEach(constraint -> constraint.initialise(solver));
+            constraints.forEach(OptimiserConstraint::updateCoefficients);
             status = Status.NOT_SOLVED;
         }
         if (isOptimal()) return status;
@@ -88,7 +88,7 @@ public class Optimiser {
     }
 
     private boolean constraintsChanged() {
-        return iterate(constraints).anyMatch(c -> !c.isInitialised);
+        return iterate(constraints).anyMatch(c -> !c.isUpToDate);
     }
 
     private void initialiseSolver() {
@@ -98,7 +98,7 @@ public class Optimiser {
         parameters.setIntegerParam(PRESOLVE, PRESOLVE_ON.swigValue());
         parameters.setIntegerParam(INCREMENTALITY, INCREMENTALITY_ON.swigValue());
         variables.forEach(var -> var.initialise(solver));
-        constraints.forEach(constraint -> constraint.initialise(solver));
+        constraints.forEach(constraint -> { constraint.initialise(solver); constraint.updateCoefficients(); });
         applyObjective();
         applyInitialisation();
         hasSolver = true;
--- a/common/optimiser/Optimiser.java
+++ b/common/optimiser/Optimiser.java
@@ -59,14 +59,14 @@ public class Optimiser {
     public synchronized Status optimise(long timeLimitMillis) {
         if (hasSolver && constraintsChanged()) {
             constraints.forEach(constraint -> constraint.initialise(solver));
-            status = Status.FEASIBLE;
+            status = Status.NOT_SOLVED;
         }
         if (isOptimal()) return status;
         else if (!hasSolver) initialiseSolver();
         solver.setTimeLimit(timeLimitMillis);
         status = Status.of(solver.solve(parameters));
-        variables.forEach(OptimiserVariable::recordValue);
-        clearInitialisation();
+        if (status != Status.NOT_SOLVED && status != Status.ERROR)
+            variables.forEach(OptimiserVariable::recordValue);
         if (isOptimal()) releaseSolver();
         return status;
     }
--- a/common/optimiser/Optimiser.java
+++ b/common/optimiser/Optimiser.java
@@ -88,7 +88,7 @@ public class Optimiser {
     }
 
     private boolean constraintsChanged() {
-        return constraints.stream().anyMatch(c -> !c.isInitialised);
+        return iterate(constraints).anyMatch(c -> !c.isInitialised);
     }
 
     private void initialiseSolver() {
--- a/common/optimiser/Optimiser.java
+++ b/common/optimiser/Optimiser.java
@@ -46,7 +46,7 @@ public class Optimiser {
     private MPSolverParameters parameters;
     private Status status;
     private boolean hasSolver;
-    private double objectiveValueCache;
+    private double objectiveValue;
 
     public Optimiser() {
         variables = new ArrayList<>();
@@ -65,10 +65,14 @@ public class Optimiser {
         else if (!hasSolver) initialiseSolver();
         solver.setTimeLimit(timeLimitMillis);
         status = Status.of(solver.solve(parameters));
-        if (status != Status.NOT_SOLVED && status != Status.ERROR)
-            variables.forEach(OptimiserVariable::recordValue);
+        recordValues();
         if (isOptimal()) releaseSolver();
         return status;
+    }
+
+    private void recordValues() {
+        objectiveValue = solver.objective().value();
+        if (status != Status.NOT_SOLVED && status != Status.ERROR) variables.forEach(OptimiserVariable::recordValue);
     }
 
     public boolean isOptimal() {
@@ -101,7 +105,6 @@ public class Optimiser {
     }
 
     private void releaseSolver() {
-        objectiveValueCache = solver.objective().value();
         constraints.forEach(OptimiserConstraint::release);
         variables.forEach(OptimiserVariable::release);
         parameters.delete();
@@ -114,8 +117,7 @@ public class Optimiser {
     }
 
     public double objectiveValue() {
-        if (hasSolver) return solver.objective().value();
-        else return objectiveValueCache;
+        return objectiveValue;
     }
 
     private void applyInitialisation() {
--- a/common/optimiser/Optimiser.java
+++ b/common/optimiser/Optimiser.java
@@ -131,10 +131,6 @@ public class Optimiser {
         solver.setHint(mpVariables, initialisations);
     }
 
-    private void clearInitialisation() {
-        solver.setHint(new MPVariable[0], new double[0]);
-    }
-
     public void setObjectiveCoefficient(OptimiserVariable<?> var, double coeff) {
         objectiveCoefficients.put(var, coeff);
         if (hasSolver) solver.objective().setCoefficient(var.mpVariable(), coeff);
--- a/common/optimiser/OptimiserConstraint.java
+++ b/common/optimiser/OptimiserConstraint.java
@@ -42,12 +42,13 @@ public class OptimiserConstraint {
     }
 
     public void setCoefficient(OptimiserVariable<?> variable, double coeff) {
-        if (isInitialised)
+        if (isInitialised) {
             if (coefficients.containsKey(variable)) {
                 isInitialised = coefficients.get(variable) == coeff;
             } else {
                 isInitialised = false;
             }
+        }
         coefficients.put(variable, coeff);
     }
 
--- a/common/optimiser/OptimiserConstraint.java
+++ b/common/optimiser/OptimiserConstraint.java
@@ -31,35 +31,39 @@ public class OptimiserConstraint {
     private final String name;
     private final Map<OptimiserVariable<?>, Double> coefficients;
     private MPConstraint mpConstraint;
-    boolean isInitialised;
+    boolean isUpToDate;
 
     public OptimiserConstraint(double lowerBound, double upperBound, String name) {
         this.lowerBound = lowerBound;
         this.upperBound = upperBound;
         this.coefficients = new HashMap<>();
         this.name = name;
-        this.isInitialised = false;
+        this.isUpToDate = false;
     }
 
     public void setCoefficient(OptimiserVariable<?> variable, double coeff) {
-        if (isInitialised) {
+        if (isUpToDate) {
             if (coefficients.containsKey(variable)) {
-                isInitialised = coefficients.get(variable) == coeff;
+                isUpToDate = coefficients.get(variable) == coeff;
             } else {
-                isInitialised = false;
+                isUpToDate = false;
             }
         }
         coefficients.put(variable, coeff);
     }
 
     synchronized void initialise(MPSolver solver) {
-        if (this.mpConstraint == null) this.mpConstraint = solver.makeConstraint(lowerBound, upperBound, name);
+        this.mpConstraint = solver.makeConstraint(lowerBound, upperBound, name);
+    }
+
+    synchronized void updateCoefficients() {
         coefficients.forEach((var, coeff) -> mpConstraint.setCoefficient(var.mpVariable(), coeff));
-        isInitialised = true;
+        isUpToDate = true;
     }
 
     synchronized void release() {
-        this.mpConstraint.delete();
-        this.mpConstraint = null;
+        mpConstraint.delete();
+        mpConstraint = null;
+        isUpToDate = false;
     }
 }
--- a/common/optimiser/OptimiserConstraint.java
+++ b/common/optimiser/OptimiserConstraint.java
@@ -31,24 +31,34 @@ public class OptimiserConstraint {
     private final String name;
     private final Map<OptimiserVariable<?>, Double> coefficients;
     private MPConstraint mpConstraint;
+    boolean isInitialised;
 
     public OptimiserConstraint(double lowerBound, double upperBound, String name) {
         this.lowerBound = lowerBound;
         this.upperBound = upperBound;
         this.coefficients = new HashMap<>();
         this.name = name;
+        this.isInitialised = false;
     }
 
     public void setCoefficient(OptimiserVariable<?> variable, double coeff) {
+        if (isInitialised)
+            if (coefficients.containsKey(variable)) {
+                isInitialised = coefficients.get(variable) == coeff;
+            } else {
+                isInitialised = false;
+            }
         coefficients.put(variable, coeff);
     }
 
     synchronized void initialise(MPSolver solver) {
-        this.mpConstraint = solver.makeConstraint(lowerBound, upperBound, name);
+        if (this.mpConstraint == null) this.mpConstraint = solver.makeConstraint(lowerBound, upperBound, name);
         coefficients.forEach((var, coeff) -> mpConstraint.setCoefficient(var.mpVariable(), coeff));
+        isInitialised = true;
     }
 
     synchronized void release() {
         this.mpConstraint.delete();
+        this.mpConstraint = null;
     }
 }
--- a/common/optimiser/OptimiserVariable.java
+++ b/common/optimiser/OptimiserVariable.java
@@ -54,10 +54,6 @@ public abstract class OptimiserVariable<T> {
         this.initial = initial;
     }
 
-    public void clearInitial() {
-        initial = null;
-    }
-
     public boolean hasInitial() {
         return initial != null;
     }
--- a/common/optimiser/OptimiserVariable.java
+++ b/common/optimiser/OptimiserVariable.java
@@ -26,27 +26,60 @@ import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.UNE
 
 public abstract class OptimiserVariable<T> {
 
+    protected T initial;
+    protected T solution;
+    protected MPVariable mpVariable;
+
     final String name;
 
     OptimiserVariable(String name) {
         this.name = name;
     }
 
-    public abstract T solutionValue();
+    public T value() {
+        if (hasSolutionValue()) return solutionValue();
+        else return initial();
+    }
 
-    abstract MPVariable mpVariable();
+    public boolean hasSolutionValue() {
+        return solution != null;
+    }
+
+    public T solutionValue() {
+        assert hasSolutionValue();
+        return solution;
+    }
+
+    public void setInitial(T initial) {
+        this.initial = initial;
+    }
+
+    public void clearInitial() {
+        initial = null;
+    }
+
+    public boolean hasInitial() {
+        return initial != null;
+    }
+
+    public T initial() {
+        assert hasInitial();
+        return initial;
+    }
+
+    public abstract double hint();
+
+    MPVariable mpVariable() {
+        return mpVariable;
+    }
 
     abstract void recordValue();
 
-    public abstract void clearInitial();
-
-    abstract boolean hasInitial();
-
-    public abstract double getInitial();
-
     abstract void initialise(MPSolver solver);
 
-    abstract void release();
+    synchronized void release() {
+        this.mpVariable.delete();
+    }
 
     @Override
     public String toString() {
@@ -55,23 +88,8 @@ public abstract class OptimiserVariable<T> {
 
     public static class Boolean extends OptimiserVariable<java.lang.Boolean> {
 
-        private java.lang.Boolean initial;
-        private java.lang.Boolean solution;
-        private MPVariable mpVariable;
-
         public Boolean(String name) {
             super(name);
-        }
-
-        @Override
-        public java.lang.Boolean solutionValue() {
-            assert solution != null;
-            return solution;
-        }
-
-        @Override
-        MPVariable mpVariable() {
-            return mpVariable;
         }
 
         @Override
@@ -87,26 +105,7 @@ public abstract class OptimiserVariable<T> {
         }
 
         @Override
-        synchronized void release() {
-            this.mpVariable.delete();
-        }
-
-        public void setInitial(boolean initial) {
-            this.initial = initial;
-        }
-
-        @Override
-        boolean hasInitial() {
-            return initial != null;
-        }
-
-        @Override
-        public void clearInitial() {
-            initial = null;
-        }
-
-        @Override
-        public double getInitial() {
+        public double hint() {
             assert hasInitial();
             if (initial) return 1.0;
             else return 0.0;
@@ -117,25 +116,11 @@ public abstract class OptimiserVariable<T> {
 
         private final double lowerBound;
         private final double upperBound;
-        private java.lang.Integer initial;
-        private java.lang.Integer solution;
-        private MPVariable mpVariable;
 
         public Integer(double lowerBound, double upperBound, String name) {
             super(name);
             this.lowerBound = lowerBound;
             this.upperBound = upperBound;
-        }
-
-        @Override
-        public java.lang.Integer solutionValue() {
-            assert solution != null;
-            return solution;
-        }
-
-        @Override
-        MPVariable mpVariable() {
-            return mpVariable;
         }
 
         @Override
@@ -149,26 +134,7 @@ public abstract class OptimiserVariable<T> {
         }
 
         @Override
-        synchronized void release() {
-            this.mpVariable.delete();
-        }
-
-        @Override
-        boolean hasInitial() {
-            return initial != null;
-        }
-
-        public void setInitial(int initial) {
-            this.initial = initial;
-        }
-
-        @Override
-        public void clearInitial() {
-            initial = null;
-        }
-
-        @Override
-        public double getInitial() {
+        public double hint() {
             assert hasInitial();
             return initial;
         }
--- a/traversal/planner/GraphPlanner.java
+++ b/traversal/planner/GraphPlanner.java
@@ -251,7 +251,7 @@ public class GraphPlanner implements Planner {
 
     @SuppressWarnings("NonAtomicOperationOnVolatileField")
     private void optimise(GraphManager graphMgr, boolean singleUse) {
-        computeTraversalCosts(graphMgr);
+        updateTraversalCosts(graphMgr);
         if (isUpToDate() && isOptimal()) {
             if (LOG.isDebugEnabled()) LOG.debug("GraphPlanner still optimal and up-to-date");
             return;
@@ -286,7 +286,7 @@ public class GraphPlanner implements Planner {
         if (LOG.isTraceEnabled()) LOG.trace(optimiser.toString());
     }
 
-    private void computeTraversalCosts(GraphManager graphMgr) {
+    private void updateTraversalCosts(GraphManager graphMgr) {
         if (snapshot < graphMgr.data().stats().snapshot()) {
             // TODO: we should not include the graph's uncommitted writes, but only the persisted counts in the costs
             snapshot = graphMgr.data().stats().snapshot();
--- a/traversal/planner/GraphPlanner.java
+++ b/traversal/planner/GraphPlanner.java
@@ -160,16 +160,16 @@ public class GraphPlanner implements Planner {
     }
 
     private void initialise() {
-        initialiseVariables();
-        initialiseConstraints();
+        createVariables();
+        createConstraints();
     }
 
-    private void initialiseVariables() {
-        vertices.values().forEach(PlannerVertex::initialiseVariables);
-        edges.forEach(PlannerEdge::initialiseVariables);
+    private void createVariables() {
+        vertices.values().forEach(PlannerVertex::createVariables);
+        edges.forEach(PlannerEdge::createVariables);
     }
 
-    private void initialiseConstraints() {
+    private void createConstraints() {
         String conPrefix = "planner_con_";
         for (int i = 0; i < vertices.size(); i++) {
             OptimiserConstraint conOneVertexAtOrderI = optimiser.constraint(1, 1, conPrefix + "one_vertex_at_order_" + i);
@@ -177,8 +177,8 @@ public class GraphPlanner implements Planner {
                 conOneVertexAtOrderI.setCoefficient(vertex.varOrderAssignment[i], 1);
             }
         }
-        vertices.values().forEach(PlannerVertex::initialiseConstraints);
-        edges.forEach(PlannerEdge::initialiseConstraints);
+        vertices.values().forEach(PlannerVertex::createConstraints);
+        edges.forEach(PlannerEdge::createConstraints);
     }
 
     @Override
@@ -398,8 +398,7 @@ public class GraphPlanner implements Planner {
         }
         assert vertexOrder == vertices.size();
 
-        vertices.values().forEach(PlannerVertex::inferInitialStartingVertexFromOrder);
-        vertices.values().forEach(PlannerVertex::setOutgoingEdgesInitialValues);
-        edges.forEach(PlannerEdge::setInitialMinimal);
+        vertices.values().forEach(PlannerVertex::initialise);
+        edges.forEach(PlannerEdge::initialise);
     }
 }
--- a/traversal/planner/GraphPlanner.java
+++ b/traversal/planner/GraphPlanner.java
@@ -231,7 +231,6 @@ public class GraphPlanner implements Planner {
         return optimiser;
     }
 
-    // FIXME rename (tryOptimise() is taken)
     void mayOptimise(GraphManager graphMgr, boolean singleUse) {
         if (procedure == null) {
             try {
@@ -272,10 +271,7 @@ public class GraphPlanner implements Planner {
         if (isError()) throwPlanningError();
         else assert isPlanned();
 
-        // TODO: occasionally optimiser produces worse results on repeat calls,
-        // we should be able to catch that and reuse the old procedure
         createProcedure();
-        LOG.debug(procedure.toString());
         end = Instant.now();
 
         isUpToDate = true;
@@ -314,14 +310,14 @@ public class GraphPlanner implements Planner {
             if (costChangeSignificant(e)) setOutOfDate();
         });
 
-        double vertexCost = vertices.values().stream().mapToDouble(PlannerVertex::getCost).sum();
-        double edgeCost = edges.stream().mapToDouble(PlannerEdge::getCost).sum();
+        double vertexCost = vertices.values().stream().mapToDouble(PlannerVertex::cost).sum();
+        double edgeCost = edges.stream().mapToDouble(PlannerEdge::cost).sum();
         totalCost = vertexCost + edgeCost;
         if (totalCostChangeSignificant()) setOutOfDate();
     }
 
     boolean costChangeSignificant(PlannerVertex<?> vertex) {
-        return costChangeSignificant(vertex.recordedCost, vertex.cost);
+        return costChangeSignificant(vertex.recordedCost, vertex.cost());
     }
 
     boolean costChangeSignificant(PlannerEdge<?, ?> edge) {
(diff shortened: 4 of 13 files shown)


common/optimiser/Optimiser.java
```diff
@@ -56,12 +57,16 @@ public Optimiser() {
...
+            status = Status.NOT_SOLVED;
+        }
         if (isOptimal()) return status;
         else if (!hasSolver) initialiseSolver();
         solver.setTimeLimit(timeLimitMillis);
         status = Status.of(solver.solve(parameters));
-        variables.forEach(OptimiserVariable::recordValue);
-        clearInitialisation();
+        if (status != Status.NOT_SOLVED && status != Status.ERROR)
+            variables.forEach(OptimiserVariable::recordValue);
```
dmitrii-ubskii, line 69: Only record values if the optimisation pass was successful, reuse initialisation as solution values otherwise.

common/optimiser/Optimiser.java
```diff
@@ -56,12 +57,16 @@ public Optimiser() {
     }
 
     public synchronized Status optimise(long timeLimitMillis) {
+        if (hasSolver && constraintsChanged()) {
+            constraints.forEach(constraint -> constraint.initialise(solver));
+            status = Status.NOT_SOLVED;
+        }
```
dmitrii-ubskii, line 63: New set of constraints depends on knowing the traversal costs, so the constraints may change along with the objective function coefficients.

common/optimiser/OptimiserVariable.java
```diff
@@ -26,27 +26,60 @@
...
+        return initial != null;
+    }
+
+    public T initial() {
+        assert hasInitial();
+        return initial;
+    }
 
-    public abstract void clearInitial();
+    public abstract double hint();
```
dmitrii-ubskii, line 70: We now have `initial()` to retrieve the native initial value, and `hint()` to convert it to a `double` for MPSolver.

database/RocksIterator.java
```diff
@@ -28,6 +28,7 @@
 import com.vaticle.typedb.core.graph.common.Storage.Key;
```
dmitrii-ubskii, line 28: RocksIterator & RocksStorage: statistics on the number and timings of DB interactions for internal use.

traversal/planner/GraphPlanner.java
```diff
@@ -107,55 +100,6 @@ static GraphPlanner create(Structure structure) {
...
-    }
-
-    private boolean isError() {
-        return optimiser.isError();
-    }
-
-    Optimiser optimiser() {
-        return optimiser;
-    }
-
```
dmitrii-ubskii, line 158: This looks like a bigger change than it is. All the creation routines are now pushed together and the accessors were moved further down in the file.

traversal/planner/GraphPlanner.java
```diff
@@ -218,134 +162,101 @@ private PlannerVertex.Type typeVertex(StructureVertex.Type structureVertex) {
...
     }
 
-    private void setInitialValues() {
-        new Initialiser().execute();
+    Optimiser optimiser() {
+        return optimiser;
     }
 
-    void mayOptimise(GraphManager graph, boolean singleUse) {
+    void mayOptimise(GraphManager graphMgr, boolean singleUse) {
```
dmitrii-ubskii, line 234: `GraphManager graph` -> `GraphManager graphMgr` everywhere, the planner seemed to be the only place where that naming convention was violated.

traversal/planner/GraphPlanner.java
```diff
@@ -218,134 +162,101 @@ private PlannerVertex.Type typeVertex(StructureVertex.Type structureVertex) {
...
     @SuppressWarnings("NonAtomicOperationOnVolatileField")
-    private void optimise(GraphManager graph, boolean singleUse) {
-        updateObjective(graph);
+    private void optimise(GraphManager graphMgr, boolean singleUse) {
+        computeTraversalCosts(graphMgr);
         if (isUpToDate() && isOptimal()) {
             if (LOG.isDebugEnabled()) LOG.debug("GraphPlanner still optimal and up-to-date");
             return;
         }
+        updateOptimiserCoefficients();
```
dmitrii-ubskii, line 259: We only want to change the ILP problem if we plan to run the solver.

traversal/planner/GraphPlanner.java
```diff
@@ -410,47 +385,34 @@ public String toString() {
         return str.toString();
     }
 
-    private class Initialiser {
```
dmitrii-ubskii, line 413: Initialiser acted as a method: the only use of it was `new Initialiser().execute()`, then the initialiser state was immediately discarded. It is explicitly a method now.

traversal/planner/GraphPlanner.java
```diff
@@ -410,47 +385,35 @@ public String toString() {
...
+        open.add(start);
+        int vertexOrder = 0;
+        while (!open.isEmpty()) {
+            PlannerVertex<?> v = open.poll();
+            assert !closedSet.contains(v);
+            closedSet.add(v);
+            v.setOrderInitial(vertexOrder++);
+            v.outs().stream().map(PlannerEdge.Directional::to).filter(x -> !closedSet.contains(x) && !open.contains(x))
+                    .forEach(open::add);
         }
```
dmitrii-ubskii, line 405: Greedy initialisation: start with the vertex with the smallest size. On each iteration, select the vertex from the open set with the smallest traversal cost, which is estimated as the minimum cost of all incoming edges. The neighbourhood of this vertex is then added to the open set for consideration. This is repeated until all vertices are in the closed set.

traversal/planner/PlannerEdge.java
```diff
@@ -129,47 +140,59 @@ public String toString() {
...
         private final Encoding.Direction.Edge direction;
-        private boolean hasInitialValue;
-        private boolean isInitialisedVariables;
-        private boolean isInitialisedConstraints;
-        private double costNext;
-        double costLastRecorded;
+
+        private boolean didInitialiseVariables;
+        private boolean didInitialiseConstraints;
+        private boolean hasInitialValues;
```
dmitrii-ubskii, line 153: Renamed + reordered to follow the initialisation order.

traversal/planner/PlannerEdge.java
```diff
@@ -129,47 +140,59 @@ public String toString() {
...
+
+        private final int tieBreaker;
+        private static int nextTieBreaker = 0;
```
dmitrii-ubskii, line 156: Needed so that there can be only one minimal incoming edge into a vertex at a time.

traversal/planner/PlannerEdge.java
```diff
@@ -129,47 +140,59 @@ public String toString() {
...
-            this.isInitialisedConstraints = false;
+            this.recordedCost = INIT_ZERO;
+
+            this.didInitialiseVariables = false;
+            this.didInitialiseConstraints = false;
+            this.hasInitialValues = false;
+
             this.varPrefix = "edge_var_" + this.toString() + "_";
             this.conPrefix = "edge_con_" + this.toString() + "_";
+            this.tieBreaker = nextTieBreaker++;
```
dmitrii-ubskii, line 175: Integer overflow is safe in Java as far as I can tell.

traversal/planner/PlannerEdge.java
```diff
@@ -1053,26 +1048,24 @@ private Forward(PlannerVertex.Thing from, PlannerVertex.Thing to) {
...
+                                double div = graphMgr.data().stats().thingVertexCount(owner);
+                                if (div > 0) {
+                                    cost += graphMgr.data().stats().hasEdgeSum(owner, attTypes) / div;
+                                }
                             }
+                            assert !ownerTypes.isEmpty();
+                            cost /= ownerTypes.size();
                         }
-                        assert !ownerTypes.isEmpty();
-                        cost /= ownerTypes.size();
```
dmitrii-ubskii, line 1073: Indented into the `else` branch, no actual algorithm changes.

traversal/planner/PlannerEdge.java
```diff
@@ -1083,26 +1076,24 @@ private Backward(PlannerVertex.Thing from, PlannerVertex.Thing to) {
...
+                                double div = graphMgr.data().stats().thingVertexCount(attr);
+                                if (div > 0) {
+                                    cost += graphMgr.data().stats().hasEdgeSum(ownerTypes, attr) / div;
+                                }
                             }
+                            assert !attTypes.isEmpty();
+                            cost /= attTypes.size();
                         }
-                        assert !attTypes.isEmpty();
-                        cost /= attTypes.size();
```
dmitrii-ubskii, line 1103: Ditto

traversal/planner/PlannerVertex.java
```diff
@@ -225,15 +228,15 @@ protected Properties.Thing newProperties() {
...
-                setObjectiveCoefficient(1);
+                cost = 1;
             } else {
                 assert !props().types().isEmpty();
                 if (iterate(props().predicates()).anyMatch(p -> p.operator().equals(EQ))) {
-                    setObjectiveCoefficient(props().types().size());
+                    cost = props().types().size();
                 } else {
-                    setObjectiveCoefficient(graph.data().stats().thingVertexSum(props().types()));
+                    cost = graphMgr.data().stats().thingVertexSum(props().types());
```
dmitrii-ubskii, line 239: As in `PlannerEdge`, replace `setObjectiveCoefficient()` calls with directly setting `cost`.

traversal/procedure/CombinationProcedure.java
```diff
@@ -161,22 +161,22 @@ private void registerLoopEdges(ProcedureVertex.Type from, Set<StructureEdge<?, ?
 
     private ProcedureEdge<?, ?> createOut(ProcedureVertex.Type from, ProcedureVertex.Type to,
                                           StructureEdge<?, ?> structureEdge) {
-        ProcedureEdge<?, ?> edge = ProcedureEdge.of(from, to, structureEdge, -1, true);
+        ProcedureEdge<?, ?> edge = ProcedureEdge.of(from, to, structureEdge, true);
```
dmitrii-ubskii, line 164: Procedure edges no longer have ordering associated with them.

traversal/procedure/GraphProcedure.java
```diff
@@ -39,23 +38,23 @@
...
 import static com.vaticle.typedb.core.concurrent.producer.Producers.async;
+import static java.util.Comparator.comparing;
 
 public class GraphProcedure implements PermutationProcedure {
 
     private static final Logger LOG = LoggerFactory.getLogger(GraphProcedure.class);
 
     private final ProcedureVertex<?, ?>[] vertices;
-    private ProcedureVertex<?, ?> startVertex;
+    private Set<ProcedureVertex<?, ?>> startVertices;
```
dmitrii-ubskii, line 57: More than one!

traversal/procedure/GraphProcedure.java
```diff
@@ -67,19 +66,23 @@ public static GraphProcedure create(GraphPlanner planner) {
         Set<PlannerVertex<?>> registeredVertices = new HashSet<>();
         Set<PlannerEdge.Directional<?, ?>> registeredEdges = new HashSet<>();
         planner.vertices().forEach(vertex -> builder.registerVertex(vertex, registeredVertices, registeredEdges));
-        return builder.orderAndbuild();
+        builder.streamlineVertices();
```
dmitrii-ubskii, line 69: To avoid zigzagging traversal, see L192 for details.

traversal/procedure/GraphProcedure.java
```diff
@@ -181,39 +184,39 @@ public Builder() {
...
+            while (!open.empty()) {
+                ProcedureVertex<?, ?> vertex = open.pop();
+                vertex.setOrder(vertexOrder++);
+                closed.add(vertex);
+                vertex.outs().stream().map(ProcedureEdge::to).distinct().sorted(comparing(v -> -v.ins().size())).filter(
+                        v -> v.ins().stream().map(ProcedureEdge::from).allMatch(closed::contains)
+                ).forEach(open::add);
             }
-            return orderedVertices;
+            assert vertexOrder == vertices.size();
```
dmitrii-ubskii, line 206: The optimiser ultimately only cares about the directions of the edges in the plan. `0 -> 2 -> 4 <- 3 <- 1` and `0 -> 1 -> 4 <- 3 <- 2` are therefore equivalent plans, which lead to inefficient traversal. 
This procedure is meant to rectify that by essentially performing a depth-first topological sort on the vertices. For the top vertex in the `open` stack, we put on the `open` stack that part of its neighbourhood that already has its requirements fulfilled, i.e. all the immediately preceding vertices are already present in the `closed` set. These newly considered vertices are sorted such that the vertex with the greatest amount of incoming edges ends up on the top, so that during traversal we can start backtracking earlier from the intersection vertices without having to consider the other branches.
This algorithm can be improved, but it is a simple and reasonably effective first approximation.

traversal/procedure/GraphProcedure.java
```diff
@@ -67,19 +66,23 @@ public static GraphProcedure create(GraphPlanner planner) {
...
         return vertices;
     }
 
-    public ProcedureVertex<?, ?> startVertex() {
-        if (startVertex == null) {
-            startVertex = iterate(vertices()).filter(ProcedureVertex::isStartingVertex)
-                    .first().orElseThrow(() -> TypeDBException.of(ILLEGAL_STATE));
+    public ProcedureVertex<?, ?> initialVertex() {
+        return vertices[0];
+    }
```
dmitrii-ubskii, line 79: `vertices` are now ordered.

traversal/procedure/GraphProcedure.java
```diff
@@ -324,167 +327,167 @@ public ProcedureVertex.Thing setPredicate(ProcedureVertex.Thing thing, Predicate
         public ProcedureEdge.Native.Type.Sub.Forward forwardSub(
                 ProcedureVertex.Type child, ProcedureVertex.Type parent, boolean isTransitive) {
             ProcedureEdge.Native.Type.Sub.Forward edge =
-                    new ProcedureEdge.Native.Type.Sub.Forward(child, parent, -1, isTransitive);
+                    new ProcedureEdge.Native.Type.Sub.Forward(child, parent, isTransitive);
```
dmitrii-ubskii, line 330: From here to the end of the file: procedure edges no longer take an `order` parameter.

traversal/procedure/ProcedureEdge.java
```diff
@@ -79,25 +79,22 @@
         VERTEX_FROM extends ProcedureVertex<?, ?>, VERTEX_TO extends ProcedureVertex<?, ?>
         > extends TraversalEdge<VERTEX_FROM, VERTEX_TO> {
 
-    private final int order;
```
dmitrii-ubskii, line 82: Most of the changes here are just removing `order`.

traversal/procedure/ProcedureVertex.java
```diff
@@ -63,7 +63,6 @@
         > extends TraversalVertex<ProcedureEdge<?, ?>, PROPERTIES> {
```
dmitrii-ubskii, line 63: Removing crutches that were needed during transition from edge-ordered to vertex-ordered plan.

traversal/procedure/ProcedureVertex.java
```diff
@@ -167,7 +151,6 @@ public Thing asThing() {
 
         @Override
         public Forwardable<? extends ThingVertex, Order.Asc> iterator(GraphManager graphMgr, Traversal.Parameters parameters) {
-            assert isStartingVertex();
```
dmitrii-ubskii, line 170: This precondition was only satisfied accidentally. While it is true that GraphProcedure / GraphIterator should never call this on a non-starting vertex, CombinationProcedure also uses this class, and it does not assign any kind of ordering.

traversal/procedure/ProcedureVertex.java
```diff
@@ -321,7 +304,7 @@ protected Properties.Type newProperties() {
 
         @Override
         public Forwardable<? extends TypeVertex, Order.Asc> iterator(GraphManager graphMgr, Traversal.Parameters parameters) {
-            assert isStartingVertex() && id().isVariable();
+            assert id().isVariable();
```
dmitrii-ubskii, line 307: Ditto.

traversal/procedure/VertexProcedure.java
```diff
@@ -54,7 +54,7 @@ public static VertexProcedure create(StructureVertex<?> structureVertex) {
         if (procedureVertex.isType()) procedureVertex.asType().props(structureVertex.asType().props());
         else procedureVertex.asThing().props(structureVertex.asThing().props());
         for (StructureEdge<?, ?> structureEdge : structureVertex.loops()) {
-            ProcedureEdge<?, ?> edge = ProcedureEdge.of(procedureVertex, procedureVertex, structureEdge, -1, true);
+            ProcedureEdge<?, ?> edge = ProcedureEdge.of(procedureVertex, procedureVertex, structureEdge, true);
```
dmitrii-ubskii, line 57: No more `order` on edges.

common/optimiser/Optimiser.java
```diff
@@ -46,6 +46,7 @@ public class Optimiser {
     private MPSolverParameters parameters;
     private Status status;
     private boolean hasSolver;
+    private double objectiveValueCache;
```
flyingsilverfin, line 49: I think we should refactor this to be `objectiveValue`, and always record the solver's objective value into it every time. It looks a bit odd to have a variable called `cache` (caches tend to be things we can remove and still maintain correctness) - but if it is always used that would make sense.

common/optimiser/Optimiser.java
flyingsilverfin, line 69: we can turn this into a method called `recordValues` which will record the variable values and the objective value together

common/optimiser/Optimiser.java
```diff
@@ -78,8 +83,12 @@ public boolean isError() {
         return status == Status.ERROR;
     }
 
+    private boolean constraintsChanged() {
+        return constraints.stream().anyMatch(c -> !c.isInitialised);
+    }
+
     private void initialiseSolver() {
-        solver = MPSolver.createSolver("SCIP");
+        solver = MPSolver.createSolver("SAT");
```
flyingsilverfin, line 91: worth a comment on why/what this change is!!

Also worth mentionining in the PR description's second section

database/RocksIterator.java
```diff
@@ -44,6 +45,8 @@
     private boolean isClosed;
     org.rocksdb.RocksIterator internalRocksIterator;
 
+    public static Stats stats = new Stats();
```
flyingsilverfin, line 48: i had a chat with ganesh and we think we should delete these objects and usages for now until we can implement it properly with per-query context to log the statistics against a particular query plan with

common/optimiser/Optimiser.java
```diff
@@ -56,12 +57,16 @@ public Optimiser() {
...
+            constraints.forEach(constraint -> constraint.initialise(solver));
+            status = Status.NOT_SOLVED;
+        }
         if (isOptimal()) return status;
         else if (!hasSolver) initialiseSolver();
         solver.setTimeLimit(timeLimitMillis);
         status = Status.of(solver.solve(parameters));
-        variables.forEach(OptimiserVariable::recordValue);
-        clearInitialisation();
+        if (status != Status.NOT_SOLVED && status != Status.ERROR)
```
flyingsilverfin, line 68: so it's our style to use only one of these two styles:
```
if (condition) single-line-operation;
```
or

```
if (condition) {
  cant fit in the same line as the condition operation
}
```
for loops we always use `{}`

I noticed this in a couple places in this PR that we can bring to consistency with the rest of the code base

flyingsilverfin (changes requested):
I've done some syntactic/style comments, still working through understanding how you implemented the functionality changes to the query planner (this will take some time)
have been pulled off this for a today though! will leave another review soon

common/optimiser/Optimiser.java
dmitrii-ubskii: You're right!
We switch from a mixed integer programming solver to an integer linear programming solver that is optimised around the variables and coefficients being integers.

traversal/planner/GraphPlanner.java
```diff
@@ -368,6 +279,70 @@ private void optimise(GraphManager graph, boolean singleUse) {
...
+    }
+
+    boolean costChangeSignificant(double costPrevious, double costNext) {
+        assert recordedTotalCost > 0;
+        assert costPrevious > 0;
+        assert costNext > 0;
+
+        return (costNext / costPrevious >= OBJECTIVE_VARIABLE_COST_MAX_CHANGE ||
+                costNext / costPrevious <= 1.0 / OBJECTIVE_VARIABLE_COST_MAX_CHANGE) &&
+                abs(costNext - costPrevious) / recordedTotalCost >= OBJECTIVE_VARIABLE_TO_PLANNER_COST_MIN_CHANGE;
```
flyingsilverfin, line 338: this is actually a pretty significant fix (?) compared the previous check we had implemented? Definitely worth a comment PR if so!

traversal/planner/PlannerVertex.java
```diff
@@ -41,43 +43,62 @@
...
     PlannerVertex(Identifier identifier, @Nullable GraphPlanner planner) {
         super(identifier);
         this.planner = planner;
-        isInitialisedVariables = false;
-        isInitialisedConstraints = false;
-        costLastRecorded = 0.01; // non-zero value for safe division
+        didInitialiseVariables = false;
+        didInitialiseConstraints = false;
+        didInitialiseValues = false;
+        recordedCost = 0.01; // non-zero value for safe division
```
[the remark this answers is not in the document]
flyingsilverfin: maybe that lets use remove the `max()` method that the `cost()` does internally?

common/optimiser/Optimiser.java
```diff
@@ -78,8 +87,12 @@ public boolean isError() {
         return status == Status.ERROR;
     }
 
+    private boolean constraintsChanged() {
+        return constraints.stream().anyMatch(c -> !c.isInitialised);
```
flyingsilverfin, line 91: something here is odd I will look into it after lunch in the next part of the review!

common/optimiser/OptimiserConstraint.java
```diff
@@ -31,24 +31,35 @@ public class OptimiserConstraint {
...
+            }
+        }
         coefficients.put(variable, coeff);
     }
 
     synchronized void initialise(MPSolver solver) {
-        this.mpConstraint = solver.makeConstraint(lowerBound, upperBound, name);
+        if (this.mpConstraint == null) this.mpConstraint = solver.makeConstraint(lowerBound, upperBound, name);
         coefficients.forEach((var, coeff) -> mpConstraint.setCoefficient(var.mpVariable(), coeff));
+        isInitialised = true;
```
flyingsilverfin, line 58: I think you're using the `isInitialised` to do too many things here (and this is compounded by the naming). If we rename `isInitialised` to `coefficientsChanged` or `coefficientsApplied` or `coefficientsSet`, we realise that the `setCoefficient` method can be simplified massively to:
```

    public void setCoefficient(OptimiserVariable<?> variable, double coeff) {
        if (mpConstraint != null) coefficientsApplied = coefficients.get(variable) == coeff;
        coefficients.put(variable, coeff);
    }
```

The new naming also indicates that we should set it to `false` when we `release()` if that doesn't affect the logical integrity.

(note that the `.get()` will return a null if it doesn't exist which will not match the value you pass in anyway)

This way, we use one signal (`mpConstraint` being null or not) to indicate whether we've initialised with the previous set of constraints, and then the `coefficientsApplied` to record whether active constraint contains any outdated coefficients

traversal/planner/PlannerEdge.java
```diff
@@ -96,28 +98,37 @@ void initialiseVariables() {
...
 
-    void resetInitialValue() {
-        forward.resetInitialValue();
-        backward.resetInitialValue();
+    void resetInitialValues() {
+        forward.resetInitialValues();
+        backward.resetInitialValues();
+    }
+
+    public void initialiseMinimal() {
```
flyingsilverfin, line 129: this should also be called `setInitialMinimal` i suppose!

traversal/planner/PlannerEdge.java
flyingsilverfin: let's say the DB is around for a loooong time and doesn't get restarted - Java doesn't do overflows per say but it's still awkward to permit them. I'd say we should still promote this to a `long`!
Also this currently isn't thread safe: two DirectionalEdges that are created in parallel can have the same tie breaker (we don't currently do this within the same query, but it's possible). Because it's static we should use an `AtomicLong`, or choose a different strategy...
choosing a random number still has a chance of collision huh? Same with a hash...

The other option is that we can create the `PlannerEdge` with an ID (long) we generate when creating the `GraphPlanner`, which it can pass into the `Directional` edges as a disambiguator.
This is probably actually the cleanest solution, what do you think?

traversal/procedure/GraphProcedure.java
```diff
@@ -181,39 +184,39 @@ public Builder() {
...
-         * we need to remove it once the query planner natively orders vertices instead of edges
-         */
-        private ProcedureVertex<?, ?>[] orderedVertices() {
-            ProcedureVertex<?, ?>[] orderedVertices = new ProcedureVertex[vertices.size()];
-            List<ProcedureVertex<?, ?>> vertexList = iterate(vertices.values()).toList();
-            vertexList.sort(Comparator.comparing(v -> v.lastInEdge() == null ? 0 : v.lastInEdge().order()));
-            for (int i = 0; i < vertexList.size(); i++) {
-                orderedVertices[i] = vertexList.get(i);
-                orderedVertices[i].setOrder(i);
+        private void streamlineVertices() {
```
flyingsilverfin, line 192: so interestingly, the traversal engine will already be able to backtrack through the zigzag in an optimal way.
Additionally, this kind of "push towards DFS" operation feels like a planning task, rather than a procedure creation task - notice that because we have this, we still need the `setOrder()` method on the vertex, which ideally we'd get rid of and make the `ProcedureVertex` less stateful.

Given all that I'm inclined to remove it from here for now. We can then choose convert it into a planning task that will allow us to construct the streamlined procedure from the plan directly and delete the capability from the traversal engine instead if we deem it more correct.

traversal/procedure/GraphProcedure.java
```diff
@@ -181,39 +184,39 @@ public Builder() {
...
+            Stack<ProcedureVertex<?, ?>> open = new Stack<>();
+            vertices.values().stream().filter(ProcedureVertex::isStartingVertex).forEach(open::add);
+            Set<ProcedureVertex<?, ?>> closed = new HashSet<>();
+
+            int vertexOrder = 0;
+            while (!open.empty()) {
+                ProcedureVertex<?, ?> vertex = open.pop();
+                vertex.setOrder(vertexOrder++);
```
flyingsilverfin, line 200: let's aim to have this method deleted and make `ProcedureVertex` more stateless

traversal/procedure/ProcedureEdge.java
```diff
@@ -1114,15 +1131,20 @@ public Identifier.Variable scope() {
                     else return to.id().asVariable();
                 }
 
+                @Override
+                public boolean equals(Object o) {
+                    return super.equals(o) && ((ProcedureEdge.Native.Thing.RolePlayer) o).roleTypes.equals(roleTypes);
```
flyingsilverfin, line 1136: note: you can do this in all the `equals()` methods you changed above ^^

flyingsilverfin (changes requested):

traversal/planner/GraphPlanner.java
```diff
@@ -410,47 +386,35 @@ public String toString() {
...
-    private class Initialiser {
-
-        private final LinkedHashSet<PlannerVertex<?>> queue;
-        private int edgeCount;
-
-        private Initialiser() {
-            queue = new LinkedHashSet<>();
-            edgeCount = 0;
+    private void setInitialValues() {
+        resetInitialValues();
```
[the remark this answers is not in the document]
dmitrii-ubskii: I think even if we do, this is unnecessary as they just get overwritten anyway.

traversal/planner/PlannerVertex.java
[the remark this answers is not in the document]
dmitrii-ubskii: No, `cost` is set to zero so that we don't have to bother with `INIT_ZERO` until we access it. You'll notice that `cost` is never accessed directly, only through `cost()`. `max()` stays because of the way we determine whether the model is still applicable to the updated costs, which involves division between latest and recorded costs.

common/optimiser/Optimiser.java
```diff
@@ -56,16 +57,24 @@ public Optimiser() {
...
-        variables.forEach(OptimiserVariable::recordValue);
-        clearInitialisation();
+        recordValues();
         if (isOptimal()) releaseSolver();
         return status;
     }
 
+    private void recordValues() {
+        objectiveValue = solver.objective().value();
+        if (status != Status.NOT_SOLVED && status != Status.ERROR) variables.forEach(OptimiserVariable::recordValue);
```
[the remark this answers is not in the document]
dmitrii-ubskii: If the optimiser leaves the model in an unsolved state, i.e. the values of the variables do not satisfy the constraints, we do not want to overwrite their previously recorded values. That is also part of the reason `OptimiserVariable::value()` was introduced.

traversal/planner/PlannerEdge.java
dmitrii-ubskii: We only care that the tie breakers are distinct within a single planner's set of edges, and given that the planner is not threaded, there is no danger of collisions. That said, AtomicLong seems to be a cleaner solution overall, with only a tiny and extremely rare penalty during the planner setup.
I don't think involving the planner in determining the tie breakers—or ids, or whatever we call them—is a good idea, because that exposes an implementation detail that the planner really shouldn't know anything about.

traversal/planner/PlannerVertex.java
```diff
@@ -41,43 +43,62 @@
...
+
+    public boolean didInitialiseConstraints() {
+        return didInitialiseConstraints;
     }
 
-    public boolean isInitialisedVariables() {
-        return isInitialisedVariables;
+    public boolean didInitialiseValues() {
+        return didInitialiseValues;
     }
```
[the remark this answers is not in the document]
dmitrii-ubskii: First time for everything. Maybe we should start using `did...`, `can...`, and `has...` in addition to `is...`? That reads so much better than `isVerbed...` or `isHas...`.
That said, I would actually like to get rid of these flags as much as I can by restructuring the planner construction a bit.

traversal/planner/PlannerVertex.java
```diff
@@ -104,93 +125,70 @@ void loop(PlannerEdge<?, ?> edge) {
...
-        assert !Double.isNaN(cost);
-        if (cost < INIT_ZERO) cost = INIT_ZERO;
-        double exp = planner.edges().size() * planner.costExponentUnit;
-        double coeff = cost * Math.pow(planner.branchingFactor, exp);
-        planner.optimiser().setObjectiveCoefficient(varIsStartingVertex, coeff);
-        costNext = cost;
-        planner.updateCostNext(costLastRecorded, costNext);
+    protected void updateOptimiserCoefficients() {
+        assert !Double.isNaN(recordedCost);
+        planner.optimiser().setObjectiveCoefficient(varIsStartingVertex, log(1 + recordedCost));
```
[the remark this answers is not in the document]
dmitrii-ubskii: That makes a lot of sense, I think you're right!

traversal/planner/GraphPlanner.java
```diff
@@ -410,47 +386,35 @@ public String toString() {
...
+
+        PlannerVertex<?> start = vertices.values().stream().min(comparing(PlannerVertex::cost)).get();
+        Set<PlannerVertex<?>> closedSet = new HashSet<>();
+        PriorityQueue<PlannerVertex<?>> open = new PriorityQueue<>(comparing(
+                v -> v.ins().stream().filter(e -> closedSet.contains(e.from())).mapToDouble(PlannerEdge.Directional::cost).min().orElse(v.cost())
```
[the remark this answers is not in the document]
dmitrii-ubskii: As far as I can tell, `PriorityQueue` doesn't recalculate the priority once the element was inserted.
I don't think this is that big of a problem, we just want the best value we can get at the time we add the vertex to the queue. Just using the vertex's `cost()` is not great, because we will end up just sorting the vertices without considering the edges at all. This could be rewritten without using a priority queue at all, and I think I just will.

traversal/procedure/GraphProcedure.java
dmitrii-ubskii: `ProcedureVertex::order()` is used in a ton of places in `GraphIterator`, I think that change might be better suited for its own PR.

traversal/procedure/GraphProcedure.java
dmitrii-ubskii: That wasn't the impression I got from the traversal engine, but you probably know better. Removed.

common/optimiser/Optimiser.java
```diff
@@ -56,16 +57,24 @@ public Optimiser() {
     }
 
     public synchronized Status optimise(long timeLimitMillis) {
+        if (hasSolver && constraintsChanged()) {
+            constraints.forEach(constraint -> constraint.initialise(solver));
```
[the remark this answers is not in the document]
dmitrii-ubskii: To be clear, the constraint is what is initialised for a given solver, not the solver itself. Renamed to `updateCoefficients()` and `isInitialised` -> `isUpToDate`.

traversal/planner/GraphPlanner.java
dmitrii-ubskii: I guess? The logic is that if doubling a cost is a significant change, then so is halving.

traversal/planner/GraphPlanner.java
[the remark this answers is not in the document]
flyingsilverfin: Just responding to: "As far as I can tell, PriorityQueue doesn't recalculate the priority once the element was inserted."

That may be the case in the specific implementation used, but we shouldn't generally assume how a class is implementing its internal state. Additionally, that assumption isn't actually in line with my expection of how a priority queue might be implemented (when adding new elements to the queue, you have to re-evaluate the comparator to place elements correctly).

The refactored version without the `PriorityQueue` looks good!!

flyingsilverfin (changes requested):

common/optimiser/Optimiser.java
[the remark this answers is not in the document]
dmitrii-ubskii: Pretty much just that. Updating the constraints can put the existing solution in an invalid state. Now, I've never actually encountered a problem with that, but it's not inconceivable that on a bigger query the solver might never reach a valid state.

common/optimiser/OptimiserVariable.java
```diff
@@ -26,27 +26,56 @@
...
     }
 
-    public abstract T solutionValue();
+    public T value() {
+        if (hasSolutionValue()) return solutionValue();
+        else return initial();
+    }
 
-    abstract MPVariable mpVariable();
+    public boolean hasSolutionValue() {
```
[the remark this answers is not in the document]
dmitrii-ubskii: All the methods that are package protected are used elsewhere, usually by Optimiser, and the public ones are accessors that, even if they aren't used elsewhere, it doesn't make sense to restrict access to. I think we're good.

traversal/planner/PlannerVertex.java
```diff
@@ -104,93 +113,57 @@ void loop(PlannerEdge<?, ?> edge) {
...
 
-    void setStartingVertexInitial() {
-        varIsStartingVertex.setInitial(true);
-        varIsEndingVertex.setInitial(false);
-        varHasIncomingEdges.setInitial(false);
+    void inferInitialStartingVertexFromOrder() {
+        assert varOrderNumber.hasInitial();
+        double initialOrder = varOrderNumber.initial();
+        varIsStartingVertex.setInitial(iterate(outs()).allMatch(e -> e.to().varOrderNumber.initial() > initialOrder));
+        isInitialised = true;
```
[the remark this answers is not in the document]
dmitrii-ubskii: Re: edit, I think I'm also going to rename the `initialiseVariables/Constraints` method to remove ambiguity.

traversal/planner/PlannerVertex.java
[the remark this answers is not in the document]
dmitrii-ubskii: It might read better if the accessor is called smth like `safeCost()` to emphasize the difference. It can definitely be confusing as is.

traversal/planner/GraphPlanner.java
```diff
@@ -217,135 +160,102 @@ private PlannerVertex.Type typeVertex(StructureVertex.Type structureVertex) {
     }
 
     private void initialise() {
-        initialiseVariables();
-        initialiseConstraintsForVariables();
-        initialiseConstraintsForEdges();
+        createVariables();
+        createConstraints();
```
flyingsilverfin, line 164: To connec the storyline even more fully we could call these `createOptimiserVariables()` and `createOptimiserConstraints()`

Then we've definitely set on the vocabulary:
1. `Planner` objects are our own objects 
2. `Optimiser` is the linear optimiser
3. we `register` planner vertices/edges, then `create` optimiser variables and constraints
4. finally, we greedily `initialise` the `Optimiser` variables with a valid solution.

by that logic `setInitialValues` should become `initialiseOptimiserValues` and `PlannerVertex/Edge.initialise()` should become`initialiseOptimiserValues()` as well!

flyingsilverfin:
Update: we consider whether our cost function can also naturally be used to consider disconnected graphs.

```
0   2
|   |
1   3
```

In this mode, we don't have to do anything differently:
```
Π{v ∈ S} (1 + α(v)) * Π{v ∈ V}(1 + min({β(x, v): edge(x, v) is selected}))
```

the product of the `α(v)` already captures multiple possible starting points, and calculates cost according to our nested-loop formulation. The cost function however cannot capture the difference between executing the ordered plan: `0 - 1 - 2 - 3` versus `0 - 2 - 1 - 3`. We allow the traversal engine to perform the required short circuiting to make the two plans roughly equivalent in performance.

PR #6594, opened 2022-06-22
Improve negative-cycle detection during rule-validation

### LINKED ISSUE #6500: Trying to force rule order gives non-deterministic results or hangs the database.
## Description

The first setup never finishes, the second gives sometimes two results, sometimes three.

## Environment

1. OS (where TypeDB server runs): Ubuntu 20.04
2. TypeDB version (and platform): TypeDB 2.5.0
3. TypeDB client: client-python

## Reproducible Steps
The text files contain a define section, an insert section and a query section.
[never_finishes.txt](https://github.com/vaticle/typedb/files/7746775/never_finishes.txt)
[non_deterministic_result_set.txt](https://github.com/vaticle/typedb/files/7746777/non_deterministic_result_set.txt)

I ran & generated those in Python (see additional information)
```
options = TypeDBOptions.core()
options.explain = True
options.infer = True
with session.transaction(TransactionType.READ, options) as transaction:
```

## Expected Output
First query:
Some output, or at least a way to interrupt the database, but SIGTERM doesn't seem to work.

Second query:
A stable size of the result set.

## Actual Output

First query:
It hangs, pegged at 100% CPU, stable memory usage, ignores SIGTERM.

Second query:

Sometimes two results, sometimes three results.
The one that goes missing 50% of the time is the `door` `common room` combination.
 
## Additional information

Full Python code for the first query: [main.txt](https://github.com/vaticle/typedb/files/7746841/main.txt)
Full python code for second query: [door.txt](https://github.com/vaticle/typedb/files/7746837/door.txt)

### LINKED ISSUE #6538: Incorrect validation for negated attribute values in rules (invalidates valid rules)
## Description

The following rule is invalidated because of a cyclical negation, but the negation is on a different attribute value in the antecedent to the one inferred by the consequent.

```
rule if-data-source-not-trusted-then-untrusted:
    when {
        $ds isa data_source;
        not { $ds has is_trusted true;};
    }
    then {
        $ds has is_trusted false;
    };
```

## Environment

1. Windows or Ubuntu
2. TypeDB 2.8.0

## Reproducible Steps

1. Load the following schema:

```
define

is_trusted sub attribute, value boolean;
data_source sub entity, owns is_trusted;

rule if-data-source-not-trusted-then-untrusted:
    when {
        $ds isa data_source;
        not { $ds has is_trusted true;};
    }
    then {
        $ds has is_trusted false;
    };
```

## Expected Output

A successful schema commit

## Actual Output

Validation error:

```
Invalid Rule Write: A cycle containing negation(s) that can cause inference contradictions has been detected in rules:
if-data-source-not-trusted-then-untrusted ->
if-data-source-not-trusted-then-untrusted
```
 
## Additional information
If the rule contains  a more general not { $ds has is_trusted $it;}; in the antecedent, then validation should still fail.

### LINKED ISSUE #6599: Rule validation wrongly flags cases without negative cycles
Please replace every line in curly brackets { like this } with an appropriate answer, and remove this line.

## Description

{ Please provide a clear and concise description of the bug. }

## Environment

1. OS (where TypeDB server runs): Mac OS 12.4
2. TypeDB version (and platform): 2.11
3. TypeDB client: N/A, server bug.
4. Other environment details:

## Reproducible Steps
```
  define
  concept sub entity,
    plays binary:first,
    plays binary:second;

    binary sub relation, relates first, relates second;
    road sub binary;
    reachable sub binary;

    roadblock sub binary;
    sleeping-snorlax-on-route sub binary;

    rule reachable-is-transitive:
    when{
      (first: $a, second: $b) isa road;
      (first: $b, second: $c) isa reachable;
      not {(first: $a, second: $b) isa roadblock; };
    } then {
      (first: $a, second: $c) isa reachable;
    };

    rule snorlax-blocks-roads:
    when{
      (first: $a, second: $b) isa sleeping-snorlax-on-route;
    }then{
      (first: $a, second: $b) isa roadblock;
    };
```

## Expected Output
Passes validation

## Actual Output
Fails validation

krishnangovindraj:
The part of rule validation which prevents negative cycles (validateRulesThroughNegativeCycles) currently:
1.  Hangs on rule-commit in certain situations because of an infinite loop (#6500)
2. Falsely flags rules without negative cycles (#6599)

This PR fixes both.

When traversing the dependency tree, the set of visited nodes was tracked using the valueSet of a back-pointer map. Duplicate keys caused an overwrite of certain nodes, allowing infinite looping. This is fixed by using the keyset.

Any cycle involving a rule containing negated concludables was flagged. This was narrowed down to only flag rules where the cycle starts at a negated concludable.

- fixes #6500 
- fixes #6599

### DIFF
--- a/logic/LogicManager.java
+++ b/logic/LogicManager.java
@@ -130,26 +130,31 @@ public class LogicManager {
     }
 
     private void validateCyclesThroughNegations(ConceptManager conceptMgr, LogicManager logicMgr) {
+        // Problem: When a derivation of an inferred concept contains its own negation.
+        // Tracking this at runtime is impractical - statically detect & disallow.
+        // Detect: A cycle which passes through atleast one negation.
+        //      Equivalently - a cycle starting from a negated concept in a rule.
         Set<Rule> negationRulesTriggeringRules = logicMgr.rulesWithNegations()
                 .filter(rule -> !rule.condition().negatedConcludablesTriggeringRules(conceptMgr, logicMgr).isEmpty())
                 .toSet();
 
         for (Rule negationRule : negationRulesTriggeringRules) {
             Map<Rule, RuleDependency> visitedDependentRules = new HashMap<>();
-            visitedDependentRules.put(negationRule, RuleDependency.of(negationRule, null));
-            List<RuleDependency> frontier = new LinkedList<>(ruleDependencies(negationRule, conceptMgr, logicMgr));
-            RuleDependency dependency;
+            LinkedList<RuleDependency> frontier = new LinkedList<>(negatedRuleDependencies(negationRule, conceptMgr, logicMgr));
             while (!frontier.isEmpty()) {
-                dependency = frontier.remove(0);
+                RuleDependency dependency = frontier.removeFirst();
+                visitedDependentRules.put(dependency.recursiveRule, dependency);
                 if (negationRule.equals(dependency.recursiveRule)) {
                     List<Rule> cycle = findCycle(dependency, visitedDependentRules);
                     String readableCycle = cycle.stream().map(Rule::getLabel).collect(Collectors.joining(" -> \n", "\n", "\n"));
                     throw TypeDBException.of(CONTRADICTORY_RULE_CYCLE, readableCycle);
                 } else {
-                    visitedDependentRules.put(dependency.recursiveRule, dependency);
                     Set<RuleDependency> recursive = ruleDependencies(dependency.recursiveRule, conceptMgr, logicMgr);
-                    recursive.removeAll(visitedDependentRules.values());
-                    frontier.addAll(recursive);
+                    iterate(recursive)
+                        .filter(rule -> !visitedDependentRules.containsKey(rule.recursiveRule))
+                        .forEachRemaining( ruleDependency -> {
+                            frontier.add(ruleDependency);
+                        });
                 }
             }
         }
@@ -158,6 +163,12 @@ public class LogicManager {
     private Set<RuleDependency> ruleDependencies(Rule rule, ConceptManager conceptMgr, LogicManager logicMgr) {
         return link(iterate(rule.condition().concludablesTriggeringRules(conceptMgr, logicMgr)),
                     iterate(rule.condition().negatedConcludablesTriggeringRules(conceptMgr, logicMgr)))
+                .flatMap(concludable -> concludable.getApplicableRules(conceptMgr, logicMgr))
+                .map(recursiveRule -> RuleDependency.of(recursiveRule, rule)).toSet();
+    }
+
+    private Set<RuleDependency> negatedRuleDependencies(Rule rule, ConceptManager conceptMgr, LogicManager logicMgr) {
+        return iterate(rule.condition().negatedConcludablesTriggeringRules(conceptMgr, logicMgr))
                 .flatMap(concludable -> concludable.getApplicableRules(conceptMgr, logicMgr))
                 .map(recursiveRule -> RuleDependency.of(recursiveRule, rule)).toSet();
     }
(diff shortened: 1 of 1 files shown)


logic/LogicManager.java
```diff
@@ -130,34 +130,45 @@ private Rule fromStructure(RuleStructure ruleStructure) {
...
         Set<Rule> negationRulesTriggeringRules = logicMgr.rulesWithNegations()
                 .filter(rule -> !rule.condition().negatedConcludablesTriggeringRules(conceptMgr, logicMgr).isEmpty())
                 .toSet();
 
         for (Rule negationRule : negationRulesTriggeringRules) {
             Map<Rule, RuleDependency> visitedDependentRules = new HashMap<>();
-            visitedDependentRules.put(negationRule, RuleDependency.of(negationRule, null));
-            List<RuleDependency> frontier = new LinkedList<>(ruleDependencies(negationRule, conceptMgr, logicMgr));
-            RuleDependency dependency;
+            LinkedList<RuleDependency> frontier = new LinkedList<>(negatedRuleDependencies(negationRule, conceptMgr, logicMgr));
```
flyingsilverfin, line 143: interesting - so this indirectly relies on the unifier having pruned the rule from being recursive because the `true` and `false` don't match up right?

logic/LogicManager.java
```diff
@@ -130,34 +130,45 @@ private Rule fromStructure(RuleStructure ruleStructure) {
...
                 } else {
-                    visitedDependentRules.put(dependency.recursiveRule, dependency);
                     Set<RuleDependency> recursive = ruleDependencies(dependency.recursiveRule, conceptMgr, logicMgr);
-                    recursive.removeAll(visitedDependentRules.values());
-                    frontier.addAll(recursive);
+                    iterate(recursive)
+                            .filter(rule -> !visitedDependentRules.containsKey(rule.recursiveRule))
+                            .forEachRemaining(ruleDependency -> {
+                                frontier.add(ruleDependency);
+                            });
```
flyingsilverfin, line 157: you can use `.forEachremaining(frontier::add)` here - if you remove the `{}`, intellij will suggest it for you!

logic/LogicManager.java
```diff
@@ -130,34 +130,45 @@ private Rule fromStructure(RuleStructure ruleStructure) {
     }
 
     private void validateCyclesThroughNegations(ConceptManager conceptMgr, LogicManager logicMgr) {
+        // Problem: When a derivation of an inferred concept contains its own negation.
+        // Tracking this at runtime is impractical - statically detect & disallow.
+        // Detect: A cycle which passes through atleast one negation.
+        //      Equivalently - a cycle starting from a negated concept in a rule.
```
flyingsilverfin, line 136: if you want to leave a comment describing this, the habit is to use a 
```
/**
*/
``` 
style comment before the function. This is called a Java-doc comment!

however, since this is just a bug fix of an incorrect implementation, we don't need to leave a comment at all. The namings and variables should be sufficient!
If you want to make it clearer, let's think about how we can further modularise the code to help explain the intent... maybe the inner loop should be its own method called `mayCycle(Rule rule)` or something like that

krishnangovindraj: Works.
I guess this PR will show up in the blame if anyone needs an explanation.

flyingsilverfin:
@krishnangovindraj The "goal" of the PR section shouldn't really reference a particular variable name or piece of code. Consider this from the random community member's point of view - they don't know/want to know the methods names deep in the code. 

An improved PR goal description is probably "This PR improves the validation of rules that contain negation. We fix cases  that could cause infinite loops (outlined in #1) and..."

flyingsilverfin:
@krishnangovindraj  also note that let's be careful to acknowledge we've only fixed half of: https://github.com/vaticle/typedb/issues/6500 right?

PR #6595, opened 2022-06-23
Fix erroneous session timeouts

dmitrii-ubskii:

We fix a bug which allowed session idle timeouts to trigger unexpectedly.

If `mayStartIdleTimeout()` was called at the same time in two different threads, as long as the session had no open transactions, two timeouts were scheduled. The object only holds on to one reference though, so if a new transaction is opened, only one future is cancelled. If, when the timer runs out, no transactions happen to be open, the other future—which could never be cancelled—would close the session. From the outside that looks like occasionally the timeout triggers immediately (within milliseconds) after the last transaction is closed.

We now refuse to schedule a timeout if one is already scheduled.

### DIFF
--- a/server/SessionService.java
+++ b/server/SessionService.java
@@ -53,7 +53,11 @@ public class SessionService implements AutoCloseable {
         this.accessLock = new StampedLock().asReadWriteLock();
         this.isOpen = new AtomicBoolean(true);
         this.transactionServices = new ConcurrentSet<>();
-        mayStartIdleTimeout();
+        startIdleTimeout();
+    }
+
+    private synchronized void startIdleTimeout() {
+        idleTimeoutTask = scheduled().schedule(this::idleTimeout, options.sessionIdleTimeoutMillis(), MILLISECONDS);
     }
 
     void register(TransactionService transactionSvc) {
@@ -95,28 +99,39 @@ public class SessionService implements AutoCloseable {
     }
 
     private void cancelIdleTimeout() {
-        assert idleTimeoutTask != null;
         idleTimeoutTask.cancel(false);
     }
 
     private synchronized void mayStartIdleTimeout() {
-        if (isOpen() && transactionServices.isEmpty()) {
-            idleTimeoutTask = scheduled().schedule(this::idleTimeout, options.sessionIdleTimeoutMillis(), MILLISECONDS);
+        if (isOpen() && transactionServices.isEmpty() && !idleTimeoutActive()) {
+            startIdleTimeout();
         }
     }
 
+    private boolean idleTimeoutActive() {
+        return !idleTimeoutTask.isDone();
+    }
+
     private void idleTimeout() {
-        if (transactionServices.isEmpty()) {
-            close();
-            LOG.warn("Session with ID " + session.uuid() + " timed out due to inactivity");
-        } else resetIdleTimeout();
+        try {
+            accessLock.writeLock().lock();
+            if (transactionServices.isEmpty()) {
+                if (isOpen.compareAndSet(true, false)) {
+                    session.close();
+                    typeDBSvc.closed(this);
+                    LOG.warn("Session with ID " + session.uuid() + " timed out due to inactivity");
+                }
+            }
+        } finally {
+            accessLock.writeLock().unlock();
+        }
     }
 
     @Override
     public void close() {
         try {
             accessLock.writeLock().lock();
-            if (idleTimeoutTask != null) idleTimeoutTask.cancel(false);
+            idleTimeoutTask.cancel(false);
             if (isOpen.compareAndSet(true, false)) {
                 transactionServices.forEach(TransactionService::close);
                 session.close();
--- a/server/SessionService.java
+++ b/server/SessionService.java
@@ -56,10 +56,6 @@ public class SessionService implements AutoCloseable {
         startIdleTimeout();
     }
 
-    private synchronized void startIdleTimeout() {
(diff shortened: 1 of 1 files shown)


server/SessionService.java
```diff
@@ -100,7 +100,7 @@ private void cancelIdleTimeout() {
     }
 
     private synchronized void mayStartIdleTimeout() {
-        if (isOpen() && transactionServices.isEmpty()) {
+        if (isOpen() && transactionServices.isEmpty() && idleTimeoutTask == null) {
```
lolski, line 103: Is this completely thread-safe?

In theory, it is still possible for multiple concurrent invocations of the `mayStartIdleTimeout` to evaluate the if condition to `true` and then execute the if body, right?

Maybe adding a lock, or wrapping `idleTimeoutTask` within an `AtomicRef` solution would be the 100% foolproof solution?

dmitrii-ubskii: The methpd is `synchronized`, which means entering the method acquires a lock on the object and returning from it releases the lock. We're a-okay.

server/SessionService.java
```diff
@@ -95,28 +99,39 @@ public synchronized void resetIdleTimeout() {
     }
 
     private void cancelIdleTimeout() {
-        assert idleTimeoutTask != null;
```
dmitrii-ubskii, line 98: Always true, as `idleTimeoutTask` is only set in `startIdleTimeout()`.

server/SessionService.java
```diff
@@ -95,28 +99,39 @@ public synchronized void resetIdleTimeout() {
...
+    }
+
     private void idleTimeout() {
-        if (transactionServices.isEmpty()) {
-            close();
-            LOG.warn("Session with ID " + session.uuid() + " timed out due to inactivity");
-        } else resetIdleTimeout();
+        try {
+            accessLock.writeLock().lock();
+            if (transactionServices.isEmpty()) {
```
dmitrii-ubskii, line 118: Make sure no transactions were open while we were waiting for the write lock.

server/SessionService.java
```diff
@@ -95,28 +99,39 @@ public synchronized void resetIdleTimeout() {
...
+                if (isOpen.compareAndSet(true, false)) {
+                    session.close();
+                    typeDBSvc.closed(this);
+                    LOG.warn("Session with ID " + session.uuid() + " timed out due to inactivity");
+                }
+            }
```
flyingsilverfin, line 124: what about if the `transactionServices` are not empty?

server/SessionService.java
```diff
@@ -53,7 +53,11 @@ public SessionService(TypeDBService typeDBSvc, TypeDB.Session session, Options.S
         this.accessLock = new StampedLock().asReadWriteLock();
         this.isOpen = new AtomicBoolean(true);
         this.transactionServices = new ConcurrentSet<>();
-        mayStartIdleTimeout();
+        startIdleTimeout();
+    }
+
+    private synchronized void startIdleTimeout() {
+        idleTimeoutTask = scheduled().schedule(this::idleTimeout, options.sessionIdleTimeoutMillis(), MILLISECONDS);
```
flyingsilverfin, line 60: would expect to find this function near the other methods to do with starting/stopping idle timeouts

flyingsilverfin (changes requested):

server/SessionService.java
dmitrii-ubskii: Then the transaction must have been opened between entering the method and acquiring the lock. We do nothing and release the lock.

server/SessionService.java
dmitrii-ubskii: I definitely want this method to be right after the constructor to make it crystal clear that `idleTimeoutTask` is initialised at construction and therefore never null. I considered pulling the rest of the related functions up to group with this, but that would push the accessors further down than I'd like.

flyingsilverfin: hmm I think it's clear enough in the constructor and we can put `startIdleTimeout` near the others, that is the stronger grouping

server/SessionService.java
dmitrii-ubskii: Yes, it would try to cancel a timeout, which wouldn't do anything, and try to restart it. The last transaction may have been closed by the time we reach this line after checking the condition, so we also might cancel a new active timeout and replace it with a new future. Either way, essentially a very busy no-op.

flyingsilverfin: right - i actually meant we could delete the method then!

server/SessionService.java
```diff
@@ -95,28 +99,39 @@ public synchronized void resetIdleTimeout() {
...
 
     private synchronized void mayStartIdleTimeout() {
-        if (isOpen() && transactionServices.isEmpty()) {
-            idleTimeoutTask = scheduled().schedule(this::idleTimeout, options.sessionIdleTimeoutMillis(), MILLISECONDS);
+        if (isOpen() && transactionServices.isEmpty() && !idleTimeoutActive()) {
+            startIdleTimeout();
         }
     }
 
+    private boolean idleTimeoutActive() {
```
flyingsilverfin, line 111: also, for boolean methods we tend to use `isIdleTimeoutActive()`

server/SessionService.java
dmitrii-ubskii: No, it's used in TypeDBService (L215)

flyingsilverfin: ah okay

flyingsilverfin (approved):

PR #6598, opened 2022-06-27
Correctly add RocksDB JAR to the distribution

alexjpwalker:

We updated our distribution targets to ensure the RocksDB JAR is correctly added.

The RocksDB JAR is added to the output distribution via its own `java_deps` target, named `//server:server-deps-prod`. `java_deps` used to package both the `data_runfiles` and `files` of the `java_library` and `java_binary` targets passed to it. This was largely superfluous, as when called on a `java_binary`, `data_runfiles` would contain all necessary files. But `:server-deps-prod` was incidentally working because of it.

To fix it, we've introduced a `java_binary` target, `:server-bin-deps-prod`, in line with our other uses of `java_deps` in this repo (e.g: `server-bin-deps-mac`.

### DIFF
--- a/server/BUILD
+++ b/server/BUILD
@@ -147,6 +147,13 @@ java_binary( # TODO: this target should not be needed if :server-deps-windows ca
     tags = ["maven_coordinates=com.vaticle.typedb:typedb-core-server-bin-deps:{pom_version}"],
 )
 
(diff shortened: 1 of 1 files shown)


haikalpribadi:
How did it work before and what caused it to break @alexjpwalker ?

PR #6602, opened 2022-07-01
Update TypeDB runner

lolski:

We updated TypeDB runner usage according to the latest version of the API

1.  Updated the version of `@vaticle_typedb_common`.
2. Updated the declaration of TypeDB runner according to the latest version of the API / syntax.


flyingsilverfin:
looks fine!

PR #6604, opened 2022-07-07
Parallel reads within a single transaction

alexjpwalker:

Read operations within a single transaction may now run in parallel.

It is a fact that TypeDB allows parallel reads. However, this was not really in line with the behaviour of the transaction RPC model.

Transaction RPCs were gated by a mutex in the form of a `synchronized` block, allowing for - within one transaction - precisely one RPC call to be handled concurrently. In the context of queries, there are a few kinds of calls:

- Write query executions (define, undefine, insert, etc.)
- Read query executions (match, match-aggregate, etc.)
- Answer iterations (reading N answers from an open iterator)

Concurrent reads were supported, somewhat, in that when we ran two `match` queries in the same transaction, we could interleave the iteration of their answers (by, say, reading 50 answers from query 1, then 50 answers from query 2, then 50 more answers from query 1). But this felt like a watered-down version of the concurrency that we had the power to support.

We've rewritten the safeguard by categorising each Transaction RPC as either a _read_ or a _write_ operation. For example:

- Define, Undefine, Insert queries are writes
- Commits are writes
- Match query executions are reads
- Answer iterations are reads

For simplicity, we currently treat Concept API calls as writes (this is equivalent to the old mutex behaviour, so is not a regression)

As a result, read operations within the same transaction are now truly run in parallel, rather than just being interleaved. The concurrency model and architecture are also more explainable: _write operations need a write lock, read operations need a read lock._

### DIFF
--- a/server/TransactionService.java
+++ b/server/TransactionService.java
@@ -52,6 +52,9 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.StampedLock;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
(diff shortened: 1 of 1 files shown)


alexjpwalker:
Do not merge until all clients have been tested with the new TypeDB snapshot artifact.

server/TransactionService.java
```diff
@@ -132,10 +137,12 @@ public void onError(Throwable error) {
         close(error);
     }
 
-    private synchronized void execute(TransactionProto.Transaction.Req request) {
+    private void execute(TransactionProto.Transaction.Req request) {
```
alexjpwalker, line 140: We've replaced our old mutex lock (`synchronized`) with a read-write lock. Read operations acquire a read lock; write operations acquire a write lock.

PR #6606, opened 2022-07-08
Reasoning test oracle uses stratified forward chaining

krishnangovindraj:
Fix how the forward-chaining oracle ( used in reasoner behaviour tests) behaves in the presence of negation.
The current implementation iteratively re-evaluates every rule to derive new inferred concepts, and terminates when no new concepts are inferred. This fails with negated rules, since it may prematurely conclude a negation is satisfied before a contradictory concept is inferred. 

The oracle now computes the stratification of rules (for [stratified negation](https://en.wikipedia.org/wiki/Stratification_(mathematics)#In_mathematical_logic)) in the schema and evaluates rules accordingly - it first derives all concepts inferable from a given strata before moving on to the next.

### DIFF
--- a/test/behaviour/reasoner/verification/ForwardChainingMaterialiser.java
+++ b/test/behaviour/reasoner/verification/ForwardChainingMaterialiser.java
@@ -59,13 +59,13 @@ public class ForwardChainingMaterialiser {
     private final Map<com.vaticle.typedb.core.logic.Rule, Rule> rules;
     private final CoreTransaction tx;
     private final Materialisations materialisations;
-    private List<Set<Rule>> strata;
+    private List<Set<Rule>> ruleStrata;
 
     private ForwardChainingMaterialiser(CoreSession session) {
         this.rules = new HashMap<>();
         this.tx = session.transaction(Arguments.Transaction.Type.WRITE, new Options.Transaction().infer(false));
         this.materialisations = new Materialisations();
-        this.strata = null;
+        this.ruleStrata = null;
     }
 
     public static ForwardChainingMaterialiser materialise(CoreSession session) {
@@ -76,7 +76,7 @@ public class ForwardChainingMaterialiser {
 
     private void materialise() {
         tx.logic().rules().forEachRemaining(rule -> this.rules.put(rule, new Rule(rule)));
-        for (Set<Rule> stratum : getStratifiedRules()) {
+        for (Set<Rule> stratum : computeNegationInducedRuleStratification()) {
             boolean reiterateRules = true;
             while (reiterateRules) {
                 reiterateRules = false;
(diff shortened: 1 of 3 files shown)


test/behaviour/reasoner/verification/ForwardChainingMaterialiser.java
```diff
@@ -110,10 +116,58 @@ Optional<Materialisation> conditionMaterialisations(com.vaticle.typedb.core.logi
...
+                if (!inLesserStratum.contains(dependency)) {
+                    strataDFS(dependency, inLesserStratum, inLesserOrEqualStratum);
+                }
+            }
+        }
+    }
+
+    private void updateStrata(Set<Rule> inLesserStratum, Set<Rule> inLesserOrEqualStratum) {
+        if (!inLesserOrEqualStratum.isEmpty()) {
+            Set<Rule> inThisStratum = iterate(inLesserOrEqualStratum).filter(rule -> !inLesserStratum.contains(rule)).toSet();
```
jmsfltchr, line 159: This is a set operation so we should use the provided set operations:
```
Set<Rule> inThisStratum = new HashSet<>(inLesserOrEqualStratum);
inThisStratum.removeAll(inLesserStratum);
```

jmsfltchr (changes requested):

test/behaviour/reasoner/verification/ForwardChainingMaterialiser.java
```diff
@@ -110,10 +116,58 @@ Optional<Materialisation> conditionMaterialisations(com.vaticle.typedb.core.logi
         return materialisations.forCondition(rules.get(rule), conditionAnswer);
     }
 
+    private List<Set<Rule>> getStratifiedRules() {
```
flyingsilverfin, line 119: fyi we don't really use the logic terminology like this very often - only if we can't find a name that requires less context/understanding for future readers. For example, the validation logic in TypeDB does stratification checking without using the word stratification...

flyingsilverfin (commented):
Just got some email notifications and dropped in :P

test/behaviour/reasoner/verification/ForwardChainingMaterialiser.java
flyingsilverfin: That's true, but it would be even easier if you can get away without googling by naming it something like "ruleCyclesWithoutNegations" or something, that would be more self-explanatory

not sure exactly that stratification is anymore after i understood it once and sidestepped needing the term in core since then :)

flyingsilverfin: which illustrates my point actually

flyingsilverfin: well if you're using `strata` you can keep the naming using `strata` consistent and don't need the long name... i have totally forgotten stratification so i will have to think about this separately! feel free to proceed as it was

flyingsilverfin (approved):

flyingsilverfin:
Looks good!

PR #6611, opened 2022-07-19
Check explanation completeness in reasoner tests

krishnangovindraj:

Reasoner correctness checks, specifically Completness checks, are enabled on explanations, and bugs uncovered as a result are fixed.

 This PR adds a check for the completeness of explanations - The number of explanations must match the the number of  materialisations found for the same materializable ( or, the number of conditions which lead to the same conclusion)

This PR also includes fixes bugs found in the explain-specific parts of the reasoner:
1. Disallow creating an inferred attribute where a nonInferred attribute exists, when an explicit-has is being inferred ([more](https://github.com/vaticle/typedb/pull/6611/files#r924398135)).
2. An explanation did not enforce the unifier requirements on answers received, leading to answers with mismatched role-variables being accepted as valid-explanations. ([more](https://github.com/vaticle/typedb/pull/6611/files#r924400333))

### DIFF
--- a/reasoner/controller/ConcludableController.java
+++ b/reasoner/controller/ConcludableController.java
@@ -259,7 +259,19 @@ public abstract class ConcludableController<INPUT, OUTPUT,
 
             @Override
             protected FunctionalIterator<ConceptMap> filterNonInferred(ConceptMap conceptMap) {
-                if (!concludable.isInferredAnswer(conceptMap)) return Iterators.empty();
+                Concept conceptToCheck = null;
+                if (concludable.isAttribute()) {
+                    conceptToCheck = conceptMap.get(concludable.asAttribute().attribute().id());
+                } else if (concludable.isIsa()) {
+                    conceptToCheck = conceptMap.get(concludable.asIsa().isa().owner().id());
+                }
+                if (conceptToCheck != null) {
+                    if (conceptToCheck.asThing().isInferred()) {
+                        return Iterators.single(conceptMap);
+                    } else {
+                        return Iterators.empty();
+                    }
+                }
                 return Iterators.single(conceptMap);
             }
 
@@ -330,7 +342,21 @@ public abstract class ConcludableController<INPUT, OUTPUT,
 
             @Override
             protected FunctionalIterator<Explanation> filterNonInferred(Explanation explanation) {
-                if (!concludable.isInferredAnswer(explanation.conclusionAnswer().WHATCOMESHERE!)) return Iterators.empty();
+                Set<Variable> conclusionConceptsToCheck = null;
+                if (concludable.isAttribute()) {
+                    conclusionConceptsToCheck = explanation.variableMapping().get(concludable.asAttribute().attribute().id());
+                } else if (concludable.isIsa()) {
+                    conclusionConceptsToCheck = explanation.variableMapping().get(concludable.asIsa().isa().owner().id());
+                }
+                if (conclusionConceptsToCheck != null) {
+                    for (Variable toCheck : conclusionConceptsToCheck) {
+                        if (explanation.conclusionAnswer().concepts().get(toCheck).asThing().isInferred()) {
+                            return Iterators.single(explanation);
+                        } else {
+                            return Iterators.empty();
+                        }
+                    }
+                }
                 return Iterators.single(explanation);
             }
 
--- a/reasoner/controller/ConcludableController.java
+++ b/reasoner/controller/ConcludableController.java
@@ -49,6 +49,7 @@ import java.util.function.Supplier;
 
 import static com.vaticle.typedb.common.collection.Collections.set;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
+import static com.vaticle.typedb.core.common.iterator.Iterators.empty;
 import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
 import static com.vaticle.typedb.core.reasoner.processor.reactive.PoolingStream.BufferedFanStream.fanInFanOut;
 
@@ -257,9 +258,9 @@ public abstract class ConcludableController<INPUT, OUTPUT,
             private FunctionalIterator<ConceptMap> filterInferred(ConceptMap conceptMap) {
                 // TODO: Requires duplicate effort to filter out inferred concludable answers. Instead create a new
                 //  traversal mode that doesn't return inferred concepts
-                if (concludable.isInferredAnswer(conceptMap)) return Iterators.empty();
+                if (concludable.isInferredAnswer(conceptMap)) return empty();
                 for (Concept c : conceptMap.concepts().values()) {
-                    if (c.isThing() && c.asThing().isInferred()) return Iterators.empty();
+                    if (c.isThing() && c.asThing().isInferred()) return empty();
                 }
                 return Iterators.single(conceptMap);
             }
@@ -276,7 +277,7 @@ public abstract class ConcludableController<INPUT, OUTPUT,
                     if (conceptToCheck.asThing().isInferred()) {
                         return Iterators.single(conceptMap);
                     } else {
-                        return Iterators.empty();
+                        return empty();
                     }
                 }
                 return Iterators.single(conceptMap);
@@ -360,7 +361,7 @@ public abstract class ConcludableController<INPUT, OUTPUT,
                         if (explanation.conclusionAnswer().concepts().get(toCheck).asThing().isInferred()) {
                             return Iterators.single(explanation);
                         } else {
-                            return Iterators.empty();
+                            return empty();
                         }
                     }
                 }
@@ -376,7 +377,11 @@ public abstract class ConcludableController<INPUT, OUTPUT,
             protected Publisher<Explanation> transformInput(Publisher<PartialExplanation> input,
                                                             Unifier unifier,
                                                             Unifier.Requirements.Instance requirements) {
-                return input.map(
+                return input.flatMap(p -> {
+                    FunctionalIterator<ConceptMap> exists = unifier.unUnify(p.conclusionAnswer().concepts(), requirements);
+                    if (exists.hasNext()) return iterate(p);
+                    else return empty();
+                }).map(
                         p -> new Explanation(p.rule(), unifier.mapping(), p.conclusionAnswer(), p.conditionAnswer())
                 );
             }
--- a/reasoner/controller/ConcludableController.java
+++ b/reasoner/controller/ConcludableController.java
@@ -158,7 +158,7 @@ public abstract class ConcludableController<INPUT, OUTPUT,
             //  concludable. They should be filtered before being passed to the concludableProcessor's constructor
             assert bounds.equals(this.bounds);
             return new Processor.Explain(
-                    explainDriver, driver(), processorContext(), bounds, set(), conclusionUnifiers, reasonerConsumer,
+                    explainDriver, driver(), processorContext(), concludable, bounds, set(), conclusionUnifiers, reasonerConsumer,
                     () -> Traversal.traversalIterator(registry(), concludable.pattern(), bounds),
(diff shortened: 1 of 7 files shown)


test/behaviour/reasoner/verification/SoundnessVerifier.java
```diff
@@ -101,14 +114,46 @@ private boolean canExplanationBeVerified(Explanation explanation) {
...
                 if (collectedExplanations.contains(explanation)) return;
                 else collectedExplanations.add(explanation);
                 verifyAnswerAndCollectExplanations(explanation.conditionAnswer(), tx);
                 verifyVariableMapping(answer, explainable, explanation);
             });
+
+            Pair<Conjunction, ConceptMap> deferredBindableConjunction = new Pair<>(explainable.conjunction(), answer.filter(explainable.conjunction().retrieves()));
+            assert !deferredExplanationCount.containsKey(deferredBindableConjunction) ||
+                    deferredExplanationCount.get(deferredBindableConjunction).equals(explanations.size());
+            deferredExplanationCount.put(deferredBindableConjunction, explanations.size());
```
krishnangovindraj, line 130: We can't call verifyNumberOfExplanations here because the inferredConcepts are not guaranteed to have been mapped. We instead store the conjunction and the pre-image of the mapped concepts, along with the expected explanation count.

reasoner/controller/ConcludableController.java
```diff
@@ -248,9 +258,27 @@ protected void mayAddTraversal() {
...
+                    conceptToCheck = conceptMap.get(concludable.asAttribute().attribute().id());
+                } else if (concludable.isIsa()) {
+                    conceptToCheck = conceptMap.get(concludable.asIsa().isa().owner().id());
+                }
+                if (conceptToCheck != null) {
+                    if (conceptToCheck.asThing().isInferred()) {
+                        return Iterators.single(conceptMap);
+                    } else {
+                        return empty();
+                    }
```
jmsfltchr, line 281: This is coded to specifically catch the scenario where: 
1. the rule conclusion is a `has explicit`
2. the attribute concept is actually an inserted fact
3. the concludable only concerns the attribute and not the `has` edge, in which case this answer no longer represents an inferred answer to the concludable

Not rejecting this answer at this stage leads to an extra explanation, which we deem invalid according to our definition of what should be explained (only inferences should be explained).

reasoner/controller/ConcludableController.java
```diff
@@ -317,6 +348,26 @@ public void setUp() {
...
+                        if (explanation.conclusionAnswer().concepts().get(toCheck).asThing().isInferred()) {
+                            return Iterators.single(explanation);
+                        } else {
+                            return empty();
+                        }
+                    }
+                }
+                return Iterators.single(explanation);
+            }
+
```
jmsfltchr, line 370: This is coded to specifically catch the scenario where: 
1. the rule conclusion is a `has explicit`
2. the attribute concept is actually an inserted fact
3. the concludable only concerns the attribute and not the `has` edge, in which case this answer no longer represents an inferred answer to the concludable

Not rejecting this answer at this stage leads to an extra explanation, which we deem invalid according to our definition of what should be explained (only inferences should be explained).

reasoner/controller/ConcludableController.java
```diff
@@ -326,7 +377,11 @@ public void rootPull() {
             protected Publisher<Explanation> transformInput(Publisher<PartialExplanation> input,
                                                             Unifier unifier,
                                                             Unifier.Requirements.Instance requirements) {
-                return input.map(
+                return input.flatMap(p -> {
+                    FunctionalIterator<ConceptMap> exists = unifier.unUnify(p.conclusionAnswer().concepts(), requirements);
+                    if (exists.hasNext()) return iterate(p);
+                    else return empty();
+                }).map(
```
jmsfltchr, line 384: We should only return an explanation if the answer it contains can actually answer this concludable, which requires that it can ununify.

test/behaviour/reasoner/verification/SoundnessVerifier.java
```diff
@@ -101,14 +114,46 @@ private boolean canExplanationBeVerified(Explanation explanation) {
...
+                    numExplanationsExpected.getAndAdd(unifier.unUnify(materialisation.boundConclusion().bounds(), boundsAndRequirements.get().second()).toSet().size());
+                });
+            });
+            if (deferredExplanationCount.get(deferredBindableConjunction) != numExplanationsExpected.get()) {
+                throw new SoundnessException(String.format(
+                        "Soundness testing found an incorrect number of explanations for query \"%s\" for explainable" +
+                                " \"%s\".\nExplanations expected: %s\nExplanations found: %s",
+                        inferenceQuery, deferredBindableConjunction.first(), numExplanationsExpected, deferredExplanationCount.get(deferredBindableConjunction)
+                ));
+            }
```
jmsfltchr, line 156: Add a check that the number of explanations is as expected. We do this as we do the soundness check which queries for the user query and queries the body of each rule used. For a given query find the number of explanations that there should be for each concludable, getting this number from the forward-chaining oracle. Compare this with the number of explanations actually found for that concludable (actually from the explainable that represents that concludable).

Doing this required some extra trickery to find the number of actual explanations that should be present since the materialiser doesn't store an exact equivalent to explanations, it has `concludableMaterialisations` instead.

test/behaviour/reasoner/verification/ForwardChainingMaterialiser.java
```diff
@@ -183,13 +181,15 @@ private Materialisations() {
...
             FunctionalIterator<Materialisation> materialisations = empty();
-            Optional<Thing> inferredConcept = boundConcludable.inferredConcept();
-            if (inferredConcept.isPresent()) {
-                materialisations = materialisations.link(forThing(inferredConcept.get()));
-            }
+
             Optional<Pair<Thing, Attribute>> inferredHas = boundConcludable.inferredHas();
             if (inferredHas.isPresent()) {
                 materialisations = materialisations.link(forHas(inferredHas.get().first(), inferredHas.get().second()));
+            }else{
```
krishnangovindraj, line 188: returned the attribute materialisation in addition to the "has" materialisation when a "has" concludable was passed

reasoner/controller/ConcludableController.java
```diff
@@ -158,7 +159,7 @@ protected Processor.Explain createProcessorFromDriver(Driver<Processor.Explain>
             //  concludable. They should be filtered before being passed to the concludableProcessor's constructor
             assert bounds.equals(this.bounds);
             return new Processor.Explain(
-                    explainDriver, driver(), processorContext(), bounds, set(), conclusionUnifiers, reasonerConsumer,
+                    explainDriver, driver(), processorContext(), concludable, bounds, set(), conclusionUnifiers, reasonerConsumer,
```
flyingsilverfin, line 162: this should have a comment as to why we ned this new argument

reasoner/controller/ConcludableController.java
flyingsilverfin, line 281: our style for these short `if` blocks is not to have `{}` and inline the `if` with the body:
```
if (condition) return ...;
else return empty()
```
Generally the rule is that if the `if (condition)` together with the body fits in one line, we don't need the `{}` or newlines

test/behaviour/reasoner/verification/BoundPattern.java
```diff
@@ -156,21 +159,38 @@ public Concludable concludable() {
     static class BoundConclusion {
         private final BoundConjunction conjunction;
         private final Rule.Conclusion unboundConclusion;
+        private final Map<Variable, Concept> bounds;
 
-        private BoundConclusion(BoundConjunction conjunction, Rule.Conclusion unboundConclusion) {
+        private BoundConclusion(BoundConjunction conjunction, Rule.Conclusion unboundConclusion, Map<Variable, Concept> bounds) {
```
flyingsilverfin, line 164: suspicous that we have both `BoundConjunction` (presumably the conjunction and bounds), then the `unboundConclusion` (which contains the simple conjunction within the conclusion), and `bounds`... we should be able to derive the `BoundConclusion` from the other two, right?

test/behaviour/reasoner/verification/CompletenessVerifier.java
```diff
@@ -73,7 +73,7 @@ private void verifyConjunction(BoundConjunction boundConjunction) {
                     verifyConclusionReasoning(materialisation.boundConclusion());
                 });
                 verifyConcludableReasoning(boundConcludable);
-            }
+            }  // TODO: If not inferred should we do a traversal to check?
```
flyingsilverfin, line 76: can we resolve this TODO?

test/behaviour/reasoner/verification/ForwardChainingMaterialiser.java
flyingsilverfin, line 188: formatting - use auto-format on modified files before submitting the PR/undrafting it

test/behaviour/reasoner/verification/SoundnessVerifier.java
```diff
@@ -101,14 +114,46 @@ private boolean canExplanationBeVerified(Explanation explanation) {
     private void verifyAnswerAndCollectExplanations(ConceptMap answer, Transaction tx) {
         verifyExplainableVars(answer);
         answer.explainables().iterator().forEachRemaining(explainable -> {
-            tx.query().explain(explainable.id()).forEachRemaining(explanation -> {
+            List<Explanation> explanations = tx.query().explain(explainable.id()).toList();
+            iterate(explanations).forEachRemaining(explanation -> {
```
flyingsilverfin, line 118: this can be shortened back to the old style of `tx.query().explain(...).forEachRemaining(...)`

flyingsilverfin (changes requested):

test/behaviour/reasoner/verification/CompletenessVerifier.java
krishnangovindraj: @jmsfltchr , 
I see no harm in doing it. We'd just be checking whether the reasoner returns retrieved answers as well.

krishnangovindraj:
```
int numReasonedAnswersReceived = numReasonedAnswers(boundConcludable.pattern(), boundConcludable.concludable().retrieves());
                if (numReasonedAnswersReceived != 1) { /*throw*/ }
```
I thought this would work, but it seems to find 0 answers quite often. Am I doing something obviously wrong?

krishnangovindraj: Second thought: How would we do it? And is there a point to it? 
We'd be using traversal to fetch it both times. Maybe once we should do it with infer turned on and once with it turned off and see whether both get the same (0 or 1) number of results

reasoner/controller/ConcludableController.java
jmsfltchr: True - we needed this argument to be able to check whether the concludable fits the criteria that mean that may have received a non-inferred fact from a conclusion to answer this concludable.

test/behaviour/reasoner/verification/CompletenessVerifier.java
jmsfltchr: There's no such thing as inferred on or off for a traversal. Where we have non-inferred answers only coming from a traversal then we've implemented that by adding a filter.
Is there a point to it? I'm not sure there is. In doing so I think we might actually be testing `isInferredAnswer` more than anything else. I'm satisfied to skip it.

krishnangovindraj: If a thing can be both inferred and retrieved, wouldn' it pass through a ConcludableProcessor in case reasoning is enabled?
But my previous comment was wrong. To verify, Completeness verifier would just do a query (with infer on) and see whether the answer is being returned. It seems like a harmless check worth doing (and it uncovered a weird/unrelated bug in ForwardChainingMaterializer when I tried to do it)
So Joshua and I think we'll do it, but in a different PR since it's unrelated to checking explanation completeness.

PR #6612, opened 2022-07-21
Query planner emulates incremental solving

flyingsilverfin:

Query planner implement incremental solving, on top of OR-Tools solvers that do not support incremental solving. To implement this, the planner always restarts a new solver and provides the previous solution as a hint to the next invocation.

* Refactor `Optimiser` to hide unnecessary APIs dealing with dynamically reconfiguring the optimisation model
* Add sanity check to validate that the solution held in each `OptimiserVariable` satisfies each constraint
* Always set the hints before invocing the OR-Tools solver to emulate incremental solutions
* SAT solver doesn't need increasing amounts of time allocated, we can just provide a new 100ms window each time

### DIFF
--- a/common/optimiser/Optimiser.java
+++ b/common/optimiser/Optimiser.java
@@ -123,7 +123,7 @@ public class Optimiser {
         variables.forEach(var -> var.initialise(solver));
         constraints.forEach(constraint -> constraint.initialise(solver));
         setConstraintCoefficients();
-        applyObjective();
+        setObjective();
     }
 
     private void setConstraintCoefficients() {
@@ -150,7 +150,7 @@ public class Optimiser {
         solver.setHint(mpVariables, hints);
     }
 
-    private void applyObjective() {
+    private void setObjective() {
         objectiveCoefficients.forEach((var, coeff) -> solver.objective().setCoefficient(var.mpVariable(), coeff));
     }
 
--- a/common/optimiser/Optimiser.java
+++ b/common/optimiser/Optimiser.java
@@ -28,6 +28,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import static com.google.ortools.linearsolver.MPSolverParameters.IncrementalityValues.INCREMENTALITY_ON;
@@ -39,10 +40,10 @@ import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
 
 /**
  * A wrapper around an OR-Tools solver, which supports an incremental solving mode on top of non-incremental solvers.
- * This is implemented by using hints to pass the previous solution to the solver.
+ * This implementation expects that a value is always set on each OptimiserVariable that satisfies the entire model.
  *
- * We guarantee subsequent solutions are valid within the variables/constraints and at least as good whenever
- * the objective function does not change.
+ * Optimiser guarantees solutions are better than the last whenever the objective has not changed.
+ * Optimiser models can be modified dynamically by updating coefficients on the objective or on constraints.
  */
 public class Optimiser {
 
@@ -52,40 +53,43 @@ public class Optimiser {
     private MPSolver solver;
     private MPSolverParameters parameters;
     private Status status;
-    private boolean hasSolver;
-    private double objectiveValue = Double.MAX_VALUE;
+    private boolean constraintsChanged;
+    private Double objectiveValue;
 
     public Optimiser() {
         variables = new ArrayList<>();
         constraints = new HashSet<>();
         objectiveCoefficients = new HashMap<>();
         status = Status.NOT_SOLVED;
-        hasSolver = false;
+        objectiveValue = null;
+        constraintsChanged = false;
     }
 
     public synchronized Status optimise(long timeLimitMillis) {
-        assert isSatisfied();
+        assert modelSatisfied();
         if (isOptimal()) return status;
+        if (solver == null) createSolver();
+        else if (constraintsChanged) {
+            solver.reset();
+            applyConstraintCoefficients();
+        }
         setValuesAsHints();
         solver.setTimeLimit(timeLimitMillis);
         status = Status.of(solver.solve(parameters));
-        recordValues();
+        recordSolverValues();
         if (isOptimal()) releaseSolver();
-        assert isSatisfied();
+        assert modelSatisfied();
         return status;
     }
 
-    private void recordValues() {
-        if (status != Status.NOT_SOLVED && status != Status.ERROR) {
-            System.out.println("Old objective: " + objectiveValue + " - new objective: " + solver.objective().value());
-            if (solver.objective().value() < objectiveValue) {
-                objectiveValue = solver.objective().value();
-                variables.forEach(OptimiserVariable::recordSolutionValue);
-            }
+    private void recordSolverValues() {
+        if ((isFeasible() || isOptimal()) && (objectiveValue == null || solver.objective().value() < objectiveValue)) {
+            objectiveValue = solver.objective().value();
+            variables.forEach(OptimiserVariable::recordSolutionValue);
         }
     }
 
-    private boolean isSatisfied() {
+    private boolean modelSatisfied() {
         return iterate(variables).allMatch(OptimiserVariable::isSatisfied) &&
                 iterate(constraints).allMatch(OptimiserConstraint::isSatisfied);
     }
@@ -102,14 +106,11 @@ public class Optimiser {
         return status == Status.ERROR;
     }
 
-    public synchronized void restartFromSolution() {
-        if (hasSolver) {
-            solver.reset();
-            setSolverModel();
-        } else initialiseSolver();
+    public void setConstraintsChanged() {
+        constraintsChanged = true;
     }
 
-    private void initialiseSolver() {
+    private void createSolver() {
         solver = MPSolver.createSolver("SAT");
         solver.objective().setMinimization();
         parameters = new MPSolverParameters();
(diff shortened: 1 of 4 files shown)


common/optimiser/Optimiser.java
```diff
@@ -45,30 +53,49 @@ public class Optimiser {
     private MPSolver solver;
     private MPSolverParameters parameters;
     private Status status;
-    private boolean hasSolver;
```
flyingsilverfin, line 48: we got rid of this in favor of `solver == null` checks

common/optimiser/Optimiser.java
```diff
@@ -45,30 +53,49 @@ public class Optimiser {
...
-    private double objectiveValue;
+    private boolean isConstraintsUpToDate;
+    private Double objectiveValue;
```
flyingsilverfin, line 57: `objectiveValue` may not exist yet until we have actually got some objective function. A nullable object makes more sense untl a objective is defined

common/optimiser/Optimiser.java
```diff
@@ -45,30 +53,49 @@ public class Optimiser {
...
         status = Status.NOT_SOLVED;
-        hasSolver = false;
+        objectiveValue = null;
+        isConstraintsUpToDate = false;
     }
 
     public synchronized Status optimise(long timeLimitMillis) {
-        assert hasSolver;
         if (isOptimal()) return status;
+        assert solutionSatisfiesModel();
```
flyingsilverfin, line 70: we always expect to move from one valid solution to another valid solution

common/optimiser/Optimiser.java
```diff
@@ -45,30 +53,49 @@ public class Optimiser {
...
+        maySetSolver();
+        setSolutionAsHints();
```
flyingsilverfin, line 72: always apply the previous solution as hints to the model - since the solver always restarts this is the only way to "carry on" solving from the last solution

common/optimiser/Optimiser.java
```diff
@@ -45,30 +53,49 @@ public class Optimiser {
...
-        hasSolver = false;
+        objectiveValue = null;
+        isConstraintsUpToDate = false;
     }
 
     public synchronized Status optimise(long timeLimitMillis) {
-        assert hasSolver;
         if (isOptimal()) return status;
+        assert solutionSatisfiesModel();
+        maySetSolver();
```
flyingsilverfin, line 71: We got rid of an API: `restartFromSolution()`. From the point of view of users of `Optimiser`, this is an implementation-specific method that we had to call to update the solver whenever the `Constraint` coefficients had changed, or we created the model for the first time. However, this can all be transparently be handled.

Now, whenever we call `optimise()`, we already know if the solver needs to be initialised from scratch versus the coefficients on constraints need to be updated only. This leads to a cleaner UX of the Optimiser class.

common/optimiser/Optimiser.java
```diff
@@ -83,67 +110,76 @@ public boolean isError() {
...
-        if (hasSolver) {
-            solver.reset();
-            setSolverModel();
-        } else initialiseSolver();
+    public void setConstraintsChanged() {
+        isConstraintsUpToDate = true;
     }
 
-    private void initialiseSolver() {
+    private void createSolver() {
```
flyingsilverfin, line 117: I prefer `create` to `initialise` here because create/release are more like opposites. Also, `initialise` kind of has a connotation that it only happens once - initialisation normally occurs once per class. `create` is more flexible in the english language.

common/optimiser/Optimiser.java
```diff
@@ -45,30 +53,49 @@ public class Optimiser {
...
+            solver.reset();
+            setConstraintCoefficients();
+        }
+    }
+
+    private void recordSolverValues() {
+        if ((isFeasible() || isOptimal()) && solver.objective().value() < objectiveValue()) {
+            objectiveValue = solver.objective().value();
+            variables.forEach(OptimiserVariable::recordSolutionValue);
+        }
```
flyingsilverfin, line 93: we now only record the objective and solutions to optimiser variables whenever a solution is found that is actually better than the prior one!
We require this because we aren't using a "true" incremental solver, but restarting each time with the hints from the prior one.

common/optimiser/Optimiser.java
```diff
@@ -83,67 +110,76 @@ public boolean isError() {
         return status == Status.ERROR;
     }
 
-    public synchronized void restartFromSolution() {
-        if (hasSolver) {
-            solver.reset();
-            setSolverModel();
-        } else initialiseSolver();
+    public void setConstraintsChanged() {
+        isConstraintsUpToDate = true;
```
flyingsilverfin, line 114: we allow `Constraint` to call back and notify the `Optimiser` that owns it that the constraint has changed

common/optimiser/Optimiser.java
```diff
@@ -83,67 +110,76 @@ public boolean isError() {
...
 
     public void setObjectiveCoefficient(OptimiserVariable<?> var, double coeff) {
-        objectiveCoefficients.put(var, coeff);
-        if (hasSolver) solver.objective().setCoefficient(var.mpVariable(), coeff);
-        else if (isOptimal()) status = Status.FEASIBLE;
+        Double prevCoeff = objectiveCoefficients.put(var, coeff);
+        if (!Objects.equals(coeff, prevCoeff)) {
+            if (solver != null) solver.objective().setCoefficient(var.mpVariable(), coeff);
+            else if (isOptimal()) status = Status.FEASIBLE;
+            objectiveValue = null;
```
flyingsilverfin, line 176: because we likely set many coefficients at a time, we just `null` the objective value and re-compute it when asked to the next time via `objectiveValue()`

traversal/planner/GraphPlanner.java
```diff
@@ -262,18 +261,16 @@ private void optimise(GraphManager graphMgr, boolean singleUse) {
         // TODO: we should have a more clever logic to allocate extra time
         long allocatedDuration = singleUse ? HIGHER_TIME_LIMIT_MILLIS : DEFAULT_TIME_LIMIT_MILLIS;
         Instant start, endSolver, end;
-        totalDuration += allocatedDuration;
 
         start = Instant.now();
-        optimiser.optimise(totalDuration);
+        optimiser.optimise(allocatedDuration);
```
flyingsilverfin, line 264: we find that the SAT solver always restarts so we don't need to increase the allocated duration over multiple invocations

flyingsilverfin: And so we don't need to keep the variable at all.

common/optimiser/Optimiser.java
dmitrii-ubskii: Did you mean to set it to `false` here?

flyingsilverfin: oh shoot ahaha wow

common/optimiser/Optimiser.java
```diff
@@ -45,30 +53,49 @@ public class Optimiser {
...
 
-    private void recordValues() {
-        objectiveValue = solver.objective().value();
-        if (status != Status.NOT_SOLVED && status != Status.ERROR) variables.forEach(OptimiserVariable::recordSolutionValue);
+    private void maySetSolver() {
+        if (solver == null) createSolver();
+        else if (!isConstraintsUpToDate) {
+            solver.reset();
+            setConstraintCoefficients();
+        }
```
dmitrii-ubskii, line 86: We should reset `objectiveValue` to `null` here, otherwise down at line 90 we're comparing with the objective value based on outdated costs.

common/optimiser/Optimiser.java
dmitrii-ubskii: If the solver is deterministic, we might just be stuck in a Sisyphean loop here. Which isn't the problem with SAT solver, to my knowledge.

common/optimiser/Optimiser.java
```diff
@@ -189,7 +225,7 @@ static Status of(MPSolver.ResultStatus mpStatus) {
 
     @Override
     public String toString() {
-        return "Optimiser[" + "hasSolver=" + hasSolver + ", variables=" + variables.size() +
-                ", constraints=" + constraints.size() + "]" + (hasSolver ? solver.exportModelAsLpFormat() : "");
+        return "Optimiser[" + "hasSolver=" + (solver == null) + ", variables=" + variables.size() +
+                ", constraints=" + constraints.size() + "]" + (solver != null ? solver.exportModelAsLpFormat() : "");
```
dmitrii-ubskii, line 229: `hasSolver` ⇔ `solver != null`, which is correct in the second case, but not the first.

common/optimiser/OptimiserConstraint.java
```diff
@@ -23,36 +23,49 @@
...
+        this.optimiser = optimiser;
         this.lowerBound = lowerBound;
         this.upperBound = upperBound;
         this.coefficients = new HashMap<>();
         this.name = name;
     }
 
     public void setCoefficient(OptimiserVariable<?> variable, double coeff) {
-        coefficients.put(variable, coeff);
+        assert mpConstraint == null || coefficients.containsKey(variable);
```
dmitrii-ubskii, line 46: Why would adding a coefficient in a changed constraint be an error we want to assert against?

common/optimiser/Optimiser.java
flyingsilverfin: it's true - i'm thinking once we can background plan we can incrementally give it more time if it doesn't make progress somehow?

common/optimiser/Optimiser.java
flyingsilverfin: actually not required, because the Constraint coefficients don't affect the objective - the objective is updated only in `setObjectiveCoefficient` and/or nulled out

common/optimiser/Optimiser.java
```diff
@@ -83,67 +110,76 @@ public boolean isError() {
...
-            solver.reset();
-            setSolverModel();
-        } else initialiseSolver();
+    public void setConstraintsChanged() {
+        isConstraintsUpToDate = true;
     }
 
-    private void initialiseSolver() {
+    private void createSolver() {
         solver = MPSolver.createSolver("SAT");
```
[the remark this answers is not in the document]
flyingsilverfin: why do we need to change the behaviour here? Not sure I see the issue

common/optimiser/OptimiserConstraint.java
flyingsilverfin: i'm more strictly defining the type of solving we support as "The coefficients may be dynamic, but not the number of terms in constraints or objective"

common/optimiser/Optimiser.java
dmitrii-ubskii: Right, missed that.

common/optimiser/Optimiser.java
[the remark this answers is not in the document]
dmitrii-ubskii: Maybe we don't. It looks like SWIG relies on finalizers to free the heap.

dmitrii-ubskii:
Looks good to me!

dmitrii-ubskii (approved):

PR #6615, opened 2022-07-25
Reasoner tracing: Fix initialization race & handle case when directory does not exist

krishnangovindraj:
Fix bugs occurring whe reasoner tracing is used in multiple transactions during the same session.

The TraceWriter member used to be initialised lazily by the traceWriter method when first accessed. This has a concurrency bug which is triggered with multiple writers. Now it  is initialised in the constructor of Tracer.

Also, initialisation would fail when the output directory already exists. This is fixed with a check for directory existence.

### DIFF
--- a/reasoner/common/Tracer.java
+++ b/reasoner/common/Tracer.java
@@ -48,6 +48,7 @@ public class Tracer {
     public Tracer(long id, Path logDir) {
         this.id = id;
         this.logDir = logDir;
+        traceWriter();
     }
 
     private TraceWriter traceWriter() {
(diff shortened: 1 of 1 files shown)


reasoner/common/Tracer.java
```diff
@@ -149,7 +150,7 @@ private void start() {
             try {
                 LOG.debug("Writing reasoner traces to {}", path.get().toAbsolutePath());
                 File file = path.get().toFile();
-                if (!file.getParentFile().mkdirs()) {
+                if (!file.getParentFile().exists() && !file.getParentFile().mkdirs()) {
```
krishnangovindraj, line 153: mkdirs succeeds iff the directory was created (hence, fails if it already exists).
Adding a .exists() handles this.

jmsfltchr (approved):

flyingsilverfin (approved):

PR #6616, opened 2022-07-26
Enforce query planning time limit using background worker

flyingsilverfin:

Strictly enforce a time limit for query planning to create more predictable query execution latencies. Query planning using the OR-Tools SAT solver may not respect our fine-granularity time limits (eg. 100ms), and may overrun up several multiples of this before stopping. We now plan in a background thread, and use the an updated plan if it is available within the time limit, and otherwise fall back to the previous query plan.

* Query planning executes in a background thread using `CompletableFuture`
* Refactor `GraphPlanner.optimise()` to allow first-time optimisation or re-optimisation to utilise the same `Future`

### DIFF
--- a/traversal/planner/GraphPlanner.java
+++ b/traversal/planner/GraphPlanner.java
@@ -40,15 +40,18 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.locks.ReadWriteLock;
-import java.util.concurrent.locks.StampedLock;
 
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.UNEXPECTED_PLANNING_ERROR;
 import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+import static com.vaticle.typedb.core.concurrent.executor.Executors.async2;
 import static java.lang.Math.abs;
 import static java.time.Duration.between;
 import static java.util.Comparator.comparing;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 public class GraphPlanner implements Planner {
 
@@ -65,14 +68,13 @@ public class GraphPlanner implements Planner {
     private final Map<Identifier, PlannerVertex<?>> vertices;
     private final Set<PlannerEdge<?, ?>> edges;
     private final AtomicBoolean isOptimising;
-    private final ReadWriteLock firstOptimisingLock;
 
     protected volatile GraphProcedure procedure;
+    private volatile double totalCostLastRecorded;
     private volatile boolean isUpToDate;
     private volatile boolean isVertexOrderInitialised;
     private volatile long snapshot;
-
-    private volatile double totalCostLastRecorded;
+    private volatile CompletableFuture<Void> backgroundOptimisation;
     private double totalCost;
 
     private GraphPlanner() {
@@ -80,7 +82,6 @@ public class GraphPlanner implements Planner {
         vertices = new HashMap<>();
         edges = new HashSet<>();
         isOptimising = new AtomicBoolean(false);
-        firstOptimisingLock = new StampedLock().asReadWriteLock();
         isUpToDate = false;
         isVertexOrderInitialised = false;
         totalCostLastRecorded = INIT_ZERO;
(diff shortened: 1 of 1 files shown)


traversal/planner/GraphPlanner.java
```diff
@@ -65,9 +68,9 @@ public class GraphPlanner implements Planner {
     private final Map<Identifier, PlannerVertex<?>> vertices;
     private final Set<PlannerEdge<?, ?>> edges;
     private final AtomicBoolean isOptimising;
-    private final ReadWriteLock firstOptimisingLock;
 
     protected volatile GraphProcedure procedure;
+    private volatile CompletableFuture<Void> backgroundOptimisation;
```
flyingsilverfin, line 73: background planning task that may or may not be completed at any given time but can always be waited on with a time limit

traversal/planner/GraphPlanner.java
```diff
@@ -231,37 +233,43 @@ Optimiser optimiser() {
...
-            }
-        } else if (isOptimising.compareAndSet(false, true)) {
-            try {
-                optimise(graphMgr, singleUse);
-            } finally {
-                isOptimising.set(false);
-            }
+        long timeLimitMillis = singleUse ? HIGHER_TIME_LIMIT_MILLIS : DEFAULT_TIME_LIMIT_MILLIS;
+        if (backgroundOptimisation == null) beginFirstOptimise(graphMgr, timeLimitMillis);
+        else if (isOptimising.compareAndSet(false, true)) beginReOptimise(graphMgr, timeLimitMillis);
```
flyingsilverfin, line 238: if we have not started any optimisation yet, we start the first optimisation. Otherwise, only 1 thread is allowed to go and reset the future to run another optimisation.

The `isOptimising` AtomicBoolean is set back to `false` within the background optimisation task.

traversal/planner/GraphPlanner.java
```diff
@@ -231,37 +233,43 @@ Optimiser optimiser() {
...
+
+        try {
+            backgroundOptimisation.get(timeLimitMillis, MILLISECONDS);
```
flyingsilverfin, line 241: all threads wanting to use this plan give the background optimisation the required time to finish, and then return (this approach will immediately return if the optimisation is already done or as soon as it finishes, which is better than a `sleep()` that always consumes the maximum time)

traversal/planner/GraphPlanner.java
```diff
@@ -231,37 +233,43 @@ Optimiser optimiser() {
...
+            LOG.trace("Query plan did not finish updating in " + timeLimitMillis + " ms.");
+        }
+    }
+
+    private synchronized void beginFirstOptimise(GraphManager graphMgr, long timeLimitMillis) {
+        if (backgroundOptimisation == null) {
+            isOptimising.set(true);
+            updateTraversalCosts(graphMgr);
+            updateOptimiser();
+            createProcedure();
```
flyingsilverfin, line 252: create an initial `procedure` based on the greedy initialisation that is the fallback in case the first call to `optimise()` takes much longer than expected

traversal/planner/GraphPlanner.java
```diff
@@ -231,37 +233,43 @@ Optimiser optimiser() {
...
-        long allocatedDuration = singleUse ? HIGHER_TIME_LIMIT_MILLIS : DEFAULT_TIME_LIMIT_MILLIS;
-        Instant start, endSolver, end;
+        backgroundOptimisation = backgroundOptimisation.thenRunAsync(() -> optimise(timeLimitMillis), async2());
+    }
 
+    private void optimise(long timeLimitMillis) {
+        Instant start, endSolver, end;
         start = Instant.now();
-        optimiser.optimise(allocatedDuration);
+        optimiser.optimise(timeLimitMillis);
```
flyingsilverfin, line 272: the computation of the costs and `isUpToDate`-ness is checked before calling `optimise()` now. This method is purely responsible for running the optimiser, setting the state to `isUpToDate = true`, creating the new procedure, and finishing with `isOptimising = false`.

We could argue that we should set the `isOptimising = false` outside of this method, but inside the future's runnable.

traversal/planner/GraphPlanner.java
```diff
@@ -270,6 +278,7 @@ private void optimise(GraphManager graphMgr, boolean singleUse) {
 
         isUpToDate = true;
         printDebug(start, endSolver, end);
+        isOptimising.set(false);
```
flyingsilverfin, line 281: must release the boolean to allow another optimisation to be queued

traversal/planner/GraphPlanner.java
dmitrii-ubskii: Unless the solver converges to an optimal solution in well under the allotted time, the optimisation will almost never finish within this time limit, because updating the costs and creating the procedure takes on the order of milliseconds. I'd suggest increasing the timeout here.

flyingsilverfin: from my profiling updating traversal costs can take around 2-5ms, and procedure creation a few more ms. I'll 10ms to the overall time allocated.

haikalpribadi (approved):

traversal/planner/GraphPlanner.java
dmitrii-ubskii: That feels about right.

dmitrii-ubskii (approved):

traversal/planner/GraphPlanner.java
dmitrii-ubskii: But you didn't actually change anything, now the optimiser limit is 110 ms, and you wait for 110 ms. This line needs to be `timeLimitMillis + 10`

dmitrii-ubskii (changes requested):

PR #6617, opened 2022-08-01
Refactor reasoner oracle test rule partitioning

flyingsilverfin:

Simplify and restructure the rule partitioning (aka Stratification in the literature) approach implemented in the Oracle test framework. We improve readability by creating smaller methods and more canonical state.

* deleted derived/unnecessary state
* apply new naming to help the reader distinguish the state
* refactor revealed common code paths that could be deduplicated

### DIFF
--- a/test/behaviour/reasoner/verification/ForwardChainingMaterialiser.java
+++ b/test/behaviour/reasoner/verification/ForwardChainingMaterialiser.java
@@ -118,50 +118,46 @@ public class ForwardChainingMaterialiser {
     }
 
     private List<Set<Rule>> computeOrderedRuleEvaluationPartitions() {
-        // Computes the stratification of rules required by stratified-negation.
+        // Compute the stratification of rules required by stratified-negation.
         if (rulePartitions == null) {
             rulePartitions = new LinkedList<>();
-            Set<Rule> inLowerPartitions = new HashSet<>(); // Tracks nodes already added to some partition.
             for (Rule rule : rules.values()) {
-                if (!inLowerPartitions.contains(rule)) {
-                    Set<Rule> inLowerOrSamePartition = new HashSet<>();
-                    partitioningDFS(rule, inLowerPartitions, inLowerOrSamePartition);
-                    updatePartitions(inLowerPartitions, inLowerOrSamePartition);
-                }
+                if (!inPartition(rule)) computePartitionsFrom(rule);
             }
         }
         return rulePartitions;
     }
 
-    private void partitioningDFS(Rule at, Set<Rule> inLowerPartitions, Set<Rule> InLowerOrSamePartition) {
-        // If a rule is reachable through a path with a negated edge, it is in a strictly lower partition/stratum.
-        // If a rule is reachable only through paths with only positive edges, it is in a lower or equal partition/stratum.
-        if (inLowerPartitions.contains(at) || InLowerOrSamePartition.contains(at)) {
-            return;
-        } else {
-            InLowerOrSamePartition.add(at);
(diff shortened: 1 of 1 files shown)


test/behaviour/reasoner/verification/ForwardChainingMaterialiser.java
```diff
@@ -118,52 +118,48 @@ Optional<Materialisation> conditionMaterialisations(com.vaticle.typedb.core.logi
...
-        // Computes the stratification of rules required by stratified-negation.
+        // Compute the stratification of rules required by stratified-negation.
         if (rulePartitions == null) {
             rulePartitions = new LinkedList<>();
-            Set<Rule> inLowerPartitions = new HashSet<>(); // Tracks nodes already added to some partition.
             for (Rule rule : rules.values()) {
-                if (!inLowerPartitions.contains(rule)) {
-                    Set<Rule> inLowerOrSamePartition = new HashSet<>();
-                    partitioningDFS(rule, inLowerPartitions, inLowerOrSamePartition);
-                    updatePartitions(inLowerPartitions, inLowerOrSamePartition);
```
flyingsilverfin, line 129: push derived logic one method down, which also makes it re-usable

test/behaviour/reasoner/verification/ForwardChainingMaterialiser.java
```diff
@@ -118,52 +118,48 @@ Optional<Materialisation> conditionMaterialisations(com.vaticle.typedb.core.logi
...
-            for (Rule dependency : at.negatedDependencies()) {
-                if (!inLowerPartitions.contains(dependency)) {
-                    Set<Rule> inLowerOrSamePartitionAsDependency = new HashSet<>();
-                    partitioningDFS(dependency, inLowerPartitions, inLowerOrSamePartitionAsDependency);
-                    updatePartitions(inLowerPartitions, inLowerOrSamePartitionAsDependency);
-                }
-            }
+    private boolean inPartition(Rule rule) {
+        return iterate(rulePartitions).anyMatch(partition -> partition.contains(rule));
+    }
```
flyingsilverfin, line 133: replace separate `Set` with looping over the previously created partitions

test/behaviour/reasoner/verification/ForwardChainingMaterialiser.java
```diff
@@ -118,52 +118,48 @@ Optional<Materialisation> conditionMaterialisations(com.vaticle.typedb.core.logi
...
+        extractPartition(positivelyReachable);
+    }
+
+    private void partitionDependencies(Rule rule, Set<Rule> positivelyReachable) {
+        // Negated dependencies of a rule are strictly in a lower partition.
+        // Positive dependencies of a rule are in the current partition or a lower one.
+        rule.negatedDependencies().forEach(negatedDependency -> {
+            if (!inPartition(negatedDependency)) computePartitionsFrom(negatedDependency);
+        });
+        rule.unnegatedDependencies().forEach(dependency -> expandPartition(dependency, positivelyReachable));
```
flyingsilverfin, line 148: simplified the dependency partitioning - now it's very clear that we have two separate behaviours for negated dependencies and positive dependencies

krishnangovindraj:
This refactor describes the code in the properties of the nodes (connectedness) rather than the final result which the properties tell you (which stratum they are in). It also makes explicit the difference between steps into lower stratum (computePartitionFrom) and expanding the current stratum (partitionDependencies).
This makes it easier to follow what the code is doing, which we know will lead to the answer as post-condition.
The earlier approach was harder to decipher as it talked in-terms of the post-condition before it was achieved, leading to some confusion.

PR #6619, opened 2022-08-03
Sorted iterators do not require matched comparator and equality

flyingsilverfin:

We relax sorted iterators' implicit contract, which previously required that `equals() == true` iif `compareTo() == 0`. TypeDB's sorted iterators now require that when if `equals() == true`, then `compareTo() == 0`, but not vice versa. This allows using comparators that order based on a subset of the comparable elements' state.

Future uses of this includes sorted iterators of comparable `ConceptMap`s by particular variables, for example.

* remove `toLinkedSet()` API on `FunctionalIterator` since it can be reduced to the more minimal existing API: `toSet(Set<T> collector)`
* `Distinct` sorted iterators now keep a deduplication set as long as the next element has `compareTo(last) == 0`. This implementation means that when equality and comparator are perfectly aligned, the deduplication set never grows larger than a size of 1, while also handling the more general case of deduplicating repeated and but comparably equal values.
* `Intersection` iterator now performs the intersection _purely_ by `compareTo() == 0`. To preserve correctness, the intersection now returns ALL elements that are comparably equal but not equal by equality function.

For example:
Intersecting `[(1,2), (2,10)]` and `[(1,5), (1,2), (3,20)]`, comparing just by the first element, yields `[(1,2), (1,5)]`, where no ordering is guaranteed between these two elements.

### DIFF
--- a/common/iterator/sorted/DistinctSortedIterator.java
+++ b/common/iterator/sorted/DistinctSortedIterator.java
@@ -21,37 +21,87 @@ package com.vaticle.typedb.core.common.iterator.sorted;
 import com.vaticle.typedb.core.common.exception.TypeDBException;
 import com.vaticle.typedb.core.common.iterator.sorted.SortedIterator.Order;
 
+import java.util.HashSet;
 import java.util.NoSuchElementException;
+import java.util.Set;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_ARGUMENT;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
 
+/**
+ * Distinct iterator implemented first by comparator, then by equality.
+ *
+ * Assumes elements which are equal by comparison MAY be non-equal, but that
+ * all elements which are not-equal by comparator are NEVER equal.
+ *
+ * This leads to an efficient distinct iterator, which uses the comparator to cut the size of the deduplication set.
+ * When the comparator and equality are perfectly aligned, the space complexity is O(1).
+ */
 public class DistinctSortedIterator<T extends Comparable<? super T>, ORDER extends Order, ITER extends SortedIterator<T, ORDER>>
         extends AbstractSortedIterator<T, ORDER> {
 
     final ITER iterator;
+    final Set<T> equalByComparator;
     T last;
+    State state;
+
+    private enum State {INIT, EMPTY, FETCHED, COMPLETED}
 
     public DistinctSortedIterator(ITER iterator) {
         super(iterator.order());
         this.iterator = iterator;
+        this.equalByComparator = new HashSet<>();
         last = null;
+        state = State.INIT;
     }
 
     @Override
     public boolean hasNext() {
-        while (iterator.hasNext()) {
-            if (iterator.peek().equals(last)) iterator.next();
-            else return true;
+        switch (state) {
+            case INIT:
+                return fetchFirst();
+            case EMPTY:
+                return fetchNext();
+            case FETCHED:
+                return true;
+            case COMPLETED:
+                return false;
+            default:
+                throw TypeDBException.of(ILLEGAL_STATE);
         }
-        return false;
+    }
+
+    private boolean fetchFirst() {
+        if (iterator.hasNext()) {
+            equalByComparator.add(iterator.peek());
+            state = State.FETCHED;
+        } else state = State.COMPLETED;
+        return state == State.FETCHED;
+    }
+
+    private boolean fetchNext() {
+        while (iterator.hasNext()) {
+            T next = iterator.peek();
+            int comparison = next.compareTo(last);
+            if (comparison == 0 && equalByComparator.contains(next)) {
+                iterator.next();
+                continue;
+            }
+            if (comparison != 0) equalByComparator.clear();
+            equalByComparator.add(next);
+            state = State.FETCHED;
+            break;
+        }
+        return state == State.FETCHED;
     }
 
     @Override
     public T next() {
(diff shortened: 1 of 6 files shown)


common/iterator/sorted/IntersectForwardableIterator.java
```diff
@@ -21,40 +21,52 @@
...
         extends AbstractSortedIterator<T, ORDER>
         implements SortedIterator.Forwardable<T, ORDER> {
 
     private final List<Forwardable<T, ORDER>> iterators;
 
     private T candidate;
-    private int candidateSource;
+    private Forwardable<T, ORDER> candidateSource;
+    private final Set<T> intersectionValues;
+    private final List<Forwardable<T, ORDER>> intersectionIterators;
```
flyingsilverfin, line 50: keep track of iterators that are intersect at the last candidate, and the set of values they have produced.
`intersectionValues` ensures we do not return duplicates from the intersection.

common/iterator/sorted/IntersectForwardableIterator.java
```diff
@@ -64,17 +76,43 @@ public boolean hasNext() {
...
+            Forwardable<T, ORDER> iterator = intersectionIterators.get(0);
+            assert iterator.hasNext();
+            T next = iterator.peek();
+            if (isIntersection(candidate, next)) {
+                iterator.next();
+                if (!intersectionValues.contains(next)) {
+                    candidate = next;
+                    state = State.FETCHED;
+                    return true;
+                }
```
flyingsilverfin, line 91: find a new candidate from within this intersection

common/iterator/sorted/IntersectForwardableIterator.java
```diff
@@ -64,17 +76,43 @@ public boolean hasNext() {
...
+        intersectionValues.clear();
+        return false;
+    }
+
+    private boolean computeIntersection() {
+        assert state == State.INIT || state == State.EMPTY;
         Forwardable<T, ORDER> iterator = iterators.get(0);
         if (!iterator.hasNext()) state = State.COMPLETED;
         else {
-            candidateSource = 0;
```
flyingsilverfin, line 71: we replace an `int` index with the object it is indexing directly to simplify the understanding

common/iterator/sorted/DistinctSortedIterator.java
```diff
@@ -78,6 +129,7 @@ public Forwardable(SortedIterator.Forwardable<T, ORDER> source) {
         public void forward(T target) {
             if (last != null && !order.isValidNext(last, target)) throw TypeDBException.of(ILLEGAL_ARGUMENT);
             iterator.forward(target);
+            if (state == State.FETCHED) state = State.EMPTY;
```
flyingsilverfin, line 132: reset the state so we are forced to re-compute the intersection

common/iterator/sorted/IntersectForwardableIterator.java
```diff
@@ -112,12 +153,15 @@ public T peek() {
...
         return candidate;
     }
 
     @Override
     public void forward(T target) {
         iterators.forEach(iterator -> iterator.forward(target));
+        intersectionIterators.clear();
+        intersectionValues.clear();
         state = State.INIT;
     }
```
dmitrii-ubskii, line 166: This smells like trouble to me.
This next part is under an assumption that the order of the extra fields is not guaranteed, that is if an iterator is sorted by the first field, we could get `[(1,2), (1,5)]` or `[(1,5), (1,2)]`.
Say we have two iterators A `[..., (1,2), >(1,5), ...]` and B `[..., >(1,5), (1,2), ...]`, where `>` is the iterator's current position. This intersection was queried once, so the set of `intersectionValues` contains `(1,2)`, and both A and B are in `intersectionIterators`.
The iterator is forwarded to `(1,5)`. Both iterators are already pointing to `(1,5)`, so they don't move, but the deduplication info is lost. Next two times the iterator is queried, we get `(1,5)`, then `(1,2)`, despite the fact that `(1,2)` was already output before the `forward()` call.
I think a solution could be gating the two added lines behind an `isIntersection(candidate, target)` check.

dmitrii-ubskii: Upon reflection, the different order isn't required either. `[..., (1,2), >(1,5), ...]` and `[..., >(1,2), (1,5), ...]` would spell the same issue. In fact, we could be in a  `[..., (1,2), (1,5), >...]`, `[..., >(1,7) ...]`  state, in which case the first line of this method would be an error.

common/iterator/sorted/IntersectForwardableIterator.java
```diff
@@ -159,10 +159,15 @@ public T next() {
 
     @Override
     public void forward(T target) {
+        if (state == State.COMPLETED) return;
         iterators.forEach(iterator -> iterator.forward(target));
```
dmitrii-ubskii, line 163: I think this line needs to be under a check as well. If A is in `[(1,2), >(2,3), …]`, and B is in `[>(1,3), …]`, we should still be able to forward to `(1,3)`. Instead, this line will attempt to advance A from `(2,3)` to `(1,3)`, triggering an exception.

dmitrii-ubskii (changes requested):

common/iterator/sorted/IntersectForwardableIterator.java
flyingsilverfin: i see what you mean - if we're partway through an intersection containing many values (eg a large of tuples all starting with `(1, X))`, but some of the iterators then ready to start producing `(2, X)`. Forwarding them to `(1,X)` is a valid no-op, but will cause an exception in the iterators already placed at `(2,X)`!

dmitrii-ubskii (approved):
LGTM

flyingsilverfin:
High-level idea of intersection based on just the comparator is approved by @lolski  verbally, implementation approved by Dmitrii!

PR #6620, opened 2022-08-04
Refactor reasoner planner to be able to use statistics

krishnangovindraj:
We move the reasoner-planner to an architecture which enables recursive computation of costs through triggered rules.

We also implement a version which emulates the old local heuristic planner by returning appropriate dummy costs. This can serve as a baseline to compare against. 

A Plannable class maps the logical concepts to the planning domain - We consider the unit of planning to be a logical entity + the set of bound variables it receives. We capture this in a **Plannable** class.

We create an abstract **ReasonerPlanner** which serves as a coordinator the planning - allowing a Plannable to access the cost of the Plannables it recursively depends on. The actual planning algorithm is left to the subclasses to implement.  

Finally, we implement a GreedyAnswerSizeSearch subclass which (in the next phases) will be developed into the simplest prototype plan search based on a cost estimated from data statistics. At this point, the cost is set to 1 for any Plannable.

Porting the old planning algorithm to the new architecture, we implement a new ReasonerPlanner subclass - **OldPlannerEmulator** which returns the same result as the old-planning algorithm would. This will serve as a sanity-check and a good base-line in future, as well as allowing us to re-use the existing integration tests to cover some of the new code-paths.


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

reasoner/controller/ConditionController.java
```diff
@@ -45,11 +43,6 @@ public class ConditionController extends ConjunctionController<
         this.condition = condition;
     }
 
-    @Override
-    Set<Concludable> concludablesTriggeringRules() {
-        return condition.concludablesTriggeringRules(registry().conceptManager(), registry().logicManager());
-    }
-
```
krishnangovindraj, line 52: Removing this means the physical Resolvable instances returned by condition.concludablesTriggeringRules are different to the ones used by the reasoner.

reasoner/planner/ReasonerPlanner.java
```diff
@@ -0,0 +1,142 @@
...
+                    Set<Resolvable<?>> resolvables = new HashSet<>();
+                    resolvables.addAll(concludablesTriggeringRules);
+                    resolvables.addAll(Retrievable.extractFrom(conjunction.pattern(), concludablesTriggeringRules));
+                    resolvables.addAll(conjunction.negations());
+                    compiled.put(conjunction, resolvables);
+                }
+            }
+        }
+        return compiled.get(conjunction);
+    }
```
krishnangovindraj, line 75: Transaction bound mapping of conjunction -> concludables and retrievables

test/integration/reasoner/planner/PlannerTest.java
```diff
@@ -63,15 +65,37 @@ public class PlannerTest {
...
+        public Set<Resolvable<?>> compile(ResolvableConjunction conjunction){
+            return mockResolvables;
+        }
+
+        private static List<Resolvable<?>> planResolvables(Set<Resolvable<?>> resolvables, Set<Identifier.Variable.Retrievable> inputBounds) {
+            ReasonerPlannerTestWrapper planner = new ReasonerPlannerTestWrapper(transaction, resolvables);
+            return planner.planConjunction(planner.mockConjunction, inputBounds).planOrder();
+        }
+    }
+
```
krishnangovindraj, line 89: Override to mock compilation of conjunction to concludables & retrievables

reasoner/planner/ReasonerPlanner.java
```diff
@@ -0,0 +1,142 @@
...
+        Pair<ResolvableConjunction, Set<Identifier.Variable.Retrievable>> plannableKey = new Pair<>(conjunction, bounds);
+        if (!planCache.containsKey(plannableKey)) {
+            synchronized (planCache) {
+                if (!planCache.containsKey(plannableKey)) {
+                    planCache.put(plannableKey, planConjunction(conjunction, bounds));
+                }
+            }
+        }
+        return planCache.get(plannableKey);
+    }
```
krishnangovindraj, line 87: Cache plans to avoid recompilation

reasoner/planner/ReasonerPlanner.java
```diff
@@ -0,0 +1,142 @@
...
+        planCache = new HashMap<>();
+        this.traversalEng = traversalEng;
+        this.conceptMgr = conceptMgr;
+        this.logicMgr = logicMgr;
+        this.compiled = new HashMap<>();
+    }
+
+    public static ReasonerPlanner create(TraversalEngine traversalEng, ConceptManager conceptMgr, LogicManager logicMgr) {
+        return new GreedyCostSearch.OldPlannerEmulator(traversalEng, conceptMgr, logicMgr);
+    }
```
krishnangovindraj, line 57: Create the planner instance. Called by Reasoner

reasoner/planner/GreedyCostSearch.java
```diff
@@ -0,0 +1,101 @@
...
+        while (!remaining.isEmpty()) {
+            Optional<Resolvable<?>> nextResolvableOpt = remaining.stream()
+                    .filter(r -> dependenciesSatisfied(r, bounds, dependencies))
+                    .min(Comparator.comparing(r -> estimateCost(r, bounds)));
+
+            if (nextResolvableOpt.isEmpty()) {
+                nextResolvableOpt = remaining.stream()
+                        .min(Comparator.comparing(r -> estimateCost(r, bounds)));
+            }
+            assert nextResolvableOpt.isPresent();
```
krishnangovindraj, line 68: Needs to be revised.

krishnangovindraj (commented):
Add comments

test/integration/reasoner/planner/PlannerTest.java
```diff
@@ -180,12 +206,42 @@ public void test_planner_disconnected_conjunction() {
...
+        Set<Resolvable<?>> resolvables = new HashSet<>();
+        resolvables.add(negated);
+        resolvables.addAll(concludables);
+
+        List<Resolvable<?>> plan = ReasonerPlannerTestWrapper.planResolvables(resolvables, set());
+
+        assertEquals(plan.size(), 2);
+        assertEquals(plan.get(1), negated);
+    }
+
```
krishnangovindraj, line 244: Add a case to ensure the negation (as failure) is evaluated only after the unnegated dependency.

reasoner/planner/ReasonerPlanner.java
```diff
@@ -0,0 +1,144 @@
...
+
+        for (Resolvable<?> resolvable : resolvables) {
+            if (resolvable.isNegated()) {
+                for (Identifier.Variable.Retrievable v : resolvable.retrieves()) {
+                    if (unnegatedRefCount.getOrDefault(v, 0) > 0) {
+                        deps.get(resolvable).add(v);
+                    }
+                }
+            }
+        }
```
krishnangovindraj, line 126: If a variable in a negated appears unnegated in some concludable, the negated is dependent on that concludable.
Here, circular dependencies are impossible since negated cannot be 'generating' any variables

krishnangovindraj (commented):
Commented how we ensure negations are scheduled after their dependencies

reasoner/controller/ConjunctionController.java
```diff
@@ -79,23 +78,20 @@
...
 
     @Override
     protected void setUpUpstreamControllers() {
         assert resolvables.isEmpty();
-        Set<Concludable> concludables = concludablesTriggeringRules().toSet();
-        Set<Retrievable> retrievables = Retrievable.extractFrom(conjunction.pattern(), concludables);
-        resolvables.addAll(concludables);
-        resolvables.addAll(retrievables);
-        iterate(concludables).forEachRemaining(c -> {
+        resolvables.addAll(registry().planner().compile(conjunction));
```
flyingsilverfin, line 86: I think `compile()` should be in the `LogicManager`

reasoner/controller/ConjunctionController.java
```diff
@@ -107,12 +103,10 @@ protected void setUpUpstreamControllers() {
...
 
     List<Resolvable<?>> plan(Set<Variable.Retrievable> boundVariables) {
-        if (!plans.containsKey(boundVariables)) {
-            List<Resolvable<?>> plan = Planner.plan(resolvables, new HashMap<>(), boundVariables);
-            plan.addAll(conjunction.negations());
-            plans.put(boundVariables, plan);
-        }
-        return plans.get(boundVariables);
+        ReasonerPlanner.Plan<Resolvable<?>> planObj = registry().planner().getPlan(conjunction, boundVariables);
+        List<Resolvable<?>> newPlan = planObj.planOrder();
```
flyingsilverfin, line 107: we do we need to expose a `Plan` object at all? can we just get the `planOrder` directly?

reasoner/planner/ReasonerPlanner.java
```diff
@@ -0,0 +1,144 @@
...
+
+        for (Resolvable<?> resolvable : resolvables) {
+            if (resolvable.isNegated()) {
+                for (Identifier.Variable.Retrievable v : resolvable.retrieves()) {
+                    if (unnegatedRefCount.getOrDefault(v, 0) > 0) {
+                        deps.get(resolvable).add(v);
+                    }
+                }
+            }
```
flyingsilverfin, line 125: you can write the inner loop compactly with iterators:
`iterate(resolvable.retrieves()).filter(id -> unnegatedRefCount.getOrDefault(v, 0) > 0)).forEachRemaining(id -> deps.get(resolvable).add(id))`

flyingsilverfin: or something like that!

reasoner/planner/GreedyCostSearch.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /* TODO:
+     * AnswerSize -> For each resolvable: Add the number of calls + the answers produced
+     * AnswersProduced? TraversalEngine tells us this based on constraints in the rule-body
+     * #calls? Product of the cardinality of each variable in the bounds (This number reduces as you evaluate a new resolvable which further constrains the variable - but min or less?)ssssssss
+     * Effect? Schedule stronger constraints first.
+     */
+    @Override
+    protected Plan<Resolvable<?>> planConjunction(ResolvableConjunction conjunction, Set<Identifier.Variable.Retrievable> inputBounds) {
```
flyingsilverfin, line 50: this method feels too long - is there a logical way to partition it up into subroutines or multiple parts?

reasoner/planner/GreedyCostSearch.java
```diff
@@ -0,0 +1,102 @@
...
+                    .min(Comparator.comparing(r -> estimateCost(r, bounds)));
+
+            if (nextResolvableOpt.isEmpty()) {
+                nextResolvableOpt = remaining.stream()
+                        .filter(r -> !dependenciesSatisfied(r, bounds, dependencies) && !r.isNegated())
+                        .min(Comparator.comparing(r -> estimateCost(r, bounds)));
+            }
+            assert nextResolvableOpt.isPresent();
+            Resolvable<?> nextResolvable = nextResolvableOpt.get();
+            cost += estimateCost(nextResolvable, bounds); // TODO: Eliminate double work
```
flyingsilverfin, line 71: I think we should implement this TODO, we can discuss how we may be able to do that if it's not easy!
Note it's not just this place we do duplicate work: in the first `stream()` we may be doing the same resolvable + bounds multiple times.

flyingsilverfin: not sure that itself is solvable without a good equality function on Retrievables so maybe we just don't worry about it for now

reasoner/planner/GreedyCostSearch.java
```diff
@@ -0,0 +1,102 @@
...
+            remaining.remove(nextResolvable);
+            bounds.addAll(nextResolvable.retrieves());
+        }
+        assert resolvables.size() == orderedResolvables.size() && iterate(orderedResolvables).allMatch(r -> resolvables.contains(r));
+        return new Plan<>(orderedResolvables, cost);
+    }
+
+    abstract long estimateCost(Resolvable<?> resolvable, Set<Identifier.Variable.Retrievable> bounds);
+
+    public static class OldPlannerEmulator extends GreedyCostSearch {
```
flyingsilverfin, line 82: I assume that once we have the cost estimator implemented we'll get rid of this right?
I'd expect the `GreedyCostSearch` to merge into the `ReasonerPlanner` as well?

reasoner/planner/ReasonerPlanner.java
```diff
@@ -0,0 +1,144 @@
...
+        this.conceptMgr = conceptMgr;
+        this.logicMgr = logicMgr;
+        this.compiled = new HashMap<>();
+    }
+
+    public static ReasonerPlanner create(TraversalEngine traversalEng, ConceptManager conceptMgr, LogicManager logicMgr) {
+        return new GreedyCostSearch.OldPlannerEmulator(traversalEng, conceptMgr, logicMgr);
+    }
+
+    public Set<Resolvable<?>> compile(ResolvableConjunction conjunction) {
```
flyingsilverfin, line 59: move into LogicManager

reasoner/planner/ReasonerPlanner.java
```diff
@@ -0,0 +1,144 @@
...
+                        deps.get(resolvable).add(v);
+                    }
+                }
+            }
+        }
+
+        return deps;
+    }
+
+    public static class Plan<PLANELEMENT> {
```
flyingsilverfin, line 131: do we need this?

reasoner/planner/ReasonerPlanner.java
```diff
@@ -0,0 +1,144 @@
...
+                    resolvables.addAll(Retrievable.extractFrom(conjunction.pattern(), concludablesTriggeringRules));
+                    resolvables.addAll(conjunction.negations());
+                    compiled.put(conjunction, resolvables);
+                }
+            }
+        }
+        return compiled.get(conjunction);
+    }
+
+    public Plan<Resolvable<?>> getPlan(ResolvableConjunction conjunction, Set<Identifier.Variable.Retrievable> bounds) {
```
flyingsilverfin, line 77: We only really use `get` for real getters, and even then only sometimes (to match APIs that are available to the user on the client-side concept API). On the server side this would just be `plan(...)`

reasoner/planner/ReasonerPlanner.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+
+public abstract class ReasonerPlanner {
+    protected final Map<Pair<ResolvableConjunction, Set<Identifier.Variable.Retrievable>>, Plan<Resolvable<?>>> planCache;
```
flyingsilverfin, line 41: this should use our `CommonCache` with a sensible size limit, what do you think?

reasoner/planner/ReasonerPlanner.java
```diff
@@ -0,0 +1,144 @@
...
+    protected final ConceptManager conceptMgr;
+    protected final TraversalEngine traversalEng;
+    protected final LogicManager logicMgr;
+    private final HashMap<ResolvableConjunction, Set<Resolvable<?>>> compiled;
```
flyingsilverfin, line 45: also note that this will go into `LogicManager`. When you do that, this can just be `Map` rather than `HashMap`

test/integration/reasoner/planner/PlannerTest.java
```diff
@@ -63,15 +68,37 @@ public class PlannerTest {
...
+        @Override
+        public Set<Resolvable<?>> compile(ResolvableConjunction conjunction) {
+            return mockResolvables;
+        }
+
+        private static List<Resolvable<?>> planResolvables(Set<Resolvable<?>> resolvables, Set<Identifier.Variable.Retrievable> inputBounds) {
+            ReasonerPlannerTestWrapper planner = new ReasonerPlannerTestWrapper(transaction, resolvables);
+            return planner.planConjunction(planner.mockConjunction, inputBounds).planOrder();
+        }
```
flyingsilverfin, line 90: It feels strange to create this extra class just to be able to create a new `planResolvables(Set<Resolvable>)` entry point. This tells me that we should have a `package-private` method in `ReasonerPlanner` (or `GreedyCostSearch)` that does exactly this.

test/integration/reasoner/planner/PlannerTest.java
```diff
@@ -97,7 +124,7 @@ public void test_planner_retrievable_dependent_upon_concludable() {
         Retrievable retrievable = new Retrievable(resolvedConjunction("{ $c($b); }", logicMgr));
 
         Set<Resolvable<?>> resolvables = set(concludable, retrievable);
-        List<Resolvable<?>> plan = Planner.plan(resolvables, new HashMap<>(), set());
+        List<Resolvable<?>> plan = ReasonerPlannerTestWrapper.planResolvables(resolvables, set());
```
flyingsilverfin, line 127: then here to do the planning you ask the reasoner to do the planning:  `tx().reasoner().planner().planResolvables(...)`

test/integration/reasoner/planner/PlannerTest.java
```diff
@@ -180,12 +220,42 @@ public void test_planner_disconnected_conjunction() {
...
         Set<Resolvable<?>> resolvables = set(concludable, concludable2);
-        List<Resolvable<?>> plan = Planner.plan(resolvables, new HashMap<>(), set());
+        List<Resolvable<?>> plan = ReasonerPlannerTestWrapper.planResolvables(resolvables, set());
 
         assertEquals(2, plan.size());
         assertEquals(set(concludable, concludable2), set(plan));
     }
 
+    @Test
+    public void test_planner_negations_with_dependencies() {
```
flyingsilverfin, line 230: useful test!

flyingsilverfin (changes requested):

test/integration/reasoner/planner/PlannerTest.java
krishnangovindraj: Ah this reminds me.
A key convenience of having the ReasonerPlanner do the compilation was that the compilation method was easy to mock like this, and didn't have to update every test in this file to actually construct a ResolvableConjunction and then select out the interesting concludables from there.

reasoner/planner/ReasonerPlanner.java
krishnangovindraj: Already, or when we cache plans across transactions?

reasoner/planner/ReasonerPlanner.java
krishnangovindraj: Useful to keep plans and costs together

reasoner/planner/GreedyCostSearch.java
krishnangovindraj: Could. But for now I'd rather split the planner's interfacing & caching to be separate to the actual implementation of the search to be separate
I guess I could create a new class called PlanSearch and make it a member.

krishnangovindraj (commented):
Add comments

reasoner/planner/GreedyCostSearch.java
krishnangovindraj: That's ok. We can use object equality.
But this will probably be redundant as soon as we update the search.

reasoner/planner/ReasonerPlanner.java
flyingsilverfin: we can get rid of generic

reasoner/controller/ConjunctionController.java
```diff
@@ -107,12 +103,10 @@ protected void setUpUpstreamControllers() {
...
 
     List<Resolvable<?>> plan(Set<Variable.Retrievable> boundVariables) {
-        if (!plans.containsKey(boundVariables)) {
-            List<Resolvable<?>> plan = Planner.plan(resolvables, new HashMap<>(), boundVariables);
-            plan.addAll(conjunction.negations());
-            plans.put(boundVariables, plan);
-        }
-        return plans.get(boundVariables);
+        ReasonerPlanner.Plan<Resolvable<?>> planObj = registry().planner().plan(conjunction, boundVariables);
+        List<Resolvable<?>> newPlan = planObj.planOrder();
```
flyingsilverfin, line 107: `Planner.Plan` should be a private object used for the optimisation search, and not be exposed outwardly.

reasoner/planner/GreedyCostSearch.java
flyingsilverfin: it will be nested in the future

reasoner/planner/ReasonerPlanner.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.Optional;
+import java.util.Set;
+
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+
+public abstract class ReasonerPlanner {
+    protected final ConceptManager conceptMgr;
+    protected final TraversalEngine traversalEng;
+    protected final LogicManager logicMgr;
+    protected final CommonCache<Pair<ResolvableConjunction, Set<Identifier.Variable.Retrievable>>, Plan<Resolvable<?>>> planCache;
```
flyingsilverfin, line 45: these can be `package-private` if used from a subclass in the same package

logic/LogicManager.java
```diff
@@ -56,12 +59,14 @@ public class LogicManager {
     private final ConceptManager conceptMgr;
     private final TypeInference typeInference;
     private final LogicCache logicCache;
+    private final Map<ResolvableConjunction, Set<Resolvable<?>>> compiled;
```
flyingsilverfin, line 62: `compiled` is not enough to provide the reader an indication of what the compilation is about... `compiledConjunctions` might be better since it implies it's "many" and to do with "conjunctions", so the name is a bit more standalone

flyingsilverfin (approved):

PR #6622, opened 2022-08-08
Refactor traversal engine to handle overlapping role edges

flyingsilverfin:

We fix the traversal behaviour under queries like `(friend: $x, friend: $y)` when the role players `$x` and `$y` actually are the same instance, and queries like `(friend: $x, role: $x)` where the relation queries overlapping sets of roles.

* Remove a lot of complexity of the traversal engine by de-optimising relation traversals that would generate partially overlapping sets of role instances. We can still handle exactly equal sets of role types optimally `(friend: $x, friend: $y)`, which is a common query. Thus, we only de-optimise rare queries like `(friend: $x, role: $y)` where the players are neither playing mutually exclusive nor identical role types.
* Simplify the traversal engine by filtering occupied roles out within the edge iterators, rather than generating and checking the occupied roles later. This allows correctness when combined with the above assumption.
* Encode implicit vertex dependencies in the Traversal engine to force backtracking between vertices that would otherwise not be directly depending on each other, such as two role players that are traversed after visiting the relation.
* Remove traversal engine lookaheads, as they overly complicated the role filtering design above
* Push `repetition` of a role player into all the role player `TraversalEdge` (`Procedure` and `Planner` variants) to be able to distinguish parallel edges from the same start and end vertex in the traversal engine

* optimise hot code paths by removing redundant operations and object construction
* remove outdated tests


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

traversal/planner/PlannerEdge.java
```diff
@@ -1194,11 +1194,14 @@ void computeCost(GraphManager graphMgr) {
             public static class RolePlayer extends Thing {
 
                 private final Set<Label> roleTypes;
+                private final int repetition;
 
-                RolePlayer(PlannerVertex.Thing from, PlannerVertex.Thing to, Set<Label> roleTypes) {
+                RolePlayer(PlannerVertex.Thing from, PlannerVertex.Thing to, int repetition, Set<Label> roleTypes) {
```
flyingsilverfin, line 1199: pass `repetition` into the equality of a RolePlayer edge in all traversal edges

test/integration/traversal/TraversalTest.java
```diff
@@ -186,7 +186,7 @@ public void mixed_variable_and_edge_role_players() {
             proc.forwardRelating(_0, role_inst);
             proc.backwardPlaying(role_inst, x);
             proc.forwardIsa(role_inst, role, true);
-            proc.forwardRolePlayer(_0, e, set(Label.of("employer", "employment")));
+            proc.forwardRolePlayer(_0, e, 0, set(Label.of("employer", "employment")));
```
flyingsilverfin, line 189: * rolePlayer edge now requires `repetition` number to construct to differentiate parallel edges

test/integration/traversal/TraversalTest.java
```diff
@@ -633,11 +633,11 @@ public void test_closure_backtrack_clears_scopes() {
             proc.backwardHas(n, x);
             proc.backwardHas(n, y);
             proc.forwardHas(x, _0);
-            proc.backwardRolePlayer(x, refl, set(Label.of("friend", "friendship")));
-            proc.backwardRolePlayer(x, refl, set(Label.of("friend", "friendship")));
-            proc.backwardRolePlayer(x, f1, set(Label.of("friend", "friendship")));
+            proc.backwardRolePlayer(x, refl, 0, set(Label.of("friend", "friendship")));
+            proc.backwardRolePlayer(x, refl, 1, set(Label.of("friend", "friendship")));
```
flyingsilverfin, line 637: parallel edges need to be differentiated

traversal/procedure/ProcedureEdge.java
```diff
@@ -1075,11 +1076,13 @@ public boolean isClosure(GraphManager graphMgr, Vertex<?, ?> fromVertex, Vertex<
 
             public static abstract class RolePlayer extends Thing {
 
+                final int repetition;
                 final Set<Label> roleTypes;
 
                 private RolePlayer(ProcedureVertex.Thing from, ProcedureVertex.Thing to,
-                                   Encoding.Direction.Edge direction, Set<Label> roleTypes) {
+                                   Encoding.Direction.Edge direction, int repetition, Set<Label> roleTypes) {
```
flyingsilverfin, line 1083: again, pass in `repetition` number into role player traversal edges

traversal/procedure/ProcedureEdge.java
```diff
@@ -1114,25 +1117,30 @@ public RolePlayer asRolePlayer() {
...
+
                 public Identifier.Variable scope() {
                     if (direction().isForward()) return from.id().asVariable();
                     else return to.id().asVariable();
                 }
 
                 @Override
                 public boolean equals(Object o) {
-                    return super.equals(o) && ((RolePlayer) o).roleTypes.equals(roleTypes);
+                    return super.equals(o) && ((RolePlayer) o).repetition == repetition &&
```
flyingsilverfin, line 1131: updated equality function

traversal/scanner/GraphIterator.java
```diff
@@ -349,8 +349,9 @@ private void reset() {
                 else if (procedureVertex.isStartingVertex()) iterator = createIteratorFromStart();
                 else {
                     iterator = createIteratorFromEdges();
-                    lastIntersection().ifPresent(value -> iterator.forward(value));
+                    cachedIntersection().ifPresent(value -> iterator.forward(value));
                 }
+                // TODO: we may only need to find one valid answer if all dependents are not included in the filter and also find an answer
```
flyingsilverfin, line 354: future optimisation we can apply

traversal/scanner/GraphIterator.java
```diff
@@ -377,18 +378,72 @@ private void reset() {
...
-            else return intersect(iterate(iters), ASC);
+            Map<Scopes.Scoped, List<ProcedureEdge.Native.Thing.RolePlayer>> rpEdgesByScope = new HashMap<>();
+            List<ProcedureEdge<?, ?>> otherEdges = new ArrayList<>();
+            edgeInputs.forEach((edge, vertex) -> {
+                if (edge.isRolePlayer()) {
+                    Scopes.Scoped scoped = scopes.getOrInitialise(edge.asRolePlayer().scope());
+                    rpEdgesByScope.computeIfAbsent(scoped, s -> new ArrayList<>()).add(edge.asRolePlayer());
+                } else otherEdges.add(edge);
+            });
+            return intersectEdges(rpEdgesByScope, otherEdges);
```
flyingsilverfin, line 389: we are doing the edge iterator intersection in two layers:
first we intersect RolePlayer edges using pairs of `(relation, role)`, which we guarantee to not occuply the same role across parallel edges. After this intersection we filter down to just the relations.

Then we intersect the remaining edges with the intersection of the role player edges.

traversal/scanner/GraphIterator.java
```diff
@@ -401,30 +456,29 @@ private void reset() {
...
-                //       an entire tree, where none of the leaves are referenced by name
-                toIter = toIter.limit(1);
-            }
 
             // TODO: this cast is a tradeoff between losing a lot of type safety in the ProcedureEdge branch() return type vs having a cast
             return (Forwardable<Vertex<?, ?>, Order.Asc>) toIter;
         }
 
+        private Forwardable<KeyValue<ThingVertex, ThingVertex>, Order.Asc> branch(Vertex<?, ?> fromVertex,
+                                                                                  ProcedureEdge.Native.Thing.RolePlayer edge) {
```
flyingsilverfin, line 468: split out branching of RolePlayer edges separately

pattern/constraint/thing/RelationConstraint.java
```diff
@@ -94,20 +94,14 @@ public void addTo(GraphTraversal.Thing traversal) {
...
+                Identifier.Scoped role = Identifier.Scoped.of(owner.id(), rolePlayer.roleType().get().id(), player.id(), rolePlayer.repetition());
                 traversal.relating(owner.id(), role);
                 traversal.playing(player.id(), role);
                 traversal.isa(role, rolePlayer.roleType().get().id());
                 traversal.types(role, rolePlayer.inferredRoleTypes());
-            } else if (rep > 1) {
-                Identifier.Scoped role = Identifier.Scoped.of(owner.id(), null, player.id(), rep);
-                traversal.relating(owner.id(), role);
-                traversal.playing(player.id(), role);
-                traversal.types(role, rolePlayer.inferredRoleTypes());
```
flyingsilverfin, line 108: we no longer need to de-optimise parallel edges since the traversal engine handles the more general case:

ie. `(friend: $x, friend: $x)` can be treated as two optimised traversal edges, because the traversal has to handle the case where `(friend: $x, friend: $y)` concidentally ends up with `$x` and `$y` being the same concept.

traversal/scanner/GraphIterator.java
```diff
@@ -377,18 +377,70 @@ private void reset() {
         }
 
         private Forwardable<Vertex<?, ?>, Order.Asc> createIteratorFromEdges() {
-            List<Forwardable<Vertex<?, ?>, Order.Asc>> iters = new ArrayList<>();
-            edgeInputs.forEach((procedureEdge, vertex) -> iters.add(branch(vertex, procedureEdge)));
-            if (iters.size() == 1) return iters.get(0);
-            else return intersect(iterate(iters), ASC);
+            Map<Vertex<?, ?>, List<ProcedureEdge.Native.Thing.RolePlayer>> overlappingRPEdges = new HashMap<>();
+            List<ProcedureEdge<?, ?>> otherEdges = new ArrayList<>();
```
flyingsilverfin, line 381: to handle parallel/overlapping role player edges from the same vertex, we intersect them separately to make sure the roles they occupy are valid.

traversal/scanner/GraphIterator.java
```diff
@@ -377,18 +377,70 @@ private void reset() {
...
+        private Forwardable<Vertex<?, ?>, Order.Asc> intersectEdges(Map<Vertex<?, ?>, List<ProcedureEdge.Native.Thing.RolePlayer>> rpEdges,
+                                                                    List<ProcedureEdge<?, ?>> otherEdges) {
+            List<Forwardable<Vertex<?, ?>, Order.Asc>> iterators = new ArrayList<>();
+            otherEdges.forEach(e -> iterators.add(branch(edgeInputs.get(e), e)));
+            rpEdges.forEach((vertex, edges) -> {
+                if (edges.size() == 1) {
+                    iterators.add(branch(edgeInputs.get(edges.get(0)), edges.get(0))
+                            .mapSorted(KeyValue::key, thing -> KeyValue.of(thing.asThing(), null), ASC));
+                } else iterators.add(intersectOverlapping(edges));
+            });
```
flyingsilverfin, line 399: normal edges or role player edges that are not parallel are simply added to the list of iterators. If there are multiple parallel role player edges, we handle them separately

traversal/scanner/GraphIterator.java
```diff
@@ -377,18 +377,70 @@ private void reset() {
...
+                    .forEach(e -> {
+                        // rare cases have RP edges in different directions from the same vertex, so look up scope each time
+                        Scopes.Scoped scoped = scopes.getOrInitialise(e.scope());
+                        iterators.add(branch(edgeInputs.get(e), e)
+                                .filter(thingAndRole -> isRoleAvailable(thingAndRole.value(), e, scoped, edges))
+                                .mapSorted(KeyValue::key, v -> KeyValue.of(v.asThing(), null), ASC));
+                    });
+            // WARN: we rely on the intersection operating in order of the iterators provided...
+            return intersect(iterate(iterators), ASC);
+        }
```
flyingsilverfin, line 424: the current solution to handling role player edges from the same start.
We have definitely handled the `identical` and `disjoint` cases perfectly with this approach. However, the subset/superset is only iffy since we rely on the intersection operating a specific way...

we can improve the solution for this case somehow...

dmitrii-ubskii: To summarise the in-person discussion: the subset case relies on the intersection iterator advancing the iterators in the order provided within the intersection. Put simply, we want to exclude the subset's values from the superset, not the other way around. This is the cleanest solution we came up with in the course of the discussion.

traversal/procedure/ProcedureVertex.java
```diff
@@ -142,11 +155,20 @@ public Thing asThing() {
...
-        public Set<ProcedureEdge<?, ?>> inRolePlayers() {
-            if (inRolePlayers == null) {
-                inRolePlayers = iterate(ins()).filter(ProcedureEdge::isRolePlayer).toSet();
+        public boolean isScope() {
+            if (isScope == null) {
+                isScope = iterate(outs()).anyMatch(e -> e.direction().isForward() && (e.isRolePlayer() || e.isRelating())) ||
+                        iterate(ins()).anyMatch(e -> e.direction().isBackward() && (e.isRolePlayer() || e.isRelating()));
+            }
+            return isScope;
+        }
```
flyingsilverfin, line 164: a vertex with `RolePlayer` or `Relating` edges in certain directions mandates that the vertex is a Scope, ie. only generates Relations.

traversal/scanner/GraphIterator.java
```diff
@@ -84,15 +82,55 @@ public GraphIterator(GraphManager graphMgr, Vertex<?, ?> initial, GraphProcedure
...
+
+    private void setup() {
+        // set up scopes
+        iterate(procedure.vertices()).filter(v -> v.isThing() && v.asThing().isScope())
+                .forEachRemaining(v -> scopes.put(v.id().asVariable(), new Scope()));
+        // set up traversers
+        iterate(procedure.vertices()).forEachRemaining(v -> vertexTraversers.put(v, new VertexTraverser(v)));
+        // add implicit dependencies between traversers
+        iterate(procedure.vertices()).filter(v -> v.isThing() && v.asThing().isScope())
+                .forEachRemaining(this::setupImplicitDependencies);
```
flyingsilverfin, line 101: prepare Scopes, Traversers, and Implicit dependencies between vertices

traversal/scanner/GraphIterator.java
```diff
@@ -176,82 +214,29 @@ private boolean computeAnswer() {
...
-    private void removeFromSuccessors(ProcedureVertex<?, ?> procedureVertex) {
-        procedureVertex.outs().forEach(edge -> {
-            vertexTraversers.get(edge.to()).removeEdgeInput(edge);
-            removeFromSuccessors(edge.to());
-        });
+    private void success(VertexTraverser vertexTraverser) {
+        iterate(vertexTraverser.procedureVertex.dependents()).link(iterate(vertexTraverser.implicitDependents))
+                .forEachRemaining(v -> {
+                    toTraverse.add(v);
+                    vertexTraversers.get(v).clear();
```
flyingsilverfin, line 225: when a traverser finds a vertex, we add its direct dependents & implicit dependents to the set of vertices to traverse next

traversal/scanner/GraphIterator.java
```diff
@@ -176,82 +214,29 @@ private boolean computeAnswer() {
...
+                });
     }
 
-    private void failed(ProcedureVertex<?, ?> procedureVertex) {
-        procedureVertex.ins().forEach(edge -> toRevisit.add(edge.from()));
-        vertexTraversers.get(procedureVertex).reset();
-        toTraverse.add(procedureVertex);
+    private void failed(VertexTraverser vertexTraverser) {
```
flyingsilverfin, line 229: on failure to find yet another vertex, we note all the predecessors that could have caused the failure (eg. dependees and implicit dependees) in the `toRevisit` set and backtrack

traversal/scanner/GraphIterator.java
```diff
@@ -176,82 +214,29 @@ private boolean computeAnswer() {
...
+        iterate(vertexTraverser.procedureVertex.dependees()).forEachRemaining(dependee -> toRevisit.add(dependee));
+        vertexTraverser.implicitDependees.forEach(toRevisit::add);
+        vertexTraverser.clear();
+        toTraverse.add(vertexTraverser.procedureVertex);
```
flyingsilverfin, line 233: we have to remember to come back here! It's possible that there are no dependees but we have to come here anyway. (eg. multiple start points)

traversal/scanner/GraphIterator.java
```diff
@@ -176,82 +214,29 @@ private boolean computeAnswer() {
...
-            return verifyScopedUpToOrder(scoped, procedureVertex.order(), verifyFailureCauses);
-        } else if (procedureVertex.isThing() && !procedureVertex.asThing().inRolePlayers().isEmpty()) {
-            for (ProcedureEdge<?, ?> edge : procedureVertex.asThing().inRolePlayers()) {
-                assert edge.isRolePlayer();
-                Scopes.Scoped scoped = scopes.get(edge.asRolePlayer().scope());
-                if (!verifyScopedUpToOrder(scoped, procedureVertex.order(), verifyFailureCauses)) return false;
-            }
-        }
-        return true;
-    }
```
flyingsilverfin, line 223: scopes are now purely handled by filtering & vertex backtracking. The latter handles all complex cases when considered with the fact that we guarantee de-optimisation of partially overlapping role traversals

traversal/scanner/GraphIterator.java
```diff
@@ -176,82 +214,29 @@ private boolean computeAnswer() {
...
-            toVertexTraverser.addEdgeInput(edge, vertexTraverser.vertex());
-            verified.add(edge);
-            if (!toVertexTraverser.hasNextVertex()) {
-                verified.forEach(e -> vertexTraversers.get(e.to()).removeEdgeInput(e));
-                toVertexTraverser.edgeInputs().forEach(e -> verifyFailureCauses.add(e.from()));
-                return false;
-            }
-        }
-        return true;
-    }
```
flyingsilverfin, line 249: we no longer do an edge lookahead because it made the role filtering & backtracking overly difficult and burdensome. The engine is now more intuitive: when we traverse a vertex and need to create an iterator, we look at all the input edges and create an iterator out of them at that point.

traversal/scanner/GraphIterator.java
```diff
@@ -176,82 +214,29 @@ private boolean computeAnswer() {
...
-
     private void revisit(ProcedureVertex<?, ?> procedureVertex) {
-        removeFromSuccessors(procedureVertex);
-        iterate(procedureVertex.transitiveOuts()).map(ProcedureEdge::to).forEachRemaining(toTraverse::remove);
         toTraverse.add(procedureVertex);
+        vertexTraversers.get(procedureVertex).implicitDependents.forEach(implicit -> vertexTraversers.get(implicit).clear());
```
flyingsilverfin, line 239: when re-visiting a vertex to generate the next one, we need to release all the vertices/scoped roles held by the implicit dependents. So for example:
```
1) $r --RP--> $x
2) $r --RP--> $y
```
when we backtrack to the next $x, we must release the role held by $y to guarantee `$x` will find all valid $x's. This is captured by the implicit dependencies

traversal/scanner/GraphIterator.java
```diff
@@ -263,34 +248,32 @@ public void recycle() {
     private class VertexTraverser {
 
         private final ProcedureVertex<?, ?> procedureVertex;
-        private final Map<ProcedureEdge<?, ?>, Vertex<?, ?>> edgeInputs;
-        private final SortedMap<Map<ProcedureEdge<?, ?>, Vertex<?, ?>>, Vertex<?, ?>> edgeInputIntersectionCache;
+        private final Scope localScope;
```
flyingsilverfin, line 251: cached local scope, if this Traverser is traversing a scoped ProcedureVertex

traversal/scanner/GraphIterator.java
```diff
@@ -263,34 +248,32 @@ public void recycle() {
...
+            this.implicitDependees = new HashSet<>();
         }
 
-        private boolean hasNextVertex() {
-            return getIterator().hasNext();
+        private boolean findNextVertex() {
+            Forwardable<Vertex<?, ?>, Order.Asc> iterator = getIterator();
+            while (iterator.hasNext()) {
+                vertex = getIterator().next();
+                if (verifyLoops()) return true;
```
flyingsilverfin, line 268: verify loops within the `Traverser`, and only return `true` if we both find a vertex & the loop edges are satisfied

traversal/scanner/GraphIterator.java
```diff
@@ -301,200 +284,146 @@ private boolean verifyLoops() {
...
             }
+            clearCurrentVertex();
+            if (procedureVertex.id().isScoped()) localScope.remove(procedureVertex);
+            else {
+                iterate(procedureVertex.ins()).filter(ProcedureEdge::isRolePlayer).forEachRemaining(e -> {
+                    Scope scope = scopes.get(e.asRolePlayer().scope());
+                    scope.remove(e);
+                });
+            }
+        }
```
flyingsilverfin, line 307: `reset()`clears the iterator and the vertex, along with unscoping any scoped roles.

traversal/scanner/GraphIterator.java
```diff
@@ -301,200 +284,146 @@ private boolean verifyLoops() {
...
-                Identifier.Variable scope = edge.to().id().asScoped().scope();
-                Scopes.Scoped scoped = scopes.getOrInitialise(scope);
-                toIter = edge.branch(graphMgr, fromVertex, params).mapSorted(
-                        role -> {
-                            scoped.record(edge.to(), role.asThing().asThing());
-                            return role;
-                        },
-                        role -> role,
-                        ASC
-                );
```
flyingsilverfin, line 403: this has become `applyLocalsScope(edge.branch())`

traversal/scanner/GraphIterator.java
```diff
@@ -301,200 +284,146 @@ private boolean verifyLoops() {
...
-                Identifier.Variable scope = edge.asRolePlayer().scope();
-                Scopes.Scoped scoped = scopes.getOrInitialise(scope);
-                toIter = edge.asRolePlayer().branchEdge(graphMgr, fromVertex, params).mapSorted(
-                        thingAndRole -> {
-                            scoped.record(edge, thingAndRole.value());
-                            return thingAndRole.key();
-                        },
-                        key -> KeyValue.of(key, null),
-                        ASC
-                );
```
flyingsilverfin, line 414: this has become `applyEdgeScope(edge.branch())`

traversal/scanner/GraphIterator.java
```diff
@@ -301,200 +284,146 @@ private boolean verifyLoops() {
...
+                        scope.record(edge, thingAndRole.value());
+                        return thingAndRole.key();
+                    },
+                    thing -> KeyValue.of(thing.asThing(), null),
+                    ASC
+            );
+        }
     }
 
-    public static class Scopes {
```
flyingsilverfin, line 430: got rid of the nested scoping structure, we just have a `Scope` object that corresponds to the roles held by vertex/edges within on Relation.

traversal/structure/Structure.java
```diff
@@ -51,7 +51,7 @@ public Structure() {
 
     public StructureVertex.Thing thingVertex(Identifier identifier) {
         return vertices.computeIfAbsent(identifier, id -> {
-            StructureVertex.Thing v = new StructureVertex.Thing(this, id);
+            StructureVertex.Thing v = new StructureVertex.Thing(id);
```
flyingsilverfin, line 54: remove unneeded input argument

traversal/scanner/GraphIterator.java
```diff
@@ -176,82 +214,29 @@ private boolean computeAnswer() {
...
-            } else {
-                vertexTraverser.clearCurrentVertex();
-                removeFromSuccessors(procedureVertex);
-            }
-        }
-        toRevisit.addAll(verifyFailureCauses);
-        if (verified) procedureVertex.outs().forEach(edge -> toTraverse.add(edge.to()));
-        else failed(procedureVertex);
+        if (vertexTraverser.findNextVertex()) success(vertexTraverser);
+        else failed(vertexTraverser);
```
flyingsilverfin, line 218: vertexTraversers now do a search internally, rather than having the `while` loop externally

traversal/scanner/GraphIterator.java
```diff
@@ -301,200 +284,146 @@ private boolean verifyLoops() {
...
-            }
-            edgeInputs.put(edge, fromVertex);
-            vertex = null;
-        }
-
-        private void removeEdgeInput(ProcedureEdge<?, ?> edge) {
-            edgeInputIntersectionCache.remove(edgeInputs);
-            edgeInputs.remove(edge);
-            reset();
-        }
```
flyingsilverfin, line 325: we no longer have to "activate" and "deactivate" different input edges to use when traverse this vertex - by the time we arrive at traversing a vertex, all predecessors/dependees must have been explored & valid! This is much simpler to understand

traversal/scanner/GraphIterator.java
dmitrii-ubskii: I assume you meant `clear()`

traversal/scanner/GraphIterator.java
```diff
@@ -301,200 +284,146 @@ private boolean verifyLoops() {
...
-
-        private void reset() {
+        void clear() {
             if (iterator != null) {
                 iterator.recycle();
                 iterator = null;
             }
+            clearCurrentVertex();
+            if (procedureVertex.id().isScoped()) localScope.remove(procedureVertex);
```
dmitrii-ubskii, line 300: The condition should probably just check that `localScope` is not `null`.

traversal/scanner/GraphIterator.java
```diff
@@ -301,200 +284,146 @@ private boolean verifyLoops() {
...
-                Scopes.Scoped scoped = scopes.getOrInitialise(scope);
-                toIter = edge.branch(graphMgr, fromVertex, params).mapSorted(
-                        role -> {
-                            scoped.record(edge.to(), role.asThing().asThing());
-                            return role;
-                        },
-                        role -> role,
-                        ASC
-                );
+            if (procedureVertex.id().isScoped()) {
```
dmitrii-ubskii, line 347: Also can be `localScope != null`

dmitrii-ubskii (changes requested):
It certainly looks like `GraphIterator` is a lot more streamlined now! Outside of a few minor kinks, looks pretty good.

traversal/procedure/ProcedureVertex.java
```diff
@@ -88,27 +90,38 @@ public int order() {
...
+    void setOrder(int order) {
+        this.order = order;
+    }
+
+    public Set<ProcedureVertex<?, ?>> dependents() {
+        if (dependents == null) {
+            Set<ProcedureVertex<?, ?>> vertices = new HashSet<>();
+            outs().forEach(e -> vertices.add(e.to()));
+            dependents = vertices;
         }
```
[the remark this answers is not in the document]
flyingsilverfin: this is because this `Procedure` classes are shared across threads. To avoid the possibility of race conditions and not require locking, we first in each thread collect to a set and then do an atomic assignment operation (assignment is always atomic in Java), where one writer will win. 
Technically because set addition is idempotent we could avoid doing this if the set we use also supports concurrent writers - however, using a `ConcurrentSet` will also incur read time overheads we don't want to pay.

traversal/procedure/ProcedureVertex.java
```diff
@@ -88,27 +90,38 @@ public int order() {
...
-            outs().forEach(edge -> {
-                transitive.add(edge);
-                transitive.addAll(edge.to().transitiveOuts());
-            });
-            transitiveOuts = transitive;
+    public Set<ProcedureVertex<?, ?>> dependees() {
+        if (dependees == null) {
+            Set<ProcedureVertex<?, ?>> vertices = new HashSet<>();
+            ins().forEach(e -> vertices.add(e.from()));
+            dependees = vertices;
```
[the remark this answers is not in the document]
flyingsilverfin: same as above

traversal/procedure/ProcedureVertex.java
```diff
@@ -88,27 +90,38 @@ public int order() {
...
         }
-        return transitiveOuts;
+        return dependees;
     }
 
-    void setOrder(int order) {
-        this.order = order;
+    public Set<ProcedureVertex<?, ?>> transitiveDependees() {
```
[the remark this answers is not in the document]
flyingsilverfin: good point!

common/iterator/BaseIterator.java
```diff
@@ -18,31 +18,28 @@
...
-import com.vaticle.typedb.common.collection.Either;
-
 import java.util.Iterator;
 
 class BaseIterator<T> extends AbstractFunctionalIterator<T> {
 
-    private final Either<FunctionalIterator<T>, Iterator<T>> iterator;
+    private final Iterator<T> iterator;
 
-    BaseIterator(Either<FunctionalIterator<T>, Iterator<T>> iterator) {
```
flyingsilverfin, line 29: profiling suggested that making the `BaseIterator` even a little bit more complex than it had to could have an impact on performance since it's one of the hottest code paths in the TypeDB! Here we simplify it here to remove an API that we never used: a `BaseIterator` wrapping a `Functionalterator`

traversal/procedure/ProcedureEdge.java
```diff
@@ -1157,25 +1195,23 @@ public Forwardable<KeyValue<ThingVertex, ThingVertex>, Order.Asc> branchEdge(Gra
...
-                            );
+                            iter = instanceRoleTypes.flatMap(
+                                    rt -> {
+                                        return iterate(to.props().types())
+                                                .map(l -> graphMgr.schema().getType(l))
+                                                .map(t -> rel.outs()
+                                                        .edge(ROLEPLAYER, rt, PrefixIID.of(t.encoding().instance()), t.iid())
+                                                        .toAndOptimised()
+                                                );
+                                    }).mergeMap(Function.identity(), ASC);
```
flyingsilverfin, line 1206: we remove 1 `mergeMap` call that could incur some overheads in terms of creating extra priority queues under the hood

dmitrii-ubskii (approved):
No notes, lgtm.

PR #6623, opened 2022-08-09
Fix storage iterator forwarding

flyingsilverfin:

We fix the behaviour of our storage iterators when forwarding them to a target. Storage iterators are bound to a particular prefix internally. However forwarding them to a value before the prefix should still be a valid operation that returns the entire prefix range as before, rather than returning nothing.

* Previously, forwarding a storage iterator to a value before the prefix would be allowed to proceed to performing a `seek()` operation on RocksDB
* However, this mean the next value retrieved from RocksDB wouldn't have the prefix that our storage iterator was bound to, and the storage iterator would immediately return with no answers
* Now, we do a no-op if the target is before the prefix that the iterator is bound to.

### DIFF
--- a/database/RocksIterator.java
+++ b/database/RocksIterator.java
@@ -215,7 +215,7 @@ public abstract class RocksIterator<T extends Key, ORDER extends Order>
 
         @Override
         public synchronized void forward(KeyValue<T, ByteArray> target) {
-            if (state == State.COMPLETED || ASC.isValidNext(target.key().bytes(), prefix.bytes())) return;
+            if (state == State.COMPLETED || !ASC.isValidNext(prefix.bytes(), target.key().bytes())) return;
             if (state == State.INIT) initialiseInternalIterator();
             internalRocksIterator.seek(target.key().bytes().getBytes());
             state = State.FORWARDED;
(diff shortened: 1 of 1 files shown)


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

database/RocksIterator.java
```diff
@@ -215,8 +215,8 @@ synchronized boolean fetchAndCheck() {
 
         @Override
         public synchronized void forward(KeyValue<T, ByteArray> target) {
-            if (state == State.COMPLETED) return;
-            else if (state == State.INIT) initialiseInternalIterator();
+            if (state == State.COMPLETED || ASC.isValidNext(target.key().bytes(), prefix.bytes())) return;
```
dmitrii-ubskii, line 218: Why is this of the form `isValidNext(to, from)` as opposed to `!isvalidNext(from, to)`?

flyingsilverfin: Checking if it is not valid you'd normally expect there to be an exception. So, mostly because I wanted to leave it "positive" to convey intent that this is OK behaviour.

dmitrii-ubskii: I see. What this tells me is that we need an `isValidPrev()` method. This version reads like "if it's valid (to do the opposite), then do nothing" which is more than a little confusing.
Alternatively, I'd argue that "if it isn't valid to move forward, don't move at all" reads quite a bit better, and it is a kind of error handling, as we can't legally forward to a value that comes before our current position.

flyingsilverfin: i see your point, I'll swap it around with the `!`

dmitrii-ubskii (approved):
LGTM.

PR #6628, opened 2022-08-11
Unifying a concludable with a conclusion considers value constraints

### LINKED ISSUE #6538: Incorrect validation for negated attribute values in rules (invalidates valid rules)

krishnangovindraj:
The current rule-validation incorrectly flags valid & useful scenarios as a negative cycle, such as in #6538 
These occur when the conclusion is a `has` with a constant attribute value, and the concludable has constraints with constant values. This PR incorporates evaluating the constraints in the concludable against the concluded value in such cases.  

* Extends Concludable.Has.unificationSatisfiable to evaluate constant value constraints against constant-value conclusions 
* Implements  the constraint consistency check in pattern::ValueConstraint.
- Currently only supports the case where one constraint is the an exact equality. The other can be any constraint. (Thus, the implementation checks whether a value satisfies a constraint, rather than checking whether two constraints are compatible)
* Extends unification requirements to include other value constraints in addition to equality.

This would satisfy/ fix #6538


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

logic/resolvable/Concludable.java
```diff
@@ -188,8 +188,13 @@ boolean unificationSatisfiable(ThingVariable concludableThingVar, ThingVariable
...
         if (!concludableThingVar.value().isEmpty() && !conclusionThingVar.value().isEmpty()) {
-            // TODO: detect value contradictions between constant predicates
-            satisfiable &= true;
+            assert conclusionThingVar.value().size() == 1;
+            ValueConstraint<?> conclusionValueConstraint = conclusionThingVar.value().stream().findFirst().get();
+
+            assert conclusionValueConstraint.predicate().isEquality() &&
+                    conclusionValueConstraint.predicate().asEquality().equals(EQ);
+
+            satisfiable &=  concludableThingVar.value().stream().allMatch( v -> v.isSatisfiedByValueOf(conclusionValueConstraint) );
```
krishnangovindraj, line 197: allMatch may be misleading, as different constraints on the same variable may be spread across concludables.

logic/resolvable/ConcludableTest.java
```diff
@@ -310,4 +311,338 @@ public void test_conjunction_anonymous_variable_constraints_are_not_conflated()
             assertTrue(hasConcludable.has().attribute().isa().isPresent());
         });
     }
+
+    @Test
+    public void test_var_attribute_unification_satisfiable(){
+        Pair<String,String>[] concludableConclusionPairs = new Pair[]{
```
krishnangovindraj, line 317: We test the various cases for comparison based predicate (assignment, =, !=, <, >, <=, >= ) against a constant value "explicit-has"(?) conclusion.

Does the concludable pair.first() have a satisfiable 'unification' with the conclusion pair.second()?

logic/resolvable/ConcludableTest.java
```diff
@@ -310,4 +311,338 @@ public void test_conjunction_anonymous_variable_constraints_are_not_conflated()
...
+            Concludable conclusion = Concludable.create(parseConjunction(pair.second())).stream().findFirst().get();
+
+            assertTrue(concludable.isHas());
+            assertTrue(conclusion.isHas());
+            assertTrue(concludable.unificationSatisfiable(concludable.asHas().attribute(), conclusion.asHas().attribute()));
+        }
+    }
+
+    @Test
+    public void test_valued_attribute_unification_unsatisfiable_substring(){
```
krishnangovindraj, line 632: Test the substring tests (like, contains).

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -195,6 +197,30 @@ public FunctionalIterator<AlphaEquivalence> alphaEquals(ValueConstraint<?> that)
                 .flatMap(a -> a.alphaEqualIf(this.value.equals(that.value)));
     }
 
+    public abstract boolean isSatisfiedByValueOf(ValueConstraint<?> conclusionValueConstraint);
+
+    protected boolean evaluateEqualityPredicate(TypeQLToken.Predicate.Equality predicate, int comparatorOutput) {
```
krishnangovindraj, line 202: Comparison based 'equality' predicates use this method to evaluate the predicate against the comparator output of the concludable and conclusion value.

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -195,6 +197,30 @@ public FunctionalIterator<AlphaEquivalence> alphaEquals(ValueConstraint<?> that)
...
+            case NEQ: return comparatorOutput != 0;
+            case GT: return comparatorOutput > 0;
+            case GTE: return comparatorOutput >= 0;
+            case LT: return comparatorOutput < 0;
+            case LTE: return comparatorOutput <= 0;
+            default: throw TypeDBException.of(ILLEGAL_STATE);
+        }
+    }
+
+    protected boolean evaluateSubstringPredicate(TypeQLToken.Predicate.SubString predicate,
```
krishnangovindraj, line 214: Substring predicates use this method to evaluate the concludable pattern against the conclusion value.

logic/resolvable/ConcludableTest.java
```diff
@@ -310,4 +311,340 @@ public void test_conjunction_anonymous_variable_constraints_are_not_conflated()
...
+            // String comparisons
+            new Pair("{ $x \"one\"; }", "{$y \"one\";}"),
+            new Pair("{ $x = \"one\"; }", "{$y \"one\";}"),
+            new Pair("{ $x >= \"one\"; }", "{$y \"one\";}"),
+            new Pair("{ $x <= \"one\"; }", "{$y \"one\";}"),
+            new Pair("{ $x != \"one\"; }", "{$y \"two\";}"),
+            new Pair("{ $x > \"one\"; }", "{$y \"two\";}"),
+            new Pair("{ $x < \"two\"; }", "{$y \"one\";}"),
+
+            // String comparisons, case-sensitivity
```
krishnangovindraj, line 478: Case sensitive comparisons.

traversal/predicate/Predicate.java
```diff
@@ -39,13 +41,59 @@ private Predicate(PRED_OP operator, PRED_ARG argument) {
         this.hash = Objects.hash(operator, argument);
     }
 
+    // TODO: Move this where it belongs - Encoding?
+    public static int compareBooleans(boolean first, boolean second) {
```
krishnangovindraj, line 45: Comparison methods for values of compatible data-types. Considered the source-of-truth

traversal/predicate/Predicate.java
```diff
@@ -39,13 +41,59 @@ private Predicate(PRED_OP operator, PRED_ARG argument) {
...
+    public static int compareDateTimes(LocalDateTime first, LocalDateTime second) {
+        return first.compareTo(second);
+    }
+
+    public static int compareStrings(String first, String second) {
+        return first.compareTo(second);
+    }
+
+
+    public static boolean containsIgnoreCase(String superString, String subString) {
```
krishnangovindraj, line 77: Substring predicates

traversal/predicate/PredicateArgument.java
```diff
@@ -98,9 +98,12 @@ public boolean apply(PredicateOperator.Equality operator, AttributeVertex<?> ver
...
 
-                if (vertex.isLong()) return operator.apply(vertex.asLong().value().compareTo(value));
-                else if (vertex.isDouble())
-                    return operator.apply(Predicate.compareDoubles(vertex.asDouble().value(), value));
+                if (vertex.isLong()) {
+                    return operator.apply(Predicate.compareLongs(vertex.asLong().value(), value));
+                }
+                else if (vertex.isDouble()){
+                    return operator.apply(Predicate.compareDoubleToLong(vertex.asDouble().value(), value));
+                }
```
krishnangovindraj, line 106: Standardize comparison

traversal/predicate/PredicateOperator.java
```diff
@@ -203,45 +209,35 @@ SubString asSubString() {
...
-            boolean apply(String vertexValue, Traversal.Parameters.Value predicateValue) {
+            public boolean apply(String vertexValue, Traversal.Parameters.Value predicateValue) {
                 assert predicateValue.isRegex();
-                return predicateValue.getRegex().matcher(vertexValue).matches();
+                return apply(vertexValue, predicateValue.getRegex());
+            }
+
+            @Override
+            public boolean apply(String vertexValue, Pattern predicateValue) {
+                return Predicate.stringLike(predicateValue, vertexValue);
```
krishnangovindraj, line 234: Use newly introduced standard definitions.

traversal/predicate/PredicateOperator.java
```diff
@@ -85,9 +86,13 @@ public Equality(TypeQLToken.Predicate.Equality token) {
...
 
-        abstract boolean apply(int comparisonResult);
+        public static Equality of(TypeQLToken.Predicate.Equality operator) {
+            return Equality.operators.get(operator);
+        }
 
-        abstract Equality reflection();
+        public abstract boolean apply(int comparisonResult);
+
+        abstract public Equality reflection();
```
krishnangovindraj, line 95: Reflection doesn't need to be public, but apply does.
Should I revert this to package-private?

krishnangovindraj, line 95: Reflection doesn't need to be public, but apply does.
Should I revert this to package-private?

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -221,6 +234,15 @@ public Double asDouble() {
...
 
+        @Override
+        public boolean isCompatibleWithEQ(ValueConstraint<?> conclusionValueConstraint) {
+            return conclusionValueConstraint.isVariable() ||
+                    (conclusionValueConstraint.isLong() &&
+                            PredicateOperator.Equality.of(predicate()).apply(Predicate.compareLongs(conclusionValueConstraint.asLong().value(), value))) ||
+                    (conclusionValueConstraint.isDouble() &&
+                            PredicateOperator.Equality.of(predicate()).apply(Predicate.compareDoubleToLong(conclusionValueConstraint.asDouble().value(), value)));
+        }
+
```
krishnangovindraj, line 245: Defer type-casting responsibilities to the comparison definition.

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -326,6 +363,22 @@ public java.lang.String toString() {
...
+        @Override
+        public boolean isCompatibleWithEQ(ValueConstraint<?> conclusionValueConstraint) {
+            if (conclusionValueConstraint.isVariable()) return true;
+            if (!conclusionValueConstraint.isString()) return false;
+            if (predicate.isEquality()) {
+                return PredicateOperator.Equality.of(predicate.asEquality()).apply(Predicate.compareStrings(conclusionValueConstraint.asString().value(), value));
+            } else if (predicate.isSubString()) {
+                PredicateOperator.SubString operator = PredicateOperator.SubString.of(predicate.asSubString());
+                if (operator == PredicateOperator.SubString.CONTAINS) {
+                    return PredicateOperator.SubString.CONTAINS.apply(conclusionValueConstraint.asString().value(), value);
```
krishnangovindraj, line 376: LIKE and CONTAINS actually have two different predicate value types - Pattern and String.
(This) ValueConstraint class has only the String. 

Introducing a type-parameter on PredicateOperator.SubString is a workaround.

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -326,6 +363,22 @@ public java.lang.String toString() {
...
+                } else if (operator == PredicateOperator.SubString.LIKE) {
+                    return PredicateOperator.SubString.LIKE.apply(conclusionValueConstraint.asString().value(), Pattern.compile(value));
```
krishnangovindraj, line 378: Doing Pattern.compile here may be non-standard, but the standard way will likely require more cascading changes.

traversal/predicate/PredicateOperator.java
```diff
@@ -203,45 +209,35 @@ SubString asSubString() {
...
             }
         };
 
         private static final Map<TypeQLToken.Predicate.SubString, SubString> operators = map(
                 pair(TypeQLToken.Predicate.SubString.CONTAINS, SubString.CONTAINS),
                 pair(TypeQLToken.Predicate.SubString.LIKE, SubString.LIKE)
         );
-
-        private static SubString of(TypeQLToken.Predicate.SubString token) {
-            return operators.get(token);
```
krishnangovindraj, line 244: Moved to top & Made public.

traversal/predicate/PredicateArgument.java
```diff
@@ -154,7 +159,7 @@ public boolean apply(PredicateOperator operator, AttributeVertex<?> vertex, Trav
             public boolean apply(PredicateOperator operator, AttributeVertex<?> vertex, String value) {
                 if (!vertex.valueType().comparableTo(Encoding.ValueType.STRING)) return false;
                 assert vertex.isString() && operator.isEquality();
-                return operator.asEquality().apply(vertex.asString().value().compareTo(value));
+                return operator.asEquality().apply(Predicate.compareStrings(vertex.asString().value(), value));
             }
```
krishnangovindraj, line 163: (Expand a few lines upwards to see how SubString constraints were handled)

krishnangovindraj (commented):
add comments

logic/resolvable/Concludable.java
```diff
@@ -222,9 +227,9 @@ private static Function<com.vaticle.typedb.core.concept.thing.Attribute, Boolean
 
                 assert a.getType().isDouble() || a.getType().isLong();
                 if (a.getType().isLong())
-                    return value.asLong().value().compareTo(a.asLong().getValue()) == 0;
+                    return Predicate.compareLongs(a.asLong().getValue(), value.asLong().value()) == 0;
                 else if (a.getType().isDouble())
-                    return Predicate.compareDoubles(a.asDouble().getValue(), value.asLong().value()) == 0;
+                    return Predicate.compareDoubleToLong(a.asDouble().getValue(), value.asLong().value()) == 0;
                 else throw TypeDBException.of(ILLEGAL_STATE);
```
krishnangovindraj, line 233: Update these to use the Predicate.* comparisons as well

krishnangovindraj (commented):
more comments

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -195,6 +198,16 @@ public FunctionalIterator<AlphaEquivalence> alphaEquals(ValueConstraint<?> that)
                 .flatMap(a -> a.alphaEqualIf(this.value.equals(that.value)));
     }
 
+    public boolean maybeCompatible(ValueConstraint<?> valueConstraint) {
```
flyingsilverfin, line 201: I do wonder, when we say two predicates are "compatible", is that less clear than inverting this to say "contradicts"?

traversal/predicate/PredicateOperator.java
flyingsilverfin: Pretty sure IntelliJ Signature refactor can do that for you in one go

logic/resolvable/Concludable.java
```diff
@@ -188,8 +188,13 @@ boolean unificationSatisfiable(ThingVariable concludableThingVar, ThingVariable
         }
 
         if (!concludableThingVar.value().isEmpty() && !conclusionThingVar.value().isEmpty()) {
-            // TODO: detect value contradictions between constant predicates
-            satisfiable &= true;
```
krishnangovindraj, line 192: (Entry point)
For a concludable to unify with a conclusion, the constraints on the conclusion are compatible with the constraints on the concludable.
This is currently only useful in the case where the conclusion is an explicit-has, where the concluded value can be known statically. In all other cases, the concluded value is variables and can thus satisfy any single constraint in isolation.

(We cannot consider two constraints at a time since concludables consist of only one constraint. Trying to support this is likely out of scope of this optimisation and what the reasoner does)

flyingsilverfin (changes requested):
I'm happy with the overall architectures and high-level changes you've made! Mostly requesting changes on naming consistency and code conventions :)

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -195,6 +198,16 @@ public FunctionalIterator<AlphaEquivalence> alphaEquals(ValueConstraint<?> that)
...
+        if (valueConstraint.predicate == EQ) {
+            return isCompatibleWithEQ(valueConstraint);
+        } else { // TODO: implement inequality compatibility when useful
+            return true;
```
krishnangovindraj, line 205: Implementing this may not be hard for the (in)Equality constraints , but may be hopelessly complicated and uninformative in the case of SubString constraints. 
There is also no use case for it at the moment, so we defer implementing it till we need it.

pattern/constraint/thing/ValueConstraint.java
krishnangovindraj: "IsConsistentWith" ?

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -195,6 +198,16 @@ public FunctionalIterator<AlphaEquivalence> alphaEquals(ValueConstraint<?> that)
...
+        }
+    }
+
+    protected abstract boolean isCompatibleWithEQ(ValueConstraint<?> conclusionValueConstraint);
```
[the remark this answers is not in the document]
krishnangovindraj: We've done what?

logic/resolvable/Concludable.java
```diff
@@ -222,9 +227,9 @@ private static Function<com.vaticle.typedb.core.concept.thing.Attribute, Boolean
 
                 assert a.getType().isDouble() || a.getType().isLong();
                 if (a.getType().isLong())
-                    return value.asLong().value().compareTo(a.asLong().getValue()) == 0;
+                    return Predicate.compareLongs(a.asLong().getValue(), value.asLong().value()) == 0;
```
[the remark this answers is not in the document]
krishnangovindraj: Is the domain really clear from it?

logic/resolvable/Concludable.java
```diff
@@ -188,8 +188,13 @@ boolean unificationSatisfiable(ThingVariable concludableThingVar, ThingVariable
         }
 
         if (!concludableThingVar.value().isEmpty() && !conclusionThingVar.value().isEmpty()) {
-            // TODO: detect value contradictions between constant predicates
-            satisfiable &= true;
+            assert conclusionThingVar.value().size() == 1;
+            ValueConstraint<?> conclusionValueConstraint = conclusionThingVar.value().stream().findFirst().get();
```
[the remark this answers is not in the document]
krishnangovindraj: Replaced.

logic/resolvable/ConcludableTest.java
```diff
@@ -310,4 +311,340 @@ public void test_conjunction_anonymous_variable_constraints_are_not_conflated()
             assertTrue(hasConcludable.has().attribute().isa().isPresent());
         });
     }
+
+    @Test
+    public void test_var_attribute_unification_satisfiable(){
```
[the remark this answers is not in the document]
krishnangovindraj: Updated. Is that better?

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -221,6 +234,15 @@ public Double asDouble() {
             return new Double(owner, predicate.asEquality(), value);
         }
 
+        @Override
+        public boolean isCompatibleWithEQ(ValueConstraint<?> conclusionValueConstraint) {
+            return conclusionValueConstraint.isVariable() ||
+                    (conclusionValueConstraint.isLong() &&
+                            PredicateOperator.Equality.of(predicate()).apply(Predicate.compareLongs(conclusionValueConstraint.asLong().value(), value))) ||
```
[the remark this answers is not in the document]
krishnangovindraj: Yeah, but there's also TypeQLToken.Predicate.Equality lying about which wouldn't be out of place if it were used here.

logic/resolvable/Concludable.java
```diff
@@ -188,8 +188,13 @@ boolean unificationSatisfiable(ThingVariable concludableThingVar, ThingVariable
...
+
+            assert conclusionValueConstraint.predicate().isEquality() &&
+                    conclusionValueConstraint.predicate().asEquality().equals(EQ);
+
+            satisfiable &=  concludableThingVar.value().stream().allMatch( v -> v.maybeCompatible(conclusionValueConstraint) );
```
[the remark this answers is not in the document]
krishnangovindraj: Yes. A lot of them can be static.

A lot of files have spacing which gets fixed by the auto-formatter and would lead to random lines being updated.
I do try to run it on my changes.

logic/resolvable/Concludable.java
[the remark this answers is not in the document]
flyingsilverfin: I think once you get to know Java a bit better you'll start to recognise when it's a java built-in versus one of ours - the style is slightly different and they probably would not have a `compareLongs` method, but instead an overriden `compare` method that is abstract :)

pattern/constraint/thing/ValueConstraint.java
flyingsilverfin: well, to keep the intent we'd have to do `maybeConsistentWith` since we want to keep the `maybe`-ness about the positive check.

flyingsilverfin: if we call it `contradicts` we don't need the `maybe` because if it's not a certain contradiction it's not a contradiction. We can then also shorten the abstract method it calls to `contradictsEquality`

pattern/constraint/thing/ValueConstraint.java
flyingsilverfin: useful comment, good!

pattern/constraint/thing/ValueConstraint.java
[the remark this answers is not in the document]
flyingsilverfin: ahh okay!

krishnangovindraj:
This whole thing should be merged into Vaticle:develop, shouldn't it?

logic/resolvable/Concludable.java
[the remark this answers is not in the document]
krishnangovindraj: Probably not Java, but you'd expect it to at least be local to this package?
If this were in the common package, like Iterators.iterate, I'd understand static imports. But Predicate doesn't feel like the standard place to have it yet, so I thought it made sense to keep it explicit.

logic/resolvable/Concludable.java
```diff
@@ -147,68 +147,12 @@ public Attribute asAttribute() {
...
-        // TODO: this is cachable, and is a hot code path - analyse and see impact of cache
-        if (label.scope().isPresent()) {
-            assert conceptMgr.getRelationType(label.scope().get()) != null;
-            return conceptMgr.getRelationType(label.scope().get()).getRelates(label.name()).getSubtypes()
-                    .map(RoleType::getLabel);
-        } else {
-            return conceptMgr.getThingType(label.name()).getSubtypes().map(Type::getLabel);
-        }
-    }
-
```
krishnangovindraj, line 211: Moved to Unifier Builder

logic/resolvable/Unifier.java
```diff
@@ -255,6 +262,77 @@ public Builder clone() {
...
+        static boolean unificationSatisfiable(RelationConstraint.RolePlayer concludableRolePlayer, RelationConstraint.RolePlayer conclusionRolePlayer, ConceptManager conceptMgr) {
+            assert conclusionRolePlayer.roleType().isPresent();
+            boolean satisfiable = true;
+            if (concludableRolePlayer.roleType().isPresent()) {
+                satisfiable = unificationSatisfiable(concludableRolePlayer.roleType().get(), conclusionRolePlayer.roleType().get(), conceptMgr);
+            }
+            satisfiable &= unificationSatisfiable(concludableRolePlayer.player(), conclusionRolePlayer.player());
+            return satisfiable;
+        }
     }
```
krishnangovindraj, line 336: Moved from Concludable

krishnangovindraj (commented):
Move unification methods from Concludable to unifier
Fix stuff.

logic/resolvable/ConcludableTest.java
```diff
@@ -310,4 +312,378 @@ public void test_conjunction_anonymous_variable_constraints_are_not_conflated()
...
+        };
+
+        for (Pair<String, String> pair : constraintValuePairs) {
+            Concludable constraint = iterate(Concludable.create(parseConjunction(pair.first()))).next();
+            Concludable value = iterate(Concludable.create(parseConjunction(pair.second()))).next();
+
+            assertTrue(constraint.isAttribute());
+            assertTrue(value.isAttribute());
+            assertFalse(Unifier.Builder.unificationSatisfiable(constraint.asAttribute().attribute(), value.asAttribute().attribute()));
+        }
```
flyingsilverfin, line 652: Let's move this into `UnifierTest` as a unit test, and then not create Concludables - get the variables used for satisfiability tests from the conjunction directly

logic/resolvable/Concludable.java
```diff
@@ -281,7 +225,7 @@ private static FunctionalIterator<AlphaEquivalence> alphaEqualValueConstraints(S
         }
     }
 
-    void addConstantValueRequirements(Unifier.Builder unifierBuilder, Set<ValueConstraint<?>> values,
+    static void addConstantValueRequirements(Unifier.Builder unifierBuilder, Set<ValueConstraint<?>> values,
```
flyingsilverfin, line 228: see if this moves to `Builder` as well, but it may not be easy

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -400,6 +459,11 @@ public FunctionalIterator<AlphaEquivalence> alphaEquals(ValueConstraint<?> that)
         public Variable clone(Conjunction.ConstraintCloner cloner) {
             return cloner.cloneVariable(owner).valueVariable(predicate(), cloner.cloneVariable(value));
         }
+
+        @Override
+        public boolean isConsistentWithEquality(ValueConstraint<?> conclusionValueConstraint) {
```
flyingsilverfin, line 464: these should not be public

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -195,6 +198,16 @@ public FunctionalIterator<AlphaEquivalence> alphaEquals(ValueConstraint<?> that)
...
 
+    public boolean inconsistentWith(ValueConstraint<?> valueConstraint) {
+        if (valueConstraint.predicate == EQ) {
+            return !isConsistentWithEquality(valueConstraint);
+        } else { // TODO: implement inequality compatibility when useful
+            return true;
+        }
+    }
+
+    protected abstract boolean isConsistentWithEquality(ValueConstraint<?> conclusionValueConstraint);
```
flyingsilverfin, line 209: this should be without a modifier, ie. `package-private`, along with all the implementations below

flyingsilverfin (changes requested):

logic/resolvable/ConcludableTest.java
```diff
@@ -27,6 +28,7 @@
 import java.util.Set;
 
 import static com.vaticle.typedb.common.collection.Collections.list;
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
```
flyingsilverfin, line 31: unused deps?

logic/resolvable/Concludable.java
```diff
@@ -281,11 +174,11 @@ private static FunctionalIterator<AlphaEquivalence> alphaEqualValueConstraints(S
...
```
flyingsilverfin, line 177: you have this in here staticall and in the builder?

flyingsilverfin (changes requested):

logic/resolvable/ConcludableTest.java
krishnangovindraj: fixed

logic/resolvable/Unifier.java
```diff
@@ -333,6 +346,81 @@ static boolean unificationSatisfiable(RelationConstraint.RolePlayer concludableR
...
+                            return PredicateOperator.SubString.LIKE.apply(a.asString().getValue(), Pattern.compile(value.asString().value()));
+                        } else throw TypeDBException.of(ILLEGAL_STATE);
+                    };
+                }  else throw TypeDBException.of(ILLEGAL_STATE);
+            }  else {
+                throw TypeDBException.of(ILLEGAL_STATE);
+            }
+
+            return predicateFn;
+        }
```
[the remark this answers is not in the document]
krishnangovindraj: Actually, it's better not to include the other constraints in the unifier requirements yet.
We'll have two ways of doing things.
I'll make another branch for it and filter the constraint types here (todo: add a link to the comment where I did it)

logic/resolvable/Unifier.java
```diff
@@ -236,6 +248,14 @@ public void addLabelType(Identifier.Variable.Label source, Set<Label> allowedTyp
             unifiedRequirements.types(target, allowedTypes);
         }
 
+        void addConstantValueRequirements(Set<ValueConstraint<?>> values,
+                                          Retrievable id, Retrievable unifiedId) {
+            iterate(values).filter(v -> v.predicate().equals(EQ)).forEachRemaining(value -> {
+                unifiedRequirements().predicates(unifiedId, valuePredicate(value));
```
krishnangovindraj, line 254: Filter out the non-EQ constraints here. 
This was the original behaviour (I believe). Removing the filter would allow other constraints to be unification requirements. Should we just do that in this PR?

logic/resolvable/Unifier.java
```diff
@@ -255,6 +275,152 @@ public Builder clone() {
...
+            assert conclusionRolePlayer.roleType().isPresent();
+            boolean satisfiable = true;
+            if (concludableRolePlayer.roleType().isPresent()) {
+                satisfiable = unificationSatisfiable(concludableRolePlayer.roleType().get(), conclusionRolePlayer.roleType().get(), conceptMgr);
+            }
+            satisfiable &= unificationSatisfiable(concludableRolePlayer.player(), conclusionRolePlayer.player());
+            return satisfiable;
+        }
+
+        static Function<com.vaticle.typedb.core.concept.thing.Attribute, Boolean> valuePredicate(ValueConstraint<?> value) {
```
flyingsilverfin, line 350: add assert the value constraint is not a variable on the RHS

logic/resolvable/Unifier.java
```diff
@@ -236,6 +248,14 @@ public void addLabelType(Identifier.Variable.Label source, Set<Label> allowedTyp
             unifiedRequirements.types(target, allowedTypes);
         }
 
+        void addConstantValueRequirements(Set<ValueConstraint<?>> values,
+                                          Retrievable id, Retrievable unifiedId) {
+            iterate(values).filter(v -> v.predicate().equals(EQ)).forEachRemaining(value -> {
```
flyingsilverfin, line 253: definitely need to filter out variables: variables on the RHS cannot be part of the unification process because they would not unify with a variable in the `conclusion`. We can handle constants on the RHS because they can be evaluated during the forward/reverse unification

logic/resolvable/Unifier.java
flyingsilverfin: and is equality

logic/resolvable/Unifier.java
flyingsilverfin: we should delete this `EQ` filter, since the code can already handle all predicates with a constant on the RHS.

This means we've extended concludables to handle all constant predicates (compared to just `=` in the past). As such, we should also make sure the `Retrievable` does not contain these predicates when they are included in a `Concludable`

logic/resolvable/Unifier.java
flyingsilverfin, line 350: Add `TODO: ` that this should leverage traversal `Predicate`

logic/resolvable/RetrievableTest.java
```diff
@@ -158,10 +158,10 @@ public void test_retrievables_split_by_labelled_type_are_separated() {
     }
 
     @Test
-    public void test_has_value_constraints_are_in_retrievable_and_concludable() {
-        Set<Concludable> concludables = Concludable.create(parse("{ $a has $b; $b > 5; $b < 10; }"));
+    public void test_has_value_constraints_are_in_retrievable_iff_they_are_not_in_concludable() {
+        Set<Concludable> concludables = Concludable.create(parse("{ $a has $b; $b > 5; }"));
         Set<Retrievable> retrievables = Retrievable.extractFrom(parse("{ $a has $b; $b > 5; $b < 10; }"), concludables);
-        assertEquals(set(parse("{ $b > 5; $b < 10; }")),
+        assertEquals(set(parse("{ $b < 10; }")),
```
flyingsilverfin, line 164: Since we can now handle constraint compatibility, AND evaluate any constant predicates in the Unifier, we can move all the constant value constraints into the single concludable exclusively.

This should be more efficient because we create potentially fewer unifiers and also potentially eliminate some Retrievables totally.

Verify that `$a has $b; $c has $b; $b > 10;` ends up with two concludables and 0 retrievables: `$a has $b; $b > 10;` and `$c has $b; $b > 10;`

flyingsilverfin (approved):
WOOOP

logic/resolvable/Unifier.java
```diff
@@ -255,6 +274,158 @@ public Builder clone() {
...
+
+        /*
+        Unifying a source type variable with a target type variable is impossible if none of the source's allowed labels,
+        and their subtypes' labels, are found in the allowed labels of the target type.
+
+        Improvements:
+        * we could take information such as negated constraints into account.
+         */
+        static boolean unificationSatisfiable(TypeVariable concludableTypeVar, TypeVariable conclusionTypeVar, ConceptManager conceptMgr) {
+
```
flyingsilverfin, line 301: stray newline

test/integration/logic/resolvable/UnifyHasConcludableTest.java
```diff
@@ -488,4 +494,284 @@ public void has_all_equivalent_vars_unifier() {
...
+            put("$y", set("$x"));
+            put("$_0", set("$_0"));
+        }};
+        assertEquals(expected, result);
+
+        // test requirements
+        assertEquals(0, unifier.requirements().types().size());
+        assertEquals(2, unifier.requirements().isaExplicit().size());
+        assertEquals(set(Label.of("first-name"), Label.of("last-name")), unifier.requirements().isaExplicit().get(Identifier.Variable.anon(0)));
+        assertEquals(1, unifier.requirements().predicates().size()); // NEW CASE
```
flyingsilverfin, line 520: stray comment

Issue #6379, opened 2021-06-22
Undefining Type throws NullPointer if declared parent type doesn't exist


jmsfltchr:
## Description

Undefining a type, declaring that it subtypes a type that doesn't exist, throws a NullPointerException.

## Environment

1. OS (where TypeDB server runs): Mac OS 10
2. TypeDB version (and platform): TypeDB 2.1.2
3. TypeDB client: Found in integration test

## Reproducible Steps

Steps to create the smallest reproducible scenario:
Try this in an integration test, having defined any valid rule called `my-rule`:
```
tx.query().undefine(TypeQL.undefine(TypeQL.type("my-rule").sub("rule")));
```
Above is the Grakn 1.x syntax for undefining a rule, so is an easy mistake to make in TypeDB 2.x, which is how I came across this bug.

The correct way would be:
```
tx.query().undefine(TypeQL.undefine(Collections.singletonList(TypeQL.rule("my-rule"))));
```

## Expected Output

We expect that 

## Actual Output

```
java.lang.NullPointerException: Cannot invoke "com.vaticle.typedb.core.concept.type.ThingType.isRoleType()" because "thingType" is null
com.vaticle.typedb.core.common.exception.TypeDBException: 
	at com.vaticle.typedb.core.common.exception.TypeDBException.of(TypeDBException.java:51)
	at com.vaticle.typedb.core.rocks.RocksStorage.exception(RocksStorage.java:139)
	at com.vaticle.typedb.core.rocks.RocksStorage$TransactionBounded.exception(RocksStorage.java:274)
	at com.vaticle.typedb.core.rocks.RocksStorage$Data.exception(RocksStorage.java:325)
	at com.vaticle.typedb.core.graph.GraphManager.exception(GraphManager.java:55)
	at com.vaticle.typedb.core.concept.ConceptManager.exception(ConceptManager.java:228)
	at com.vaticle.typedb.core.query.QueryManager.undefine(QueryManager.java:184)
	at com.vaticle.typedb.core.query.QueryManager.undefine(QueryManager.java:175)
	at com.vaticle.typedb.core.test.behaviour.resolution.framework.test.TestResolution.testResolutionThrowsForTransitivityWhenRuleTriggersTooOften(TestResolution.java:107)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #6631, opened 2022-08-12
Undefiner checks type exists

flyingsilverfin:

Prevent a null pointer exception when undefining a type that doesn't exist. This is a variant of #6379, which is not an issue anymore.

* Validate that a type being undefined exists
* Remove redundant code null checks when retrieving a concept

### DIFF
--- a/query/Undefiner.java
+++ b/query/Undefiner.java
@@ -120,6 +120,7 @@ public class Undefiner {
             else if (undefined.contains(variable)) return; // do nothing
 
             ThingType type = getThingType(labelConstraint);
+            if (type == null) throw TypeDBException.of(TYPE_NOT_FOUND, labelConstraint.scopedLabel());
 
             if (!variable.plays().isEmpty()) undefinePlays(type, variable.plays());
             if (!variable.owns().isEmpty()) undefineOwns(type, variable.owns());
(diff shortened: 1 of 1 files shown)


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

query/Undefiner.java
```diff
@@ -141,9 +142,7 @@ else if (variable.valueType().isPresent()) {
 
     private ThingType getThingType(LabelConstraint label) {
         try (ThreadTrace ignored = traceOnThread(TRACE_PREFIX + "gettype")) {
-            ThingType thingType;
-            if ((thingType = conceptMgr.getThingType(label.label())) != null) return thingType;
-            else return null;
```
flyingsilverfin, line 146: the concept manager already returns a null when the label doesn't exist, so this check is redundant

query/Undefiner.java
```diff
@@ -163,7 +162,7 @@ private void undefineSub(ThingType thingType, SubConstraint subConstraint) {
             }
             ThingType supertype = getThingType(subConstraint.type().label().get());
             if (supertype == null) {
-                throw TypeDBException.of(TYPE_NOT_FOUND, subConstraint.type().label().get());
```
flyingsilverfin, line 166: printing out the `LabelConstraint` leads to an odd printing behaviour, we want to just print `TYPE` does not exist, rather than `TYPE label TYPE` which represents `[variable] label TYPE`

haikalpribadi (approved):

PR #6634, opened 2022-08-18
Remove redundant commit rule validation

flyingsilverfin:

Eliminate an expensive rule re-validation on schema transaction commit when the schema types have not changed. Rules are therefore only validated at define time, and re-validated at commit time if the types have changed.

If using rules that are costly to validate, the optimal way to define a schema is now to use one transaction to create all the Types, and a separate transaction to create all the Rules. 

We also fix rule re-indexing at commit time.

* Only re-validate rules at commit time if any schema types have been modified during the transaction


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

logic/LogicManager.java
```diff
@@ -104,21 +104,20 @@ private FunctionalIterator<Rule> rulesWithNegations() {
...
+        if (graphMgr.schema().typesModified()) {
+            // re-validate all rules are valid
+            rules().forEachRemaining(rule -> rule.validate(this, conceptMgr));
 
-        // re-index if rules are valid and satisfiable
-        if (graphMgr.schema().rules().conclusions().isOutdated()) {
-            graphMgr.schema().rules().all().forEachRemaining(s -> fromStructure(s).reindex());
-            graphMgr.schema().rules().conclusions().outdated(false);
+            // recreate rule index conclusions
+            graphMgr.schema().rules().all().forEachRemaining(s -> fromStructure(s).conclusion().reindex());
```
flyingsilverfin, line 120: we only need to validate & reindex rules if the schema types are modified. The other validation (eg. negation cycle) detection must still be done

logic/Rule.java
```diff
@@ -210,11 +210,6 @@ private Conjunction thenPattern(com.vaticle.typeql.lang.pattern.variable.ThingVa
         return conj;
     }
 
-    public void reindex() {
-        conclusion().unindex();
-        conclusion().index();
-    }
```
flyingsilverfin, line 216: move from `Rule.reindex()` to `Rule.Conclusion.reindex()`

graph/TypeGraph.java
```diff
@@ -694,33 +697,21 @@ public class Conclusions {
 
             private final Buffered buffered;
             private final Persisted persisted;
-            private boolean outdated;
```
flyingsilverfin, line 697: we don't need to store `outdated` on the `Conclusions` index, since we will scan the schema graph for modified types at commit time instead.

krishnangovindraj (approved):

PR #6640, opened 2022-09-06
Answer count model for reasoner planner

krishnangovindraj:

We introdue statistics-based answer count estimation which will be the basis of the cost of reasoner plans.

For a given conjunction, We aim to estimate:
* The number of answers for a given (sub)set of variables
* when considering only a subset of resolvables of the conjunction 
We believe this is sufficient for our formulation of the reasoner planner's cost function. 

* **reasoner.planner.AnswerCountEstimator.ConjunctionModel:**
  * Given a conjunction, builds a model and answers queries to estimate answers for 
    * The number of answers to a subset of variables
    * Considering only a subset of the specified resolvables in the conjunction. 
* **reasoner.planner.AnswerCountEstimator.LocalModel:**
  * The units of estimation. The ConjunctionModel derives its estimates from the LocalModels 
* **reasoner.planner.AnswerCountEstimator:**
  * Registry of ConjunctionModels
  * Coordinates between them while inter-dependent models are being built
* **reasoner.planner.AnswerCountEstimatorTest:**
  * TestCases for concludables & conjunctions verifying the estimates are as expected.

### The conjunction model
Rules are inherently recursive, and a planner will ask for answerEstimates for various variable and resolvable subsets multiple times - To prevent expensive computation, we build conjunction models which can answer these queries without recursing into triggered rules. 

We describe the model by starting at a naive estimate and incrementally refining the model. We use as running example:
```
$p isa person; $c isa person; $d isa dog;
$_pc (parent: $p, child: $c) isa parentship; $_pd (owner: $c, pet: $d) isa pet-ownership ;
```
We'll query the variables {$p,$c,$d} (and use $_pc, and $_pd for easy referencing )   

#### Variable type-information only
For a given conjunction, the number of answers to the variables is the number of values each variable can take-up. Take the product across variables, of the number of things in the type-graph matching the variable set of types.

`estimate({$p, $c, $d))  = thingsMatchingTypes({person}) * thingsMatchingTypes({person}) * thingsMatchingTypes({dog})`

#### Resolvables constrain variables
Notice the resolvables restrict the number of values a variable can take - $p is restricted to the number of role-players of parent in a parentship.

```
estimate({$p, $c, $d))  =
  min(thingsMatchingType({person}), rolePlayers(parentship:parent)) * 
  min(thingsMatchingType({person}, rolePlayers(parentship:child), rolePlayers(pet-ownership:owner)) *
  min(thingsMatchingType({dog}), rolePlayers(pet-ownership:pet) 
```

#### Factorising estimates over the resolvables:
It's become obvious above that the rolePlayer constraints (and hasEdge constraints, etc) are enforced only when a resolvable is in the evaluated set.
The conjunction model uses this fact by constructing a 'local model' for each resolvable (using numbers from stats and an approximate recursive answer count (see acyclic-models)). The answer count to a variable given a set of resolvables to evaluate is the minimum, across all local models, of the answers to the variable returned by each model.

#### Co-constrained variables
$p and $c  (and $c,$d) are co-constrained by being in the same relation. 
Roughly, moving our estimate from factorising over variables to factorising over resolvables looks like this:
```
estimate = min {
   answers($_pc, {$p,$c}) * answers($_pd, {$d}),
   answers($_pc, {$p}) * answers($_pd, {$c, $d})
}
```
We compute this (approximately) by doing a greedy cover - Find the set of resolvables which cover all the queried variables but has the minimal answer count.   

### Building local models and handling cycles
Building local models for concludables becomes complicated when they trigger a cycle - since the answer-count recursively depends on itself.
Building the local model for a concludable which is not in a cyclic path is simpler - we can use the conjunction models of the conditions of the triggered rules to estimate the inferred count. 

We figured answer-counts involving cycles are hard, and we approximate it by breaking the cycle: 
When constructing the model for a cyclic concludable, we only consider the acyclic concludables in the triggered rules.

This leads us to a three-phase process for building a model for a conjunction

1. Register conjunction & its dependencies, identify cyclic concludables
2. Construct acyclic models for all of them
3. Construct cyclic models based on the acyclic model estimates


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

logic/LogicManager.java
```diff
@@ -69,7 +69,7 @@ public LogicManager(GraphManager graphMgr, ConceptManager conceptMgr, TraversalE
         this.compiledConjunctions = new HashMap<>();
     }
 
-    GraphManager graph() { return graphMgr; }
+    public GraphManager graph() { return graphMgr; }
```
krishnangovindraj, line 73: Need access for the stats.

reasoner/BUILD
```diff
@@ -46,6 +46,7 @@ native_java_libraries(
         "//common:common",
         "//concurrent:concurrent",
         "//concept:concept",
+        "//graph:graph",
         "//logic:logic",
```
krishnangovindraj, line 50: Need access for the stats.

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,384 @@
...
+
+            private CostCover(long cost) {
+                this.cost = cost;
+            }
+
+            private static long costToCover(Collection<Variable> subset, Map<Variable, CostCover> coverMap) {
+                Set<CostCover> subsetCoveredBy = coverMap.keySet().stream().filter(subset::contains)
+                        .map(coverMap::get).collect(Collectors.toSet());
+                return subsetCoveredBy.stream().map(cc -> cc.cost).reduce(1L, (x, y) -> x * y);
+            }
```
krishnangovindraj, line 381: Product of covering-costs for the interesting `subset` of variables.

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,384 @@
...
+                    super(variables);
+                    this.staticEstimate = estimate;
+                }
+
+                @Override
+                long answerEstimate(ConjunctionAnswerCountEstimator costEstimator, Set<Variable> variablesOfInterest) { return staticEstimate; }
+            }
+
+            public static class CoPlayerEstimate extends LocalEstimate {
+
```
krishnangovindraj, line 333: (Possibly overkill)
Variables which play a role in the same relation are strongly co-constrained by the number of relations.
e.g. : The number of edges in a graph v/s the square of number of vertices. 

Considers the fact that multiple entities can play the same role in the same relation.
Supports answering queries on subsets of the role-players.

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,384 @@
...
+            return new LocalEstimate.SimpleEstimate(list(var), estimate);
+        }
+
+        private LocalEstimate estimatesFromHasEdges(Concludable.Has concludableHas) {
+
+            long estimate = countPersistedHasEdges(concludableHas.asHas().owner().inferredTypes(), concludableHas.asHas().attribute().inferredTypes()) +
+                            estimateInferredAnswerCount(concludableHas, set(concludableHas.asHas().owner(), concludableHas.asHas().attribute()));
+
+            return new LocalEstimate.SimpleEstimate(list(concludableHas.asHas().owner(), concludableHas.asHas().attribute()), estimate);
+        }
```
krishnangovindraj, line 271: Captures that attribute ownership co-constrains the owner and the attribute

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,384 @@
...
+                        assert rule.conclusion().isVariableHas();
+                        return 0L;
+                    }
+                }).reduce(0L, Long::sum);
+
+                mostConstrainedAnswer = Math.min(mostConstrainedAnswer, answersGeneratedByThisResolvable);
+            }
+            assert mostConstrainedAnswer != Long.MAX_VALUE;
+            return mostConstrainedAnswer;
+        }
```
krishnangovindraj, line 235: For attribute types:, adds one to the unary estimate for each ExplicitHas rule concluding the type (Even if it already occurs in the data)
For relation-types: Sums across each rule concluding a matching type,  the estimated answers to the rule (i.e., to the condition of the rule projected to the variables in the head of the rule)

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,384 @@
...
+                    } else if (rule.conclusion().isExplicitHas()) {
+                        return 1L;
+                    }  else {
+                        assert rule.conclusion().isVariableHas();
+                        return 0L;
+                    }
+                }).reduce(0L, Long::sum);
+
+                mostConstrainedAnswer = Math.min(mostConstrainedAnswer, answersGeneratedByThisResolvable);
+            }
```
krishnangovindraj, line 232: If multiple resolvables constrain the same attribute/relation variable, pick the strongest constrained one (min answers).

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,384 @@
...
+                        Set<Variable> answerVariables = iterate(rule.conclusion().pattern().variables())
+                                .filter(v -> rule.condition().pattern().variables().contains(v)).toSet();
+                        return costEstimator.estimateAnswers(rule.condition().conjunction(), answerVariables);
+                    } else if (rule.conclusion().isExplicitHas()) {
+                        return 1L;
+                    }  else {
+                        assert rule.conclusion().isVariableHas();
+                        return 0L;
+                    }
+                }).reduce(0L, Long::sum);
```
krishnangovindraj, line 229: Sum number of answers across all applicable rules

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,384 @@
...
+                    Optional<Unifier> firstUnifier = unifiers.get(rule).stream().findFirst();
+                    // Increment the estimates for variables generated by the rule
+                    assert rule.conclusion().generating().isPresent() && firstUnifier.isPresent();
+
+                    if (rule.conclusion().isRelation()) { // Is a relation
+                        Set<Variable> answerVariables = iterate(rule.conclusion().pattern().variables())
+                                .filter(v -> rule.condition().pattern().variables().contains(v)).toSet();
+                        return costEstimator.estimateAnswers(rule.condition().conjunction(), answerVariables);
+                    } else if (rule.conclusion().isExplicitHas()) {
+                        return 1L;
```
krishnangovindraj, line 224: In retrospect, I could collect the values and take the set size

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,384 @@
...
+                iterate(concludablesInRetrievable).forEachRemaining(concludable -> {
+                    multivarEstimates.get(retrievable).addAll(computeEstimatesFromConcludable(concludable));
+                });
+            });
+
+            iterate(resolvables()).filter(this::isInferrable).map(Resolvable::asConcludable).forEachRemaining(concludable -> {
+                multivarEstimates.get(concludable).addAll(computeEstimatesFromConcludable(concludable));
+            });
+
+            // TODO: Negateds for the total -cost?
```
krishnangovindraj, line 183: TODO.

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,384 @@
...
+            return resolvable.isConcludable() &&
+                    !logicMgr.applicableRules(resolvable.asConcludable()).isEmpty();
+        }
+
+        private FunctionalIterator<Resolvable<?>> resolvables() {
+            return iterate(logicMgr.compile(conjunction));
+        }
+
+
+        private void initializeOnce() {
```
krishnangovindraj, line 145: Collects the "estimates" which are all we need to answer any of the queries of interest

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,384 @@
...
+            return Math.max(ret, 1);    // Don't do stupid stuff in prod when it happens.
+        }
+
+        private void registerTriggeredRules() {
+            resolvables().filter(this::isInferrable).map(Resolvable::asConcludable).forEachRemaining(concludable -> {
+                Map<Rule, Set<Unifier>> unifiers = logicMgr.applicableRules(concludable);
+                iterate(unifiers.keySet()).forEachRemaining(rule -> {
+                    costEstimator.registerConjunction(rule.condition().conjunction());
+                });
+            });
```
krishnangovindraj, line 132: Recursively register & initialise rules triggered by this conjunction. Cannot handle cycles yet.

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,384 @@
...
+            multivarEstimates.sort(Comparator.comparing(x -> x.answerEstimate(this, projectToVariables)));
+            for (LocalEstimate multivarEstimate : multivarEstimates) {
+                Set<Variable> interestingSubsetOfVariables = multivarEstimate.variables.stream()
+                        .filter(projectToVariables::contains).collect(Collectors.toSet());
+
+                long currentCostToCover = CostCover.costToCover(interestingSubsetOfVariables, costCover);
+                if (currentCostToCover > multivarEstimate.answerEstimate(this, interestingSubsetOfVariables)) {
+                    CostCover variablesNowCoveredBy = new CostCover(multivarEstimate.answerEstimate(this, interestingSubsetOfVariables));
+                    interestingSubsetOfVariables.forEach(v -> costCover.put(v, variablesNowCoveredBy));
+                }
```
krishnangovindraj, line 118: Do a greedy cover for the set of variables being queried, based on constraints applied by the resolvables included.

krishnangovindraj (commented):
Add comments

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,417 @@
...
+        }
+
+        private FunctionalIterator<Resolvable<?>> resolvables() {
+            return iterate(logicMgr.compile(conjunction));
+        }
+
+        private Set<Variable> allVariables() {
+            return iterate(conjunction.pattern().variables())
+                    .filter(v -> v.isThing())
+                    .toSet();
```
krishnangovindraj, line 152: We don't know how to deal with type variables yet, or whether we need to at all.

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,417 @@
...
+            if (concludable.isHas()) {
+                return list(estimatesFromHasEdges(concludable.asHas()));
+            } else if (concludable.isRelation()) {
+                return list(estimatesFromRolePlayersAssumingEvenDistribution(concludable.asRelation()));
+            } else {
+                return list();
+            }
+        }
+
+        private long computeInferredUnaryEstimates(ThingVariable generatedVariable) {
```
krishnangovindraj, line 219: For attribute types:, adds one to the unary estimate for each ExplicitHas rule concluding the type (Even if it already occurs in the data)
For relation-types: Sums across each rule concluding a matching type, the estimated answers to the rule (i.e., to the condition of the rule projected to the variables in the head of the rule)

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,417 @@
...
+                for (Unifier unifier : unifiers.get(rule)) {
+                    Set<Identifier.Variable> ruleSideIds = iterate(variablesOfInterest)
+                            .flatMap(v -> iterate(unifier.mapping().get(v.id()))).toSet();
+                    Set<Variable> ruleSideVariables = iterate(ruleSideIds).map(id -> rule.condition().conjunction().pattern().variable(id))
+                            .toSet();
+                    inferredEstimate += costEstimator.estimateAnswers(rule.condition().conjunction(), ruleSideVariables);
+                }
+            }
+            return inferredEstimate;
+        }
```
krishnangovindraj, line 277: Sums answer counts across all (rule, unifier)s

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,417 @@
...
+                    unaryEstimates.add(estimateFromTypes(v));
+                });
+
+                unaryEstimates.sort(Comparator.comparing(x -> x.answerEstimate(this, new HashSet<>(x.variables))));
+                for (LocalEstimate unaryEstimate : unaryEstimates) {
+                    if (!unaryCostCover.containsKey(unaryEstimate.variables.get(0))) {
+                        unaryCostCover.put(unaryEstimate.variables.get(0), new CostCover(unaryEstimate.answerEstimate(this, new HashSet<>(unaryEstimate.variables))));
+                    }
+                }
+            }
```
krishnangovindraj, line 174: A basic estimate based on types which will be refined.

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,417 @@
...
+            resolvables().forEachRemaining(resolvable -> multivarEstimates.put(resolvable, new ArrayList<>()));
+
+            //TODO: How does traversal handle type Variables?
+
+            resolvables().filter(Resolvable::isRetrievable).map(Resolvable::asRetrievable).forEachRemaining(retrievable -> {
+                Set<Concludable> concludablesInRetrievable = ResolvableConjunction.of(retrievable.pattern()).positiveConcludables();
+                iterate(concludablesInRetrievable).forEachRemaining(concludable -> {
+                    multivarEstimates.get(retrievable).addAll(computeEstimatesFromConcludable(concludable));
+                });
+            });
```
krishnangovindraj, line 187: (Retrievables processed first since they won't trigger rules)
Refine estimates by consider how concludables co-constrain variables

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,417 @@
...
+
+            iterate(resolvables()).filter(this::isInferrable).map(Resolvable::asConcludable).forEachRemaining(concludable -> {
+                multivarEstimates.get(concludable).addAll(computeEstimatesFromConcludable(concludable));
+            });
+
```
krishnangovindraj, line 192: Refine estimates by consider how concludables co-constrain variables

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,417 @@
...
+            if (this.negatedsCost == -1) {
+                this.negatedsCost = resolvables().filter(Resolvable::isNegated).map(Resolvable::asNegated)
+                        .flatMap(negated -> iterate(negated.disjunction().conjunctions()))
+                        .map(negatedConjunction -> {
+                            costEstimator.registerConjunction(negatedConjunction);
+                            return costEstimator.estimateAllAnswers(negatedConjunction);
+                        }).reduce(0L, Long::sum);
+            }
```
krishnangovindraj, line 200: Negations can't contribute to answers (though they do rule them out, but that's out of scope)
They do contribute to the cost of evaluating this condition though.

reasoner/planner/CostEstimator.java
```diff
@@ -0,0 +1,417 @@
...
+
+            // Can we prune the multiVarEstimates stored?
+            // Not if we want to order-resolvables. Else, yes based on queryableVariables.
+
+            Map<Variable, CostCover> fullCostCover = computeCostCover(iterate(multivarEstimates.values()).flatMap(Iterators::iterate).toList(), allVariables());
+            this.fullAnswerCount = CostCover.costToCover(allVariables(), fullCostCover) + this.negatedsCost;
+        }
```
krishnangovindraj, line 207: By now we have all the estimates needed to handle queries asking answerEstimates.

But we also will need an overall cost of evaluating the conjunction - approximated as the number of answers to all relevant variables in the conjunction.

krishnangovindraj (commented):
More comments

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,505 @@
...
+
+            // In cyclic paths
+            if (cycleHeads.size() == 1 && cycleHeads.contains(conjunction)) {
+                // We can reset and recompute all!
+                for (ResolvableConjunction conjunctionToReset : toReset) {
+                    estimators.get(conjunctionToReset).reset();
+                }
+                estimators.get(conjunction).initialize(); // no loop needed. This is the maximal cycle.
+            }
+        }
```
krishnangovindraj, line 98: Cycle-handling at the highest cycle-head.
Reset all the costEstimators in the cycle path.
Be lazy about re-evaluating them.

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,505 @@
...
+    }
+
+    private void reportInitializationCycle(ResolvableConjunction conjunction) {
+        cycleHeads.add(conjunction);
+        for (int i = initializationStack.size() - 1; initializationStack.get(i) != conjunction; i--) {
+            toReset.add(initializationStack.get(i));
+        }
```
krishnangovindraj, line 105: Mark all nodes in the cycle (except the cycle-head)  one as needing to be reset.

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,505 @@
...
+                }
+            }
+            return costCover;
+        }
+
+        private static long costOfEstimateCover(Set<Variable> variablesToConsider, Map<Variable, LocalEstimate> coverMap) {
+            Set<LocalEstimate> subsetCoveredBy = coverMap.keySet().stream().filter(variablesToConsider::contains)
+                    .map(coverMap::get).collect(Collectors.toSet());
+            return subsetCoveredBy.stream().map(estimate -> estimate.answerEstimate(variablesToConsider)).reduce(1L, (x, y) -> x * y);
+        }
```
krishnangovindraj, line 208: Computes the cost of covering a set of variables given a (partial) estimate cover.

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,487 @@
...
+                    long singleRelationEstimate = 1L;
+                    for (TypeVariable key : rolePlayerCounts.keySet()) {
+                        assert rolePlayerEstimates.containsKey(key);
+                        long avgRolePlayers = Double.valueOf(Math.ceil(rolePlayerEstimates.get(key) / relationTypeEstimate)).longValue();
+                        singleRelationEstimate *= nPermuteKforSmallK(avgRolePlayers, rolePlayerCounts.get(key));
+                    }
+
+                    long fullEstimate = Double.valueOf(Math.ceil(relationTypeEstimate * singleRelationEstimate)).longValue() + inferredRelationEstimate;
+                    return fullEstimate;
+                }
```
krishnangovindraj, line 389: Store just enough information to be able to query answers to subset of variables.

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,486 @@
...
+            for (Rule rule : unifiers.keySet()) {
+                for (Unifier unifier : unifiers.get(rule)) {
+                    Set<Identifier.Variable> ruleSideIds = iterate(variableFilter)
+                            .flatMap(v -> iterate(unifier.mapping().get(v.id()))).toSet();
+                    Set<Variable> ruleSideVariables = iterate(ruleSideIds).map(id -> rule.conclusion().conjunction().pattern().variable(id))
+                            .toSet();
+                    if (rule.conclusion().generating().isPresent() && ruleSideVariables.contains(rule.conclusion().generating().get())) {
+                        // There is one generated variable per combination of ALL variables in the conclusion
+                        ruleSideVariables = rule.conclusion().pattern().variables().stream().filter(Variable::isThing).collect(Collectors.toSet());
+                    }
```
krishnangovindraj, line 419: Generated variables are only in the conclusion and not in the condition - handle that.

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,486 @@
...
+                    rolePlayerEstimates.put(key, countPersistedRolePlayers(rp));
+                }
+            }
+
+            // TODO: Can improve estimate by collecting List<List<LocalEstimate>> from the triggered rules and doign sum(costCover).
+            // Consider owner in the inferred estimate call only if it's not anonymous
+            if (concludable.relation().owner().id().isName()) {
+                constrainedVars.add(concludable.relation().owner());
+            }
+            long inferredRelationsEstimate = estimateInferredAnswerCount(concludable, new HashSet<>(constrainedVars));
```
krishnangovindraj, line 457: Assumption made to keep things local. We can't have role-based projection without propagating constraints from the rules (possible, but likely overkill)

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,486 @@
...
+            if (this.negatedsCost == -1) this.negatedsCost = computeNegatedsCost();
+
+            List<LocalEstimate> allEstimates = resolvables().flatMap(resolvable -> iterate(estimatesFromResolvable(resolvable))).toList();
+            Map<Variable, LocalEstimate> fullCostCover = computeGreedyEstimateCoverForVariables(allVariables(), allEstimates);
+            this.fullAnswerCount = costOfEstimateCover(allVariables(), fullCostCover) + this.negatedsCost;
+
+            // Can we prune the multiVarEstimates stored?
+            // Not if we want to order-resolvables. Else, yes based on queryableVariables.
+            initializationStatus = InitializationStatus.COMPLETE;
+        }
```
krishnangovindraj, line 254: Initialisation recursively computes all the rules and prepares all the LocalEstimates needed to answer queries. 
All recursion is handled at initialisation. The queries are answered without needing to recurse.

krishnangovindraj (commented):
Add comments

logic/LogicManager.java
```diff
@@ -69,7 +69,7 @@ public LogicManager(GraphManager graphMgr, ConceptManager conceptMgr, TraversalE
...
```
flyingsilverfin, line 72: I think rather than getting the `GraphManager` from the `LogicManager` in the `CostEstimator` (and making this public), it would be better to pass it in as another argument to the constructor of the estimator.

One reason we have constructor arguments is to give a high-level signal of the main collaborators of the class. In your domain, answer estimation cannot be done without access to the graph, so we should highlight that by passing in the constructor explicitly.

reasoner/BUILD
flyingsilverfin: I had a hard think about this and only see one option for how we could avoid the need for this dependency.

The reason it's important to be deliberate here is that opening up a new top-level dependency (look at `docs/package-structure.png`) is always a big risk - you're allowing access between two packages that were otherwise isolated. If you do this without much regard you kind of end up with everything depending on everything, which makes it much harder later to refactor packages & untangle the code later.

The option I see is that we move the `AnswerCountEstimator` into the `logic` package. This would avoid opening up this major new dependency. It also wouldn't by so bad domain-wise because answer estimation of a `ResolvableConjunction` could be located near where `Resolvables` exist.

Then, the estimator could be a member of the `LogicManager`? What do we think of the proposal...

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,486 @@
...
+    private final LogicManager logicMgr;
+    private final Map<ResolvableConjunction, ConjunctionAnswerCountEstimator> estimators;
+
+    // Cycle handling
+    private final ArrayList<ResolvableConjunction> initializationStack;
+    private final Set<ResolvableConjunction> cycleHeads;
+    private final Set<ResolvableConjunction> toReset;
+    private final AnswerCountModel answerCountModel;
+
+    public AnswerCountEstimator(LogicManager logicMgr) {
```
flyingsilverfin, line 59: add `GraphManager` argument explicitly

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,486 @@
...
+    private void registerConjunctionAndInitialize(ResolvableConjunction conjunction) {
+        if (!estimators.containsKey(conjunction)) {
+            estimators.put(conjunction, new ConjunctionAnswerCountEstimator(this, conjunction));
+        }
+
+        if (estimators.get(conjunction).mayInitialize()) {
+            initializationStack.add(conjunction);
+            estimators.get(conjunction).initialize();
+            assert initializationStack.get(initializationStack.size() - 1) == conjunction;
+            initializationStack.remove(initializationStack.size() - 1);
```
flyingsilverfin, line 91: correct me if I'm wrong, but the fact that the `stack` is a class variable not a local variable that is passed through the recursion explicitly is for convenience?

What keeping it as a class variable is that it looks like somehow the stack is going to be important in the next invocation by the user of `estimateAnswers`... however, here we clearly `add` and then `remove` the same item from the stack!

I think the right process would be to pass the stack into the `registerConjunctionAndInitialise` method. The entry points (`estimateAllAnswers` etc.) will initialise an empty stack to use.

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,486 @@
...
+    public long estimateAllAnswers(ResolvableConjunction conjunction) {
+        registerConjunctionAndInitialize(conjunction);
+        return estimators.get(conjunction).estimateAllAnswers();
+    }
+
+    public long estimateAnswers(ResolvableConjunction conjunction, Set<Variable> variableFilter) {
+        return estimateAnswers(conjunction, variableFilter, logicMgr.compile(conjunction));
+    }
+
+    public long estimateAnswers(ResolvableConjunction conjunction, Set<Variable> variableFilter, Set<Resolvable<?>> includedResolvables) {
```
flyingsilverfin, line 77: finer detail:
the order of the arguments would be better done as `Conjunction, Resolvables, Filter`.
Because we tend to read left to right, it makes more sense to think of the arguments as "estimate answers from this conjunction within these resolvables then narrowed to this filter". The other way around, "estimate answers from this conjunction, narrows to this filter, then within these resolvables" is conceptually reversing the dependencies

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,486 @@
...
+        private final AnswerCountModel answerCountModel;
+        private Map<Variable, LocalEstimate> unaryEstimateCover;
+        private Map<Resolvable<?>, List<LocalEstimate>> retrievableEstimates;
+        private Map<Resolvable<?>, LocalEstimate> inferrableEstimates;
+        private Map<Resolvable<?>, LocalEstimate> unaryEstimates;
+
+        private long fullAnswerCount;
+        private long negatedsCost;
+
+        private enum InitializationStatus {NOT_STARTED, RESET, IN_PROGRESS, COMPLETE}
```
flyingsilverfin, line 123: "initialisation" is not something that really has states normally. Something else usually has the state of "initialising" for example. I'd propose `EstimationStatus {EMPTY, RESET, PARTIAL, COMPLETE}` for example.

If using this naming, we further get hints that the `mayInitialise` can be renamed - maybe `hasEstimate`? Then, `initialise()` could be `computeEstimate()`?

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,486 @@
...
+    private static class ConjunctionAnswerCountEstimator {
+
+        private final ResolvableConjunction conjunction;
+        private final AnswerCountEstimator answerCountEstimator;
+        private final LogicManager logicMgr;
+        private final AnswerCountModel answerCountModel;
+        private Map<Variable, LocalEstimate> unaryEstimateCover;
+        private Map<Resolvable<?>, List<LocalEstimate>> retrievableEstimates;
+        private Map<Resolvable<?>, LocalEstimate> inferrableEstimates;
+        private Map<Resolvable<?>, LocalEstimate> unaryEstimates;
```
flyingsilverfin, line 118: a `Map` can usually be final, and initialised in the constructor. `reset()` can then `clear()` the map.

I believe you're making them nullable also to know when to recompute these values - however I expect that this signal should be derivable from the `Status` flag!

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,486 @@
...
+            this.initializationStatus = InitializationStatus.NOT_STARTED;
+        }
+
+        private FunctionalIterator<Resolvable<?>> resolvables() {
+            return iterate(logicMgr.compile(conjunction));
+        }
+
+        private boolean isInferrable(Resolvable<?> resolvable) {
+            return resolvable.isConcludable() &&
+                    !logicMgr.applicableRules(resolvable.asConcludable()).isEmpty();
```
flyingsilverfin, line 143: in the context of where you're using this, you're always getting the resolvables from the `compile()` method. This means we've already filtered out the concludables that do not have applicable rules! This means this method can simply be replaced by `resolvable.isConcludable()`

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,486 @@
...
+        private void registerTriggeredRules() {
+            resolvables().filter(this::isInferrable).map(Resolvable::asConcludable).forEachRemaining(concludable -> {
+                Map<Rule, Set<Unifier>> unifiers = logicMgr.applicableRules(concludable);
+                iterate(unifiers.keySet()).forEachRemaining(rule -> {
+                    answerCountEstimator.registerConjunctionAndInitialize(rule.condition().conjunction());
+                });
+            });
+        }
+
+        private void initialize() {
```
flyingsilverfin, line 227: if this is `computeEstimate()`, which does the main work of computing the estimate...
Then if we see what `registerTriggeredRules` does, it actually computes the estimates of the recursive rules so it should be called `computeRecursiveEstimates`

We can then follow that naming convention to replace the use of `estimatexxxxInferrables` with using the `recursive` term so we're consistent (general practice - a different term somewhere can really help align the different methods and state everywhere that weren't quite smooth reading before!)

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,486 @@
...
+
+            registerTriggeredRules();
+
+            if (inferrableEstimates == null) {
+                inferrableEstimates = deriveEstimatesFromInferrables();
+                unaryEstimates = computeUnaryEstimatesWithInference(unaryEstimateCover);
+                unaryEstimateCover = computeFinalUnaryEstimateCover(unaryEstimateCover); // recompute with inferred
+            }
+
+            if (this.negatedsCost == -1) this.negatedsCost = computeNegatedsCost();
```
flyingsilverfin, line 245: interestingly, by the time we get there, these should always be null? because we only get past the initial `if` when the state is the first state or the RESET state.

We acutally mostly want to make sure that we only do the `unaryEstimateCover` and `retrievableEstimates` once, because they aren't affected by recursion! So we can protect that with the `state == NOT_STARTED/EMPTY`!

flyingsilverfin (changes requested):
I've done a pass over the first couple layers of implementation.
On the whole I think this is a big step in the right direction already! I believe another pass or two can make it even simpler and easier to understand. There are some big high level ideas are probably worth doing first before reviewing the details.

reasoner/planner/AnswerCountEstimator.java
krishnangovindraj: Does this still hold when you have two variants of the same method, where one has an extra argument?
This is the equivalent of having a default argument value in languages which support it. Such languages often require arguments with default values to be after all arguments without default values.

reasoner/planner/AnswerCountEstimator.java
krishnangovindraj: Then the AnswerCountEstimators could be aware of their place in the graph, which they don't need to be.
Is it worth passing an argument to a class if it's only going to be passed back to this one?

reasoner/planner/AnswerCountEstimator.java
flyingsilverfin: it can get tricky with overloading to be honest. It's a good principle but sometimes the "extra argument" must come at the end to maintain symmetry with the other versions of the method:
`run(A, B)`
`run(A,B,C)` -- this version C would actually read better before `B` but we prioritise the symmetry with the other version of this method

reasoner/planner/AnswerCountEstimator.java
flyingsilverfin: I think it makes sense:
```computeAnswers() {
  visited = new Stack()
  estimator.computeEstimate(visited)
}

Estimator {

  computeEstimate(visited) {
    visited.add(conjunction)
    computeRecursiveEstimates(visited)
    visited.pop()
    }
 }
```

krishnangovindraj: That does look nice and pure.
But now, Should I also pass cycle_heads and reset through as well? Because they're all related to each other. I could also wrap them into a context object.

flyingsilverfin: do `cycle_heads` and `reset` affect the operation of the next call to `computeAnswers`? I believe so right? In that case, they're still ok. The `stack` however is purely used within a single recursive computation.

Will have to see how it looks after this change

reasoner/BUILD
krishnangovindraj: The estimator has very little to do with the rest of the contents of the Logic package.
I would imagine there are layers in the packages where higher layers are built upon the lower layers. Adding a dependency between two packages which are clearly in different layers wouldn't lead to cyclic dependencies.
Also, they weren't directly dependent, but they weren't exactly isolated either, given it (and almost every other package) transitively depends on the graph package.

Between moving the AnswerCountEstimator to logic (where it's out of place, given it's effectively a cost-model for planning) and adding a direct edge to a dependency which is already transitively present, I'd expect the latter to be the lesser sin.

reasoner/planner/AnswerCountEstimator.java
krishnangovindraj: They all work together in a single-call to do cycle-detection. By the time the stack is empty, reset and cycle_heads will be empty.

reasoner/planner/AnswerCountEstimator.java
krishnangovindraj: That would need me to rename the estimateAnswers methods to avoid ambiguity.

Currently, there's a (recursive) initialisation phase which collects the local estimates, and a query phase which uses these  local estimates to answer complex queries (without having to recurse).
`needsPreparation` and `prepareLocalEstimates` may be a better fit? `hasEstimate` is not the write set of conditions, because that would correspond to `state==COMPLETE`.

reasoner/planner/AnswerCountEstimator.java
krishnangovindraj: I agree that the status flag is the ground-truth for deciding whether to (re-)compute the fields.
But I would rather not initialize them to an empty set, since the final result can also be an empty set. Isn't it better to have a loud crash when it's accessed before it's been computed, than to have a silent failure?

reasoner/planner/AnswerCountEstimator.java
krishnangovindraj: That would move the setting of InitializationState = IN_PROGRESS to somewhere in the middle of the function which is a bit weird for tracking what's going on. I could use a temporary flag, but the ==null check was used in a lot of places in reasoner to enforce that it's computed only once and then cached .

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,507 @@
...
+
+                public CoPlayerEstimate(List<Variable> variables, double relationTypeEstimate,
+                                        Map<TypeVariable, Long> rolePlayerEstimates, Map<TypeVariable, Integer> rolePlayerCounts,
+                                        long inferredRelationEstimate) {
+                    super(variables);
+                    this.relationTypeEstimate = relationTypeEstimate;
+                    this.rolePlayerEstimates = rolePlayerEstimates;
+                    this.rolePlayerCounts = rolePlayerCounts;
+                    this.inferredRelationEstimate = inferredRelationEstimate;
+                }
```
krishnangovindraj, line 396: Store just enough information to be able to reasonably answer queries on subsets of role-players.
Since we can't accurately estimate the co-constraints between role-players in inferred relations (without propagating constraints from triggered rules), that's an approximation.

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -0,0 +1,500 @@
...
+
+                initializationStatus = InitializationStatus.COMPLETE;
+            }
+        }
+
+        private void initializeAcyclicEstimates() {
+            this.estimatesFromResolvable.putAll(initializeEstimatesFromRetrievables());
+            this.estimatesFromResolvable.putAll(initializeEstimatesFromNegations());
+            this.estimatesFromResolvable.putAll(initializeEstimatesFromAcyclicConcludables());
+            iterate(cyclicConcludables).forEachRemaining(concludable -> this.estimatesFromResolvable.put(concludable, list()));
```
krishnangovindraj, line 239: Empty estimates to allow us to initialise dependencies which lead to cycles.

krishnangovindraj (commented):
Add comments

flyingsilverfin (approved):
Looks good!

PR #6641, opened 2022-09-08
Optimise sort queries by utilising sorted data on disk

flyingsilverfin:

TypeDB can leverage the sorted nature of the data storage engine for non-string attributes, to implement optimised sort queries. Previously, TypeDB would inefficiently gather all answers to a query, sort them, and then output them.

TypeDB also now supports sorting different each variable in their own direction (eg. ascending or descending).

The main change is that TypeDB pins the sort variables at the start of the query execution
  * This guarantees that the sort variables are produced in order!
  * The rest of the query is traversed from the fixed start points, guaranteeing we find all answers in the user's sort order
  * the query planner implements this with new ordering constraints, and an updated plan initialiser  
  * we introduce a validation before any of query executes to ensure that all concepts retrieved within the sorted variable are comparable - eg. if `$a` can be `datetime` and `long`, these are not comparable, so the query is rejected.

We also push the traversal modifiers (`filter` and `sort`) into the query planner and traversal execution, and wrap them in new server-side constructs nested in `//traversal/common:Modifiers`. This enables the query planner to be aware fo the sorting requirements, and in the future of the filtering requirements as well. In addition, the traversal can then use the right ascending or descending sort order for the sorted vertices.

`ConceptMap` has a new comparable subtype called `ConceptMap.Sortable`, which contains a comparator for sorting the concept maps according to some of the variables in the concept map. This allows us to use our standard `SortedIterator` implementations to handle ordered concept map streams.
* as a result, we also use `? extends ConceptMap` wherever the `Matcher.execute()` returns to handle the possibility of getting sortable concept maps.
* we also move the comparators used to sort concept map lists into `ConceptMap.Comparator`

Other changes
* update `FunctionalIterator`s and `SortedIterator`s to allow merge sorting and mapping more freely, as well as renaming to disambiguate methods overloaded purely generic parameters, which Java can't otherwise distinguish
* remove redundant `Context` arguments throughout the querying APIs, such as in `Reasoner.java`
* low-level APIs in the `//graph` package have optional arguments to enforce an ordering of the returned iterator of vertices. For example, `Forwadable<ThingVertx, ORDER> thingGraph.getReadable(TypeVertex type, ORDER order)` now has an order argument and parameter


query/Matcher.java
```diff
@@ -118,71 +109,13 @@ public static Matcher.Group.Aggregator create(Reasoner reasoner, TypeQLMatch.Gro
...
         assert context != null;
         return execute(context);
     }
 
-    FunctionalIterator<ConceptMap> execute(Context.Query context) {
-        FunctionalIterator<ConceptMap> answers = reasoner.execute(disjunction, query.modifiers(), context);
-        // TODO: we should remove these and handle them in the traversal engine or reasoner ONLY. Currently in reasoner already
-        if (query.modifiers().sort().isPresent()) answers = sort(answers, query.modifiers().sort().get());
-        if (query.modifiers().offset().isPresent()) answers = answers.offset(query.modifiers().offset().get());
-        if (query.modifiers().limit().isPresent()) answers = answers.limit(query.modifiers().limit().get());
```
flyingsilverfin, line 131: sort, offset, and limit are all moved into the `Reasoner`. The sorting `Comparator` is moved into `ConceptMap.Ordered`

reasoner/Reasoner.java
```diff
@@ -73,41 +78,31 @@ public Reasoner(ConceptManager conceptMgr, LogicManager logicMgr,
         this.conceptMgr = conceptMgr;
         this.traversalEng = traversalEng;
         this.logicMgr = logicMgr;
-        Context.Query defaultContext = new Context.Query(context, new Options.Query());
-        defaultContext.producer(Either.first(EXHAUSTIVE));
```
flyingsilverfin, line 77: the default context turns out to be redundant, so we delete it

reasoner/Reasoner.java
```diff
@@ -73,41 +78,31 @@ public Reasoner(ConceptManager conceptMgr, LogicManager logicMgr,
...
-        return mayReason(disjunction);
-    }
-
-    private boolean mayReason(Disjunction disjunction) {
-        for (Conjunction conj : disjunction.conjunctions()) {
-            Set<Variable> vars = conj.variables();
-            List<Negation> negs = conj.negations();
-            if (iterate(vars).flatMap(v -> iterate(v.inferredTypes())).distinct().anyMatch(this::hasRule)) return true;
-            if (!negs.isEmpty() && iterate(negs).anyMatch(n -> mayReason(n.disjunction()))) return true;
+    public FunctionalIterator<? extends ConceptMap> execute(Disjunction disjunction, TypeQLMatch.Modifiers modifiers, Context.Query context) {
```
flyingsilverfin, line 88: `execute()` is actually the main entry point into this class, so we move it to the top and do all the main switching logic at the top

reasoner/Reasoner.java
```diff
@@ -73,41 +78,31 @@ public Reasoner(ConceptManager conceptMgr, LogicManager logicMgr,
...
+        FunctionalIterator<? extends ConceptMap> answers;
+        Filter filter = Filter.create(modifiers.filter());
+        if (mayReason(disjunction, context)) {
+            answers = executeReasoner(disjunction, filter, context);
+            if (modifiers.sort().isPresent()) answers = eagerSort(answers, Sorting.create(modifiers.sort().get()));
+        } else if (modifiers.sort().isPresent() && isNativelySortable(disjunction, modifiers.sort().get())) {
+            answers = executeTraversalSorted(disjunction, filter, Sorting.create(modifiers.sort().get()));
+        } else {
+            answers = executeTraversal(disjunction, context, filter);
+            if (modifiers.sort().isPresent()) answers = eagerSort(answers, Sorting.create(modifiers.sort().get()));
```
flyingsilverfin: branches:
1. execute reasoner, and possibly apply eager sorting (ie. retrieve all answers, then sort)
2. use natively sorted traversal
3. use a regular traversal and, and possibly apply eager sorting

reasoner/Reasoner.java
```diff
@@ -73,41 +78,31 @@ public Reasoner(ConceptManager conceptMgr, LogicManager logicMgr,
...
-    private boolean hasRule(Label type) {
-        return logicMgr.rulesConcluding(type).hasNext() || logicMgr.rulesConcludingHas(type).hasNext();
-    }
-
-    public FunctionalIterator<ConceptMap> execute(Disjunction disjunction, TypeQLMatch.Modifiers modifiers, Context.Query context) {
-        inferAndValidateTypes(disjunction);
-        if (mayReason(disjunction, context)) return executeReasoner(disjunction, filter(modifiers.filter()), context);
-        else return executeTraversal(disjunction, context, filter(modifiers.filter()));
+        if (modifiers.offset().isPresent()) answers = answers.offset(modifiers.offset().get());
+        if (modifiers.limit().isPresent()) answers = answers.limit(modifiers.limit().get());
```
flyingsilverfin, line 104: offset and limit moved in from the `Matcher`

reasoner/Reasoner.java
```diff
@@ -140,16 +129,56 @@ private Set<Conjunction> incoherentConjunctions(Disjunction disjunction) {
         return causes;
     }
 
-    public FunctionalIterator<ConceptMap> executeReasoner(Disjunction disjunction, Set<Identifier.Variable.Retrievable> filter,
-                                                          Context.Query context) {
+    private boolean mayReason(Disjunction disjunction, Context.Query context) {
```
flyingsilverfin, line 132: these methods were all just moved around

reasoner/Reasoner.java
```diff
@@ -158,42 +187,70 @@ public FunctionalIterator<ConceptMap> executeTraversal(Disjunction disjunction,
...
+                    .filter(ans -> !isNegated(ans, conjunction.negations()))
+                    .map(conceptMap -> conceptMap.filter(filter)).distinct();
+        }
+    }
+
+    private SortedIterator<ConceptMap.Ordered, SortedIterator.Order.Asc> iteratorSorted(Conjunction conjunction,
+                                                                                        Filter filter, Sorting sorting) {
+        ConceptMap.Ordered.Comparator comparator = ConceptMap.Ordered.Comparator.create(sorting);
+        SortedIterator<ConceptMap.Ordered, SortedIterator.Order.Asc> answers = traversalEng.iterator(conjunction.traversal(filter, sorting))
+                .mapSorted(vertexMap -> conceptMgr.conceptMapOrdered(vertexMap, comparator), ASC);
```
flyingsilverfin, line 236: execute a traversal, knowing that the answers will respect the sorting required, then map it to the ordered iterator of comparable concept maps

flyingsilverfin: this is a compromise between creating a parallel API that returns a sorted iterator of comparable `VertexMap` on the traversal engine (more type safe) and simplicity...

reasoner/Reasoner.java
```diff
@@ -158,42 +187,70 @@ public FunctionalIterator<ConceptMap> executeTraversal(Disjunction disjunction,
...
+        if (conjunction.negations().isEmpty()) return answers;
+        else {
+            return answers.filter(ans -> !isNegated(ans, conjunction.negations()))
+                    .mapSorted(conceptMap -> conceptMap.filter(filter), ASC).distinct();
         }
     }
 
     private Conjunction bound(Conjunction conjunction, ConceptMap bounds) {
-        Conjunction newClone = conjunction.clone();
-        newClone.bound(bounds.toMap(Type::getLabel, Thing::getIID));
```
flyingsilverfin, line 195: we deleted the `toMap` method on `ConceptMap` in order to simplify it, and move the implementation into the only usage here.

reasoner/ReasonerProducer.java
```diff
@@ -170,10 +169,10 @@ public synchronized void receiveAnswer(ConceptMap answer) {
         public static class Conjunction extends Match {
 
             private final com.vaticle.typedb.core.pattern.Conjunction conjunction;
-            private final Set<Identifier.Variable.Retrievable> filter;
+            private final Modifiers.Filter filter;
```
flyingsilverfin, line 172: new class to hold filter

reasoner/controller/ControllerRegistry.java
```diff
@@ -74,7 +76,7 @@ public class ControllerRegistry {
     private TypeDBException terminationCause;
 
     public ControllerRegistry(ActorExecutorGroup executorService, TraversalEngine traversalEngine, ConceptManager conceptMgr,
-                              LogicManager logicMgr, com.vaticle.typedb.core.common.parameters.Context.Query context) {
```
flyingsilverfin, line 77: we don't actually need the context anywhere, so we delete it

reasoner/controller/ControllerRegistry.java
```diff
@@ -141,15 +143,15 @@ public void terminate(Throwable e) {
         return controller;
     }
 
-    public void createRootConjunction(Conjunction conjunction, Set<Variable.Retrievable> filter,
+    public void createRootConjunction(Conjunction conjunction, Modifiers.Filter filter,
```
flyingsilverfin, line 146: use `Modifiers.Filter`, which holds a filter for a query, where

traversal/BUILD
```diff
@@ -42,6 +42,7 @@ native_java_libraries(
         "@vaticle_typedb_common//:common",
         "@vaticle_typeql//java/common:common",
         "@vaticle_typeql//java/pattern:pattern",
+        "@vaticle_typeql//java/query:query",
```
flyingsilverfin, line 45: beause we can construct `Modifier.Filter` and `Modifier.Sorting` from the TypeQL representations, we have to allow accessing the TypeQL construct from the `//traversal` package. This seems acceptable, especially since we do something similar for `Identifier` already

traversal/GraphTraversal.java
```diff
@@ -56,23 +54,26 @@
 import static com.vaticle.typedb.core.graph.common.Encoding.Edge.Type.SUB;
 import static com.vaticle.typeql.lang.common.TypeQLToken.Predicate.SubString.LIKE;
 
+// TODO: We should not this object as a builder, as the hash and equality functions would change
+//       We should introduce a separate "builder pattern" to Traversal, such that users of this library will build
+//       traversals with Traversal.Builder, and call .build() in the end to produce a final Object.
```
flyingsilverfin, line 59: moved this TODO from below

traversal/GraphTraversal.java
```diff
@@ -56,23 +54,26 @@
...
 
-    abstract Set<Identifier.Variable.Retrievable> filter();
+    public void filter(Modifiers.Filter filter) {
+        modifiers.filter(filter);
+    }
 
-    public abstract void filter(Set<? extends Identifier.Variable.Retrievable> filter);
+    public void sort(Modifiers.Sorting sorting) {
+        modifiers.sorting(sorting);
+    }
```
flyingsilverfin, line 72: old API `filter` takes the new type `Modifiers.Filter`, and we add an analogous `sort` API

traversal/GraphTraversal.java
```diff
@@ -130,154 +131,108 @@ public Type() {
...
-        //       traversals with Traversal.Builder, and call .build() in the end to produce a final Object.
-        @Override
-        Set<Identifier.Variable.Retrievable> filter() {
-            if (filter.isEmpty()) {
-                modifiable = false;
-                iterate(structure.vertices()).filter(v -> v.id().isRetrievable()).map(v -> v.id().asVariable().asRetrievable())
-                        .toSet(filter);
-            }
-            return filter;
-        }
```
flyingsilverfin, line 198: creating the default filter has moved up to the user, eg. `Conjunction.traversal(filter, sorting)`, which sets the filter to be all retrievable variables if the user filter is empty.

traversal/GraphTraversal.java
```diff
@@ -130,154 +131,108 @@ public Type() {
...
-
-        @Override
-        public void filter(Set<? extends Identifier.Variable.Retrievable> filter) {
-            assert modifiable && iterate(filter).noneMatch(Identifier::isLabel);
-            this.filter.addAll(filter);
-        }
-
         public void equalThings(Identifier.Variable thing1, Identifier.Variable thing2) {
-            assert modifiable;
```
flyingsilverfin, line 207: using this to protect us against using the class as a Builder object was incorrect anyway, and now the place where we set `modifiable` to false is no longer existing. Instead, we delete the check and move the TODO to the top level of the class, where it should be implemented for all types of `GraphTraversal`

traversal/GraphTraversal.java
```diff
@@ -289,34 +244,14 @@ public void predicate(Identifier.Variable attribute, TypeQLToken.Predicate.Equal
...
-            // We compare this.filter() instead of this.filter, as the property is dynamically calculated
-            return (this.structure.equals(that.structure) &&
-                    this.parameters.equals(that.parameters) &&
-                    this.filter().equals(that.filter()));
-        }
-
-        @Override
-        public int hashCode() {
-            return Objects.hash(this.structure, this.parameters, this.filter);
-        }
```
flyingsilverfin, line 319: equality is actually the same for all `Traversal` objects

traversal/planner/GraphPlanner.java
```diff
@@ -78,8 +79,10 @@ public class GraphPlanner implements ConnectedPlanner {
 
     private volatile double totalCostLastRecorded;
     private double totalCost;
+    private final Modifiers modifiers;
 
-    private GraphPlanner() {
+    private GraphPlanner(Modifiers modifiers) {
+        this.modifiers = modifiers;
```
flyingsilverfin, line 85: planner is now modifiers-aware

traversal/planner/GraphPlanner.java
```diff
@@ -179,6 +182,13 @@ private void createOptimiserConstraints() {
                 conOneVertexAtOrderI.setCoefficient(vertex.varOrderAssignment[i], 1);
             }
         }
+        for (int i = 0; i < modifiers.sorting().variables().size(); i++) {
+            Identifier.Variable.Retrievable sortVariable = modifiers.sorting().variables().get(i);
+            PlannerVertex<?> vertex = vertices.get(sortVariable);
+            assert vertex != null;
+            OptimiserConstraint conSetVertexAtOrderI = optimiser.constraint(1, 1, conPrefix + "set_vertex_at_order_" + i);
+            conSetVertexAtOrderI.setCoefficient(vertex.varOrderAssignment[i], 1);
+        }
```
flyingsilverfin, line 195: the ordering variables are pinned to the beginning of the plan. This guarantees that we generate answers in an ordered fashion

traversal/planner/Planner.java
```diff
@@ -28,19 +29,19 @@
...
-        else return MultiPlanner.create(retrievedStructures);
+    static Planner create(Structure structure, Modifiers modifiers) {
+        List<Structure> retrievedStructures = splitStructure(structure, modifiers);
+        if (retrievedStructures.size() == 1) return ConnectedPlanner.create(retrievedStructures.get(0), modifiers);
+        else return MultiPlanner.create(retrievedStructures, modifiers);
     }
 
     /**
-     * Filter out structures that are not retrievable, such as disconnected type vertices
-     * TODO: We can further improve this if we have access to the filter as well
```
flyingsilverfin, line 39: we implement the optimisation that removes subgraphs which are not retrieved at all

traversal/procedure/GraphProcedure.java
```diff
@@ -109,49 +110,41 @@ public int vertexCount() {
...
             LOG.trace(params.toString());
             LOG.trace(this.toString());
         }
-        assertWithinFilterBounds(filter);
-        if (initialVertex().id().isRetrievable() && filter.contains(initialVertex().id().asVariable().asRetrievable())) {
-            return async(initialVertex().iterator(graphMgr, params).map(v ->
-                    new GraphIterator(graphMgr, v, this, params, filter).distinct()
+        if (initialVertex().id().isRetrievable() && modifiers.filter().variables().contains(initialVertex().id().asVariable().asRetrievable())) {
+            return async(initialVertex().iterator(graphMgr, params, modifiers.sorting()).map(v ->
+                    new GraphIterator(graphMgr, v, this, params, modifiers).distinct()
```
flyingsilverfin, line 122: traversal execution now receives the modifiers. Openinga vertex iterator requires the sorting modifiers to decide whether to retrieve in ascending or descending order

traversal/procedure/ProcedureEdge.java
```diff
@@ -240,8 +240,8 @@ public boolean equals(Object o) {
             assert fromVertex.isThing() && fromVertex.asThing().isAttribute();
 
             Forwardable<? extends ThingVertex, Order.Asc> toIter;
-            if (to.props().hasIID()) toIter = to.iterateAndFilterFromIID(graphMgr, params);
-            else toIter = to.iterateAndFilterFromTypes(graphMgr, params);
+            if (to.props().hasIID()) toIter = to.iterateAndFilterFromIID(graphMgr, params, ASC);
+            else toIter = to.iterateAndFilterFromTypes(graphMgr, params, ASC);
```
flyingsilverfin, line 244: require that vertex iterators provide an ordering

traversal/procedure/ProcedureEdge.java
```diff
@@ -834,7 +834,7 @@ Forwardable<ThingVertex, Order.Asc> forwardBranchToRole(GraphManager graphMgr, V
                         graphMgr.schema().relatedRoleTypes(type) : graphMgr.schema().playedRoleTypes(type);
                 return iterate(roleTypes)
                         .filter(rt -> to.props().types().contains(rt.properLabel()))
-                        .mergeMap(t -> relationOrPlayer.outs().edge(encoding, PrefixIID.of(VERTEX_ROLE), t.iid()).to(), ASC);
+                        .mergeMapForwardable(t -> relationOrPlayer.outs().edge(encoding, PrefixIID.of(VERTEX_ROLE), t.iid()).to(), ASC);
```
flyingsilverfin, line 837: new naming throughout

traversal/procedure/ProcedureVertex.java
```diff
@@ -157,69 +160,74 @@ public boolean overlaps(Thing other, Traversal.Parameters params) {
         }
 
         @Override
-        public Forwardable<? extends ThingVertex, Order.Asc> iterator(GraphManager graphMgr, Traversal.Parameters parameters) {
-            if (props().hasIID()) return iterateAndFilterFromIID(graphMgr, parameters);
-            else return iterateAndFilterFromTypes(graphMgr, parameters);
+        public Forwardable<? extends ThingVertex, Order> iterator(GraphManager graphMgr, Traversal.Parameters parameters,
+                                                                  Modifiers.Sorting sorting) {
+            Order order = sorting.isAscending(id()) ? ASC : DESC;
```
flyingsilverfin, line 165: convert sorting direction into iterator ordering, which we use throughout this class in order to get an iterator ordered in the right way

traversal/procedure/ProcedureVertex.java
```diff
@@ -157,69 +160,74 @@ public boolean overlaps(Thing other, Traversal.Parameters params) {
...
-                                                                              Traversal.Parameters parameters) {
+        <ORDER extends Order> Forwardable<? extends AttributeVertex<?>, ORDER> filterPredicates(Forwardable<? extends AttributeVertex<?>, ORDER> iterator,
+                                                                                                Traversal.Parameters parameters) {
             return filterPredicates(iterator, parameters, null);
         }
 
-        Forwardable<? extends AttributeVertex<?>, Order.Asc> filterPredicates(Forwardable<? extends AttributeVertex<?>, Order.Asc> iterator,
-                                                                              Traversal.Parameters parameters,
-                                                                              @Nullable Predicate.Value<?> exclude) {
+        <ORDER extends Order> Forwardable<? extends AttributeVertex<?>, ORDER> filterPredicates(Forwardable<? extends AttributeVertex<?>, ORDER> iterator,
```
flyingsilverfin, line 228: these methods are all parametrised by the order, which is a typing guarantee that we always maintain the ordering of the iterator we receive/return or meet the ordering required indicated by the arguments

traversal/procedure/ProcedureVertex.java
```diff
@@ -310,20 +319,22 @@ protected Properties.Type newProperties() {
...
-                if (mustBeAttributeType()) return graphMgr.schema().attributeTypes();
-                else if (mustBeRelationType()) return graphMgr.schema().relationTypes();
-                else if (mustBeRoleType()) return graphMgr.schema().roleTypes();
-                else if (mustBeThingType()) return graphMgr.schema().thingTypes();
-                else iterator = graphMgr.schema().thingTypes().merge(graphMgr.schema().roleTypes());
+                if (mustBeAttributeType()) return graphMgr.schema().attributeTypes(order);
+                else if (mustBeRelationType()) return graphMgr.schema().relationTypes(order);
+                else if (mustBeRoleType()) return graphMgr.schema().roleTypes(order);
+                else if (mustBeThingType()) return graphMgr.schema().thingTypes(order);
+                else iterator = graphMgr.schema().thingTypes(order).merge(graphMgr.schema().roleTypes(order));
```
flyingsilverfin, line 337: pass required `order` into each of the iterator generation methods

traversal/procedure/VertexProcedure.java
```diff
@@ -60,42 +59,42 @@ public static VertexProcedure create(StructureVertex<?> structureVertex) {
...
+        FunctionalIterator<? extends Vertex<?, ?>> iterator = vertex.iterator(graphMgr, params, modifiers.sorting());
         for (ProcedureEdge<?, ?> e : vertex.loops()) {
             iterator = iterator.filter(v -> e.isClosure(graphMgr, v, v, params));
         }
 
         return iterator.map(v -> VertexMap.of(map(pair(vertex.id().asVariable().asRetrievable(), v))));
     }
 
+    @Override
+    public String toString() {
```
flyingsilverfin, line 85: reordering

traversal/structure/Structure.java
```diff
@@ -107,24 +110,36 @@ private void recordEdge(StructureEdge<?, ?> edge) {
...
-                splitGraph(verticesToVisit.iterator().next(), newStructure, verticesToVisit, edgesToVisit);
-                if (newStructure.vertices().size() > 1 || newStructure.vertices().iterator().next().id().isRetrievable()) {
-                    structures.add(newStructure);
-                }
+
+    public List<Structure> splitConnected() {
+        return splitConnected(emptyList());
+    }
+
+    public List<Structure> splitConnected(List<? extends Identifier> forceConnect) {
```
flyingsilverfin, line 117: when we split a structure into subgraphs, we can mandate that some vertices are in the same subgraph, even if they have no direct edge connections

traversal/structure/Structure.java
```diff
@@ -107,24 +110,36 @@ private void recordEdge(StructureEdge<?, ?> edge) {
...
+        Set<StructureEdge<?, ?>> edgesToVisit = new HashSet<>(this.edges);
+        Structure structure;
+        if (!forceConnect.isEmpty()) {
+            structure = new Structure();
+            while (!verticesToConnect.isEmpty()) {
+                StructureVertex<?> vertex = verticesToConnect.iterator().next();
+                splitConnectedTo(vertex, structure, verticesToVisit, edgesToVisit);
+                verticesToConnect.remove(vertex);
             }
+            structures.add(structure);
```
flyingsilverfin, line 130: create one subgraph which contains all the required forced connections

traversal/common/Modifiers.java
```diff
@@ -0,0 +1,144 @@
...
+            return new Sorting(variables, ascending);
+        }
+
+        public List<Identifier.Variable.Retrievable> variables() {
+            return variables;
+        }
+
+        public boolean isAscending(Identifier id) {
+            if (!id.isRetrievable() || !ascending.containsKey(id.asVariable().asRetrievable())) return true;
+            return ascending.get(id.asVariable().asRetrievable());
```
flyingsilverfin, line 128: note: we're encoding a key assumtion here that anything that is not listed or not explicitly descending is ascending!

traversal/planner/MultiPlanner.java
```diff
@@ -78,9 +81,10 @@ private void mayOptimise(GraphManager graphMgr, boolean singleUse) {
     private void createProcedure() {
         // create procedure based on optimal ordering of all the planners
         // rather than solving the optimisation, we estimate the most expensive traversals are the largest ones
-        Comparator<ConnectedPlanner> comparator = Comparator.<ConnectedPlanner, Integer>comparing(planner ->
-                planner.isVertex() ? 1 : planner.asGraph().vertices().size()
-        ).reversed();
+        Comparator<ConnectedPlanner> comparator = Comparator.<ConnectedPlanner, Integer>comparing(planner -> {
+            if (iterate(modifiers.sorting().variables()).anyMatch(planner.vertices()::contains)) return -1;
```
flyingsilverfin, line 85: the sub-structure that contains the sort vertices must come first

test/integration/traversal/TraversalTest.java
```diff
@@ -76,6 +86,143 @@ public void teardown() {
         databaseMgr.close();
     }
 
+    @Test
+    public void native_sorting_longs() {
```
dmitrii-ubskii, line 90: What part of these actually tests that the sorting is performed "natively"? Or do you just test that native sorting didn't break anything? In that case I'd remove `native` from the test names.

dmitrii-ubskii (changes requested):
Some naming questions mostly. Looks great!

traversal/TraversalCache.java
```diff
@@ -18,35 +18,39 @@
...
 import com.vaticle.typedb.core.traversal.structure.Structure;
 
 import java.util.function.Function;
 
 public class TraversalCache {
 
-    private final CommonCache<Structure, Planner> activePlanners;
-    private final CommonCache<Structure, Planner> optimalPlanners;
+    private final CommonCache<Pair<Structure, Modifiers>, Planner> activePlanners;
+    private final CommonCache<Pair<Structure, Modifiers>, Planner> optimalPlanners;
```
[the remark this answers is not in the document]
flyingsilverfin: That is true... not sure I can be bothered to extract the difference! We'll just end up with a couple more plans

traversal/common/Modifiers.java
```diff
@@ -0,0 +1,146 @@
...
+            return Objects.hash(variables);
+        }
+    }
+
+    public static class Sorting {
+
+        public static Sorting EMPTY = new Sorting(list(), map());
+
+        List<Identifier.Variable.Retrievable> variables;
+        Map<Identifier.Variable.Retrievable, Boolean> ascending;
```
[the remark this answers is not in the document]
flyingsilverfin: Because the `ASC/DESC` is a part of `SortedIterator`, and it feels odd domain-wise to have a Modifier depending on an `SortedIterator.Order` order...
but you're right, the boolean isn't correct. THUS, I started moving the `ASC/DESC` out of Sorted iterator. HOWEVER this leads to a massive change that I will submit in a separate PR :P

traversal/procedure/GraphProcedure.java
```diff
@@ -109,50 +112,42 @@ public int vertexCount() {
...
+        if (initialVertex().id().isRetrievable() && modifiers.filter().variables().contains(initialVertex().id().asVariable().asRetrievable())) {
+            return async(initialVertex().iterator(graphMgr, params, modifiers.sorting().isAscending(initialVertex().id()) ? ASC : DESC)
+                    // TODO we can reduce the size of the distinct() set if the traversal engine doesn't overgenerate as much
+                    .map(v -> new GraphIterator(graphMgr, v, this, params, modifiers).distinct()), parallelisation);
         } else {
-            // TODO we can reduce the size of the distinct() set if the traversal engine doesn't overgenerate as much
-            return async(initialVertex().iterator(graphMgr, params).map(v ->
-                    new GraphIterator(graphMgr, v, this, params, filter)
-            ), parallelisation).distinct();
+            return async(initialVertex().iterator(graphMgr, params, modifiers.sorting().isAscending(initialVertex().id()) ? ASC : DESC)
```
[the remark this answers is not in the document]
flyingsilverfin: will be solved by making the `Sorting` hold ASC/DESC in the next PR

traversal/structure/Structure.java
```diff
@@ -107,24 +109,36 @@ private void recordEdge(StructureEdge<?, ?> edge) {
...
-                    structures.add(newStructure);
-                }
+
+    public List<Structure> splitConnected() {
+        return splitConnected(emptyList());
+    }
+
+    public List<Structure> splitConnected(List<? extends Identifier> forceConnect) {
+        List<Structure> structures = new ArrayList<>();
+        Set<? extends StructureVertex<?>> verticesToConnect = iterate(forceConnect).map(vertices::get).toSet();
```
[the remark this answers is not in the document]
flyingsilverfin: updated!

concept/answer/ConceptMap.java
```diff
@@ -177,6 +150,122 @@ public int hashCode() {
...
+                            if (att1.isString()) {
+                                return att1.asString().getValue().compareToIgnoreCase(att2.asString().getValue());
+                            } else if (att1.isBoolean()) {
+                                return att1.asBoolean().getValue().compareTo(att2.asBoolean().getValue());
+                            } else if (att1.isLong() && att2.isLong()) {
+                                return att1.asLong().getValue().compareTo(att2.asLong().getValue());
+                            } else if (att1.isDouble() || att2.isDouble()) {
+                                Double double1 = att1.isLong() ? att1.asLong().getValue() : att1.asDouble().getValue();
+                                Double double2 = att2.isLong() ? att2.asLong().getValue() : att2.asDouble().getValue();
+                                return double1.compareTo(double2);
```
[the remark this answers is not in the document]
flyingsilverfin: to solve this, I added a pre-validation step that ensures that all the possible types produced by the answer iterator for each variable that is comparable, is actually comparable

flyingsilverfin:
Merging with approval to move forward from Haikal

PR #6642, opened 2022-09-13
Bump dependencies

lukas-slezevicius:
## What is the goal of this PR?

Bump dependencies which bump the version of rules_docker. Changes in rules_docker API has allowed us to improve the way Docker-related Bazel targets are written.

## What are the changes implemented in this PR?

* BUILD file uses `container_push` rules instead of `docker_push`
* The `container_bundle` targets are not needed with the new `container_push`
* There are 2 Docker-pushing targets now, 1 for latest tag, 1 for versioned tag.
* Automation `deploy-docker` job now calls the 2 Docker pushing targets
* `workspace_status.sh` is not needed anymore since `container_push` can now read versions from files
* `.bazelrc` was updated to not use the workspace status
* `.bazelrc` build step has a new flag to fix a MacOS C++ toolchain issue

PR #6644, opened 2022-09-19
Fix tar.gz assembly having incorrect folder structure

lukas-slezevicius:

Fix the issues in tar.gz assembly that were caused by #6642.

PR #6642 has changed the upstream repository of `rules_docker` to be from `bazelbuild`. Previously, the upstream repository was the Vaticle's fork of `rules_docker`. That fork contained a patch that fixed a bug in `rules_pkg` that stopped `pkg_tar` rule from working with long paths (the bug caused targz assembly of TypeDB to be incorrect) . Incidentally, an identical patch was applied in Vaticle's `bazel-distribution`. By loading, `rules_docker` from `vaticle_dependencies` before `bazel-distribution`, we ended up using the unpatched `rules_pkg` in the assembly targets that were depending on `rules_pkg` (we can't have multiple versions of a workspace dependency at the same time). This caused the targz assembly of TypeDB to contain invalid paths.

The solution in this PR simply swaps the order of loading, so that `typedb` ends up depending on the patched `rules_pkg` that is provided by `bazel-distribution`.

The ideal solution for this problem would be to bump the `rules_pkg` dependency in `bazel-distribution`, however, the newer versions of `rules_pkg` changed their subpackage visibility modifiers. More precisely, in the newer versions, `pkg_tar` depends on private subpackages. Unfortunately, `stardoc`, the tool that we use to generate documentation, does not work with rules that depend on rules that are invisible to our project (https://github.com/bazelbuild/stardoc/issues/93, https://github.com/bazelbuild/stardoc/issues/128). Therefore, in order to continue using `stardoc`, we must continue using the older version of `rules_pkg`.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -89,6 +89,18 @@ rules_pkg()
 load("@rules_pkg//:deps.bzl", "rules_pkg_dependencies")
 rules_pkg_dependencies()
 
+# Load //github
+load("@vaticle_bazel_distribution//github:deps.bzl", github_deps = "deps")
+github_deps()
+
+# Load //pip
+load("@vaticle_bazel_distribution//pip:deps.bzl", pip_deps = "deps")
+pip_deps()
+
+######################################
+# Load @vaticle_dependencies//distribution docker #
+######################################
+
 # Load //distribution/docker
 load("@vaticle_dependencies//distribution/docker:deps.bzl", docker_deps = "deps")
 docker_deps()
@@ -106,14 +118,6 @@ container_pull(
   repository = "vaticle/ubuntu",
   tag = "4ee548cea883c716055566847c4736a7ef791c38"
 )
-
-# Load //github
-load("@vaticle_bazel_distribution//github:deps.bzl", github_deps = "deps")
-github_deps()
-
-# Load //pip
-load("@vaticle_bazel_distribution//pip:deps.bzl", pip_deps = "deps")
-pip_deps()
 
 #####################################
 # Load @vaticle/typedb dependencies #
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -76,6 +76,19 @@ sonarcloud_dependencies()
 load("@vaticle_dependencies//tool/unuseddeps:deps.bzl", unuseddeps_deps = "deps")
 unuseddeps_deps()
 
+######################################
+# Load @vaticle_bazel_distribution #
+######################################
+
+load("@vaticle_dependencies//distribution:deps.bzl", "vaticle_bazel_distribution")
+vaticle_bazel_distribution()
+
+# Load //common
+load("@vaticle_bazel_distribution//common:deps.bzl", "rules_pkg")
+rules_pkg()
+load("@rules_pkg//:deps.bzl", "rules_pkg_dependencies")
+rules_pkg_dependencies()
+
 # Load //distribution/docker
 load("@vaticle_dependencies//distribution/docker:deps.bzl", docker_deps = "deps")
 docker_deps()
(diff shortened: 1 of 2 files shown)


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

lukas-slezevicius:
Seems like similar problems were encountered before at Vaticle.
https://github.com/vaticle/bazel-distribution/issues/251
https://github.com/vaticle/bazel-distribution/commit/732c5b0359e32543e266e7a7c74f73515c03d9f7

Would patching the latest version of `rules_pkg` to change subpackage visibilities allow us to implement the ideal solution outlined in the description?

lukas-slezevicius:
Am I correct in believing that this CI job proves that the produced docker images are functional? https://grabl.io/lukas-slezevicius/typedb/adf3e9ed0929b9f7e11c564e071f1cfa11666852/build/1/correctness/1/test-assembly-docker/1

lukas-slezevicius:
Deployments of artifact snapshots are also functional: https://grabl.io/lukas-slezevicius/typedb/ccd1392798d378b4826d0ee540fbab670089adb9/build/1/correctness/1/deploy-artifact-snapshot/1.

alexjpwalker:
> Am I correct in believing that this CI job proves that the produced docker images are functional? https://grabl.io/lukas-slezevicius/typedb/adf3e9ed0929b9f7e11c564e071f1cfa11666852/build/1/correctness/1/test-assembly-docker/1

Yes, I believe it means exactly that.

alexjpwalker:
> Seems like similar problems were encountered before at Vaticle. [vaticle/bazel-distribution#251](https://github.com/vaticle/bazel-distribution/issues/251) [vaticle/bazel-distribution@732c5b0](https://github.com/vaticle/bazel-distribution/commit/732c5b0359e32543e266e7a7c74f73515c03d9f7)
> 
> Would patching the latest version of `rules_pkg` to change subpackage visibilities allow us to implement the ideal solution outlined in the description?

I was thinking the same thing when reading through the commit :)

Not an ideal solution (as Adam mentioned in his issue, the _ideal_ solution is for the upstream maintainers to fix the issue) but should be a fairly quick one for us.

However, because it's not an ideal solution, I would still hold back on implementing it for now. But I've added a comment to https://github.com/vaticle/bazel-distribution/issues/251 .

alexjpwalker:
> Deployments of artifact snapshots are also functional: https://grabl.io/lukas-slezevicius/typedb/ccd1392798d378b4826d0ee540fbab670089adb9/build/1/correctness/1/deploy-artifact-snapshot/1.

Try downloading the snapshot that was deployed by this CI job and verify that the folder structure is correct.

lukas-slezevicius:
> > Deployments of artifact snapshots are also functional: https://grabl.io/lukas-slezevicius/typedb/ccd1392798d378b4826d0ee540fbab670089adb9/build/1/correctness/1/deploy-artifact-snapshot/1.
> 
> Try downloading the snapshot that was deployed by this CI job and verify that the folder structure is correct.

The folder structures seem correct. Screenshot 2022-09-20 at 10 25 33


Also, I believe Haikal uses (or used) the produced snapshots in his fork.

WORKSPACE
```diff
@@ -115,6 +97,28 @@ github_deps()
 load("@vaticle_bazel_distribution//pip:deps.bzl", pip_deps = "deps")
 pip_deps()
 
+######################################
+# Load @vaticle_dependencies//distribution/docker #
+######################################
+
```
alexjpwalker, line 103: I think this technical debt is severe enough to warrant an issue, and a code comment.
```suggestion

# must be loaded after `vaticle_bazel_distribution` to ensure
# `rules_pkg` is correctly patched (bazel-distribution #issue_no)

```
And raise an issue in `bazel-distribution`, something like "The `rules_pkg` patch to allow long filenames is fragile".

The issue description should say that the patch needs to be applied before repos such as `rules_docker` are loaded in (that themselves depend on `rules_pkg`), and we should reference https://github.com/vaticle/bazel-distribution/issues/251 in the issue.

alexjpwalker (changes requested):

WORKSPACE
lukas-slezevicius: Created the issue: https://github.com/vaticle/bazel-distribution/issues/355.

alexjpwalker (approved):

haikalpribadi (approved):

PR #6646, opened 2022-09-20
Simplify and fix traversal forward and back-tracking

flyingsilverfin:

We identify an issue where the forward and back-tracking will lead to an incorrect state. To fix this, we simplify the traversal's progression through the query plan, and only optimise the back-tracking mechanism.

* traversal forward progression simplified to progress to the next vertex in the plan
* follow query plan traversal edges only for backtracking, rather than the forward-stepping
* backtracking in the traversal can skip backward (short circuit) whenever a query vertex failed to find a single answer

* reduce the permutation space of the traversal plan permutator

### Problematic case

Imagine the following query plan:

```
    0
 /     \
1        2
|        |
4        3
```
Let's say the first answer was already found, so every vertex above has a valid solution. 
1. We try to find the next answer, by bumping `4` to the next answer, and it has nothing. 
2. Backtracking will proceed to `3`, because that is another leaf in the `revisit` set, and we add `1` to the `revisit` set as well.
3. Let's say `3` and `2` fail to find an answer, adding `0` to the `revisit` set. 
4. Next, we revisit `1`, which was recorded before, and it finds the next vertex.
5. `2` and `3` will find the same vertices they found for the first answer
6. `4` fails to find a vertex! Here, backtracking would have skipped to `1`. However, this breaks our invariants that everything after the currently explored vertex is cleared/empty.

In the new solution, skipping backward from `4` to `1` will clear everything between them, restoring our invariant. The forward step will then re-visit `2` and `3` by proceeding through the ordered vertices in the query plan.

### DIFF
--- a/test/behaviour/typeql/TypeQLSteps.java
+++ b/test/behaviour/typeql/TypeQLSteps.java
@@ -182,7 +182,7 @@ public class TypeQLSteps {
         for (Conjunction conjunction : disjunction.conjunctions()) {
             GraphTraversal.Thing traversal = conjunction.traversal(filter);
             // limited permutation space to avoid timeouts
-            FunctionalIterator<GraphProcedure> procedurePermutations = ProcedurePermutator.generate(traversal.structure()).limit(20000);
+            FunctionalIterator<GraphProcedure> procedurePermutations = ProcedurePermutator.generate(traversal.structure()).limit(40320);
             GraphProcedure procedure = procedurePermutations.next();
             Set<VertexMap> answers = procedure.iterator(tx().concepts().graph(),
                     traversal.parameters(), filter).toSet();
--- a/test/behaviour/typeql/TypeQLSteps.java
+++ b/test/behaviour/typeql/TypeQLSteps.java
@@ -181,12 +181,14 @@ public class TypeQLSteps {
                 .toSet();
         for (Conjunction conjunction : disjunction.conjunctions()) {
             GraphTraversal.Thing traversal = conjunction.traversal(filter);
-            // limited permutation space to avoid OOMs and timeouts
-            FunctionalIterator<GraphProcedure> procedurePermutations = ProcedurePermutator.generate(traversal.structure()).limit(5000);
-            Set<VertexMap> answers = procedurePermutations.next().iterator(tx().concepts().graph(),
+            // limited permutation space to avoid timeouts
+            FunctionalIterator<GraphProcedure> procedurePermutations = ProcedurePermutator.generate(traversal.structure()).limit(20000);
+            GraphProcedure procedure = procedurePermutations.next();
+            Set<VertexMap> answers = procedure.iterator(tx().concepts().graph(),
                     traversal.parameters(), filter).toSet();
             for (int i = 0; procedurePermutations.hasNext(); i++) {
-                Set<VertexMap> permutationAnswers = procedurePermutations.next().iterator(tx().concepts().graph(),
+                procedure = procedurePermutations.next();
+                Set<VertexMap> permutationAnswers = procedure.iterator(tx().concepts().graph(),
                         traversal.parameters(), filter).toSet();
                 assertEquals(answers, permutationAnswers);
             }
--- a/traversal/scanner/GraphIterator.java
+++ b/traversal/scanner/GraphIterator.java
@@ -136,8 +136,7 @@ public class GraphIterator extends AbstractFunctionalIterator<VertexMap> {
     }
 
     private void setupImplicitDependency(ProcedureVertex<?, ?> from, ProcedureVertex<?, ?> to) {
-        vertexTraversers.get(from).implicitDependents.add(to);
-        vertexTraversers.get(to).implicitDependees.add(from);
+        vertexTraversers.get(to).addImplicitDependee(from);
     }
 
     @Override
@@ -194,7 +193,7 @@ public class GraphIterator extends AbstractFunctionalIterator<VertexMap> {
     }
 
     private void initialiseStarts() {
-        toTraverse.addAll(procedure.startVertices());
+        toTraverse.add(procedure.vertex(0));
         direction = Direction.TRAVERSE;
     }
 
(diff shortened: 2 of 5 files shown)


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

test/behaviour/typeql/TypeQLSteps.java
```diff
@@ -181,12 +181,14 @@ private void assertQueryPlansCorrect(TypeQLMatch typeQLQuery) {
                 .toSet();
         for (Conjunction conjunction : disjunction.conjunctions()) {
             GraphTraversal.Thing traversal = conjunction.traversal(filter);
-            // limited permutation space to avoid OOMs and timeouts
-            FunctionalIterator<GraphProcedure> procedurePermutations = ProcedurePermutator.generate(traversal.structure()).limit(5000);
-            Set<VertexMap> answers = procedurePermutations.next().iterator(tx().concepts().graph(),
+            // limited permutation space to avoid timeouts
+            FunctionalIterator<GraphProcedure> procedurePermutations = ProcedurePermutator.generate(traversal.structure()).limit(20000);
+            GraphProcedure procedure = procedurePermutations.next();
```
flyingsilverfin, line 186: introduce a procedure variable so we can easily breakpoint and investigate!

traversal/scanner/GraphIterator.java
```diff
@@ -226,27 +225,32 @@ private void traverse(ProcedureVertex<?, ?> procedureVertex) {
...
 
     private void success(VertexTraverser vertexTraverser) {
-        for (ProcedureVertex<?, ?> vertex : vertexTraverser.procedureVertex.dependents()) {
-            toTraverse.add(vertex);
-            vertexTraversers.get(vertex).clear();
-        }
-        for (ProcedureVertex<?, ?> vertex : vertexTraverser.implicitDependents) {
-            toTraverse.add(vertex);
-            vertexTraversers.get(vertex).clear();
-        }
```
flyingsilverfin, line 236: essentially, as explained in the PR description, we cannot skip forward just along the procedure edges because that can lead to inconsistent states. We must always proceed to the next vertex in the plan.
However, we can do some short-circuit backtracking when failing to find a vertex

traversal/scanner/GraphIterator.java
```diff
@@ -226,27 +225,32 @@ private void traverse(ProcedureVertex<?, ?> procedureVertex) {
...
         toRevisit.addAll(vertexTraverser.procedureVertex.dependees());
         toRevisit.addAll(vertexTraverser.implicitDependees);
+        if (!vertexTraverser.anyAnswerFound() && !vertexTraverser.procedureVertex.ins().isEmpty()) {
+            // short circuit everything not required to be re-visited
+            for (int i = vertexTraverser.lastDependee.order() + 1; i < vertexTraverser.procedureVertex.order(); i++) {
+                ProcedureVertex<?, ?> skip = procedure.vertex(i);
+                toRevisit.remove(skip);
+                vertexTraversers.get(skip).clear();
+            }
+        }
```
flyingsilverfin, line 244: when we can short-circuit backtrack to the dependees (eg. when we failed to find a single answer for the traverser), we need to clear all the intermediate vertices that we are skipping past.

traversal/test/ProcedurePermutator.java
```diff
@@ -20,26 +20,36 @@
...
-                    }
-                    return GraphProcedure.create(structure, orderingMap);
-                });
+        // to reduce the permutation space, we always will put the type vertices at the start
+        Set<Identifier> retrievables = new HashSet<>();
+        List<Identifier.Variable.Label> labels = new ArrayList<>();
+        for (StructureVertex<?> vertex : structure.vertices()) {
+            if (vertex.id().isLabel()) labels.add(vertex.id().asVariable().asLabel());
+            else retrievables.add(vertex.id());
+        }
```
flyingsilverfin, line 46: because this permutation spaces grows factorially, reducing the permutation size is crucial. One thing we do here is to just lump all the constant labelled vertices at the start of the plan, since they're always independent and have no real impact on the traversals

traversal/scanner/GraphIterator.java
```diff
@@ -284,6 +295,10 @@ private boolean findNextVertex() {
             return vertex;
         }
 
+        private boolean anyAnswerFound() {
+            return anyAnswerFound;
+        }
```
flyingsilverfin, line 300: used for short-circuiting backtracks

traversal/scanner/GraphIterator.java
```diff
@@ -194,7 +193,7 @@ private void setCompleted() {
     }
 
     private void initialiseStarts() {
-        toTraverse.addAll(procedure.startVertices());
+        toTraverse.add(procedure.vertex(0));
```
flyingsilverfin, line 196: since we always proceed to the next ordered vertex, we can just start at the 0th vertex now

test/behaviour/typeql/TypeQLSteps.java
dmitrii-ubskii: Why the limit increase?

traversal/scanner/GraphIterator.java
```diff
@@ -226,27 +225,29 @@ private void traverse(ProcedureVertex<?, ?> procedureVertex) {
...
-            vertexTraversers.get(vertex).clear();
-        }
-        for (ProcedureVertex<?, ?> vertex : vertexTraverser.implicitDependents) {
-            toTraverse.add(vertex);
-            vertexTraversers.get(vertex).clear();
-        }
+        if (vertexTraverser.procedureVertex.order() + 1 == procedure.vertexCount()) return;
+        ProcedureVertex<?, ?> next = procedure.vertex(vertexTraverser.procedureVertex.order() + 1);
+        toTraverse.add(next);
+        vertexTraversers.get(next).clear();
```
dmitrii-ubskii, line 231: This could definitely be more readable with some helper functions. `vertexTraverser.procedureVertex.isLast()`, `.nextVertex()` alone would do wonders.

dmitrii-ubskii (changes requested):
Is there a reason why the solution is this and not relabeling the graph like this with e.g. a topological DFS?
```
   0  
 /  \
1     3
|     |
2     4
```
So long as the edge directions don't change, the planner can't tell the difference.

test/behaviour/typeql/TypeQLSteps.java
flyingsilverfin: Mostly because we can! They still run fast for all the behaviour scenarios.

However, now that I'm thinking of it, we should choose a factorial, like 8! = 40320 to ensure we can get all query plans for queries with that many traversal vertices?

traversal/scanner/GraphIterator.java
flyingsilverfin: that is the correct solution, yes!

flyingsilverfin: In the previous version, we shouldn't clear `2` and `3` when `4` fails

traversal/scanner/GraphIterator.java
flyingsilverfin: updated

flyingsilverfin:
The reason we didn't require the Query Planner to enforce the correctness, is that it is more decoupled for the traversal iterator itself to be handle an arbitrary query procedure.

By decoupling them, we don't rely on two components to get a correct query plan (eg. query planner works in tandem with the internals of the traversal execution to guarantee all answers are found correctly), and they can be considered separately.

In the future, I do expect that the query planner will start ordering the vertices considering the choices between BFS and DFS as another part of the optimisation. If the traversal can already handle that it'll be easier.

dmitrii-ubskii:
> The reason we didn't require the Query Planner to enforce the correctness, is that it is more decoupled for the traversal iterator itself to be handle an arbitrary query procedure.

That makes sense.

dmitrii-ubskii (approved):
LGTM

PR #6648, opened 2022-09-22
Refactor internal sort ordering parameters

flyingsilverfin:

To make the `Order` parameter of a sorted iterator more generally applicable, we extract it to become a parameter that lives in the `//common` package.

* Move `//common/iterator/sorted/SortedIterator.Order` into `//common/parameters/Order`
* Propagate all the changes required by the move to all the users of `Order`
* Traversal sorting modifiers directly contain an `Order`, rather than a boolean representing an order
* Producers no longer eagery generate all answers as soon as a sort variable is present. Instead, we use `incremental` only when there is no sort, otherwise we use `exhaustive` producers


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

PR #6650, opened 2022-09-27
Reasoner planner treats IID and value-equality constraints as bounds

krishnangovindraj:
The reasoner planner must treat constraints binding a variable to a specific value (or IID) similar to how it treats the variable being bound to a specific thing.
 
When planning the order of resolvables constituting a conjunction, the current reasoner-planner prioritises a connected resolvable over a disconnected one.
A variable which is bound to an IID (or a specific value) has a similarly strong restricting effect. With this PR, the reasoner planner treats these variables as being bound (and hence, all resolvables featuring them as connected) AFTER it runs out of connected resolvables. (Note: for the top-level query, this is always the case).  

- Reasoner planner: When all resolvables left to schedule are disconnected, the boundVariables are extended to include the variables bound to an IID or a specific value.

### DIFF
--- a/reasoner/common/Planner.java
+++ b/reasoner/common/Planner.java
@@ -116,7 +116,6 @@ public class Planner {
                     }
                 }
 
-
                 // Retrievable where:
                 // it can be disconnected
                 // all of it's dependencies are already satisfied (should be moot),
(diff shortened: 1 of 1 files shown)


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

flyingsilverfin (approved):
The boolean to avoid looping over the bounds extension multiple times is a code smell in that we could re-architect the while loop somehow! But since this code will be deleted anyway, let's just merge :)

PR #6656, opened 2022-10-05
Reasoner Planning: plan resolvable orderings

krishnangovindraj:
Introduce a cost-metric for resolvable-orderings, and a planner to choose the resolvable ordering for each conjunction-callmode (a conjunction & the set of bound variables)
Since the plans of each callmode in a cycle depends on the others,  these are optimised together. More precisely, a combined-optimisation is done for each strongly-connected component.

* Introduce `RecursivePorPlanner`
  * Recursively computes the optimal resolvable ordering for a conjunction
  * Plans a callmode assuming it will be called for all possible bindings
  * Uses a (scaling-factor * unbound-cost) to estimate ordering effects
* Move cyclic-dependency detection into `ConjunctionSummary` 
* `AnswerCountEstimator` sheds it's local assumptions, introducing bound-propagation - when a new resolvable is added to the prefix, and its estimate for a variable v is lower than the current lowest estimate for v, all other resolvables are scaled down so their estimates for v match this new lowest estimate.

### The reasoner-planner problem
The high-level problem is to find the best order to evaluate resolvables in each conjunction (the user query and triggered rules). We cannot just call the traversal-planner on each conjunction because reasoning works differently to traversal. 

1. **Buffered answers:** Unlike during traversal, when an answer to a 'resolvable' is produced, it is buffered in memory and returned the next time the resolvable is evaluated - (In the traversal where it would be re-traversed). This makes the computational cost of a reasoning problem different from that of a traversal.
2. The unit of computation is a 'call' - A `Conjunction` and input `ConceptMap`. Each 'call' spawns a processor which buffers answer ConceptMaps involving the remaining (output) variables in the call. 
3. **Cycles**: Rules can depend on themselves - walking the rule-graph and estimating on the go won't work. 
4. **Interdependent plans**: The order of evaluating a conjunction affects the ideal plan of any rules it invokes. This in turn influencing the total cost of evaluating the conjunction.
5. The unit of planning is a mutually recursive subgraph. Each node in the subgraph is identified by a 'callmode' - The conjunction and the set of variables which will be passed as input bounds to the conjunction. 

### The global planning problem
The true planning problem is a global optimisation where we must choose the resolvable ordering for the user-query and every rule (recursively) triggered, such that the total cost of evaluation is minimized. 
Each global-plan can be represented by the call-graph, where each-node is a CallMode and there is an edge from a call-mode A to another CallMode B, if the ordering for A triggers B. The weight of this edge is the set of calls generated  A to B.

Consider the directed-graph:
```
$a isa node, has name "a"; 
$b isa node, has name "b";
$c isa node, has name "c";

(from: $a, to: $b) isa edge;
(from: $b, to: $c) isa edge;

rule A-path-is-transitive-edge-base:
when { (from: $a, to: $b) isa edge; }
then {  (from: $a, to: $b) isa path; };

rule B-path-is-transitive-edge-recursive:
when { (from: $a, to: $b) isa path; (from: $b, to: $c) isa edge; }
then {  (from: $a, to: $c) isa path; };
```
image

For the following query:
`$start isa node, has name "a"; (from: $start, to: $end) isa path;`

A good plan would be:
image

A bad plan would be: 
image
 
The cost of the global plan is just the total cost of evaluating each CallMode node across all calls on the incoming edges.

#### Cost of evaluating a conjunction
It remains to define the "cost of evaluating a callmode" used above.
The cost of recursive computation is captured through the nodes corresponding to the triggered call-modes. The only remaining which needs to be accounted is the cost of retrieving the answers to a resolvable (+ the cost of processing the inferred answers). 

#### Naive approach:
The naive way of solving this is to back-track over the choice of orderings for each conjunction. 
The way of generating one set of choices is:

1. Initially, the unplanned-set = { user-query }
2. Until the unplanned-set is empty:
  * Choose an ordering for the unplanned-conjunction. 
  * determine the call-mode of each triggered rule and estimate the number of calls to each of these call-mode.
  * Add any unplanned call-mode to the unplanned-set

Enumerating and evaluating each plan is clearly intractable. We now describe how we decompose the global-problem into smaller problems over subgraphs and the assumptions made while doing so.

#### All-bounds costs & scaling-factors
We can't keep track of the actual set of incoming calls. Instead, we use an estimate of the number of calls as the weight of the edge. It was also easier to reason about the cost of finding ALL answers to a conjunction rather than a single-answer. So, we define the cost-contribution of a single call-mode as:
(#incoming-calls/allPossibleAnswers) * cost-of-finding-all-answers

#### Decomposing into subgraphs
In the naive setting, the choice of plan for a conjunction is dependent on its ancestors through the number of calls coming from the ancestors. To decouple this, we assume __"the optimal plan for a conjunction is the one with the lowest cost when evaluated for a single-call coming from its ancestors"__
It doesn't always hold, but it seems reasonable given most queries specify concepts .

While choosing the plan for a conjunction, we replace all incoming edges from ancestors with a 'dummy' edge of weight 1. This decomposes the graph into disconnected subgraphs of mutually-recursive conjunctions - our unit of planning. We must plan all nodes in the subgraph together, as they all affect each other through the number of calls in the cyclic-edges. 

It still remains to choose the plan of the ancestors so that we minimise the work done in the descendants. 

#### Incorporating planned descendants
We now have the plan for each descendant CallMode, we have not yet described how the cost of a descendant callee is integrated into a caller.  We simply add it to the cost of evaluating the resolvable triggering it - treating it as an expensive retrieval. Since we have to account for the actual number of calls to the descendant being generated, the term being added is more precisely
#calls-generated/#answers-to-resolvable * cost-of-finding-all-answers-to-conjunction

### Concrete definitions: The cost of a given global plan
A global plan for a query defines the ordering of each CallMode triggered when evaluating the query. Whenever a CallMode is specified, we implicitly mean the plan of the CallMode.

The total cost of a given global plan for a user-query $Q$ is the PlanCost of the CallMode for the conjunction $Q$ in the mode $\{\}$

$$ TotalCost(Q) = PlanCost( <Q,\{\}> ) $$

For a given CallMode, the PlanCost of the is the cost of the mutually-recursive subgraph that the CallMode is a part of:

$$ PlanCost(CallMode) = SubgraphCost( mutuallyRecursiveSubgraph(CallMode) ) $$

The cost of a subgraph plan (not accounting for entries):

$$ SubgraphCost(G) = \sum_{CallMode \in G} CyclicScalingFactorSum(G, CallMode) * ScaledCost(CallMode) + UnscaledCost(CallMode) $$

The ScaledCost and UnscaledCost are the cost of evaluating a certain resolvable-ordering. We make the difference to support multiple-starting points - Work done by resolvables disconnected from the input-variables should not be scaled down with a reduced-number of entry-calls. Thus, whether a resolvable contributes to the ScaledCost or UnscaledCost depends on whether there is a path from the input-mode variables to the resolvable through the preceding resolvables, captured here by the indicator $I[connected(Order,r)]$

$$ ScaledCost(Order) = \sum_{r \in Order} I[connected(Order,r)] *  ScalingFactor(Order, r) * resolvableCost(r) $$ 
$$ UnscaledCost(Order) = \sum_{r \in Order} (1-I[connected(Order,r]) * ScalingFactor(Order, r) * resolvableCost(r) $$ 

The ResolvableCost of a resolvable r is:

$$ ResolvableCost(r) = LocalCost(r) + AcyclicCallCost(r) $$

$$ LocalCost(r) = AnswerEstimate(r) $$

$$ AcyclicCallCost(r) = \sum_{CallMode \in acyclicCalls(r)} PlanCost(CallMode) $$

And the scaling factor of a resolvable r in a specified Order

$$ ScalingFactor(Order, r) = \frac{AnswerEstimate(pre(Order,r), mode(Order,r))}{AnswerEstimate({r}, mode(Order,r))} $$ 

Where $mode(Order, r)$ is the set of variables which would be input to r when evaluated according to Order.

The CyclicScalingFactorSum of a Callmode in a subgraph is just the sum of CyclicScalingFactor's associated with calls to CallMode in orders in G. A CyclicScalingFactor for a cyclic-call triggered by the resolvable $r \in Order$ is: 
* 0 if all variables to the recursive call are input-variables to this call.
* Otherwise 

$$ \frac{AnswerEstimate(pre(Order,r), mode(Order,r) \setminus mode(CallMode) )}{AnswerEstimate(\{r\}, mode(Order,r))} $$

### Finding the optimal cost plan

When planning a given CallMode, we:
1. Recursively generate orderings for all reachable CallModes.
    * Recursively plan any Descendant subgraphs thus encounter.
2. Summarize any generated ordering. A summary consists of information useful for the subgraph-planning:
    * The ScaledCost &UnscaledCost
    * CyclicScalingFactor for any calls to other nodes in the subgraph
    * The estimate for the number of answers the CallMode generates (*the conjunction generates, since its independent of mode)
3. Start a back-tracking search over choices of orderings to generate all SubgraphPlans and pick the best one (in an interleaved manner)  
    * The definition of the 'best' subgraph plan is (This estimates the cost of evaluating the plan for a single entry-call):

$$ G* = argmin_G \left ( SubgraphCost(G) + \frac{1}{|entryCalls|} * ScaledCost(entryCallMode) \right ) $$ 
    
### Assumptions & Known inaccuracies
* **Assumption:** The optimal plan for a given callmode (for any number of calls) is the same as the optimal plan for callmode for a single call . **Danger:** We may pick a plan which is good for small number of bounds but far more expensive than one which is optimised for a large number of calls.
* **Assumption:** We add descendant subgraph plan costs to the resolvable cost . **Inaccuracy:** We double count these costs (e.g., the call graph is a diamond)
* **Assumption**: Cyclic-scaling-factor is 0 if the resolvableMode consists only of input variables - This is only true if the call is to the conjunction itself.


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

reasoner/planner/ConjunctionSummarizer.java
```diff
@@ -0,0 +1,232 @@
...
+        iterate(stack.visited()).forEachRemaining(conjunction -> {
+            assert !conjunctionSummaries.containsKey(conjunction);
+            conjunctionSummaries.put(conjunction, createConjunctionSummary(conjunction, cyclicConcludables.getOrDefault(conjunction, new HashSet<>())));
+        });
+    }
+
+    private void findCyclicConcludables(ResolvableConjunction conjunction, ConjunctionConcludableStack stack, Map<ResolvableConjunction, Set<Pair<Concludable, ResolvableConjunction>>> cyclicConcludables) {
+        if (conjunctionSummaries.containsKey(conjunction)) {
+            return; // We only add conjunction summaries at the end of a search -> this does not loop back.
+        }
```
krishnangovindraj, line 93: Optimisation added: We start our search at a root which is not part of a cycle. We summarise the conjunctions in a subgraph AFTER the search. This ensures all cyclic dependencies in the subgraph have been found, and we don't have to recurse down this path from other entry points.

reasoner/planner/GreedyCostSearch.java
```diff
@@ -40,54 +40,67 @@ public GreedyCostSearch(TraversalEngine traversalEng, ConceptManager conceptMgr,
...
+            if (!boundsExtended && (nextResolvableCost == null || nextResolvableCost.second() > 10)) {
+                boundsExtended = true;
+                iterate(remaining).filter(Resolvable::isRetrievable).flatMap(resolvable -> iterate(resolvable.asRetrievable().pattern().variables()))
+                        .flatMap(variable -> iterate(variable.constraints()))
+                        .filter(constraint -> constraint.isThing()).map(constraint -> constraint.asThing())
+                        .filter(thingConstraint -> (thingConstraint.isIID() || (thingConstraint.isValue() && thingConstraint.asValue().isValueIdentity())))
+                        .map(thingConstraint -> thingConstraint.owner())
+                        .forEachRemaining(bounds::add);
+                continue;
+            }
```
krishnangovindraj, line 72: The IID improvement added into development.

reasoner/planner/ReasonerPlanner.java
```diff
@@ -50,34 +50,44 @@ public ReasonerPlanner(TraversalEngine traversalEng, ConceptManager conceptMgr,
...
      */
-    static Map<Resolvable<?>, Set<Identifier.Variable.Retrievable>> dependencies(Set<Resolvable<?>> resolvables) {
-        Map<Resolvable<?>, Set<Identifier.Variable.Retrievable>> deps = new HashMap<>();
-        Set<Identifier.Variable.Retrievable> generated = iterate(resolvables).map(Resolvable::generating).filter(Optional::isPresent)
-                .map(Optional::get).map(ThingVariable::id).toSet();
-
-        Map<Identifier.Variable.Retrievable, Integer> unnegatedRefCount = new HashMap<>();
+    static Map<Resolvable<?>, Set<Variable>> dependencies(Set<Resolvable<?>> resolvables) {
+        // TODO: There may not be any generated->used dependencies since every use either triggers the rule or is unsatisfiable
+        Map<Resolvable<?>, Set<Variable>> deps = new HashMap<>();
```
krishnangovindraj, line 79: This is a todo for later. Relaxing all generated->used dependencies does causes tests to fail

reasoner/planner/ReasonerPlanner.java
```diff
@@ -50,34 +50,44 @@ public ReasonerPlanner(TraversalEngine traversalEng, ConceptManager conceptMgr,
...
      * Determine the resolvables that are dependent upon the generation of each variable
      */
-    static Map<Resolvable<?>, Set<Identifier.Variable.Retrievable>> dependencies(Set<Resolvable<?>> resolvables) {
-        Map<Resolvable<?>, Set<Identifier.Variable.Retrievable>> deps = new HashMap<>();
-        Set<Identifier.Variable.Retrievable> generated = iterate(resolvables).map(Resolvable::generating).filter(Optional::isPresent)
-                .map(Optional::get).map(ThingVariable::id).toSet();
-
-        Map<Identifier.Variable.Retrievable, Integer> unnegatedRefCount = new HashMap<>();
+    static Map<Resolvable<?>, Set<Variable>> dependencies(Set<Resolvable<?>> resolvables) {
+        // TODO: There may not be any generated->used dependencies since every use either triggers the rule or is unsatisfiable
```
krishnangovindraj, line 78: This is a todo for later. Relaxing all generated->used dependencies does causes tests to fail

reasoner/controller/ConjunctionController.java
```diff
@@ -101,7 +101,8 @@ protected void setUpUpstreamControllers() {
 
     abstract FunctionalIterator<Concludable> concludablesTriggeringRules();
 
-    List<Resolvable<?>> plan(Set<Variable.Retrievable> boundVariables) {
+    List<Resolvable<?>> plan(Set<Variable.Retrievable> bounds) {
+        Set<com.vaticle.typedb.core.pattern.variable.Variable> boundVariables = iterate(bounds).map(id -> conjunction.pattern().variable(id)).toSet();
```
krishnangovindraj, line 105: The planner deals exclusively with pattern.Variable instead of Identifier.Variable

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -60,186 +59,72 @@
 public class AnswerCountEstimator {
```
krishnangovindraj, line 59: Stripped out cycle-detection & ConjunctionContext
Rearranged the functions

reasoner/planner/RecursivePorPlanner.java
```diff
@@ -0,0 +1,361 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.Stack;
+
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+
+public class RecursivePorPlanner extends ReasonerPlanner {
```
krishnangovindraj, line 43: Where do I start >.<

krishnangovindraj (commented):
(Comments, except for RecursivePorPlanner)

reasoner/planner/RecursivePorPlanner.java
```diff
@@ -0,0 +1,361 @@
...
+        assert planCache.getIfPresent(new Pair<>(conjunction, inputBounds)) != null;
+        return planCache.getIfPresent(new Pair<>(conjunction, inputBounds));
+    }
+
+    private void computeOrderingsForCycles(CallKey callKey) {
+        // This should be called only from conjunctions not on cycle-paths or at the lasso-knot FOR THE CURRENT LASSO ROPE
+        Map<CallKey, OrderingSummary> chosenSummaries = new HashMap<>();
+        Set<CallKey> pendingKeys = new HashSet<>();
+        pendingKeys.add(callKey);
+        Pair<List<OrderingSummary>, Double> bestPlan = cyclePlanSearch(callKey, pendingKeys, chosenSummaries);
```
krishnangovindraj, line 82: Since the plans of conjunctions in a cycle affect each other, they have to be optimised together.
cyclePlanSearch walks the rule-graph, backtracking over candidate orderings[1] at each level and generating the graph based on concludables in the chosen ordering.

[1] - Some orderings have been pruned - When multiple orderings have the same input-bounds and the same bindings for each cyclic-concludable, we only need to consider the one with the lowest cost.

reasoner/planner/RecursivePorPlanner.java
```diff
@@ -0,0 +1,361 @@
...
+            chosenSummaries.put(key, orderingSummary);
+
+            Set<CallKey> nextPendingKeys = new HashSet<>(pendingKeys);
+            Set<CallKey> triggeredCalls = orderingSummary.cyclicCallsTriggered(conjunctionSummary);
+            iterate(triggeredCalls).filter(call -> !chosenSummaries.containsKey(call)).forEachRemaining(nextPendingKeys::add);
+            Pair<List<OrderingSummary>, Double> newPlan = cyclePlanSearch(root, nextPendingKeys, chosenSummaries);
+            bestPlan = (bestPlan == null || newPlan.second() < bestPlan.second()) ? newPlan : bestPlan;
+
+            chosenSummaries.remove(key);
+        }
```
krishnangovindraj, line 114: Loop over the backtrack-choices.

reasoner/planner/RecursivePorPlanner.java
```diff
@@ -0,0 +1,361 @@
...
+            // So a call from a different location would receive the cost it would expect?
+            planCache.put(new Pair<>(bestPlanForCall.callKey.conjunction, bestPlanForCall.callKey.bounds), new Plan(bestPlanForCall.ordering, bestGlobalCost));
+        }
+    }
+
+    private Pair<List<OrderingSummary>, Double> cyclePlanSearch(CallKey root, Set<CallKey> pendingKeys, Map<CallKey, OrderingSummary> chosenSummaries) { // The value contains the scaling factors needed for the other conjunctions in the globalCost.
+        // Walk the cycle, backtrack over choices of orders.
+        if (pendingKeys.isEmpty()) {
+            return new Pair<>(new ArrayList<>(chosenSummaries.values()), cyclePlanCost(root, chosenSummaries));
+        }
```
krishnangovindraj, line 96: No more choices to make! Evaluate the cost of the plan generated and return.

reasoner/planner/RecursivePorPlanner.java
```diff
@@ -0,0 +1,361 @@
...
+            for (Rule rule : logicMgr.applicableRules(resolvable.asConcludable()).keySet()) {
+                ResolvableConjunction ruleConjunction = rule.condition().conjunction();
+                recursivelySummarizeOrderings(new CallKey(ruleConjunction, resolvableBounds));
+                if (!cyclicDependencies.contains(ruleConjunction)) {
+                    plan(ruleConjunction, resolvableBounds); // Acyclic dependencies can be fully planned
+                }
+            }
+        }
+    }
+
```
krishnangovindraj, line 140: The initialise methods generate candidate orderings for each triggered (conjunction, bounds) and prune the redundant, suboptimal ones.

.grabl/automation.yml
```diff
@@ -214,7 +217,8 @@ release:
       dependencies: [deploy-github]
       command: |
         docker login -u $REPO_DOCKER_USERNAME -p $REPO_DOCKER_PASSWORD
-        bazel run //:deploy-docker
+        bazel run //:deploy-docker-latest
+        bazel run //:deploy-docker-versioned
     deploy-artifact-release:
       image: vaticle-ubuntu-21.04
       filter:
```
krishnangovindraj, line 224: Copy from development

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -60,190 +59,76 @@
...
                                  boolean isCyclic) {
-            this.conjunctionContext = conjunctionContext;
+            this.conjunctionSummary = conjunctionSummary;
             this.variableModels = variableModels;
             this.constraintModels = constraintModels;
             this.isCyclic = isCyclic;
         }
 
         private long estimateAnswers(Set<Variable> variables, Set<Resolvable<?>> resolvables) {
+            if (resolvables.size() ==0 ) return 1; // TODO: Is this right, or should we use the type-upper-bound for the unconstrained variables?
```
flyingsilverfin, line 121: make sure you run auto-format and import-cleanup on the changed classes as a habit

flyingsilverfin: why is it possible to have an empty set of resolvables at all? Same as having an empty set of variables?

reasoner/planner/ConjunctionSummarizer.java
```diff
@@ -0,0 +1,232 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+
+public class ConjunctionSummarizer {
```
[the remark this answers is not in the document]
flyingsilverfin: note: use British spelling, `Summariser`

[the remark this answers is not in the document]
flyingsilverfin: note: apply throughout

logic/resolvable/ResolvableConjunction.java
```diff
@@ -62,4 +62,9 @@ public FunctionalIterator<Concludable> allConcludables() {
                 iterate(negations()).flatMap(negated -> iterate(negated.disjunction().conjunctions()))
                         .flatMap(conj -> conj.allConcludables()));
     }
+
+    @Override
+    public String toString() {
+        return "ResolvableConjunction{" +  conjunctionPattern + '}';
```
flyingsilverfin, line 68: note: we try to avoid the default `toString` - you can get rid of the `ResolvableConjunction{` part

common/util/PerfCounter.java
```diff
@@ -0,0 +1,42 @@
...
+    public abstract long get(Object identifier);
+
+    public abstract void add(Object identifier, long delta);
+
+    private static class PerfCounterImpl extends PerfCounter {
+
+        private final Map<Object, AtomicLong> counters;
+
+        public PerfCounterImpl() {
+            counters = new HashMap<>();
```
flyingsilverfin, line 22: this needs to be a `ConcurrentHashMap`, and then you can use normal non-atomic longs since it's protected by the concurrent hash map protection

reasoner/planner/PartialOrderReductionTest.java
```diff
@@ -0,0 +1,231 @@
...
+
+    private Set<Variable> getVariablesByName(Set<Resolvable<?>> resolvables, Set<String> names) {
+        Set<Variable> vars = iterate(resolvables).flatMap(r -> iterate(r.variables()))
+                .filter(v -> v.id().isName() && names.contains(v.id().asName().name()))
+                .toSet();
+        assert vars.size() == names.size();
+        return vars;
+    }
+
+    private void debugPrint(List<Resolvable<?>> resolvables, List<String> names, List<List<Resolvable<?>>> allOrderings) {
```
flyingsilverfin, line 219: you should use a Logger instance, and call it something more specific like `logOrderings` rather than just "debugPrint" -- what is being printed? The printing kinda implies use for debugging as well, so no point having that in the name

common/util/PerfCounter.java
```diff
@@ -0,0 +1,42 @@
+package com.vaticle.typedb.core.common.util;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
+
+public abstract class PerfCounter {
```
flyingsilverfin, line 7: why not an interface?

reasoner/planner/ReasonerPlanner.java
```diff
@@ -101,6 +138,51 @@ static Map<Resolvable<?>, Set<Identifier.Variable.Retrievable>> dependencies(Set
...
+                Set<Variable> ruleSideBounds = iterate(ruleSideIds)
+                        .filter(id -> ruleConjunction.pattern().retrieves().contains(id)) // avoids constant has
+                        .map(id -> ruleConjunction.pattern().variable(id)).toSet();
+                calls.add(new ReasonerPlanner.CallKey(ruleConjunction, ruleSideBounds));
+            }
+        }
+        return calls;
+    }
+
+    static class CallKey {
```
flyingsilverfin, line 160: hadn't we agreed to rename this to something more intuitive?

reasoner/planner/RecursivePorPlanner.java
```diff
@@ -0,0 +1,343 @@
...
+public class RecursivePorPlanner extends ReasonerPlanner {
+    // Inaccuracies:
+    //      retrieval costs are treated the same as reasoning-overhead cost (Both approximated as the number of answers retrieved for all variables)
+    //      Excess calls are not penalized, since the scaling factor is capped at one (Solve using (calls + scaling-factor * rec-cost?) )
+    //      Acyclic dependencies are counted once for each occurence
+    //          (Can be fixed by including the optimal plans for the bound in the globalPlan chosenSummaries - The scaling factors will be capped at 1 then)
+    // So far, these are acceptable because the difference in cost is some constant factor, and our aim is to avoid horrible plans.
+    //
+    //      !!! The cost does not depend on the binding mode !!! because of the formulation - Handle this with connectedness restriction when generating orders?
```
flyingsilverfin, line 52: make sure these comments are up to date when we start changing the impls

reasoner/planner/RecursivePorPlanner.java
```diff
@@ -0,0 +1,343 @@
...
+        this.candidateOrderings = new HashMap<>();
+    }
+
+    private static Set<Variable> consideredVariables(Resolvable<?> resolvable) {
+        return iterate(resolvable.variables()).filter(Variable::isThing).toSet();
+    }
+
+    @Override
+    Plan computePlan(CallKey callKey) {
+        recursivelySummarizeOrderings(callKey);
```
flyingsilverfin, line 71: use UK spelling, `s` not `z`

reasoner/planner/RecursivePorPlanner.java
```diff
@@ -0,0 +1,343 @@
...
+        computeOrderingsForCycles(callKey);
+        assert planCache.getIfPresent(callKey) != null;
+        return planCache.getIfPresent(callKey);
+    }
+
+    // plan-search: combined plan-search for all nodes in a cycle
+    private void computeOrderingsForCycles(CallKey callKey) {
+        // This should be called only from conjunctions not on cycle-paths or at the lasso-knot FOR THE CURRENT LASSO ROPE
+        Map<CallKey, OrderingSummary> chosenSummaries = new HashMap<>();
```
flyingsilverfin, line 80: I thought we had a better name for `OrderingSummary`?

reasoner/controller/ControllerRegistry.java
```diff
@@ -161,13 +162,19 @@ public void createRootConjunction(ResolvableConjunction conjunction, Set<Variabl
...
                                       boolean explain, ReasonerConsumer<ConceptMap> reasonerConsumer) {
+        disjunction.conjunctions().forEach(conjunction -> planner().plan(conjunction, new HashSet<>()));
         Function<Driver<RootDisjunctionController>, RootDisjunctionController> actorFn =
                 driver -> new RootDisjunctionController(driver, disjunction, filter, explain, controllerContext, reasonerConsumer);
         LOG.debug("Create Root Disjunction for: '{}'", disjunction);
         createRootController(reasonerConsumer, actorFn);
     }
 
     public void createExplainableRoot(Concludable concludable, ConceptMap bounds, ReasonerConsumer<Explanation> reasonerConsumer) {
+        Set<com.vaticle.typedb.core.pattern.variable.Variable> boundVariables = bounds.concepts().keySet().stream()
```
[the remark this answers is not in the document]
krishnangovindraj: Okay. Only left in GreedyCostSearch where I can't find an alternative to Stream.min

common/util/PerfCounter.java
krishnangovindraj: Why would you use a ConcurrentHashMap when you can use atomic operations?

reasoner/controller/ControllerRegistry.java
```diff
@@ -161,13 +162,19 @@ public void createRootConjunction(ResolvableConjunction conjunction, Set<Variabl
...
         createRootController(reasonerConsumer, actorFn);
     }
 
     public void createExplainableRoot(Concludable concludable, ConceptMap bounds, ReasonerConsumer<Explanation> reasonerConsumer) {
+        Set<com.vaticle.typedb.core.pattern.variable.Variable> boundVariables = Iterators.iterate(bounds.concepts().keySet())
+                .map(id -> concludable.pattern().variable(id))
+                .filter(v -> v != null).toSet();
+        planner().planAllDependencies(concludable, boundVariables);
+
         Function<Driver<ConcludableController.Explain>, ConcludableController.Explain> actorFn =
```
krishnangovindraj, line 178: Since we plan the whole sub-graph at once, we plan it at creation of the root.
> It is single threaded, but we can fix that later if needed.

reasoner/planner/ReasonerPlanner.java
```diff
@@ -51,37 +54,71 @@ public ReasonerPlanner(TraversalEngine traversalEng, ConceptManager conceptMgr,
...
+    static boolean dependenciesSatisfied(Resolvable<?> resolvable, Set<Variable> bounds, Map<Resolvable<?>, Set<Variable>> dependencies) {
         return bounds.containsAll(dependencies.get(resolvable));
     }
 
+    public void plan(ResolvableConjunction conjunction, Set<Variable> boundVariables) {
+        CallMode callMode = new CallMode(conjunction, estimateableVariables(boundVariables));
+        // We can't use `planCache.get(Key, Func)` below because of the recursive-update error
+        if (planCache.getIfPresent(callMode) == null) {
+            synchronized (this) {
+                if (planCache.getIfPresent(callMode) == null) {
```
flyingsilverfin, line 77: what's the recursive update error? This is basically saying let's use a thread-safe `CommonCache`, but still require all planning happens in this single-threaded block -- can we avoid the recursive planning problem somehow?

reasoner/planner/ReasonerPlanner.java
```diff
@@ -101,6 +138,56 @@ static Map<Resolvable<?>, Set<Identifier.Variable.Retrievable>> dependencies(Set
...
+        }
+
+        @Override
+        public String toString() {
+            return conjunction + "::{" + iterate(bounds).reduce("", (x, y) -> (x + ", " + y)) + "}";
+        }
+    }
+
     public static class Plan {
         private final List<Resolvable<?>> elementOrder;
```
flyingsilverfin, line 192: just realised this should be `resolvableOrder` right? why introduce a new term "element"?

reasoner/planner/ReasonerPlanner.java
```diff
@@ -101,6 +138,56 @@ static Map<Resolvable<?>, Set<Identifier.Variable.Retrievable>> dependencies(Set
...
+        public int hashCode() {
+            return this.hash;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            CallMode that = (CallMode) o;
+            return conjunction.equals(that.conjunction) &&
```
flyingsilverfin, line 181: these can go inline if they fit within the line limits

reasoner/planner/ReasonerPlanner.java
```diff
@@ -101,6 +138,56 @@ static Map<Resolvable<?>, Set<Identifier.Variable.Retrievable>> dependencies(Set
...
+
+        CallMode(ResolvableConjunction conjunction, Set<Variable> bounds) {
+            this.conjunction = conjunction;
+            this.bounds = bounds;
+            this.hash = Objects.hash(conjunction, bounds);
+        }
+
+        @Override
+        public int hashCode() {
+            return this.hash;
```
flyingsilverfin, line 173: don't need `this` when it's unambiguous

reasoner/planner/ReasonerPlanner.java
```diff
@@ -51,37 +54,71 @@ public ReasonerPlanner(TraversalEngine traversalEng, ConceptManager conceptMgr,
...
+            synchronized (this) {
+                if (planCache.getIfPresent(callMode) == null) {
+                    plan(callMode);
+                }
+            }
+        }
+    }
+
+    public void planAllDependencies(Concludable concludable, Set<Variable> boundVariables) {
+        triggeredCalls(concludable, estimateableVariables(boundVariables), Optional.empty()).forEach(callMode -> plan(callMode.conjunction, callMode.bounds));
```
flyingsilverfin, line 85: in this case, you do want a line-break before the `.forEach` so it doesn't overrun the line limits

flyingsilverfin (commented):
Before I go any further, I left some small comments, but here is some high level feedback:
Your naming is often reflecting the computer-science or logical term ("component" as in SCC -- instead of CallGraph, "OrderingChoice" for the a plan -- instead of PlanChoice)

In general, once you create a construct (method, class, module, project) always look at what it contains inside, its APIs, and try to verify the name of the construct against its content.

Here's what I noticed:
1. An `OrderingChoice` is simply a CallMode, plus an actual ordering of resolvables, plus some costs. Compare this to a `Plan` - which is a list of resolvables, plus a cost. So why introduce a new term "order"? Did you notice their similary (naming the variables inside consistently helps too)? If you make them converge you'd see that perhaps a name like `PlanChoice` or `CostedPlanChoice` would work quite well.

2. A "ComponentPlan" is a set of plans for each called conjunction (in their respective modes). This becomes more obvious after you do the above rename. So, if you already have term "CallMode", used inside of a `ComponentPlan`, and we notice a `ConjunctionGraph` exists, we may create the name out of existing terms, such as `CallGraphPlans`.

make sure you do those renames consistently - for example, the current state still has `Key` in use in some places (`pendingKeys`), even though `CallKey` is gone. Granted, in a large PR this is harder.

I'll keep adding comments for a bit. Doing the two above should already help you notice when methods are doing similar work.

reasoner/planner/RecursivePlanner.java
```diff
@@ -0,0 +1,383 @@
...
+
+    private void recursivelyGenerateOrderingChoices(CallMode callMode) {
+        if (!candidateOrderings.containsKey(callMode)) {
+            candidateOrderings.put(callMode, null); // Guard
+            ConjunctionNode conjunctionNode = conjunctionGraph.conjunctionNode(callMode.conjunction);
+            answerCountEstimator.buildConjunctionModel(callMode.conjunction);
+
+            // Two orderings for the same CallMode with identical cyclic-concludable modes are interchangeable in the component-plan
+            //      -> We only need to keep the cheaper one.
+            Map<Set<Pair<Concludable, Set<Variable>>>, OrderingChoice> bestChoice = new HashMap<>();
```
flyingsilverfin, line 163: you really love "best" haha!

The point of this method, if we use the new naming and keep it verbose for now, is to `generateCallGraphPlanChoices`. So really, the main point of this variable is not to store the "best" choice right? That's a detail of resolving two interchangable plans. The main thing is to actually create the choices. So i'd call this `planChoices`, and move the comment you have above it to the `if` statement below

reasoner/planner/RecursivePlanner.java
```diff
@@ -0,0 +1,383 @@
...
+            PartialOrderReductionSearch porSearch = new PartialOrderReductionSearch(logicMgr.compile(callMode.conjunction), callMode.bounds);
+            for (List<Resolvable<?>> ordering : porSearch.allOrderings()) {
```
flyingsilverfin, line 165: i'd still call this variable `plans`, even though the search returns orderings. This is where the domain changes: the user of an API is allowed to bring it into a new domain (a plan is a specialisation of an order), so we can name it so here.

reasoner/planner/RecursivePlanner.java
```diff
@@ -0,0 +1,383 @@
...
+                if (!bestChoice.containsKey(orderingChoice.cyclicConcludableBounds) ||
+                        bestChoice.get(orderingChoice.cyclicConcludableBounds).acyclicCost > orderingChoice.acyclicCost) {
+                    bestChoice.put(orderingChoice.cyclicConcludableBounds, orderingChoice);
+                }
+            }
+            candidateOrderings.put(callMode, new HashSet<>(bestChoice.values()));
+        }
+    }
+
+    private OrderingChoice summariseOrdering(ConjunctionNode conjunctionNode, List<Resolvable<?>> ordering, Set<Variable> inputBounds) {
```
flyingsilverfin, line 177: everywhere you use `bounds`, should we be using `mode`?

reasoner/planner/RecursivePlanner.java
```diff
@@ -0,0 +1,383 @@
...
+        assert bestPlan != null;
+        return bestPlan;
+    }
+
+    private void initialiseOrderingDependencies(ConjunctionNode conjunctionNode, List<Resolvable<?>> ordering, Set<Variable> inputBounds) {
+        Set<Variable> currentBounds = new HashSet<>(inputBounds);
+        for (Resolvable<?> resolvable : ordering) {
+            Set<Variable> resolvableBounds = iterate(consideredVariables(resolvable)).filter(currentBounds::contains).toSet();
+            initialiseDependencies(conjunctionNode, resolvable, resolvableBounds);
+            if (!resolvable.isNegated()) currentBounds.addAll(consideredVariables(resolvable));
```
flyingsilverfin, line 133: if you're going to do an operation twice, `consideredVariables(resolvable)` it's normally better to save it to a variable to avoid double-computation

reasoner/planner/ReasonerPlanner.java
krishnangovindraj: I don't think you can insert into the cache if you're evaluating the function passed to get()

reasoner/planner/RecursivePlanner.java
krishnangovindraj: Good point

reasoner/planner/RecursivePlanner.java
krishnangovindraj: We've been somewhat consistent that a plan has a cost associated with it.

reasoner/planner/RecursivePlanner.java
```diff
@@ -0,0 +1,382 @@
...
+        } else if (resolvable.isNegated()) {
+            iterate(resolvable.asNegated().disjunction().conjunctions()).forEachRemaining(conjunction -> {
+                CallMode callMode = new CallMode(conjunction, resolvableMode);
+                recursivelyGeneratePlanChoices(callMode);
+                plan(callMode);
+            });
+        }
+    }
+
+    private void recursivelyGeneratePlanChoices(CallMode callMode) {
```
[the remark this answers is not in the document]
flyingsilverfin: the naming of `recursivelyGeneratePlanChoices` is informative. the `initialiseOrderingDependencies` is also good once we have established what an ordering is - some things to build on

reasoner/planner/RecursivePlanner.java
flyingsilverfin: yeah alright

reasoner/planner/RecursivePlanner.java
```diff
@@ -0,0 +1,382 @@
...
+                if (!planChoices.containsKey(planChoice.cyclicConcludableModes) ||
+                        planChoices.get(planChoice.cyclicConcludableModes).acyclicCost > planChoice.acyclicCost) {
+                    planChoices.put(planChoice.cyclicConcludableModes, planChoice);
+                }
+            }
+            this.planChoices.put(callMode, new HashSet<>(planChoices.values()));
+        }
+    }
+
+    private PlanChoice summarisePlanChoice(ConjunctionNode conjunctionNode, List<Resolvable<?>> ordering, Set<Variable> mode) {
```
flyingsilverfin, line 174: do we still want to keep the term "summarise?

reasoner/planner/RecursivePlanner.java
```diff
@@ -0,0 +1,382 @@
...
+        Set<Variable> restrictedBounds = new HashSet<>(); // Restricted by preceding resolvables
+        double acyclicCost = 0L;
+
+        Map<Concludable, Double> scalingFactors = new HashMap<>();
+        Set<Pair<Concludable, Set<Variable>>> cyclicConcludableModes = new HashSet<>();
+        AnswerCountEstimator.IncrementalEstimator estimator = answerCountEstimator.createIncrementalEstimator(conjunctionNode.conjunction());
+        for (Resolvable<?> resolvable : ordering) {
+            Set<Variable> estimateableVars = estimateableVariables(resolvable);
+            Set<Variable> resolvableMode = Collections.intersection(estimateableVars, bounds);
+            Set<Variable> restrictedResolvableBounds = Collections.intersection(estimateableVars, restrictedBounds);
```
flyingsilverfin, line 185: `bounds` -> Mode right?

reasoner/planner/RecursivePlanner.java
```diff
@@ -0,0 +1,382 @@
...
+
+            if (enableConnectednessRestriction) {
+                List<Resolvable<?>> connectedEnabled = iterate(enabled)
+                        .filter(r -> iterate(estimateableVariables(r)).anyMatch(currentBounds::contains)).toList();
+                if (!connectedEnabled.isEmpty()) {
+                    enabled = connectedEnabled;
+                }
+            }
+
+            Set<Resolvable<?>> newSleepSet = new HashSet<>(sleepSet);
```
flyingsilverfin, line 359: hmm is `sleep` and `awaken` common terminology? Not heard it before

reasoner/planner/RecursivePlanner.java
```diff
@@ -0,0 +1,382 @@
...
+
+            if (enabled.isEmpty()) {
+                enabled = iterate(remaining).filter(r -> !r.isNegated() && !sleepSet.contains(r)).toList();
+            }
+
+            if (enabled.isEmpty()) {
+                return; // All enabled are sleeping
+            }
+
+            if (enableConnectednessRestriction) {
```
flyingsilverfin, line 351: this is interesting! What's this boolean and why do we have it?

reasoner/planner/GreedyCostSearch.java
```diff
@@ -40,54 +41,69 @@ public GreedyCostSearch(TraversalEngine traversalEng, ConceptManager conceptMgr,
...
-        assert resolvables.size() == orderedResolvables.size() && iterate(orderedResolvables).allMatch(r -> resolvables.contains(r));
+        assert resolvables.size() == orderedResolvables.size() && iterate(orderedResolvables).allMatch(resolvables::contains);
         return new Plan(orderedResolvables, cost);
     }
 
-    abstract long estimateCost(Resolvable<?> resolvable, Set<Identifier.Variable.Retrievable> bounds);
+    abstract long estimateCost(Resolvable<?> resolvable, Set<Variable> bounds);
 
-    public static class OldPlannerEmulator extends GreedyCostSearch {
+    public static class HeuristicPlannerEmulator extends GreedyCostSearch {
```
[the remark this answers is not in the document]
flyingsilverfin: shall we still do this?

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -392,13 +407,15 @@ private RelationModel(RelationConstraint relation, long relationTypeEstimate,
                 this.rolePlayerTypes = new HashMap<>();
 
                 relation.players().forEach(player -> {
```
flyingsilverfin, line 409: by the way, not sure if you need/want to use this, but the relation role player always should have a set of `inferredRoleTypes` even if there is not role type variable?

reasoner/controller/ControllerRegistry.java
```diff
@@ -153,6 +153,7 @@ public void terminate(Throwable e) {
 
     public void createRootConjunction(ResolvableConjunction conjunction, Set<Variable.Retrievable> filter,
                                       boolean explain, ReasonerConsumer<ConceptMap> reasonerConsumer) {
+        planner().plan(conjunction, new HashSet<>());
```
flyingsilverfin, line 156: hmm okay so it's odd to do planning in the registry... the registry should only deal with registering actors & managing their lifetimes, and not know/care about planning - that's too low-level.
So - can the root controllers themselves create the plans?

flyingsilverfin (changes requested):

reasoner/planner/RecursivePlanner.java
krishnangovindraj: Reduces the search-space by only scheduling resolvables which are connected to some bound.
We can get rid of it and make it the default behaviour, because I think the alternative will just be too slow.

reasoner/planner/RecursivePlanner.java
krishnangovindraj: Not awaken, but sleep-sets are a standard way of doing POR.
Awaken is just the best name I could think off for things which need to be removed from the sleep-set

reasoner/planner/RecursivePlanner.java
krishnangovindraj: No, but it's a bit hard to think of a better term. createPlanChoice?

reasoner/planner/RecursivePlanner.java
krishnangovindraj: Unsure. Because a mode is like a function-signature. This is the set of variables bound within the conjunction.

reasoner/planner/RecursivePlanner.java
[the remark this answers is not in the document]
krishnangovindraj: I feel like 
recursivelyGenerateOrderings
initialiseOrderingDependencies
initialiseResolvableDependencies 
is quite tidy. planChoices are just orderings with extra info on them

reasoner/planner/GreedyCostSearch.java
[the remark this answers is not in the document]
krishnangovindraj: Yeah. The code for doing the greedy stuff isn't complicated enough to justify keeping.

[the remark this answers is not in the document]
krishnangovindraj: Or maybe it is.
I feel like it's going to be useful to have to quickly test out greedy planners if we want to.

reasoner/planner/AnswerCountEstimator.java
krishnangovindraj: Too complicated to be useful.

krishnangovindraj: Thank you!
Updated to consider.

reasoner/planner/RecursivePlanner.java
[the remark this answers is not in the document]
flyingsilverfin: the only reason i would object is that oyu have `PlanChoice` as a class, but Ordering is not... how about renaming `PlanChoice` to `OrderingChoice` and then going with your naming?

reasoner/planner/RecursivePlanner.java
flyingsilverfin: better, make sure you change it to createOrderingChoice if you go with that naming

reasoner/planner/RecursivePlanner.java
flyingsilverfin: `restrictedVariables`?

reasoner/planner/RecursivePlanner.java
```diff
@@ -0,0 +1,382 @@
...
+                List<Resolvable<?>> connectedEnabled = iterate(enabled)
+                        .filter(r -> iterate(estimateableVariables(r)).anyMatch(currentBounds::contains)).toList();
+                if (!connectedEnabled.isEmpty()) {
+                    enabled = connectedEnabled;
+                }
+            }
+
+            Set<Resolvable<?>> newSleepSet = new HashSet<>(sleepSet);
+            for (Resolvable<?> next : enabled) {
+                Set<Resolvable<?>> awaken = iterate(newSleepSet) //  Don't need to iterate(remaining), do I?
```
flyingsilverfin, line 361: Make sure you resolve and remove this kind of comment. We should never let this kind of comment make it into production code!

reasoner/planner/RecursivePlanner.java
flyingsilverfin: maybe `processing`?

reasoner/planner/RecursivePlanner.java
flyingsilverfin: okay, let's do that

reasoner/planner/RecursivePlanner.java
krishnangovindraj: Not really. Why processing?

reasoner/planner/RecursivePlanner.java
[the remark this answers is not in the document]
krishnangovindraj: Yeah I like that. It's where we came from but it feels right.

reasoner/planner/RecursivePlanner.java
krishnangovindraj: Yes, that would work

krishnangovindraj: Now I have restrictedVars, boundVars and Mode to mean 3 different things, but I like it.
restricted -> The space is restricted by this conjunction
bound -> During execution, this variable will have a concept associated with it.
mode -> The set of variables which will be bound when the rule is invoked.

PR #6657, opened 2022-10-05
Centralise cross-type value comparison

flyingsilverfin:

We introduce a generic `compare(ValueType<T> lhsType, T lhs, ValueType<U> rhsType, U rhs)` method to handle cross-value comparison. This strips out code duplication across TypeDB: Previously `Unifier`, `ValueConstraint` and `Predicate` all handled cross-type casting and comparison themselves.

* introduce `Encoding.ValueType.compare(ValueType<T> lhsType, T lhs, ValueType<U> rhsType, U rhs)`
* using this generic method, we can simplify `Predicate` to the point that both `Unifier` and `ValueConstraint` can direct themselves to use the `Predicate`, rather than the `PredicateArgument`.

This change prepares us for introducing a `ValueVertex` that will do cross-type value comparison and enable merging value across attribute types.


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

encoding/Encoding.java
```diff
@@ -409,7 +411,7 @@ public Infix infix() {
         );
 
         private static final Map<ValueType<?>, Set<ValueType<?>>> ASSIGNABLES = map(
-                pair(OBJECT, set(OBJECT)),
+                pair(OBJECT, set()),
```
flyingsilverfin, line 414: `OBJECT` valueType never has instances - it's only used for types... conceptually it is not comparable or assignable to anything since it cannot have instances either

encoding/Encoding.java
```diff
@@ -476,6 +478,31 @@ public static ValueType<?> of(TypeQLArg.ValueType typeQLValueType) {
             else throw TypeDBException.of(UNRECOGNISED_VALUE);
         }
 
+        public static <T, U> int compare(ValueType<T> firstType, T firstValue, ValueType<U> secondType, U secondValue) {
```
flyingsilverfin, line 481: our new generic, centralised method for comparing across value types. In the future, we can extend this to handle more types.

In addition, should `ValueType` grow, we can split it into actual sub-classes, eg. `Encoding.ValueType.Long/Double/String/Boolean/DateTime` so that these kind of switching logic and the comparables & assignables maps above can be absorbed.

logic/resolvable/Unifier.java
```diff
@@ -358,48 +359,33 @@ static Set<? extends ValueConstraint.Constant<?>> constantValueConstraints(Set<V
...
-                PredicateOperator.Equality operator = PredicateOperator.Equality.of(value.predicate().asEquality());
-                if (value.isLong()) {
-                    predicateFn = (a) ->
-                            PredicateArgument.Value.LONG.apply(operator, ((AttributeImpl<?>) a).readableVertex(), value.asLong().value());
-                } else if (value.isDouble()) {
-                    predicateFn = (a) ->
-                            PredicateArgument.Value.DOUBLE.apply(operator, ((AttributeImpl<?>) a).readableVertex(), value.asDouble().value());
-                } else if (value.isBoolean()) {
-                    predicateFn = (a) ->
-                            PredicateArgument.Value.BOOLEAN.apply(operator, ((AttributeImpl<?>) a).readableVertex(), value.asBoolean().value());
```
flyingsilverfin, line 374: it's was strange that we had to reach in and apply a `PredicateArgument` operation directly. Now, we only access `Predicate` for normal comparisons, and apply `PredicateOperator` directly for special cases like regexes and `contains`.

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -157,7 +157,7 @@ public java.lang.String toString() {
         return owner.toString() + SPACE + predicate.toString() + SPACE + value.toString();
     }
 
-    public boolean inconsistentWith(ValueConstraint<?> valueConstraint) {
+    public <TYPE> boolean inconsistentWith(ValueConstraint<TYPE> valueConstraint) {
```
flyingsilverfin, line 160: type parameter lets us access a `ValueConstraint`'s value encoding and actual value bound to the same type

pattern/constraint/thing/ValueConstraint.java
```diff
@@ -259,16 +263,10 @@ public Double asDouble() {
...
+            public <TYPE> boolean isConsistentWith(ValueConstraint.Constant<TYPE> other) {
                 if (other.predicate() != EQ) return true;
-                if (!valueEncoding.comparableTo(other.valueEncoding)) return false;
-                if (other.isDouble()) {
-                    return Predicate.Value.Numerical.of(predicate.asEquality(), PredicateArgument.Value.DOUBLE)
-                            .apply(other.asDouble().value(), value.doubleValue());
-                } else {
-                    return Predicate.Value.Numerical.of(predicate.asEquality(), PredicateArgument.Value.LONG)
-                            .apply(other.asLong().value, value);
-                }
```
flyingsilverfin, line 271: we eliminated the casting and checking here! Now we can actually just redirect into the `Predicate` to do all the required operations, which in turn uses `Encoding.ValueType.compare()`

traversal/predicate/Predicate.java
```diff
@@ -95,8 +95,12 @@ public boolean apply(AttributeVertex<?> vertex, Traversal.Parameters.Value value
             return argument.apply(operator, vertex, value);
         }
 
-        public boolean apply(ARG_TYPE lhs, ARG_TYPE rhs) {
-            return argument.apply(operator, lhs, rhs);
+        public <T> boolean apply(AttributeVertex<T> vertex, ARG_TYPE rhs) {
+            return apply(vertex.valueType(), vertex.value(), rhs);
+        }
+
+        public <T> boolean apply(Encoding.ValueType<T> lhsType, T lhs, ARG_TYPE rhs) {
```
flyingsilverfin, line 102: new API on `Predicate` lets us decouple the `Preidcate` from having to be used from an `AttributeVertex` or a `Traversal.Params`, and enables the completely generic execution of a language predicate on two values

traversal/predicate/PredicateArgument.java
```diff
@@ -36,7 +36,7 @@ public abstract class PredicateArgument {
 
     private final String symbol;
 
-    protected PredicateArgument(String symbol) {
+    PredicateArgument(String symbol) {
```
flyingsilverfin, line 39: since the `Predicate` is more generally useful now, we don't need to expose any methods in this class to be accessed publicly

haikalpribadi (approved):

PR #6659, opened 2022-10-05
Refine estimated costs of traversing intersections

dmitrii-ubskii:

We refine the estimate of the costs of an intersection of multiple edges incoming into a vertex to better reflect the number of operations performed during traversal. Because the cost of traversing to a vertex increases with the number of edges intersecting at the vertex, penalties may be incurred when intersecting many cheap edges. The query planner model is made aware of these penalties. 

Consider the following query structure:

```
A  -  B
 \   /
   C
```

Suppose:
- we always begin traversal from A,
- the cost of traversal of edges AC and BC is 1, that is, for each concept match by A or B we will match exactly one C concept,
- the cost of AB is cheaper than that of CB.

The two possible plans are as follows, traversed top-to-bottom:

```
A            A
|\          /|
| B   or   C |
|/          \|
C            B
```

The cost of the plan is estimated as the product of the costs of querying each vertex. The cost of starting at a vertex is that vertex's cardinality, i.e. the number of concepts that vertex would match in isolation. The cost of traversing to a vertex via edges is estimated as the minimum of the costs of each incoming edge. The cost of an edge, in turn, is an estimate of the average number of concepts that would be matched by traversing this vertex given a fixed value at the edge's source, plus one to account for verifying that there are no further answers.

Given all that, the estimated costs of the plans in question are estimated as:

```
|A| * (1 + |AB|) * (1 + min(|AC|, |BC|)) = |A| * (1 + |AB|) * 2

|A| * (1 + |AC|) * (1 + min(|AB|, |CB|)) = |A| * 2 * (1 + |AB|)
```

which are equivalent. In other words, the planner cannot tell the difference between the two plans.

It turns out that for most sets of data the first plan will perform significantly worse. A more correct estimate of the work done to traverse to a vertex via some set of edges is `N * (1 + |V| * Π |E|/|V|)`, where N is the number of edges intersecting at the vertex. `|E|/|V|` is the "filtering power" of an edge, which is to say, it's the portion of the concepts at its destination that traversing the edge would match, on average. `1 + |V| * Π |E|/|V|` therefore represents the total expected number of matches produced by the intersection as a whole, once again accounting for verifying that there are no more answers by adding one. Finally, the factor of `N` represents the fact that each of the `N` edges must be queried at least once to produce each answer.

With that in mind, the estimated costs of the two plans from before would instead be:

```
|A| * (1 + |AB|) * 2 * (1 + |AC| * |BC| / |C|) = 
    |A| * (1 + |AB|) * 2 * (1 + 1 / |C|) ≈
    2 * |A| * (1 + |E|) * (1 + 1/|V|) ≈
    2 * |A| * (1 + |E|)

|A| * (1 + |AC|) * 2 * (1 + |AB| * |CB| / |B|) =
    |A| * 2 * 2 * (1 + |AB| * |CB| / |B|) ≈
    4 * |A| * (1 + |E|² / |V|) ≈
    4 * |A|
```

Assuming `|AB| ≈ |CB| ≈ |E|`, `|B| ≈ |C| ≈ |V|`, `|E| << |V|`.

Because the cost of traversing to a vertex via an edge is capped at N from below by the need to verify that the iterator has no more matches, intersecting many cheap edges is not as powerful as intersecting few expensive ones. This change is now reflected in the planner model. The additional complexity slows down the convergence quite a bit, but intermediate solutions are generally better due to more accurate estimation.

### DIFF
--- a/common/optimiser/Optimiser.java
+++ b/common/optimiser/Optimiser.java
@@ -54,6 +54,7 @@ public class Optimiser {
     private MPSolverParameters parameters;
     private Status status;
     private boolean isConstraintsUpToDate;
+    private long lastTimeLimit;
     private Double objectiveValue;
 
     public Optimiser() {
@@ -63,6 +64,7 @@ public class Optimiser {
         status = Status.NOT_SOLVED;
         objectiveValue = null;
         isConstraintsUpToDate = true;
+        lastTimeLimit = 0;
     }
 
     public synchronized Status optimise(long timeLimitMillis) {
@@ -70,7 +72,8 @@ public class Optimiser {
         assert solutionSatisfiesModel();
         maySetSolver();
         setSolutionAsHints();
-        solver.setTimeLimit(timeLimitMillis);
+        lastTimeLimit += timeLimitMillis;
+        solver.setTimeLimit(lastTimeLimit);
         status = Status.of(solver.solve(parameters));
         recordSolverValues();
         if (isOptimal()) releaseSolver();
@@ -79,10 +82,13 @@ public class Optimiser {
     }
 
     private void maySetSolver() {
-        if (solver == null) createSolver();
-        else if (!isConstraintsUpToDate) {
+        if (solver == null) {
+            createSolver();
+            lastTimeLimit = 0;
+        } else if (!isConstraintsUpToDate) {
             solver.reset();
             setConstraintCoefficients();
+            lastTimeLimit = 0;
         }
     }
 
--- a/traversal/planner/PlannerEdge.java
+++ b/traversal/planner/PlannerEdge.java
@@ -198,7 +198,7 @@ public abstract class PlannerEdge<VERTEX_FROM extends PlannerVertex<?>, VERTEX_T
             for (int i = 0; i < numAdjacent; i++) {
                 OptimiserConstraint conMinimalWithMultiplicity = planner.optimiser().constraint(0, 1, conPrefix + "minimal_with_multiplicity[" + i + "]");
                 conMinimalWithMultiplicity.setCoefficient(varIsMinimal, 1);
-                conMinimalWithMultiplicity.setCoefficient(to.varNumInsEncoding[i + 1], 1);
+                conMinimalWithMultiplicity.setCoefficient(to.varNumInsSelectedOneHot[i + 1], 1);
                 conMinimalWithMultiplicity.setCoefficient(varIsMinimalWithMultiplicity[i], -2);
             }
 
@@ -281,7 +281,7 @@ public abstract class PlannerEdge<VERTEX_FROM extends PlannerVertex<?>, VERTEX_T
                                     noneMatch(e -> e.cheaperThan(this))
             );
             for (int i = 0; i < to.ins().size(); i++) {
-                varIsMinimalWithMultiplicity[i].setValue(varIsMinimal.value() && to.varNumIns.value() == i+1);
+                varIsMinimalWithMultiplicity[i].setValue(varIsMinimal.value() && to.varNumInsSelected.value() == i+1);
             }
         }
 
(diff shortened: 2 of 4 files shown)


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

common/optimiser/Optimiser.java
```diff
@@ -63,14 +64,16 @@ public Optimiser() {
         status = Status.NOT_SOLVED;
         objectiveValue = null;
         isConstraintsUpToDate = true;
+        lastTimeLimit = 0;
```
dmitrii-ubskii, line 67: Needed to emulate incrementality in the MPSolver.

traversal/planner/PlannerEdge.java
```diff
@@ -186,6 +194,14 @@ void createOptimiserConstraints() {
             conIsSelected.setCoefficient(to.varOrderNumber, -1);
 
             int numAdjacent = to.ins().size();
+
+            for (int i = 0; i < numAdjacent; i++) {
+                OptimiserConstraint conMinimalWithMultiplicity = planner.optimiser().constraint(0, 1, conPrefix + "minimal_with_multiplicity[" + i + "]");
+                conMinimalWithMultiplicity.setCoefficient(varIsMinimal, 1);
+                conMinimalWithMultiplicity.setCoefficient(to.varNumInsEncoding[i + 1], 1);
+                conMinimalWithMultiplicity.setCoefficient(varIsMinimalWithMultiplicity[i], -2);
+            }
```
dmitrii-ubskii, line 203: Edge is minimal-with-multiplicity-i iff 1) it is minimal, and 2) its destination vertex has i+1 incoming edges.

traversal/planner/PlannerEdge.java
```diff
@@ -231,7 +247,14 @@ private void recordCost() {
 
         protected void updateOptimiserCoefficients() {
             assert costLastRecorded == safeCost();
-            planner.optimiser().setObjectiveCoefficient(varIsMinimal, log(1 + safeCost()));
+            
+            double[] filters = to.ins().stream().filter(e -> !e.equals(this)).filter(this::cheaperThan).mapToDouble(e -> e.safeCost() / to.safeCost()).toArray();
+            double averageFilter = 1.0;
+            if (filters.length > 0) averageFilter = pow(Arrays.stream(filters).reduce(1.0, (a,b) -> a*b), 1.0 / filters.length);
```
dmitrii-ubskii, line 253: Geometric mean of the "filtering power" of all the adjacent edges which are more expensive than this one. The filtering power is defined as the edge's cardinality divided by the cardinality of its destination vertex, i.e. "what portion of all valid concepts will traversing this edge match?".

traversal/planner/PlannerVertex.java
```diff
@@ -48,9 +48,10 @@
     double cost;
     double costLastRecorded;
 
-    OptimiserVariable.Boolean varIsStartingVertex;
```
dmitrii-ubskii, line 51: Replaced by `varNumInsEncoding[0]`

traversal/planner/PlannerVertex.java
```diff
@@ -125,15 +128,23 @@ void createOptimiserConstraints() {
...
+            conNumInsEncoding.setCoefficient(varNumInsEncoding[i], i);
+            conNumInsEncodingValid.setCoefficient(varNumInsEncoding[i], 1);
+        }
     }
 
     protected void updateOptimiserCoefficients() {
         assert costLastRecorded == safeCost();
-        planner.optimiser().setObjectiveCoefficient(varIsStartingVertex, log(1 + safeCost()));
+        planner.optimiser().setObjectiveCoefficient(varNumInsEncoding[0], log(1 + safeCost()));
+        for (int i = 1; i < ins().size() + 1; i++) planner.optimiser().setObjectiveCoefficient(varNumInsEncoding[i], log(i));
```
dmitrii-ubskii, line 147: Getting each answer on an intersection of N edges requires N seeks.

flyingsilverfin (approved):
Looks good!

PR #6660, opened 2022-10-05
Linearise query traversal

dmitrii-ubskii:

We ensure that the traversal of a query procedure is performed with as little "jumping" between unrelated concepts as possible.

When a graph iterator fetches the candidates for ea, it does so in the order specified in the procedure. Because the graph planner only considers the direction of the edges and not the exact ordering of the vertices, the resulting procedure may look as follows:

```
   0
 /   \
1     2
|     |
4     3
```

In this example, if the vertex no. 4 does not match any things in the database, the iterator would backtrack back to vertex no. 1 and revisit vertices no. 2 & 3, duplicating work unnecessarily.

Through a simple depth-first topological sort, we ensure that the traversal follows edges whenever possible, eliminating this problem.

### DIFF
--- a/traversal/planner/GraphPlanner.java
+++ b/traversal/planner/GraphPlanner.java
@@ -37,6 +37,7 @@ import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -296,15 +297,15 @@ public class GraphPlanner implements ComponentPlanner {
 
     private void linearise() {
         Set<PlannerVertex<?>> visited = new HashSet<>();
-        List<PlannerVertex<?>> stack = vertices.values().stream().filter(PlannerVertex::isStartingVertex).collect(Collectors.toList());
+        LinkedList<PlannerVertex<?>> toVisit = vertices.values().stream().filter(PlannerVertex::isStartingVertex).collect(Collectors.toCollection(LinkedList::new));
         int vertexOrder = 0;
-        while (!stack.isEmpty()) {
-            PlannerVertex<?> vertex = stack.remove(stack.size() - 1);
+        while (!toVisit.isEmpty()) {
+            PlannerVertex<?> vertex = toVisit.removeFirst();
             vertex.setOrder(vertexOrder++);
             visited.add(vertex);
             for (PlannerVertex<?> v : vertex.outs().stream().filter(PlannerEdge.Directional::isSelected).map(PlannerEdge.Directional::to).collect(Collectors.toSet())) {
                 if (!visited.contains(v) && visited.containsAll(v.ins().stream().filter(PlannerEdge.Directional::isSelected).map(PlannerEdge.Directional::from).collect(Collectors.toSet()))) {
-                    stack.add(v);
+                    toVisit.addFirst(v);
                 }
             }
         }
--- a/traversal/planner/GraphPlanner.java
+++ b/traversal/planner/GraphPlanner.java
@@ -310,8 +310,7 @@ public class GraphPlanner implements ComponentPlanner {
                 }
             }
         }
-        assert visited.size() == vertices.size();
-        assert vertexOrder == vertices.size();
+        assert visited.size() == vertices.size() && vertexOrder == vertices.size();
     }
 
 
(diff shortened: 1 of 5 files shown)


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

traversal/procedure/GraphProcedure.java
```diff
@@ -183,10 +185,29 @@ public Builder() {
         }
 
         public GraphProcedure build() {
+            linearise();
             return new GraphProcedure(vertices.values().stream().sorted(comparing(ProcedureVertex::order))
                     .toArray(ProcedureVertex[]::new));
         }
 
+        private void linearise() {
+            Set<ProcedureVertex<?, ?>> orderedSet = new HashSet<>();
```
flyingsilverfin, line 194: shouldn't an `orderedSet` actually be an ordered set? where do we depend on the ordering? Either the naming is off or the type is wrong...

traversal/procedure/GraphProcedure.java
```diff
@@ -183,10 +185,29 @@ public Builder() {
         }
 
         public GraphProcedure build() {
+            linearise();
```
flyingsilverfin, line 188: note that we actually have a procedure permutation generator, that generates all possible permutations of a query procedure using the `Builder`. However, now that we have this, that testing framework won't be as effective...
I recommend making two APIs: one called `lineariseAndBuild` and `build`. The `linearseAndBuild` should be called from the engine, and the `build` can be used from the test.

traversal/procedure/GraphProcedure.java
dmitrii-ubskii: It's meant to be a set of the ordered vertices. `orderedVertices` then.

flyingsilverfin: `visited`?

traversal/procedure/GraphProcedure.java
flyingsilverfin: let's do this whole operation in the `GraphPlanner` actually!

traversal/planner/GraphPlanner.java
```diff
@@ -291,11 +294,31 @@ private void optimise(long timeLimitMillis) {
         isOptimising.set(false);
     }
 
+    private void linearise() {
+        Set<PlannerVertex<?>> visited = new HashSet<>();
+        List<PlannerVertex<?>> stack = vertices.values().stream().filter(PlannerVertex::isStartingVertex).collect(Collectors.toList());
```
flyingsilverfin, line 299: in this case, you should actually be making sure that you're using a `LinkedList` because of the operations you'r doing on it.

Also notice that `visited` and `stack` are not symmetrical... one is an abstract term and one is a contextualised term. Can you find a contextualised term for `stack`?

traversal/planner/GraphPlanner.java
```diff
@@ -291,11 +294,31 @@ private void optimise(long timeLimitMillis) {
...
+            vertex.setOrder(vertexOrder++);
+            visited.add(vertex);
+            for (PlannerVertex<?> v : vertex.outs().stream().filter(PlannerEdge.Directional::isSelected).map(PlannerEdge.Directional::to).collect(Collectors.toSet())) {
+                if (!visited.contains(v) && visited.containsAll(v.ins().stream().filter(PlannerEdge.Directional::isSelected).map(PlannerEdge.Directional::from).collect(Collectors.toSet()))) {
+                    stack.add(v);
+                }
+            }
+        }
+        assert visited.size() == vertices.size();
+        assert vertexOrder == vertices.size();
```
flyingsilverfin, line 312: merge with &&

traversal/planner/GraphPlanner.java
```diff
@@ -291,11 +295,31 @@ private void optimise(long timeLimitMillis) {
         isOptimising.set(false);
     }
 
+    private void linearise() {
+        Set<PlannerVertex<?>> visited = new HashSet<>();
+        LinkedList<PlannerVertex<?>> toVisit = vertices.values().stream().filter(PlannerVertex::isStartingVertex).collect(Collectors.toCollection(LinkedList::new));
```
flyingsilverfin, line 300: this can also be done with our iterators ;)

dmitrii-ubskii: It can now :)

flyingsilverfin (approved):

PR #6664, opened 2022-10-12
Enforce key uniqueness across subtypes

### LINKED ISSUE #6543: Key uniqueness not respected for subtypes
## Description

A subtype of a type that has a key attribute can be inserted with a key set to a value already taken by an instance of the base type. This is, at the very least, unexpected behaviour.

## Environment

1. OS (where TypeDB server runs): Ubuntu 21.10
2. TypeDB version (and platform): TypeDB 2.8.1 for Linux
3. TypeDB client: console

## Reproducible Steps

1. Create a database with the following schema:
```
define
  name sub attribute, value string;
  base sub entity, owns name @key;
  derived sub base;
```
2. Load a base and a derived entity with equivalent keys, e.g.:
```
insert $x isa base, has name "unique";
insert $y isa derived, has name "unique";
```
3. (For completeness) Query the key:
```
match $x isa base, has name "unique";
```

## Expected Output

Error during data loading (step 2) like "Attempted to assign a key 'unique' of type 'name' that had been taken by another 'base'."

NB, this error does fire when trying to insert two entities of type `base` with the same key value.

## Actual Output

No error on data write, the query in step 3 results in the following output:
```
{ $x iid 0x826e80018000000000000000 isa base; }
{ $x iid 0x826e80028000000000000000 isa derived; }
```

dmitrii-ubskii:

We enforce uniqueness of a key owned by a supertype across all its types.

Fetch the supertype before checking whether a key attribute is owned by another thing.

In case of a collision, report the supertype that owns the key.

Update typedb-behaviour to include the tests for this scenario.

Closes #6543

### DIFF
--- a/concept/thing/impl/AttributeImpl.java
+++ b/concept/thing/impl/AttributeImpl.java
@@ -88,9 +88,7 @@ public abstract class AttributeImpl<VALUE> extends ThingImpl implements Attribut
 
     @Override
     public FunctionalIterator<ThingImpl> getOwners(ThingType ownerType) {
-        ThingType topLevelOwnerType = ownerType;
-        while (topLevelOwnerType.getSupertype().getOwns(getType().getValueType()).hasNext()) topLevelOwnerType = topLevelOwnerType.getSupertype();
-        return topLevelOwnerType.getSubtypes().map(ot -> ((ThingTypeImpl) ot).vertex).flatMap(
+        return ownerType.getSubtypes().map(ot -> ((ThingTypeImpl) ot).vertex).flatMap(
                 v -> readableVertex().ins().edge(HAS, (PrefixIID.of(v.encoding().instance())), v.iid()).from()
         ).map(ThingImpl::of);
     }
--- a/concept/thing/impl/AttributeImpl.java
+++ b/concept/thing/impl/AttributeImpl.java
@@ -88,7 +88,9 @@ public abstract class AttributeImpl<VALUE> extends ThingImpl implements Attribut
 
     @Override
     public FunctionalIterator<ThingImpl> getOwners(ThingType ownerType) {
-        return ownerType.getSubtypes().map(ot -> ((ThingTypeImpl) ot).vertex).flatMap(
+        ThingType topLevelOwnerType = ownerType;
+        while (topLevelOwnerType.getSupertype().getOwns(getType().getValueType()).hasNext()) topLevelOwnerType = topLevelOwnerType.getSupertype();
+        return topLevelOwnerType.getSubtypes().map(ot -> ((ThingTypeImpl) ot).vertex).flatMap(
                 v -> readableVertex().ins().edge(HAS, (PrefixIID.of(v.encoding().instance())), v.iid()).from()
         ).map(ThingImpl::of);
     }
--- a/concept/thing/impl/ThingImpl.java
+++ b/concept/thing/impl/ThingImpl.java
@@ -116,7 +116,7 @@ public abstract class ThingImpl extends ConceptImpl implements Thing {
     public void setHas(Attribute attribute, boolean isInferred) {
         validateIsNotDeleted();
         AttributeVertex.Write<?> attrVertex = ((AttributeImpl<?>) attribute).writableVertex();
-        if (!getType().getOwns().findFirst(attribute.getType()).isPresent()) {
+        if (getType().getOwns().findFirst(attribute.getType()).isEmpty()) {
             throw exception(TypeDBException.of(THING_CANNOT_OWN_ATTRIBUTE, attribute.getType().getLabel(), readableVertex().type().label()));
         } else if (getType().getOwns(true).findFirst(attribute.getType()).isPresent()) {
             if (getHas(attribute.getType()).first().isPresent()) {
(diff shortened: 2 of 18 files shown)


concept/thing/impl/AttributeImpl.java
```diff
@@ -88,7 +88,9 @@ public FunctionalIterator<ThingImpl> getOwners() {
 
     @Override
     public FunctionalIterator<ThingImpl> getOwners(ThingType ownerType) {
-        return ownerType.getSubtypes().map(ot -> ((ThingTypeImpl) ot).vertex).flatMap(
+        ThingType topLevelOwnerType = ownerType;
+        while (topLevelOwnerType.getSupertype().getOwns(getType().getValueType()).hasNext()) topLevelOwnerType = topLevelOwnerType.getSupertype();
+        return topLevelOwnerType.getSubtypes().map(ot -> ((ThingTypeImpl) ot).vertex).flatMap(
```
dmitrii-ubskii, line 93: Figure out which supertype owns this key attribute, then check all its subtypes.

concept/thing/impl/ThingImpl.java
```diff
@@ -122,7 +122,7 @@ public void setHas(Attribute attribute, boolean isInferred) {
                 throw exception(TypeDBException.of(THING_KEY_OVER, attribute.getType().getLabel(), getType().getLabel()));
             } else if (attribute.getOwners(getType()).first().isPresent()) {
                 throw exception(TypeDBException.of(THING_KEY_TAKEN, ((AttributeImpl<?>) attribute).getValue(),
-                        attribute.getType().getLabel(), getType().getLabel()));
+                        attribute.getType().getLabel(), attribute.getOwners(getType()).first().get().getType().getLabel()));
```
dmitrii-ubskii, line 125: Make sure the message reports the base type that owns the key, i.e. "owned by another 'base'", not "owned by another 'derived'". The latter can be misleading.

dependencies/vaticle/repositories.bzl
```diff
@@ -48,8 +48,8 @@ def vaticle_typedb_protocol():
 def vaticle_typedb_behaviour():
     git_repository(
         name = "vaticle_typedb_behaviour",
-        remote = "https://github.com/vaticle/typedb-behaviour",
-        commit = "403fd2cbd4d338dd8627ca79867140ae2451a4ce", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
+        remote = "https://github.com/dmitrii-ubskii/typedb-behaviour",
+        commit = "66d8d122dc82a600d70cf379390c511d0624e04f", # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
```
dmitrii-ubskii, line 52: Pending https://github.com/vaticle/typedb-behaviour/pull/228

concept/thing/impl/AttributeImpl.java
```diff
@@ -88,7 +88,9 @@ public FunctionalIterator<ThingImpl> getOwners() {
 
     @Override
     public FunctionalIterator<ThingImpl> getOwners(ThingType ownerType) {
```
flyingsilverfin, line 90: I think this method is doing what it should be doing: getting owners of this attribute from the owner type, including its subtypes. It shouldn't traverse up the types.

concept/thing/impl/ThingImpl.java
```diff
@@ -122,7 +122,7 @@ public void setHas(Attribute attribute, boolean isInferred) {
                 throw exception(TypeDBException.of(THING_KEY_OVER, attribute.getType().getLabel(), getType().getLabel()));
             } else if (attribute.getOwners(getType()).first().isPresent()) {
```
flyingsilverfin, line 123: but you can do the validation differently here:
from `Attribute.getType().getOwnersExplicit(true)` you can get the direct owners of this attribute type. Find the one that is the supertype of this `Thing` - `intersect(getOwnersExplicit(true).toSet(), superTypes().toSet())`, then use that type to check the uniqueness of the ownership with the old version of `getOwners`

dmitrii-ubskii: There can be many supertypes that match, and we only want the highest level one. I don't think the solution you're suggesting is any better than traversing the inheritance structure.

concept/thing/impl/ThingImpl.java
```diff
@@ -120,9 +121,13 @@ public void setHas(Attribute attribute, boolean isInferred) {
         } else if (getType().getOwns(true).findFirst(attribute.getType()).isPresent()) {
             if (getHas(attribute.getType()).first().isPresent()) {
                 throw exception(TypeDBException.of(THING_KEY_OVER, attribute.getType().getLabel(), getType().getLabel()));
-            } else if (attribute.getOwners(getType()).first().isPresent()) {
-                throw exception(TypeDBException.of(THING_KEY_TAKEN, ((AttributeImpl<?>) attribute).getValue(),
-                        attribute.getType().getLabel(), getType().getLabel()));
+            } else {
+                ThingType topLevelOwnerType = getType();
+                while (topLevelOwnerType.getSupertype().getOwns(attribute.getType().getValueType()).hasNext()) topLevelOwnerType = topLevelOwnerType.getSupertype();
+                if (attribute.getOwners(getType()).first().isPresent()) {
```
flyingsilverfin, line 127: should `getOwners()` use `toplevelOwnerType`?
I'd still do `intersect(attribute.getType().getSuperTypes(), getSuperTypes()).iterator().next()`. Shorter!
Also, the variable could be more briefly named `ownerType` and let the reader figure it out from the RHS :)

PR #6666, opened 2022-10-14
Propagate bound-restrictions through conjunction model

krishnangovindraj:
Address the locality assumption of the conjunction answer count model by propagating the reduction in the number of possible bounds for a variable to other resolvable models. The current model (without this) was prone to wildly over-estimating the number of answers to a conjunction containing a strongly selective resolvable.

* The answer-count scales resolvable models according to the estimate of bounds propagated from other resolvables.
* Introduces an `IncrementalEstimator` which can return answers for any prefix when building up a conjunction resolvable by resolvable.
* The propagation implementation is handled by the IncrementalEstimator - Every time a resolvable is added, any reduction  on the estimates of bounds sizes for the variables are propagated  to the scale the other resolvables.

### DIFF
--- a/reasoner/planner/AnswerCountEstimator.java
+++ b/reasoner/planner/AnswerCountEstimator.java
@@ -361,7 +361,6 @@ public class AnswerCountEstimator {
             }
         }
 
-
         private static class RelationModel extends LocalModel {
             private final RelationConstraint relation;
             private final Map<TypeVariable, Long> rolePlayerEstimates;
@@ -383,7 +382,6 @@ public class AnswerCountEstimator {
                     TypeVariable roleType = player.roleType().isPresent() ? player.roleType().get() : null;
                     this.rolePlayerTypes.put(player.player(), roleType);
                 });
-
             }
 
             @Override
@@ -523,7 +521,6 @@ public class AnswerCountEstimator {
 
             inferredRelationsEstimate = Math.min(inferredRelationsEstimate, relationUpperBound - persistedRelationEstimate);
 
-
             for (RelationConstraint.RolePlayer rp : rolePlayers) {
                 TypeVariable key = rp.roleType().orElse(null);
                 rolePlayerCounts.put(key, rolePlayerCounts.getOrDefault(key, 0) + 1);
@@ -546,10 +543,7 @@ public class AnswerCountEstimator {
                     boolean rulesConcludeOwner = iterate(hasConstraint.owner().inferredTypes()).flatMap(ownerType -> iterate(answerCountEstimator.logicMgr.rulesConcluding(ownerType))).hasNext();
                     if (rulesConcludeOwner) ownerEstimate = hasEdgeEstimate / attributeEstimate;
                 }
-
-
             }
-
             return new LocalModel.HasModel(hasConstraint, hasEdgeEstimate, ownerEstimate, attributeEstimate);
         }
 
--- a/reasoner/planner/AnswerCountEstimator.java
+++ b/reasoner/planner/AnswerCountEstimator.java
@@ -113,13 +113,13 @@ public class AnswerCountEstimator {
         private final ConjunctionModel conjunctionModel;
         private final Map<LocalModel, Pair<Double, Optional<Variable>>> modelScale;
         private final Map<Variable, Double> minVariableEstimate;
-        private final Map<Variable, Set<LocalModel>> modelsWithVar;
+        private final Map<Variable, Set<LocalModel>> affectedModels;
 
         private IncrementalEstimator(ConjunctionModel conjunctionModel) {
             this.conjunctionModel = conjunctionModel;
             this.modelScale = new HashMap<>();
             this.minVariableEstimate = new HashMap<>();
-            this.modelsWithVar = new HashMap<>();
+            this.affectedModels = new HashMap<>();
         }
 
         public void extend(Resolvable<?> resolvable) {
@@ -127,7 +127,7 @@ public class AnswerCountEstimator {
             List<LocalModel> models = conjunctionModel.modelsForResolvable(resolvable);
             assert models.stream().allMatch(model -> model.variables.size() > 0);
 
-            iterate(models).flatMap(model -> iterate(model.variables)).forEachRemaining(v -> modelsWithVar.computeIfAbsent(v, v1 -> new HashSet<>()));
+            iterate(models).flatMap(model -> iterate(model.variables)).forEachRemaining(v -> affectedModels.computeIfAbsent(v, v1 -> new HashSet<>()));
 
             iterate(models).filter(model -> model.variables.size() == 1).forEachRemaining(model -> {
                 Variable v = model.variables.stream().findFirst().get();
@@ -138,7 +138,7 @@ public class AnswerCountEstimator {
             });
 
             iterate(models).filter(model -> model.variables.size() > 1).forEachRemaining(model -> {
-                model.variables.forEach(v -> modelsWithVar.get(v).add(model));
+                model.variables.forEach(v -> affectedModels.get(v).add(model));
                 modelScale.put(model, new Pair<>(1.0, Optional.empty()));
                 evaluatePropagationEffectsOn(model, cascadingEffectsAccumulator);
             });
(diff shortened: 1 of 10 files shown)


grabl:
### PR Review Checklist

----
**Code**

**Architecture**

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -60,210 +59,195 @@
...
+        public void extend(Resolvable<?> resolvable) {
+            Map<Variable, Double> unaryUpdates = new HashMap<>();
+            List<LocalModel> models = conjunctionModel.modelsForResolvable(resolvable);
+            for (LocalModel model : models) {
+                addAndCollectUnaryUpdates(model, unaryUpdates);
+            }
+            propagate(unaryUpdates);
+        }
+
+        private void addAndCollectUnaryUpdates(LocalModel model, Map<Variable, Double> unaryUpdates) {
```
krishnangovindraj, line 134: When a model is 'added' its estimate for each variable is compared to the best-known estimate for each variable to determine the scaling-factor is needed. The scaling factor is the ratio of best-known estimate to this model's estimate (at most 1).

The scaling-factor is then used to scale down this model's estimates for each variable - If any of these are lower they are collected to be propagated later.

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -60,210 +59,195 @@
...
+            // Ideally, we'd just remove and add each model again till unaryUpdates is empty.
+            int maxIters = 2 * modelScale.size(); // TODO: Consider pruning out small changes
+            while (!unaryUpdates.isEmpty() && maxIters > 0) {
+                Set<LocalModel> affectedModels = iterate(unaryUpdates.keySet()).flatMap(v -> iterate(modelsWithVar.get(v))).toSet();
+                assert iterate(unaryUpdates.entrySet()).allMatch(update -> update.getValue() < bestUnary.getOrDefault(update.getKey(), Double.MAX_VALUE));
+                bestUnary.putAll(unaryUpdates);
+                unaryUpdates.clear();
+                affectedModels.forEach(model -> addAndCollectUnaryUpdates(model, unaryUpdates));
+                maxIters--;
             }
```
krishnangovindraj, line 183: When adding a model causes a revision in the best-known estimate, we updated the estimate and 'propagate' the change to other models by recomputing their scaling-factor based on the new estimate. This is the same process as adding them for the first time.
Since one bound reducing can lead to other ones being reduced and so on, we restrict the number of iterations of the propagation.

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -60,210 +59,195 @@
...
+            Set<Variable> variables = new HashSet<>();
+            iterate(estimateVariables).filter(model.variables::contains).forEachRemaining(variables::add);
+            double ans = model.estimateAnswers(variables);
+            if (scale.second().isPresent()) {
+                assert scale.first() <= 1.0;
+                variables.add(scale.second().get());
+                double scaledAns = scale.first() * model.estimateAnswers(variables);
+                ans = Math.min(ans, scaledAns);
+            }
+            return ans;
```
krishnangovindraj, line 228: The effect of the scaling factor on the estimate for a set of variables - 
The scaled estimate is the minimum of 
- The original estimate for the queried set of variables
- The scaling factor * The estimate for the (set of queried variables $$\cup$ the variable which caused the scaling)

krishnangovindraj (commented):
first self-review

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -60,210 +59,204 @@
...
+        if (!conjunctionModels.containsKey(conjunction)) buildConjunctionModel(conjunction);
+        if (!fullConjunctionEstimators.containsKey(conjunction)) {
+            IncrementalEstimator incrementalEstimator = createIncrementalEstimator(conjunction);
+            conjunctionModels.get(conjunction).conjunctionSummary.resolvables().forEach(incrementalEstimator::extend);
+            fullConjunctionEstimators.put(conjunction, incrementalEstimator);
+        }
+        IncrementalEstimator estimator = fullConjunctionEstimators.get(conjunction);
+        Set<Variable> estimateableVariables = ReasonerPlanner.estimateableVariables(variables);
+        return estimator.answerEstimate(estimateableVariables);
     }
```
krishnangovindraj, line 105: Cached scaled models for when all resolvables are included - used by retrieval cost

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -60,210 +59,204 @@
...
+        return new IncrementalEstimator(conjunctionModels.get(conjunction));
+    }
 
-        private void add(ResolvableConjunction conjunction) {
-            stack.add(conjunction);
-            concludables.put(conjunction, null);
-        }
+    public static class IncrementalEstimator {
+        private final ConjunctionModel conjunctionModel;
+        private final Map<LocalModel, Pair<Double, Optional<Variable>>> modelScale;
```
flyingsilverfin, line 114: I think to be honest if we can get rid of the Optional we should - not having a scale for a model we can just reflect by having a `getOrDefault`?
It might even better to split this into two maps: Model -> Double for modelScale, and Model -> Variable for modelScaleSource ?

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -60,210 +59,204 @@
...
+            this.modelsWithVar = new HashMap<>();
         }
 
-        private boolean contains(ResolvableConjunction conjunction) {
-            return concludables.containsKey(conjunction);
+        public void extend(Resolvable<?> resolvable) {
+            Map<Variable, Double> unaryUpdates = new HashMap<>();
+            List<LocalModel> models = conjunctionModel.modelsForResolvable(resolvable);
+            for (LocalModel model : models) {
+                includeModelAndCheckUpdates(model, unaryUpdates);
```
flyingsilverfin, line 129: try to make the `includeModelAndCheckUpdates` do no changes to the class state, and treat it as a function that communicates with `propagate` purely by reading class state and updating the `unaryUpdates`.
Then, you can even have the function return a new `unaryUpdates` object rather than passing in an empty map and updating in C-style.

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -60,210 +59,204 @@
...
+    }
 
-        private void add(ResolvableConjunction conjunction) {
-            stack.add(conjunction);
-            concludables.put(conjunction, null);
-        }
+    public static class IncrementalEstimator {
+        private final ConjunctionModel conjunctionModel;
+        private final Map<LocalModel, Pair<Double, Optional<Variable>>> modelScale;
+        private final Map<Variable, Double> bestUnary;
```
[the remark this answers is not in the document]
flyingsilverfin: can we actually delete this by getting the current minimum estimate for a variable by multiplying the `modelScale` for the `variableSource` from the `LocalModel` for the variable? 
**but**
For this to be relatively efficient, the `LocalModel` is designed in a way that the project for that variable is cached and doesn't have to recompute each time!

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -60,210 +59,204 @@
...
+        private final Map<Variable, Set<LocalModel>> modelsWithVar;
```
flyingsilverfin, line 116: `variableModels` is a better name more consistent with our existing conventions

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -60,210 +59,204 @@
...
+                    if (v == bestScalingVar) continue;
+                    double newUnary = scaledEstimate(model, modelScale.get(model), set(v));
+                    if (newUnary < bestUnary.getOrDefault(v, Double.MAX_VALUE)) {
+                        unaryUpdates.put(v, Math.min(newUnary, unaryUpdates.getOrDefault(v, Double.MAX_VALUE)));
+                    }
+                }
+            } else {
+                assert bestScaler == 1;
+                modelScale.put(model, new Pair<>(bestScaler, Optional.empty()));
+            }
```
flyingsilverfin, line 179: redundant

reasoner/planner/AnswerCountEstimator.java
krishnangovindraj: They're not variableModels though. Those were models for a variable. This is more "models estimating a variable" or "models affected by a variable"

flyingsilverfin (approved):
Ok!

flyingsilverfin:
Let's move onto the next refactor :)

PR #6678, opened 2022-11-01
Only check key uniqueness within exact type

dmitrii-ubskii:

Key uniqueness is now only respected within a single specific type, without considering its super- and subtypes.

- revert the changes to `ThingImpl::setHas()` that checked whether a supertype of the owner owns the key;
- restrict the above check to only check the owners of the specific type.

### DIFF
--- a/concept/thing/impl/AttributeImpl.java
+++ b/concept/thing/impl/AttributeImpl.java
@@ -27,7 +27,6 @@ import com.vaticle.typedb.core.concept.type.impl.ThingTypeImpl;
 import com.vaticle.typedb.core.encoding.Encoding;
 import com.vaticle.typedb.core.encoding.iid.PrefixIID;
 import com.vaticle.typedb.core.graph.vertex.AttributeVertex;
-import com.vaticle.typedb.core.graph.vertex.TypeVertex;
 
 import java.time.LocalDateTime;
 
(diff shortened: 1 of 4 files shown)


concept/thing/impl/AttributeImpl.java
```diff
@@ -88,9 +89,13 @@ public FunctionalIterator<ThingImpl> getOwners() {
 
     @Override
     public FunctionalIterator<ThingImpl> getOwners(ThingType ownerType) {
-        return ownerType.getSubtypes().map(ot -> ((ThingTypeImpl) ot).vertex).flatMap(
-                v -> readableVertex().ins().edge(HAS, (PrefixIID.of(v.encoding().instance())), v.iid()).from()
-        ).map(ThingImpl::of);
+        return ownerType.getSubtypes().flatMap(this::getOwnersExplicit);
+    }
```
dmitrii-ubskii, line 93: The insides of the `map`s are replicated within `getOwnersExplicit`, and we can use it here instead.

concept/thing/impl/AttributeImpl.java
```diff
@@ -88,9 +89,13 @@ public FunctionalIterator<ThingImpl> getOwners() {
...
+
+    @Override
+    public FunctionalIterator<ThingImpl> getOwnersExplicit(ThingType ownerType) {
```
flyingsilverfin, line 96: @dmitrii-ubskii  ahh I didn't realise - we don't have the pattern of `explicit` methods on things - only Types. It doesn't make sense to have "explicit" ownership on the instance level I think?

dmitrii-ubskii: The idea is to get owners of this explicit type, as the `getOwners()` also gets subtypes.

flyingsilverfin (approved):

PR #6693, opened 2022-11-08
Type inference differentiates illegal and queries with no answer

flyingsilverfin:

We update the usage of type inference to differentiate between an impossible-to-satisfy query (eg. using relations between types that are not playing the right roles, mismatched value types, etc.) and a query that will not have any answers but is semantically sound - such as querying for instances of an abstract type that has no concrete subtypes.

* Type inference sets two booleans on `Conjunction`: `isCoherent` is set to `false` when the query represents a semantic schema violation. `isAnswerable` is set to `false` when the conjunction will never create any answers in the current schema.
* Short-circuit the execution of a query that does not have any answerable conjunctions and return an empty iterator

To implement these changes, we run type inference twice.
1. run type inference, including abstract types in the type inference output
2. If no types are found, then the query is illegal and we throw an exception
3. Run type inference disallowing abstract types. If any variables do not have any types, we set the query ans unanswerable

This restores the built-in assumption implementation of traversal, planning, reasoner, rule validation etc. that only queries where each variable contains a non-empty set of concrete types on each variable shall be processed.

### DIFF
--- a/common/exception/ErrorMessage.java
+++ b/common/exception/ErrorMessage.java
@@ -279,14 +279,12 @@ public abstract class ErrorMessage extends com.vaticle.typedb.common.exception.E
                 new Pattern(14, "Tye type variable '%s' has multiple 'value' constraints.");
         public static final Pattern MULTIPLE_TYPE_CONSTRAINT_REGEX =
                 new Pattern(15, "The type variable '%s' has multiple 'regex' constraints.");
-        public static final Pattern UNSATISFIABLE_PATTERN =
+        public static final Pattern INCOHERENT_PATTERN =
                 new Pattern(16, "The pattern '%s' is illegal in the current schema.");
-        public static final Pattern UNSATISFIABLE_SUB_PATTERN =
+        public static final Pattern INCOHERENT_SUB_PATTERN =
                 new Pattern(17, "The pattern '%s' is illegal in the current schema, due to '%s'.");
-        public static final Pattern UNSATISFIABLE_PATTERN_VARIABLE =
-                new Pattern(18, "The pattern '%s' is illegal in the current schema, due to contradicting types for '%s'.");
-        public static final Pattern UNSATISFIABLE_PATTERN_VARIABLE_VALUE =
-                new Pattern(19, "The pattern '%s' is illegal in the current schema, due to contradicting attribute value types for '%s'.");
+        public static final Pattern INCOHERENT_PATTERN_VARIABLE_VALUE =
+                new Pattern(18, "The pattern '%s' is illegal in the current schema, due to contradicting attribute value types for '%s'.");
 
         private static final String codePrefix = "QRY";
         private static final String messagePrefix = "Invalid Query Pattern";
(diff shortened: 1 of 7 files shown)


common/exception/ErrorMessage.java
```diff
@@ -278,13 +278,13 @@ public static class Pattern extends ErrorMessage {
         public static final Pattern MULTIPLE_TYPE_CONSTRAINT_REGEX =
                 new Pattern(15, "The type variable '%s' has multiple 'regex' constraints.");
         public static final Pattern UNSATISFIABLE_PATTERN =
-                new Pattern(16, "The pattern '%s' can never be satisfied in the current schema.");
+                new Pattern(16, "The pattern '%s' is illegal in the current schema.");
```
flyingsilverfin, line 281: I find "cannot be satisfied" is too ambiguous when compared with un-answerable. Illegal is a stronger term.

logic/LogicCache.java
```diff
@@ -31,23 +31,30 @@
 
 public class LogicCache {
 
-    private final CommonCache<GraphTraversal.Type, Optional<Map<Identifier.Variable.Retrievable, Set<Label>>>> typeInferenceCache;
     private final CommonCache<String, Rule> ruleCache;
     private final CommonCache<Concludable, Map<Rule, Set<Unifier>>> unifiers;
+    private final CommonCache<GraphTraversal.Type, Optional<Map<Identifier.Variable.Retrievable, Set<Label>>>> typeInferenceCache;
+    private final CommonCache<GraphTraversal.Type, Boolean> queryCoherenceCache;
```
flyingsilverfin, line 37: we have two caches: the type inference cache, and query coherence cache. Query coherence is simply a boolean about whether a query is illegal or not. Type inference is the same as the previous cache

reasoner/Reasoner.java
```diff
@@ -96,23 +98,25 @@ public ControllerRegistry controllerRegistry() {
 
     public FunctionalIterator<? extends ConceptMap> execute(Disjunction disjunction, TypeQLMatch.Modifiers modifiers, Context.Query context) {
         inferAndValidateTypes(disjunction);
-        FunctionalIterator<? extends ConceptMap> answers;
         Filter filter = Filter.create(modifiers.filter());
         Optional<Sorting> sorting = modifiers.sort().map(Sorting::create);
         sorting.ifPresent(value -> validateSorting(disjunction, value));
-        if (mayReason(disjunction, context)) {
-            answers = executeReasoner(disjunction, filter, context);
+        Disjunction answerableDisjunction = filterUnanswerable(disjunction);
```
flyingsilverfin, line 104: we eliminate disjunction branches that are unaswerable

reasoner/Reasoner.java
```diff
@@ -231,6 +239,8 @@ public SortedIterator<ConceptMap.Sortable, Order.Asc> executeTraversalSorted(Dis
     }
 
     private Producer<ConceptMap> producer(Conjunction conjunction, Filter filter) {
+        assert conjunction.isCoherent();
+        if (!conjunction.isAnswerable()) return Producers.empty();
```
flyingsilverfin, line 243: short circuit conjunctions which are not answerable but are coherent

haikalpribadi (approved):

PR #6700, opened 2022-11-18
Fix the overwriting of owns edges

flyingsilverfin:

We fix the overwriting of owns edges in the type graph by explicitly deleting pre-existing owns edges from the type vertex. In addition, we fix the conceptual model of type-edge equality functions by removing the actual java type from the equality function definitions, allowing persisted and buffered edges to be treated as equal.

* Explicitly delete pre-existing owns edges when putting a new one into the schema graph
* Collapse separated equality functions for type edges into one consistent equality function

### DIFF
--- a/graph/edge/impl/TypeEdgeImpl.java
+++ b/graph/edge/impl/TypeEdgeImpl.java
@@ -19,11 +19,11 @@
 package com.vaticle.typedb.core.graph.edge.impl;
 
 import com.vaticle.typedb.core.common.exception.TypeDBException;
-import com.vaticle.typedb.core.graph.TypeGraph;
 import com.vaticle.typedb.core.encoding.Encoding;
-import com.vaticle.typedb.core.graph.edge.TypeEdge;
 import com.vaticle.typedb.core.encoding.iid.EdgeViewIID;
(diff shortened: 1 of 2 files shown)


graph/edge/impl/TypeEdgeImpl.java
```diff
@@ -64,6 +65,26 @@ public View.Backward backwardView() {
...
     abstract EdgeViewIID.Type computeBackwardIID();
 
+    abstract VertexIID.Type fromIID();
+
+    abstract VertexIID.Type toIID();
+
+    @Override
+    public final boolean equals(Object object) {
+        if (this == object) return true;
+        if (!(object instanceof TypeEdgeImpl)) return false;
```
flyingsilverfin, line 75: we make all type edges equality-compatible with each other

graph/edge/impl/TypeEdgeImpl.java
```diff
@@ -418,6 +403,16 @@ public TypeVertex to() {
             return to;
         }
 
+        @Override
+        VertexIID.Type fromIID() {
```
flyingsilverfin, line 407: we introduce this method here in order to avoid looking up a cached `from()` in order to get the IID for the equality function.

PR #6701, opened 2022-11-18
Unset overriddes when replacing pre-existing type edges

flyingsilverfin:

We introduce a new API that allows removing the overridden types on a Type edge. This is used to fix a class of bugs that expected edge "overwrite" behaviour in the type graph to unset overridden types on edges, but actually encounter "put" behaviour, meaning the overridden type is never replaced if the edge pre-existed.

* Introduce `TypeEdge.unsetOverridden`, an idempotent operation that removes any existing overridden vertex on the edge
* Refactor setters on the Type concept API that were trying to overwrite any existing edge to use the new `unset` when the exact edge pre-exists

### DIFF
--- a/concept/type/impl/RelationTypeImpl.java
+++ b/concept/type/impl/RelationTypeImpl.java
@@ -146,9 +146,7 @@ public class RelationTypeImpl extends ThingTypeImpl implements RelationType {
                 roleType.setSupertype(RoleTypeImpl.of(graphMgr, graphMgr.schema().rootRoleType()));
             }
             assert roleType.getSupertype() != null;
-            TypeEdge existingEdge = vertex.outs().edge(RELATES, roleType.vertex);
-            if (existingEdge != null) existingEdge.delete();
-            vertex.outs().edge(RELATES, roleType.vertex).overridden(roleType.getSupertype().vertex);
+            vertex.outs().edge(RELATES, roleType.vertex).setOverridden(roleType.getSupertype().vertex);
         }
     }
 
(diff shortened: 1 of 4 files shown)


concept/type/impl/RelationTypeImpl.java
```diff
@@ -146,7 +146,7 @@ public void setRelates(String roleLabel) {
                 roleType.setSupertype(RoleTypeImpl.of(graphMgr, graphMgr.schema().rootRoleType()));
             }
             assert roleType.getSupertype() != null;
-            vertex.outs().edge(RELATES, roleType.vertex).overridden(roleType.getSupertype().vertex);
+            vertex.outs().edge(RELATES, roleType.vertex).setOverridden(roleType.getSupertype().vertex);
```
flyingsilverfin, line 149: rename to `set` to disambiguate explicit new set/unset APIs

concept/type/impl/ThingTypeImpl.java
```diff
@@ -329,11 +329,12 @@ private void ownsAttribute(AttributeTypeImpl attributeType) {
 
         TypeVertex attVertex = attributeType.vertex;
         TypeEdge existingEdge;
-        if ((existingEdge = vertex.outs().edge(OWNS_KEY, attVertex)) != null ||
-                (existingEdge = vertex.outs().edge(OWNS, attVertex)) != null) {
-            existingEdge.delete();
+        if ((existingEdge = vertex.outs().edge(OWNS, attVertex)) != null) {
+            existingEdge.unsetOverridden();
```
flyingsilverfin, line 333: if this exact edge already exists, we just unset the overridden vertex. 
If it is the wrong type of edge (eg. OWNS_KEY), then delete it before creating it. If there is no edge, we just create it immediately

concept/type/impl/ThingTypeImpl.java
```diff
@@ -439,7 +440,9 @@ public void setPlays(RoleType roleType) {
         } else if (getSupertypes().filter(t -> !t.equals(this)).mergeMapForwardable(ThingType::getPlays, ASC).findFirst(roleType).isPresent()) {
             throw exception(TypeDBException.of(PLAYS_ROLE_NOT_AVAILABLE, getLabel(), roleType.getLabel()));
         }
-        vertex.outs().put(Encoding.Edge.Type.PLAYS, ((RoleTypeImpl) roleType).vertex);
+        TypeEdge existingEdge = vertex.outs().edge(PLAYS, ((RoleTypeImpl) roleType).vertex);
+        if (existingEdge != null) existingEdge.unsetOverridden();
```
flyingsilverfin, line 444: if the exact edge exists, we remove the overridden, otherwise we put it without an overridden

haikalpribadi (approved):

PR #6702, opened 2022-11-21
Avoid division by zero in reasoner planner

krishnangovindraj:
Correctly handle code-paths with division so the case where the divisor may be zero.

* Avoid dividing by zero with specific checks in the reasoner planner


flyingsilverfin (approved):

flyingsilverfin:
But it is red :o

krishnangovindraj:
> But it is red :o

Now it is green

flyingsilverfin (approved):

PR #6704, opened 2022-11-22
Simplify isolation manager

flyingsilverfin:

We discover that a ConcurrentSkipListMap does not implement locking/atomic `compute` operations. This behaviour was relied on to implement Isolation Manager's cleanup of transaction state, and errors could manifest as concurrent modification exceptions. To fix this, we refactor the IsolationManager with a simpler state model to optimally allow both isolation validation and deletion.

* `IsolationManager` cleanup of committed transactions no longer relies on a `timeline`, but just the current set of transactions and the state they are in. We also delete the single concurrent hash map of `CommitState`
* Introduce three maps: `uncommitted`, `committing`, and `committed`. These allow us to optimise the single-threaded lock duration in the isolation manager, and implement a simplified cleanup mechanism.

### DIFF
--- a/database/CoreDatabase.java
+++ b/database/CoreDatabase.java
@@ -671,6 +671,7 @@ public class CoreDatabase implements TypeDB.Database {
          * and correct them if we have enough information to do so.
          */
         protected void correctMiscounts() {
+            if (!databaseMgr.isOpen()) return;
             try (CoreTransaction.Data txn = session.transaction(WRITE)) {
                 if (mayCorrectMiscounts(txn)) cache.incrementStatisticsVersion();
             }
--- a/database/CoreDatabase.java
+++ b/database/CoreDatabase.java
@@ -27,13 +27,13 @@ import com.vaticle.typedb.core.common.iterator.FunctionalIterator;
 import com.vaticle.typedb.core.common.parameters.Arguments;
 import com.vaticle.typedb.core.common.parameters.Options;
 import com.vaticle.typedb.core.concurrent.executor.Executors;
-import com.vaticle.typedb.core.encoding.key.Key;
-import com.vaticle.typedb.core.graph.TypeGraph;
 import com.vaticle.typedb.core.encoding.Encoding;
+import com.vaticle.typedb.core.encoding.iid.VertexIID;
+import com.vaticle.typedb.core.encoding.key.Key;
 import com.vaticle.typedb.core.encoding.key.KeyGenerator;
 import com.vaticle.typedb.core.encoding.key.StatisticsKey;
+import com.vaticle.typedb.core.graph.TypeGraph;
 import com.vaticle.typedb.core.graph.edge.ThingEdge;
-import com.vaticle.typedb.core.encoding.iid.VertexIID;
 import com.vaticle.typedb.core.graph.vertex.AttributeVertex;
 import com.vaticle.typedb.core.graph.vertex.ThingVertex;
 import com.vaticle.typedb.core.graph.vertex.TypeVertex;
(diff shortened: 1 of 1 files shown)


database/CoreDatabase.java
```diff
@@ -495,15 +492,15 @@ public void delete() {
...
 
         private enum CommitState {UNCOMMITTED, COMMITTING, COMMITTED}
 
         IsolationManager() {
-            this.cleanupRunning = new AtomicBoolean(false);
-            this.commitStates = new ConcurrentHashMap<>();
-            this.commitTimeline = new ConcurrentSkipListMap<>();
+            cleanupRunning = new AtomicBoolean(false);
+            commitStates = new ConcurrentHashMap<>();
+            approximateLastCommitSnapshot = new AtomicLong(0);
```
flyingsilverfin, line 503: For cleanup, we either need to know the last uncommitted transaction that means we can delete everything before that, OR we need to know the last commit snapshot that we can delete up to (because at any point we can open a new transaction that cannot be cleaned up from the isolation manager)

database/CoreDatabase.java
```diff
@@ -556,24 +551,19 @@ void closed(CoreTransaction.Data txn) {
...
-                                txn.delete();
-                                commitStates.remove(txn);
-                            }
-                            return null;
-                        })
-                );
+                long lastCommittedSnapshot = approximateLastCommitSnapshot.get();
+                Long cleanupUntil = oldestUnCommittedSnapshot().orElse(lastCommittedSnapshot + 1);
+                commitStates.forEach((txn, state) -> {
+                    if (txn.snapshotEnd().isPresent() && txn.snapshotEnd().get() < cleanupUntil) {
```
flyingsilverfin, line 557: we delete any transaction that ended before the cleanup target

database/CoreDatabase.java
```diff
@@ -494,39 +492,39 @@ public void delete() {
...
 
-        private enum CommitState {UNCOMMITTED, COMMITTING, COMMITTED}
-
         IsolationManager() {
-            this.cleanupRunning = new AtomicBoolean(false);
-            this.commitStates = new ConcurrentHashMap<>();
-            this.commitTimeline = new ConcurrentSkipListMap<>();
+            uncommitted = new ConcurrentSet<>();
+            committing = new ConcurrentSet<>();
+            committed = new ConcurrentSet<>();
```
flyingsilverfin, line 503: collapse the `CommitState` into three sparate maps

database/CoreDatabase.java
```diff
@@ -494,39 +492,39 @@ public void delete() {
...
+            uncommitted.add(transaction);
         }
 
         public Set<CoreTransaction.Data> validateOverlappingAndStartCommit(CoreTransaction.Data txn) {
             Set<CoreTransaction.Data> transactions;
             synchronized (this) {
                 transactions = commitMayConflict(txn);
                 transactions.forEach(other -> validateIsolation(txn, other));
-                commitStates.put(txn, CommitState.COMMITTING);
+                committing.add(txn);
```
flyingsilverfin, line 516: almost move from the earlier-stage man into the higher-level stage map

database/CoreDatabase.java
```diff
@@ -494,39 +492,39 @@ public void delete() {
...
         private Set<CoreTransaction.Data> commitMayConflict(CoreTransaction.Data txn) {
             if (!txn.dataStorage.hasTrackedWrite()) return set();
-            return iterate(commitStates.entrySet())
-                    .filter(e -> e.getValue() == CommitState.COMMITTING ||
-                            (e.getValue() == CommitState.COMMITTED && e.getKey().snapshotEnd().get() > txn.snapshotStart()))
-                    .map(Map.Entry::getKey)
-                    .toSet();
+            Set<CoreTransaction.Data> mayConflict = new HashSet<>(committing);
+            iterate(committed).filter(committed -> committed.snapshotEnd().get() > txn.snapshotStart())
+                    .forEachRemaining(mayConflict::add);
```
flyingsilverfin, line 526: this should be faster than the old way because we don't have to filter out all the uncommitted transactions at all, we only need to actually check the committed transaction age in the filter

database/CoreDatabase.java
```diff
@@ -540,51 +538,47 @@ private void validateIsolation(CoreTransaction.Data txn, CoreTransaction.Data ma
...
-                            for (CoreTransaction.Data txn : txns) {
-                                txn.delete();
-                                commitStates.remove(txn);
-                            }
-                            return null;
-                        })
-                );
+                long lastCommittedSnapshot = newestCommittedSnapshot();
+                Long cleanupUntil = oldestUncommittedSnapshot().orElse(lastCommittedSnapshot + 1);
+                committed.forEach(txn -> {
```
flyingsilverfin, line 557: cleanup is simplified by to only deleting the `committed` transactions up to a certain age

haikalpribadi (approved):

PR #6707, opened 2022-11-24
Optimise traversal over non-retrieved variables

flyingsilverfin:

Query plans which retrieve a tail sequence of vertices to visit, which are not in the user's query filter, can short-circuit their backtracking after a full answer is found. This means only permutations of vertices up to the last retrieved vertex are generated. This can dramatically speed up large queries in which only a few variables are retrieved. This also noticably speeds up rule validation, where only a few variables of the `when` need to be retrieved.

To illustrate the implementation: given ordered traversal variables [A, B, C, D, E], of which only [A, C] are retrieved by the query filter, after finding a successful permutation for all vertices D and E can immediately backtrack up to C, since they are not in the query filter.

Code changes:
* Implement optimised backtracking through un-retrieved vertices in the traversal engine

### DIFF
--- a/traversal/scanner/GraphIterator.java
+++ b/traversal/scanner/GraphIterator.java
@@ -75,7 +75,7 @@ public class GraphIterator extends AbstractFunctionalIterator<VertexMap> {
     private Direction direction;
     private IteratorState iteratorState;
 
-    private enum Direction {TRAVERSE, REVISIT}
+    private enum Direction {TRAVERSE, REVISIT_ALL, REVISIT_RETRIEVED}
 
     private enum IteratorState {INIT, EMPTY, FETCHED, COMPLETED}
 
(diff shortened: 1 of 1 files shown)


traversal/scanner/GraphIterator.java
```diff
@@ -206,28 +206,51 @@ private void initialiseStart() {
     private void initialiseEnds() {
         assert toTraverse.isEmpty();
         toRevisit.addAll(procedure.endVertices());
-        direction = Direction.REVISIT;
+        direction = Direction.REVISIT_RETRIEVED;
```
flyingsilverfin, line 209: `initialiseEnds` is called when a full answer is found and we look for the next complete answer

traversal/scanner/GraphIterator.java
```diff
@@ -206,28 +206,51 @@ private void initialiseStart() {
...
             if (direction == Direction.TRAVERSE) {
                 toTraverse.remove(vertex = toTraverse.first());
                 traverse(vertex);
-            } else {
+            } else if (direction == Direction.REVISIT_ALL) {
                 toRevisit.remove(vertex = toRevisit.last());
                 revisit(vertex);
+            } else {
+                toRevisit.remove(vertex = toRevisit.last());
+                revisitRetrieved(vertex);
```
flyingsilverfin, line 224: either we are in the revisiting any vertex state, or in the revisiting retrieved vertices only state

traversal/scanner/GraphIterator.java
```diff
@@ -206,28 +206,51 @@ private void initialiseStart() {
...
+            direction = Direction.REVISIT_ALL;
+        }
+    }
+
+    private void revisit(ProcedureVertex<?, ?> procedureVertex) {
+        toTraverse.add(procedureVertex);
+        direction = Direction.TRAVERSE;
+    }
+
+    private void revisitRetrieved(ProcedureVertex<?, ?> vertex) {
```
flyingsilverfin, line 247: retrieved vertex revisiting redirects into regular `revisit` if the variable is retrieved, otherwise immediately fails the vertex search for the vertex and backtracks

haikalpribadi (approved):

PR #6712, opened 2022-12-07
Shutdown gracefully on unexpected server exception

flyingsilverfin:

To prevent zombie or broken server states due to unexpected exceptions, we no longer catch and swallow unexpected exceptions. Instead, we shutdown the server. The catch-and-swallow behaviour is dangerous because 1) an uncaught exception means a thread is also killed somewhere, and could put the server into a zombie or otherwise broken state, and specifically can cause 2) zombie state by catching an OOM exception which would stall the server instead of letting it die. 

* Allow TypeDB to exit if there's an unexpected exception by removing the uncaught exception handler

### DIFF
--- a/common/exception/ErrorMessage.java
+++ b/common/exception/ErrorMessage.java
@@ -42,7 +42,7 @@ public abstract class ErrorMessage extends com.vaticle.typedb.common.exception.E
         public static final Server EXITED_WITH_ERROR =
                 new Server(3, "Exited with error.");
         public static final Server UNCAUGHT_EXCEPTION =
-                new Server(4, "Uncaught exception thrown at thread '%s'.");
+                new Server(4, "Uncaught exception thrown at thread '%s':\n'%s'");
         public static final Server CONFIG_FILE_NOT_FOUND =
                 new Server(5, "Could not find/read the configuration file '%s'.");
         public static final Server UNRECOGNISED_CLI_COMMAND =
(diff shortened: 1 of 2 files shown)


flyingsilverfin:
@lolski  is this the kind of thing you were thinking about?

@haikalpribadi  do you know why we didn't use this approach in the first place?

server/TypeDBServer.java
```diff
@@ -104,7 +104,10 @@ protected TypeDBServer(CoreConfig config, boolean debug, Factory factory, CoreLo
         databaseMgr = factory.databaseManager(options);
         server = rpcServer();
         Thread.setDefaultUncaughtExceptionHandler(
-                (t, e) -> logger().error(UNCAUGHT_EXCEPTION.message(t.getName() + ": " + e.getMessage()), e)
+                (t, e) -> {
```
lolski, line 107: Rather than killing the process and exiting with code 1, it would be more elegant if we can simply propagate the uncaught exception to the main thread, and let the JVM exit in the most appropriate way.

Would it be possible to do so, @flyingsilverfin ?

flyingsilverfin: you can't propagate from this handler to the main thread

flyingsilverfin: we decided just to delete the handler

haikalpribadi: Why not exit the whole server gracefully, @flyingsilverfin ? Log the error, print the stack trace, then gracefully shutdown? Surely that has to be the better way then crash right there and then?

common/exception/ErrorMessage.java
```diff
@@ -42,7 +42,7 @@ public static class Server extends ErrorMessage {
         public static final Server EXITED_WITH_ERROR =
                 new Server(3, "Exited with error.");
         public static final Server UNCAUGHT_EXCEPTION =
-                new Server(4, "Uncaught exception thrown at thread '%s'.");
```
flyingsilverfin, line 45: we were trying to add the exception before but the string formatting would've not matched the second argument

server/TypeDBServer.java
```diff
@@ -209,7 +213,7 @@ protected void serve() {
     }
 
     @Override
-    public void close() {
+    public synchronized void close() {
```
flyingsilverfin, line 216: now that this is called from two threads, it must be synchronized

haikalpribadi (approved):

PR #6717, opened 2022-12-13
Static server constructor with earlier logging configuration

flyingsilverfin:

We introduce a static TypeDB server constructor in which the server's logging is configured ahead of time, instead of in the constructor. This allows TypeDB Cluster have more freedom about when the logging is configured.

* introduce `TypeDBServer.create()` static constructor
* move logging configuration into the static constructor, and make it protected to be accessed in cluster's server constructor
* remove unneeded argument to TypeDBServer

### DIFF
--- a/server/TypeDBServer.java
+++ b/server/TypeDBServer.java
@@ -71,7 +71,6 @@ public class TypeDBServer implements AutoCloseable {
 
     protected final Factory factory;
     protected final CoreConfig config;
-    protected final CoreLogback logback;
     protected final CoreDatabaseManager databaseMgr;
     protected final io.grpc.Server server;
     protected final boolean debug;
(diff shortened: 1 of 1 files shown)


server/TypeDBServer.java
```diff
@@ -71,24 +71,28 @@ public class TypeDBServer implements AutoCloseable {
 
     protected final Factory factory;
     protected final CoreConfig config;
-    protected final CoreLogback logback;
```
flyingsilverfin, line 74: this is an uneeded field, it is just required to configure the logger which we can do statically

server/TypeDBServer.java
```diff
@@ -71,24 +71,28 @@ public class TypeDBServer implements AutoCloseable {
...
     protected final CoreDatabaseManager databaseMgr;
     protected final io.grpc.Server server;
     protected final boolean debug;
     protected TypeDBService typeDBService;
 
-    private TypeDBServer(CoreConfig config, boolean debug) {
-        this(config, debug, new CoreFactory(), new CoreLogback());
+    private static TypeDBServer create(CoreConfig config, boolean debug) {
```
flyingsilverfin, line 79: new setatic constructor which configures the logging as well

server/TypeDBServer.java
```diff
@@ -130,11 +134,6 @@ private void verifyDataDir() {
         }
     }
 
-    private static void configureLogging(CoreLogback logback, CoreConfig config) {
```
flyingsilverfin, line 133: reorganise to be near usage

PR #6718, opened 2022-12-19
Update brew formula

dmitrii-ubskii:

We update the formula template used for our own homebrew tap to be more in line with recent changes to the homebrew-core formula.

* Use server configuration to set the persistent data and log diretories;
* depend on the most recent version of jdk as opposed to strictly 11.

### DIFF
--- a/config/brew/typedb.rb
+++ b/config/brew/typedb.rb
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
+# IMPORTANT: any changes to the formula should be propagated to Homebrew/homebrew-core
 class Typedb < Formula
   desc "Strongly-typed database with a rich and logical type system"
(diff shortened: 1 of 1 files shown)


config/brew/typedb.rb
```diff
@@ -15,27 +15,23 @@
...
+  depends_on "openjdk"
 
   def install
     libexec.install Dir["*"]
-    setup_directory "data"
-    setup_directory "logs"
+    mkdir_p var/"typedb/data"
+    inreplace libexec/"server/conf/config.yml", "server/data", var/"typedb/data"
+    mkdir_p var/"log/typedb"
+    inreplace libexec/"server/conf/config.yml", "server/logs", var/"typedb/logs"
```
flyingsilverfin, line 33: these paths dont match up?

flyingsilverfin (approved):

PR #6720, opened 2022-12-20
Open import-export modules for extension

flyingsilverfin:

We extend the core Server and Migrators for extension in TypeDB Cluster.

* rename `ServerSubcommand*` classes to `CoreSubcommand*` to reflect our existing naming convention where overridden components are called `Core*` or `Cluster*` instead of `Server*` which are shared components
* open Config and Parser files for extension
* refactor `MigratorClient` to accept an injected `stub` to allow Cluster to inject a stub with call credentials
* rename `MigratorClient` to `CoreMigratorClient` to matching our naming convention
* move a testing package into a subdirectory matching the tested classes to access package-private methods
* fix a bug in subcommand parsing logic & add missing `toString()` method


server/TypeDBServer.java
```diff
@@ -233,18 +237,18 @@ public static void main(String[] args) {
             printASCIILogo();
 
             CoreConfigParser configParser = new CoreConfigParser();
-            ArgsParser<ServerSubcommand> argsParser = new ArgsParser<ServerSubcommand>()
-                    .subcommand(new ServerSubcommandParser.Server(configParser))
-                    .subcommand(new ServerSubcommandParser.Import())
-                    .subcommand(new ServerSubcommandParser.Export());
-            Optional<ServerSubcommand> subcmd = argsParser.parse(args);
+            ArgsParser<CoreSubcommand> argsParser = new ArgsParser<CoreSubcommand>()
```
flyingsilverfin, line 240: We follow the existing naming convention where anything called `Server` is shared by Core and Cluster, otherwise we use the `Core` and `Cluster` prefixes. This change is caused by Cluster needing to extend command-line commands and not use shared bootup commands anymore

server/parameters/CoreSubcommandParser.java
```diff
@@ -34,34 +34,35 @@
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Server.CLI_OPTION_UNRECOGNISED;
 import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
 
-public class ServerSubcommandParser {
+public class CoreSubcommandParser {
 
-    private static void validateRequiredOptions(Set<OptionParser> requiredParsers, Set<Option> options) {
+    protected static void validateRequiredOptions(Set<OptionParser> requiredParsers, Set<Option> options) {
```
flyingsilverfin, line 39: allow usage within the `Cluster` variant of this class

server/parameters/CoreSubcommandParser.java
```diff
@@ -34,34 +34,35 @@
...
             }
         });
     }
 
-    private static void validateUnrecognisedOptions(Set<OptionParser> recognisedParsers, Set<Option> options) {
-        recognisedParsers.forEach(parser -> {
-            if (iterate(options).noneMatch(option -> option.name().equals(parser.name()))) {
-                throw TypeDBException.of(CLI_OPTION_UNRECOGNISED, parser);
+    protected static void validateUnrecognisedOptions(Set<OptionParser> recognisedParsers, Set<Option> options) {
+        options.forEach(option -> {
```
flyingsilverfin, line 48: this was a bug - we throwing `UNRECOGNISED` whenever there was a parser that was not in the options, rather than the other way around.

server/parameters/CoreSubcommandParser.java
```diff
@@ -34,34 +34,35 @@
...
+    public static class Server extends SubcommandParser<CoreSubcommand.Server> {
+
+        public static final String[] tokens = new String[]{};
+        public static final String description = "Run TypeDB server";
 
-        private static final OptionParser.Flag debug = new OptionParser.Flag("debug", "Run server in debug mode.");
-        private static final OptionParser.Flag help = new OptionParser.Flag("help", "Print help menu.");
-        private static final OptionParser.Flag version = new OptionParser.Flag("version", "Print version number.");
-        private static final OptionParser.Path configPath =
+        public static final OptionParser.Flag debug = new OptionParser.Flag("debug", "Run server in debug mode.");
```
flyingsilverfin, line 60: we make these `public` so that Cluster can re-use them without extending the class directly.

server/parameters/CoreSubcommandParser.java
```diff
@@ -71,24 +72,24 @@ public Server(CoreConfigParser configParser) {
...
             CoreConfig config = configPath
                     .map(path -> CoreConfigFactory.config(path, configOptions, configParser))
                     .orElseGet(() -> CoreConfigFactory.config(configOptions, configParser));
-            return new ServerSubcommand.Server(debug.parse(auxOptions), help.parse(auxOptions),
+            return new CoreSubcommand.Server(debug.parse(auxOptions), help.parse(auxOptions),
                     version.parse(auxOptions), config);
         }
 
-        private Set<Option> findAuxiliaryOptions(Set<Option> options) {
+        public static Set<Option> findAuxiliaryOptions(Set<OptionParser> auxiliaryParsers, Set<Option> options) {
```
flyingsilverfin, line 86: make `static` so that the Cluster implementation can reference without re-impelementing

server/parameters/CoreSubcommandParser.java
```diff
@@ -99,27 +100,27 @@ public List<Help> helpList() {
...
     }
 
-    public static class Import extends SubcommandParser<ServerSubcommand.Import> {
+    public static class Import extends SubcommandParser<CoreSubcommand.Import> {
 
         public static final String[] tokens = new String[]{"import"};
         public static final String description = "Run TypeDB import.";
 
-        private static final OptionParser.String database =
+        public static final OptionParser.String database =
```
flyingsilverfin, line 108: same as above - allow referencing from within the equivalent Cluster implementation

server/parameters/util/OptionParser.java
```diff
@@ -58,6 +58,11 @@ FunctionalIterator<Option> matchingOptions(Set<Option> options) {
         return iterate(options).filter(opt -> opt.name().equals(name()));
     }
 
+    @Override
+    public java.lang.String toString() {
```
flyingsilverfin, line 62: if we end up throwing an exception with an `OptionParser` in it - we should have a useful `toString()`, which should just be the help entry for that parser

server/test/parameters/CoreConfigTest.java
```diff
@@ -16,14 +16,11 @@
  *
  */
 
-package com.vaticle.typedb.core.server.test;
+package com.vaticle.typedb.core.server.parameters;
```
flyingsilverfin, line 19: we move the `Config` testing into a mirror package (eg. `//server/test` to `//server/test/parameters`) so that we can tighten some access methods to be `package-private` instead of `public`.

server/test/BUILD
```diff
@@ -29,7 +29,7 @@ package(default_visibility = ["//visibility:private",])
 host_compatible_java_test(
     name = "test-configuration",
     srcs = [
-        "CoreConfigTest.java",
+        "parameters/CoreConfigTest.java",
```
flyingsilverfin, line 32: we have relocated the test contents into a package that mirrors the source to be able to access package-private methods

server/parameters/CoreSubcommand.java
```diff
@@ -25,33 +25,33 @@
...
         return false;
     }
 
-    public Export asExport() {
+    default Export asExport() {
         throw TypeDBException.of(ILLEGAL_CAST, className(getClass()), className(Export.class));
     }
 
-    public static class Server extends ServerSubcommand {
+    class Server implements CoreSubcommand {
```
flyingsilverfin, line 54: no longer need to be public

lolski (approved):

PR #6732, opened 2023-01-16
Introduce expressions and computed value variables

### LINKED ISSUE #6654: Arithmetic
## Problem to Solve
Support basic arithmetic (or more generally, value-functions) in queries.

## Current Workaround
None.

## Proposed Solution

{ Please describe the solution you would like TypeDB to provide, to solve the problem above. }

### Major design concerns
* Currently, new concepts can only be created in the 'then' of a rule. Doing an assignment in a conjunction opens the door to  concepts being created in the `when` of a rule, leading to evaluation of the model becoming hard.

{ Any additional information, including logs or screenshots if you have any. }

krishnangovindraj:

Implement support for expressions and value variables. The user-facing changes that drive this change are outlined in https://github.com/vaticle/typeql/pull/260.

In particular TypeDB now supports the evaluation and return of Values, which are distinct from the previously existing data instances Entity/Relation/Attribute. Values are never materialised permanently, and only used within the scope of a particular query or rule. They are defined as an actual inner value, plus one of the defined TypeDB value types: `long`, `double`, `boolean`, `string` or `datetime`.

Rules also now allow using of value variables to infer new attribute values:
```
define
rule computation-cost-dollars-to-pounds: when {
  $c isa computation, has cost-dollars $c;
  $_ isa exchange-rate, has name "dollar-pound", has rate $rate;
  ?pounds = $rate * $c;
} then {
  $c has cost-pounds ?pounds;
};
```

Querying for an attribute `match $x isa thing;` will then also trigger this inference. Note that accidentally creating infinite reasoning is much easier now: if implement `n = n + 1` in a rule, **it will run as long as the transaction is alive** - and likely cause an OOM. Future work will offer more advanced validation to protect against these cases.

Closes #6654.

Patterns and constraints:
Introduce:
* Value(Concept) (`concept.value.Value`)
  * A (virtual) concept defined by a ValueType and a raw value of the said type.
  * Implemented `ValueImpl` and subclasses for each ValueType.  
* ValueVariable  (`pattern.variable.ValueVariable`)
  * A new class of variables which unifies with `ValueConcept` and holds `ValueConstraint`
* `ValueConstraint` (`pattern.constraint.ValueConstraint) and subclasses:
  * `AssignmentConstraint`: assigns a ValueVariable with the result of some TypeQL `Expression`
  * `PredicateConstraint`: Adds a predicate constraint on the ValueVariable - It can be an (inequality) constraint where the value is compared against one of: 
     * A constant of the supported valueTypes
     * A `ThingVariable` (which has to be an attribute to have a value)
     * Another `ValueVariable`
* A `ValueVertex` similar to Value Concept 
* An IID definition for `ValueVertex` in `encoding.iid.VertexIID`
  * Defined by  Prefix + ValueType + EncodedValue
* `ExpressionCompiler` as a sibling to TypeResolution
    * Convertes TypeQL expressions into executable traversal expressions with unambiguous value types

Refactor:
* Introduce a `vertex.Value` interface implemented by ThingVertex and ValueVertex
   * supports reading the value for comparisons in predicates.
   * supports reading the as input in expressions
* Rename `ThingConstraint.ValueConstraint` to be `ThingConstraint.PredicateConstraint`
   * Supports comparing a ThingVariable to a ValueVariable .

### Traversal
* Introduce the Value subclasses for StructureVertex, PlannerVertex, ProcedureVertex
* Introduce `Argument` edge to connect variables on the RHS of an assignment to the variable being assigned to (i.e. on the LHS)
* Update the planner to enforce `Argument` edges are always traversed forwards (i.e., all vertices of the RHS are evaluated before the LHS)
* Update PredicateEdge to be able to deal with both ValueVertex and ThingVertex
* Introduce the implementation of every operator / function in `traversal.expression`


traversal/test/ProcedurePermutator.java
```diff
@@ -45,11 +46,29 @@ public static FunctionalIterator<GraphProcedure> generate(Structure structure) {
...
+        for (Identifier id : idPermutation) {
+            StructureVertex<?> vertex = vertices.get(id);
+            if (vertex.isValue()) {
+                if (!iterate(vertex.ins()).filter(StructureEdge::isInput).allMatch(inputEdge -> seen.contains(inputEdge.from().id()))) {
+                    return false;
+                }
+            }
+            seen.add(id);
+        }
+        return true;
```
krishnangovindraj, line 72: Fix to support value variables

traversal/structure/Structure.java
```diff
@@ -82,11 +90,21 @@ public void equalEdge(StructureVertex<?> from, StructureVertex<?> to) {
         recordEdge(edge);
     }
 
-    public void predicateEdge(StructureVertex.Thing from, StructureVertex.Thing to, Predicate.Variable predicate) {
+    public void predicateEdge(StructureVertex<?> from, StructureVertex<?> to, Predicate.Variable predicate) {
```
krishnangovindraj, line 93: Had to weaken them to support Thing-Value comparisons

traversal/procedure/ProcedureVertex.java
```diff
@@ -132,6 +145,8 @@ public String toString() {
         return str;
     }
 
+    public abstract Forwardable<Vertex<?, ?>, Order.Asc> intersect(Map<ProcedureEdge<?, ?>, Forwardable<Vertex<?, ?>, Order.Asc>> iterators, Traversal.Parameters parameters, Order.Asc order);
```
krishnangovindraj, line 148: Support overloading the intersection operator.

Made sense at the time as a way to support "partial specifications" of the to-vertex, rather than an exact vertex identity.
e.g. for a virtual/functional vertex such as a ValueVertex, the inputs only partially specify the final identity of the vertex. All the specifications from the input-vertices together uniquely determine the output vertex.

This may be the wrong approach when there are multiple possible output vertices for a given specification.

traversal/procedure/ProcedureVertex.java
```diff
@@ -643,4 +670,87 @@ public ProcedureVertex.Type asType() {
...
+            return true;
+        }
+
+        @Override
+        public Value asValue() {
+            return this;
+        }
+
+        @Override
+        public Forwardable<Vertex<?, ?>, Order.Asc> intersect(Map<ProcedureEdge<?, ?>, Forwardable<Vertex<?, ?>, Order.Asc>> iterators, Traversal.Parameters parameters, Order.Asc order) {
```
krishnangovindraj, line 712: Value vertex implementation of intersect.
Takes the assignment inputs, makes an input map, evaluates the value
Collects the predicate inputs, evaluates each against the result of the evaluation

traversal/procedure/ProcedureVertex.java
```diff
@@ -643,4 +670,87 @@ public ProcedureVertex.Type asType() {
...
+            ValueVertexImpl<?> tmpResult;
+            try {
+                tmpResult = props().expression().evaluateToVertex(vertexMap);
+            } catch (TypeDBException | TypeDBCheckedException e) {
+                throw evaluationError(vertexMap, e);
+            }
+
+            ValueVertexImpl<?> result = tmpResult;
+            if (iterate(props().predicates()).allMatch(predicate -> checkPredicate(predicate, result, parameters.getValues(id().asVariable(), predicate))) &&
+                    iterate(variablePredicates.entrySet()).allMatch(entry -> checkPredicate(entry.getKey().predicate(), entry.getValue(), result))) {
```
krishnangovindraj, line 734: constant predicates & variable predicates resp.

traversal/procedure/ProcedureEdge.java
```diff
@@ -244,24 +283,31 @@ public boolean equals(Object o) {
...
+                ProcedureVertex.Thing to = this.to.asThing();
+                Forwardable<? extends ThingVertex, Order.Asc> toIter;
+                if (to.props().hasIID()) toIter = to.iterateAndFilterFromIID(graphMgr, params, ASC);
+                else toIter = to.iterateAndFilterFromTypes(graphMgr, params, ASC);
+
+                return toIter.filter(toVertex -> predicate.apply(fromValue, toVertex.asAttribute()));
+            } else if (to.isValue()) {
+                ValueVertex<?> predicateSpec = ValueVertexImpl.ValueVertexSpec.of(fromValue);
+                return SortedIterators.Forwardable.iterateSorted(ASC, predicateSpec);
+            } else throw TypeDBException.of(ILLEGAL_STATE); // Use predicate for now
```
krishnangovindraj, line 299: Thing & value predicates behave differently.

krishnangovindraj (commented):
some comments

encoding/Expression.java
```diff
@@ -0,0 +1,200 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
```
krishnangovindraj, line 18: Declares supported functions  by mapping symbol & arguments types -> static instance (which in turn declares the return type). 
Used during expression resolution to ensure infer return types of intermediate results.

encoding/Expression.java
```diff
@@ -0,0 +1,200 @@
...
+                Class.forName(innerClass.getName(), true, innerClass.getClassLoader());
+            } catch (ClassNotFoundException e) {
+                throw TypeDBException.of(e);
+            }
+        }
+    }
+
+    private static void register(Operation op) {
+        if (null != Expression.OPS.put(new Pair<>(op.symbol, new Pair<>(op.arg1Type, op.arg2Type)), op))
+            throw TypeDBException.of(DUPLICATE_EXPRESSION_REGISTERED, op);
```
krishnangovindraj, line 61: Ensure only one declaration per signature.

encoding/Expression.java
```diff
@@ -0,0 +1,200 @@
...
+    private static void register(ThingVar thingVar) {
+        if (null != Expression.THING_VARS.put(thingVar.type, thingVar))
+            throw TypeDBException.of(DUPLICATE_EXPRESSION_REGISTERED, thingVar);
+    }
+
+    private static void register(ValVar valVar) {
+        if (null != Expression.VAL_VARS.put(valVar.type, valVar))
+            throw TypeDBException.of(DUPLICATE_EXPRESSION_REGISTERED, valVar);
+    }
+
```
krishnangovindraj, line 83: Helpers to resolve the right instance from the signature.

graph/vertex/AttributeVertex.java
```diff
@@ -95,7 +85,7 @@
     /**
      * An vertex wrapper that sorts based on its value, then type -- as opposed to type, then value.
      */
-    interface Value<VALUE> extends AttributeVertex<VALUE> {
+    interface Sort<VALUE> extends AttributeVertex<VALUE> {
```
krishnangovindraj, line 88: Can we rename this to something less generic than Value? We only use this for sorting.

graph/vertex/ValueHolder.java
```diff
@@ -0,0 +1,50 @@
+/*
```
krishnangovindraj, line 1: Common interface implemented by Attribute & Value vertices. Used in Predicates & expressions to access the value.

graph/vertex/impl/AttributeVertexImpl.java
```diff
@@ -101,18 +101,18 @@ public AttributeVertex.Write<VALUE> toWrite() {
         }
 
         @Override
-        public boolean isValue() {
```
krishnangovindraj, line 104: Changes as result of renaming AttributeVertex.Value -> AttributeVertex.Sort.

krishnangovindraj (commented):
more comments.

dependencies/vaticle/repositories.bzl
```diff
@@ -27,8 +27,8 @@ def vaticle_dependencies():
 def vaticle_typeql():
     git_repository(
         name = "vaticle_typeql",
-        remote = "https://github.com/vaticle/typeql",
-        tag = "2.14.0",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typeql
+        remote = "https://github.com/krishnangovindraj/typeql",
+        commit = "0f19624892839460fd66db843cf34a131113bb64"
```
krishnangovindraj, line 31: These need to be pointed back.

krishnangovindraj (commented):
Add review so I don't forget dependencies.

concept/answer/ConceptMap.java
```diff
@@ -219,7 +228,25 @@ public static Comparator create(Modifiers.Sorting sorting) {
...
+                            } else if (val1.isDouble() || val2.isDouble()) {
+                                Double double1 = val1.isLong() ? val1.asLong().value() : val1.asDouble().value();
+                                Double double2 = val2.isLong() ? val2.asLong().value() : val2.asDouble().value();
+                                return Encoding.ValueType.DOUBLE.comparator().compare(double1, double2);
+                            } else if (val1.isDateTime()) {
+                                return Encoding.ValueType.DATETIME.comparator().compare(val1.asDateTime().value(), val2.asDateTime().value());
+                            } else {
+                                throw TypeDBException.of(ILLEGAL_STATE);
+                            }
+                        } else if (concept1.isAttribute() && concept2.isAttribute()) {
```
krishnangovindraj, line 249: Adds the case of value variables. 
(The diff is a bit unintuitive because I added Values as the first case)

concept/value/impl/ValueImpl.java
```diff
@@ -0,0 +1,264 @@
+/*
```
krishnangovindraj, line 1: Very boring implementation

database/CoreDatabaseManager.java
```diff
@@ -49,6 +50,7 @@ public class CoreDatabaseManager implements TypeDB.DatabaseManager {
         RocksDB.loadLibrary();
         Loader.loadNativeLibraries();
         ErrorMessage.loadConstants();
+        Expression.loadConstants();
```
krishnangovindraj, line 53: Creates the directory of signature -> static instance.

encoding/Encoding.java
```diff
@@ -99,7 +99,8 @@ public boolean isIncrement() {
     public enum Status {
         BUFFERED(0),
         COMMITTED(1),
-        PERSISTED(2);
+        PERSISTED(2),
+        VALUE(3);
```
krishnangovindraj, line 103: Create a new prefix for Value Vertices, though they should never be stored.

encoding/iid/VertexIID.java
```diff
@@ -422,4 +427,173 @@ public DateTime asDateTime() {
...
+            @Override
+            public Boolean asBoolean() {
+                return this;
+            }
+
+            @Override
+            public java.lang.Boolean value() {
+                return byteToBoolean(bytes.get(VALUE_INDEX));
+            }
+        }
```
krishnangovindraj, line 510: The encoded IID remains the source of truth.

encoding/iid/VertexIID.java
```diff
@@ -422,4 +427,173 @@ public DateTime asDateTime() {
...
+
+        public static class String extends Value<java.lang.String> {
+
+            private String(ByteArray bytes) {
+                super(STRING, bytes);
+            }
+
+            public String(java.lang.String value) throws TypeDBCheckedException {
+                this(encodeStringAsSorted(value, STRING_ENCODING));
+            }
```
krishnangovindraj, line 564: Exception thrown is if the encoding exceeds 64k

krishnangovindraj (commented):
some more comment

graph/vertex/impl/ValueVertexImpl.java
```diff
@@ -0,0 +1,308 @@
...
+                return new ValueVertexImpl.ValueVertexSpec<>(new VertexIID.Value.Long(valueVertex.asLong().value()), valueVertex.asLong());
+            else if (valueVertex.isDouble())
+                return new ValueVertexImpl.ValueVertexSpec<>(new VertexIID.Value.Double(valueVertex.asDouble().value()), valueVertex.asDouble());
+            else if (valueVertex.isString()) {
+                try { return new ValueVertexImpl.ValueVertexSpec<>(new VertexIID.Value.String(valueVertex.asString().value()), valueVertex.asString()); }
+                catch (TypeDBCheckedException e) { throw TypeDBException.of(ILLEGAL_STATE, e); }
+            } else if (valueVertex.isDateTime())
+                return new ValueVertexImpl.ValueVertexSpec<>(new VertexIID.Value.DateTime(valueVertex.asDateTime().value()), valueVertex.asDateTime());
+            else throw TypeDBException.of(ILLEGAL_STATE);
+        }
```
krishnangovindraj, line 286: This approach of assigning IIDs doesn't really make sense. 
Should I introduce a new class of VertexIIDs for VertexSpecs? Or we rethinking VertexSpecs altogether.

krishnangovindraj (commented):
Flag ValueVertexSpec IID for review

logic/Rule.java
```diff
@@ -723,12 +749,23 @@ public Materialisable materialisable(ConceptMap whenConcepts, ConceptManager con
...
+                    else throw TypeDBException.of(ILLEGAL_STATE);
+                }
+
+                private PredicateConstraint.Constant<?> predicateConstraintForValueAssignment(com.vaticle.typedb.core.concept.value.Value<?> value) {
+                    if (value.isBoolean()) return new PredicateConstraint.Constant.Boolean(attribute(), TypeQLToken.Predicate.Equality.EQ, value.asBoolean().value());
+                    else if (value.isLong()) return new PredicateConstraint.Constant.Long(attribute(), TypeQLToken.Predicate.Equality.EQ, value.asLong().value());
+                    else if (value.isDouble()) return new PredicateConstraint.Constant.Double(attribute(), TypeQLToken.Predicate.Equality.EQ, value.asDouble().value());
+                    else if (value.isString()) return new PredicateConstraint.Constant.String(attribute(), TypeQLToken.Predicate.Equality.EQ, value.asString().value());
+                    else if (value.isDateTime()) return new PredicateConstraint.Constant.DateTime(attribute(), TypeQLToken.Predicate.Equality.EQ, value.asDateTime().value());
+                    else throw TypeDBException.of(ILLEGAL_STATE);
```
krishnangovindraj, line 768: Converts the ValueConcept to the Equality.EQ constraint expected by the materialiser, as it would for Constant attribute inference.

encoding/Expression.java
```diff
@@ -0,0 +1,200 @@
...
+            this.argumentTypes = list(argumentTypes);
+            Expression.register(this);
+        }
+
+        public List<Encoding.ValueType<?>> argumentTypes() {
+            return argumentTypes;
+        }
+    }
+
+    public static class ThingVar extends Expression {
```
krishnangovindraj, line 161: I can do away with having a separate `ThingVar` and `ValueVar` if preferred - they make little difference at the traversal level.

traversal/scanner/GraphIterator.java
```diff
@@ -400,10 +397,9 @@ private void clearScopes() {
         }
 
         private Forwardable<Vertex<?, ?>, Order.Asc> createIteratorFromEdges() {
-            List<Forwardable<Vertex<?, ?>, Order.Asc>> iterators = new ArrayList<>();
-            procedureVertex.ins().forEach(edge -> iterators.add(branch(vertexTraversers.get(edge.from()).vertex(), edge)));
-            if (iterators.size() == 1) return iterators.get(0);
-            else return intersect(iterate(iterators), ASC);
+            Map<ProcedureEdge<?,?>, Forwardable<Vertex<?, ?>, Order.Asc>> iterators = new HashMap<>();
+            procedureVertex.ins().forEach(edge -> iterators.put(edge, branch(vertexTraversers.get(edge.from()).vertex(), edge)));
+            return procedureVertex.intersect(iterators, params, ASC);
```
flyingsilverfin, line 402: If the vertex is a Value procedure vertex:
          1. Find the expression input edges, get the input edges' values, compute the output value
          2. apply all remaining edges as predicates using isClosures instead of branch()

          else:
          1. open all the iterators with Branch
          2. intersect them all
          
          this should delete `ValueVertexSpec` and also remove the need for special intersection operations

flyingsilverfin (approved):

graph/vertex/AttributeVertex.java
```diff
@@ -50,21 +50,11 @@
...
-    boolean isBoolean();
-
-    boolean isLong();
-
-    boolean isDouble();
-
-    boolean isString();
-
-    boolean isDateTime();
+    Sort<VALUE> toSort();
```
flyingsilverfin, line 57: `isValueSortable`
`asValueSortable`
`toValueSortable`

common/exception/ErrorMessage.java
```diff
@@ -587,8 +623,10 @@ public static class RuleWrite extends ErrorMessage {
                 new RuleWrite(8, "The rule '%s' has a then clause '%s' that can never have answers in the current schema.");
         public static final RuleWrite RULE_THEN_INVALID_VALUE_ASSIGNMENT =
                 new RuleWrite(9, "The rule '%s' has a then clause with an invalid assignment of '%s' into a '%s'.");
+        public static final RuleWrite RULE_THEN_INVALID_VALUE_ASSIGNMENT_FROM_NON_VALUE_VARIABLE =
+                new RuleWrite(10, "The rule '%s' has a then clause which assigns from a variable '%s' which is not a value variable.");
```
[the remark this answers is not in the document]
krishnangovindraj: Deleted it instead. it was no longer valid.

traversal/scanner/GraphIterator.java
krishnangovindraj: Weren't we doing away with isClosure?

traversal/expression/Expression.java
```diff
@@ -0,0 +1,199 @@
...
+
+        @Override
+        default Expression.String asString() {
+            return this;
+        }
+    }
+
+    interface DateTime extends Expression<LocalDateTime> {
+        @Override
+        default ValueVertexImpl.DateTime evaluateToVertex(Map<Identifier, ValueHolder<?>> vertexMap) {
```
flyingsilverfin, line 180: Expressions should not compute Vertices as output directly, they should return values I think

traversal/expression/ExpressionFactory.java
```diff
@@ -0,0 +1,110 @@
...
+            Pair<ValueConstraint.Assignment.Expression, ValueConstraint.Assignment.Expression> operands = expr.asOperation().operands();
+            return instantiateOperation(resolvedExpressions.get(expr), build(operands.first(), resolvedExpressions), build(operands.second(), resolvedExpressions));
+        } else if (expr.isFunction()) {
+            List<Expression<?>> args = new ArrayList<>();
+            iterate(expr.asFunction().arguments()).map(argExpr -> build(argExpr, resolvedExpressions)).forEachRemaining(args::add);
+            return instantiateFunction(resolvedExpressions.get(expr), args);
+        } else throw TypeDBException.of(ILLEGAL_STATE);
+    }
+
+    public static Constant.Boolean instantiateConstant(java.lang.Boolean value) { return new Constant.Boolean(value); }
```
flyingsilverfin, line 73: you don't need `instantiate` on a static method, these are factory methods already

flyingsilverfin (commented):
I'm almost through, just need to look at the Expression package and Identifiers... please don't push until that's done so I can keep track :o 

This is a huge PR, so sorry it's taking so long!

encoding/iid/VertexIID.java
```diff
@@ -422,4 +426,173 @@ public DateTime asDateTime() {
...
+
+
+            @Override
+            public DateTime asDateTime() {
+                return this;
+            }
+
+            @Override
+            public java.time.LocalDateTime value() {
+                return bytes.view(VALUE_INDEX, bytes.length()).decodeSortedAsDateTime(TIME_ZONE_ID);
```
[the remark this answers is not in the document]
krishnangovindraj: Eep, I fixed it at the wrong place. Sorry.
It looks like the Attribute IIDs also use the second argument just to be explicit.
And there are places in the code where you need to specify the second argument as well, so I now think it's better to keep the end argument everywhere

common/exception/ErrorMessage.java
```diff
@@ -556,6 +566,30 @@ public static class TypeWrite extends ErrorMessage {
...
 
+    public static class Expression extends ErrorMessage {
+        public static final Expression UNRESOLVABLE_OPERATION =
+                new Expression(1, "The operation could not be resolved for the data-types: '%s' %s '%s'.");
+        public static final Expression UNRESOLVABLE_FUNCTION =
+                new Expression(2, "The function could not be resolved for the signature: '%s(%s)'.");
+        public static final Expression AMBIGUOUS_EXPRESSION_ARGUMENT_TYPE =
+                new Expression(3, "The variable '%s' can have multiple value types: [%s].");
+        public static final Expression EVALUATION_ERROR =
+                new Expression(4, "An error occured while evaluating an expression:\nExpression: '%s'; \nInput: '%s'.\nError: '%s'");
```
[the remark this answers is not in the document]
krishnangovindraj: Yes. Good idea

encoding/Expression.java
```diff
@@ -0,0 +1,200 @@
...
+import com.vaticle.typedb.core.common.exception.TypeDBException;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static com.vaticle.typedb.common.collection.Collections.list;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.DUPLICATE_EXPRESSION_REGISTERED;
+
+public abstract class Expression {
```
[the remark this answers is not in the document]
krishnangovindraj: Traversal isn't really the right place to have this.
But since we've now explicitly defined the available functions in TypeQL, I feel like this should be in TypeQL too.

graph/vertex/ValueHolder.java
krishnangovindraj: That breaks the symmetry a bit much doesn't it?

graph/vertex/impl/ValueVertexImpl.java
```diff
@@ -0,0 +1,254 @@
...
+        return new ValueVertexImpl.DateTime(new VertexIID.Value.DateTime(value), value);
+    }
+
+    public static ValueVertexImpl<?> from(AttributeVertex<?> attributeVertex) {
+        if (attributeVertex.isBoolean()) return ValueVertexImpl.of(attributeVertex.asBoolean().value());
+        else if (attributeVertex.isLong()) return ValueVertexImpl.of(attributeVertex.asLong().value());
+        else if (attributeVertex.isDouble()) return ValueVertexImpl.of(attributeVertex.asDouble().value());
+        else if (attributeVertex.isString()) {
+            try { return ValueVertexImpl.of(attributeVertex.asString().value()); }
+            catch (TypeDBCheckedException e) { throw TypeDBException.of(ILLEGAL_STATE, e); } // Should not happen
```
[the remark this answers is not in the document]
krishnangovindraj: Yes. Hence the illegal state.

pattern/variable/ValueVariable.java
```diff
@@ -0,0 +1,196 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import static com.vaticle.typedb.common.collection.Collections.set;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Pattern.REASSIGNING_VALUE_VARIABLE;
+
+public class ValueVariable extends Variable implements AlphaEquivalent<ValueVariable> {
+    private ExpressionConstraint assignedBy;
+    private final Set<Constraint> constraining; // What expressions are we an operand in?
```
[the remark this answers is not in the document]
krishnangovindraj: Yes, predicates too. I'll remove the comment

test/integration/logic/resolvable/AlphaEquivalenceTest.java
```diff
@@ -480,7 +480,7 @@ public void test_isa() {
     public void test_attribute() {
         schema("define number sub attribute, value long; other-number sub attribute, value long;");
         List<Concludable> concludables = iterate(
-                concludable("$a = 3"),
```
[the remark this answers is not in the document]
krishnangovindraj: Is AlphaEquivalence for expressions useful?

test/integration/logic/resolvable/UnifyHasConcludableTest.java
```diff
@@ -760,16 +760,16 @@ public void has_attribute_like_unifies_rule_has_var() {
...
+                pair(Identifier.Variable.namedConcept("x"), instanceOf("person")),
+                pair(Identifier.Variable.namedConcept("n"), instanceOf("age"))
         );
         unified = unifier.unUnify(concepts, new Unifier.Requirements.Instance(map()));
         assertFalse(unified.hasNext());
 
         // filter out invalid value
         concepts = map(
-                pair(Identifier.Variable.name("x"), instanceOf("person")),
-                pair(Identifier.Variable.name("n"), instanceOf("first-name", "reasoner@grakn.ai"))
```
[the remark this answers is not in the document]
krishnangovindraj: No, It should be illegal syntax at the moment to write anything that would lead a value variable to unify with a rule.

test/behaviour/typeql/TypeQLSteps.java
```diff
@@ -201,6 +203,11 @@ public void typeql_match_throws_exception(String typeQLQueryStatements) {
         assertThrows(() -> typeql_match(typeQLQueryStatements));
     }
 
+    @When("typeql match; throws exception containing {string}")
+    public void typeql_match_throws_exception(String exception, String typeQLQueryStatements) {
+        assertThrowsWithMessage(() -> typeql_match(typeQLQueryStatements), exception);
+    }
```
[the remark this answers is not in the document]
krishnangovindraj: I wrote some yes.
I couldn't find one which checked the exception message so I added it.

pattern/constraint/value/PredicateConstraint.java
```diff
@@ -0,0 +1,501 @@
...
+import static com.vaticle.typedb.common.util.Objects.className;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Pattern.INVALID_CASTING;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Pattern.MISSING_CONSTRAINT_VALUE;
+
+import static com.vaticle.typeql.lang.common.TypeQLToken.Char.QUOTE_DOUBLE;
+import static com.vaticle.typeql.lang.common.TypeQLToken.Char.SPACE;
+import static com.vaticle.typeql.lang.common.TypeQLToken.Predicate.Equality.EQ;
+
+public abstract class PredicateConstraint<T> extends ValueConstraint implements AlphaEquivalent<PredicateConstraint<?>> {
```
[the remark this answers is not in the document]
krishnangovindraj: Might work. Where would it sit?

pattern/constraint/value/ValueConstraint.java
```diff
@@ -0,0 +1,110 @@
...
+    public ValueConstraint(ValueVariable owner, Set<Variable> additionalVariables) {
+        this.owner = owner;
+        Set<Variable> vars = new HashSet<>(additionalVariables);
+        vars.add(owner);
+        this.variables = Collections.unmodifiableSet(vars);
+    }
+
+    public static ValueConstraint of(ValueVariable valueVariable, com.vaticle.typeql.lang.pattern.constraint.ValueConstraint constraint, VariableRegistry registry) {
+        if (constraint.isAssignment()) {
+            return ExpressionConstraint.of(valueVariable, constraint.asAssignment(), registry);
```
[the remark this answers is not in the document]
krishnangovindraj: Yes

concept/answer/ConceptMap.java
krishnangovindraj: We could see if we can go even further and make an interface for values themselves.

traversal/structure/StructureEdge.java
```diff
@@ -50,6 +50,10 @@ public boolean isNative() {
         return false;
     }
 
+    public boolean isInput() {
+        return false;
+    }
```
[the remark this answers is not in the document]
krishnangovindraj: Works. I'll change all of the Input's to Argument

traversal/structure/Structure.java
```diff
@@ -82,11 +90,21 @@ public void equalEdge(StructureVertex<?> from, StructureVertex<?> to) {
...
         StructureEdge.Predicate edge = new StructureEdge.Predicate(from, to, predicate);
         recordEdge(edge);
     }
 
+    public void dependencyEdge(StructureVertex.Value operand, StructureVertex.Value result) {
```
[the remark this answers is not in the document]
krishnangovindraj: Yeah, I thought I'd need dependencies elsewhere (in predicates) as well so I figured it could be a single dependency edge at this level/the planner level and then the actual edge in the procedure. I'll change these to argument as well.

traversal/procedure/ProcedureEdge.java
```diff
@@ -270,6 +309,46 @@ public boolean isClosure(GraphManager graphMgr, Vertex<?, ?> fromVertex, Vertex<
         }
     }
 
+    public static class Input extends ProcedureEdge<ProcedureVertex<?, ?>, ProcedureVertex<?, ?>> {
```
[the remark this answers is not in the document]
krishnangovindraj: That's not really true. But also, is there still a case where we use isClosure?

traversal/planner/PlannerVertex.java
```diff
@@ -271,4 +275,34 @@ public PlannerVertex.Type asType() {
             return this;
         }
     }
+
+    public static class Value extends PlannerVertex<Properties.Value> {
+        Value(Identifier id) {
+            this(id, null);
+        }
```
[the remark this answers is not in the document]
krishnangovindraj: No, Neither are the other 3. I can remove all of them

traversal/expression/Expression.java
krishnangovindraj: I do agree. The actual implementations have signature in `T evaluate(Map<Identifier, ValueHolder<?>> valueMap);`
This is the interfacing for the traversal engine, but that can be pushed out if needed.
I think the traversal engine has always kept vertices as black boxes, so we wouldn't want it to have to deal with the datatypes either.

traversal/expression/ExpressionFactory.java
krishnangovindraj: So just `constant`?

encoding/Expression.java
krishnangovindraj: Just plain var should be fine. 
Now that the ValueHolder interface exists, it should be possible.

.factory/automation.yml
```diff
@@ -78,6 +78,7 @@ build:
       command: |
         bazel test //test/behaviour/typeql/language/match/... --test_output=streamed
         bazel test //test/behaviour/typeql/language/get/... --test_output=streamed
+        bazel test //test/behaviour/typeql/language/expressions/... --test_output=streamed
```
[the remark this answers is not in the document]
krishnangovindraj: Yes. there's just one though

common/exception/ErrorMessage.java
```diff
@@ -556,6 +566,30 @@ public static class TypeWrite extends ErrorMessage {
...
+        public static final Expression INVALID_EXPRESSION_RESULT =
+                new Expression(5, "The expression produced an invalid result: '%s'");
+        public static final Expression INVALID_VALUE_CASTING =
+                new Expression(6, "Invalid concept conversion from '%s' to '%s'.");
```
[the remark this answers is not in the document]
krishnangovindraj: There's an INVALID_THING_CASTING and INVALID_THING_IID_CASTING in ThingRead

common/exception/ErrorMessage.java
```diff
@@ -556,6 +566,30 @@ public static class TypeWrite extends ErrorMessage {
         }
     }
 
+    public static class Expression extends ErrorMessage {
+        public static final Expression UNRESOLVABLE_OPERATION =
+                new Expression(1, "The operation could not be resolved for the data-types: '%s' %s '%s'.");
+        public static final Expression UNRESOLVABLE_FUNCTION =
```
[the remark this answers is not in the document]
krishnangovindraj: Do I need to change the text as well? Or do I keep that until we decide on the naming of the process?

common/exception/ErrorMessage.java
```diff
@@ -556,6 +566,30 @@ public static class TypeWrite extends ErrorMessage {
...
+                new Expression(2, "The function could not be resolved for the signature: '%s(%s)'.");
+        public static final Expression AMBIGUOUS_EXPRESSION_ARGUMENT_TYPE =
+                new Expression(3, "The variable '%s' can have multiple value types: [%s].");
+        public static final Expression EVALUATION_ERROR =
+                new Expression(4, "An error occured while evaluating an expression:\nExpression: '%s'; \nInput: '%s'.\nError: '%s'");
+        public static final Expression INVALID_EXPRESSION_RESULT =
+                new Expression(5, "The expression produced an invalid result: '%s'");
```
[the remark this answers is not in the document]
krishnangovindraj: I've changed it to `INVALID_RESULT` "An operation produced an invalid result: '%s'";
The actual expression isn't available at the point this happens. Should I move this to the `Internal` Error messages? I currently wrap this in an EVALUATION_ERROR and throw it with the expression included.

logic/tool/ExpressionResolution.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    private Encoding.ValueType<?> resolveExpressionTypes(ValueVariable v) {
+        if (v.assignedBy().returnType() == null) {
+            Map<Identifier, Encoding.ValueType<?>> inputTypes = new HashMap<>();
+            // No circular dependencies -> No infinite recursion
+            iterate(v.assignedBy().inputs()).forEachRemaining(input -> {
+                if (input.isThing()) {
+                    inputTypes.put(input.id(), ensureAndGetSingleValueType(input.asThing()));
+                } else if (v.isValue()) {
```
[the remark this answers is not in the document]
krishnangovindraj: It is indeed a bug.
There aren't any tests.
But they wouldn't have caught it either since the condition is always true (and is expected to be true).

logic/tool/ExpressionResolution.java
```diff
@@ -0,0 +1,162 @@
...
+                    throw TypeDBException.of(UNASSIGNED_VALUE_VARIABLE, v);
+                });
+        Set<ValueVariable> safe = new HashSet<>();
+        Stack<ValueVariable> current = new Stack<>();
+        valueVariables.forEach(v -> validateCyclicAssignment(v, safe, current));
+
+        valueVariables.forEach(v -> resolveExpressionTypes(v.asValue()));
+    }
+
+    private void validateCyclicAssignment(ValueVariable at, Set<ValueVariable> safe, Stack<ValueVariable> current) {
```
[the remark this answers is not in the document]
krishnangovindraj: 1 - yes. 
validateAcyclicAssignments - yes.

2. Isn't that making it more stateful depending on the class? 
 - Can I put it in a while loop? Yes, but then I have to handle the branches in the tree explicitly (which the forEachRemaining is doing implicitly for me). This feels like an intuitive signature for a DFS

pattern/constraint/value/ExpressionConstraint.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+
+public class ExpressionConstraint extends ValueConstraint {
+    private final com.vaticle.typeql.lang.pattern.constraint.ValueConstraint.Assignment.Expression expression;
```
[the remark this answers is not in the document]
krishnangovindraj: For the top level class or nested classes as well?
Our style guide really needs to be written down :(

traversal/expression/Operation.java
```diff
@@ -0,0 +1,346 @@
...
+
+            @Override
+            public Expression<?> build(Expression<Long> x1, Expression<Long> x2) {
+                return thenFactory.build(new Function.CastLongToDouble(x1), new Function.CastLongToDouble(x2));
+            }
+        }
+    }
+
+
+    public static class Unimplemented extends Operation<Void, Void> implements Expression<Void> {
```
krishnangovindraj, line 322: Can't merge till I've implemented them all.

krishnangovindraj (commented):
Flag unimplemented functions.

traversal/procedure/GraphProcedure.java
```diff
@@ -503,5 +525,12 @@ public ProcedureEdge.Native.Thing.RolePlayer.Backward backwardRolePlayer(
             attachEdge(player, relation, edge);
             return edge;
         }
+
+        public ProcedureEdge.Input forwardInput(ProcedureVertex<?,?> input, ProcedureVertex.Value result) {
+            // backwardInput is illegal
```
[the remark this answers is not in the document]
krishnangovindraj: This looks like a builder method. Do I want to provide a builder method just to throw ILLEGAL_STATE ?

common/exception/ErrorMessage.java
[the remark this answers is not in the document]
flyingsilverfin: It's best to keep the error name and the text in line - but you can refactor both when we finish deciding on the process naming.

traversal/expression/Operation.java
```diff
@@ -0,0 +1,293 @@
...
+
+import java.time.LocalDateTime;
+import java.util.HashMap;
+import java.util.Map;
+
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.DUPLICATE_EXPRESSION_DEFINITION;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.UNIMPLEMENTED;
+
+public abstract class Operation<T1, T2> implements ExpressionFactory.OperationFactory<T1, T2> {
+    private static Map<Triple<java.lang.String, Encoding.ValueType<?>, Encoding.ValueType<?>>,ExpressionFactory.OperationFactory> registry = null;
```
krishnangovindraj, line 36: Had to use an unparametrised OperationFactory here. 
The operand types vary from signature to signature, So I had trouble getting the `build` call in  `ExpressionFactory.instantiateOperation` to compile. I might be able to do parametrise that function, but that just pushes the problem to the caller.

logic/tool/ExpressionCompiler.java
```diff
@@ -0,0 +1,150 @@
...
+
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Expression.AMBIGUOUS_ARGUMENT_TYPE;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Expression.FUNCTION_NOT_FOUND;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Expression.OPERATION_NOT_FOUND;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Pattern.VALUE_ASSIGNMENT_CYCLE;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Pattern.VALUE_VARIABLE_UNASSIGNED;
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+
+public class ExpressionCompiler {
```
krishnangovindraj, line 50: This now does both ExpressionResolution & Compilation.

krishnangovindraj (commented):
Describe changes.

common/exception/ErrorMessage.java
```diff
@@ -556,6 +566,30 @@ public static class TypeWrite extends ErrorMessage {
...
+        public static final Expression AMBIGUOUS_ARGUMENT_TYPE =
+                new Expression(3, "The variable '%s' can have multiple value types: [%s].");
+        public static final Expression EVALUATION_ERROR =
+                new Expression(4, "An error occured while evaluating an expression:\nExpression: '%s = %s'; \nInput: '%s'.\nError: '%s'");
+        public static final Expression INVALID_RESULT =
+                new Expression(5, "An operation produced an invalid result: '%s'");
+        public static final Expression INVALID_VALUE_CASTING =
+                new Expression(6, "Invalid value conversion from '%s' to '%s'.");
+        public static final Expression INVALID_VALUE_IID_CASTING =
+                new Expression(7, "Invalid Thing IID casting to '%s'.");
```
flyingsilverfin, line 583: The key point is that these errors to expressions at all - this should be in the something analogous to the `ThingRead` error message section.

I think you need a new section around the `Thing` errors called `Value`, which includes these two errors inside.

flyingsilverfin: eg. `VALUE_CASTING` and `VALUE_IID_CASTING`

traversal/expression/ExpressionFactory.java
flyingsilverfin: Yep!

traversal/expression/ExpressionFactory.java
```diff
@@ -0,0 +1,83 @@
...
+import com.vaticle.typedb.core.common.exception.TypeDBException;
+import com.vaticle.typedb.core.encoding.Encoding;
+import com.vaticle.typedb.core.traversal.common.Identifier;
+
+import java.util.List;
+
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
+
+public abstract class ExpressionFactory {
+    private static final Expression<?>[] toArrayTemplate = new Expression[0];
```
flyingsilverfin, line 30: You're missing some newlines. I would also order them like this:
```
loadConstants()

toArrayTemplate

interface

interface
```

traversal/expression/Operation.java
```diff
@@ -0,0 +1,293 @@
...
+        this.x1 = x1;
+        this.x2 = x2;
+    }
+
+    public static abstract class DoubleOp extends Operation<java.lang.Double, java.lang.Double> implements Expression.Double {
+        public DoubleOp(Expression<java.lang.Double> x1, Expression<java.lang.Double> x2) {
+            super(x1,x2);
+        }
+
+        public static class Plus extends DoubleOp {
```
flyingsilverfin, line 89: Invert this - group by Operation name first, then return type - we think about what plus is defined for rather than what can return a double

pattern/variable/VariableRegistry.java
```diff
@@ -51,6 +52,7 @@ public class VariableRegistry {
     private final boolean allowDerived;
     private final Map<Reference, TypeVariable> types;
     private final Map<Reference, ThingVariable> things;
+    private final Map<Reference, ValueVariable> values;
```
flyingsilverfin, line 55: can we specialise this to `Reference.Value`?

pattern/variable/VariableRegistry.java
```diff
@@ -164,31 +173,49 @@ public Set<ThingVariable> things() {
...
-        else return things.computeIfAbsent(reference, constructor);
+        if (types.containsKey(reference)) {
+            throw TypeDBException.of(VARIABLE_CONTRADICTION, reference);
+        } else if (reference.isName() && values.containsKey(Reference.value(reference.name()))) {
+            throw TypeDBException.of(VARIABLE_NAME_CONFLICT, reference.name());
+        } else return things.computeIfAbsent(reference, constructor);
+    }
+
+    public ValueVariable computeValueIfAbsent(Reference reference, Function<Reference, ValueVariable> constructor) {
+        Reference clashingConceptRef = Reference.name(reference.name());
```
flyingsilverfin, line 216: this function should take a `Reference.Value` and then the clashing reference should be `Reference asNameReference = Reference.name(reference.name())`.

Right now it's not clear why you'd unwrap a reference's name only to wrap it in a name again... the fact it must be a `Value` reference is not clear!

pattern/constraint/value/ValueConstraint.java
```diff
@@ -0,0 +1,110 @@
...
+    public boolean equals(Object o) {
+        assert false;
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        ValueConstraint that = (ValueConstraint) o;
+        return this.owner.equals(that.owner);
+    }
+
+    @Override
+    public abstract int hashCode();
```
flyingsilverfin, line 109: I think if you define the `equals` here you should be able to define the equality function too?

flyingsilverfin: This stuff can lead to nasty random bugs so let's just be extra careful

pattern/constraint/value/AssignmentConstraint.java
```diff
@@ -0,0 +1,624 @@
...
+        this.hash = Objects.hash(this.owner, this.expression);
+    }
+
+    public static AssignmentConstraint of(ValueVariable valueVariable, com.vaticle.typeql.lang.pattern.constraint.ValueConstraint.Assignment constraint, VariableRegistry registry) {
+        Set<Variable> arguments = new HashSet<>();
+        for (BoundVariable v : constraint.variables()) {
+            if (v.isThing()) arguments.add(registry.register(v.asThing()));
+            else if (v.isValue()) arguments.add(registry.register(v.asValue()));
+            else throw TypeDBException.of(ILLEGAL_STATE);
+        }
```
flyingsilverfin, line 68: It's a bit different from our usual pattern to pre-register all the variables first, then retrieve them. However I don't think this approach is too bad in this context

pattern/constraint/value/AssignmentConstraint.java
```diff
@@ -0,0 +1,624 @@
...
+
+    public abstract static class Expression {
+
+        private com.vaticle.typedb.core.traversal.expression.Expression<?> compiled;
+
+        Expression() {
+            this.compiled = null;
+        }
+
+        public static Expression build(com.vaticle.typeql.lang.pattern.constraint.ValueConstraint.Assignment.Expression expr, VariableRegistry registry) {
```
flyingsilverfin, line 150: I think this should also be called `of` instead of `build()`

pattern/constraint/value/AssignmentConstraint.java
```diff
@@ -0,0 +1,624 @@
...
+        public abstract Expression clone(Map<Identifier, Variable> argMap);
+
+        public static class Operation extends Expression {
+
+            private final Op op;
+            private final Expression a;
+            private final Expression b;
+            private final int hash;
+
+            public Operation(Op op, Expression a, Expression b) {
```
flyingsilverfin, line 242: Let's use the same oeprator terminology everywhere - it should be `first` and `second` to match the traversal expression operation naming terminology.

The point here is to be precise about every variable name, and be as consistent as possible everywhere

pattern/constraint/value/AssignmentConstraint.java
```diff
@@ -0,0 +1,624 @@
...
+            }
+
+            @Override
+            public Expression clone(Map<Identifier, Variable> argMap) {
+                return new Operation(op, a.clone(argMap), b.clone(argMap));
+            }
+        }
+
+        public static class Function extends Expression {
+            private final String symbol;
```
flyingsilverfin, line 292: Hmm I think this should also use  TypeQL symbol if possible? What do you think... I left a comment there that we're missing `TypeQLTokens` for Functions

traversal/common/Identifier.java
```diff
@@ -26,6 +26,7 @@
 
 import static com.vaticle.typedb.common.util.Objects.className;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_CAST;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
 
 public abstract class Identifier {
```
flyingsilverfin, line 31: Let's rename `.Name` to `.NamedConcept` and `.Anonymous` to `.AnonymousConcept`. It's not really great but it'll have to get fixed in Rust

logic/tool/ExpressionResolution.java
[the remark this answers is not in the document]
flyingsilverfin: We can refactor it in the future, but I think needing an `assert` to validate the operation of the recursion is a bit of an anti-pattern (or rather, means the logic can be improved)

logic/tool/ExpressionCompiler.java
```diff
@@ -0,0 +1,150 @@
...
+        ValueVariable poppedAt = stack.pop();
+        assert poppedAt == var;
+        validated.add(var);
+    }
+
+    private Encoding.ValueType<?> compileExpression(ValueVariable v) {
+        if (v.assignment().valueType() == null) {
+            Map<Identifier, Encoding.ValueType<?>> argTypes = new HashMap<>();
+            // No circular dependencies -> No infinite recursion
+            iterate(v.assignment().arguments()).forEachRemaining(argument -> {
```
flyingsilverfin, line 89: fyi java Collections have a built-in `forEach` that mean you don't have to use a iterator and `forEachRemaining` explicitly:
```
v.assignment().arguments().forEach(argument -> ...)
```

flyingsilverfin: might be ever so slightly more performant

pattern/constraint/value/ValueConstraint.java
krishnangovindraj: As in?
I think I wanted to force the subclasses to implement them.
I can also remove this.

query/Inserter.java
```diff
@@ -131,6 +138,8 @@ private static void validate(ThingVariable var) {
         }
     }
 
+    private static void validate(ValueVariable var) { }
```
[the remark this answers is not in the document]
krishnangovindraj: We should yes.
Here and in delete I suppose.

common/exception/ErrorMessage.java
flyingsilverfin: @krishnangovindraj  i think this still needs implementing

common/exception/ErrorMessage.java
```diff
@@ -285,6 +287,14 @@ public static class Pattern extends ErrorMessage {
                 new Pattern(17, "The schema does not allow for data matching the pattern '%s', due to the included pattern '%s'.");
         public static final Pattern INCOHERENT_PATTERN_VARIABLE_VALUE =
                 new Pattern(18, "The schema does not allow for data matching the pattern '%s', due to contradicting attribute value types for '%s'.");
+        public static final Pattern VARIABLE_NAME_CONFLICT =
+                new Pattern(19, "The variable name '%s' was used both for a concept variable and a value variable.");
+        public static final Pattern VALUE_VARIABLE_UNASSIGNED =
+                new Pattern(20, "The variable '%s' is never assigned to.");
```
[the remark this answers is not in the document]
flyingsilverfin: this needs resolving

flyingsilverfin:
@krishnangovindraj  can you please double check comments from my last review one by one and make sure they're implemented? I started commenting on the ones that are not done but i think you should just validate them again yourself. You should resolve the ones that are implemented unambiguously and comment on the ones that are not done for a specific reason/up for discussion

traversal/expression/Operation.java
```diff
@@ -0,0 +1,101 @@
...
+
+            default:
+                throw TypeDBException.of(ILLEGAL_STATE);
+        }
+    }
+
+    private static Expression<?> instantiatePlus(Expression<?> first, Expression<?> second) {
+        if (first.isLong()) {
+            if (second.isLong()) return new ExpressionBase.Binary.Long<>(first.asLong(), second.asLong(), (a,b) -> a + b);
+            if (second.isDouble()) return new ExpressionBase.Binary.Double<>(first.asLong(), second.asDouble(), (a,b) -> a.doubleValue() + b);
```
flyingsilverfin, line 33: these are still better as `else-if` as a better signal these are mutually exclusive, same below

flyingsilverfin: in fact you can get the same effect with:
```
if {
  if
  else if
} else if {
  if
  else if
}

throw
```

traversal/expression/Operation.java
```diff
@@ -0,0 +1,101 @@
...
+        if (first.isDouble()) {
+            if (second.isLong()) return new ExpressionBase.Binary.Double<>(first.asDouble(), second.asLong(), (a,b) -> a * b.doubleValue());
+            if (second.isDouble()) return new ExpressionBase.Binary.Double<>(first.asDouble(), second.asDouble(), (a,b) -> a * b);
+        }
+
+        throw TypeDBException.of(OPERATION_NOT_FOUND,
+                first.returnType().name(), TypeQLToken.Expression.Operation.TIMES, second.returnType().name());
+    }
+
+    private static Expression<?> instantiateDiv(Expression<?> first, Expression<?> second) {
```
flyingsilverfin, line 76: Do we want to throw our own exception rather than Java's for a Division by 0 error? Then we can show which expression caused it, etc.
Secondly, we should add a test for division by 0 to BDD

pattern/constraint/value/AssignmentConstraint.java
```diff
@@ -0,0 +1,622 @@
...
+        public abstract Expression clone(Map<Identifier, Variable> argMap);
+
+        public static class Operation extends Expression {
+
+            private final TypeQLToken.Expression.Operation op;
+            private final Expression a;
+            private final Expression b;
+            private final int hash;
+
+            public Operation(TypeQLToken.Expression.Operation op, Expression a, Expression b) {
```
flyingsilverfin, line 240: pick consistent naming scheme throughout the expression-realted changes. Operations should always use first/second, or lhs/rhs, etc.

flyingsilverfin (changes requested):

encoding/iid/VertexIID.java
```diff
@@ -422,4 +426,176 @@ public DateTime asDateTime() {
...
+        public abstract VALUE value();
+
+        @Override
+        public Encoding.Vertex.Value encoding() {
+            return Encoding.Vertex.Value.VALUE;
+        }
+
+        @Override
+        public java.lang.String toString() {
+            return "ValueVertex[" + valueType + " :" + value() + "]";
```
benjaminasmall, line 471: You sure about " :"?

pattern/variable/VariableRegistry.java
```diff
@@ -94,36 +97,33 @@ public static VariableRegistry createFromVariables(List<? extends BoundVariable>
...
-                if (typeQLVar.isConcept()) unboundedVariables.add(typeQLVar.asConcept());
-                else registry.register(typeQLVar);
+                if (typeQLVar.isThing()) registry.register(typeQLVar.asThing());
+                else if (typeQLVar.isType()) registry.register(typeQLVar.asType());
+                else if (typeQLVar.isValue()) registry.register(typeQLVar.asValue());
+                else if (typeQLVar.isConcept()) unboundedVariables.add(typeQLVar.asConcept());
+                else throw TypeDBException.of(ILLEGAL_STATE);
             });
-            unboundedVariables.forEach(registry::register);
+            unboundedVariables.forEach(registry::register); // Also need to bring back the Base type Variable register method.
```
benjaminasmall, line 106: Has this been done?

traversal/GraphTraversal.java
```diff
@@ -212,45 +213,108 @@ public void clearLabels(Identifier.Variable type) {
...
+            structure.predicateEdge(structure.valueVertex(valueVariable), structure.thingVertex(attr), predicate);
+        }
+
+        public void predicateValueValue(Identifier.Variable eval1, TypeQLToken.Predicate.Equality token, Identifier.Variable eval2) {
+            Predicate.Variable predicate = Predicate.Variable.of(token);
+            structure.predicateEdge(structure.valueVertex(eval1), structure.valueVertex(eval2), predicate);
+        }
+
+        public void assignment(Identifier.Variable result, Set<Identifier.Variable> arguments, Expression<?> expression) {
+            StructureVertex.Value resultVertex = structure.valueVertex(result);
```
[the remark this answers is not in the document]
krishnangovindraj: It does play the role of the result in the assignment, so it could still be?
Anyway, will change.

traversal/expression/Operation.java
krishnangovindraj: We should. But I'd rather do it higher up so there's less chance of missing them.

pattern/variable/VariableRegistry.java
krishnangovindraj: Nope. And I don't think I need to anymore.

traversal/expression/ExpressionBase.java
```diff
@@ -0,0 +1,259 @@
+package com.vaticle.typedb.core.traversal.expression;
```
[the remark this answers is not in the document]
krishnangovindraj: Shall I do it when all files in development is updated to 2023?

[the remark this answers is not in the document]
flyingsilverfin: Yeah ok just change it to 2022 and not 2023 for now and then we'll do a sweeping change

traversal/expression/Operation.java
krishnangovindraj: Turns out we do have a BDD test for division by 0 and we do throw our own error.
But it's specifically handled by the check in `traversal.expression.Expression.Double.evaluateToVertex`

I'd like to handle all errors around there

pattern/constraint/value/AssignmentConstraint.java
```diff
@@ -0,0 +1,623 @@
...
+            else throw TypeDBException.of(ILLEGAL_STATE);
+        });
+        Map<Identifier, Variable> argMap = new HashMap<>();
+        clonedVariables.forEach(v -> argMap.put(v.id(), v));
+        AssignmentConstraint cloned = cloner.cloneVariable(this.owner).assign(expression.clone(argMap), clonedVariables);
+        cloned.compiledExpression(compiledExpression);
+        return cloned;
+    }
+
+    public abstract static class Expression {
```
krishnangovindraj, line 157: I don't think we gain anything out of having this instead of the TypeQL expression.

traversal/GraphTraversal.java
```diff
@@ -212,45 +213,108 @@ public void clearLabels(Identifier.Variable type) {
...
+            parameters.pushValue(valueVariable, predicate, new Parameters.Value.DateTime(value));
+            structure.thingVertex(valueVariable).props().predicate(predicate);
+        }
+
+        public void predicateValueThing(Identifier.Variable valueVariable, TypeQLToken.Predicate.Equality token, Identifier.Variable attr) {
+            Predicate.Variable predicate = Predicate.Variable.of(token);
+            structure.predicateEdge(structure.valueVertex(valueVariable), structure.thingVertex(attr), predicate);
+        }
+
+        public void predicateValueValue(Identifier.Variable eval1, TypeQLToken.Predicate.Equality token, Identifier.Variable eval2) {
```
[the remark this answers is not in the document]
flyingsilverfin: Yes i'd probably go with `left` and `right` for all? At least let's pick 1 and stick with it consistently

pattern/constraint/value/AssignmentConstraint.java
flyingsilverfin: I think you might be right - does it all work if we just keep the `compiledExpression` on the `AssignmentConstraint`?  What about the `toString`, should just print the TypeQL expression?

test/behaviour/typeql/TypeQLSteps.java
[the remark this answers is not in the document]
flyingsilverfin: @krishnangovindraj  let's make sure we only assert it throws for now and add this as an item to an issue that is about stabilising the error codes? I think someone was talking about that recently - if it doesn't already exist you can create the issue of it and add this as an item to it.

A part of that issue should be to refactor BDD to use error codes for exception checking. This is a large issue tied into https://github.com/vaticle/typedb/issues/6146 that is to do with a centralised error repository.

flyingsilverfin (approved):

pattern/constraint/value/AssignmentConstraint.java
flyingsilverfin: I think the compiled 'traversal' level to-string can be printed in a very functional way just like we printed edges like
`--PLAYS-->` etc. So yes, the symbol should be there (eg. should be readable in a query plan).

The `toString` on `Constraint`s should be pretty much parse-able by TypeQL though. However, if we go with this then there is no Constraint-equivalent of Expressions. Thus, we would have to embed the TypeQL expressions & use their `toString()`. Would that work?

krishnangovindraj: We could indeed embed the TypeQL expression into the compiled expression, but we can't call toString, because it would recurse into the TypeQL sub-expressions, while we'd want it to recurse into compiled sub-expressions.

Do we really need traversal level expressions to be parseable by typeql?

pattern/constraint/value/AssignmentConstraint.java
```diff
@@ -0,0 +1,164 @@
...
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Pattern.VALUE_VARIABLE_DUPLICATE_ASSIGMENT;
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+import static com.vaticle.typedb.core.common.iterator.Iterators.link;
+
+public class AssignmentConstraint extends ValueConstraint {
+
+    private final Expression expression;
+    private final Set<ThingVariable> thingArguments;
+    private final Set<ValueVariable> valueArguments;
+    private com.vaticle.typedb.core.traversal.expression.Expression<?> compiledExpression;
```
flyingsilverfin, line 50: these variables can be called 'traversalExpression'

logic/tool/ExpressionCompiler.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.Stack;
+
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Expression.AMBIGUOUS_ARGUMENT_TYPE;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILLEGAL_STATE;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Pattern.VALUE_ASSIGNMENT_CYCLE;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.Pattern.VALUE_VARIABLE_UNASSIGNED;
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+import static java.lang.String.join;
+
+public class ExpressionCompiler {
```
flyingsilverfin, line 51: Let's call this `ExpressionResolution` and move it into a better package if we can - it doesn't feel like "logic" @flyingsilverfin

PR #6733, opened 2023-01-18
Short-cut rules where possible & relax reasoner planner orderings

krishnangovindraj:
We revise the criteria for a constraint to be dependent on a variable being bound. This refines the search-space of the reasoner planner to include more legal plans, and allows an optimisation for reasoning involving relations playing roles in other relations.

In certain cases during reasoning, we can determine that all answers for a given constraint have already been computed. In such cases, we can bypass inference and simply look up the answers.

* Loosen the criteria for resolvable dependencies:
  * A resolvable _r_ is only dependent on a variable _v_ if : 
      1. _v_ is 'generated' by some other concludable in the conjunction AND
      2. _r_ contains a constraint involving `v`
      3. no constraint in _r_  generates values for `v` 
      * (Currently, predicates are the only constraints which can involve a variable but not generate inferred values for it)
  * Negations remain dependent on variables which occur unnegated in conjunction(s) which contain it.    
* `RecursivePlanner` no longer relaxes dependencies to generate orderings when it cannot schedule any resolvable.
* `ConcludableProcessor.Match` will not trigger rules (i.e. create conclusion processors upstream)  if the answer is already guaranteed to be materialised. It will create instead do a traversal which includes inferred concepts.

### DIFF
--- a/reasoner/controller/ConcludableController.java
+++ b/reasoner/controller/ConcludableController.java
@@ -72,7 +72,7 @@ public abstract class ConcludableController<INPUT, OUTPUT,
 
     @Override
     public void setUpUpstreamControllers() {
-        iterate(registry().logicManager().applicableRules(concludable).entrySet()).forEachRemaining( ruleAndUnifiers -> {
+        iterate(registry().logicManager().applicableRules(concludable).entrySet()).forEachRemaining(ruleAndUnifiers -> {
                 Rule rule = ruleAndUnifiers.getKey();
                 Driver<? extends ConclusionController<INPUT, ?, ?>> controller = registerConclusionController(rule);
                 conclusionControllers.put(rule.conclusion(), controller);
@@ -200,11 +200,7 @@ public abstract class ConcludableController<INPUT, OUTPUT,
             this.requestedConnections = new HashSet<>();
         }
 
-        @Override
-        public void setUp() {
-            setHubReactive(fanInFanOut(this));
-            // TODO: Add a find first optimisation when all variables are bound
-            mayAddTraversal(false);
+        void addRules() {
             conclusionUnifiers.forEach((conclusion, unifiers) -> {
                 unifiers.forEach(unifier -> unifier.unify(bounds).ifPresent(boundsAndRequirements -> {
                     InputPort<INPUT> inputPort = createInputPort();
@@ -229,7 +225,6 @@ public abstract class ConcludableController<INPUT, OUTPUT,
         protected abstract REQ createRequest(Reactive.Identifier identifier, Conclusion conclusion,
                                              ConceptMap bounds);
 
-        protected abstract void mayAddTraversal(boolean includeInferred);
 
         private void mayRequestConnection(REQ conclusionRequest) {
             if (!requestedConnections.contains(conclusionRequest.id())) {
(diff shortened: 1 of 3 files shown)


reasoner/controller/ConcludableController.java
```diff
@@ -180,20 +180,22 @@ protected Driver<ConclusionController.Explain> registerConclusionController(Rule
...
+        final ConceptMap bounds;
         private final Set<Variable.Retrievable> unboundVars;  // TODO: Can just use a boolean to indicate if fully bound
         private final Map<Conclusion, Set<Unifier>> conclusionUnifiers;
         private final Set<Identifier> requestedConnections;
         final java.util.function.Supplier<FunctionalIterator<ConceptMap>> traversalSuppplier;
 
         Processor(Driver<PROCESSOR> driver,
                   Driver<? extends AbstractController<?, INPUT, OUTPUT, REQ, PROCESSOR, ?>> controller,
-                  Context context, ConceptMap bounds,
+                  Concludable concludable, Context context, ConceptMap bounds,
```
flyingsilverfin, line 192: Concludable should come after context

reasoner/controller/ConcludableController.java
```diff
@@ -241,22 +243,33 @@ private void mayRequestConnection(REQ conclusionRequest) {
...
+
+
+            @Override
+            public void setUp() {
+                // TODO: Add a find first optimisation when all variables are bound
+                if (concludable.isRelation() && concludable.generating().isPresent() && bounds.contains(concludable.generating().get().id())) {
+                    // Optimisation: If we know the relation instance, just do a lookup.
+                    setHubReactive(fanInFanOut(this));
+                    mayAddTraversal(true);
+                } else {
```
flyingsilverfin, line 264: we can generalise this optimisation:
```
                if (concludable.isInferredAnswer(bounds)) {
                    // don't trigger rules ever
                    if (!bounds are fully binding concludable) {
                        mayAddTraversal(true);
                    } else {
                        return singleton bounds as answer
                    }
                } else {
                    mayAddTraversal(false);
                    triggerRules();
                }
```

you'll need to add `Concludable.isAnswer` which returns true if the concept map provided is an answer that fits as either an inferred or non-inferred answer

reasoner/planner/ReasonerPlanner.java
```diff
@@ -95,38 +101,47 @@ Plan getPlan(CallMode callMode) {
 
     abstract Plan computePlan(CallMode callMode);
 
+    private static boolean isDependency(Resolvable<?> resolvable, Variable v) {
```
flyingsilverfin, line 104: put below the user

reasoner/controller/ConcludableController.java
krishnangovindraj: It comes before in the controller constructor. Should I change both?

reasoner/planner/ReasonerPlanner.java
```diff
@@ -95,38 +101,47 @@ Plan getPlan(CallMode callMode) {
...
+
+        // Add dependency between negateds and shared variables
+        Set<Variable> unnegatedVars = iterate(resolvables).filter(r -> !r.isNegated()).flatMap(r -> iterate(r.variables())).toSet();
+        iterate(resolvables).filter(Resolvable::isNegated)
+                .forEachRemaining(resolvable -> deps.get(resolvable).addAll(intersection(resolvable.variables(), unnegatedVars)));
+
+        // Add dependency for resolvable to any variables for which it can't (independently) generate all satisfying values
+        iterate(resolvables).filter(resolvable -> !resolvable.isNegated())
+                .forEachRemaining(resolvable -> deps.get(resolvable).addAll(
+                        iterate(resolvable.variables()).filter(ReasonerPlanner::hasNoGeneratingConstraint).toSet()
```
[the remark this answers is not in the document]
krishnangovindraj: The RHS part doesn't seem to hold. 

The BDD scenario "a rule can be applied based on concept inequality" has a query
```
        $value isa! retailer;
        $unwantedValue isa! retailer;
        $x has $value;
        $y has $unwantedValue;
        $unwantedValue "Ocado";
        $value != $unwantedValue;
```
which is (strangely) planned to be:
1. `Concludable: $y has $unwantedValue;`
2. `Retrievable: $value != $unwantedValue;`
3. `Concludable: $x has $value;` 

2 being a separate retrievable is part of the problem. Another problem is a weak planning of predicates and the connectedness approach to planning.

reasoner/planner/RecursivePlanner.java
```diff
@@ -373,10 +374,6 @@ private void porDFS(Stack<Resolvable<?>> currentPath, Set<Variable> currentBound
                     .filter(r -> ReasonerPlanner.dependenciesSatisfied(r, currentBoundVars, dependencies) && !sleepSet.contains(r))
                     .toList();
 
-            if (enabled.isEmpty()) {
-                enabled = iterate(remaining).filter(r -> !r.isNegated() && !sleepSet.contains(r)).toList();
```
krishnangovindraj, line 377: When our dependencies were too strict, we'd relax them to be able to generate orderings.
Now that they're strict enough, We should prevent that.

reasoner/controller/ConcludableController.java
```diff
@@ -241,22 +234,44 @@ private void mayRequestConnection(REQ conclusionRequest) {
...
+                    addRules();
+                }
+            }
+
+            private boolean canBypassReasoning(ConceptMap bounds) {
+                if (concludable.isHas()) {
+                    Concept owner; Concept attribute;
+                    return  (owner = bounds.get(concludable.asHas().owner().id())) != null &&
+                            (attribute = bounds.get(concludable.asHas().attribute().id())) != null &&
+                            (owner.asThing().hasInferred(attribute.asAttribute()) || owner.asThing().hasNonInferred(attribute.asAttribute()));
```
krishnangovindraj, line 263: We check the existence of the edge here, to determine whether we can bypass traversal.
Will see how hard it is to bypass the full traversal for non-relation cases.

flyingsilverfin (approved):

PR #6775, opened 2023-03-24
Unique attributes

krishnangovindraj:
**For the title of this PR:** please follow the grammatical rules of a usual publication title, without capitalisation (except for the first letter). Thus, the title should NOT CONTAIN CODE: no dots, no parentheses, no backticks, no brackets, etc. It needs to be distinctive (not detailed) and succinct (not lengthy). Details of this PR will go in the description. **For the description of this PR:** please replace every line in curly brackets ( { like this } ) with an appropriate description following the guidance. Finally, **please remove this paragraph**.

  - its responsibility, 


dependencies/vaticle/repositories.bzl
```diff
@@ -27,8 +27,8 @@ def vaticle_dependencies():
 def vaticle_typeql():
     git_repository(
         name = "vaticle_typeql",
-        remote = "https://github.com/vaticle/typeql",
-        tag = "2.14.0",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typeql
+        remote = "https://github.com/krishnangovindraj/typeql",
+        commit = "61cdbdb3aff9fd16beb4d99b1ba6778e6ee2459a",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typeql
```
krishnangovindraj, line 31: Shouldn't merge till this is done.

graph/edge/impl/TypeEdgeImpl.java
```diff
@@ -344,6 +379,69 @@ public void commit() {
      */
     public static class Persisted extends TypeEdgeImpl implements TypeEdge {
 
+        public static class EncodedValue {
+            private static final byte KEY_OVERRIDEN = 1;
```
krishnangovindraj, line 383: We store the information that an OWNS edge has the unique constraint in the RocksDB value for the TypeEdge.
Originally, this value was only used for the overriden type (`owns concrete-attribute AS abstract-attributes`).

We convert the value into a key-value list. Thankfully, the overriden bytes were vertices, so they have a prefix. 
We recognise & handle this case for backward compatibility, but we insert new ones with KEY_OVERRIDEN as a key.

This can be extended to include cardinality information.

graph/edge/impl/TypeEdgeImpl.java
```diff
@@ -344,6 +379,69 @@ public void commit() {
...
+            private static final byte KEY_UNIQUE = 2;
+            // We used to store the overridden type directly
+            //  Encoding.prefix.VERTEX_ATTRIBUTE_TYPE.key = 111;
+            //  Encoding.prefix.VERTEX_ROLE_TYPE.key = 113;
```
krishnangovindraj, line 387: Need a better way to mark that these are expected values.
I'm afraid of using `private static final byte KEY_LEGACY_ROLE_TYPE = Encoding.prefix.VERTEX_ROLE_TYPE.key()` and introducing static initialisation dependencies.

krishnangovindraj (commented):
self review

server/concept/TypeService.java
```diff
@@ -352,13 +353,14 @@ private void getOwnsOverridden(ThingType thingType, ConceptProto.ThingType.GetOw
 
     private void setOwns(ThingType thingType, ConceptProto.ThingType.SetOwns.Req setOwnsRequest, UUID reqID) {
         AttributeType attributeType = getThingType(setOwnsRequest.getAttributeType()).asAttributeType();
-        boolean isKey = setOwnsRequest.getIsKey();
+        // TODO: Support unique over ConceptAPI
+        TypeEdge.UniquenessConstraint unique = setOwnsRequest.getIsKey() ? TypeEdge.UniquenessConstraint.KEY : TypeEdge.UniquenessConstraint.NONE;
```
benjaminasmall, line 357: Don't forget this.

pattern/constraint/type/OwnsConstraint.java
```diff
@@ -77,14 +83,18 @@ public Optional<TypeVariable> overridden() {
...
     }
 
+    public TypeEdge.UniquenessConstraint uniqueness() {
+        return uniqueness;
+    }
+
     public boolean isKey() {
-        return isKey;
+        return uniqueness == TypeEdge.UniquenessConstraint.KEY;
     }
```
benjaminasmall, line 92: This method should no longer be useful. If it is still used, you'll want to trace where and update that code based on the new `UniquenessConstraint`.

concept/type/impl/ThingTypeImpl.java
```diff
@@ -440,6 +493,21 @@ public AttributeType getOwnsOverridden(AttributeType attributeType) {
...
+    }
+
+    @Override
+    public boolean isOwnedUniqueExplicit(AttributeType attributeType) {
+        TypeVertex attrVertex = graphMgr.schema().getType(attributeType.getLabel());
+        if (attrVertex != null) {
+            TypeEdge ownsEdge = vertex.outs().edge(OWNS, attrVertex);
+            return ownsEdge != null && ownsEdge.unique();
+        }
+        return false;
```
benjaminasmall, line 508: I'm surprised you don't have an else here. Not important; just an observation.

concept/type/impl/ThingTypeImpl.java
```diff
@@ -379,49 +439,53 @@ FunctionalIterator<AttributeTypeImpl> overriddenOwns(boolean onlyKey, boolean tr
...
-    @Override
-    public Forwardable<AttributeType, Order.Asc> getOwnsExplicit(AttributeType.ValueType valueType) {
-        return getOwnsExplicit(valueType, false);
+    boolean ownsHasAnnotations(AttributeTypeImpl attributeType, Set<TypeQLToken.Annotation> annotations) {
+        return getSupertypes().anyMatch(ownerType -> attributeType.getSupertypes().anyMatch(ownedType -> {
+            TypeEdge edge;
+            edge = ownerType.vertex.outs().edge(OWNS, ownedType.vertex);
+            if (edge == null) edge = ownerType.vertex.outs().edge(OWNS_KEY, ownedType.vertex);
+            return edge != null && edge.annotations().containsAll(annotations);
+        }));
```
krishnangovindraj, line 463: Function to filter on annotations.
Not very efficient because we apply this to every type returned by `getOwns`/`getOwners`

lveillard:
I see that it is too late, but there is one related issue that could be interesting
https://github.com/vaticle/typedb/issues/6658

Prisma has a great way to differentiate between @unique and @keys, which also enables **composite Ids**

flyingsilverfin:
TODO: optimise data verification during a key or unique ownership

flyingsilverfin:
> 

I think that is not in scope for this change :) Good to know people are watching our work ;)

flyingsilverfin (approved):

PR #6783, opened 2023-04-05
Revise cost model for reasoner planner

krishnangovindraj:
We improve the cost-model for recursive rules from a 'worst-case' estimate to an 'average-case' one, enabling better informed planning. We also introduce a cost for the combining the results of reasoning within a conjunction - discouraging generating large permutations of cheap answers.

* Revises the upper-bound for the number of recursively inferable things from the product of the number of the vertices which could be involved to the square root of the product.
* Whenever we add a resolvable to a plan, we only used cost of evaluating that resolvable. Now, we also add the cost of combining the answers of the resolvable to the current answer sets the unextended plan would produce - This is equal to the estimate for all variables produced by the extended plan.

### DIFF
--- a/reasoner/planner/AnswerCountEstimator.java
+++ b/reasoner/planner/AnswerCountEstimator.java
@@ -165,7 +165,7 @@ public class AnswerCountEstimator {
             Variable bestScalingVar = scale.second().orElse(null);
             // Find scaling factor
             for (Variable v : model.variables) {
-                double ans = (double) Math.max(1,model.estimateAnswers(set(v)));
+                double ans = (double) Math.max(1, model.estimateAnswers(set(v)));
                 if (minVariableEstimate.containsKey(v) && minVariableEstimate.get(v) / ans < bestScaler) {
                     bestScaler = minVariableEstimate.get(v) / ans;
                     bestScalingVar = v;
@@ -228,6 +228,10 @@ public class AnswerCountEstimator {
             return Math.round(Math.ceil(answerEstimateFromCover(variables, cover)));
         }
 
+        public double answerSetSize() {
+            return answerEstimate(minVariableEstimate.keySet());
+        }
+
         private static double scaledEstimate(LocalModel model, Pair<Double, Optional<Variable>> scale, Set<Variable> estimateVariables) {
             assert scale.second().isPresent() || scale.first() == 1.0;
             Set<Variable> variables = new HashSet<>();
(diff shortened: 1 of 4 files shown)


reasoner/planner/AnswerCountEstimator.java
```diff
@@ -620,17 +629,31 @@ private LocalModel.RelationModel modelForRelation(RelationConstraint relationCon
...
+
+            double hasEdgeEstimate;
             if (correspondingConcludable != null) {
-                hasEdgeEstimate += estimateInferredAnswerCount(correspondingConcludable, set(hasConstraint.owner(), hasConstraint.attribute()));
                 attributeEstimate += attributesCreatedByExplicitHas(correspondingConcludable);
-                if (attributeEstimate != 0 && ownerEstimate < hasEdgeEstimate / attributeEstimate) {
+
+                double inferredHasEdgeCount = estimateInferredAnswerCount(correspondingConcludable, set(hasConstraint.owner(), hasConstraint.attribute()));
+
+                if (attributeEstimate != 0 && ownerEstimate < inferredHasEdgeCount / attributeEstimate) {
```
benjaminasmall, line 642: I see this existed before this change, but it always scares me to compare `double`s with `==0`.

krishnangovindraj: Makes sense. It used to be a long but I didn't update these.
I can do a sweep now. Is there an easier way to do it than `Math.abs(x-y) < EPS` ?

benjaminasmall: Only if you're confident about the sign.

krishnangovindraj: The sign of what?

benjaminasmall: Of `x-y`, such that you don't have to take the `abs()`. For example, in these cases, `attributeEstimate > 0.0` or `attributeEstimate > -epsilon` might suffice.

reasoner/planner/OrderingCoster.java
```diff
@@ -79,6 +84,11 @@ LocalAllCallsCosting createAllCallsCosting(ReasonerPlanner.CallMode callMode, Li
                 double allAnswersForUnrestrictedMode = answerCountEstimator.localEstimate(conjunctionNode.conjunction(), resolvable, resolvableMode);
                 double cyclicScalingFactor = projectionVars.isEmpty() && allAnswersForUnrestrictedMode != 0 ? 0.0 :
                         (double) estimator.answerEstimate(projectionVars) / allAnswersForUnrestrictedMode;
+
+                // Terrible overestimate. Let's take the square-root to be more realistic.
+                if (allAnswersForUnrestrictedMode != 0) {
+                    cyclicScalingFactor = Math.min(cyclicScalingFactor, Math.sqrt(allAnswersForUnrestrictedMode) / allAnswersForUnrestrictedMode);
```
krishnangovindraj, line 90: Need to decide if this is still needed, given we've revised the answer count itself.

flyingsilverfin (approved):

PR #6784, opened 2023-04-05
Bump Protocol and Common

jamesreprise:

We've bumped protocol and common for changes required downstream in Cluster.

We've bumped protocol and common.

### DIFF
--- a/dependencies/vaticle/repositories.bzl
+++ b/dependencies/vaticle/repositories.bzl
@@ -32,18 +32,26 @@ def vaticle_typeql():
     )
 
 def vaticle_typedb_common():
-    git_repository(
+    native.local_repository(
         name = "vaticle_typedb_common",
-        remote = "https://github.com/vaticle/typedb-common",
-        tag = "2.16.0",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_common
+        path = "../typedb-common",
     )
+#    git_repository(
+#        name = "vaticle_typedb_common",
+#        remote = "https://github.com/vaticle/typedb-common",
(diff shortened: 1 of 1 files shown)


server/parameters/util/YAMLParser.java
```diff
@@ -398,5 +403,38 @@ private static boolean isValidSizeString(String size) {
...
+            private static final Pattern DURATION_PATTERN = Pattern.compile(LENGTH_PART + UNIT_PART, Pattern.CASE_INSENSITIVE);
+
+            private static long parse(String duration) {
+                Matcher matcher = DURATION_PATTERN.matcher(duration);
+                long coefficient;
+                if (matcher.matches()) {
+                    String lenStr = matcher.group(LENGTH_GROUP);
+                    String unitStr = matcher.group(UNIT_GROUP);
+                    long lenValue = Long.parseLong(lenStr);
+                    if (unitStr.equalsIgnoreCase("w")) coefficient = com.vaticle.typedb.core.common.collection.Duration.WEEK;
```
flyingsilverfin, line 422: can we avoid creating our own constants and either use java's `Duration` directly, or it's `ChronoUnit`?

jamesreprise: We can, but we'd have to be happy to not support weeks. Java doesn't feel that weeks is an exact duration:
```
external/vaticle_typedb/server/parameters/util/YAMLParser.java:424: error: [DurationTemporalUnit] Duration APIs only work for DAYS or exact durations.
                    if (unitStr.equalsIgnoreCase("w")) result = java.time.Duration.of(lenValue, ChronoUnit.WEEKS).toSeconds();
                                                                                     ^
```
https://errorprone.info/bugpattern/DurationTemporalUnit

PR #6785, opened 2023-04-05
Add configuration support for parsing Durations 

flyingsilverfin:

We've added support for parsing Durations to the YAML parser.

We've bumped protocol and common for changes required downstream in Cluster.

We've:
* added the `Duration` parser to `YAMLParser`, allowing for the parsing of durations.
* bumped protocol and common.


flyingsilverfin (commented):
Approved before, re-merging

PR #6790, opened 2023-04-12
Throw exception when attribute value type is modified

### LINKED ISSUE #6789: Modifying an attribute value type neither throws exception nor has effects
## Description
Trying to redefine an attribute type as follows throw an exception as expected:

```typeql
define 
redefine-me sub attribute, value long;
define
redefine-me value string;

# Exception:
# [TYW15] Invalid Type Write: An attribute value type (in this case 'STRING') can only be set onto an attribute type (in this case 'redefine-me') when it was defined for the first time.
```

However, the following query does not:

```typeql
define 
redefine-me sub attribute, value long;
define
redefine-me sub attribute, value string;

# Success
```
## Environment

1. OS (where TypeDB server runs): macOS Ventura { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. TypeDB version (and platform): 2.16.1
3. TypeDB client: BDD/Studio
4. Other environment details:

## Reproducible Steps
As in the description

## Expected Output
The following error message
```
 [TYW15] Invalid Type Write: An attribute value type (in this case 'STRING') can only be set onto an attribute type (in this case 'redefine-me') when it was defined for the first time.
```
## Actual Output
Query succeeds and commits.

krishnangovindraj:
Throw an exception when a `define` query attempts to modify the an existing the attribute type with a different value type.

* Throw an exception when a `define` query attempts to modify the redefine the attribute type with a different value type.

- fixes #6789

### DIFF
--- a/query/Definer.java
+++ b/query/Definer.java
@@ -195,7 +195,7 @@ public class Definer {
                 else if (!supertype.isRoot()) valueType = supertype.asAttributeType().getValueType();
                 else throw TypeDBException.of(ATTRIBUTE_VALUE_TYPE_MISSING, labelConstraint.label());
                 if (thingType == null) thingType = conceptMgr.putAttributeType(labelConstraint.label(), valueType);
-                else if (supertype.isRoot() && !valueType.equals(thingType.asAttributeType().getValueType())) {
+                else if (!valueType.equals(thingType.asAttributeType().getValueType())) {
                     throw TypeDBException.of(ATTRIBUTE_VALUE_TYPE_MODIFIED, valueType, labelConstraint.label());
                 }
                 thingType.asAttributeType().setSupertype(supertype.asAttributeType());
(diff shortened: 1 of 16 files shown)


query/Definer.java
```diff
@@ -195,6 +195,9 @@ private ThingType defineSub(ThingType thingType, SubConstraint subConstraint, Ty
                 else if (!supertype.isRoot()) valueType = supertype.asAttributeType().getValueType();
                 else throw TypeDBException.of(ATTRIBUTE_VALUE_TYPE_MISSING, labelConstraint.label());
                 if (thingType == null) thingType = conceptMgr.putAttributeType(labelConstraint.label(), valueType);
+                else if (supertype.isRoot() && !valueType.equals(thingType.asAttributeType().getValueType())) {
```
krishnangovindraj, line 198: If super type is not root, a more appropriate error is throw.

flyingsilverfin (approved):

PR #6793, opened 2023-04-14
Replace schema export formatter with TypeQL formatter

flyingsilverfin:

We replace the custom schema formatter in TypeDB with TypeQL's native pretty-print capability. This means that all changes to the standard way to format TypeQL schemas wil be automatically reflected upon updating the TypeQL dependency in the future.

* Remove all special formatting TypeDB used to pretty-print exported schema
* Parse and re-stringify the generated TypeDB schema using TypeQL, using pretty printing, to produce a well-formatted schema string
* Remove custom Rule export in favor of returning the Rule's persisted TypeQL definition
* Delete now-used methods and classes


database/BUILD
```diff
@@ -45,6 +45,7 @@ native_java_libraries(
     deps = [
         # External dependencies from Vaticle
         "@vaticle_typedb_common//:common",
+        "@vaticle_typeql//java:typeql-lang",
```
flyingsilverfin, line 48: Required in order for the schema export to be done in the `CoreDatabase` directly. I believe this is a sensible dependency to have and there is no good alternative location for this dependency to live.

Explored the alternative of having the parsing in `ConceptManager` and `LogicManager` separately, but this leads to having to combine the parsed definitions in a way that could affect correctness of the final schema, so we leave eliminate this option.

database/CoreDatabase.java
```diff
@@ -433,21 +434,21 @@ public Stream<TypeDB.Session> sessions() {
     @Override
     public String schema() {
         try (TypeDB.Session session = databaseMgr.session(name, DATA); TypeDB.Transaction tx = session.transaction(READ)) {
-            return "define\n\n" + tx.concepts().typesSyntax() + tx.logic().rulesSyntax();
+            return TypeQL.parseQuery("define\n\n" + tx.concepts().typesSyntax() + tx.logic().rulesSyntax()).toString(true);
```
flyingsilverfin, line 437: `true` makes enforces pretty-printing

PR #6794, opened 2023-04-18
Implement fast termination for reasoner

flyingsilverfin:

We force each transaction-bound reasoner instance to terminate rapidly when any exception occurs, such as those generated when the transaction is closed while queries are still executing. This fixes a bug where queued up reasoning work would continue running for some time even after transaction close. Previously, scenarios where many queries ran concurrently and over a longer period of time, the incurred resource consumption could add up to bottleneck the entire server.

* Introduce preemption (eg. skip the message queue) in the Actor model, by switching the internal actor queue to be a `BlockingDeque` instead of a `TransferQueue`
* Introduce a new `terminate` API on actors, which should generally be used with the new `Actor.Driver.executeNext()` execution mechanism.
* Tasks submitted to the actor executors automatically check whether they should execute, by checking if the actor that submitted them is terminated or not, allowing the messages in the queue to drain quickly.

### Performance implications:
We did a a few synthetic benchmarks to check whether the replacement of the `LinkedTransferQueue` is much slower than the `LinkedBlockingDeque` and found that:
1. Under no contention (single writer/single reader) the `LinkedTransferQueue` can be faster (variance is higher, but between -10 and +30% faster) compared to the `LinkedTransferDeque`
2. Under any contention, the `Deque` and the `Queue` are actually very comparable, and the `Deque` is sometimes faster.

These tradeoffs are acceptable for our use cases so we proceed with the `Deque` that provides the ability to insert at the start of the queue natively.

### DIFF
--- a/concurrent/actor/Actor.java
+++ b/concurrent/actor/Actor.java
@@ -17,6 +17,10 @@
 
 package com.vaticle.typedb.core.concurrent.actor;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
@@ -29,21 +33,27 @@ public abstract class Actor<ACTOR extends Actor<ACTOR>> {
     private static final String ERROR_ACTOR_DRIVER_IS_NULL = "driver() must not be null.";
     private final Driver<ACTOR> driver;
     private final Supplier<String> debugName;
+    boolean isTerminated;
 
     public static <A extends Actor<A>> Driver<A> driver(Function<Driver<A>, A> actorFn, ActorExecutorGroup service) {
         return new Driver<>(actorFn, service);
     }
 
-    protected abstract void exception(Throwable e);
-
     protected Actor(Driver<ACTOR> driver, Supplier<String> debugName) {
         this.driver = driver;
         this.debugName = debugName;
+        this.isTerminated = false;
     }
 
     public Driver<ACTOR> driver() {
         assert this.driver != null : ERROR_ACTOR_DRIVER_IS_NULL;
         return this.driver;
+    }
+
+    protected abstract void exception(Throwable e);
+
+    public void terminate(@Nullable Throwable cause) {
+        this.isTerminated = true;
     }
 
     public Supplier<String> debugName() {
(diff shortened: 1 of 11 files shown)


reasoner/processor/AbstractProcessor.java
```diff
@@ -156,27 +153,7 @@ public void onFinished(Identifier finishable) {
...
-    }
-
-    public void terminate(Throwable cause) {
-        LOG.debug("Actor terminated.", cause);
-        this.terminated = true;
-    }
-
-    private boolean isTerminated() {
-        return terminated;
+        controller.executePreemptive(controller -> controller.exception(e));
```
flyingsilverfin, line 156: we let the `Controller` plane decide what to do with each exception - they will propagate termination commands to the registry (eg. other controllers) and their processors (including this one).

reasoner/controller/ConcludableController.java
```diff
@@ -83,7 +84,6 @@ public void setUpUpstreamControllers() {
 
     @Override
     public void routeConnectionRequest(REQ req) {
-        if (isTerminated()) return;
```
flyingsilverfin, line 86: we no longer need manual check for whether the actor is terminated, messages are ignored if the actor is terminated at the higher level now

concurrent/actor/ActorExecutor.java
```diff
@@ -52,8 +53,7 @@ public class ActorExecutor {
     public ActorExecutor(ThreadFactory threadFactory, Supplier<Long> clock) {
         this.thread = threadFactory.newThread(this::run);
         this.clock = clock;
-        // TODO: Benchmark and verify that LinkedTransferQueue is actually most performant
-        submittedTasks = new LinkedTransferQueue<>();
+        submittedTasks = new LinkedBlockingDeque<>();
```
flyingsilverfin, line 56: We hope that this queue is performant enough for our purposes!

concurrent/actor/Actor.java
```diff
@@ -89,20 +109,28 @@ public CompletableFuture<Void> complete(Consumer<ACTOR> consumer) {
...
+                        () -> {
+                            if (!actor.isTerminated) future.complete(function.apply(actor));
+                        },
+                        e -> {
+                            actor.exception(e);
+                            future.completeExceptionally(e);
+                        }
+                );
+                return future;
+            } else return null;
```
flyingsilverfin, line 124: we could debate whether this should return `null` or a `CompletedFuture` - what are the semantics of submitted a job to be run on a terminated Actor?

PR #6795, opened 2023-04-20
Disallow deletion of root types

dmitrii-ubskii:

We override `delete()` for all `..Type.Root` classes to throw error `[TYW01] Root types are immutable.`

Side note: `undefine entity sub thing;` and similar queries fail already, because despite `entity` being a subtype of `thing`, root entity type's `getSupertypes()` method (which is used to validate type deletion) only returns the root type itself. The query fails with error `TYW38` `The type 'entity' cannot be undefined, as the provided supertype 'thing' is not a valid supertype.`

### DIFF
--- a/concept/type/impl/TypeImpl.java
+++ b/concept/type/impl/TypeImpl.java
@@ -47,6 +47,7 @@ import static com.vaticle.typedb.core.common.exception.ErrorMessage.ThingWrite.I
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Transaction.SESSION_SCHEMA_VIOLATION;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.TypeRead.INVALID_TYPE_CASTING;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.TypeWrite.CYCLIC_TYPE_HIERARCHY;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.TypeWrite.ROOT_TYPE_MUTATION;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.TypeWrite.TYPE_HAS_BEEN_DELETED;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.TypeWrite.TYPE_REFERENCED_IN_RULES;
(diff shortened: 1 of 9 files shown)


dependencies/vaticle/repositories.bzl
```diff
@@ -48,8 +48,8 @@ def vaticle_typedb_protocol():
 def vaticle_typedb_behaviour():
     git_repository(
         name = "vaticle_typedb_behaviour",
-        remote = "https://github.com/vaticle/typedb-behaviour",
-        commit = "04ec1f75cc376524b3cac75af37e1716230845e0",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
+        remote = "https://github.com/dmitrii-ubskii/typedb-behaviour",
+        commit = "aa03cf16cd9b2b2cf2600778d58763dbdf7e56a3",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
```
dmitrii-ubskii, line 52: Pending https://github.com/vaticle/typedb-behaviour/pull/245

test/behaviour/config/Parameters.java
```diff
@@ -130,7 +130,8 @@ public List<Arguments.Transaction.Type> transaction_types(List<String> values) {
     public enum RootLabel {
         ENTITY("entity"),
         ATTRIBUTE("attribute"),
-        RELATION("relation");
+        RELATION("relation"),
+        THING("thing");
```
dmitrii-ubskii, line 134: Added solely for us to be able to do `Then delete thing type: thing; throws exception` in BDD tests.

concept/type/impl/TypeImpl.java
```diff
@@ -149,6 +150,7 @@ <TYPE extends TypeImpl> Forwardable<TYPE, Order.Asc> getSubtypesExplicit(Functio
     }
 
     void validateDelete() {
+        if (isRoot()) throw exception(TypeDBException.of(ROOT_TYPE_MUTATION));
```
dmitrii-ubskii, line 153: Not strictly necessary as long as all `Root` types are covered, but I do think that having this unconditional validation step is nice to have.

flyingsilverfin (approved):

PR #6800, opened 2023-04-26
Reasoner benchmarking based on the demo IAM schema

krishnangovindraj:

Introduce tests based on queries from the TypeDB IAM example to measure reasoner performance and flag regressions. We include two types of tests:

1. basic tests: these are simple baselines for the reasoner such as inferring relations, attribute ownerships, etc.
2. complex tests: these are designed to target particularly difficult spots of the reasoning engine, such as query planning, large state spaces, complex rule graphs, and more.

The 'test' portion of these benchmarks are designed to pass/fail based on the amount of state & work created in the execution of the benchmark, rather than time, as these should be stable across machine types.

Introduce generalised `common/PerfCounters` to measure metrics in the system.
- Can be enabled or disabled through options.
- A `PerfCounters` object holds multiple counters.
- A counter must be registered to a `PerfCounters` object with a name.
  - this returns a counter exposing an `add(long delta)` method, and get method. 
  - The counting is done using an AtomicLong.

We create a transaction-bound set of performance counters for reasoning.
  - The `ReasonerPerfCounters` count planning time, conjunction branches, sub-goals, materialisations,
  - They can be enabled or disabled through the config file
  - When enabled, they will print the counters at the end of reasoning

Create`test/benchmark/reasoner/iam/*`
* Add dataset schema & data in `resources` folder
  - Loaded data using the DataImporter.
  - Tests of basic query shapes
     - Multiple starting points, 
     - high-arity concludable
  - Tests characterised by data/result size
    - Individual result sets combine combinatorially
    - Individual result sets combine linearly due to high selectivity.
  - Tests characterised by rule graph features
    - Mutually recursive rules resulting in combinatorial proofs 
  - Tests For language features
    - Inferred attribute, ownership, relation, relation-with-ownership,  
    - Filter attributes on value and not type 
    - variabilised rules
  - Basic tests of fundamental constructs
    - inferred ownerships, inferred relations, inferred ownerships on inferred relations
    - Relations queried by role-player, by relation variable.
    - Unpacking queries on bound relation and bypassing reasoning 
  - Basic tests based on joins of different structures
    - A x B
    - (A x B) x (B x C)
    - (A x B) x (A x B)
* Assert run time, planning time and certain perf-counters against a maximum acceptable value.


test/benchmark/iam/Benchmark.java
```diff
@@ -0,0 +1,115 @@
...
+                run.reasonerPerfCounters.get(counter.name()) <= maxValue));
+    }
+
+    void mayPrintResults(CSVResults printTo) {
+        if (printTo != null) {
+            printTo.append(this);
+        }
+    }
+
+    static class CSVResults {
```
flyingsilverfin, line 81: I think this is actually a `CSVBuilder` and you can convert the `flush()` to return a string - then you don't need the `out` variable at all?

flyingsilverfin (changes requested):
Okay, we're getting there! Good work :)

reasoner/processor/AbstractProcessor.java
```diff
@@ -182,6 +186,36 @@ public Driver<Monitor> monitor() {
...
+            return perfCounters;
+        }
+
+        public static class ReasonerPerfCounters {
+
+            public static final String KEY_TIME_PLANNING_MS = "time_planning_ms";
+            public static final String KEY_COUNT_MATERIALISATIONS = "count_materialisations";
+            public static final String KEY_COUNT_CONJUNCTION_PROCESSORS = "count_conjunction_processors";
+            public static final String KEY_COUNT_COMPOUND_STREAMS = "count_compound_streams";
+
```
krishnangovindraj, line 199: Moved perf-counters here and made the keys static strings. Could also make them enums?

test/benchmark/iam/BenchmarkRunner.java
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        private void appendLine(List<String> entries) {
+            entries.forEach(entry -> sb.append(entry).append(","));
+            sb.append("\n");
+        }
+
+        public String build() {
+            return sb.toString();
+        }
```
krishnangovindraj, line 197: Renamed class to CSV builder and renamed method to build.

test/benchmark/iam/BenchmarkRunner.java
```diff
@@ -0,0 +1,199 @@
...
+        Files.createDirectory(dataDir);
+
+        databaseMgr = CoreDatabaseManager.open(new Options.Database().dataDir(dataDir).storageDataCacheSize(MB).storageIndexCacheSize(MB));
+        databaseMgr.create(database);
+    }
+
+    void tearDown() {
+        databaseMgr.close();
+        if (this.csvBuilder != null) System.out.println(csvBuilder.build());
```
krishnangovindraj, line 79: Prints at tearDown

test/benchmark/iam/ComplexRuleGraphTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junit.Test;
+
+import java.io.IOException;
+
+public class ComplexRuleGraphTest {
+
+    private static final String database = "iam-benchmark-rules";
+    private static final BenchmarkRunner benchmarker = new BenchmarkRunner(database);
+
+    @BeforeClass
```
krishnangovindraj, line 33: schema/data is loaded once before the tests.

krishnangovindraj (commented):
Flag changes in response to feedback.

reasoner/controller/RootConjunctionController.java
```diff
@@ -49,7 +51,9 @@
 
     @Override
     public void initialise() {
+        long start = PerfCounters.getNanos();
```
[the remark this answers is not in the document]
flyingsilverfin: Let's make sure we implement this change - we don't need `getNanos` on `PerfCounters` for no reason

reasoner/controller/RootConjunctionController.java
```diff
@@ -49,7 +50,9 @@
...
         planner().plan(conjunction, new HashSet<>());
+        processorContext().perfCounters().timePlanningMs.add((PerfCounters.getNanos() - start) / 1_000_000);
```
flyingsilverfin, line 55: as noted before, let's not do any division because it's going to be non-performant. Keep it in nanos

reasoner/controller/RootDisjunctionController.java
```diff
@@ -48,7 +49,9 @@
 
     @Override
     public void initialise() {
+        long start = PerfCounters.getNanos();
         disjunction.conjunctions().forEach(conjunction -> planner().plan(conjunction, new HashSet<>()));
+        processorContext().perfCounters().timePlanningMs.add((PerfCounters.getNanos() - start) / 1_000_000);
```
flyingsilverfin, line 54: I wonder -do you just want to move the perf counter timing into the `Planner` itself? Then you don't need to be careful about adding it around every planner call.

reasoner/processor/AbstractProcessor.java
```diff
@@ -182,6 +185,36 @@ public Driver<Monitor> monitor() {
...
+        public ReasonerPerfCounters perfCounters() {
+            return perfCounters;
+        }
+
+        public static class ReasonerPerfCounters {
+
+            public static final String KEY_TIME_PLANNING_MS = "time_planning_ms";
+            public static final String KEY_COUNT_MATERIALISATIONS = "count_materialisations";
+            public static final String KEY_COUNT_CONJUNCTION_PROCESSORS = "count_conjunction_processors";
+            public static final String KEY_COUNT_COMPOUND_STREAMS = "count_compound_streams";
```
flyingsilverfin, line 197: you don't need to prefix this with `KEY_`, remove.
Also, you should suffix `_COUNT` instead of having them be a prefix (see how you did `_MS`) above.

reasoner/processor/AbstractProcessor.java
```diff
@@ -182,6 +185,36 @@ public Driver<Monitor> monitor() {
...
+                perfCounters = new PerfCounters(enabled);
+                timePlanningMs = perfCounters.register(KEY_TIME_PLANNING_MS);
+                countMaterialisations = perfCounters.register(KEY_COUNT_MATERIALISATIONS);
+                countConjunctionProcessors = perfCounters.register(KEY_COUNT_CONJUNCTION_PROCESSORS);
+                countCompoundStreams = perfCounters.register(KEY_COUNT_COMPOUND_STREAMS);
+            }
+
+            public Map<String, Long> snapshotUnsynchronised() {
+                return perfCounters.snapshotUnsynchronised();
+            }
```
flyingsilverfin, line 216: I've changed my mind - think this class should actually extend the common `PerfCounter` class. Sorry! 

Then we can get rid of the duplication and the awkwardness that a `PerfCounters` holds onto a `PerfCounters` object. However, if it was initialised with a set of `Counter` directly (eg. doesn't create a `new PerfCounters()` itself) then I would still think it is better not extending `PerfCounters`

test/benchmark/iam/BenchmarkRunner.java
```diff
@@ -0,0 +1,211 @@
...
+        new DataImporter(databaseMgr, database, Paths.get(RESOURCE_DIRECTORY + filename), Version.VERSION).run();
+    }
+
+    void warmUp() {
+        long start = System.nanoTime();
+        try (TypeDB.Session session = dataSession()) {
+            try (TypeDB.Transaction tx = session.transaction(Arguments.Transaction.Type.READ, new Options.Transaction().infer(true))) {
+                AnswerCountEstimator estimator = new AnswerCountEstimator(tx.logic(), tx.concepts().graph(), new ConjunctionGraph(tx.logic()));
+                tx.logic().rules().flatMap(rule -> iterate(rule.condition().disjunction().conjunctions()))
+                        .forEachRemaining(conjunction -> estimator.buildConjunctionModel(conjunction));
```
flyingsilverfin, line 118: this will get cached so there's basically no more planning to do except the search right?

separately, why are you just warming up the planner objects and none of the other things? Why not run some queries? That is a much bigger code path to warm up OS/JVM for...

reasoner/processor/AbstractProcessor.java
```diff
@@ -182,6 +185,36 @@ public Driver<Monitor> monitor() {
             return monitor;
         }
 
+        public ReasonerPerfCounters perfCounters() {
+            return perfCounters;
+        }
+
+        public static class ReasonerPerfCounters {
```
flyingsilverfin, line 192: Ok so after our discussion on usage of the Perf Counters, let's go back to your previous architecture: move this to `//reasoner/common/PerfCounters`. Then you can create the `PerfCounters` in the top-level `Reasoner` entry point, and pass it into the `ControllerRegistry`, which passes it through to the `Controller` and `Processor` contexts. Then we have wider access to it, such as being able to pass it into `Planner`.

flyingsilverfin (changes requested):

test/benchmark/iam/BenchmarkRunner.java
krishnangovindraj: Now I remember why I tried to do this - to cache the statistics.
But looking at the code again, it looks like the cache is a member of ThingGraph, which is transaction bound and won't make a difference?

reasoner/processor/AbstractProcessor.java
krishnangovindraj: I'm currently creating ReasonerPerfCounters as an extension of PerfCounters, but we can remove the PerfCounter*s* class and just keep the counters themselves if we don't see them as a tightly grouped set.

reasoner/controller/ControllerRegistry.java
```diff
@@ -124,6 +125,14 @@ public void terminate(Throwable cause) {
         }
     }
 
+    public ReasonerPlanner planner() {
+        return controllerContext.planner();
+    }
```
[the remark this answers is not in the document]
krishnangovindraj: deleted

krishnangovindraj (commented):
replies

reasoner/planner/OrderingCoster.java
```diff
@@ -94,7 +94,7 @@ LocalAllCallsCosting createAllCallsCosting(ReasonerPlanner.CallMode callMode, Li
 
             estimator.extend(resolvable);
 
-            boolean isConnectedToInput = !Collections.intersection(resolvableMode, inputConnectedVars).isEmpty();
+            boolean isConnectedToInput = callMode.mode.isEmpty() || !Collections.intersection(resolvableMode, inputConnectedVars).isEmpty();
```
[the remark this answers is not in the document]
krishnangovindraj: I think it won't have had much of an effect, because the scaling factor in the case of empty bounds always 1 so the disconnected and acyclic costs should have equal weightage.

reasoner/processor/AbstractProcessor.java
flyingsilverfin: right but we can still put the unit at the end - PROCESSOR_CONJUNCTION_COUNT and PLANNER_PLANNING_TIME_MS

test/benchmark/iam/LanguageFeatureTest.java
```diff
@@ -0,0 +1,179 @@
...
+    private static final BenchmarkRunner benchmarker = new BenchmarkRunner(database);
+
+    public LanguageFeatureTest() { }
+
+    @BeforeClass
+    public static void setup() throws IOException {
+        benchmarker.setUp();
+        benchmarker.loadSchema("schema_types.tql");
+        benchmarker.loadSchema("schema_rules_optimised.tql");
+        benchmarker.loadSchema("schema_rules_test_specific.tql");
```
flyingsilverfin, line 40: is there way to factor these schema files into a set of "common" types and rules, and then have one set of rules for `schema_rules_language_feature.tql` which matches this test name, then another for `conjunction_structure`, etc.?

test/benchmark/iam/LanguageFeatureTest.java
```diff
@@ -0,0 +1,179 @@
...
+    @After
+    public void reset() {
+        benchmarker.reset();
+    }
+
+    @Test
+    public void testOwnership() {
+        String query = "match\n" +
+                "$x has parent-company-name $n;\n";
+        Benchmark benchmark = new Benchmark("ownership", query, 0); // TODO: Data only exists in the full data-set
```
flyingsilverfin, line 59: what are these TODOS? What's the full data set?

test/benchmark/iam/LanguageFeatureTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+    @Test
+    public void testRelationWithoutRolePlayer() {
+        String query = "match\n" +
+                "$p isa permission;\n";
+        Benchmark benchmark = new Benchmark("relation-without-role-players", query, 2783);
+        benchmarker.runBenchmark(benchmark);
+
+        benchmark.assertAnswerCountCorrect();
+        benchmark.assertRunningTime(5000);
```
flyingsilverfin, line 102: this is already a pretty interesting test from the baseline! This means we're doing approximately 2700 answers in 5 seconds, or about 1 answer every 2 milliseconds. For simple inference I'd hope we can get this to 10x faster in Rust.

flyingsilverfin: These are the kinds of high-level pointers that is exactly the point of a simple inference test (to address your push back on "whats the point" of these tests).

test/benchmark/iam/LanguageFeatureTest.java
```diff
@@ -0,0 +1,179 @@
...
+    @Test
+    public void testInferredRelationAndOwnership() {
+        String query = "match\n" +
+                "$p (subject: $s, access: $ac) isa permission, has validity $v;\n";
+        Benchmark benchmark = new Benchmark("inferred-relation-and-ownership", query, 2783);
+        benchmarker.runBenchmark(benchmark);
+
+        benchmark.assertAnswerCountCorrect();
+        benchmark.assertRunningTime(10_000);
+        benchmark.assertCounters(500, 5677, 10000, 15000);
```
flyingsilverfin, line 115: As another point, we get to get some high-level summaries of how much state we are creating and can assess that against if we think it's 'reasonable'. 15k streams for 2700 answers... That seems like a lot! Something we might look into in the future to see if we can reduce state somehow.

test/benchmark/iam/LanguageFeatureTest.java
```diff
@@ -0,0 +1,179 @@
...
+                "$f isa file, has $path; $path = \"root/engineering/typedb-studio/src/README.md\";\n" +
+                "$o isa operation, has $operation; $operation = \"edit file\";\n" +
+                "$a (object: $f, action: $o) isa access;\n" +
+                "$pe (subject: $p, access: $a) isa permission, has validity true;\n";
+        Benchmark benchmark = new Benchmark("value-predicate-filtering", query, 1);
+        benchmarker.runBenchmark(benchmark);
+
+        benchmark.assertAnswerCountCorrect();
+        benchmark.assertRunningTime(2500);
+        benchmark.assertCounters(1000, 200, 500, 1500);
```
flyingsilverfin, line 150: 1500 compound streams to find 1 answer in 2.5 seconds! Seems like there's gotta be a better way to do this one too. 

I hope you start to see the point of why we wanted these!

test/benchmark/iam/LanguageFeatureTest.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+
```
flyingsilverfin, line 153: careful with hanging newlines

test/benchmark/iam/LargeDataTest.java
```diff
@@ -0,0 +1,105 @@
...
+package com.vaticle.typedb.core.reasoner.benchmark.iam;
+
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.io.IOException;
+
+public class LargeDataTest {
```
flyingsilverfin, line 28: Do these fall under 'language' tests or the literature-derived/analytically derived ones?

test/integration/reasoner/planner/RecursivePlannerTest.java
```diff
@@ -104,7 +106,7 @@ public void tearDown() {
     @Test
     public void test_single_retrieval() {
         initialise(Arguments.Session.Type.DATA, Arguments.Transaction.Type.READ);
-        RecursivePlanner planSpaceSearch = RecursivePlanner.create(transaction.traversal(), transaction.concepts(), transaction.logic());
+        RecursivePlanner planSpaceSearch = RecursivePlanner.create(transaction.traversal(), transaction.concepts(), transaction.logic(), new ReasonerPerfCounters(false));
```
flyingsilverfin, line 109: This is a nicer signature than passing a random boolean in with a bunch of higher-level components!

flyingsilverfin:
Let's use `data_large.typedb` instead of `data.big.typedb

flyingsilverfin (changes requested):

flyingsilverfin:
There's also a (heavy?) rebase to do? Not sure how difficult the conflicts will be...

test/benchmark/iam/LanguageFeatureTest.java
krishnangovindraj: This isn't actually a simple baseline test.
Computing permissions recursively computes 3 transitive relations and is the difficult part of the iam-dataset.
There's no simple relation to infer in the IAM dataset. 
My question was never, "What's the point of simple tests?". It was: "If we do want baseline tests, why not add them in the existing reasoner benchmark?" as opposed to trying to awkwardly fit them into the iam dataset

test/benchmark/iam/LanguageFeatureTest.java
krishnangovindraj: There's one stream per branch in the conjunction. 
I already have a branch where we can re-use streams, but I don't know how widely applicable it is - we just need this benchmark done so I can justify the trade-off.

(Along with local-doneness, Trying to replace these streams with recompilation is a major goal for the reasoner rewrite)

test/benchmark/iam/LanguageFeatureTest.java
krishnangovindraj: The number of answers is in no way correlated to the amount of work to be done.

This query is equivalent to "Is there a path between A and B?" which has one answer, but requires a traversal of the whole graph.

test/benchmark/iam/LargeDataTest.java
krishnangovindraj: Literature.

test/integration/reasoner/planner/RecursivePlannerTest.java
krishnangovindraj: That was on the other branch. It will still be re-introduced after the rebase
.

test/benchmark/iam/LanguageFeatureTest.java
krishnangovindraj: Materialisations + conjunction processors + compound streams is meant to reflect the memory usage. We're critically missing the intermediate answers created/buffered in the compound streams, which don't show up anywhere else.
I'll add that in somewhere.

reasoner/processor/AbstractProcessor.java
krishnangovindraj: Refactored and omitted the count because they're 'counters' anyway.

test/benchmark/iam/BenchmarkRunner.java
flyingsilverfin: I think there are statistics per-transaction and the ones from before the snapshot so I think this is still useful. The new initialisation looks good.

test/benchmark/iam/LanguageFeatureTest.java
flyingsilverfin: It's true, but maybe there's a more optimal way to find that 1 answer

krishnangovindraj: Possibly - the planner picks the best plan to compute a fraction of all answers. We could also see it as computing a multiple of a a single answer.

But in this case, I think it's somewhat unlikely there's a significantly better way to do it.

test/benchmark/reasoner/iam/basic/BUILD
```diff
@@ -0,0 +1,30 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+load("@vaticle_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
+load("@vaticle_dependencies//builder/java:rules.bzl", "host_compatible_java_test", "native_java_libraries")
+
+host_compatible_java_test(
+    name = "basic-test",
```
flyingsilverfin, line 22: All our test targets are prefixed with `test-` rather than suffixed with it. Please change throughout

test/benchmark/reasoner/iam/basic/BasicTest.java
```diff
@@ -0,0 +1,224 @@
...
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class BasicTest {
```
flyingsilverfin, line 33: This looks like what we wanted! Good!

test/benchmark/reasoner/iam/basic/BasicTest.java
```diff
@@ -0,0 +1,224 @@
...
+                "match\n" +
+                        "$a isa object, has id \"%s\";\n" +
+                        "(start: $a, end: $b) isa object-pair;\n" +
+                        "(start: $b, end: $c) isa object-pair;\n",
+                queryParams.basicTestObject);
+        Benchmark benchmark = new Benchmark("double-join", query, NOBJECTS * NOBJECTS);
+        benchmarker.runBenchmark(benchmark);
+
+        benchmark.assertAnswerCountCorrect();
+        benchmark.assertRunningTime(5000);
```
flyingsilverfin, line 159: looks like we're at 2 milliseconds per answer here again huh

test/benchmark/reasoner/iam/common/BUILD
```diff
@@ -0,0 +1,51 @@
...
+
+load("@vaticle_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
+load("@vaticle_dependencies//builder/java:rules.bzl", "native_java_libraries")
+
+package(default_visibility = [
+    "//test/benchmark/reasoner/iam:__subpackages__",
+])
+
+native_java_libraries(
+    name = "benchmark-lib",
```
flyingsilverfin, line 26: you don't need the `-lib` suffix, just call it `common` - to match the package name.

test/benchmark/reasoner/iam/complex/ConjunctionStructureTest.java
```diff
@@ -0,0 +1,109 @@
...
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class ConjunctionStructureTest {
+    static final Path RESOURCE_DIRECTORY =  Paths.get("test", "benchmark", "reasoner", "iam", "complex");
+    private static final Path COMMON_RESOURCE_DIR = Paths.get("test", "benchmark", "reasoner", "iam", "resources");
+
+    private static final String database = "iam-benchmark-conjunctions";
+    private static final BenchmarkRunner benchmarker = new BenchmarkRunner(database);
```
flyingsilverfin, line 38: I think a better name would be `runner` or `benchmarkRunner`, we don't really have the convention of verbalising nounds with '-er' suffixes

test/benchmark/reasoner/iam/basic/BasicTest.java
krishnangovindraj: Yes, but also no.
There's still too much variance from the warmup for this to be indicative of the actual performance.
I'm just setting the bounds loosely till I get warmup to work better

test/benchmark/reasoner/iam/common/Benchmark.java
```diff
@@ -0,0 +1,125 @@
...
+import static com.vaticle.typedb.core.common.iterator.Iterators.iterate;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class Benchmark {
+
+    private static final int DEFAULT_NRUNS = 5;
+    private static final double COUNTER_LOWER_MARGIN = 0.75;
+    private static final double COUNTER_UPPER_MARGIN = 1.25;
+    private static final double RUNNING_TIME_UPPER_MARGIN = 3.0;
```
krishnangovindraj, line 39: Adjusted these values. The running time in particular has a large margin because the numbers were adjusted for my local performance and CI seems to be unto 3x slower.

.factory/automation.yml
```diff
@@ -122,6 +122,20 @@ build:
...
+#        branch: [master, development]
+      dependencies: [ build ]
+      command: |
+        bazel run @vaticle_dependencies//tool/bazelinstall:remote_cache_setup.sh
+        bazel test //test/benchmark/reasoner/iam/basic:test-basic --test_output=streamed
+        bazel test //test/benchmark/reasoner/iam/complex:test-conjunction-structure --test_output=streamed
+        bazel test //test/benchmark/reasoner/iam/complex:test-complex-rule-graph --test_output=streamed
+        bazel test //test/benchmark/reasoner/iam/complex:test-language-features --test_output=streamed
+        bazel test //test/benchmark/reasoner/iam/complex:test-large-data --test_output=streamed
+        bazel test //test/benchmark/reasoner/iam/complex:test-real-queries --test_output=streamed
```
krishnangovindraj, line 138: tests

.factory/automation.yml
```diff
@@ -122,6 +122,20 @@ build:
         bazel test //test/behaviour/reasoner/tests/type_hierarchy:test --test_output=streamed
         bazel test //test/behaviour/reasoner/tests/value_predicate:test --test_output=streamed
         bazel test //test/behaviour/reasoner/tests/variable_roles:test --test_output=streamed
+    test-benchmark-reasoner:
+      image: vaticle-ubuntu-22.04
+#      filter:
+#        owner: vaticle
+#        branch: [master, development]
```
krishnangovindraj, line 129: Need to enable the filter once we're convinced it's reasonably reliable.

krishnangovindraj (commented):
changes

krishnangovindraj (commented):
changes

.factory/automation.yml
```diff
@@ -122,6 +122,20 @@ build:
...
+      dependencies: [ build ]
+      command: |
+        bazel run @vaticle_dependencies//tool/bazelinstall:remote_cache_setup.sh
```
flyingsilverfin, line 132: remove

common/perfcounter/PerfCounters.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private final Function<String, Counter> counterConstructor;
+    private final Collection<Counter> counters;
+    private final boolean enabled;
+
+    public PerfCounters(boolean enabled) {
+        this.enabled = enabled;
+        this.counterConstructor = enabled ? AtomicLongCounter::new : NoOpCounter::new;
+        this.counters = new ConcurrentLinkedQueue<>();
```
flyingsilverfin, line 43: use a `ConcurrentSet`, which is instantiated as a `ConcurrentHashSet`

server/parameters/config.yml
```diff
@@ -49,10 +49,14 @@ log:
...
-    reasoner:
+    reasoner-tracer:
       enable: false
-      type: reasoner
+      type: reasoner-tracer
       output: file
+    reasoner-perfcounters:
+      enable: false
+      type: reasoner-perfcounters
+      output: stdout
```
krishnangovindraj: rename reasoner to reasoner-tracer;  Introduce reasoner-perfcounters

common/parameters/Options.java
```diff
@@ -103,6 +105,17 @@ public SELF explain(boolean explain) {
...
+        if (reasonerPerfCounters != null) return reasonerPerfCounters;
+        else if (parent != null) return parent.reasonerPerfCounters();
+        else return DEFAULT_PERFCOUNTERS;
+    }
+
+    public SELF reasonerPerfCounters(boolean perfCounters) {
+        this.reasonerPerfCounters = perfCounters;
+        return getThis();
+    }
+
```
krishnangovindraj, line 118: New option

reasoner/Reasoner.java
```diff
@@ -80,14 +81,16 @@ public class Reasoner {
...
+    private final ReasonerPerfCounters perfCounters;
 
     public Reasoner(ConceptManager conceptMgr, LogicManager logicMgr,
                     TraversalEngine traversalEng, Context.Transaction context) {
         this.conceptMgr = conceptMgr;
         this.traversalEng = traversalEng;
         this.logicMgr = logicMgr;
-        this.planner = ReasonerPlanner.create(traversalEng, conceptMgr, logicMgr, context.options().explain());
-        this.controllerRegistry = new ControllerRegistry(actor(), traversalEng, conceptMgr, logicMgr, planner, context);
+        this.perfCounters = new ReasonerPerfCounters(context.options().infer() && context.options().reasonerPerfCounters());
```
krishnangovindraj, line 91: Enabled if infer is on & reasonerPerfCounters is enabled, which is set through config file.

server/parameters/CoreConfig.java
```diff
@@ -330,6 +337,19 @@ public Output.Type.File output() {
...
+                private final boolean enable;
+
+                ReasonerPerfCounters(boolean enable) {
+                    this.enable = enable;
+                }
+
+                public boolean isEnabled() {
+                    return enable;
+                }
+            }
```
krishnangovindraj, line 352: data structure for new config branch

server/TypeDBServer.java
```diff
@@ -101,7 +101,8 @@ protected TypeDBServer(CoreConfig config, boolean debug, Factory factory) {
                 .dataDir(config.storage().dataDir())
                 .storageDataCacheSize(config.storage().databaseCache().dataSize())
                 .storageIndexCacheSize(config.storage().databaseCache().indexSize())
-                .reasonerDebuggerDir(config.log().debugger().reasoner().output().path());
+                .reasonerDebuggerDir(config.log().debugger().reasonerTracer().output().path())
+                .reasonerPerfCounters(config.log().debugger().reasonerPerfCounters().isEnabled());
```
krishnangovindraj, line 105: Set option based on config.

server/parameters/CoreConfigParser.java
```diff
@@ -420,6 +424,31 @@ public List<com.vaticle.typedb.core.server.parameters.util.Help> helpList(String
...
+
+                @Override
+                public CoreConfig.Log.Debugger.ReasonerPerfCounters parse(YAML yaml, String path) {
+                    if (yaml.isMap()) {
+                        validatePredefinedKeys(parsers, yaml.asMap().keys(), path);
+                        String type = typeParser.parse(yaml.asMap(), path);
+                        assert ReasonerPerfCounters.type.equals(type);
+                        return new CoreConfig.Log.Debugger.ReasonerPerfCounters(enable.parse(yaml.asMap(), path));
+                    } else throw TypeDBException.of(CONFIG_SECTION_MUST_BE_MAP, path);
+                }
```
krishnangovindraj, line 445: Parsing logic for new section in config.

server/parameters/config.yml
```diff
@@ -49,10 +49,13 @@ log:
...
   debugger:
-    reasoner:
+    reasoner-tracer:
       enable: false
-      type: reasoner
+      type: reasoner-tracer
       output: file
+    reasoner-perfcounters:
+      enable: false
+      type: reasoner-perfcounters
```
krishnangovindraj, line 58: New section in config.

server/test/BUILD
```diff
@@ -35,7 +35,7 @@ host_compatible_java_test(
         "//common:common",
         "//server:server"
     ],
-    test_class = "com.vaticle.typedb.core.server.test.ConfigurationTest",
+    test_class = "com.vaticle.typedb.core.server.parameters.CoreConfigTest",
```
krishnangovindraj, line 38: This test seems very broken.

krishnangovindraj (commented):
Self review for adding config option to enable perf counters

reasoner/controller/ControllerRegistry.java
```diff
@@ -276,6 +281,7 @@ public void setExecutorService(ActorExecutorGroup executorService) {
 
     public void close() {
         controllerContext.tracer().ifPresent(Tracer::finishTrace);
+        controllerContext.processor().perfCounters().logCounters();
```
krishnangovindraj, line 284: Print counters at the end of reasoning.

common/parameters/Options.java
```diff
@@ -41,6 +41,7 @@
     public static final boolean DEFAULT_INFER = false;
     public static final boolean DEFAULT_TRACE_INFERENCE = false;
     public static final boolean DEFAULT_EXPLAIN = false;
+    public static final boolean DEFAULT_PERFCOUNTERS = false;
```
flyingsilverfin, line 44: `DEFAULT_REASONER_PERF_COUNTERS`
you should match the naming and also the casing (`cameCase` -> `_`)

server/parameters/CoreConfigParser.java
```diff
@@ -376,26 +376,30 @@ private static class Debugger extends Compound<CoreConfig.Log.Debugger> {
...
             private static final String description = "Debuggers that may be enabled at runtime.";
 
-            private static final Predefined<CoreConfig.Log.Debugger.Reasoner> reasoner =
-                    predefined("reasoner", "Configure reasoner debugger.", new Reasoner());
-            private static final Set<Predefined<?>> parsers = set(reasoner);
+            private static final Predefined<CoreConfig.Log.Debugger.ReasonerTracer> reasonerTracer =
+                    predefined("reasoner-tracer", "Configure reasoner tracer.", new ReasonerTracer());
+
+            private static final Predefined<CoreConfig.Log.Debugger.ReasonerPerfCounters> reasonerPerfCounters =
+                    predefined("reasoner-perfcounters", "Configure reasoner perfcounters.", new ReasonerPerfCounters());
```
flyingsilverfin, line 383: Use the full description string "Configure per-transaction reasoner performance counters."

server/parameters/CoreConfigParser.java
```diff
@@ -420,6 +424,31 @@ public List<com.vaticle.typedb.core.server.parameters.util.Help> helpList(String
                     return list(typeParser.help(path), output.help(path), enable.help(path));
                 }
             }
+
+            private static class ReasonerPerfCounters extends Compound<CoreConfig.Log.Debugger.ReasonerPerfCounters> {
+
+                private static final String type = "reasoner-perfcounters";
```
flyingsilverfin, line 430: `reasoner-perf-counters`

server/parameters/CoreConfigParser.java
```diff
@@ -420,6 +424,31 @@ public List<com.vaticle.typedb.core.server.parameters.util.Help> helpList(String
...
+                private static final Predefined<String> typeParser =
+                        predefined("type", "Type of this debugger.", restricted(STRING, list(type)));
+                private static final Predefined<Boolean> enable =
+                        predefined("enable", "Enable perf-counting and logging.", BOOLEAN);
```
flyingsilverfin, line 434: add _in each transaction_

server/parameters/CoreConfigParser.java
krishnangovindraj: Doesn't that sounds like you're configuring it per transaction?

flyingsilverfin: Humm good point, maybe "Configure transaction-bound reasoner performance counters"?

flyingsilverfin (approved):

PR #6808, opened 2023-05-18
Optimise queries by introducing concept-level schema cache

flyingsilverfin:

Many data operations rely on the schema easily and performantly available. However, we only cached the schema at the lower "graph" layer - which has no knowledge of semantic constructs like ownerships. 

As of #6775, caching at the Concept layer becomes particularly useful since it introduces constructs to represent ownerships. To avoid re-computing type ownerships, we had already introduced caching in the owner Type. However, the Type itself was repeatedly constructed, negating the usage of this cache. 

This change introducing a schema cache in the Concept Manager, and redirects all constructions of Type concepts via the Concept Manager to leverage the cache. This cache is only active when the schema is read-only.

With this change, we can see heavy write performance of improve by up to 35%, in particular when writing attribute ownerships.

* Introduce a cache in the ConceptManager to avoid re-creating schema Types
* Remove all constructors of shape `Type.of(vertex)` - the Concept Manager now directly creates the appropriate type of Type concept itself.
* Redirect all construction of Types by wrapping vertices to go through the Concept Manager, which allows us to return cached Types
  * To implement this, all Concepts now carry a `ConceptManager` within them, instead of a `GraphManager`

### DIFF
--- a/concept/ConceptImpl.java
+++ b/concept/ConceptImpl.java
@@ -46,6 +46,12 @@ import static com.vaticle.typedb.core.common.exception.ErrorMessage.TypeRead.INV
 
 public abstract class ConceptImpl implements Concept {
 
+    protected final ConceptManager conceptMgr;
+
+    protected ConceptImpl(ConceptManager conceptMgr) {
+        this.conceptMgr = conceptMgr;
+    }
+
     @Override
     public boolean isType() {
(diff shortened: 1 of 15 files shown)


flyingsilverfin:
Merging with verbal approval from Haikal

PR #6810, opened 2023-05-23
Fix assertions in Data Importer

krishnangovindraj:
Remove one overly-restrictive assert and fix a bug which was triggering another assert.
This bug would not have affected correctness or completeness of the import.

* `getMappedThing` can return null in the case of nested relations, since the relation may only be processed in a later batch.
* clearing incompleteIDs on commit. After committing, the responsibility of incompleteIIDs lies with the conceptTracker

### DIFF
--- a/migrator/data/DataImporter.java
+++ b/migrator/data/DataImporter.java
@@ -248,6 +248,7 @@ public class DataImporter {
                 conceptTracker.recordMapped(originalID, pair.second());
                 if (incompleteIDs.contains(originalID)) conceptTracker.recordIncomplete(originalID);
             });
+            incompleteIDs.clear();
             bufferedToOriginalIDs.clear();
             originalToBufferedIDs.clear();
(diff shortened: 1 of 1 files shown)


migrator/data/DataImporter.java
```diff
@@ -248,6 +248,7 @@ private void commitBatch() {
                 conceptTracker.recordMapped(originalID, pair.second());
                 if (incompleteIDs.contains(originalID)) conceptTracker.recordIncomplete(originalID);
             });
+            incompleteIDs.clear();
```
krishnangovindraj, line 251: I could also add an assert to line 244, asserting the incompleteIDs are either in the originalToBufferedIDs set, or in the conceptTracker's incomplete set.

flyingsilverfin (approved):

flyingsilverfin (approved):

PR #6815, opened 2023-06-01
Relax rule conclusion validation to facilitate rules over abstract types

flyingsilverfin:

A follow on from #6801 , we finish the rule validation relaxation implemented there. We further relax the 'then' of the rule to operate over concrete  variables, as long as all the type variables that define _which_ instances to create during materialisation are concrete.

For example:
```
define
abstract-person sub entity, abstract, plays friendship:friend; #abstract
friendship sub relation, relates friend;  #non-abstract

rule concrete-relation-over-abstract-players: 
when { 
   $x isa abstract-person;
} then {
   (friend: $x) isa friendship;
};
```

Is now a legal rule to write since the 'friendship' that would be created is an instance of a concrete type.

* Relax rule 'then' validation to only validate that the type provided in 'isa' constraints may not contain abstract types

### DIFF
--- a/logic/Rule.java
+++ b/logic/Rule.java
@@ -69,8 +69,8 @@ import static com.vaticle.typedb.core.common.exception.ErrorMessage.Internal.ILL
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.Pattern.INVALID_CASTING;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.RuleWrite.RULE_CONCLUSION_ILLEGAL_INSERT;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.RuleWrite.RULE_THEN_INCOHERENT;
+import static com.vaticle.typedb.core.common.exception.ErrorMessage.RuleWrite.RULE_THEN_INSERTS_ABSTRACT_TYPES;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.RuleWrite.RULE_THEN_INVALID_VALUE_ASSIGNMENT;
-import static com.vaticle.typedb.core.common.exception.ErrorMessage.RuleWrite.RULE_THEN_UNANSWERABLE;
 import static com.vaticle.typedb.core.common.exception.ErrorMessage.RuleWrite.RULE_WHEN_INCOHERENT;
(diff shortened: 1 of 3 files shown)


logic/Rule.java
```diff
@@ -188,7 +188,6 @@ private void validateSatisfiable() {
             }
         }
         if (!then.isCoherent()) throw TypeDBException.of(RULE_THEN_INCOHERENT, structure.label(), then);
-        if (!then.isAnswerable()) throw TypeDBException.of(RULE_THEN_UNANSWERABLE, structure.label(), then);
```
flyingsilverfin, line 191: We allow un-answerable 'then' clauses now

logic/Rule.java
```diff
@@ -413,10 +412,18 @@ public Has.WithoutIsa asHasWithoutIsa() {
...
 
         public void validate(LogicManager logicMgr, ConceptManager conceptMgr) {
-            validateInsertable(logicMgr);
+            validateInsertable(logicMgr, conceptMgr);
         }
 
-        private void validateInsertable(LogicManager logicMgr) {
+        private void validateInsertable(LogicManager logicMgr, ConceptManager conceptMgr) {
+            // All types (labelled or variabilised) must represent exclusively concrete types
+            if (isIsa() && iterate(asIsa().isa().type().inferredTypes())
```
flyingsilverfin, line 420: slightly more relaxed 'then' validation instead

logic/tool/TypeInference.java
```diff
@@ -194,7 +194,9 @@ private FunctionalIterator<Map<Identifier.Variable.Name, Label>> typePermutation
             Map<Retrievable, Order> sortOrder = new HashMap<>();
             sortVars.forEach(var -> sortOrder.put(var, Order.Asc.ASC));
             traversal.modifiers().filter(inferenceFilter).sorting(Modifiers.Sorting.create(sortVars, sortOrder));
-            return traversalEng.iterator(traversal).map(vertexMap -> {
+            return traversalEng.iterator(traversal).filter(vertexMap ->
+                !iterate(vertexMap.map().entrySet()).anyMatch(e -> thingInferenceVars().contains(e.getKey()) && e.getValue().asType().isAbstract())
```
flyingsilverfin, line 198: note: we filter out permutations containing abstract 'thing' variable types in type inference permutations, which are used when validating rules' insertability: we can guarantee that thing variables will never provide instances of abstract types, so we can eliminate these permutations.

PR #6822, opened 2023-06-08
Revise how reasoner planner does answer count estimates

krishnangovindraj:
This PR introduces a 'connectivity-graph' abstraction to simplify how the reasoner planner's answer count estimator models the effect of constraints in a conjunction on each other.

The reasoner planner was originally designed to estimate the answer count of a conjunction of constraints as the greedy-cover of the estimates of the variables by the individual constraints. Later, we would scale down estimates based on the variables it shared with other constraints. e.g.: If a constraint A had an estimate of A.x for a variable `$x`, And a constraint B had an estimate of B.x, such that B.x < A.x; We would scale the model of A by a factor (B.X/A.x) - applying the selectivity of B to A.

The drawback of the greedy-cover is that the connectivity constraints were local, and would not propagate across multiple resolvables.[1]

In the new implementation, we create a "connectivity-graph" representing the expected answer graph. The graph reflects the structure of the query, with each vertex (or edge) labelled with the expected number of corresponding vertices (or edges) in the answer graph. Further, we do a minimal-spanning-tree style estimate to replace the greedy set-cover - by construction of the connectivity graph, the connectivity constraints apply across multiple resolvables.

#### Constructing the connectivity graph
The AnswerCountEstimator has a model for each constraint in the conjunction. Each of these can be mapped to a connectivity subgraph, where the model gives us the labels of the vertices and edges. e.g., The model for a `has` concludable will have an estimate for each variable and for the number of `has` edges between them. In the corresponding connectivity (sub)graphs, these will be the labels of the vertices and edges respectively.

When we combine two connectivity (sub)graphs, we "scale" down the variable labels on each so they agree on the estimates of common variables. We then propagate this by scaling down the number of edges connected to the variable. If the number of edges is less than the vertex on the other end, we scale it down to the number of edges and propagate it further.

#### Estimating answer count from the graph
The answer count of a conjunction is the set of all (combinations) of answers to the variables in the conjunction.

If $v(x)$ is the label for the vertex corresponding to `$x` in the graph, and $e(x, y)$ is the label for the (undirected) edge between `$x` and `$y`, We define the directed connectivity from `$x` to `$y` as $c(x, y) = e(x, y)/x$.

First, consider the simplified problem of finding the answers to all variables (v/s some subset) for a query with a tree structure.
We could start an any vertex `$x` and compute the product of connectivities for each edge in the tree rooted at `$x` (for the direction root -> leaf).

$ answers =  v(x) \times \prod_{y\in adj(x)} \{ c(x,y) \prod_{z \in adj(y)} \{c(y,z) * ...\} \} $

Two complications arise: 

1. When the query is not tree structured, neither is our connectivity graph, and there is not a unique tree. We ideally would do the same for the tree which yields the minimum result for the $answers$ . This corresponds to the minimum weight directed spanning tree. Since we're taking the products of connectivity, the weights of the edge have to be the log of the actual connectivities of the directed edges.
1. When we are only querying the answer count of a subset of variables, we cannot simply multiply the connectivities because the estimate for how many vertices of variable `$x` we are connected to might exceed the number of unique vertices possible for `$x` (i.e. $v(x)$) (see [3]). For this, we have to cap the product we're computing at each vertex so it does not exceed $v(x)$ % 

 Note that the minimum spanning tree for the set of all variables may not be the same for a subset of the variables. Ideally, we would find a combined solution to both the problems.

[1]  if we had 15 each of $x, $y, and $z. One $x is connected to exactly one unique $y through a constraint (x,y) and one $y is connected to one unique $z through ($y,$z). The greedy cover would pick both (x,y) and (y,z) but would have an estimate of 15 from (x,y) and 15 for (y,z), yielding an answer of 225. The correct answer would be to see that each $x is connected to one $y and one $z in turn, for a total answer count of 15.

[2]: When querying the answers to {$x,$y} for a query-graph ($x-$y-$z), We may have one $x, connected to five $y, each connected to one $z. The minimal spanning tree would have a weight of 5, whereas the number of answers cannot exceed 1 (the only pair of $x, $y which exist in the database).

### DIFF
--- a/reasoner/planner/AnswerCountEstimator.java
+++ b/reasoner/planner/AnswerCountEstimator.java
@@ -121,7 +121,7 @@ public class AnswerCountEstimator {
         private final ConjunctionModel conjunctionModel;
         private final LocalModelFactory localModelFactory;
         private final Map<Variable, Double> minVariableEstimate;
-        private final Map<Variable, Set<LocalModel.EdgeEstimate.Directed>> outgoingEdges; // TODO: Only store the min edge per (u,v) ?
+        private final Map<Variable, Set<LocalModel.EdgeEstimate.Directed>> outgoingEdges;
 
         private IncrementalEstimator(ConjunctionModel conjunctionModel, LocalModelFactory localModelFactory, Set<Variable> initialBounds) {
             this.conjunctionModel = conjunctionModel;
@@ -157,7 +157,7 @@ public class AnswerCountEstimator {
         }
 
         private void propagate(Map<Variable, Double> firstUpdatesToApply) {
-            int maxIters = Math.max(1, 2 * (firstUpdatesToApply.size() + minVariableEstimate.size())); // TODO: Ignoring small changes can force early convergence
+            int maxIters = Math.max(1, 2 * (firstUpdatesToApply.size() + minVariableEstimate.size()));
             Map<Variable, Double> updatesToApply = firstUpdatesToApply;
             while (!updatesToApply.isEmpty() && maxIters > 0) {
                 assert iterate(updatesToApply.entrySet()).allMatch(update -> update.getValue() <= minVariableEstimate.getOrDefault(update.getKey(), Double.MAX_VALUE));
@@ -177,7 +177,8 @@ public class AnswerCountEstimator {
             Double newFromEstimate = minVariableEstimate.get(from);
             outgoingEdges.get(from).forEach(edge -> {
                 edge.scaleDownFrom(newFromEstimate);
-                if (edge.toEstimate() < minVariableEstimate.get(edge.to())) {
+                double delta = (minVariableEstimate.get(edge.to()) - edge.toEstimate());
+                if ( delta > 0.1 && delta / minVariableEstimate.get(edge.to()) > 0.01) { // Ignore reductions less than 1%
                     cascadingEffects.put(edge.to(), edge.toEstimate());
                 }
             });
@@ -191,13 +192,13 @@ public class AnswerCountEstimator {
             double estimate = 1.0;
             while (!remainingVars.isEmpty()) {
                 Variable start = remainingVars.stream().findAny().get();
+                // Question: Does the tree which minimises the answer count for all variables also minimise it for the queried variables?
                 Map<Variable, Set<LocalModel.EdgeEstimate.Directed>> mdst = findTree(remainingVars, start);
                 estimate *= minVariableEstimate.get(start) * effectiveConnectivityFromMdst(estimateableVariables, mdst, start, 1.0, null).first();
             }
             return estimate;
         }
 
-        // TODO: Read up and see if I can use a standard rooted minimum directed spanning tree approach.
         private Map<Variable, Set<LocalModel.EdgeEstimate.Directed>> findTree(Set<Variable> remainingQueryVars, Variable start) {
             Map<Variable, Set<LocalModel.EdgeEstimate.Directed>> tree = new HashMap<>();
             tree.put(start, new HashSet<>());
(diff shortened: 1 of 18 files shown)


reasoner/planner/OrderingCoster.java
```diff
@@ -178,7 +178,7 @@ Map<Resolvable<?>, Double> resolvableHeuristics(ReasonerPlanner.CallMode callMod
             } else if (res.isNegated()) {
                 estimate = iterate(res.asNegated().disjunction().conjunctions()).map(conj -> {
                     Set<Variable> conjVariables = estimateableVariables(conj.pattern().variables());
-                    Set<Variable> commonVars = Collections.intersection(conjVariables, resVars);
+                    Set<Variable> commonVars = Collections.intersection(conjVariables, conjunctionNode.conjunction().pattern().variables());
```
krishnangovindraj, line 181: Bugfix: Can apply to development separately if needed.

test/benchmark/reasoner/iam/basic/BasicTest.java
```diff
@@ -73,7 +73,7 @@ public void testOwnership() {
 
         benchmark.assertAnswerCountCorrect();
         benchmark.assertRunningTime(500);
-        benchmark.assertCounters(10, NACCESS, NACCESS + 1, NACCESS + 1);
+        benchmark.assertCounters(10, NACCESS, 2, 2);
```
krishnangovindraj, line 76: Improvement, apparently.

reasoner/planner/ReasonerPlanner.java
```diff
@@ -72,22 +72,22 @@ static boolean dependenciesSatisfied(Resolvable<?> resolvable, Set<Variable> bou
...
-        perfCounters.timePlanning.add(System.nanoTime() - start);
-    }
-
     public void planAllDependencies(Concludable concludable, Set<Variable> mode) {
         triggeredCalls(concludable, estimateableVariables(mode), null)
-                .forEach(callMode -> plan(callMode));
+            .forEach(callMode -> plan(callMode));
+    }
+
+    public void planRoot(ResolvableConjunction conjunction) {
```
krishnangovindraj, line 81: New, neater interface imo

test/benchmark/reasoner/iam/complex/ConjunctionStructureTest.java
```diff
@@ -83,7 +83,8 @@ public void testMultipleStartingPoints() {
 
         benchmark.assertAnswerCountCorrect();
         benchmark.assertRunningTime(1000);
-        benchmark.assertCounters(500, 17, 50, 120);
+        benchmark.assertCountersBelow(500, 14, 100, 140, 1.0);
+        benchmark.assertCountersAbove(12, 48, 80, 1.0); // TODO: Assert lower bound for CompoundStreams
```
krishnangovindraj, line 87: Had to separate the variance was too high to do it in a single number.

The variance is caused indeed by the new scaling, but it seems to be the correct behaviour, since the plans being generated are equal cost as far as the planner can see 
(The specific case is the variables involved in the inter-changeable constraints are bound, hence have estimate 1, so even our multiplicative cost cannot differentiate the plans)

test/integration/reasoner/planner/AnswerCountEstimatorTest.java
```diff
@@ -200,7 +197,7 @@ public void test_owns_based_estimate_two_concludables() {
             ResolvableConjunction conjunction = ResolvableConjunction.of(resolvedConjunction("{ $m isa man, has name $n; $p isa! person, has $n; }", transaction.logic()));
 
             double answers = answerCountEstimator.estimateAnswers(conjunction, getVariablesByName(conjunction.pattern(), set("m", "p")));
-            assertEquals(4.0, answers);
+            assertEquals(2.0, answers);
```
krishnangovindraj, line 200: All the fixes here are just to reflect the better scaling.

krishnangovindraj (commented):
Quick round of self-review of the trivial changes. Self-review of the meat coming soon.

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -118,157 +120,162 @@ public IncrementalEstimator createIncrementalEstimator(ResolvableConjunction con
...
+                for (Set<LocalModel.EdgeEstimate.Directed> toCorrect : groupedEdges) {
+                    if (toCorrect.size() > 1) {
+                        double permutations = 1;
+                        double nPowerN = 1;
+                        for (int i = 1; i <= toCorrect.size(); i++) {
+                            permutations *= i;
+                            nPowerN *= toCorrect.size();
+                        }
+                        connectivity *= (permutations / nPowerN);
+                    }
```
krishnangovindraj, line 259: Correction for duplicated role-players

reasoner/planner/AnswerCountEstimator.java
```diff
@@ -366,9 +371,15 @@ private LocalModel buildConstraintModel(Constraint constraint, @Nullable Conclud
...
-                } else throw TypeDBException.of(UNSUPPORTED_OPERATION);
+                } else if (asThingConstraint.isIID()) {
+                    return localModelFactory.modelForIID(asThingConstraint.asIID());
+                } else if (asThingConstraint.isPredicate()) {
+                    return localModelFactory.modelForPredicate(asThingConstraint.asPredicate(), correspondingConcludable);
+                } else if (asThingConstraint.isIs()) {
+                    return localModelFactory.modelForIs(asThingConstraint.asIs(), correspondingConcludable);
+                } else throw TypeDBException.of(ILLEGAL_STATE);
+            } else if (constraint.isValue()) {
+                return new LocalModel(new HashSet<>(localModelFactory.variableModelsBasedOnType(new HashSet<>(constraint.asValue().variables()), null)), set());
```
krishnangovindraj, line 382: Add the remaining predicates to model

krishnangovindraj (commented):
Self review AnswerCountEstimator.

flyingsilverfin:
@krishnangovindraj  im going to close this?

PR #6826, opened 2023-06-19
Conjunction stream graph optimisation

krishnangovindraj:

Enable the reasoner to efficiently cache and reuse results to sub-conjunctions where possible. This can dramatically increasing the performance of reasoning in some large conjunctions configurations.

* Reuse results to sub conjunction wherever possible
  - The current architecture requires us to buffer the results if we have multiple readers to the stream.
* Aggressively prune out variables which are not relevant to the remaining resolvables in the plan
  - If the variable is an output of the conjunction, we extend the result of the sub conjunction with it.
  - else we just project it out.
* Introduce ConjunctionStreamPlan, which divides the conjunction in a way that maximises reusability.

### Motivation
The stream plan forms a tree, with ResolvablePlans  as the leaves, and CompoundStreamPlan as internal nodes. Each CompoundStreamPlan has a left and right child. Each node in the tree represents some sub-conjunction. The order of the leaves from left-to-right matches the ordering in the plan chosen by ReasonerPlanner. 

The actual stream graph forms a tree, with the individual Resolvables as the leaves and compound streams as internal nodes. Each node is identified by the bounds to a set of variables - henceforth '**identifiers**'. A compound stream has a unique left child, and a family of right-children. Each right-child corresponds to the same sub conjunction but differs on the identifiers. The changes correspond to normalising intermediate tables if we saw everything as tables.
  
**Current system**
In the current system, Every conjunction is divided into its first resolvable on the left, and the remaining resolvable on the right. We create a new compound stream for each branch in the conjunction, whose identifiers are the current bounds for all variables in the resolvables processed so far.

For the goal :  `z(A,D) :- p(A, B), q(B, C), r(C, D).` (with no bounds to Z) this would correspond to the plan: 

```
⨂: pqr(A,B,C,D)
+---?: p(A,B)
^---⨂: qr(a, b,C,D)
    +---?: q(b, C)
    ^---?: r(c,D)
```
Where ,  
* ⨂ represents the a compound stream node
* ? represents a resolvable node
* `q(b,C)` indicates it only contains the rows in `q` which are queried by backward-chaining (with values for `b` resulting from the resolvables before it, here {`p`}), 
* `qr(a,b,C,D)` indicates the intermediate result of joining `q` with `r`, again with only results `a,b` from p.

The plan produces the stream graph:
**The amount of work done is approximately number of answers propagated up each edge**
image

**Reusing results**
Considering that we only require {A,D} as output, we transform the query after each join - projecting away the variables not used in the suffix, and joining any outputs on the way back up. Further, it's clear that the value of `qr`does not actually depend on the value of `a`, so could be reused by different values of a which share the same value for `b`.

This results in the plan:
```
⨂: pqr(A,D)
+---?: p'(A,B)
^---⨁: (a,)
    +---⨂: qr(b, D)          (reusable)
        +---?: q'(b, C)
        ^---?: r'(c,D)
```

Where, 
* ⨁" (a,) is the newly introduced map stream which appends the **extension** bounds (the values of `a`) to the results from the subtree.

This would be the resulting stream graph. Notice we also de-duplicate answers this way.
image

** Optimising reusability **
Suppose our goal was instead,  `z(A,D) :- p(A, B), q(B, C), r(C, D), s(A,D).` [^1]

We would not be able to reuse results for the suffix `qrs` because of the dependency on `a`. i.e. `qrs(a,b,D)` and `rs(a,c,D)` would be the sub conjunctions.
```
⨂: pqrs(A,D)
+---?: p'(A,B)
+---⨂: qrs(a, b, D)
       +---?: q'(b, C)
       ^---?: rs'(a,c,D)             (reusable)
               +---?: r(c, D)
               ^---?: s(a,d)
```
with the stream graph:
image

There may be multiple ways to restructure the stream, but we aim to pick a structure such that the bounds of every right-child is bound with exactly the set of variables expected by its left-child - The only violation allowed is for a sub conjunction of size 2, where it's unavoidable.

```
⨂: pqrs(A,D)
+---⨂: pqr'(A,B)             
|   +---?: p'(A,B)
|   ^---⨁: (a,)
|       ^---⨂: qr(b, D)             (reusable)
|           +---?: q'(b, C)
|           ^---?: r'(c,D)
^---?: s(a,d)
```
with the stream graph:
image

##### Flattening  
In cases where a compound stream node is redundant, we directly add it's children to it parent, reducing one hop.
This is the case when
a) The parent is the exclusive reader AND
b) No bounds for the compound stream are bound by the parent's other (more-left) children.

##### reusing answers
When we create a compound stream for a conjunction
1. Create a for the left child with the relevant subset of bounds as identifier
  * The left child will 'stream' results in as and when it computes one.

For each answer received back from its left child, We create (or reuse) a right-child by:
1. Create (or get an existing) compound stream with the relevant identifier bounds.
2. Compute a set of extension bounds - consisting of the variables bound by the left-child, not used by the right BUT is either used by a later resolvable or an output of the conjunction.
  -  If this set is non-empty, we attach the compound stream through a MapStream which extends any answers coming from it with the extension bounds.
  - else, we attache the compound stream from one directly.
* The right child will stream results in as and when it computes them (through the map stream)

We propagate any answer coming from our right children up towards the parent compound-stream or processor.


reasoner/controller/ConditionController.java
```diff
@@ -40,7 +42,7 @@ public class ConditionController extends DisjunctionController<
     private final Rule.Condition condition;
 
     ConditionController(Driver<ConditionController> driver, Rule.Condition condition, Context context) {
-        super(driver, condition.disjunction(), context);
+        super(driver, condition.disjunction(), iterate(condition.rule().conclusion().retrievableIds()).filter(v -> !v.isAnonymous()).toSet(), context);
```
krishnangovindraj, line 45: outputVariables are the named variables in the conclusion

reasoner/controller/ConjunctionController.java
```diff
@@ -184,80 +194,115 @@ NegatedRequest asNegated() {
...
                   Driver<? extends ConjunctionController<?, PROCESSOR>> controller,
-                  Context context, ConceptMap bounds, List<Resolvable<?>> plan,
+                  Context context, ConceptMap bounds, ConjunctionStreamPlan conjunctionStreamPlan,
                   Supplier<String> debugName) {
             super(driver, controller, context, debugName);
             this.bounds = bounds;
-            this.plan = plan;
+            this.outputVariables = controller.actor().outputVariables;
+            this.plan = conjunctionStreamPlan;
+            this.compoundStreamRegistry = new HashMap<>();
```
krishnangovindraj, line 209: Map bounds -> compound stream to reuse sub conjunction results.
CompoundStreams are internal to the processor (and cannot be shared across processors without violating actor isolation).

reasoner/controller/ConjunctionController.java
```diff
@@ -184,80 +194,115 @@ NegatedRequest asNegated() {
...
+            private final ConceptMap identifierBounds;
             private final AbstractProcessor<?, ?, ?, ?> processor;
+            private final Map<Publisher<ConceptMap>, Integer> whichChild;
 
-            CompoundStream(AbstractProcessor<?, ?, ?, ?> processor, List<Resolvable<?>> plan,
-                           ConceptMap initialPacket) {
-                super(processor, new SubscriberRegistry.Single<>(), new PublisherRegistry.Multi<>());
+
+            CompoundStream(AbstractProcessor<?, ?, ?, ?> processor, ConjunctionStreamPlan plan, ConceptMap identifierBounds) {
+                super(processor, new SubscriberRegistry.Multi<>(), new PublisherRegistry.Multi<>());
```
krishnangovindraj, line 221: Since compound stream results are reused, they can have multiple readers

reasoner/processor/AbstractProcessor.java
```diff
@@ -188,6 +190,10 @@ public Driver<Monitor> monitor() {
         public ReasonerPerfCounters perfCounters() {
             return perfCounters;
         }
+
+        public boolean explainEnabled() {
+            return explainEnabled;
+        }
```
krishnangovindraj, line 196: Since we're factoring the conjunction into small bits for re-usable answers, we break the present approach to explanations. We now use this as a switch to turn off reasoner optimisations which break explanations.

reasoner/processor/reactive/AbstractStream.java
```diff
@@ -63,8 +63,10 @@ void propagatePull(Publisher<INPUT> publisher) {
 
     @Override
     public void registerPublisher(Publisher<INPUT> publisher) {
-        publisherRegistry().add(publisher);
-        subscriberDelegate().registerPath(publisher);
+        if (!publisherRegistry.contains(publisher)) {
+            publisherRegistry().add(publisher);
+            subscriberDelegate().registerPath(publisher);
+        }
```
krishnangovindraj, line 69: Prevents the duplicate edges between the same two streams. 

This shouldn't happen with the optimised conjunction stream structure (but might with explanations disabling the optimisation) and I should look into it and reply here on why/when it happens or add/replace it with an assert.

test/benchmark/reasoner/iam/basic/BasicTest.java
```diff
@@ -160,6 +160,24 @@ public void testDoubleJoin() {
...
+                        "(start: $c, end: $d) isa object-pair;\n" +
+                        "get $a, $d;\n",
+                queryParams.basicTestObject);
+        Benchmark benchmark = new Benchmark("double-join", query, NOBJECTS);
+        benchmarker.runBenchmark(benchmark);
+
+        benchmark.assertAnswerCountCorrect();
+        benchmark.assertRunningTime(1500);
+        benchmark.assertCounters(25, NOBJECTS * NOBJECTS, NOBJECTS, NOBJECTS + NOBJECTS + 3);
+    }
```
krishnangovindraj, line 179: Minimal example to show why the re-usability is useful.

TODO: See if I can add one to create a ring so the stream-graph restructuring is tested.

reasoner/controller/ConjunctionController.java
```diff
@@ -184,80 +194,115 @@ NegatedRequest asNegated() {
...
-                this.leadingPublisher.registerSubscriber(this);
+                this.identifierBounds = identifierBounds;
+                this.plan = plan;
+                Publisher<ConceptMap> leftChild;
+                if (plan.isCompoundStream()) {
+                    leftChild = spawnPlanElement(plan.asCompoundStreamPlan().ithChild(0), identifierBounds);
+                } else {
+                    leftChild = spawnPlanElement(plan.asResolvablePlan(), identifierBounds);
+                }
+                leftChild.registerSubscriber(this);
```
krishnangovindraj, line 231: Creates the first child based on input bounds.
Remaining children are created when children on their left (before them in the plan) return an answer.

reasoner/controller/ConjunctionController.java
```diff
@@ -184,80 +194,115 @@ NegatedRequest asNegated() {
...
+                ConceptMap identifiers = bounds.filter(planElement.identifierVariables());
+                assert planElement.identifierVariables().size() == identifiers.concepts().size() &&  identifiers.concepts().keySet().containsAll(planElement.identifierVariables());
+                Publisher<ConceptMap> publisher;
+                if (planElement.isResolvable()) {
+                    publisher = spawnResolvableElement(planElement.asResolvablePlan(), identifiers);
+                } else if (planElement.isCompoundStream()) {
+                    publisher = spawnCompoundStream(planElement.asCompoundStreamPlan(), identifiers);
+                } else throw TypeDBException.of(ILLEGAL_STATE);
+
+                return extendWithBounds(publisher, extension);
```
krishnangovindraj, line 248: If a variable that's been found by a previous resolvable/subtree was not relevant to this subtree, we don't pass it down to our subtree but keep it in the compound stream and extend any answer from our subtree with it.

(These are the diamond join nodes in the figures we drew)

reasoner/controller/ConjunctionController.java
```diff
@@ -184,80 +194,115 @@ NegatedRequest asNegated() {
...
-                        if (remainingPlan.size() == 1) {
-                            follower = nextCompoundLeader(remainingPlan.get(0), mergedPacket);
-                        } else {
-                            follower = new CompoundStream(processor, remainingPlan, mergedPacket).buffer();
-                        }
-                        publisherPackets.put(follower, mergedPacket);
-                        return Either.first(follower);
-                    }
+                context().perfCounters().compoundStreamAccepts.add(1);
+                ConceptMap mergedPacket = merge(identifierBounds, packet);
```
krishnangovindraj, line 255: This line is a bit sus. I would like to see if we can just keep these identifier bounds in the extension (join) variables to handle this.

reasoner/controller/ConjunctionController.java
```diff
@@ -184,80 +194,115 @@ NegatedRequest asNegated() {
...
+                int nextChild = whichChild.get(publisher) + 1;
+
+                // assert on packet. NOT mergedPacket.  I added a join node for identifier bounds to the resolvable when spawning, breaking the assert.
+                assert plan.isResolvable() || plan.asCompoundStreamPlan().ithChild(nextChild-1).isResolvable() || (
+                        ConjunctionStreamPlan.union(plan.asCompoundStreamPlan().ithChild(nextChild-1).outputVariables(), plan.asCompoundStreamPlan().ithChild(nextChild-1).extendOutputWithVariables()).size() == packet.concepts().size() &&
+                                packet.concepts().keySet().containsAll(ConjunctionStreamPlan.union(plan.asCompoundStreamPlan().ithChild(nextChild-1).outputVariables(),
+                                plan.asCompoundStreamPlan().ithChild(nextChild-1).extendOutputWithVariables())));
```
krishnangovindraj, line 262: assert to remove when I have flattening working

reasoner/controller/ConjunctionController.java
```diff
@@ -184,80 +194,115 @@ NegatedRequest asNegated() {
...
-            } else {
-                throw TypeDBException.of(ILLEGAL_STATE);
+                return input.map(conceptMap -> merge(filterOutputsWithExplainables(conceptMap, resolvablePlan.outputVariables()), identifiers));
+            }
+            private Publisher<ConceptMap> extendWithBounds(Publisher<ConceptMap> s, ConceptMap extension) {
+                return extension.concepts().isEmpty() ? s : s.map(conceptMap -> merge(conceptMap, extension));
+            }
+
+            private ConceptMap filterOutputsWithExplainables(ConceptMap packet, Set<Variable.Retrievable> toVariables) {
+                return context().explainEnabled() ? packet : packet.filter(toVariables);
```
krishnangovindraj, line 304: If we're explaining, we can't filter out any variables.

krishnangovindraj (commented):
Comment on ConjunctionController. Will do ConjunctionStreamPlan after implementing flattening.

reasoner/planner/ConjunctionStreamPlan.java
```diff
@@ -0,0 +1,337 @@
...
+                nextSuffixVariables.addAll(subConjunction.get(r).retrieves());
+                Set<Retrievable> subtreeBounds = VariableSets.intersection(boundsAt.get(r), nextSuffixVariables);
+                Set<Retrievable> candidateFirstChildBounds = VariableSets.intersection(subConjunction.get(r).retrieves(), boundsAt.get(r));
+
+                if (!VariableSets.difference(subtreeBounds, candidateFirstChildBounds).isEmpty()) {
+                    break;
+                }
+            }
+            return r;
+        }
```
krishnangovindraj, line 245: (Is supposed to) split the (sub)conjunction into two so that the right-tree satisfies the properties described in the PR.

reasoner/planner/ConjunctionStreamPlan.java
```diff
@@ -0,0 +1,337 @@
...
+                return new ResolvablePlan(subConjunction.get(0), variableSets.rightInputs, variableSets.extendOutputWith, resolvableOutputs);
+            } else {
+                List<Resolvable<?>> suffix = subConjunction.subList(1, subConjunction.size());
+                VariableSets variableSets = VariableSets.determineVariableSets(subConjunction.subList(0, 1), subConjunction.subList(1, subConjunction.size()), availableInputs, requiredOutputs);
+                ConjunctionStreamPlan leftPlan = new ResolvablePlan(subConjunction.get(0), variableSets.leftIdentifiers, set(), variableSets.leftOutputs);
+                ConjunctionStreamPlan rightPlan = buildRecursiveSuffix(suffix, variableSets.rightInputs, variableSets.rightOutputs);
+                return new CompoundStreamPlan(list(leftPlan, rightPlan),
+                        variableSets.identifiers, variableSets.extendOutputWith, requiredOutputs);
+            }
+        }
```
krishnangovindraj, line 218: Building the suffix should be easy because it satisfies the structure described in the PR description.

reasoner/planner/ConjunctionStreamPlan.java
```diff
@@ -0,0 +1,337 @@
...
+                return new ResolvablePlan(subConjunction.get(0), variableSets.rightInputs, variableSets.extendOutputWith, resolvableOutputs);
+            } else {
+                Pair<List<Resolvable<?>>, List<Resolvable<?>>> divided = divide(subConjunction);
+                VariableSets variableSets = VariableSets.determineVariableSets(divided.first(), divided.second(), availableInputs, requiredOutputs);
+                ConjunctionStreamPlan leftPlan = buildRecursivePrefix(divided.first(), variableSets.leftIdentifiers, variableSets.leftOutputs);
+                ConjunctionStreamPlan rightPlan = buildRecursiveSuffix(divided.second(), variableSets.rightInputs, variableSets.rightOutputs);
+
+                return new CompoundStreamPlan(list(leftPlan, rightPlan), variableSets.identifiers, variableSets.extendOutputWith, requiredOutputs);
+            }
+        }
```
krishnangovindraj, line 200: Computes the split into the left and right subtrees, and recurses on the left.
The right has a predictable structure so there's nothing to compute.

reasoner/planner/ConjunctionStreamPlan.java
```diff
@@ -0,0 +1,337 @@
...
+
+        private static boolean rightChildBoundsSatisfied(ConjunctionStreamPlan.CompoundStreamPlan conjunctionStreamPlan, ConjunctionStreamPlan.CompoundStreamPlan unflattenedNextPlan) {
+            return VariableSets.difference(
+                            unflattenedNextPlan.ithChild(1).identifierVariables,
+                            VariableSets.union(conjunctionStreamPlan.identifierVariables, unflattenedNextPlan.ithChild(0).outputVariables))
+                    .isEmpty();
+        }
+
+        private static class VariableSets {
```
krishnangovindraj, line 280: Holds the logic for determining the sets of input, join and output variables for the left and right subtrees.

krishnangovindraj (commented):
More self-review

flyingsilverfin (approved):

PR #6836, opened 2023-07-04
Add RPC logging for session and transaction creation

flyingsilverfin:

To monitor production instances of TypeDB and TypeDB enterprise more effectively, when using debug logging the server logs open client connections (each sessions and number of transactions each) at a rate of once per minute. When there are no connections, the server does not log to avoid noise in the idle state.

To enable these logs, set the logger for 'com.vaticle.typedb.core.server` to DEBUG in the TypeDB configuration file.

* Add periodic logging of the global connection state when the server is in DEBUG mode

### Design

We considered alternatives:
1. Not logging the global state each time -- this will not be useful for debugging and monitoring
2. Logging global connection state on client connection -- this makes sense, but probably isn't often enough to be useful
3. Logging at each connection/session/transaction  -- this makes sense, and is consistent, but can get expensive. It also requires transaction open/close to query the _other_ sessions to log their information, which is an awkward control flow

In the end, the best tradeoff of usefulness, performance, usability, and 'correct' control flow by domain was decided to lie in a polling loop that lives at the top-level service.

Note that we ultimately want a client RPC method to query the current server state, which can retrieve network state (sessions/transactions) as well as storage statistics, etc.

### DIFF
--- a/server/SessionService.java
+++ b/server/SessionService.java
@@ -62,10 +62,12 @@ public class SessionService implements AutoCloseable {
             if (isOpen.get()) {
                 transactionServices.add(transactionSvc);
                 cancelIdleTimeout();
-                LOG.debug(
-                        "Session '{}' for database '{}' opened transaction. Transactions open in this session: {}.",
-                        UUID(), session.database().name(), transactionServices.size()
-                );
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug(
+                            "Session '{}' for database '{}' opened transaction. Transactions open in this session: {}.",
+                            UUID(), session.database().name(), transactionServices.size()
+                    );
+                }
             } else throw TypeDBException.of(SESSION_CLOSED);
         } finally {
             accessLock.readLock().unlock();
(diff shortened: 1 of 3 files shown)


server/SessionService.java
```diff
@@ -62,6 +62,12 @@ void register(TransactionService transactionSvc) {
             if (isOpen.get()) {
                 transactionServices.add(transactionSvc);
                 cancelIdleTimeout();
+                if (LOG.isDebugEnabled()) {
```
flyingsilverfin, line 65: We protect the logging in an `if` when the logging can be expensive: computing the size of concurrent hash maps/sets could induce expensive locking and block other operations - the contract does not guarantee this is a fast operation.

server/TypeDBService.java
```diff
@@ -71,6 +77,39 @@ public class TypeDBService extends TypeDBGrpc.TypeDBImplBase {
...
+                        return sessionInfos;
+                    })
+            );
+            StringBuilder connectionStates = new StringBuilder("Server connections: ").append("\n");
+            databaseConnections.entrySet().stream().sorted(Map.Entry.comparingByKey()).forEach(entry -> {
+                connectionStates.append("Database '").append(entry.getKey()).append("' has ").append(entry.getValue().size()).append(" sessions:\n");
+                entry.getValue().forEach(state -> connectionStates.append("\t").append(state).append("\n"));
+            });
+            LOG.debug(connectionStates.toString());
+        } catch (Exception e) {
```
flyingsilverfin, line 110: this catch will prevent an exception from killing the thread

PR #6842, opened 2023-07-07
Fix server hang-ups during shutdown

flyingsilverfin:

Under exceptional circumstances, such as when the server runs out of memory, the server could fail to shut down. We modify the server shutdown process and unexpected exception handler to shutdown in several stages, in the most severe case halting the JVM runtime immediately.

Implement graceful and forceful shutdown from within the process:

1. If an unexpected exception occurs, we try to perform the server `close()`. Only 1 thread is allowed to do this one time, and if it fails to do so it throws a checked exception. The caller of `close()` is expected to handle this case and force a halt of the JVM
2. If the user hits ctrl+c, we add a shutdown hook to gracefully close the server, but do nothing if it fails to since the JVM is already shutting down

During the server `close()`, we also now force the grpc server layer to shut down in 10 seconds.

### DIFF
--- a/common/exception/ErrorMessage.java
+++ b/common/exception/ErrorMessage.java
@@ -41,8 +41,8 @@ public abstract class ErrorMessage extends com.vaticle.typedb.common.exception.E
                 new Server(2, "The expected data directory '%s' is not writable.");
         public static final Server EXITED_WITH_ERROR =
                 new Server(3, "Exited with error.");
-        public static final Server UNCAUGHT_EXCEPTION =
-                new Server(4, "Uncaught exception thrown at thread '%s':\n'%s'");
+        public static final Server UNCAUGHT_ERROR =
+                new Server(4, "Uncaught error thrown at thread '%s':\n'%s'");
         public static final Server CONFIG_FILE_NOT_FOUND =
                 new Server(5, "Could not find/read the configuration file '%s'.");
         public static final Server UNRECOGNISED_CLI_COMMAND =
--- a/common/exception/TypeDBCheckedException.java
+++ b/common/exception/TypeDBCheckedException.java
@@ -37,12 +37,12 @@ public class TypeDBCheckedException extends Exception {
         this.errorMessage = error;
     }
 
-    private TypeDBCheckedException(Exception e) {
+    private TypeDBCheckedException(Throwable e) {
         super(e);
         errorMessage = null;
     }
 
-    public static TypeDBCheckedException of(Exception e) {
+    public static TypeDBCheckedException of(Throwable e) {
         return new TypeDBCheckedException(e);
     }
 
--- a/reasoner/controller/ConjunctionController.java
+++ b/reasoner/controller/ConjunctionController.java
@@ -97,19 +97,15 @@ public abstract class ConjunctionController<
         assert resolvables.isEmpty();
         resolvables.addAll(registry().logicManager().compile(conjunction));
 
-        iterate(resolvables).filter(Resolvable::isConcludable).map(Resolvable::asConcludable).forEachRemaining(c -> {
-            concludableControllers.put(c, registry().getOrCreateConcludable(c));
-        });
-        iterate(resolvables).filter(Resolvable::isRetrievable).map(Resolvable::asRetrievable).forEachRemaining(r -> {
-            retrievableControllers.put(r, registry().createRetrievable(r));
-        });
-        iterate(resolvables).filter(Resolvable::isNegated).map(Resolvable::asNegated).forEachRemaining(negated -> {
-            try {
-                negationControllers.put(negated, registry().createNegation(negated, conjunction));
-            } catch (TypeDBException e) {
-                terminate(e);
-            }
-        });
+        iterate(resolvables).filter(Resolvable::isConcludable).map(Resolvable::asConcludable).forEachRemaining(c ->
+                concludableControllers.put(c, registry().getOrCreateConcludable(c))
+        );
+        iterate(resolvables).filter(Resolvable::isRetrievable).map(Resolvable::asRetrievable).forEachRemaining(r ->
+                retrievableControllers.put(r, registry().createRetrievable(r))
+        );
+        iterate(resolvables).filter(Resolvable::isNegated).map(Resolvable::asNegated).forEachRemaining(negated ->
+                negationControllers.put(negated, registry().createNegation(negated, conjunction))
+        );
     }
 
     ConjunctionStreamPlan getPlan(Set<Variable.Retrievable> bounds) {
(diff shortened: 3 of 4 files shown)


reasoner/controller/ConjunctionController.java
```diff
@@ -97,19 +97,15 @@ protected void setUpUpstreamControllers() {
...
         resolvables.addAll(registry().logicManager().compile(conjunction));
 
-        iterate(resolvables).filter(Resolvable::isConcludable).map(Resolvable::asConcludable).forEachRemaining(c -> {
-            concludableControllers.put(c, registry().getOrCreateConcludable(c));
-        });
-        iterate(resolvables).filter(Resolvable::isRetrievable).map(Resolvable::asRetrievable).forEachRemaining(r -> {
-            retrievableControllers.put(r, registry().createRetrievable(r));
-        });
-        iterate(resolvables).filter(Resolvable::isNegated).map(Resolvable::asNegated).forEachRemaining(negated -> {
-            try {
```
flyingsilverfin, line 107: unnecessary try-catch-terminate

common/exception/ErrorMessage.java
```diff
@@ -41,8 +41,8 @@ public static class Server extends ErrorMessage {
                 new Server(2, "The expected data directory '%s' is not writable.");
         public static final Server EXITED_WITH_ERROR =
                 new Server(3, "Exited with error.");
-        public static final Server UNCAUGHT_EXCEPTION =
-                new Server(4, "Uncaught exception thrown at thread '%s':\n'%s'");
+        public static final Server UNCAUGHT_ERROR =
```
flyingsilverfin, line 44: we print this for both errors and exceptions, which are both errors from the user's point of view

common/exception/TypeDBCheckedException.java
```diff
@@ -37,12 +37,12 @@ private TypeDBCheckedException(ErrorMessage error, Object... parameters) {
         this.errorMessage = error;
     }
 
-    private TypeDBCheckedException(Exception e) {
+    private TypeDBCheckedException(Throwable e) {
```
flyingsilverfin, line 40: allow wrapping a throwable in one of our Checked exception

server/TypeDBServer.java
```diff
@@ -109,14 +111,28 @@ protected TypeDBServer(CoreConfig config, boolean debug, Factory factory) {
...
-                    System.exit(1);
+                    try {
+                        logger().error(UNCAUGHT_ERROR.message(t.getName(), e), e);
+                        close();
+                        System.exit(1);
+                    } catch (TypeDBCheckedException ex) {
+                        logger().error("Failed to shut down cleanly, performing hard stop.");
+                        Runtime.getRuntime().halt(1);
+                    } catch (Throwable error) {
+                        // another thread will do the close
```
flyingsilverfin, line 122: this clause prevents an OOM error inside a thread that isn't doing the shutdown from propagating to the top level

server/TypeDBServer.java
```diff
@@ -109,14 +111,28 @@ protected TypeDBServer(CoreConfig config, boolean debug, Factory factory) {
...
-                    logger().error(UNCAUGHT_EXCEPTION.message(t.getName(), e), e);
-                    close();
-                    System.exit(1);
+                    try {
+                        logger().error(UNCAUGHT_ERROR.message(t.getName(), e), e);
+                        close();
+                        System.exit(1);
+                    } catch (TypeDBCheckedException ex) {
+                        logger().error("Failed to shut down cleanly, performing hard stop.");
+                        Runtime.getRuntime().halt(1);
```
flyingsilverfin, line 120: if this block triggers, the thread that is shutting down failed to shut down cleanly (because of OOM or other reasons) so we force a halt

server/TypeDBServer.java
```diff
@@ -109,14 +111,28 @@ protected TypeDBServer(CoreConfig config, boolean debug, Factory factory) {
...
                 }
         );
         Runtime.getRuntime().addShutdownHook(
-                NamedThreadFactory.create(TypeDBServer.class, "shutdown").newThread(this::close)
+                NamedThreadFactory.create(TypeDBServer.class, "shutdown").newThread(() -> {
+                    try {
+                        close();
+                    } catch (Throwable error) {
+                        logger().error("Error during shutdown: ", error);
+                    }
```
flyingsilverfin, line 132: If the user hits ctrl+c, we want to initiate an orderly shutdown. However, if this fails for any reason we log the error before returning control to the JVM shutdown sequence

server/TypeDBServer.java
```diff
@@ -109,14 +111,28 @@ protected TypeDBServer(CoreConfig config, boolean debug, Factory factory) {
...
+                })
         );
+        isOpen = new AtomicBoolean(true);
```
flyingsilverfin, line 135: we make the `close()` idempotent

server/TypeDBServer.java
```diff
@@ -210,66 +226,67 @@ protected void serve() {
             server.awaitTermination();
         } catch (InterruptedException e) {
             // server is terminated
-            close();
-            Thread.currentThread().interrupt();
+            try {
+                close();
+                System.exit(0);
```
flyingsilverfin, line 231: we expect every user of `close()` to have their own error handling and exit code use.

server/TypeDBServer.java
```diff
@@ -210,66 +226,67 @@ protected void serve() {
...
-            System.runFinalization();
-            logger().info("{} has been shutdown", name());
-        } catch (InterruptedException e) {
-            logger().error(FAILED_AT_STOPPING.message(), e);
-            Thread.currentThread().interrupt();
+    public synchronized void close() throws TypeDBCheckedException {
+        if (isOpen.compareAndSet(true, false)) {
+            try {
+                logger().info("");
+                logger().info("Closing {} instance...", name());
```
flyingsilverfin, line 244: put these into the `try` because they can trigger OOM when right the limit of memory

server/TypeDBServer.java
```diff
@@ -210,66 +226,67 @@ protected void serve() {
...
+                assert typeDBService != null;
+                typeDBService.close();
+                logger().info("Stopping network layer...");
+                server.shutdown();
+                if (!server.awaitTermination(10, TimeUnit.SECONDS)) {
```
flyingsilverfin, line 249: we give the grpc server 10 seconds to shutdown cleanly, before shutting it down forcefully

server/TypeDBServer.java
```diff
@@ -210,66 +226,67 @@ protected void serve() {
...
+                    server.shutdownNow();
+                }
+                logger().info("Stopping storage layer...");
+                databaseMgr.close();
+                System.runFinalization();
+                logger().info("{} instance has been closed.", name());
+            } catch (Throwable e) {
+                logger().error(FAILED_AT_STOPPING.message(), e);
+                throw TypeDBCheckedException.of(e);
```
flyingsilverfin, line 258: any failure to close is wrapped in a checked exception, which the caller must handle

server/TypeDBServer.java
```diff
@@ -210,66 +226,67 @@ protected void serve() {
...
-                } else if (subcmd.get().isExport()) {
-                    runExport(subcmd.get().asExport());
-                } else throw TypeDBException.of(ILLEGAL_STATE);
-            }
-        } catch (Exception e) {
-            if (e instanceof TypeDBException) {
-                LOG.error(e.getMessage());
-            } else {
-                LOG.error(e.getMessage(), e);
-                LOG.error(EXITED_WITH_ERROR.message());
```
flyingsilverfin, line 267: we now expect every branch of this `if` block to handle their own exits and exit codes and never fail

benjaminasmall (approved):

PR #6848, opened 2023-07-13
Fix NullPointerExceptions during reasoner stream graph construction

krishnangovindraj:
Fix a concurrency bug leading to rare NullPointerExceptions during the construction of the reasoner graph.

The NPE occurs when one actor executes `getConcludable` when the another actor is between 

* `concludableControllers.put(concludable, controllerView.controller());` and 
* `controllerConcludables.put(controllerView.controller(), concludables);`
* Swapping the two lines to avoids the `controllerConcludables.get` prevents `getConcludable` from finding a controller before the key in `controllerConcludables.put`.

### DIFF
--- a/reasoner/controller/ControllerRegistry.java
+++ b/reasoner/controller/ControllerRegistry.java
@@ -201,10 +201,10 @@ public class ControllerRegistry {
                     driver -> new ConcludableController.Match(driver, concludable, controllerContext);
             LOG.debug("Create ConcludableController: '{}'", concludable.pattern());
             controllerView = ControllerView.concludable(createController(actorFn), identity(concludable));
-            concludableControllers.put(concludable, controllerView.controller());
             Set<Concludable> concludables = new HashSet<>();
             concludables.add(concludable);
             controllerConcludables.put(controllerView.controller(), concludables);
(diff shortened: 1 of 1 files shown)


benjaminasmall (approved):

reasoner/controller/ControllerRegistry.java
```diff
@@ -215,6 +216,9 @@ private static Map<Variable.Retrievable, Variable.Retrievable> identity(Resolvab
     }
 
     private Optional<ControllerView.MappedConcludable> getConcludable(Concludable concludable) {
+        // TODO: Due to a lack of synchronisation,
+        //      this does not guarantee that all alpha-equivalent concludables will re-use the same controller.
+        //      We should revisit what we gain out of re-using concludables instead of conclusions.
```
flyingsilverfin, line 221: Just trying to recall this right now, some historical possibilities that pop into mind:
- we didn't want to compute unification many times (not very expensive anyway)
- caching was done per-concludable and not per conclusion so reusing them was critical


Let's revisit this in 3.0

flyingsilverfin (approved):

PR #6863, opened 2023-08-10
Respect archives size cap configuration value

jamesreprise:

We've added a hardcoded `maxHistory` to the rolling policy for our logging appender. Consequently, our policy now respects the supplied `totalCapSize` which we expose as `log.output.file.archives-size-cap` in our server configuration file.

Added a constant `TYPEDB_LOG_MAX_HISTORY` and it as our maximum history in our logging policy.

## Additional info

We chose 10,000 rather than a larger number (or `Integer.MAX_VALUE`) as we found during that our logging framework behaves strangely when `maxHistory` is set to numbers much larger than 10,000 (e.g. 25,000), deleting archives rather than keeping them.

Fixes https://github.com/vaticle/typedb/issues/6854


flyingsilverfin:
Closing in favor of #6864 - the change was much bigger than expected @jamesreprise ! Always the risk of rushing things ;)

PR #6865, opened 2023-08-14
Disable stamping to fix windows builds

flyingsilverfin:

As of #6853 we include an updated version of `rules_jvm_external`, which is used to transform maven dependencies into valid Bazel targets. However, the upgrade to version 5 also included a new default feature called 'stamping' (https://bazel.build/docs/user-manual#workspace-status), which works in general but breaks builds on Windows.

This PR disables stamping by default (which supposedly also improves cache hits on remote caches), as well as shortens the Windows build directory to avoid potential long path issues in the future.

- Disable stamping in `rules_jvm_external` (if it were on, this would manifest as jars prefixed with `processed_*`
- Shorten paths lengths on Windows by placing the sandbox close to the Windows root
- Remove unexpected deletion of `.bazelrc` from windows execution setup


.circleci/windows/prepare.bat
```diff
@@ -32,7 +32,3 @@ C:\Python37\python.exe -m pip install wheel
 REM permanently set variables for Bazel build
 SETX BAZEL_SH "C:\Program Files\Git\usr\bin\bash.exe"
 SETX BAZEL_PYTHON C:\Python37\python.exe
-
-REM TODO: https://github.com/vaticle/typedb/issues/6200
-REM temporary workaround around "AsAbsoluteWindowsPath failed: Unix-style paths are unsupported" error
-DEL .bazelrc
```
flyingsilverfin, line 38: This is no longer relevant, and unintuitive since we expect windows to read the `.bazelrc` too.

.circleci/windows/test-assembly.bat
```diff
@@ -21,7 +21,7 @@ REM by Chocolatey in prepare.bat is accessible
 CALL refreshenv
 
 REM build typedb-all-windows archive
-bazel test //test/assembly:assembly --test_output=streamed --enable_runfiles --test_env=PATH --java_language_version=11 --javacopt="--release 11"
+bazel --output_user_root=C:/bzl test //test/assembly:assembly --test_output=streamed --enable_runfiles --test_env=PATH --java_language_version=11 --javacopt="--release 11"
```
flyingsilverfin, line 24: Shorten windows paths to avoid issues

dependencies/maven/artifacts.bzl
```diff
@@ -23,7 +23,6 @@ artifacts = [
     "com.google.code.findbugs:jsr305",
     "com.google.errorprone:error_prone_annotations",
     "com.google.guava:guava",
-    "com.google.guava:listenablefuture",
```
flyingsilverfin, line 26: The fix to explicitly include listenable future for windows building was not necesary, it turns out!

PR #6864, opened 2023-08-11
Configure logging to respect size and time limits

flyingsilverfin:
## What is the goal of this PR?

TypeDB's logger previously could over-run its expected archives size cap. This occurred because the logging framework expected to have both the time and size window configured for either to work correctly. 

TypeDB logger file configuration now has an updated set of options to control both archive retention policies for size and age.
```
log:
  output:
    file:
      type: file
      base-dir: server/logs        // renamed from 'directory'
      file-size-limit: 50mb        // renamed from 'file-size-cap'
      archive-grouping: month      // new option, using: minute(s) | hour(s) | day(s) | week(s) | month(s) | year(s)
      archive-age-limit: 1 year    // new option, using: <N> minute(s) | hour(s) | day(s) | week(s) | month(s) | year(s)
      archives-size-limit: 1gb     // renamed from 'archives-size-cap'
```

The `archive-grouping` option configures the rollover and naming policy of archives produced by the logger. 
The `archive-age-limit` option configures how long each archive files are kept. Note that old archives are only deleted when new ones are produced.

The execution semantics are that every period of time defined by 'grouping', files exceeding the time limit are asynchronously deleted, and then oldest files are asynchronously deleted until the total size cap is respected.

In the above example, log files would be compacted into archives monthly, with naming pattern like:
```
typedb_202306.0.log.gz
typedb_202307.0.log.gz
typedb_202308.0.log.gz
...
```

Where 1 year's worth of log archives are retained.

Resolves #6854. 

## What are the changes implemented in this PR?

* Add options to configure file archive creation and retention
* Rename existing options
* Introduce new value parsers for configuration file parsing
* We also introduce a new type of integration test, which boots up TypeDB server entirely, rather than just booting up the database manager directly. We can extend this test set up for future configuration and logging testing

PR #6868, opened 2023-08-15
Add logic to allow unlimited archive age or sizes

flyingsilverfin:

The change in #6864 enabled configuring log archive retention policies by age or time, but would actually fail if trying to set unlimited time and size limits. The user can now set age limits or aggregate size limits by using the value `0` for either `archive-age-limit` or `archives-size-limit` configurations respectively.

* Allow setting `0` for time and size limits for log file archives

### DIFF
--- a/server/logging/CoreLogback.java
+++ b/server/logging/CoreLogback.java
@@ -36,6 +36,8 @@ import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.Map;
 
+import static ch.qos.logback.core.CoreConstants.UNBOUNDED_TOTAL_SIZE_CAP;
+import static ch.qos.logback.core.CoreConstants.UNBOUND_HISTORY;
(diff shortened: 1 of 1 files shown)


server/logging/CoreLogback.java
```diff
@@ -36,6 +36,8 @@
 import java.util.HashMap;
 import java.util.Map;
 
+import static ch.qos.logback.core.CoreConstants.UNBOUNDED_TOTAL_SIZE_CAP;
+import static ch.qos.logback.core.CoreConstants.UNBOUND_HISTORY;
```
benjaminasmall, line 40: "ED"?

benjaminasmall (changes requested):

benjaminasmall (approved):

server/logging/CoreLogback.java
flyingsilverfin: Not our constants, can't do anything about em sadly

PR #6883, opened 2023-08-30
Allow type variables in insert and delete clauses

flyingsilverfin:

We allow using type variables in delete and insert clauses. This allows much more expressivity when doing deletes, in particular allowing disjunctions over types in the `match`, which are then operated over in turn in the `delete` clause to clean up relations or attributes.

This change resolves https://github.com/vaticle/typedb/issues/6755

* Allow type variables or labels to be used in insert queries and delete queries
* Update behaviour tests to test both insert and delete queries with variabilisied types

### DIFF
--- a/common/exception/ErrorMessage.java
+++ b/common/exception/ErrorMessage.java
@@ -402,46 +402,42 @@ public abstract class ErrorMessage extends com.vaticle.typedb.common.exception.E
                 new ThingWrite(14, "Attempted to re-insert pre-existing thing of matched variable '%s' as a new instance (isa) of type '%s'.");
         public static final ThingWrite THING_ISA_MISSING =
                 new ThingWrite(15, "The thing variable '%s' cannot be inserted as a new instance without providing its type (isa).");
-        public static final ThingWrite ILLEGAL_TYPE_VARIABLE_IN_INSERT =
-                new ThingWrite(16, "Illegal type variable '%s' found in insert query. Types can only be referred to by their labels in insert queries or then clauses.");
-        public static final ThingWrite ILLEGAL_TYPE_VARIABLE_IN_DELETE =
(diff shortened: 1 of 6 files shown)


common/exception/ErrorMessage.java
```diff
@@ -396,48 +396,52 @@ public static class ThingWrite extends ErrorMessage {
                 new ThingWrite(11, "Attempted to put an instance of '%s' with value '%s' that does not satisfy the regular expression '%s'.");
         public static final ThingWrite THING_IID_NOT_INSERTABLE =
                 new ThingWrite(12, "The variable '%s' tries to insert iid '%s'. IIDs are prohibited in insert clauses. You may want to query the variable using IID in the match clause.");
+        public static final ThingWrite THING_INSERT_ISA_NOT_THING_TYPE =
+                new ThingWrite(13, "The type '%s' cannot be used for creating an instance as it is not a thing type.");
         public static final ThingWrite THING_ISA_REINSERTION =
-                new ThingWrite(13, "Attempted to re-insert pre-existing thing of matched variable '%s' as a new instance (isa) of type '%s'.");
+                new ThingWrite(14, "Attempted to re-insert pre-existing thing of matched variable '%s' as a new instance (isa) of type '%s'.");
```
lveillard, line 402: Sorry for the intromission lol, my product manager soul got triggered 😂

Totally opinionated and I don't know the surrounding code so probably there is a good reason to do it like this. But I would personally not change the IDs of the existing errors in case any customer catches and routes them in the client side and for stack-overflow / forum consistency.

flyingsilverfin: That's alright - we're doing this intentionally right now, you still get consistency when using a specific TypeDB Version + error code :)

flyingsilverfin: You're right about the error codes catching though! In rust we'll do a bit of re-think of them - like versioning the changes and then prepending the version to the code.

Issue #6882, opened 2023-08-30
Deletes as inverses of Inserts


flyingsilverfin:
## Problem to Solve

Deletes should operate as the inverse of inserts. We should allow the introduction of new variables in the `delete` clause. 

As a summary, the result of a delete should be as if that pattern no longer exists in the database.

## Current Workaround

N/A

## Proposed Solution

1) We should make deletes 'best-effort', ie. if a matched answer has already deleted a concept, its re-deletion does nothing (no error)
2) We can introduce new variables/patterns in the `delete` clause.

This introduces interesting semantic questions, such as the idea that every pattern in the `delete` becomes optional/best-effort. This would require some thought.

Following that thread, there is an instinct that we should allow writing a `delete` without a `match`:
```
delete $x isa person;
```
In this case, we would expect all `person` instances to be gone.

However this becomes ambiguous when unbounded:
```
delete $x isa person, has name "john";
```
This can be interpreted as 1) delete all people if they have the name `john`, then delete the ownership and the name? Or alternatively 2) Delete all people, all names 'john', and all ownerships between them.

The latter is what would be implied if every part of the delete becomes optional.

This becomes better if we require that the `delete` be bounded by a `match`:
```
match
$x isa person;
delete 
$x isa person, has name 'john';
```
would more strongly imply we're going to find all people, and delete them and optionally also delete their name `john` if it exists.

PR #6884, opened 2023-09-01
Deletes are idempotent instead of throwing

flyingsilverfin:

We move partially towards the 'assert-style' deletion outlined in #6882 to allow idempotent deletion. Previously, deletes would throw when trying to delete a previously-removed/non-existent connection.

We now guarantee that all deletes will continue even if trying to delete a relationship, ownership, role player, or concept that doesn't exist (ie. was previously deleted).

For example, the following is now valid:
```
match 
$x has $a, has $b; $b = $a;
delete
$x has $a;
```
Because the same attribute would be generated twice by the `match`, the second delete would fail and throw an error. This is now an idempotent operation, which will continue without an error.

We also introduce a new operation to disallow a new class of semantically unsafe queries. We add an explicit 'runtime' type check before performing the idempotent delete operation. This means the following semantically invalid query will still throw an exception when any data matches:
```
match
$x isa person;
$id isa company-id;
delete
$x has $id;
```

* Relax delete semantics to allow idempotent deletion
* Add a 'runtime' type check that before trying to perform any delete-if-present operation, can throw a meaningful error if a connection being deleted is not allowed in the schema at all.

### DIFF
--- a/query/Inserter.java
+++ b/query/Inserter.java
@@ -33,8 +33,6 @@ import com.vaticle.typedb.core.concept.type.RoleType;
 import com.vaticle.typedb.core.concept.type.ThingType;
 import com.vaticle.typedb.core.concept.type.Type;
 import com.vaticle.typedb.core.concept.value.Value;
-import com.vaticle.typedb.core.logic.LogicManager;
-import com.vaticle.typedb.core.pattern.Conjunction;
 import com.vaticle.typedb.core.pattern.constraint.common.Predicate;
 import com.vaticle.typedb.core.pattern.constraint.thing.HasConstraint;
 import com.vaticle.typedb.core.pattern.constraint.thing.IsaConstraint;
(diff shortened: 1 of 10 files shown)


query/Inserter.java
```diff
@@ -113,18 +111,27 @@ public static Inserter create(Reasoner reasoner, ConceptManager conceptMgr, Type
...
+    public static void validate(Variable var, @Nullable Matcher matcher) {
         try (FactoryTracingThreadStatic.ThreadTrace ignored = traceOnThread(TRACE_PREFIX + "validate")) {
-            if (var.isThing()) validate(var.asThing());
+            if (var.isType()) validate(var.asType(), matcher);
+            else if (var.isThing()) validate(var.asThing());
             else if (var.isValue()) validate(var.asValue());
         }
     }
 
+    private static void validate(TypeVariable var, @Nullable Matcher matcher) {
```
flyingsilverfin, line 128: Drive-by fix: we should throw a useful exception if trying to use an unbound type variable in an insert query

PR #6878, opened 2023-08-22
Changes for audit logging

krishnangovindraj:
## What is the goal of this PR?
Refactors configuration & grpc service classes for audit logging.

## What are the changes implemented in this PR?
* Refactors grpc service implementation classes, introducing methods which subtypes can override to log events
* Refactors configuration `CoreConfigParser.Log.Output.File` to have filename awareness, allowing it to be re-used for audit log outputs configuration.

PR #6894, opened 2023-09-06
Upgrade speeDB to 2.6.0

flyingsilverfin:

Upgrade SpeeDB to 2.6.0. This upgrade does not change any configurations. Even so we observe a 3-5% performance improvement in read-heavy workloads.

Upgrade SpeeDB to 2.6.0.


flyingsilverfin:
This was rolled into #6878

PR #6761, opened 2023-03-09
Simplify Concept API & unify gRPC API with Cluster

dmitrii-ubskii:
## What is the goal of this PR?

We simplify the `Concept` interface by unifying methods that were previously split into their general and `Explicit` variants. This allows us to simplify the gRPC protocol and deduplicate the implementation of the methods. 

In addition, we adjust the way TypeDB server responds to database and server manager requests to align it with TypeDB Cluster.

## What are the changes implemented in this PR?

- update the protocol to the version in which we merged Core and Cluster gRPC services and simplified Concept API (https://github.com/vaticle/typedb-protocol/pull/183);
- simplify Concept API in correspondence with the changes in the protocol;
- replace boolean switches (`isKey`, `isInferred`, would-be `isExplicit`) with enums (reusing `Annotation.KEY` in place of `isKey`);
- make `Concept`/`Thing`/`TypeService` request handlers uniform and guard against missed requests by using `EnumMap` instead of `switch-case`

PR #6895, opened 2023-09-15
Simplify Concept API & unify gRPC API with Cluster (#6761)

dmitrii-ubskii:

We simplify the `Concept` interface by unifying methods that were previously split into their general and `Explicit` variants. This allows us to simplify the gRPC protocol and deduplicate the implementation of the methods.

In addition, we adjust the way TypeDB server responds to database and server manager requests to align it with TypeDB Cluster.

- update the protocol to the version in which we merged Core and Cluster gRPC services and simplified Concept API
(https://github.com/vaticle/typedb-protocol/pull/183);
- simplify Concept API in correspondence with the changes in the protocol;
- replace boolean switches (`isKey`, `isInferred`, would-be `isExplicit`) with enums (reusing `Annotation.KEY` in place of `isKey`);
- make `Concept`/`Thing`/`TypeService` request handlers uniform and guard against missed requests by using `EnumMap` instead of `switch-case`


server/TypeDBService.java
```diff
@@ -59,17 +68,85 @@ public class TypeDBService extends TypeDBGrpc.TypeDBImplBase {
...
+            responder.onCompleted();
+        } catch (RuntimeException e) {
+            LOG.error(e.getMessage(), e);
+            responder.onError(exception(e));
+        }
+    }
+
+    @Override
+    public void usersContains(UserManager.Contains.Req request, StreamObserver<UserManager.Contains.Res> responder) {
+        ErrorMessage error_message = USER_MANAGEMENT_NOT_AVAILABLE;
```
flyingsilverfin, line 94: Java doesn't use underscores :P

server/concept/ThingService.java
```diff
@@ -198,35 +192,97 @@ private void getPlayersByRoleType(Relation relation, UUID reqID) {
...
+        return conceptMgr.getRelationType(protoType.getLabel());
+    }
+
+    private AttributeType getAttributeType(ConceptProto.AttributeType protoType) {
+        return conceptMgr.getAttributeType(protoType.getLabel());
+    }
+
+    private RoleType getRoleType(ConceptProto.RoleType protoRoleType) {
+        RelationType relationType = conceptMgr.getRelationType(protoRoleType.getScope());
+        if (relationType != null) return relationType.getRelates(protoRoleType.getLabel());
```
flyingsilverfin, line 245: You can actually now call `conceptMgr.getType(Label.of(...)).asRoleType()` :) This looks a lot like what should already exist in ConceptManager

server/concept/ThingService.java
```diff
@@ -198,35 +192,97 @@ private void getPlayersByRoleType(Relation relation, UUID reqID) {
...
+        else return null;
+    }
+
+    private Thing getThing(ConceptProto.Thing.Req thingReq) {
+        return notNull(conceptMgr.getThing(ByteArray.of(thingReq.getIid().toByteArray())));
```
flyingsilverfin, line 250: I believe we had `getIid` as `getIID` capitalised before? That is more in line with our style

flyingsilverfin, line 250: Will `notNull` throw a user-friendly exception if it is null?

server/common/ResponseBuilder.java
```diff
@@ -703,20 +837,42 @@ public static TransactionProto.Transaction.ResPart getOwnersExplicitResPart(
     public static class Thing {
 
         public static ConceptProto.Thing protoThing(com.vaticle.typedb.core.concept.thing.Thing thing) {
-            ConceptProto.Thing.Builder protoThing = ConceptProto.Thing.newBuilder()
+            if (thing.isEntity()) return ConceptProto.Thing.newBuilder().setEntity(protoEntity(thing.asEntity())).build();
+            if (thing.isRelation()) return ConceptProto.Thing.newBuilder().setRelation(protoRelation(thing.asRelation())).build();
+            if (thing.isAttribute()) return ConceptProto.Thing.newBuilder().setAttribute(protoAttribute(thing.asAttribute())).build();
```
flyingsilverfin, line 842: these should be `else if`'s

server/concept/ThingService.java
dmitrii-ubskii: `getType()` does not (and did not) exist on `ConceptManager`. Are you saying I should add it?

server/concept/ThingService.java
dmitrii-ubskii: It throws a `MISSING_CONCEPT` ("Concept does not exist."), which is caught by the `TransactionService` and propagated to the user (`server/TransactionService.java:156`)

server/concept/ThingService.java
flyingsilverfin: Ah yeah you can add it like this (Sorry I forgot I had it in the Fetch PR):
```
  public Type getType(Label label) {
      TypeVertex vertex = graphMgr.schema().getType(label);
      if (vertex != null) {
          if (label.scope().isPresent()) return convertRoleType(vertex);
          else return convertThingType(vertex);
      } else return null;
  }
```
Then you can create a core Label structure from the message and use that

flyingsilverfin (approved):

PR #6913, opened 2023-10-11
Optional config values

flyingsilverfin:

We implement a new paradigm for configurations in TypeDB: optional primitive values. The outcome of this is that we have the following three top-level paradigms in our configuration file:

1. All configuration file options are present in the configuration file
2. 'Compound' (eg. nested blocks) of options that are optional are by convention protected by an `enable: true|false` option on the same level.
3. For any optional primitive/leaf configurations, they may be left empty (or by YAML conventions set to `null` or `~`).

Points 2 and 3 derive from point 1 in that they allow us to keep all configurations present, but introduce optionality for either entire blocks or leaf configurations.

* Introduce Optional primitive value parsers that can be used for future configurations and in TypeDB Enterprise
* Refactor Help menu generation and formatting for clarity

### DIFF
--- a/server/parameters/util/YAMLParser.java
+++ b/server/parameters/util/YAMLParser.java
@@ -151,6 +151,14 @@ public class YAMLParser {
             throw TypeDBException.of(ILLEGAL_CAST, className(getClass()), className(Restricted.class));
         }
 
+        public boolean isOptional() {
+            return false;
+        }
+
+        public Optional<?> asOptional() {
+            throw TypeDBException.of(ILLEGAL_CAST, className(getClass()), className(Optional.class));
+        }
+
         boolean isCompound() {
             return false;
         }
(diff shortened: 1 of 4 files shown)


server/parameters/util/YAMLParser.java
```diff
@@ -258,6 +265,40 @@ Restricted<T> asRestricted() {
             }
         }
 
+        public static class Optional<T> extends Value<java.util.Optional<T>> {
```
flyingsilverfin, line 268: New optional value parser that wraps a primitive value parser, the say way `Restricted` does.

server/parameters/util/YAMLParser.java
```diff
@@ -133,13 +126,7 @@ public Map<String, TYPE> parseFrom(YAML.Map yaml, String path, Predefined<?>...
 
             @Override
             public Help help(String path) {
-                if (valueParser.isPrimitive()) {
```
flyingsilverfin, line 136: Simplifications

server/parameters/util/YAMLParser.java
```diff
@@ -462,22 +508,24 @@ private static boolean isValidString(String durationString) {
 
         public enum TimePeriodName {
 
-            MINUTE("minute", "minutes", ChronoUnit.MINUTES),
-            HOUR("hour", "hours", ChronoUnit.HOURS),
-            DAY("day", "days", ChronoUnit.DAYS),
-            WEEK("week", "weeks", ChronoUnit.WEEKS),
-            MONTH("month", "months", ChronoUnit.MONTHS),
-            YEAR("year", "years", ChronoUnit.YEARS);
+            MINUTE("minute", "s", ChronoUnit.MINUTES),
```
flyingsilverfin, line 511: This approach makes the printed help menu much shorter

server/parameters/util/YAMLParser.java
```diff
@@ -243,8 +245,13 @@ public T parse(YAML yaml, String path) {
                 else throw TypeDBException.of(CONFIG_VALUE_ENUM_UNEXPECTED, path, value, allowed);
             }
 
-            public String help() {
-                return "<" + String.join("+", iterate(this.allowed).map(Object::toString).toList()) + ">";
+            public String description() {
```
flyingsilverfin, line 248: refactored terminology

PR #6916, opened 2023-10-16
Disallow explicit use of role type aliases

flyingsilverfin:

We disallow using role type aliases in labels, only permitting their use when querying for role types in relation constraints. This finalises the implementation of behaviours outlined in https://github.com/vaticle/typeql/issues/274#issuecomment-1766214447.

Given
```
define
employment sub relation, relates employee;
part-time-employment sub employment;
```
We now disallow
```
define
person plays part-time-employment:employee;
```
As this role type is an alias for `employment:employee` and therefore generally used incorrectly/confusing.

However, matching is still flexibly allowed:
```
match
(employee: $x) isa part-time-employment;
```

* Add validation to reject aliased role types in define, match, and undefine queries
* Ensure that role type aliases are allowed in `match` statements where a relation constraint is specified: `(role: $x) isa friendship;` remains permitted
* We update the BDD dependency, which required refactoring the paths to the query language tests


logic/Rule.java
```diff
@@ -175,6 +175,12 @@ public final int hashCode() {
     }
 
     public void validate(LogicManager logicMgr, ConceptManager conceptMgr) {
+        iterate(when.conjunctions())
+                .flatMap(c -> iterate(c.variables())).flatMap(v -> iterate(v.constraints()))
+                .filter(c -> c.isType() && c.asType().isLabel() && c.asType().asLabel().properLabel().scope().isPresent())
+                // only validate labels that are used outside of relation constraints - this allows role type aliases in relations
+                .filter(label -> label.owner().constraining().isEmpty() || iterate(label.owner().constraining()).anyMatch(c -> !(c.isThing() && c.asThing().isRelation())))
```
flyingsilverfin, line 182: Exception: allow aliases in relation constraints

concept/ConceptManager.java
```diff
@@ -325,6 +329,23 @@ private BiFunction<ConceptManager, TypeVertex, AttributeType> attributeTypeConst
         throw exception(TypeDBException.of(BAD_VALUE_TYPE, valueType));
     }
 
+    public void validateNotRoleTypeAlias(Label label) {
```
flyingsilverfin, line 332: We put role type alias validation in the `ConceptManager` to centralise the semantics in one place

query/Matcher.java
```diff
@@ -66,37 +67,43 @@ public class Matcher {
     private final Disjunction disjunction;
     private final Context.Query context;
 
-    public Matcher(Reasoner reasoner, TypeQLMatch query) {
-        this(reasoner, query, null);
+    public Matcher(Reasoner reasoner, ConceptManager conceptMgr, TypeQLMatch query) {
```
flyingsilverfin, line 70: Matcher now needs concept manager to validate the labels

PR #6927, opened 2023-10-31
Allow disconnected relations

flyingsilverfin:

We implement the suggestion from https://github.com/vaticle/typedb/issues/6920, which will allow relations to exist without role players for the duration of a write transaction. At the end of the transaction, any relations without role players are deleted (preserving the existing behaviour of automatically deleting relations without role players, but we delay the operation until commit time).

This change addresses https://github.com/vaticle/typedb/issues/6902, https://github.com/vaticle/typedb/issues/6920, 

We delay the deletion of relations until the end of the transaction. During commit, we delete any relations without role players.

Adding a test for this case led us to discover and fix a bug in the traversal engine

### DIFF
--- a/graph/vertex/impl/AttributeVertexImpl.java
+++ b/graph/vertex/impl/AttributeVertexImpl.java
@@ -589,6 +589,16 @@ public abstract class AttributeVertexImpl {
         }
 
         @Override
+        public ThingAdjacency.In ins() {
+            return attributeVertex.ins();
+        }
+
+        @Override
+        public ThingAdjacency.Out outs() {
+            return attributeVertex.outs();
+        }
+
+        @Override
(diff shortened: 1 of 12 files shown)


graph/vertex/impl/AttributeVertexImpl.java
```diff
@@ -588,6 +588,16 @@ public ValueSortable<VALUE> toValueSortable() {
             return this;
         }
 
+        @Override
+        public ThingAdjacency.In ins() {
```
flyingsilverfin, line 592: This is a bugfix where we tried to traverse a Sortable wrapper of a vertex but failed to find vertices because the inner vertex contained edges, but the wrapper did not - now we use it as a pass through instead.

PR #6947, opened 2023-11-22
Update the readme file: fix TypeQL link

izmalk:
## Release notes: usage and product changes

Update the readme to fix the TypeQL link.

## Implementation

Fixed the TypeQL link in the readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 [![TypeDB](./docs/banner.png)](https://typedb.com/introduction)
 
-<div align="center">
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,7 @@
 [![TypeDB](./docs/banner.png)](https://typedb.com/introduction)
 
+<div align="center">
```
flyingsilverfin, line 3: Can we get rid of this? It's ugly when you read the `md` file in plain text, which is the most common use case for us developers. I think them not being centered is okay right?

PR #6949, opened 2023-11-29
Fetch output format update: attribute type includes "value_type"

dmitrii-ubskii:
## Usage and product changes

We update the serializiation schema of TypeQL Fetch query outputs: attribute type serialization now includes its `value_type`. This change makes the output symmetric between raw values and attributes:

```json
{
"raw_value": { "value": "...", "value_type": "string" },
"attribute": { "value": "...", "type": { "label": "T", "value_type": "string", "root": "attribute" } }
}
```


concept/Concept.java
```diff
@@ -93,12 +93,16 @@ interface Readable {
 
         boolean isType();
 
+        boolean isAttributeType();
```
flyingsilverfin, line 96: Let's remove this and cast through `isType` -- discussed in person

concept/answer/ReadableConceptTree.java
```diff
@@ -280,6 +282,14 @@ private static String getType(Type type) {
                         Map.CURLY_RIGHT;
```
flyingsilverfin, line 282: You'll end up with a switch in here if its an attribute type or else

PR #6956, opened 2023-12-12
Fetch query can infer fetched attributes

### LINKED ISSUE #6952: Attribute fetch should trigger reasoning
## Description

Fetching attributes should trigger reasoning if there are rules to apply and return inferred ownerships, as well as stored ones.

## Environment

1. TypeDB distribution: All
2. TypeDB version: 2.25.x
3. Environment: All
4. Client and version: -
5. Other details: -

## Reproducible Steps

1. Set up
- define a schema with an inferred attribute ownership

2. Test/Query

Run a fetch query in a transaction with inference enabled 

3. Unexpected result

The inferred attributes will be absent.

## Expected result

Inferrable attribuet is present.

flyingsilverfin:
## Usage and product changes

We fix #6952 , by allowing attribute fetching to trigger reasoning. For example;

Given:
```
define
rule has-name: when { $x isa person; } then { $x has name "John"; };
```

Then the query:
```
match $x isa person;
fetch $x: name;
```

Can now retrieve `name "John"` as part of the fetch response.

### DIFF
--- a/concept/answer/ConceptMap.java
+++ b/concept/answer/ConceptMap.java
@@ -32,6 +32,7 @@ import com.vaticle.typedb.core.traversal.common.Modifiers;
 import com.vaticle.typeql.lang.common.Reference;
 import com.vaticle.typeql.lang.common.TypeQLVariable;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -48,13 +49,11 @@ import static java.util.Collections.unmodifiableMap;
 
 public class ConceptMap {
 
+    public static ConceptMap EMPTY = new ConceptMap(Collections.emptyMap());
+
     private final Map<Retrievable, ? extends Concept> concepts;
     private final Explainables explainables;
     private final int hash;
-
-    public ConceptMap() {
-        this(new HashMap<>());
-    }
 
     public ConceptMap(Map<Retrievable, ? extends Concept> concepts) {
         this(concepts, new Explainables());
--- a/query/Fetcher.java
+++ b/query/Fetcher.java
@@ -30,7 +30,10 @@ import com.vaticle.typedb.core.concept.answer.ReadableConceptTree;
 import com.vaticle.typedb.core.concept.thing.Thing;
 import com.vaticle.typedb.core.concept.type.AttributeType;
 import com.vaticle.typedb.core.concept.type.ThingType;
+import com.vaticle.typedb.core.pattern.Conjunction;
 import com.vaticle.typedb.core.pattern.Disjunction;
+import com.vaticle.typedb.core.pattern.variable.ThingVariable;
+import com.vaticle.typedb.core.pattern.variable.TypeVariable;
 import com.vaticle.typedb.core.reasoner.Reasoner;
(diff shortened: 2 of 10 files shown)


concept/answer/ConceptMap.java
```diff
@@ -48,14 +49,12 @@
 
 public class ConceptMap {
 
+    public static ConceptMap EMPTY = new ConceptMap(Collections.emptyMap());
```
flyingsilverfin, line 52: This might improve performance a bit - ConceptMaps are immutable and so the empty constructor can just be re-used as a singleton

query/Fetcher.java
```diff
@@ -259,15 +266,30 @@ ReadableConceptTree.Node.Map execute(Reasoner reasoner, ConceptManager conceptMg
...
-                attributeFetches.forEach(attrFetch -> entries.add(
-                        attrFetch.name(),
-                        new ReadableConceptTree.Node.List(
-                                thing.getHas(attrFetch.attributeTypes(), emptySet()).map(ReadableConceptTree.Node.Leaf::new).toList()
-                        )
-                ));
+                Identifier.Variable.Name attrVar = Identifier.Variable.namedConcept("attr");
+                attributeFetches.forEach(attrFetch -> {
+                    // use a new context to avoid taking limits into sub-query
+                    Context.Query subContext = new Context.Query(context.parent(), context.options());
```
flyingsilverfin, line 272: Bug fix: we should not pass the outer Context into the inner query, since it may contain a limit on the Producer context.

query/Fetcher.java
```diff
@@ -259,15 +266,30 @@ ReadableConceptTree.Node.Map execute(Reasoner reasoner, ConceptManager conceptMg
...
+                    List<ReadableConceptTree.Node.Leaf<com.vaticle.typedb.core.concept.thing.Attribute>> c = reasoner.execute(createAttributeQuery(attrFetch.attributeType, attrVar, thing),
+                                    list(), TypeQLQuery.Modifiers.EMPTY, subContext, ConceptMap.EMPTY)
+                            .map(cm -> new ReadableConceptTree.Node.Leaf<>(cm.get(attrVar).asAttribute()))
+                            .toList();
+                    entries.add(attrFetch.name(), new ReadableConceptTree.Node.List(c));
+                });
                 return entries;
             }
 
+            private Disjunction createAttributeQuery(AttributeType attributeType, Identifier.Variable.Name attributeVariable, Thing attrOwner) {
```
flyingsilverfin, line 282: We now convert the attribute fetch into a reasoner query

query/Fetcher.java
```diff
@@ -331,10 +358,11 @@ public static Projection create(Reasoner reasoner, ConceptManager conceptMgr, Co
                 )) {
                     throw TypeDBException.of(SUBQUERY_UNBOUNDED, key.label());
                 }
-
+                // use a new context to avoid taking limits into sub-query
+                Context.Query subContext = new Context.Query(context.parent(), context.options());
```
flyingsilverfin, line 362: Bug fix: same as before, we do not pass the outer query's context into the inner one, creating a new context from the transaction context directly.

PR #6966, opened 2024-01-11
Improve fatal error handling and shutdown

flyingsilverfin:
## Usage and product changes

We streamline the error handling infrastructure by requiring all TypeDBExceptions have an associated error code, avoiding simple wrappers of Java or Storage exceptions without any context. We also ensure that all errors originating in TypeDB, such as TypeQL parser errors, are wrapped in TypeDBExceptions.

The result is the following invariant:
- TypeDB will always throw either unexpected Java exceptions, or TypeDBExceptions
- All TypeDBExceptions have error messages and therefore error codes

Additionally, Diagnostics are now implemented via a Singleton in order to inject different error reporting for TypeDB Core or Cloud. This means that all tests must now initialise a Diagnostics singleton to no-op via `Diagnostics.initialiseNoop()` before executing.

## Implementation

- Make all `TypeDBException`s contain `ErrorMessage`s
  - Introduce `toUnchecked()` on `TypeDBCheckedException` to facilitate converting a checked exception to an unchecked one. 
- Convert `Diagnostics` into a Singleton instance, which has an `ErrorReporter`
  - Introduce `CoreErrorReporter` which handles TypeDB exceptions only
- We fix a bug in server bootup: when receiving an error while booting any single database, we cleanly close/shutdown all resources attached to all databases, regardless of whether they have successfully booted or not.


database/CoreDatabaseManager.java
```diff
@@ -102,13 +106,23 @@ public boolean isOpen() {
...
+                    .map(directory ->
+                            CompletableFuture.runAsync(() -> {
+                                String name = directory.getName();
+                                CoreDatabase database = databaseFactory.databaseLoadAndOpen(this, name);
+                                databases.put(name, database);
+                            }, Executors.async1())
+                    ).collect(Collectors.toList());
+            try {
+                // once all future complete, we will catch any exceptions in order to close all databases cleanly
+                CompletableFuture.allOf(dbLoads.toArray(new CompletableFuture[0])).join();
```
flyingsilverfin, line 120: This will help cleanly shut down if any of the databases error during bootup.

PR #6968, opened 2024-01-12
Disable diagnostics in Docker and Assembly tests

flyingsilverfin:
## Usage and product changes

We disable diagnostics in Docker and Assembly tests. This also required allow the TypeDB Runner library to accept and transmit arguments to the binary it boots.

## Implementation

* Implement option pass-through for the TypeDB Runner
* Docker and Assembly tests pass `--diagnostics.reporting.enable = false` to the server on bootup

### DIFF
--- a/test/assembly/DockerTest.java
+++ b/test/assembly/DockerTest.java
@@ -48,11 +48,11 @@ public class DockerTest {
         String imagePath = Paths.get("assemble-docker.tar").toAbsolutePath().toString();
         ProcessResult result = execute("docker", "load", "-i", imagePath);
         LOG.info(result.outputString());
-        // TODO: disable diagnostics reporting
         StartedProcess typeDBProcess = executor.command(
                 "docker", "run", "--name", "typedb",
                 "--rm", "-t", "-p", String.format("%d:%d", typeDBPort, typeDBPort),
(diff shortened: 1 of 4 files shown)


test/assembly/DockerTest.java
```diff
@@ -48,11 +48,11 @@ public void bootup() throws InterruptedException, TimeoutException, IOException
         String imagePath = Paths.get("assemble-docker.tar").toAbsolutePath().toString();
         ProcessResult result = execute("docker", "load", "-i", imagePath);
         LOG.info(result.outputString());
-        // TODO: disable diagnostics reporting
         StartedProcess typeDBProcess = executor.command(
                 "docker", "run", "--name", "typedb",
                 "--rm", "-t", "-p", String.format("%d:%d", typeDBPort, typeDBPort),
-                "bazel:assemble-docker"
+                "bazel:assemble-docker",
+                "/opt/typedb-all-linux-x86_64/typedb", "server", "--diagnostics.reporting.enable=false"
```
flyingsilverfin, line 55: Override default docker command with custom one that also has arguments.

It might be better for the default Docker command to become an 'entrypoint', which means the `typedb server` part of the command are non-overridable.

PR #6973, opened 2024-01-25
Fix Role resolution in rules

krishnangovindraj:
## Usage and product changes
Fixes how rules resolve roles, fixing validation that types referenced in rules cannot be deleted.

## Implementation
* Translates scoped role name as it occurs in the rule to the actual scoped name of the role (i.e., from `inheritor:role`  to `declarer:role`) when computing which types occur in the rule.

### DIFF
--- a/graph/TypeGraph.java
+++ b/graph/TypeGraph.java
@@ -48,6 +48,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.NavigableSet;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


krishnangovindraj:
The bug this fixes would be caught at commit time anyway.
Further, it makes the validation for `set_supertype` more complicated if we want to achieve consistency.

krishnangovindraj:
I forget type-inference automatically downcasts to subtypes based on their relates/owns/plays constraints. This, role types cannot always be uniquely resolved.

PR #6978, opened 2024-01-29
Generate persistent server ID

flyingsilverfin:
## Usage and product changes

We update the server ID to be generated once, then persisted into the server's data directory. A fallback ID of of `_0` is also used if this fails for any reason.

## Implementation

Generate a length 16 string of  upperchase ASCII character and digits, persisted into the data directory. On bootup we read this value, if it exists.

### DIFF
--- a/server/TypeDBServer.java
+++ b/server/TypeDBServer.java
@@ -204,27 +204,23 @@ public class TypeDBServer implements AutoCloseable {
 
     protected String serverID() {
         try {
-            return serverIDStored();
+            Path serverIDFile = config().storage().dataDir().resolve(SERVER_ID_FILE_NAME);
+            if (serverIDFile.toFile().exists()) {
+                return Files.readString(serverIDFile);
+            } else {
+                Random random = new Random();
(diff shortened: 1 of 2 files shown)


server/TypeDBServer.java
```diff
@@ -202,13 +204,23 @@ Constants.DIAGNOSTICS_REPORTING_URI, new CoreErrorReporter()
...
+            if (serverIDFile.toFile().exists()) {
+                return Files.readString(serverIDFile);
+            } else {
+                Random random = new Random();
+                String serverID = IntStream.range(0, SERVER_ID_LENGTH).boxed()
+                        .map(i -> SERVER_ID_ALPHABET.charAt(random.nextInt(SERVER_ID_ALPHABET.length())))
+                        .map(String::valueOf)
+                        .collect(Collectors.joining());
+                Files.writeString(serverIDFile, serverID);
+                return serverID;
```
dmitrii-ubskii, line 217: Why aren't we using UUID here?

dmitrii-ubskii (approved):

PR #6981, opened 2024-02-06
Redesign schema modification capabilities

krishnangovindraj:
## Usage and product changes

We redesign schema modification to allow much more flexible in-place changes to the database schema. We relax various schema invariants within a schema write transaction, to allow moving and editing schema types on the fly. However, the data is validated against the schema consistency at each step, allowing full and safe use of TypeDB's existing Concept and Query API. Before committing, we can restore schema invariants guided by TypeDB's exceptions API (`ConceptManager.getSchemaExceptions()`). 

#### Expected schema migration workflow

This change facilitates large-scale database schema migration. We expect the following workflow to be adopted:

1. Open a schema session, and a write transaction. This blocks writes anywhere on the system.
2. Mutate the schema incrementally. Mutations that _expand_ schema are always possible and cheap, mutations that _restrict_ the schema are validated against the existing data for conformance to the new schema. All schema states you move through must match the current state of the data.
  a. If your data does not fit the new schema state, in 2.x you will get an exception on commit and it will roll back. You must open a data session+transaction to mutate the data into the shape it is expected to be and commit this. Then go back into schema session+transaction and retry the schema mutation.
  b. In TypeDB 3.0 these operations will be possible all within one schema write transaction, smoothing out the schema migration workflow.
3. To make schema migration simpler, some schema invariants are relaxed *within a schema write transaction*:
  a. **Dangling overrides are allowed**: overridden types (`... as TYPE`) are allowed to refer to types that are not overridable at that place in the schema. This is common when moving a type from one supertype to a different supertype.
  b. **Redeclarations are allowed**: Declarations of `owns`, `plays`,  or annotations, may be duplicated in child types. This facilitates moving types from one supertype to a different supertype, or moving declarations up or down the type hierarchy.
  c. **Relaxed abstract ownership**: Types may own abstract attribute types without themselves being abstract.
4. All of these invariants **must be restored before commit**, or the transaction will fail and the changes will be rolled back. To retrieve the set of errors that must be fixed before commit, use the api `ConceptManager.getSchemaExceptions()` (`transaction.concepts().getSchemaExceptions()` in most drivers).
 

Operations that _expand_ the schema capabilities:
- Adding a new type
- Adding a `plays` or `owns`
- Removing an override
- Removing an annotation
- Removing abstractness

Operations that _restrict_ the schema capabilities:
- Removing a type
- Removing a `plays` or `owns`
- Adding an override
- Adding an annotation
- Adding abstractness

## Implementation
* Implements a new `SubtypeValidation` class with static methods for validating the integrity of the subtrees of the types modified by a schema operation
  - Validation confirming whether the schema modification leaves type itself in a valid state wrt its ancestors is done as before in the modification functions .
  -   The ancestors do not require validation as their integrity is unaffected 
* Move some existing validation of declarations into `DeclarationValidation`

## Notes 
Only a valid schema can be committed. The validation considers the following properties, which are checked either at 1) immediately, at operation time or 2) deferred (requested via `transaction.concepts.getSchemaExceptions` or on commit):

#### Immediate validation:
* A relation/entity/attribute may only subtype another relation/entity/attribute.
  - Further, attribute types must have the same value type as their supertype (unless their supertype is the root attribute type)
* A type which is referenced in a rule may not be deleted.
* The data is always valid with respect to the schema.
* Only abstract attribute type can have subtypes.

#### Deferred validation:
Tese have been delayed to commit time to allow certain schema modifications to be performed with data in place.

* A Plays/Owns declarations may only override a types which is
  a) a supertype of the role-type/attribute-type being declared as being owned/played
  b) a type which would otherwise be played/owned via inheritance.
* There are no redundant re-declarations of owns/plays in the schema:
  - A redeclaration of an inherited plays is always redundant
  - A redeclaration of an inherited ownership is redundant if it does not make the annotations more strict.
* A non-abstract entity/relation/attribute type may not 'own' an abstract attribute type.

#### Data integrity invariants
There are never instances of types, ownerships or relations which are not allowed in the schema. Violations are:
* A vertex exists for a type that was deleted
* An instance of a relation type 'relates' an instance of a roletype it does not declare, or inherit, or is hidden by an overriding 'plays' declaration
* An instance of an entity/relation/attribute type 'owns' an attribute of a type it does not declare, or inherit, or is hidden by an overriding 'owns' declaration
* An instance of an entity/relation/attribute type 'plays' a role of a type it does not declare, or inherit, or is hidden by an overriding 'plays' declaration
* An ownership violates its `@key` or `@unique` constraint.

This is validated when:
  - Undefining types, roles
  - Undefining owns & plays declarations
  - Changing the supertype of a given type
  -  Introducing an override which causes a role or attribute to be "hidden".
  - Setting a type to be abstract.
  - Increasing the strictness of annotations on an ownership.


concept/type/impl/RoleTypeImpl.java
```diff
@@ -164,20 +163,6 @@ public void delete() {
         vertex.delete();
     }
 
-    @Override
-    void validateDelete() {
-        super.validateDelete();
-        if (getInstances().first().isPresent()) {
-            throw exception(TypeDBException.of(INVALID_UNDEFINE_RELATES_HAS_INSTANCES, getLabel()));
-        }
-    }
```
krishnangovindraj, line 173: Again, moved validation to unset relates. The other option is to move the validation to delete here & for owns.

concept/type/impl/ThingTypeImpl.java
```diff
@@ -494,8 +498,10 @@ public List<TypeDBException> exceptions() {
         return concatToList(
                 validateIsAbstractOrOwnedAttributeTypesNotAbstract(),
                 validateIsAbstractOrPlayedRoleTypesNotAbstract(),
-                validateOverriddenOwnedAttributeTypesAreSupertypes(),
-                validateOverriddenPlayedRoleTypesAreSupertypes()
+                validateOverriddenOwnedAttributeTypesAreSupertypesAndInherited(),
+                validateOverriddenPlayedRoleTypesAreSuperTypesAndInherited(),
+                validatePlaysAreNotRedeclared(),
+                validateOwnsAreNotRedeclared()
```
krishnangovindraj, line 504: Should we move the implementation of all these methods to `SchemaValidation`?

krishnangovindraj (commented):
self review

concept/validation/SchemaValidation.java
```diff
@@ -0,0 +1,407 @@
...
+            Set<RoleType> noLongerRelates = new HashSet<>();
+            RelationType oldSuperType = relationType.getSupertype();
+            oldSuperType.getRelates(TRANSITIVE).filter(roleType -> !roleType.isRoot()).forEachRemaining(noLongerRelates::add);
+            newSupertype.getRelates(TRANSITIVE).forEachRemaining(noLongerRelates::remove);
+            validateNoBrokenOverrides(relationType, noLongerRelates, exceptions);
+            validateNoLeakedInstances(relationType, noLongerRelates, exceptions);
+            return exceptions;
+        }
+
+        private static void validateNoBrokenOverrides(RelationType relationType, Set<RoleType> removed, List<TypeDBException> acc) {
```
flyingsilverfin, line 107: Overriddes should all be validated at "commit" time

concept/validation/SchemaValidation.java
```diff
@@ -0,0 +1,407 @@
...
+        }
+
+        private static void validateNoLeakedInstances(RelationType relationType, Set<RoleType> removedOrHidden, List<TypeDBException> acc) {
+            if (removedOrHidden.isEmpty()) return;
+            List<RoleType> overriddenHere = new ArrayList<>();
+            Iterators.iterate(removedOrHidden)
+                    .filter(roleType -> relationType.getInstances(EXPLICIT).anyMatch(instance -> instance.getPlayers(roleType).hasNext()))
+                    .forEachRemaining(roleType -> {
+                        acc.add(TypeDBException.of(INVALID_UNDEFINE_RELATES_HAS_INSTANCES, relationType.getLabel(), roleType.getLabel()));
+                        overriddenHere.add(roleType);
```
flyingsilverfin, line 128: Same bug here

concept/validation/SchemaValidation.java
```diff
@@ -0,0 +1,407 @@
...
+                    .forEachRemaining(roleType -> {
+                        acc.add(TypeDBException.of(OVERRIDDEN_RELATED_ROLE_TYPE_NOT_INHERITED, relationType.getLabel(), roleType.getLabel(), roleType.getSupertype().getLabel()));
+                        overriddenHere.add(roleType);
+                    });
+            removed.removeAll(overriddenHere);
+            relationType.getSubtypes(EXPLICIT).forEachRemaining(subtype -> validateNoBrokenOverrides(subtype, removed, acc));
+            removed.addAll(overriddenHere);
+        }
+
+        private static void validateNoLeakedInstances(RelationType relationType, Set<RoleType> removedOrHidden, List<TypeDBException> acc) {
```
flyingsilverfin, line 121:
```
validateInstances(RelationType relationType, Set<RoleType> noLongerInherited, List<TypeDBException> exceptions)
```

concept/validation/SchemaValidation.java
flyingsilverfin, line 107:
```
validateOverrides(RelationType relationType, Set<RoleType> noLongerInherited, List<TypeDBException> exceptions)
```

concept/validation/SchemaValidation.java
```diff
@@ -0,0 +1,407 @@
...
+
+        public static String format(String relationType, String roleType, @Nullable String overridenRoleType) {
+            return TypeQL.type(relationType).relates(roleType, overridenRoleType).toString(false);
+        }
+
+    }
+
+    public static class Plays {
+
+        public static List<TypeDBException> validateAdd(ThingType thingType, RoleType added, @Nullable RoleType overridden) {
```
flyingsilverfin, line 147: rename:
1. ValidateOverride(overridden)

concept/type/impl/AttributeTypeImpl.java
```diff
@@ -139,6 +142,10 @@ public void setSupertype(AttributeType superType) {
         } else if (!superType.isAbstract()) {
             throw exception(TypeDBException.of(ATTRIBUTE_NEW_SUPERTYPE_NOT_ABSTRACT, superType.getLabel()));
         }
+        SchemaValidation.throwIfNonEmpty(Iterators.link(
```
flyingsilverfin, line 145: Let's split the stuff above here into `validateDeclaration`-type methods, so we have some symmetry with `SubtypeValidation`.

concept/type/impl/ThingTypeImpl.java
```diff
@@ -785,71 +838,11 @@ private static void validateSchema(
...
-            }
-        }
-
-        private static void validateDataUnique(ThingTypeImpl ownerType, Thing instance, AttributeTypeImpl attributeType) {
-            instance.getHas(attributeType).forEachRemaining(attr -> {
-                if (compareSize(attr.getOwners(ownerType), 1) != 0) {
-                    throw ownerType.exception(TypeDBException.of(OWNS_UNIQUE_PRECONDITION, attributeType.getLabel(), ownerType.getLabel()));
-                }
-            });
+        public static int compareAnnotationsPermissive(Set<TypeQLToken.Annotation> first, Set<TypeQLToken.Annotation> second) {
```
flyingsilverfin, line 841: Make two boolean methods:
```
        public static boolean isFirstStricter(Set<TypeQLToken.Annotation> first, Set<TypeQLToken.Annotation> second) {

        }

        public static boolean isFirstStrictOrEqual(Set<TypeQLToken.Annotation> first, Set<TypeQLToken.Annotation> second) {

        }

```

flyingsilverfin (changes requested):

concept/type/impl/ThingTypeImpl.java
```diff
@@ -277,34 +269,36 @@ public void setOwns(AttributeType attributeType, AttributeType overriddenType) {
     @Override
     public void setOwns(AttributeType attributeType, @Nullable AttributeType overriddenType, Set<Annotation> annotations) {
         validateIsNotDeleted();
-        OwnsImpl.of(this, (AttributeTypeImpl) attributeType, overriddenType, annotations);
+        if (attributeType.isRoot()) {
+            throw exception(TypeDBException.of(ROOT_ATTRIBUTE_TYPE_CANNOT_BE_OWNED));
+        }
```
flyingsilverfin, line 274: let's put this in `validateAdd`

concept/type/impl/ThingTypeImpl.java
```diff
@@ -277,34 +269,36 @@ public void setOwns(AttributeType attributeType, AttributeType overriddenType) {
...
+        DeclarationValidation.Owns.validateAdd(this, attributeType, annotations).forEach(e -> { throw exception(e); });
```
flyingsilverfin, line 275: Let's just throw the first one explicitly

concept/validation/SubtypeValidation.java
```diff
@@ -0,0 +1,464 @@
...
+import static com.vaticle.typeql.lang.common.TypeQLToken.Annotation.UNIQUE;
+import static java.util.Collections.emptySet;
+
+public class SubtypeValidation {
+
+    @SafeVarargs
+    public static Optional<String> collectExceptions(List<TypeDBException>... validationErrors) {
+        String combinedMessage = Stream.concat(
+                Stream.of(""),
+                Arrays.stream(validationErrors).flatMap(Collection::stream).map(TypeDBException::getMessage)
```
flyingsilverfin, line 76: formatting's a bit funny here?

flyingsilverfin (approved):

concept/validation/SubtypeValidation.java
krishnangovindraj: Looks ok to me?

PR #6983, opened 2024-02-13
Fix commit-time cleanup of chains of empty relations

flyingsilverfin:
## Usage and product changes

In cases where chains of empty relations existed, for example `r1 -> r2 -> r3... -> rX`,  it was possible that TypeDB did not correctly execute the automatic relation cleanup due to the order the relations happened to be traversed.

TypeDB now retries the commit-time cleanup of empty relations until there are no new deletions. We accept the higher runtime cost of this operation since 1) we expect the number of modified relations in a transaction to be relatively small (no more than thousands) 2) relation chains are relatively rare 3) relation chains that must clean up in dependent fashion are extremely rare.

## Implementation

* Add a loop that retries relation cleanup until no more relations are deleted.

### DIFF
--- a/concept/ConceptManager.java
+++ b/concept/ConceptManager.java
@@ -376,7 +376,7 @@ public final class ConceptManager {
             deleted = graphMgr.data().writeVertices()
                     .filter(v -> v.existence().equals(STORED) && v.isModified() && v.encoding().equals(RELATION))
                     .map(v -> ThingImpl.of(this, v).asRelation().deleteIfNoPlayer())
-                    .anyMatch(d -> d);
+                    .reduce(false, ((acc, val) -> acc || val));
         }
     }
 
--- a/concept/ConceptManager.java
+++ b/concept/ConceptManager.java
@@ -371,10 +371,13 @@ public final class ConceptManager {
     }
 
     public void cleanupRelations() {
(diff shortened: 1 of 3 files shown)


concept/ConceptManager.java
```diff
@@ -371,10 +371,13 @@ else return list(getRootThingType(), getRootRelationType().getRelates().first().
...
-        graphMgr.data().writeVertices()
-                .filter(v -> v.existence().equals(STORED) && v.isModified() && v.encoding().equals(RELATION))
-                .forEachRemaining(v -> ThingImpl.of(this, v).asRelation().deleteIfNoPlayer());
-
+        boolean deleted = true;
+        while (deleted) {
+            deleted = graphMgr.data().writeVertices()
+                    .filter(v -> v.existence().equals(STORED) && v.isModified() && v.encoding().equals(RELATION))
+                    .map(v -> ThingImpl.of(this, v).asRelation().deleteIfNoPlayer())
+                    .anyMatch(d -> d);
```
krishnangovindraj, line 379: Does `anyMatch` return eagerly?
Consider replacing with `reduce` ?

krishnangovindraj (approved):
Looks good to me!

PR #6998, opened 2024-03-06
Storage layer isolation tests based on hermitage

krishnangovindraj:
## Usage and product changes

Adds test cases covering well known non-serialisable executions of concurrent transactions.

We verify that TypeDB's storage layer currently supports Snapshot Isolation and fails as expected in the tests that would show Serialisability - notably Write Skew and the IMP case. 

One place TypeDB's storager layer currently deviates from 'standard' Snapshot Isolation is that **it doesn't not prevent concurrent transactions from writing the same key with different values**. This manifests in two ways: concurrent commits are allows to write the same key with different values (G0 case) without failing. Additionally, the Cursor (lost update) case fails in which x=x+1 is executed by two concurrent transactions that both succeed and one update is clobbered by the other without prompting a user retry via an error.

Note that this PR does not verify TypeDB's semantics+isolation, only the low-level storage layer it depends on.

## Implementation

Introduces storage layer isolation tests outlined by the [hermitage](https://github.com/ept/hermitage) repository .
Though the papers referred there are the source of truth, the examples are heavily inspired by [these notes](https://sitano.github.io/theory/databases/2019/07/30/tx-isolation-anomalies/)

At our current isolation level many of these fail. These have been marked with the `fails_without_serializable` macro.

### Notes

Applying isolation to TypeDB:
1. TypeDB can have write-write dependencies, imagined in the serial timeline of commits
2. TypeDB cannot have transactions fail because of write-write dependencies, because we can never have write-write dependency cycles

It's helpful to think about what our snapshotting does to the timelines and the precedence graph. This is explained well in section 3.2 of the Postgres [_Serializable Snapshot Isolation_](https://arxiv.org/pdf/1208.4179.pdf) paper

### DIFF
--- a/storage/tests/test_isolation.rs
+++ b/storage/tests/test_isolation.rs
@@ -25,9 +25,16 @@ use storage::{
     key_value::{StorageKey, StorageKeyArray, StorageKeyReference},
     keyspace::keyspace::KeyspaceId,
     snapshot::error::{SnapshotError, SnapshotErrorKind},
+    snapshot::snapshot::WriteSnapshot,
     MVCCStorage,
 };
 use test_utils::{create_tmp_dir, init_logging};
+
+/// We use value_k_t to represent the value written by transaction t for the key k
+///     and read_k_t to represent the value read by transaction t for the key k
+///
+/// Note that there are multiple ways to disallow each anomaly.
+/// The tests are written for our current model and may have to be revised if it changes.
 
 const KEYSPACE_ID: KeyspaceId = 0;
 const KEY_1: [u8; 4] = [0x0, 0x0, 0x0, 0x1];
@@ -37,6 +44,12 @@ const VALUE_1: [u8; 1] = [0x0];
 const VALUE_2: [u8; 1] = [0x1];
 const VALUE_3: [u8; 1] = [0x88];
 const VALUE_4: [u8; 1] = [0x99];
+
+macro_rules! fails_without_serializability  {
+    ($x:expr) => {
+        assert!(!($x));
+    };
+}
 
 fn setup_storage(storage_path: &PathBuf) -> MVCCStorage {
     let mut storage = MVCCStorage::new(Rc::from("storage"), storage_path).unwrap();
@@ -88,8 +101,6 @@ fn commits_isolated() {
 /// 1. Open two snapshots on the same version
 /// 2. S1 - get_required(K1)
 /// 3. S2 - delete(K1)
-///
-///
 #[test]
 fn g0_update_conflicts_fail() {
     init_logging();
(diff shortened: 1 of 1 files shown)


storage/tests/test_isolation.rs
```diff
@@ -122,3 +130,335 @@ fn g0_update_conflicts_fail() {
...
+    let snapshot_1 = storage.open_snapshot_write();
+    let snapshot_2 = storage.open_snapshot_write();
+
+    let read_11 = snapshot_1.get::<128>(key_1);
+    snapshot_2.put_val(key_1.to_owned(), ByteArray::copy(value_12.bytes()));
+    snapshot_2.put_val(key_2.to_owned(), ByteArray::copy(value_22.bytes()));
+    let read_21 = snapshot_1.get::<128>(key_2);
+
+    assert_eq!(value_10.bytes(), read_11.unwrap().bytes());
+    assert_eq!(value_20.bytes(), read_21.unwrap().bytes());
```
krishnangovindraj, line 324: t1 reads key_1 before t2 writes, but reads key_2 after t2 writes. In our case, t1 actually reads both values before t2 writes anything.

(But then if t2 now commits, the reads would no longer be valid and t1 may have to be aborted)

storage/tests/test_isolation.rs
```diff
@@ -122,3 +130,335 @@ fn g0_update_conflicts_fail() {
...
+        snapshot_2.put_val(key_2.to_owned(), ByteArray::copy(value_1.bytes()));
+    }
+
+    let result_1 = snapshot_1.commit();
+    let result_2 = snapshot_2.commit();
+
+    assert!(result_1.is_ok() && result_2.is_ok());
+    let reader_after = storage.open_snapshot_read();
+    let sum: u8 = reader_after.get::<128>(key_1).unwrap().bytes()[0] + reader_after.get::<128>(key_2).unwrap().bytes()[0];
+    assert!(sum <= 1);
```
krishnangovindraj, line 364: t1 (t2) reads key_1 (key_2), and writes key_2 (key_1) in a way that key_1 xor key_2 = 1
This fails, as expected

storage/tests/test_isolation.rs
```diff
@@ -122,3 +130,335 @@ fn g0_update_conflicts_fail() {
...
+}
+
+#[test]
+fn otv() {
+    init_logging();
```
krishnangovindraj, line 369: I don't think I got this right.

storage/tests/test_isolation.rs
```diff
@@ -122,3 +130,335 @@ fn g0_update_conflicts_fail() {
...
+        None => false
+    };
+    let delete_went_first_2: bool = match read_2_after {
+        Some(x) => {
+            assert_eq!(value_30.bytes()[0], x.bytes()[0]);
+            true
+        }
+        None => true
+    };
+    assert_eq!(delete_went_first_1, delete_went_first_2);
```
krishnangovindraj, line 463: Based on the PMP from here: https://github.com/ept/hermitage/blob/master/postgres.md#predicate-many-preceders-pmp

krishnangovindraj (commented):
One round of comments.

storage/tests/test_isolation.rs
```diff
@@ -122,3 +130,379 @@ fn g0_update_conflicts_fail() {
...
+
+    let snapshot_setup = storage.open_snapshot_write();
+    snapshot_setup.put_val(key_1.to_owned(), ByteArray::copy(value_0.bytes()));
+    let result_snapshot = snapshot_setup.commit();
+    assert!(result_snapshot.is_ok());
+
+    let snapshot_1 = storage.open_snapshot_write();
+    let snapshot_2 = storage.open_snapshot_write();
+    let read_1: ByteArray<128> = snapshot_1.get::<128>(key_1).unwrap();
+    let read_2: ByteArray<128> = snapshot_1.get::<128>(key_1).unwrap();
```
flyingsilverfin, line 279: Snapshot_2

storage/tests/test_isolation.rs
```diff
@@ -122,3 +130,379 @@ fn g0_update_conflicts_fail() {
...
+    let to_write_1 = ByteArray::inline([read_1.bytes()[0] + 1], 1);
+    let to_write_2 = ByteArray::inline([read_1.bytes()[0] + 1], 1);
```
flyingsilverfin, line 281: read_2

storage/tests/test_isolation.rs
```diff
@@ -122,3 +130,379 @@ fn g0_update_conflicts_fail() {
...
+    let read_13 = snapshot_3.get::<128>(key_1);
+    let read_23 = snapshot_3.get::<128>(key_2);
+    assert_eq!(value_10.bytes(), read_13.unwrap().bytes());
+    assert_eq!(to_write_22.bytes(), read_23.unwrap().bytes());
+
+    let result_3 = snapshot_3.commit();
+
+    snapshot_1.put_val(key_1.to_owned(), ByteArray::copy(value_13.bytes()));
+    let result_1 = snapshot_1.commit();
+    assert!(false); // Error if it's allowed to commit?
```
flyingsilverfin, line 408: Check @flyingsilverfin

storage/tests/test_isolation.rs
```diff
@@ -122,3 +130,379 @@ fn g0_update_conflicts_fail() {
...
+
+    assert!(result_1.is_ok() && result_2.is_ok());
+    let reader_after = storage.open_snapshot_read();
+    let sum: u8 = reader_after.get::<128>(key_1).unwrap().bytes()[0] + reader_after.get::<128>(key_2).unwrap().bytes()[0];
+    assert!(sum <= 1);
+}
+
+#[test]
+fn g2_antidependency_cycles_fekete() {
+    // There's also a predicate version of g2, but I don't know if the iterators work yet
```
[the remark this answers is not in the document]
krishnangovindraj: I can add that, but it's still not as general as a predicate

storage/tests/test_isolation.rs
```diff
@@ -122,3 +130,379 @@ fn g0_update_conflicts_fail() {
...
+    let result_2 = snapshot_2.commit();
+
+    assert!(result_1.is_ok() && result_2.is_ok());
+    let reader_after = storage.open_snapshot_read();
+    let sum: u8 = reader_after.get::<128>(key_1).unwrap().bytes()[0] + reader_after.get::<128>(key_2).unwrap().bytes()[0];
+    assert!(sum <= 1);
+}
+
+#[test]
+fn g2_antidependency_cycles_fekete() {
```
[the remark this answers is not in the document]
krishnangovindraj: It's an error if it's allowed to commit, but should I verify the result of the commit instead of just asserting false at the end?

storage/tests/test_isolation.rs
```diff
@@ -123,3 +127,488 @@ fn g0_update_conflicts_fail() {
...
+    let snapshot_setup = storage.open_snapshot_write();
+    snapshot_setup.put_val(key_1.to_owned(), ByteArray::copy(value_1_0.bytes()));
+    let result_snapshot = snapshot_setup.commit();
+    assert!(result_snapshot.is_ok());
+
+    let snapshot_1 = storage.open_snapshot_write();
+    let snapshot_2 = storage.open_snapshot_write();
+    snapshot_1.put_val(key_1.to_owned(), ByteArray::copy(value_1_1i.bytes()));
+    let read_2 = snapshot_2.get::<BUFFER_VALUE_INLINE>(StorageKeyReference::from(&key_1));
+    snapshot_1.put_val(key_1.to_owned(), ByteArray::copy(value_1_1f.bytes()));
```
krishnangovindraj, line 225: This fails because overwriting a key that was written in the same transaction is not yet implemented.

storage/tests/test_isolation.rs
```diff
@@ -123,3 +127,488 @@ fn g0_update_conflicts_fail() {
...
+#[test]
+fn g2_antidependency_cycles_fekete() {
+    // From "A Read-Only Transaction Anomaly Under Snapshot Isolation" (Fekete et al.)
+    //  Without the read-only t3,  the history is: [ r_1(1), r_1(2), r_2(2), w_2(2, ?) ]
+    //      There is only one rw edge: r_1(2) -> w_2(2,?)
+    // The third transaction extends the history, [..., r_3(1), r_3(2), w_1(1, ?) ]
+    // opened after t2 commits introduces :
+    //  * a wr edge w_2(2,?) -> r_3(2)      (This is possible since t1 & t3 are not 'concurrent')
+    //  * a rw edge r_3(1) -> w_1(1,?)
+    // These form a cycle between t1, t2, t3.
```
krishnangovindraj, line 428: This is an example where the precedence graph cycle involves 3 transactions. This means we won't be able to detect this in a pairwise approach to conflict detection between transactions.

flyingsilverfin (approved):

PR #7000, opened 2024-03-14
WAL

dmitrii-ubskii:
## Usage and product changes

We implement the local sharded (16 MB shards) write-ahead log to be used as a durability service. The WAL is using lz4 compression within each written record, which also provides the integrity checking mechanism.

Expected process flow during commit:

1. write data into WAL;
2. write data into speedb;
3. flush speedb to disk;
4. checkpoint WAL so that subsequent recoveries begin from the records after this successful commit.

## Implementation

- crash test: start a process that continuously writes records into the WAL, kill that process, then verify that the records can be read back without being corrupted.
- write benchmark: measures pre-serialization throughput of writing data into the WAL.

### Architecture decisions

The WAL has several goals:
1. It needs to act as a sequencer, which assigns a sequence number to each record that it receives
2. Performantly serialises data to the file on disk, guaranteeing durability of each record by flushing after each write
3. Allows retrieving old entries in the WAL with reasonable speed
4. Allow iterating through old entries of the WAL from a specific sequence number
5. Is able to handle heterogeneous record types
6. Includes compression to minimise disk bandwidth usage.

We decided to use a sharded WAL with relatively small files to allow period cleanup of old WAL files without causing rewrites or extensive locking. These files can also be used to effectively search for the location of an older sequence number, or even search by time.

### DIFF
--- a/storage/durability/durability.rs
+++ b/storage/durability/durability.rs
@@ -85,6 +85,7 @@ pub trait DurabilityService: Sequencer {
 
     fn iter_from(&self, sequence_number: SequenceNumber) -> impl Iterator<Item = io::Result<RawRecord>>;
 
+    fn checkpoint(&self) -> Result<()>;
     fn recover(&self) -> impl Iterator<Item = io::Result<RawRecord>>;
 }
 
--- a/storage/durability/durability.rs
+++ b/storage/durability/durability.rs
@@ -33,36 +33,36 @@ pub mod wal;
 
 pub type Result<T> = std::result::Result<T, DurabilityError>;
 
-///
-/// Notes:
-///     We can provide to a DurabilityService for Records
-///     Records have a serialised Byte representations and type (ID)
-///
-///     We have two choices:
-///         1. Long files (~ 100s MB) with many records embedded into fixed-sized blocks (fixed size = seekable)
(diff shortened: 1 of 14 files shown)


storage/durability/durability.rs
```diff
@@ -15,62 +15,56 @@
...
 pub trait DurabilityService: Sequencer {
-    fn register_record_type(&mut self, record_type: DurabilityRecordType, record_name: &'static str);
+    fn register_record_type<Record: DurabilityRecord>(&mut self);
+
+    fn sequenced_write<Record>(&self, record: &Record) -> Result<SequenceNumber>
+    where
+        Record: DurabilityRecord;
 
-    fn sequenced_write(&self, record: &impl DurabilityRecord, record_name: &'static str) -> Result<SequenceNumber, DurabilityError>;
+    fn iter_from(&self, sequence_number: SequenceNumber) -> io::Result<impl Iterator<Item = io::Result<RawRecord>>>;
```
flyingsilverfin, line 55: let's add `iter_type_from` which takes a `DurabilityRecord` type and only returns deserialised records that have the right type, it shoudl be an iterator of `Result<(SequenceNumber, DurabilityRecord)>`
, where result is only for actual errors.

flyingsilverfin (approved):

PR #7012, opened 2024-03-21
Introduce atomic sequence number

krishnangovindraj:
## Usage and product changes
Introduce an atomic sequence number type providing atomic operations to increment epoch & increment sequence number

## Implementation
```
pub struct AtomicSequenceNumber {
    epoch: Cell<u64>,
    seq_sync: AtomicU64,
    epoch_increment_mutex: Mutex<()>
}
```

The lower 32 bits of `seq_sync`  hold the sequence number within the epoch, and the upper 32 bits of `seq_sync` are the 'sync' bits for the lowest 32 bits of the epoch. These overlapping sync bits are used to detect races between reads & epoch updates.
`epoch_increment_mutex` is used to guarantee single writers incrementing the epoch.


common/primitive/atomic_u128.rs
```diff
@@ -0,0 +1,168 @@
...
+        let uw_increment = if Self::uw_outdated(uw_read, lw_read) { UW_INCREMENT } else { 0 };
+        Self::compose(uw_read + uw_increment, lw_read)
+    }
+
+    pub fn add_and_get(&mut self, increment: u32) -> u128 {
+        let uw_read = self.uw_sync;
+        let lw_before_add = self.lw.fetch_add(increment as u64, SeqCst);
+        let lw_after_add = u64::wrapping_add(lw_before_add, increment as u64);
+        let uw_increment = if Self::uw_outdated(uw_read, lw_after_add) { UW_INCREMENT } else { 0 };
+        let updated_sync_bits: bool = (lw_after_add & SYNC_MASK) != (lw_before_add & SYNC_MASK);
```
krishnangovindraj, line 71: rename `updated_sync_bits` ->  `sync_bits_were_updated`

common/primitive/atomic_u128.rs
```diff
@@ -0,0 +1,168 @@
...
+        let lw_read  = self.lw.load(SeqCst);
+        let uw_increment = if Self::uw_outdated(uw_read, lw_read) { UW_INCREMENT } else { 0 };
+        Self::compose(uw_read + uw_increment, lw_read)
+    }
+
+    pub fn add_and_get(&mut self, increment: u32) -> u128 {
+        let uw_read = self.uw_sync;
+        let lw_before_add = self.lw.fetch_add(increment as u64, SeqCst);
+        let lw_after_add = u64::wrapping_add(lw_before_add, increment as u64);
+        let uw_increment = if Self::uw_outdated(uw_read, lw_after_add) { UW_INCREMENT } else { 0 };
```
krishnangovindraj, line 70: Should I re-order these branches? the 0 case is far more likely.

krishnangovindraj (commented):
More self-review comments

common/primitive/atomic_u128.rs
```diff
@@ -0,0 +1,168 @@
...
+/// It is the responsibility of the thread that updates the sync_bits (and possibly overflow) to update sync_bits in the u128.
+/// Other threads can return (UW+1)(LW) if sync_bits(LW) < sync_bits(UW);  (UW)(LW) otherwise
+pub struct AtomicU128 {
+    uw_sync: u128,
+    lw: AtomicU64,
+}
+
+const SYNC_MASK: u64 = 0xff00000000000000;
+const UW_MASK: u128 = 0xffffffffffffffff_0000000000000000;
+const UW_INCREMENT: u128 = 0x0000000000000001_0000000000000000;
```
dmitrii-ubskii, line 38: I'd suggest breaking up 64- and 128-bit literals further, so it's obvious the length is correct.
```
0xff00_0000_0000_0000
0xffff_ffff_ffff_ffff__0000_0000_0000_0000
```

common/primitive/atomic_u128.rs
```diff
@@ -0,0 +1,168 @@
...
+
+impl AtomicU128 {
+
+    pub fn new(from: u128) -> AtomicU128 {
+        AtomicU128 {
```
dmitrii-ubskii, line 43: We should be using `Self`, especially in constructors.

common/primitive/atomic_u128.rs
dmitrii-ubskii: Ordering of the branches doesn't affect the likeliness of the branching, and in this specific case we are probably going to end up with a conditional move (`cmov`) anyway.

common/primitive/atomic_u128.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    #[test]
+    fn test_add_and_get() {
+        let from: u128 = 0x0123456789abcdef__ffffffff_fedbca98;
+        let increment: u32 = 0x11111111;
+        let expected = from + increment as u128;
+        let mut t = AtomicU128::new(from);
+        let sum = t.add_and_get(increment);
+        // println!("exp:{:#18x}_{:#18x}\nact:{:#18x}_{:#18x}", (expected >> 64), expected as u64, (sum >> 64) , sum as u64);
```
dmitrii-ubskii, line 130: Should not be merged.

common/primitive/atomic_u128.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    pub fn get(&self) -> u128 {
+        let uw_read = self.uw_sync;
+        let lw_read  = self.lw.load(SeqCst);
+        let uw_increment = if Self::uw_outdated(uw_read, lw_read) { UW_INCREMENT } else { 0 };
+        Self::compose(uw_read + uw_increment, lw_read)
+    }
+
+    pub fn add_and_get(&mut self, increment: u32) -> u128 {
```
dmitrii-ubskii, line 66: We should loosely follow the `std` atomic naming. This would be called `add_fetch`.

common/primitive/atomic_u128.rs
```diff
@@ -0,0 +1,168 @@
...
+            Err(self.get())
+        }
+    }
+}
+
+pub mod tests {
+    use crate::atomic_u128::{AtomicU128, UW_INCREMENT};
+
+    #[test]
+    fn test_create_readback() {
```
dmitrii-ubskii, line 117: We don't need to prefix test functions with `test_`. They are in a module called `test` and marked with `#[test]`.

common/primitive/atomic_u128.rs
```diff
@@ -0,0 +1,168 @@
...
+                }
+            }
+        } else {
+            // If uw don't match, attempting a swap would write the wrong state to lw
+            Err(self.get())
+        }
+    }
+}
+
+pub mod tests {
```
dmitrii-ubskii, line 113:
```suggestion
#[cfg(test)]
mod test {
```
Only needs to be compiled when testing, and doesn't need to be public.

common/primitive/atomic_u128.rs
```diff
@@ -0,0 +1,168 @@
...
+    use crate::atomic_u128::{AtomicU128, UW_INCREMENT};
```
dmitrii-ubskii, line 114: `use super::{...}` makes it a little clearer that the imports are from the tested unit.

common/primitive/atomic_u128.rs
dmitrii-ubskii, line 66: `&mut` is an exclusive reference, meaning that this method and the `compare_exchange` can't be used from different threads. Atomic methods must take `&self`.

common/primitive/atomic_u128.rs
```diff
@@ -0,0 +1,168 @@
...
+
+        if updated_sync_bits {
+            self.uw_sync = ((uw_read & UW_MASK) + uw_increment) | (lw_after_add as u128);
+            Self::compose(self.uw_sync, lw_after_add)
+        } else {
+            Self::compose(uw_read + uw_increment, lw_after_add)
+        }
+    }
+
+    pub fn compare_and_exchange_incremented(&mut self, current : u128, increment: u32) -> Result<u128, u128> {
```
dmitrii-ubskii, line 81: `compare_fetch_add`? This is just like `fetch_add`, but can be rejected, right?

common/primitive/atomic_u128.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    pub fn add_and_get(&mut self, increment: u32) -> u128 {
+        let uw_read = self.uw_sync;
+        let lw_before_add = self.lw.fetch_add(increment as u64, SeqCst);
+        let lw_after_add = u64::wrapping_add(lw_before_add, increment as u64);
+        let uw_increment = if Self::uw_outdated(uw_read, lw_after_add) { UW_INCREMENT } else { 0 };
+        let updated_sync_bits: bool = (lw_after_add & SYNC_MASK) != (lw_before_add & SYNC_MASK);
+
+        if updated_sync_bits {
+            self.uw_sync = ((uw_read & UW_MASK) + uw_increment) | (lw_after_add as u128);
```
dmitrii-ubskii, line 74: We're spending a lot of time between `lw` and `uw` being updated. I would like to see more tests covering the potential data races in a multithreaded scenario. A comment covering possible orderings and outcomes would be nice too.

dmitrii-ubskii (changes requested):
This is currently not usable as an atomic.

common/primitive/atomic_u128.rs
krishnangovindraj: Ah, thanks for pointing that out. I've replaced it with a `Cell` around uw_sync

common/primitive/atomic_u128.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    pub fn load(&self) -> u128 {
+        let uw_read = self.uw_sync.get();
+        let lw_read  = self.lw.load(SeqCst);
+        let uw_increment = if Self::uw_outdated(uw_read, lw_read) { UW_INCREMENT } else { 0 };
+        Self::compose(uw_read + uw_increment, lw_read)
+    }
+
+    pub fn add_fetch(&self, increment: u32) -> u128 {
+        let uw_read = self.uw_sync.get();
```
krishnangovindraj, line 68: @dmitrii-ubskii, Is reading a u128 atomic? I think I assumed it would be, but I don't see why that would be the case. If so, I'll have to revise the whole thing to use a u64.

krishnangovindraj (commented):
Flag potential non-atomic u128 read problem

storage/durability/sequence_number.rs
```diff
@@ -0,0 +1,208 @@
...
+impl From<u128> for SequenceNumber {
+    fn from(value: u128) -> Self {
+        Self::new(U80::new(value))
+    }
+}
+
+pub struct AtomicSequenceNumber {
+    epoch: Cell<u64>,
+    seq_sync: AtomicU64,
+    epoch_increment_mutex: Mutex<()>
```
flyingsilverfin, line 85: Do you ened both `Cell` and `Mutex`? Can't you do `Mutex<u64>` or something directly? Or even `Cell<Mutex<u64>>`? 
I haven't read through the details but just noticed this from reading this set of types.

storage/durability/sequence_number.rs
```diff
@@ -0,0 +1,208 @@
...
+
+impl From<u128> for SequenceNumber {
+    fn from(value: u128) -> Self {
+        Self::new(U80::new(value))
+    }
+}
+
+pub struct AtomicSequenceNumber {
+    epoch: Cell<u64>,
+    seq_sync: AtomicU64,
```
flyingsilverfin, line 84: Why suffix with `sync`? Can it just be `seq`?

storage/durability/sequence_number.rs
```diff
@@ -0,0 +1,208 @@
...
+    epoch_increment_mutex: Mutex<()>
+}
+
+impl AtomicSequenceNumber {
+    const SEQ_NUMBER_BITS: u8 = 32;
+    const SYNC_BITS: u8 = (64 - Self::SEQ_NUMBER_BITS);
+
+    const UW_SYNC_MASK: u64 = (1 << Self::SYNC_BITS) - 1;
```
flyingsilverfin, line 92: Can we use fewer abbreviations `UW/LW` since these aren't particularly universal? (just judging since I don't know them!) Upper word? Lower word? Technically 'word' isn't always a u64... we could just use 'upper' and 'lower' throughout if it is that... or LOWER_BYTES (longer but clearer)

storage/durability/sequence_number.rs
```diff
@@ -0,0 +1,208 @@
...
+    fn epoch_overflow_oudated(epoch: u64, seq_sync: u64) -> bool {
+        let seq_sync_bits = seq_sync >> Self::SEQ_NUMBER_BITS;
+        let epoch_sync_bits = epoch & Self::UW_SYNC_MASK;
+        seq_sync_bits < epoch_sync_bits
+    }
+
+    #[inline]
+    fn compose(epoch: u64, seq_sync: u64) -> SequenceNumber {
+        SequenceNumber::from((((epoch & !Self::UW_SYNC_MASK) as u128) << Self::SEQ_NUMBER_BITS) | seq_sync as u128)
+    }
```
flyingsilverfin, line 105: In our style, `new` and other constructors always come first

storage/durability/sequence_number.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    pub fn increment_epoch_fetch(&self) -> SequenceNumber {
+        let lock = self.epoch_increment_mutex.lock().unwrap();
+        let next_epoch = self.epoch.get() + 1;
+        let seq_sync_updated = next_epoch << Self::SEQ_NUMBER_BITS; // & Self::SEQ_SYNC_MASK; is not needed
+        self.seq_sync.store(seq_sync_updated, SeqCst);
+        self.epoch.set(next_epoch);
+        Self::compose(next_epoch, seq_sync_updated)
+    }
```
flyingsilverfin, line 139: Not going to check this, but I hope it's all correct!! Maybe add a test with that spins up a bunch of threads (you can use scoped threads) that race on updating a sequence number in a loop and make sure you end up with the right sequence number at the end?

storage/durability/sequence_number.rs
krishnangovindraj: Putting the `u64` in the Mutex would make readers lock each other out. What I need is just a single-writer guarantee. I could do an RwLock, but I don't need locking on read since it should be atomic anyway (provided it's aligned? I could make it atomic to be sure.)

storage/durability/sequence_number.rs
krishnangovindraj: This isn't a constructor as much as it's a macro for putting together the two parts of our atomic sequence number. It's meant to be private as well.

flyingsilverfin:
Not going to be used for the foreseeable future, so we're closing

PR #7015, opened 2024-03-28
Implement storage error handling and propagation up to concept layer

dmitrii-ubskii:
## Usage and product changes

We introduce new error types, moving towards the new preferred error strategy:

- `SnapshotError` is partially dissolved into `SnapshotIteratorError` and `SnapshotGetError`. `SnapshotPutError` is removed.
- All fallible interactions with the storage from the concept layer now wrap the error (in future with additional data) into one of `ConceptReadError`, `ConceptWriteError` (treated as a superset of the former), or `ConceptIteratorError`.

### DIFF
--- a/storage/isolation_manager.rs
+++ b/storage/isolation_manager.rs
@@ -21,7 +21,7 @@ use std::{
     fmt,
     io::Read,
     sync::{
-        atomic::{AtomicBool, AtomicU64, AtomicU8, AtomicUsize, Ordering},
+        atomic::{AtomicU64, AtomicU8, AtomicUsize, Ordering},
         Arc, OnceLock, RwLock,
     },
 };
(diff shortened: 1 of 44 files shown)


storage/isolation_manager.rs
```diff
@@ -142,19 +130,18 @@ impl IsolationManager {
...
-            match status {
-                CommitStatus::Empty => unreachable!("Illegal state - commit status cannot move from pending"),
+            match predecessor_window.get_status(predecessor_sequence_number) {
+                CommitStatus::Empty => unreachable!("Illegal state - commit status cannot move from pending to empty"),
                 CommitStatus::Pending(_) => {
                     // TODO: we can improve the spin lock with async/await
                     // Note we only expect to have long waits in long chains of overlapping transactions that would conflict
                     // could also do a little sleep in the spin lock, for example if the validating is still far away
-                    continue;
+                    std::hint::spin_loop();
```
dmitrii-ubskii, line 144: This is better than just going to the start of the loop, as it may emit an instructions that signals to the CPU that this is a spin loop.

storage/isolation_manager.rs
```diff
@@ -610,32 +606,3 @@ impl DurabilityRecord for CommitRecord {
...
-#[derive(Debug)]
-pub struct IsolationError {
-    pub kind: IsolationErrorKind,
-}
-
-#[derive(Debug)]
-pub enum IsolationErrorKind {
-    DeleteRequiredViolation,
-    RequiredDeleteViolation,
-}
```
dmitrii-ubskii, line 623: Moved closer to usage and refactored

flyingsilverfin (approved):

PR #7017, opened 2024-04-01
Add basic MVCC tests

krishnangovindraj:
## Usage and product changes
Adds basic tests ensuring we can open views into past versions of the database & that commits increment the (default) version seen by new transactions.

## Implementation
* Adds test cases to `test_mvcc.rs`


krishnangovindraj (commented):
Self review comments.

storage/tests/test_mvcc.rs
```diff
@@ -34,3 +34,137 @@ This file should comprise a set of low-level tests relating to MVCC.
...
+    snapshot_read_01.close_resources();
+    snapshot_read_0.close_resources();
+
+    // Read from further in the past.
+    let snapshot_read_02 = storage.open_snapshot_read_at(watermark_0);
+    assert_eq!(snapshot_read_02.get::<128>(key_1.as_reference()).unwrap().unwrap().bytes(), VALUE_0);
+    snapshot_read_02.close_resources();
+}
+
+#[test]
```
krishnangovindraj, line 137: I'm not sure this test is meaningful in our context given

1. We may not need to fail transactions which concurrently write the same key.
2. We have isolation tests to test isolation is enforced.

storage/tests/test_mvcc.rs
```diff
@@ -34,3 +34,137 @@ This file should comprise a set of low-level tests relating to MVCC.
...
+fn test_commit_increments_watermark() {
+    init_logging();
+    let storage_path = create_tmp_dir();
+    let storage = setup_storage(&storage_path);
+    let wm_initial = storage.read_watermark();
+    let snapshot_0 = storage.open_snapshot_write();
+    snapshot_0.commit().unwrap();
+
+    assert_eq!(wm_initial.next().number().number() + 1, storage.read_watermark().number().number());
```
krishnangovindraj, line 94: Currently fails.

krishnangovindraj (commented):
New self review after moving everything back to test_mvcc.rs

storage/tests/test_mvcc.rs
```diff
@@ -34,3 +34,137 @@ This file should comprise a set of low-level tests relating to MVCC.
    After cleanup is run, if we iterate directly on the storage layer, we should be able to confirm the keys are actually not present anymore (Rocks may defer the disk delete till compaction, but to us they are "gone").
 
  */
+
```
krishnangovindraj, line 37: Comment #4 above is still pending.
If we're happy with the current case, I'll move it to a TODO comment.

storage/tests/test_mvcc.rs
flyingsilverfin: Hang on, just from reading the title, in our semantics writing the same key twice should not fail right?

flyingsilverfin (approved):

PR #7018, opened 2024-04-01
Isolation manager timeline optimisation

krishnangovindraj:
## Usage and product changes
Refactors how the isolation manager tracks transactions & implements cleanup of transactions which are no longer needed. 

## Implementation
* Re-implements the isolation manager timeline to work on relative_indices (which are guaranteed to be contiguous) instead of sequence numbers (which will be discontiguous at least in the distributed case)
* Implements cleaning up of commits which are no longer needed to validate active transactions.
* Add path for reading commit data straight from storage. 
* Add tests when reading from storage.

### DIFF
--- a/storage/durability/wal.rs
+++ b/storage/durability/wal.rs
@@ -46,8 +46,8 @@ impl WAL {
         let files = RwLock::new(files);
         let next = RecordIterator::new(files.read().unwrap(), SequenceNumber::MIN)?
             .last()
-            .map(|rr| rr.unwrap().sequence_number.number().number() as u64 + 1)
-            .unwrap_or(1);
+            .map(|rr| rr.unwrap().sequence_number.next().number().number() as u64)
+            .unwrap_or(SequenceNumber::MIN.next().number.number() as u64);
 
         Ok(Self { registered_types: HashMap::new(), next_sequence_number: AtomicU64::new(next), files })
     }
@@ -125,7 +125,7 @@ impl Files {
         files.sort_unstable_by(|lhs, rhs| lhs.path.cmp(&rhs.path));
 
         if files.is_empty() {
-            files.push(File::open_at(directory.clone(), SequenceNumber::from(1))?);
+            files.push(File::open_at(directory.clone(), SequenceNumber::from(SequenceNumber::MIN.next()))?);
         }
 
         let writer = files.last().unwrap().writer()?;
--- a/storage/durability/wal.rs
+++ b/storage/durability/wal.rs
@@ -79,6 +79,16 @@ impl DurabilityService for WAL {
         self.registered_types.insert(Record::RECORD_TYPE, Record::RECORD_NAME);
     }
 
+    fn unsequenced_write<Record>(&self, record: &Record) -> Result<()>
+    where
+        Record: DurabilityRecord,
+    {
+        debug_assert!(self.registered_types.get(&Record::RECORD_TYPE) == Some(&Record::RECORD_NAME));
+        let mut files = self.files.write().unwrap();
+        files.write_record(record, SequenceNumber::from(0))?;
+        Ok(())
+    }
+
     fn sequenced_write<Record>(&self, record: &Record) -> Result<SequenceNumber>
     where
         Record: DurabilityRecord,
--- a/storage/durability/wal.rs
+++ b/storage/durability/wal.rs
@@ -85,7 +85,7 @@ impl DurabilityService for WAL {
     {
         debug_assert!(self.registered_types.get(&Record::RECORD_TYPE) == Some(&Record::RECORD_NAME));
         let mut files = self.files.write().unwrap();
-        files.write_record(record, SequenceNumber::from(0))?;
+        files.write_record(record, self.previous())?;
         Ok(())
     }
 
@@ -221,6 +221,16 @@ impl FileReader {
         Ok(Self { reader: BufReader::new(StdFile::open(&file.path)?), file })
     }
 
+    fn peek_sequence_number(&mut self) -> io::Result<Option<SequenceNumber>> {
+        if self.reader.stream_position()? == self.file.len {
+            return Ok(None);
+        }
+        let mut buf = [0; U80::BYTES];
+        self.reader.read_exact(&mut buf)?;
+        self.reader.seek_relative(-1 * buf.len() as i64)?;
+        Ok(Some(SequenceNumber::new(U80::from_be_bytes(&buf))))
+    }
+
     fn read_one_record(&mut self) -> io::Result<Option<RawRecord>> {
         if self.reader.stream_position()? == self.file.len {
             return Ok(None);
@@ -264,24 +274,26 @@ impl<'a> RecordIterator<'a> {
     fn new(files: RwLockReadGuard<'a, Files>, start: SequenceNumber) -> io::Result<Self> {
         let (current, mut current_start) = files
             .iter()
-            .map_while(|file| (file.start <= start).then_some(file.start))
+            .map_while(|file| (file.start < start).then_some(file.start))
             .enumerate()
             .last()
             .unwrap_or((0, files.files[0].start));
         let mut reader = FileReader::new(files.files[current].clone())?;
+
         while current_start < start {
-            match reader.read_one_record().transpose() {
+            reader.read_one_record()?;
+            match reader.peek_sequence_number().transpose() {
                 None => {
-                    // sequence number is past the end
-                    return Ok(Self { files, current, reader: None });
+                    // sequence number is past the end of this file.
+                    break;
                 }
                 Some(Err(err)) => return Err(err),
-                Some(Ok(RawRecord { sequence_number, .. })) => {
-                    current_start = SequenceNumber::new(sequence_number.number() + 1)
+                Some(Ok(sequence_number)) => {
+                    current_start = sequence_number;
                 }
             }
         }
-        Ok(Self { files, current, reader: Some(reader) })
+        Ok( Self { files, current, reader: Some(reader) })
     }
 
     fn advance_file(&mut self) -> io::Result<Option<()>> {
@@ -320,7 +332,7 @@ mod test {
     use tempdir::TempDir;
 
     use super::WAL;
-    use crate::{DurabilityRecord, DurabilityRecordType, DurabilityService, RawRecord, SequenceNumber};
+    use crate::{DurabilityRecord, DurabilityRecordType, DurabilityService, RawRecord, SequencedDurabilityRecord, SequenceNumber};
 
     #[derive(Debug, PartialEq, Eq, Clone, Copy)]
     struct TestRecord {
@@ -342,6 +354,8 @@ mod test {
             Ok(Self { bytes })
         }
     }
+
+    impl SequencedDurabilityRecord for TestRecord {}
 
     fn open_wal(directory: &TempDir) -> WAL {
         let mut wal = WAL::recover(directory).unwrap();
--- a/storage/isolation_manager.rs
+++ b/storage/isolation_manager.rs
@@ -121,7 +121,7 @@ impl IsolationManager {
         let (windows, first_window_relative_index): (Vec<Arc<TimelineWindow<TIMELINE_WINDOW_SIZE>>>, i64) =
             self.timeline.collect_concurrent_windows(commit_record.open_sequence_number, commit_relative_index);
         let first_windowed_seq = *windows.get(0).unwrap().starting_sequence_number().unwrap();
-        if commit_record.open_sequence_number().next() < first_windowed_seq {
+        if commit_record.open_sequence_number().next() <= first_windowed_seq {
             self.validate_concurrent_from_disk(commit_record, first_windowed_seq, durability_service)?;
         }
 
@@ -143,15 +143,9 @@ impl IsolationManager {
         D: DurabilityService,
     {
         for commit_status in
-            Self::iterate_commit_status_from_disk(durability_service, commit_record.open_sequence_number.next()).unwrap()
+            Self::iterate_commit_status_from_disk(durability_service, commit_record.open_sequence_number.next(), stop_sequence_number).unwrap()
         {
             if let Ok(commit_status) = commit_status {
-                match commit_status {
-                    CommitStatus::Applied(predecessor_seq, _) | CommitStatus::Aborted(predecessor_seq) | CommitStatus::Pending(predecessor_seq, _) => {
-                        if predecessor_seq >= stop_sequence_number { break; }
-                    },
-                    CommitStatus::Empty | CommitStatus::Validated(_, _) => unreachable!(),
-                }
                 let commit_dependency = match commit_status {
                     CommitStatus::Applied(_, predecessor_record) => {
                         commit_record.compute_dependency(&predecessor_record)
@@ -178,7 +172,7 @@ impl IsolationManager {
         let mut window_index = 0;
         let mut slot_index = (0..TIMELINE_WINDOW_SIZE).find(|si| {
             match windows[0].get_sequence_number(*si) {
-                Some(seq) => seq.number() > commit_record.open_sequence_number.number(),
+                Some(seq) => seq > commit_record.open_sequence_number,
                 None => false
             }
         }).unwrap();
@@ -202,16 +196,17 @@ impl IsolationManager {
     pub(crate) fn iterate_commit_status_from_disk<'a, D>(
         durability_service: &'a D,
         start_sequence_number: SequenceNumber,
+        stop_sequence_number: SequenceNumber,
     ) -> durability::Result<impl Iterator<Item = durability::Result<CommitStatus<'a>>>>
         where
             D: DurabilityService,
     {
-        let mut statuses: HashMap<u128, bool> = HashMap::new();
-        for record in durability_service.iter_unsequenced_type_from::<StatusRecord>(SequenceNumber::from(start_sequence_number.clone().number().number())).unwrap() {
+        let mut statuses: HashMap<SequenceNumber, bool> = HashMap::new();
+        for record in durability_service.iter_unsequenced_type_from::<StatusRecord>(start_sequence_number.clone()).unwrap() {
             if let Ok(predecessor_record) = record {
                 // We can't stop early because status records may be out-of-order
                 statuses.insert(
-                    predecessor_record.commit_record_sequence_number().number().number(),
+                    predecessor_record.commit_record_sequence_number(),
                     predecessor_record.was_committed(),
                 );
             } else {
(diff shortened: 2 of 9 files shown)


storage/storage.rs
```diff
@@ -187,7 +187,7 @@ impl<D> MVCCStorage<D> {
         let name = name.as_ref();
         let (keyspaces, keyspaces_index) = recover_keyspaces::<KS>(name, &storage_dir)?;
 
-        let isolation_manager = IsolationManager::new(SequenceNumber::from(1));
+        let isolation_manager = IsolationManager::new(1);
```
flyingsilverfin, line 190: If it's now a relative index, can we make this 0 always?

storage/storage.rs
```diff
@@ -268,13 +269,14 @@ impl<D> MVCCStorage<D> {
         commit_record: CommitRecord,
     ) -> Result<(), MVCCStorageError> {
         // 2. validate commit isolation
-        self.isolation_manager.try_commit(commit_sequence_number, commit_record).map_err(|err| MVCCStorageError {
+        let relative_index : i64 = commit_sequence_number.number().number() as i64; // TODO: Determine relative_index
```
flyingsilverfin, line 272: This TODO is active right? The number.number isn't the relative index?

flyingsilverfin, line 272: I don't like 'relative index' - perhaps we can find a better unified name throughout?

krishnangovindraj (commented):
One round of self review with plenty of TODOs.

storage/isolation_manager.rs
```diff
@@ -68,90 +73,85 @@ impl IsolationManager {
...
+        window.insert_pending(window_index, commit_sequence_number.clone(), commit_record);
+        if let CommitStatus::Pending((_, commit_record)) = window.get_status(window_index) {
+            let validation_result = self.validate_all_concurrent(relative_index, commit_record);
+            if validation_result.is_ok() {
+                window.set_committed(window_index);
+            } else {
+                window.set_closed(window_index);
+            }
+            self.timeline.remove_reader(&commit_record.open_sequence_number);
+            self.timeline.may_increment_watermark(relative_index);
```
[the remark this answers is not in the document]
krishnangovindraj: Fixed. may_increment_watermark is now done either when validation fails, or when `notify_applied`.

storage/storage.rs
```diff
@@ -205,11 +207,13 @@ impl<D> MVCCStorage<D> {
     {
         for record in self.durability_service.iter_type_from_start::<CommitRecord>().unwrap() {
             if let Ok((commit_sequence_number, commit_record)) = record {
+                // TODO: Do we need to record readers during the recovery? try_commit currently decrements reader count.
```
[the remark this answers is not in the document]
krishnangovindraj: We have a separate recovery path for finalised records. So any record that needs validation does increment the reader.

storage/storage.rs
```diff
@@ -300,6 +320,13 @@ impl<D> MVCCStorage<D> {
                 })?;
             }
         }
+        // 4. Inform the isolation manager and increment the watermark.
+        self.isolation_manager.applied(relative_index);
+
+        // // 5. Persist the commit status
+        // TODO: Ideally, we'd persist the commit status here, but what happens if the persist fails?
+        //  Until we answer that, persisting the commit status is done right during validation.
```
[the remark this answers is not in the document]
krishnangovindraj: This is now back here. We should crash the server if writing the status record fails, but now it's just a `todo!()`

storage/storage.rs
```diff
@@ -203,13 +206,31 @@ impl<D> MVCCStorage<D> {
...
+                    CommitStatus::Applied(commit_sequence_number, commit_record) => {
+                        self.isolation_manager.load_applied(commit_sequence_number.number().number() as i64, commit_sequence_number, commit_record.into_owned());
+                    },
+                    CommitStatus::Closed(commit_sequence_number) => {
+                        self.isolation_manager.load_aborted(commit_sequence_number.number().number() as i64, commit_sequence_number);
+                    }
+                    CommitStatus::Pending(commit_sequence_number, commit_record) => {
+                        self.isolation_manager.opened(commit_record.open_sequence_number()); // try_commit currently decrements reader count.
+                        self.write_commit_record(commit_sequence_number, commit_record.into_owned())?;
+                    }
```
krishnangovindraj, line 226: TODO: Using the sequence number as relative_index fails as soon as relative_index =0 does not hold Sequence number 0.

krishnangovindraj (commented):
Plenty of rebasing & refactoring. More self review comments.

storage/storage.rs
```diff
@@ -206,18 +215,45 @@ impl<D> MVCCStorage<D> {
...
+                                commit_sequence_number,
+                            );
+                        }
+                        CommitStatus::Pending(commit_sequence_number, commit_record) => {
+                            self.isolation_manager.opened(commit_record.open_sequence_number()); // try_commit currently decrements reader count.
+                            self.try_write_commit_record(
+                                Self::tmp_relative_index_from_sequence_number(commit_sequence_number),
+                                commit_sequence_number,
+                                commit_record.into_owned(),
+                            ).unwrap(); // TODO: This was an unwrap in the trunk.
```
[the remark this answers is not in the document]
krishnangovindraj: This has since been updated to pass if it's a validation error .

storage/durability/wal.rs
```diff
@@ -90,6 +90,16 @@ impl DurabilityService for WAL {
         self.registered_types.insert(Record::RECORD_TYPE, Record::RECORD_NAME);
     }
 
+    fn unsequenced_write<Record>(&self, record: &Record) -> Result<()>
```
flyingsilverfin, line 93: All WAL writes _should_ be sequenced - we can't break this API.
Epoch-based sequencing is going to be something our WAL does by default, and isn't a behaviour we want for _some_ records - this is bleeding higher-level usage information down into the implementation of the components we use.

storage/isolation_manager.rs
```diff
@@ -40,130 +43,256 @@ pub(crate) struct IsolationManager {
...
+    }
+
+    pub(crate) fn iterate_commit_status_from_disk<'a, D>(
+        durability_service: &'a D,
+        start_sequence_number: SequenceNumber,
+    ) -> durability::Result<impl Iterator<Item = durability::Result<CommitStatus<'a>>>>
+    where
+        D: DurabilityService,
+    {
+        let mut statuses: HashMap<u128, bool> = HashMap::new();
```
flyingsilverfin, line 240: Can't we just make `SequenceNumber` have `Hash`? this can map from sequence number directly

storage/isolation_manager.rs
```diff
@@ -40,130 +43,256 @@ pub(crate) struct IsolationManager {
...
+                statuses.insert(
+                    predecessor_record.commit_record_sequence_number().number().number(),
+                    predecessor_record.was_committed(),
+                );
+            } else {
+                todo!()
+            }
+        }
+
+        let map_fn = move |result: durability::Result<(SequenceNumber, CommitRecord)>| match result {
```
flyingsilverfin, line 253: first time we've had to use inline functions, this feels like a smell - can you find a better way to do this/extract it out into a method in the impl block? Probably name it something related to this method so it's clear.

storage/isolation_manager.rs
```diff
@@ -241,297 +358,317 @@ struct Timeline {
...
-impl Timeline {
-    fn new(starting_sequence_number: SequenceNumber) -> Timeline {
-        debug_assert!(starting_sequence_number.number() > 0);
-        let last_sequence_number_value = starting_sequence_number.number() - U80::new(1);
-        let initial_window = Arc::new(TimelineWindow::new(SequenceNumber::new(last_sequence_number_value)));
-        let next_window_start = initial_window.end();
-        let mut windows = VecDeque::new();
-        windows.push_back(initial_window);
+// We can adjust the Window size to amortise the cost of the read-write locks to maintain the timeline
+pub const TIMELINE_WINDOW_SIZE: usize = 100;
```
flyingsilverfin, line 363: Can you move this into `constants.rs` please

storage/isolation_manager.rs
```diff
@@ -241,297 +358,317 @@ struct Timeline {
...
+            let start_of_second_window = *next_relative_index - ((windows.len() - 1) * TIMELINE_WINDOW_SIZE) as i64;
+            match windows.front() {
+                None => false,
+                Some(front) => front.get_readers() == 0 && watermark >= start_of_second_window,
+            }
+        };
+        if can_free_some {
+            let (next_relative_index, windows) = &mut *self.next_window_and_windows.write().unwrap_or_log();
+            let mut start_of_next_window = *next_relative_index - ((windows.len() - 1) * TIMELINE_WINDOW_SIZE) as i64;
+            while watermark >= start_of_next_window && windows.front().is_some() && windows.front().unwrap().get_readers() == 0
```
flyingsilverfin, line 393: this is a specific "clearable" condition right? Maybe this is worth its own function with a nice name, if possible/not introducing complexity

storage/isolation_manager.rs
```diff
@@ -606,3 +767,241 @@ impl DurabilityRecord for CommitRecord {
...
+        StatusRecord { commit_record_sequence_number: sequence_number, was_committed: committed }
+    }
+
+    pub(crate) fn was_committed(&self) -> bool {
+        self.was_committed
+    }
+
+    pub(crate) fn commit_record_sequence_number(&self) -> SequenceNumber {
+        self.commit_record_sequence_number
+    }
```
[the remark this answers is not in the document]
flyingsilverfin: i tend to make access methods, i really don't like the direct acccess

flyingsilverfin (changes requested):

storage/storage.rs
```diff
@@ -187,10 +188,18 @@ impl<D> MVCCStorage<D> {
         // FIXME proper error
         let mut durability_service = D::recover(path.join(Self::WAL_DIR_NAME)).expect("Could not create WAL directory");
         durability_service.register_record_type::<CommitRecord>();
+        durability_service.register_record_type::<StatusRecord>();
 
         let name = name.as_ref();
         let (keyspaces, keyspaces_index) = recover_keyspaces::<KS>(&storage_dir)?;
-        let isolation_manager = IsolationManager::new(SequenceNumber::from(1));
+
+        let isolation_manager = if durability_service.previous() == SequenceNumber::MIN {
```
flyingsilverfin, line 196: Let's add a new method called `is_empty()`, which uses the next sequence number internally and checks if it is the first allowed one. However, this leads to: in`WAL::open_from` which calls `.unwrap_or(1);` -- this should be the same constant (important constants like this should make us uncomfortable). Then it's all 'safe'.

storage/storage.rs
```diff
@@ -206,18 +215,50 @@ impl<D> MVCCStorage<D> {
         D: DurabilityService,
     {
         use StorageRecoverError::DurabilityServiceRead;
-        let records = self
-            .durability_service
-            .iter_type_from_start::<CommitRecord>()
+        let records =
+            IsolationManager::iterate_commit_status_from_disk(&self.durability_service, SequenceNumber::MIN)
```
flyingsilverfin, line 219: This method doesn't feel like it belongs to `IsolationManager`, since it's nothing to do with isolation. However, this also doesn't really fit in `storage` -- this is hinting maybe we're missing an abstraction.

I think if we think about a networked durability service, maybe it's time to introduce a `DurabilityManager` in `//storage`, which holds this method and maybe some other ones? If you make it, what falls into it naturally?

storage/storage.rs
```diff
@@ -187,10 +188,18 @@ impl<D> MVCCStorage<D> {
...
         let name = name.as_ref();
         let (keyspaces, keyspaces_index) = recover_keyspaces::<KS>(&storage_dir)?;
-        let isolation_manager = IsolationManager::new(SequenceNumber::from(1));
+
+        let isolation_manager = if durability_service.previous() == SequenceNumber::MIN {
+            let im = IsolationManager::new(Self::tmp_relative_index_from_sequence_number(SequenceNumber::MIN));
+            im.load_aborted(0, SequenceNumber::MIN); // Initialise the watermark to zero
+            im
+        } else {
+            IsolationManager::new(Self::tmp_relative_index_from_sequence_number(durability_service.current()))
```
flyingsilverfin, line 201: In this branch, we have to recover. In the other branch, we do not right? 
I would move the creation of storage into each `if` and only recover (`.reload()`) when we actually have records.

storage/storage.rs
```diff
@@ -206,18 +215,50 @@ impl<D> MVCCStorage<D> {
...
+                        // TODO: Relative indexes once sequence numbers become discontinuous
+                        CommitStatus::Applied(commit_sequence_number, commit_record) => {
+                            self.isolation_manager.load_applied(
+                                Self::tmp_relative_index_from_sequence_number(commit_sequence_number),
+                                commit_sequence_number,
+                                commit_record.into_owned(),
+                            );
+                        }
+                        CommitStatus::Closed(commit_sequence_number) => {
+                            self.isolation_manager.load_aborted(
```
flyingsilverfin, line 234: If you're going to switch to the 'aborted' terminology, you should do it consistently - `CommitStatus::Aborted`

storage/storage.rs
```diff
@@ -206,18 +215,50 @@ impl<D> MVCCStorage<D> {
...
         for record in records {
             match record {
-                Ok((commit_sequence_number, commit_record)) => {
-                    self.try_write_commit_record(commit_sequence_number, commit_record).unwrap();
-                }
+                Ok(commit_status) => {
+                    match commit_status {
+                        // TODO: Relative indexes once sequence numbers become discontinuous
+                        CommitStatus::Applied(commit_sequence_number, commit_record) => {
+                            self.isolation_manager.load_applied(
```
flyingsilverfin, line 227: Let's call these `set_applied` and `set_aborted` since they aren't loading anything within, they're just setting values provided?

storage/isolation_manager.rs
```diff
@@ -40,130 +43,256 @@ pub(crate) struct IsolationManager {
...
+        window.insert_pending(slot_index, sequence_number, commit_record);
+        window.set_applied(slot_index);
+        self.timeline.may_increment_watermark(relative_index);
+    }
 
-        validation_result
+    pub(crate) fn load_aborted(&self, relative_index: i64, sequence_number: SequenceNumber) {
+        let (window, slot_index) = self.timeline.get_or_create_window(relative_index);
+        window.insert_pending(slot_index, sequence_number, CommitRecord::new(KeyspaceBuffers::new(), sequence_number));
+        window.set_closed(slot_index);
```
flyingsilverfin, line 79: If we call this method `set_aborted` (see comments in `storage.rs`) we shouldn't need to insert pending and then set closed immediately right?
Also, the terminology sohuld be consistent - `closed` should become `aborted` here too right?

flyingsilverfin (changes requested):

storage/isolation_manager.rs
```diff
@@ -241,297 +358,317 @@ struct Timeline {
...
+        let (next_window, windows) = &*self.next_window_and_windows.read().unwrap_or_log();
+        let relative_index_of_window_0 = *next_window - (windows.len() * TIMELINE_WINDOW_SIZE) as i64;
+        assert!(end_relative_index >= relative_index_of_window_0);
+        let end_lies_in_window = (end_relative_index - relative_index_of_window_0) as usize / TIMELINE_WINDOW_SIZE;
+        let mut concurrent_windows: Vec<Arc<TimelineWindow<TIMELINE_WINDOW_SIZE>>> = Vec::new();
+
+        let window_after = (0..windows.len()).find(|i| {
+            let window_start_seq = windows.get(*i).unwrap().starting_sequence_number().unwrap().number();
+            window_start_seq > start_sequence_number.number()
+        });
```
flyingsilverfin, line 505: This scan would go away and just re-use the `try_get_window(relative index  +  1)` if we had a relative index right? What if we introduce a `IndexedSequenceNumber`, which is a pair of relative index and sequence number and always returned from the durability service on durable write (unless you don't provide a sequenced record/ do an unsequenced write)?

storage/isolation_manager.rs
```diff
@@ -241,297 +358,317 @@ struct Timeline {
...
+        (concurrent_windows, start_index_of_first_concurrent)
     }
 
-    fn get_window(&self, sequence_number: SequenceNumber) -> Arc<TimelineWindow<TIMELINE_WINDOW_SIZE>> {
-        self.try_get_window(sequence_number).unwrap()
-    }
+    fn try_get_window(&self, relative_index: i64) -> Option<(Arc<TimelineWindow<TIMELINE_WINDOW_SIZE>>, usize)> {
+        let (next_window, windows) = &*self.next_window_and_windows.read().unwrap_or_log();
+        let start_index: i64 = next_window - (windows.len() * TIMELINE_WINDOW_SIZE) as i64;
+        let must_be_in_window = (relative_index - start_index) as usize / TIMELINE_WINDOW_SIZE;
```
flyingsilverfin, line 527: This is not a boolean, this should be called something with index.

storage/isolation_manager.rs
```diff
@@ -241,297 +358,317 @@ struct Timeline {
...
 
-    fn try_get_window(&self, sequence_number: SequenceNumber) -> Option<Arc<TimelineWindow<TIMELINE_WINDOW_SIZE>>> {
-        let (_, windows) = &*self.next_window_and_windows.read().unwrap_or_log();
-        Self::find_window(sequence_number, windows)
+        if must_be_in_window >= 0 && must_be_in_window < windows.len() {
+            let in_window = windows.get(must_be_in_window)?;
```
flyingsilverfin, line 530: `in_window` sounds like a boolean

storage/isolation_manager.rs
```diff
@@ -241,297 +358,317 @@ struct Timeline {
...
 
-    fn find_window(
-        sequence_number: SequenceNumber,
-        windows: &VecDeque<Arc<TimelineWindow<TIMELINE_WINDOW_SIZE>>>,
-    ) -> Option<Arc<TimelineWindow<TIMELINE_WINDOW_SIZE>>> {
-        windows.iter().rev().find(|window| window.contains(sequence_number)).cloned()
+    fn get_or_create_window(&self, relative_index: i64) -> (Arc<TimelineWindow<TIMELINE_WINDOW_SIZE>>, usize) {
+        let create: bool = {
+            let (next_window, _) = &*self.next_window_and_windows.read().unwrap_or_log();
+            relative_index >= *next_window
```
flyingsilverfin, line 541: this boolean expression can be moved into the `if` so we don't have to look so far away for the logic

storage/isolation_manager.rs
```diff
@@ -241,297 +358,317 @@ struct Timeline {
...
+            debug_assert!(window.get_readers() >= 0);
+            let _readers_remaining = window.decrement_readers();
+            if _readers_remaining == 0 {
+                self.may_free_windows();
+            }
+        };
     }
 
-    fn remove_window_reader(
+    fn find_window_for_reader(
```
flyingsilverfin, line 476: I've just realised this is actually non-specialised to `reader` right? This is actually just a `try_get_window` by sequence number?
Note that in another place you're doing this operation is in `collect_concurrent_windows` right?

storage/isolation_manager.rs
```diff
@@ -40,130 +43,256 @@ pub(crate) struct IsolationManager {
...
+    fn validate_concurrent_from_disk<D>(
+        &self,
+        commit_record: &CommitRecord,
+        stop_sequence_number: SequenceNumber,
+        durability_service: &D,
+    ) -> Result<(), IsolationError>
+    where
+        D: DurabilityService,
+    {
+        for raw_record in
```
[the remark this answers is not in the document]
krishnangovindraj: Yes. That was a bug throughout the validation code.

storage/durability/wal.rs
krishnangovindraj: After an in person discussion, we decided to do unsequenced writes.

storage/isolation_manager.rs
krishnangovindraj: It can actually be inlined.

storage/storage.rs
krishnangovindraj: Might be worth discussing, because I do remember us wanting the normal path to be the recovery path.
Maybe it's better to insert an actual committed record when the wal is created.

storage/isolation_manager.rs
krishnangovindraj: We need to insert pending to initialise the OnceLock and write the sequence number.

krishnangovindraj (commented):
Updated according to last round of comments; Replied to some.

storage/isolation_manager.rs
```diff
@@ -215,342 +346,319 @@ impl fmt::Display for IsolationError {
...
+                CommitStatus::Applied(_, _) => true,
+                CommitStatus::Aborted(_) => true,
+            };
+            if should_update
+                && self.watermark.compare_exchange(watermark, watermark + 1, Ordering::SeqCst, Ordering::SeqCst).is_ok()
+            {
+                watermark += 1;
+                slot_index += 1;
+                if slot_index >= TIMELINE_WINDOW_SIZE {
+                    if let Some(res) = self.try_get_window(watermark + 1) {
```
flyingsilverfin, line 418: If watermark and slot index move together, do we need to increment again?

flyingsilverfin: You can simplify this expression a bit:
```
if let Some((window, slot_index)) = self.try_get_window(...) {
  assert_eq(slot_index, 0);
} else {
  break;
}
```

krishnangovindraj: > If watermark and slot index move together, do we need to add to the watermark again?
`watermark` is the current watermark. slot_index is the slot index of the next candidate watermark  = `watermark+1`. 

> You can simplify this expression a bit:
`if let Some((window, slot_index))` seems to redeclare slot_index as a new variable, so this doesn't work.

PR #7036, opened 2024-04-10
Simplify isolation manager based on continuous sequence number guarantee

krishnangovindraj:
## Usage and product changes
We simplify the isolation manager implementation to operate on a continuous sequencing of the 'global order' of transactions, rather than a discontinuous distributed sequence number.

## Implementation
Updates SequenceNumber to be a `u64` and uses it in the isolation manager.

The rationale was as follows: The isolation manager lies within the 'Scheduler' component described in the Calvin paper. The scheduler is aware of the total ordering of the transaction within an 'epoch'. We decide the complexity of  translation between the sequence numbers generated by the distributed sequencer and a continuous enumeration generated by the schedulers is worth it to prevent the complexity of discontinuous sequence numbers creeping into the rest of the system.

The translation between the  distributed & global sequence numbers may require tracking some historic mapping between them.


storage/durability/durability.rs
```diff
@@ -121,57 +119,94 @@ pub trait UnsequencedDurabilityRecord: DurabilityRecord {}
...
 impl fmt::Display for SequenceNumber {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "SeqNr[{}]", self.number.number())
+        write!(f, "SeqNr[{}]", self.number)
     }
 }
 
 impl SequenceNumber {
-    pub const MIN: Self = Self { number: U80::MIN };
-    pub const MAX: Self = Self { number: U80::MAX };
```
krishnangovindraj, line 135: U80 can be deleted.

storage/isolation_manager.rs
```diff
@@ -42,64 +43,62 @@ impl fmt::Display for IsolationManager {
...
+        let (window, slot_index) = self.timeline.get_or_create_window(sequence_number);
+        window.insert_pending(slot_index, commit_record);
         window.set_applied(slot_index);
-        self.timeline.may_increment_watermark(relative_index);
+        self.timeline.may_increment_watermark(sequence_number);
     }
 
-    pub(crate) fn load_aborted(&self, relative_index: i64, sequence_number: SequenceNumber) {
-        let (window, slot_index) = self.timeline.get_or_create_window(relative_index);
-        window.insert_pending(slot_index, sequence_number, CommitRecord::new(KeyspaceBuffers::new(), sequence_number));
```
krishnangovindraj, line 78: We no longer need to set `insert_pending` for aborted transactions.

storage/isolation_manager.rs
```diff
@@ -431,133 +419,110 @@ impl Timeline {
...
+            Some( (offset / TIMELINE_WINDOW_SIZE, offset % TIMELINE_WINDOW_SIZE) ) // (window_index, slot_index)
+        } else {
+            None
+        }
+    }
+
+    fn start_of_nth_window(windows: &VecDeque<Arc<TimelineWindow<TIMELINE_WINDOW_SIZE>>> , next_window_sequence_number: SequenceNumber, n: usize) -> SequenceNumber {
+        next_window_sequence_number - (windows.len() - n) * TIMELINE_WINDOW_SIZE
+    }
+
```
krishnangovindraj: Helper functions to avoid inline arithmetic.

storage/isolation_manager.rs
```diff
@@ -431,133 +419,110 @@ impl Timeline {
...
-    /// Sequence number start of window (inclusive)
-    fn starting_sequence_number(&self) -> Option<&SequenceNumber> {
-        Some(&self.commit_records[0].get()?.0)
-    }
-
-    fn insert_pending(&self, index: usize, sequence_number: SequenceNumber, commit_record: CommitRecord) {
-        self.commit_records[index].set((sequence_number, commit_record)).unwrap_or_log();
+    fn insert_pending(&self, index: usize, commit_record: CommitRecord) {
+        let res = self.commit_records[index].set(commit_record);
+        if res.is_err(){ panic!("Failed at index: {index}"); }
```
krishnangovindraj: debug code: has to be reverted.

storage/durability/wal.rs
```diff
@@ -225,10 +218,10 @@ impl FileReader {
         if self.reader.stream_position()? == self.file.len {
             return Ok(None);
         }
-        let mut buf = [0; U80::BYTES];
+        let mut buf = [0; mem::size_of::<u64>()];
```
krishnangovindraj, line 221: Should I just hardcode 8?

storage/storage.rs
```diff
@@ -182,13 +182,7 @@ impl<D> MVCCStorage<D> {
...
         let (keyspaces, keyspaces_index) = recover_keyspaces::<KS>(&storage_dir)?;
 
-        let isolation_manager = if durability_service.is_empty() {
-            let im = IsolationManager::new(Self::todo_relative_index_from_sequence_number(SequenceNumber::MIN));
-            im.load_aborted(0, SequenceNumber::MIN); // Initialise the watermark to zero
-            im
-        } else {
-            IsolationManager::new(Self::todo_relative_index_from_sequence_number(durability_service.current()))
-        };
+        let isolation_manager = IsolationManager::new(durability_service.current());
```
krishnangovindraj, line 185: Thanks to the timeline now storing the actual sequence number as watermark.

storage/isolation_manager.rs
```diff
@@ -243,21 +237,16 @@ impl IsolationManager {
     }
 
     pub(crate) fn watermark(&self) -> SequenceNumber {
-        let (window, slot_index) = self.timeline.try_get_window(self.timeline.watermark()).unwrap();
-        match window.get_status(slot_index) {
-            CommitStatus::Applied(seq, _) | CommitStatus::Aborted(seq) => seq.clone(),
-            CommitStatus::Validated(_, _) | CommitStatus::Pending(_, _) | CommitStatus::Empty =>
-                unreachable!("The isolation manager must always have the watermark commit record in the window, and it must be either committed or aborted."),
-        }
+        self.timeline.watermark()
```
krishnangovindraj, line 239: Since there's no longer a mapping to be done between sequence number & relative index, we don't require an entry in the timeline at the watermark. Instead, we just return the value stored in the watermark.

krishnangovindraj (commented):
self review comments

storage/durability/wal.rs
flyingsilverfin: This is ok

storage/isolation_manager.rs
```diff
@@ -360,61 +344,70 @@ impl Error for IsolationError {}
...
-        if watermark != relative_index - 1 {
+    fn may_increment_watermark(&self, sequence_number: SequenceNumber) {
+        if self.watermark() != sequence_number - 1 {
             return ();
         }
-        let (mut window, mut candidate_slot_index) = self.try_get_window(relative_index).unwrap();
-        let end_of_window: i64 = relative_index + (TIMELINE_WINDOW_SIZE - candidate_slot_index) as i64;
-        let mut candidate_watermark = relative_index;
+        let (mut window, mut candidate_slot_index) = self.try_get_window(sequence_number).unwrap();
+        let end_of_window: SequenceNumber = if candidate_slot_index == 0 {
```
flyingsilverfin, line 390: Let's only trigger freeing windows on removal of a reader and re-check:
get watermark
loop {
  get first window
  check readers is 0
  check window doesn't contain watermark
  if all 3: delete window
}

storage/isolation_manager.rs
```diff
@@ -360,61 +344,70 @@ impl Error for IsolationError {}
 #[derive(Debug)]
 struct Timeline {
     // We can adjust the Window size to amortise the cost of the read-write locks to maintain the timeline
-    next_relative_index_and_windows: RwLock<(i64, VecDeque<Arc<TimelineWindow<TIMELINE_WINDOW_SIZE>>>)>,
-    watermark: AtomicI64,
+    next_sequence_and_windows: RwLock<(SequenceNumber, VecDeque<Arc<TimelineWindow<TIMELINE_WINDOW_SIZE>>>)>,
```
flyingsilverfin, line 347: Let's move sequence umber back into Windows, and simplify all the related code `resolve_sequence_number` becomes `find_window(windows, sequence_number)` + `window.get_slot(sequence_number`, a more composable API.

flyingsilverfin: We also can just call this Windows and when we get the write lock to add more windows get the last window, and add a new one starting at LAST_WINDOW.sequence_number()+ WINDOW_SIZE

storage/isolation_manager.rs
krishnangovindraj: Can we keep it this way? I like this because the guarantee is: Unused windows are freed.
Not checking on incrementing the watermark also prevents free-as-you-go loading of a committed history.
So we are actually trading off robustness for simplicity.

flyingsilverfin: ok we can keep this but we should be able to get rid of `resolve` sequence number stuff - this was the key point of the next comment and it hasn't been implemented?

PR #7039, opened 2024-04-15
Replace shared references with reference counted pointers

dmitrii-ubskii:
## Usage and motivation

We replace shared references (`&`) and `Rc`s with `Arc`s to simplify multithreaded behaviour.

This decision comes from the fundamental idea that Databases are not going to have any lexical lifetime - they are purely controlled by runtime/user intent. As a result, we cannot use storage or databases with basic Rust lifetimes.

Additionally, this decision sets us up transactions management on shutdown or database delete: transactions that are operating will be guaranteed the database is open. We will wait for current operations to finish + prevent new work from entering when trying to close the database. This will prevent a class of errors from from TypeDB 2.0 which were caused by 'resource's closing during transaction operation. 

## Implementation

~~WIP: the server builds and storage integration tests pass, but nothing else.~~ Propagated to all tests so the whole codebase builds.


flyingsilverfin (approved):

flyingsilverfin (approved):

concept/type_/role_type.rs
```diff
@@ -96,26 +97,26 @@ impl<'a> RoleType<'a> {
...
     pub fn get_annotations<'m>(
         &self,
-        type_manager: &'m TypeManager<'_, impl ReadableSnapshot>,
+        type_manager: &'m TypeManager<impl ReadableSnapshot>,
     ) -> Result<MaybeOwns<'m, HashSet<RoleTypeAnnotation>>, ConceptReadError> {
         type_manager.get_role_type_annotations(self.clone().into_owned())
     }
 
-    pub fn set_annotation(&self, type_manager: &TypeManager<'_, impl WritableSnapshot>, annotation: RoleTypeAnnotation) {
+    pub fn set_annotation(
```
krishnangovindraj, line 105: @flyingsilverfin , I reverted this to `pub` from `pub(crate)` because a test wasn't compiling. There aren't usages outside test. Does this need to be `pub(crate)`?

PR #7040, opened 2024-04-15
Allow Type IDs of deleted types to be reused

krishnangovindraj:
## Usage and product changes
Changes how type IDs are allocated so Type IDs from deleted types can be reused and allocated to newly created types. 
If we delete a type, we do not reset the corresponding thing ID generator to 0 - because it's non-trivial and we don't have to since U64 will have plenty of free values. Hence, If a type was created with a re-used TypeID, its Object ID of its instances will not start from 0.

## Implementation
* Instead of just incrementing an AtomicU16, we iterate over the `zip(range_of_type_ids, type_ids_in_db)` ids and return the first "gap" ( `it.0 != it.1` )
* We store an offset to start the next iteration from, so that we do a maximum of two seeks for allocating multiple Type IDs.
* If we don't find a free TypeID from our offset, we do another scan from the beginning (offset  = 0)

### DIFF
--- a/encoding/graph/type_/vertex_generator.rs
+++ b/encoding/graph/type_/vertex_generator.rs
@@ -28,33 +28,31 @@ use crate::graph::type_::vertex::TypeIDUInt;
 use crate::graph::Typed;
 
 pub struct TypeIDAllocator {
-    root_type : Kind,
+    kind: Kind,
     last_allocated_type_id: AtomicU16,
-    to_vertex:  fn (TypeID) -> TypeVertex<'static>
+    vertex_constructor:  fn (TypeID) -> TypeVertex<'static>
 }
 
 impl TypeIDAllocator {
 
     fn new(root_type: Kind, to_vertex: fn (TypeID) -> TypeVertex<'static>) -> TypeIDAllocator {
-        Self { root_type, last_allocated_type_id : AtomicU16::new(0), to_vertex }
+        Self { kind: root_type, last_allocated_type_id : AtomicU16::new(0), vertex_constructor: to_vertex }
     }
 
-    fn to_key(&self, type_id: TypeIDUInt) -> StorageKey<'static, BUFFER_KEY_INLINE> {
-        (self.to_vertex)(TypeID::build(type_id)).into_storage_key()
-    }
-
-    fn to_type_id<'a>(&self, key : StorageKeyReference) -> TypeIDUInt {
-        TypeVertex::new(Bytes::Reference(key.byte_ref())).type_id_().as_u16()
-    }
-
-    fn iterate_and_find<Snapshot: WritableSnapshot>(&self, snapshot: &Snapshot, start: TypeIDUInt) -> Result<Option<TypeIDUInt>, EncodingError> {
-        let mut type_id_iter = snapshot.iterate_range(PrefixRange::new_inclusive(self.to_key(start), self.to_key(TypeIDUInt::MAX)));
+    fn find_unallocated<Snapshot: WritableSnapshot>(&self, snapshot: &Snapshot, start: TypeIDUInt) -> Result<Option<TypeIDUInt>, EncodingError> {
+        let mut type_vertex_iter = snapshot.iterate_range(
+            PrefixRange::new_inclusive(
+                (self.vertex_constructor)(TypeID::build(start)).into_storage_key(),
+                (self.vertex_constructor)(TypeID::build(TypeIDUInt::MAX)).into_storage_key()
+            )
+        );
         for expected_next in (start..=TypeIDUInt::MAX) {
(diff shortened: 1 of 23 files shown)


encoding/graph/type_/vertex_generator.rs
```diff
@@ -4,27 +4,85 @@
...
+use storage::key_value::{StorageKey, StorageKeyReference};
 use storage::snapshot::WritableSnapshot;
 
 use crate::{
     graph::type_::vertex::{
-        build_vertex_attribute_type, build_vertex_attribute_type_prefix, build_vertex_entity_type,
-        build_vertex_entity_type_prefix, build_vertex_relation_type, build_vertex_relation_type_prefix,
-        build_vertex_role_type, build_vertex_role_type_prefix, TypeID, TypeVertex,
+        build_vertex_attribute_type, build_vertex_entity_type, build_vertex_relation_type,
+        build_vertex_role_type, TypeID, TypeVertex,
```
krishnangovindraj, line 19: These are currently generated by macros, though they could likely be const instances of a struct.

```rust
pub struct TypeVertexConstructor {
    prefix : Prefix
}

impl TypeVertexConstructor {
    pub fn new_vertex(&self, bytes: Bytes<'_, BUFFER_KEY_INLINE>) -> TypeVertex<'_> {
        let vertex = TypeVertex::new(bytes);
        debug_assert_eq!(vertex.prefix(), self.prefix);
        vertex
    }

    pub fn build_vertex(&self, type_id: TypeID) -> TypeVertex<'static> {
        TypeVertex::build(self.prefix.prefix_id(), type_id)
    }

    pub fn build_prefix(&self) -> StorageKey<'static, { TypeVertex::LENGTH_PREFIX }> {
        StorageKey::new_ref(TypeVertex::KEYSPACE, ByteReference::new(&self.prefix.prefix_id().bytes))
    }

    pub fn is_same_kind(&self, bytes: Bytes<'_, BUFFER_KEY_INLINE>) -> bool {
        bytes.length() == TypeVertex::LENGTH && TypeVertex::new(bytes).prefix() == self.prefix
    }
}

const ENTITY_TYPE_CONSTRUCTOR : TypeVertexConstructor = TypeVertexConstructor { prefix: Prefix::VertexEntityType };
const RELATION_TYPE_CONSTRUCTOR : TypeVertexConstructor = TypeVertexConstructor { prefix: Prefix::VertexRelationType };
const ATTRIBUTE_TYPE_CONSTRUCTOR : TypeVertexConstructor = TypeVertexConstructor { prefix: Prefix::VertexAttributeType };
const ROLE_TYPE_CONSTRUCTOR : TypeVertexConstructor = TypeVertexConstructor { prefix: Prefix::VertexRoleType };
```

encoding/graph/type_/vertex_generator.rs
```diff
@@ -4,27 +4,85 @@
...
+use crate::error::EncodingErrorKind::{ExhaustedTypeIDs, FailedTypeIDAllocation};
+use crate::graph::type_::Kind;
+use crate::graph::type_::Kind::{Attribute, Entity, Relation, Role};
+use crate::graph::type_::vertex::TypeIDUInt;
+use crate::graph::Typed;
+
+pub struct TypeIDAllocator {
+    root_type : Kind,
+    last_allocated_type_id: AtomicU16,
+    to_vertex:  fn (TypeID) -> TypeVertex<'static>
```
krishnangovindraj, line 33: The fn could be replaced by the struct reference above if needed.
The root type could also be stored in the struct instead.

encoding/tests/test_type_vertex.rs
```diff
@@ -4,38 +4,162 @@
...
 use storage::MVCCStorage;
+use storage::snapshot::{CommittableSnapshot, WritableSnapshot};
 use test_utils::{create_tmp_dir, init_logging};
 
+use database::Database;
+use database::transaction::TransactionWrite;
+use encoding::graph::Typed;
+use encoding::value::label::Label;
+
+// TODO: Update all tests with higher level APIs
```
krishnangovindraj, line 26: TODO? Or just move the thing id's reset test out?

krishnangovindraj (commented):
Self review

flyingsilverfin (changes requested):

encoding/tests/test_type_vertex.rs
krishnangovindraj: We decided we're not bothering to reset the thing ID.

encoding/graph/type_/vertex_generator.rs
krishnangovindraj: Added a todo in the code to do it.

flyingsilverfin (approved):

PR #7041, opened 2024-04-18
Bump bazel-distribution

krishnangovindraj:
## Usage and product changes
Bumps dependency on bazel distribution rules to include version number in the apt package filename

## Implementation
Bumps bazel-distribution


krishnangovindraj:
Already bumped to a higher commit.

PR #7042, opened 2024-04-19
Bootup

dmitrii-ubskii:


storage/storage.rs
```diff
@@ -276,165 +144,124 @@ impl<D> MVCCStorage<D> {
...
+        Self::persist_commit_status(conflict.is_none(), commit_sequence_number, &self.durability_service)
+            .map_err(|error| Durability { name: self.name.clone(), source: error })?;
+
+        match conflict {
+            Some(conflict) => Err(StorageCommitError::Isolation { name: self.name.clone(), conflict }),
+            None => Ok(()),
+        }
+    }
+
+    fn set_initial_put_status(&self, snapshot: &impl CommittableSnapshot<Durability>)
```
flyingsilverfin, line 204: `validate_puts`? or `initialise_put_operations` or something?

storage/storage.rs
```diff
@@ -550,56 +356,137 @@ impl<D> MVCCStorage<D> {
...
+    let mut recovered_commits = BTreeMap::new();
+    for record in durability_service
+        .iter_from(next_sequence_number)
+        .map_err(|error| Durability { name: storage_name.to_owned(), source: error })?
+    {
+        let RawRecord { sequence_number, record_type, bytes } =
+            record.map_err(|error| IO { name: storage_name.to_owned(), source: error })?;
+        match record_type {
+            CommitRecord::RECORD_TYPE => {
+                let commit_record = CommitRecord::deserialise_from(&mut &*bytes).unwrap(); // FIXME
```
flyingsilverfin, line 431: What's this fixme?

flyingsilverfin: And below

storage/storage.rs
```diff
@@ -452,31 +279,23 @@ impl<D> MVCCStorage<D> {
         }
 
         let previous_checkpoints: Vec<_> = fs::read_dir(&checkpoint_dir)
-            .map_err(|error| ReadCheckpointDir { dir: checkpoint_dir.clone(), source: error })?
-            .map_ok(|entry| entry.path())
-            .try_collect()
+            .and_then(|entries| entries.map_ok(|entry| entry.path()).try_collect())
             .map_err(|error| ReadCheckpointDir { dir: checkpoint_dir.clone(), source: error })?;
 
         let current_checkpoint_dir = checkpoint_dir.join(format!("{}", Utc::now().timestamp_micros()));
```
flyingsilverfin, line 283: I think creating a checkpoint directory name and reading from a checkpoint directory name string to a UTC timestamp is critical enough we should create methods for these steps

storage/storage.rs
```diff
@@ -550,56 +356,137 @@ impl<D> MVCCStorage<D> {
...
+                            unreachable!("found second commit status for a record")
+                        };
+                        record
+                    });
+                    recovered_commits.insert(commit_record_sequence_number, CheckpointCommitStatus::Validated(record));
+                } else {
+                    recovered_commits.insert(commit_record_sequence_number, CheckpointCommitStatus::Rejected);
+                }
+            }
+            _other => (), // skip?
```
flyingsilverfin, line 449: This should be a unreachable - we should stay away from code branches that will silently continue and let us get away with not updating all affected branches of code if something changes (like a new record type).

storage/storage.rs
```diff
@@ -276,165 +144,124 @@ impl<D> MVCCStorage<D> {
...
-        //  1. make durable and get sequence number
-        let commit_sequence_number =
-            self.durability_service.sequenced_write(&commit_record).map_err(|err| MVCCStorageError {
-                storage_name: self.name.to_string(),
-                kind: MVCCStorageErrorKind::DurabilityError { source: err },
-            })?;
-
-        // 2,3,4:
-        self.try_write_commit_record(
+    fn apply_commit(
```
flyingsilverfin, line 229: Can we try to make this take the data?

flyingsilverfin: First pass: let's construct the write batches before sending the record to isolation manager.

flyingsilverfin: Let's then inline this function

storage/storage.rs
```diff
@@ -20,243 +21,111 @@ use std::{
...
-        validate_new_keyspace(keyspace_id, &keyspaces, &keyspaces_index)
-            .map_err(|error| KeyspaceValidation { source: error })?;
-        keyspaces.push(Keyspace::open(path, keyspace_id, &options).map_err(|error| KeyspaceOpen { source: error })?);
-        keyspaces_index[keyspace_id.id().0 as usize] = Some(KeyspaceId(keyspaces.len() as u8 - 1));
-    }
-    Ok((keyspaces, keyspaces_index))
-}
-
-impl<D> MVCCStorage<D> {
+impl<Durability> MVCCStorage<Durability> {
```
flyingsilverfin, line 63: Let's create a `Checkpoint` module.
We can move both create and load methods into this new module.

`Chekpoint.load(path)` makes a stateful `Checkpoint` struct, which you can query for number of applied commits, etc.

`Storage.open()` can then load the last checkpoint, and if any commits were applied, immediately create a new one before returning the completed storage.

I think we can also try to hide away the usage of isolation manager into a temporary isolation manager used only for the duration of recovery.

flyingsilverfin (changes requested):

storage/isolation_manager.rs
```diff
@@ -47,105 +48,108 @@ impl IsolationManager {
...
+        self.timeline
+            .try_get_window(sequence_number)
+            .ok_or(ExpectedWindowError { sequence_number })?
+            .set_applied(sequence_number);
         self.timeline.may_increment_watermark(sequence_number);
+        Ok(())
     }
 
-    pub(crate) fn load_applied(&self, sequence_number: SequenceNumber, commit_record: CommitRecord) {
+    pub(crate) fn load_validated(&self, sequence_number: SequenceNumber, commit_record: CommitRecord) {
```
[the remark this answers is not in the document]
dmitrii-ubskii: "Applied" means it's in storage. That was correct before we rolled back the storage state during recovery. The mistake here was not changing the internal line to `set_validated()`.

storage/checkpoint.rs
```diff
@@ -0,0 +1,292 @@
...
+    },
+    write_batches::WriteBatches,
+    MVCCStorage, StorageCommitError,
+};
+
+pub(crate) struct Checkpoint {
+    pub keyspaces: Keyspaces,
+    pub next_sequence_number: SequenceNumber,
+    pub directory: Option<PathBuf>,
+}
```
dmitrii-ubskii, line 33: Doesn't really make sense to call this struct a `Checkpoint`, this is more a result of restoring _from_ a checkpoint (and a WAL)

storage/isolation_manager.rs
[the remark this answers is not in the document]
flyingsilverfin: Righttt

flyingsilverfin (approved):

PR #7043, opened 2024-04-22
Check supertypes of relation when inferring player role types during insert

dmitrii-ubskii:
## Usage and product changes

Given the following minimal schema:
```
define
player sub entity, plays super-relation:super-role;
super-relation sub relation, relates super-role;
sub-relation sub super-relation;
```

It was impossible to insert an instance of `sub-relation` relating a `player` without explicitly specifying its role:
```
> match $player isa player; 
  insert ($player) isa sub-relation; 
[THW27] Invalid Thing Write: Unable to add role player '$player' to the relation, 
        as there is no provided or inferrable role type.
```

The reason for this was that during the handling of the `insert` query, we check if it can play a role defined directly on the requested relation type, without considering any role types it may have inherited. This PR resolves that issue.

## Implementation

Instead of iterating through the player's possible role types and checking that it's defined on the target relation, we iterate through the role types the relation relates, and check if the player's type can play those.

### DIFF
--- a/query/common/Util.java
+++ b/query/common/Util.java
@@ -32,7 +32,7 @@ public class Util {
                 throw TypeDBException.of(TYPE_NOT_FOUND, Label.of(var.label().get().label(), relationType.getLabel().name()));
             }
         } else if ((inferred = player.getType().getPlays()
-                .filter(rt -> rt.getRelationType().equals(relation.getType()))
+                .filter(rt -> relation.getType().getSupertypes().anyMatch(supertype -> supertype.equals(rt.getRelationType())))
                 .toSet()).size() == 1) {
             roleType = inferred.iterator().next();
         } else if (inferred.size() > 1) {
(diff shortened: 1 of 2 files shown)


query/common/Util.java
```diff
@@ -31,8 +31,8 @@ public static RoleType tryInferRoleType(Relation relation, Thing player, Relatio
             if ((roleType = relationType.getRelates(var.label().get().label())) == null) {
                 throw TypeDBException.of(TYPE_NOT_FOUND, Label.of(var.label().get().label(), relationType.getLabel().name()));
             }
-        } else if ((inferred = player.getType().getPlays()
-                .filter(rt -> rt.getRelationType().equals(relation.getType()))
+        } else if ((inferred = relation.getType().getRelates()
+                .filter(relatesRole -> player.getType().plays(relatesRole))
```
dmitrii-ubskii, line 35: Had to flip this, because we have `ThingType.plays(RoleType) -> boolean`, but not `RelationType.relates(RoleType) -> boolean`

PR #7050, opened 2024-04-26
Refactor type cache construction and Type Manager read-through queries to use the same path

krishnangovindraj:
## Usage and product changes
Refactors type cache construction and Type Manager read-through queries to use the same path

## Implementation
* Pulls all `storage_get` methods out into a `TypeManagerStorageSource` class.
* Replace all loading logic from type_cache with calls to `TypeManagerStorageSource` methods.
* Introduce methods for getting subtypes into the four vertex types.
* Make an attempt to use generics for meta-programming for in ConceptAPI in hopes of replacing macros in the future.

### DIFF
--- a/concept/type_/type_manager.rs
+++ b/concept/type_/type_manager.rs
@@ -386,17 +386,11 @@ impl<'_s, Snapshot: ReadableSnapshot> TypeManager<Snapshot>
         }
     }
 
-    pub(crate) fn get_owns_ordering(&self, owns: Owns<'_>) -> Result<Ordering, ConceptReadError> {
+    pub(crate) fn get_owns_ordering(&self, owns: Owns<'_s>) -> Result<Ordering, ConceptReadError> {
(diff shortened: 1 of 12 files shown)


concept/type_/type_manager.rs
```diff
@@ -234,14 +244,37 @@ macro_rules! get_type_annotations {
...
+    // }
+    //
+    // pub fn get_supertype<'b, T: TypeAPI<'_s> + ReadableType<'_s, 'b>>(&self, type_: T) -> Result<Option<T::SelfWithLifetime>, ConceptReadError>
+    // {
+    //     if let Some(cache) = &self.type_cache {
+    //         todo!("Ok(cache.$cache_method(label))")
+    //     } else {
+    //         StorageTypeManagerSource::storage_get_supertype(self.snapshot.as_ref(), type_)
+    //     }
+    // }
```
krishnangovindraj, line 273: We could replace some macros with these methods.

PR #7055, opened 2024-05-01
Fix bug when isolation manager finds & validates against concurrent windows

krishnangovindraj:
## Usage and product changes
Fixes bugs in the isolation manager when the open or commit sequence numbers lie on the edge of or across windows.


flyingsilverfin (approved):
Nice

PR #7056, opened 2024-05-02
Purge SchemaSnapshot (& WritableSnapshot)

dmitrii-ubskii:


flyingsilverfin:
We won't be proceeding with this

PR #7058, opened 2024-05-06
Fix generics on ReadableType trait

krishnangovindraj:
## Usage and product changes
Fix generics on ReadableType trait

## Implementation
Changes the output type lifetime to be a generic parameter on the method rather than on the trait.

### DIFF
--- a/concept/type_/type_cache.rs
+++ b/concept/type_/type_cache.rs
@@ -171,9 +171,9 @@ impl TypeCache {
         })
     }
 
-    fn build_common_cache<Snapshot: ReadableSnapshot, OUT: 'static, T>(snapshot: &Snapshot, type_: T) -> CommonTypeCache<OUT>
+    fn build_common_cache<Snapshot: ReadableSnapshot, OUT, T>(snapshot: &Snapshot, type_: T) -> CommonTypeCache<OUT>
     where
-        T: TypeAPI<'static> + ReadableType<'static, OUT>,
+        T: TypeAPI<'static> + ReadableType<Out<'static>=OUT>,
         OUT: TypeAPI<'static> + TypeAPITraits + 'static
     {
         let label = TypeReader::get_label(snapshot, type_.clone()).unwrap().unwrap();
--- a/concept/type_/type_cache.rs
+++ b/concept/type_/type_cache.rs
@@ -171,24 +171,23 @@ impl TypeCache {
         })
     }
 
-    fn build_common_cache<Snapshot: ReadableSnapshot, OUT, T>(snapshot: &Snapshot, type_: T) -> CommonTypeCache<OUT>
+    fn build_common_cache<Snapshot: ReadableSnapshot, T>(snapshot: &Snapshot, type_: T) -> CommonTypeCache<T>
     where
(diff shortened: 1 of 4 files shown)


concept/type_/type_cache.rs
```diff
@@ -176,7 +171,10 @@ impl TypeCache {
         })
     }
 
-    fn build_common_cache<Snapshot: ReadableSnapshot, T: ReadableType<'static, 'static, SelfRead = T>>(snapshot: &Snapshot, type_: T) -> CommonTypeCache<T::SelfRead> {
+    fn build_common_cache<Snapshot: ReadableSnapshot, T>(snapshot: &Snapshot, type_: T) -> CommonTypeCache<T>
+    where
+        T: TypeAPI<'static> + TypeAPITraits + ReadableType<Output<'static>=T>,
```
dmitrii-ubskii, line 176: I thin kwe should either stick with the `impl` shorthand when we don't need to name the type:
```suggestion
    fn build_common_cache<T>(snapshot: &impl ReadableSnapshot, type_: T) -> CommonTypeCache<T>
    where
        T: TypeAPI<'static> + TypeAPITraits + ReadableType<Output<'static>=T>,
```
or move all bounds to the `where` clause:
```suggestion
    fn build_common_cache<Snapshot, T>(snapshot: &Snapshot, type_: T) -> CommonTypeCache<T>
    where
        Snapshot: ReadableSnapshot,
        T: TypeAPI<'static> + TypeAPITraits + ReadableType<Output<'static>=T>,
```
Mixing inline bounds and `where` isn't very nice to read

concept/type_/type_manager.rs
```diff
@@ -743,62 +742,62 @@ impl<Snapshot: WritableSnapshot> TypeManager<Snapshot> {
...
 }
 
-pub trait ReadableType<'a, 'b>: TypeAPI<'a> + CommonType<'a> {
+pub trait ReadableType {
     // Consider replacing 'b with 'static
-    type SelfRead: ReadableType<'b, 'b>;
-    fn read_from(b: Bytes<'b, BUFFER_KEY_INLINE>) -> Self::SelfRead;
+    type Output<'out>: 'out;
+    fn read_from<'out>(b: Bytes<'out, BUFFER_KEY_INLINE>) -> Self::Output<'out>;
 }
```
dmitrii-ubskii, line 749: `'b` or `'bytes` is better at communicating which lifetime the output is tied to.

concept/type_/type_reader.rs
```diff
@@ -37,8 +37,7 @@ pub struct TypeReader { }
 impl<'_s> TypeReader
     where '_s : 'static {
```
dmitrii-ubskii, line 38: I still don't understand why we need `'_s: 'static`. `'static` is the only lifetime that outlives `'static`.

dmitrii-ubskii (changes requested):

concept/type_/type_reader.rs
krishnangovindraj: Ah good catch. 
That's me being non-committal to requiring the parameters all be static. I'll revert.

dmitrii-ubskii (approved):
LGTM

PR #7059, opened 2024-05-07
Update diagnostics service to version 1

farost:
## Usage and product changes
We introduce an updated version of diagnostics sent from a `TypeDB` server. 

1. `config.yml` gets a new field `deploymentID` for the `diagnostics` section. This field should be used for collecting the data from multiple servers of a single `TypeDB Cloud` deployment.
2.  The updated diagnostics data contains more information about the server resources and details for each separate database. More details can be found in the examples below.
3. For the `JSON` reporting, we calculated diffs between the current timestamp and the `sinceTimestamp` (the previous hour when the data had to be sent: it's updated even if we had errors sending the data for simplicity). For the `Prometheus` data, we send raw counts as `Prometheus` calculates diffs based on its queries and expects raw diagnostics from our side.
4. For the `JSON` monitoring, we show only the incrementing counters from the start of the server just as for the Prometheus diagnostics data (also available through the monitoring page). This way, the content is different from the reporting data.
5. The `schema` and `data` diagnostics about each specific database are sent only from the primary replica of a deployment at the moment of the diagnostics collection. The `connection` peak values diagnostics regarding a database are still reported by a non-primary replica if the database exists or there were established transactions within the last hour before the database had been deleted.
6. If the statistics reporting is turned off in the config, we send a totally safe part of the diagnostics data once to notify the server about the moment when the diagnostics were turned off. No user data is shared in this snapshot (see examples below). This action is performed only if the server is up for 1 hour (to avoid our CI tests report data), and only if the server has not successfully sent such a snapshot after turning the statistics reporting off the last time. If there is an error in sending this snapshot, the server will try again after a restart (no extra logic here).

Example diagnostics data for Prometheus (http://localhost:4104/metrics?format=prometheus):
```
# distribution: TypeDB Core
# version: 2.28.0
# os: Mac OS X x86_64 14.2.1

# TYPE server_resources_count gauge
server_resources_count{kind="memoryUsedInBytes"} 68160245760
server_resources_count{kind="memoryAvailableInBytes"} 559230976
server_resources_count{kind="diskUsedInBytes"} 175619862528
server_resources_count{kind="diskAvailableInBytes"} 1819598303232

# TYPE typedb_schema_data_count gauge
typedb_schema_data_count{database="212487319", kind="typeCount"} 74
typedb_schema_data_count{database="212487319", kind="entityCount"} 2891
typedb_schema_data_count{database="212487319", kind="relationCount"} 2466
typedb_schema_data_count{database="212487319", kind="attributeCount"} 5832
typedb_schema_data_count{database="212487319", kind="hasCount"} 13325
typedb_schema_data_count{database="212487319", kind="roleCount"} 7984
typedb_schema_data_count{database="212487319", kind="storageInBytes"} 2164793
typedb_schema_data_count{database="212487319", kind="storageKeyCount"} 94028
typedb_schema_data_count{database="3717486", kind="typeCount"} 5
typedb_schema_data_count{database="3717486", kind="entityCount"} 0
typedb_schema_data_count{database="3717486", kind="relationCount"} 0
typedb_schema_data_count{database="3717486", kind="attributeCount"} 0
typedb_schema_data_count{database="3717486", kind="hasCount"} 0
typedb_schema_data_count{database="3717486", kind="roleCount"} 0
typedb_schema_data_count{database="3717486", kind="storageInBytes"} 0
typedb_schema_data_count{database="3717486", kind="storageKeyCount"} 0

# TYPE typedb_attempted_requests_total counter
typedb_attempted_requests_total{kind="CONNECTION_OPEN"} 4
typedb_attempted_requests_total{kind="DATABASES_ALL"} 4
typedb_attempted_requests_total{kind="DATABASES_GET"} 4
typedb_attempted_requests_total{kind="SERVERS_ALL"} 4
typedb_attempted_requests_total{database="212487319", kind="DATABASES_CONTAINS"} 2
typedb_attempted_requests_total{database="212487319", kind="SESSION_OPEN"} 2
typedb_attempted_requests_total{database="212487319", kind="TRANSACTION_EXECUTE"} 70
typedb_attempted_requests_total{database="212487319", kind="SESSION_CLOSE"} 1
typedb_attempted_requests_total{database="3717486", kind="DATABASES_CONTAINS"} 2
typedb_attempted_requests_total{database="3717486", kind="SESSION_OPEN"} 2
typedb_attempted_requests_total{database="3717486", kind="TRANSACTION_EXECUTE"} 54
typedb_attempted_requests_total{database="3717486", kind="SESSION_CLOSE"} 1

# TYPE typedb_successful_requests_total counter
typedb_successful_requests_total{kind="CONNECTION_OPEN"} 4
typedb_successful_requests_total{kind="DATABASES_ALL"} 4
typedb_successful_requests_total{kind="DATABASES_GET"} 4
typedb_successful_requests_total{kind="SERVERS_ALL"} 4
typedb_successful_requests_total{kind="USER_TOKEN"} 8
typedb_successful_requests_total{database="212487319", kind="DATABASES_CONTAINS"} 2
typedb_successful_requests_total{database="212487319", kind="SESSION_OPEN"} 2
typedb_successful_requests_total{database="212487319", kind="TRANSACTION_EXECUTE"} 67
typedb_successful_requests_total{database="212487319", kind="SESSION_CLOSE"} 1
typedb_successful_requests_total{database="3717486", kind="DATABASES_CONTAINS"} 2
typedb_successful_requests_total{database="3717486", kind="SESSION_OPEN"} 2
typedb_successful_requests_total{database="3717486", kind="TRANSACTION_EXECUTE"} 47
typedb_successful_requests_total{database="3717486", kind="SESSION_CLOSE"} 1

# TYPE typedb_error_total counter
typedb_error_total{database="3717486", code="TYR03"} 5
typedb_error_total{database="3717486", code="TXN08"} 2
```

Example diagnostics JSON data from monitoring (http://localhost:4104/metrics?format=JSON):
```
{
  "version": 1,
  "deploymentID": "HTAOYJNSRYY2WOUR",
  "serverID": "HTAOYJNSRYY2WOUR",
  "distribution": "TypeDB Core",
  "timestamp": "2024-05-14T09:50:46",
  "server": {
    "version": "2.28.0",
    "uptimeInSeconds": 134,
    "os": {
      "name": "Mac OS X",
      "arch": "x86_64",
      "version": "14.2.1"
    },
    "memoryUsedInBytes": 68151644160,
    "memoryAvailableInBytes": 567832576,
    "diskUsedInBytes": 175619862528,
    "diskAvailableInBytes": 1819598303232
  },
  "load": [
    {
      "database": "212487319",
      "schema": {
        "typeCount": 74
      },
      "data": {
        "entityCount": 2891,
        "relationCount": 2466,
        "attributeCount": 5832,
        "hasCount": 13325,
        "roleCount": 7984,
        "storageInBytes": 2164793,
        "storageKeyCount": 94028
      }
    },
    {
      "database": "3717486",
      "schema": {
        "typeCount": 5
      },
      "data": {
        "entityCount": 0,
        "relationCount": 0,
        "attributeCount": 0,
        "hasCount": 0,
        "roleCount": 0,
        "storageInBytes": 0,
        "storageKeyCount": 0
      }
    }
  ],
  "actions": [
    {
      "name": "CONNECTION_OPEN",
      "attempted": 4,
      "successful": 4
    },
    {
      "name": "DATABASES_ALL",
      "attempted": 4,
      "successful": 4
    },
    {
      "name": "DATABASES_GET",
      "attempted": 4,
      "successful": 4
    },
    {
      "name": "SERVERS_ALL",
      "attempted": 4,
      "successful": 4
    },
    {
      "name": "DATABASES_CONTAINS",
      "database": "212487319",
      "attempted": 2,
      "successful": 2
    },
    {
      "name": "SESSION_OPEN",
      "database": "212487319",
      "attempted": 2,
      "successful": 2
    },
    {
      "name": "TRANSACTION_EXECUTE",
      "database": "212487319",
      "attempted": 70,
      "successful": 67
    },
    {
      "name": "SESSION_CLOSE",
      "database": "212487319",
      "attempted": 1,
      "successful": 1
    },
    {
      "name": "DATABASES_CONTAINS",
      "database": "3717486",
      "attempted": 2,
      "successful": 2
    },
    {
      "name": "SESSION_OPEN",
      "database": "3717486",
      "attempted": 2,
      "successful": 2
    },
    {
      "name": "TRANSACTION_EXECUTE",
      "database": "3717486",
      "attempted": 54,
      "successful": 47
    },
    {
      "name": "SESSION_CLOSE",
      "database": "3717486",
      "attempted": 1,
      "successful": 1
    }
  ],
  "errors": [
    {
      "code": "TYR03",
      "database": "3717486",
      "count": 5
    },
    {
      "code": "TXN08",
      "database": "3717486",
      "count": 2
    }
  ]
}
```

Example of diagnostics JSON data sent when the reporting flag is turned **on**: 
```
{
  "version":1,
  "deploymentID":"HTAOYJNSRYY2WOUR",
  "serverID":"HTAOYJNSRYY2WOUR",
  "distribution":"TypeDB Core",
  "timestamp":"2024-05-14T09:50:36",
  "periodInSeconds":3600,
  "enabled":true,
  "server":{
    "version":"2.28.0",
    "uptimeInSeconds":124,
    "os":{
      "name":"Mac OS X",
      "arch":"x86_64",
      "version":"14.2.1"
    },
    "memoryUsedInBytes":68097245184,
    "memoryAvailableInBytes":622231552,
    "diskUsedInBytes":175624044544,
    "diskAvailableInBytes":1819594121216
  },
  "load":[
    {
      "database":"212487319",
      "schema":{
        "typeCount":74
      },
      "data":{
        "entityCount":2868,
        "relationCount":2449,
        "attributeCount":5816,
        "hasCount":13247,
        "roleCount":7927,
        "storageInBytes":2164793,
        "storageKeyCount":93379
      },
      "connection":{
        "schemaTransactionPeakCount":0,
        "readTransactionPeakCount":1,
        "writeTransactionPeakCount":1
      }
    },
    {
      "database":"3717486",
      "schema":{
        "typeCount":5
      },
      "data":{
        "entityCount":0,
        "relationCount":0,
        "attributeCount":0,
        "hasCount":0,
        "roleCount":0,
        "storageInBytes":0,
        "storageKeyCount":0
      },
      "connection":{
        "schemaTransactionPeakCount":0,
        "readTransactionPeakCount":2,
        "writeTransactionPeakCount":1
      }
    }
  ],
  "actions":[
    {
      "name":"CONNECTION_OPEN",
      "successful":2,
      "failed":0
    },
    {
      "name":"DATABASES_ALL",
      "successful":2,
      "failed":0
    },
    {
      "name":"DATABASES_GET",
      "successful":2,
      "failed":0
    },
    {
      "name":"SERVERS_ALL",
      "successful":2,
      "failed":0
    },
    {
      "name":"DATABASES_CONTAINS",
      "database":"212487319",
      "successful":1,
      "failed":0
    },
    {
      "name":"SESSION_OPEN",
      "database":"212487319",
      "successful":1,
      "failed":0
    },
    {
      "name":"TRANSACTION_EXECUTE",
      "database":"212487319",
      "successful":32,
      "failed":2
    },
    {
      "name":"SESSION_CLOSE",
      "database":"212487319",
      "successful":1,
      "failed":0
    },
    {
      "name":"DATABASES_CONTAINS",
      "database":"3717486",
      "successful":1,
      "failed":0
    },
    {
      "name":"SESSION_OPEN",
      "database":"3717486",
      "successful":1,
      "failed":0
    },
    {
      "name":"TRANSACTION_EXECUTE",
      "database":"3717486",
      "successful":27,
      "failed":4
    },
    {
      "name":"SESSION_CLOSE",
      "database":"3717486",
      "successful":1,
      "failed":0
    }
  ],
  "errors":[
    {
      "code":"TYR03",
      "database":"3717486",
      "count":3
    },
    {
      "code":"TXN08",
      "database":"3717486",
      "count":1
    }
  ]
}
```

Example of diagnostics JSON data sent once when the reporting flag is turned **off**: 
```
{
  "version":1,
  "deploymentID":"HTAOYJNSRYY2WOUR",
  "serverID":"HTAOYJNSRYY2WOUR",
  "distribution":"TypeDB Core",
  "timestamp":"2024-05-14T10:03:53",
  "periodInSeconds":3600,
  "enabled":false,
  "server":{
    "version":"2.28.0"
  }
}
```

## Implementation
There is no huge refactoring as it's planned to be a cleaner feature in the incoming 3.0.


server/common/Constants.java
```diff
@@ -19,9 +19,11 @@ public class Constants {
     public static final String TYPEDB_LOG_FILE_NAME = "typedb";
     public static final String TYPEDB_LOG_FILE_EXT = ".log";
     public static final String TYPEDB_LOG_ARCHIVE_EXT = ".log.gz";
-    public static final String DIAGNOSTICS_REPORTING_URI = "https://3d710295c75c81492e57e1997d9e01e1@o4506315929812992.ingest.sentry.io/4506316048629760";
-    public static final String USAGE_STATISTICS_REPORTING_URI = "https://diagnostics.typedb.com/";
+    public static final String ERROR_REPORTING_URI = "https://3d710295c75c81492e57e1997d9e01e1@o4506315929812992.ingest.sentry.io/4506316048629760";
```
farost, line 22: Sentry is more about errors rather than diagnostics now + we call it `reporting errors` in the config.

common/diagnostics/Metrics.java
```diff
@@ -6,203 +6,752 @@
...
+        this.base = new BaseProperties(JSON_API_VERSION, deploymentID, serverID, name, reportingEnabled);
+        this.serverProperties = new ServerProperties(version, dataDirectory);
+    }
 
-    public Metrics(String serverID, String name, String version) {
-        this.system = new SystemProperties(serverID, name, version);
-        this.requests = new NetworkRequests();
-        this.usage = new CurrentCounts();
-        this.userErrors = new UserErrorStatistics();
+    public void takeSnapshot() {
```
farost, line 43: This method is called every hour (even if we don't successfully send a snapshot -> we forget about the past hour in case of an error for simplicity). 
It helps us take snapshots of the previous hour's data: reset peak values to the current ones, remember current counts to calculate diffs for the upcoming hour, etc.

common/diagnostics/Metrics.java
```diff
@@ -6,203 +6,752 @@
...
+        this.base.updateSinceTimestamp();
+        this.databaseLoad.values().forEach(DatabaseLoadDiagnostics::takeSnapshot);
+        this.requests.values().forEach(NetworkRequests::takeSnapshot);
+        this.userErrors.values().forEach(UserErrorStatistics::takeSnapshot);
     }
 
-    public void requestAttempt(Metrics.NetworkRequests.Kind kind) {
-        requests.attempt(kind);
+    private String hashAndAddDatabaseIfAbsent(@Nullable String databaseName) {
+        String databaseHash = databaseName != null ? String.valueOf(databaseName.hashCode()) : "";
```
farost, line 51: I use java's hash code to hash database names instead of a smarter algorithm here as I believe it is enough for our mission: to hide the real name and not be able to reverse this hashing, to have consistent hashes for the same values and not to have collisions within **one deployment**. I don't believe that a deployment can have hundreds of thousands databases and more, so I'm not scared of collisions here.

common/diagnostics/StatisticReporter.java
```diff
@@ -10,49 +10,117 @@
...
 import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
-import static java.util.concurrent.TimeUnit.HOURS;
+import static java.lang.Math.abs;
+import static java.util.concurrent.TimeUnit.MINUTES;
 
 public class StatisticReporter {
-    protected static final Logger LOG = LoggerFactory.getLogger(StatisticReporter.class);
+    // Modify calculateInitialDelay() if you change this value!
```
farost, line 28: I didn't come up with a good formula of `calculateInitialDelay` (that calculates at what minute of the current or the future hour this server needs to send the first snapshot of diagnostics to be in sync with other servers of the deployment) based on different intervals, so I hardcoded it considering that it should be 60 now.

common/diagnostics/StatisticReporter.java
```diff
@@ -10,49 +10,117 @@
...
             conn.connect();
-
             conn.getInputStream().readAllBytes();
+            return true;
         } catch (Exception e) {
-            if (LOG.isTraceEnabled()) LOG.trace("Failed to push metrics to {}:", reportingURI, e);
-            // do nothing
-        } finally {
-            pushScheduledTask = scheduled.schedule(this::push, 1, HOURS);
+            LOG.warn("Failed to push metrics to {}: {}", reportingURI, e.getMessage());
```
farost, line 78: @dmitrii-ubskii I think that it's a good thing to log as the diagnostics server is expected to be online most of the time. In the worst case, it's 24 lines of log entries per day. Should not be bad.

database/CoreDatabaseManager.java
```diff
@@ -50,6 +50,8 @@ public class CoreDatabaseManager implements TypeDB.DatabaseManager {
 
     protected static final String RESERVED_NAME_PREFIX = "_";
 
+    protected static final int DIAGNOSTICS_SEND_PERIOD_MINUTES = 1;
```
farost, line 53: This update of database metrics is initiated by the `DatabaseManager`. I didn't want to make it more frequent in production, so 1 minute sounds like a balanced value.

server/TransactionService.java
```diff
@@ -495,6 +504,12 @@ private boolean isClientCancelled(Throwable error) {
                 ((StatusRuntimeException) error).getStatus().getCode().equals(Status.CANCELLED.getCode());
     }
 
+    public String getDatabaseName() {
```
farost, line 507: Trying to be super safe here =)

common/diagnostics/StatisticReporter.java
```diff
@@ -10,49 +10,117 @@
...
+    private long calculateInitialDelay() {
+        int currentMinute = LocalDateTime.now().getMinute();
+        int scheduledMinute = abs(deploymentID.hashCode()) % REPORT_INTERVAL_MINUTES;
+
+        if (currentMinute > scheduledMinute) {
+            return 60 - currentMinute + scheduledMinute;
+        } else {
+            return scheduledMinute - currentMinute;
         }
     }
```
dmitrii-ubskii, line 125: This should assert that `REPORT_INTERVAL_MINUTES` is 60 so we have less of a chance to miss the change.

common/diagnostics/MonitoringServer.java
```diff
@@ -114,7 +114,7 @@ public void channelRead0(ChannelHandlerContext ctx, HttpObject msg) {
                 FullHttpResponse response = new DefaultFullHttpResponse(req.protocolVersion(), OK);
                 String query = Objects.requireNonNullElse(uri.getQuery(), "");
                 if (query.toLowerCase().contains("format=json")) {
-                    response.content().writeBytes(Unpooled.wrappedBuffer((metrics.formatJSON()).getBytes(StandardCharsets.UTF_8)));
+                    response.content().writeBytes(Unpooled.wrappedBuffer((metrics.formatJSON(false)).getBytes(StandardCharsets.UTF_8)));
```
dmitrii-ubskii, line 117: Boolean flags need to die in a fire. What's a false JSON?

common/diagnostics/Metrics.java
```diff
@@ -6,203 +6,752 @@
...
+        this.serverProperties = new ServerProperties(version, dataDirectory);
+    }
 
-    public Metrics(String serverID, String name, String version) {
-        this.system = new SystemProperties(serverID, name, version);
-        this.requests = new NetworkRequests();
-        this.usage = new CurrentCounts();
-        this.userErrors = new UserErrorStatistics();
+    public void takeSnapshot() {
+        this.base.updateSinceTimestamp();
```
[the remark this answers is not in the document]
dmitrii-ubskii: I thought the monitoring page was supposed to use the same data as Prometheus, with counters steadily ticking up?

common/diagnostics/Metrics.java
```diff
@@ -6,203 +6,761 @@
...
     }
 
-    public void requestAttempt(Metrics.NetworkRequests.Kind kind) {
-        requests.attempt(kind);
+    private String hashAndAddDatabaseIfAbsent(@Nullable String databaseName) {
+        String databaseHash = databaseName != null ? String.valueOf(databaseName.hashCode()) : "";
+
+        if (!databaseHash.isEmpty()) {
+            this.databaseLoad.computeIfAbsent(databaseHash, val -> new DatabaseLoadDiagnostics());
+        }
```
dmitrii-ubskii, line 56: These should be combined into one `if (databaseName != null)` statement for readability.

common/diagnostics/Metrics.java
```diff
@@ -6,203 +6,761 @@
...
+
+        deletedDatabases.forEach(databaseHash -> databaseLoad.get(databaseHash).setDatabaseDeleted());
+    }
+    
+    private static String formatDateTime(LocalDateTime dateTime) {
+        return dateTime.truncatedTo(ChronoUnit.SECONDS).toString();
+    }
+
+    protected JsonObject asJSON(boolean reporting) {
+        boolean isNoUsageDiagnostics = reporting && !base.getReportingEnabled();
```
dmitrii-ubskii, line 116: `isNoUsageDiagnostics` is a) negative and b) weird. `addUsageDiagnostics` would do here (`doAddUsageDiagnostics` is also acceptable and arguably more obviously boolean)

common/diagnostics/Metrics.java
```diff
@@ -6,203 +6,761 @@
...
+        JsonArray actions = new JsonArray();
+        requests.forEach((databaseHash, diagnostics) -> diagnostics.asJSON(databaseHash).forEach(actions::add));
+        metrics.add("actions", actions);
+
+        JsonArray errors = new JsonArray();
+        userErrors.forEach((databaseHash, diagnostics) -> diagnostics.asJSON(databaseHash).forEach(errors::add));
+        metrics.add("errors", errors);
+
         return metrics;
     }
```
dmitrii-ubskii, line 139: You short-circuit here already, which means this can be split and let us avoid the opaque boolan flag:

```java
    protected JsonObject asFullJSON() {
        JsonObject metrics = base.asJSON();
        metrics.add("server", serverProperties.asFullJSON());
      
        // ...

        return metrics;
    }
    
    protected JsonObject asReportingJSON() {
        if (base.getReportingEnabled()) {
            return asFullJSON();
        } else {
            JsonObject metrics = base.asJSON();        
            metrics.add("server", serverProperties.asMinimalJSON());
            return metrics;
        }
    }
```
You only switch in a few places afaics. (Names are only a suggestion).

server/TypeDBService.java
```diff
@@ -380,35 +386,33 @@ public void sessionOpen(SessionProto.Session.Open.Req request,
...
     @Override
     public void sessionClose(SessionProto.Session.Close.Req request,
                              StreamObserver<SessionProto.Session.Close.Res> responder) {
+        String databaseName = null;
         try {
-            Diagnostics.get().requestAttempt(SESSION);
             UUID sessionID = byteStringAsUUID(request.getSessionId());
             SessionService sessionSvc = sessionServices.get(sessionID);
             if (sessionSvc == null) throw TypeDBException.of(SESSION_NOT_FOUND, sessionID);
+            databaseName = sessionSvc.session().database().name();
```
dmitrii-ubskii, line 406: Not as careful here as in TransactionService, eh?

dmitrii-ubskii (changes requested):
Partial review

common/diagnostics/StatisticReporter.java
farost: Well... You're right, and since it already works, I'll return the trace, I guess.

test/integration/server/ParametersTest.java
```diff
@@ -70,8 +68,82 @@ public void test_log_archive_age_limits() throws IOException, InterruptedExcepti
...
+            assertEquals(savedServerId1, savedServerId2);
+
+            deploymentIdFiles = getDeploymentIdFiles(dataDir);
+            assertEquals(1, deploymentIdFiles.size());
+            String savedDeploymentId = Files.readString(deploymentIdFiles.get(0));
+            assertEquals(savedServerId2, savedDeploymentId);
+        } finally {
+            if (typeDBServer != null) typeDBServer.close();
+            if (dataDir != null) deleteDirectory(dataDir);
+            if (logDir != null) deleteDirectory(logDir);
```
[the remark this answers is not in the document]
farost: There is a 'smart' Util method being used, will `File.deleteOnExit()` work by itself?
`    public static void deleteDirectory(Path directory) throws IOException {
        Files.walk(directory).sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
    }`

common/diagnostics/Metrics.java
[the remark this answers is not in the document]
farost: I decided not to rename anything before 3.0 as it could become messy with multiple changes. We are used to the current naming and it will be easer to reimplement and rename it in 3.0 this way, but I'm open to renaming it if it won't disturb other engineers even more.

common/diagnostics/MonitoringServer.java
farost: Moving the conversation to `common/diagnostics/Metrics.java`

common/diagnostics/Metrics.java
farost: What do I do with `databaseHash` declaration? Will it be more readable than this version?

common/diagnostics/MonitoringServer.java
farost: I didn't want to create two methods `formatJSONMonitoring` and `formatJSONReporting` and wasn't sure what `typedb` thinks about boolean flags as arguments. I used to mark it this way in C++:
```
methodCall(
true, // flag1's name
false) // flag2's name
``` 
But everyone here (except just you?) use Intellij which shows the name of the argument inside the IDE, so I was hesitant about it. 

Based on the related question: `I thought the monitoring page was supposed to use the same data as Prometheus, with counters steadily ticking up?`, it could be a more clear difference between monitoring and reporting JSONs, so creating two methods would be better in terms of design than in the current state, so I'd just change it to two methods.

common/diagnostics/Metrics.java
farost: Could be an option. Will do the final round once we understand the purpose of the monitoring page.

server/TypeDBService.java
farost: Well, it's inside `try` block and it would be a valid throw if the request didn't contain any of `session`, `database`, or its `name`. That's what I thought.

common/diagnostics/Version.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+package com.vaticle.typedb.core.common.diagnostics;
+
+public class Version {
+    static final int JSON_API_VERSION = 1;
```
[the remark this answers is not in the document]
farost: I'm not against it. Will do.

test/integration/server/ParametersTest.java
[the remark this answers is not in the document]
dmitrii-ubskii: You're probably correct that it may not.

test/integration/server/ParametersTest.java
```diff
@@ -70,8 +68,82 @@ public void test_log_archive_age_limits() throws IOException, InterruptedExcepti
...
+        return getFiles(logDir, TYPEDB_LOG_ARCHIVE_EXT);
+    }
+
+    private List<Path> getServerIdFiles(Path dataDir) throws IOException {
+        return getFiles(dataDir, SERVER_ID_FILE_NAME);
+    }
+
+    private List<Path> getDeploymentIdFiles(Path dataDir) throws IOException {
+        return getFiles(dataDir, DEPLOYMENT_ID_FILE_NAME);
+    }
```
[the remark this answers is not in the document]
dmitrii-ubskii: `File.exists()`

common/diagnostics/Metrics.java
dmitrii-ubskii: You can have it uninitialized outside the if-statement, and initialize it in each branch. Java will complain if it's not guaranteed to be initialized at the end of it.

server/TypeDBService.java
dmitrii-ubskii: Hmm but now if `sessionSvc != null`, but `database` for some reason is, `sessionSvc.close()` won't be called. I don't know if that's even possible, but it is changing the behaviour in that case.

test/integration/server/ParametersTest.java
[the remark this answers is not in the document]
farost: Yup, forgot about `toFile()`...

common/diagnostics/Metrics.java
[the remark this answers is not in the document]
farost: Makes sense. I'd like to leave it as `Metrics` for now as it's at least another anchor which could be shown when you search for `metrics` and `diagnostics store` is a kind of another entity we'd introduce with this renaming. But I wouldn't resist if you insisted that `Metrics` is a bad option to leave in the codebase.

common/diagnostics/Metrics.java
farost: Well, if you prefer `if-else` over this, will do.

server/TypeDBService.java
farost: You're right, I didn't pay enough attention to the business logic here, my bad. And having a `throw` because of this line after everything would be dumb as well. Will be careful here as well then...

common/diagnostics/Diagnostics.java
```diff
@@ -128,16 +144,25 @@ protected static void initSentry(String serverID, String distributionName, Strin
...
             if (monitoringServer != null) monitoringServer.startServing(sslContext, middleware);
         }
 
         @Override
-        public void submitError(Throwable error) {
+        public void mayStartReporting() {
+            if (statisticReporter == null) {
+                LOG.error("Cannot start reporting because statistic reporter is not initialised yet.");
+                return;
+            }
```
dmitrii-ubskii, line 156: A `Diagnostics.Core` should never have a `null` `statisticReporter` in the current state.

common/diagnostics/Diagnostics.java
```diff
@@ -128,16 +144,25 @@ protected static void initSentry(String serverID, String distributionName, Strin
         }
 
         @Override
-        public void mayStartServing(@Nullable SslContext sslContext, ChannelInboundHandlerAdapter... middleware) {
+        public void mayStartMonitoring(@Nullable SslContext sslContext, ChannelInboundHandlerAdapter... middleware) {
             if (monitoringServer != null) monitoringServer.startServing(sslContext, middleware);
         }
```
dmitrii-ubskii, line 149: The monitoring server doesn't monitor, it serves the data which the user can use to monitor the statistics.

test/integration/server/ParametersTest.java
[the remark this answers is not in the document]
farost: Gonna leave it like this then.

common/diagnostics/Metrics.java
dmitrii-ubskii: This is a little more dangerous
```java
String databaseHash = "";

if (databaseName != null) {
    // oops
    var databaseHash = String.valueOf(databaseName.hashCode());
    this.databaseLoad.computeIfAbsent(databaseHash, val -> new DatabaseLoadDiagnostics());
}
```
`else databaseHash = "";` will allow the Java compiler to catch that edit.

database/CoreDatabaseManager.java
```diff
@@ -169,4 +179,30 @@ public void close() {
     protected static boolean isReservedName(String name) {
         return name.startsWith(RESERVED_NAME_PREFIX);
     }
+
+    protected Metrics.DatabaseDiagnostics databaseDiagnostics(
+            TypeDB.Database database, Metrics.DatabaseSchemaLoad schemaLoad, Metrics.DatabaseDataLoad dataLoad
+    ) {
+        return new Metrics.DatabaseDiagnostics(database.name(), schemaLoad, dataLoad, true);
+    }
```
dmitrii-ubskii, line 187: Does this really warrant a helper at this point? It could be inlined.

dmitrii-ubskii (changes requested):

common/diagnostics/Diagnostics.java
farost: `mayStartMonitoringService` it is =)

with the `serve` word it is really unclear from the outside if it's about reporting or monitoring.

database/CoreDatabaseManager.java
farost: It's not necessary, but not a bad thing for symmetry, I guess. At least that's why I left it this way.

common/diagnostics/Metrics.java
farost: I'm not sure why would anyone want to do it. I'm not strongly against it, but this example could be applied anywhere else. Do we really prefer to have "else" instead of a `default initialization` + one override inside `if`?

Just to understand our basic approach.

database/CoreDatabaseManager.java
dmitrii-ubskii: Symmetry with what? This is only used in `submitDatabaseDiagnostics()`, which doesn't use any similar helpers.
This is indirection for the sake of indirection.

server/TypeDBService.java
dmitrii-ubskii: Let's actually add a `databaseName()` getter to the session service so that we can at least have one fewer null check in these places.

common/diagnostics/Metrics.java
dmitrii-ubskii: The thinking is so that it's impossible to forget to initialize the variable to a sensible value rather than always risking leaving the fallback unchanged.

It also reduces complexity. If `databaseHash` is uninitialized before the conditional, and used after, that means each branch of the conditional must have made a decision what to initialize it to. If the conditional is non-exhaustive and the variable has a default value, the reader must inspect each branch to know if and when the default is overridden. (And also be cognizant of shadowing as in the "oops" example).

It's perhaps less relevant here because the snippet is so small, but a nice pattern to follow nonetheless.

farost: Okay, will do.

common/diagnostics/Metrics.java
farost: Used it, thx.

server/TypeDBService.java
farost: Good idea, done.

database/CoreDatabaseManager.java
farost: Yep, I did think that this method is used in another place. Fixed.

farost: Oh! I thought it was a different method.
This indirection is needed for `TypeDB Cloud` to override only the diagnostics creation, not the logic of getting the stats!

common/diagnostics/Metrics.java
```diff
@@ -113,32 +114,66 @@ private static String formatDateTime(LocalDateTime dateTime) {
...
+        return base.getReportingEnabled()
+                ? asFullReportingJSON()
+                : asMinimalReportingJSON();
+    }
 
-        JsonObject metrics = base.asJSON();
-        metrics.add("server", serverProperties.asJSON(!addUsageDiagnostics));
+    protected JsonObject asFullReportingJSON() {
+        JsonObject metrics = base.asReportingJSON();
+        metrics.add("server", serverProperties.asReportingFullJSON());
```
dmitrii-ubskii, line 125: This should be `serverProperties.asFullReportingJSON()`. Ditto for the `Minimal` version below.

common/diagnostics/Metrics.java
```diff
@@ -546,15 +604,31 @@ JsonArray asJSON(String database, boolean isPrimaryServer) {
...
 
         String prometheusDiagnostics(String database, boolean isPrimaryServer) {
-            if (!isPrimaryServer || isDeleted) {
-                return "";
+            if (isPrimaryServer && !isDeleted) {
+                return schemaLoad.prometheusDiagnostics(database) + dataLoad.prometheusDiagnostics(database);
             }
-            return schemaLoad.prometheusDiagnostics(database) + dataLoad.prometheusDiagnostics(database);
+            return "";
         }
```
dmitrii-ubskii, line 632: Interesting decision to flip the condition here.

Before the change, this was essentially:
```java
if (!needsOutput)
    return ""; // bail
return output;
```
This became
```java
if (needsOutput)
    return output;
return "";
```

The former is a short-circuit followed by logic, the latter implies that the empty result is the default which is overridden by some condition.

At the very least the latter needs an `else` so the structure is obvious:
```java
if (needsOutput)
    return output;
else
    return "";
```

Although I would prefer if we stuck with the former.

dmitrii-ubskii (approved):
Left a couple comments to address, but otherwise LGTM!

common/diagnostics/Metrics.java
farost: Yep, I've missed the renaming

common/diagnostics/Metrics.java
farost: I've added `else` and decided to stick to the current version of the check to maintain the same check for two methods for better readability.

flyingsilverfin (approved):

PR #7060, opened 2024-05-08
Test that deletes of committed keys are seen by deleting snapshot

krishnangovindraj:
## Usage and product changes
Add test checking deletes of committed keys are seen by the deleting snapshot

### DIFF
--- a/storage/tests/test_snapshot.rs
+++ b/storage/tests/test_snapshot.rs
@@ -159,6 +159,31 @@ fn snapshot_read_through() {
 }
 
 #[test]
+fn snapshot_read_buffered_delete_of_persisted_key() {
+    init_logging();
+    let storage_path = create_tmp_dir();
+    let storage = Arc::new(MVCCStorage::<WAL>::open::<TestKeyspaceSet>("storage", &storage_path).unwrap());
+
+    let key_1 = StorageKeyArray::<BUFFER_KEY_INLINE>::from((Keyspace, [0x0, 0x0, 0x1]));
+    let key_2 = StorageKeyArray::<BUFFER_KEY_INLINE>::from((Keyspace, [0x1, 0x0, 0x10]));
+    {
+        let mut snapshot = storage.clone().open_snapshot_write();
+        snapshot.put(key_1.clone());
+        snapshot.put(key_2.clone());
+        snapshot.commit().unwrap();
+    }
+
+    {
+        let mut snapshot = storage.clone().open_snapshot_write();;
+        assert!(snapshot.get::<48>(StorageKey::Array(key_1.clone()).as_reference()).unwrap().is_some());
+        assert!(snapshot.get::<48>(StorageKey::Array(key_2.clone()).as_reference()).unwrap().is_some());
+        snapshot.delete(key_2.clone());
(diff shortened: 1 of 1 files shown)


storage/snapshot/buffer.rs
```diff
@@ -136,11 +136,8 @@ impl WriteBuffer {
         self.writes.get(key.bytes()).is_some()
     }
 
-    pub(crate) fn get<const INLINE_BYTES: usize>(&self, key: &[u8]) -> Option<ByteArray<INLINE_BYTES>> {
-        match self.writes.get(key) {
-            Some(Write::Insert { value }) | Some(Write::Put { value, .. }) => Some(ByteArray::copy(value.bytes())),
-            Some(Write::Delete) | None => None,
-        }
+    pub(crate) fn get(&self, key: &[u8]) -> Option<&Write> {
+        self.writes.get(key)
```
krishnangovindraj, line 140: This is trivial now. But dissolving it breaks the abstraction

storage/snapshot/snapshot.rs
```diff
@@ -94,17 +94,20 @@ pub trait WritableSnapshot: ReadableSnapshot {
...
+            }
         } else {
             let storage_value = self.get_mapped(key.as_reference(), |reference| ByteArray::from(reference))?;
             if let Some(value) = storage_value {
                 self.operations_mut().lock_add(ByteArray::copy(key.bytes()), LockType::Unmodifiable);
                 Ok(value)
             } else {
-                // TODO: what if the user concurrent requires a concept while deleting it in another query
-                unreachable!("Require key exists in snapshot or in storage.");
+                Err(SnapshotGetError::GetRequiredKeyDoesntExist { key: StorageKey::Array(key.into_owned_array()) })
```
krishnangovindraj, line 110: Is this really an illegal state worthy of panic? Or should we be returning an error?

storage/snapshot/snapshot.rs
```diff
@@ -409,6 +425,7 @@ impl Error for SnapshotError {
 #[derive(Debug, Clone)]
 pub enum SnapshotGetError {
     MVCCRead { source: MVCCReadError },
+    GetRequiredKeyDoesntExist { key: StorageKey<'static, BUFFER_KEY_INLINE> }
```
krishnangovindraj, line 428: Do we need to return the key?

krishnangovindraj (commented):
self review comments

dmitrii-ubskii (changes requested):

storage/snapshot/snapshot.rs
dmitrii-ubskii: For something that's an internal error I propose we call them `Expected<Precondition>`, in this case `ExpectedRequiredKeyToExist`.
Returning the key seems helpful.

storage/snapshot/snapshot.rs
krishnangovindraj: From the global perspective, yes.
But from the perspective of a snapshot?
Should we panic here or should the callers call unwrap?

dmitrii-ubskii (approved):

PR #7062, opened 2024-05-09
Basic BDD tests for Type API

krishnangovindraj:
## Usage and product changes
Adds basic BDD test targets & steps for Type API.

### DIFF
--- a/concept/tests/test_type.rs
+++ b/concept/tests/test_type.rs
@@ -80,7 +80,7 @@ fn entity_usage() {
         assert_eq!(supertypes.len(), 2);
 
         // --- child owns age ---
-        child_type.set_owns(&mut snapshot, &type_manager, age_type.clone().into_owned(), Ordering::Unordered);
+        child_type.set_owns(&mut snapshot, &type_manager, age_type.clone().into_owned(), Ordering::Unordered).unwrap();
         let owns = child_type.get_owns_attribute(&snapshot, &type_manager, age_type.clone().into_owned()).unwrap().unwrap();
         // TODO: test 'owns' structure directly
 
@@ -206,7 +206,7 @@ fn role_usage() {
 
         // --- person plays friendship:friend ---
         let person_type = type_manager.create_entity_type(&mut snapshot, &person_label, false).unwrap();
-        person_type.set_plays(&mut snapshot, &type_manager, role_type.clone().into_owned());
+        person_type.set_plays(&mut snapshot, &type_manager, role_type.clone().into_owned()).unwrap();
         let plays = person_type.get_plays_role(&snapshot, &type_manager, role_type.clone().into_owned()).unwrap().unwrap();
         debug_assert_eq!(plays.player(), ObjectType::Entity(person_type.clone()));
         debug_assert_eq!(plays.role(), role_type);
--- a/concept/tests/test_type.rs
+++ b/concept/tests/test_type.rs
@@ -91,12 +91,26 @@ fn entity_usage() {
         assert!(child_type.has_owns_attribute(&snapshot, &type_manager, age_type.clone()).unwrap());
 
         // --- adult sub person ---
-        let adult = type_manager.create_entity_type(&mut snapshot, &Label::build("adult"), false).unwrap();
-        adult.set_supertype(&mut snapshot, &type_manager, person_type.clone()).unwrap();
+        let adult_type = type_manager.create_entity_type(&mut snapshot, &Label::build("adult"), false).unwrap();
+        adult_type.set_supertype(&mut snapshot, &type_manager, person_type.clone()).unwrap();
         assert_eq!(root_entity.get_subtypes(&snapshot, &type_manager).unwrap().len(), 1);
         assert_eq!(root_entity.get_subtypes_transitive(&snapshot, &type_manager).unwrap().len(), 3);
         assert_eq!(person_type.get_subtypes(&snapshot, &type_manager).unwrap().len(), 2);
         assert_eq!(person_type.get_subtypes_transitive(&snapshot, &type_manager).unwrap().len(), 2);
+
+        // --- owns inheritance ---
+        let height_label = Label::new_static("height");
+        let height_type = type_manager.create_attribute_type(&mut snapshot, &height_label, false).unwrap();
+        person_type.set_owns(&mut snapshot, &type_manager, height_type.clone(), Ordering::Unordered).unwrap();
+
+        match child_type.get_owns_attribute_transitive(&snapshot, &type_manager, height_type.clone()).unwrap() {
+            None => assert!(false, "child should inherit ownership of height"),
+            Some(child_owns_height) => {
+                assert_eq!(height_type, child_owns_height.attribute());
+                assert_eq!(ObjectType::Entity(person_type.clone()), child_owns_height.owner());
+            },
+        }
+
     }
     snapshot.commit().unwrap();
 
@@ -157,6 +171,16 @@ fn entity_usage() {
         assert_eq!(root_entity.get_subtypes_transitive(&snapshot, &type_manager).unwrap().len(), 3);
         assert_eq!(person_type.get_subtypes(&snapshot, &type_manager).unwrap().len(), 2);
         assert_eq!(person_type.get_subtypes_transitive(&snapshot, &type_manager).unwrap().len(), 2);
+
+        // --- owns inheritance ---
+        let height_type = type_manager.get_attribute_type(&snapshot, &Label::new_static("height")).unwrap().unwrap();
+        match child_type.get_owns_attribute_transitive(&snapshot, &type_manager, height_type.clone()).unwrap() {
+            None => assert!(false, "child should inherit ownership of height"),
+            Some(child_owns_height) => {
+                assert_eq!(height_type, child_owns_height.attribute());
+                assert_eq!(ObjectType::Entity(person_type.clone()), child_owns_height.owner());
+            },
+        }
     }
 }
 
--- a/concept/type_/entity_type.rs
+++ b/concept/type_/entity_type.rs
@@ -4,7 +4,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 
 use bytes::Bytes;
 use encoding::{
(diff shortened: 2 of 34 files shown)


concept/tests/test_type.rs
```diff
@@ -157,6 +171,16 @@ fn entity_usage() {
...
+
+        // --- owns inheritance ---
+        let height_type = type_manager.get_attribute_type(&snapshot, &Label::new_static("height")).unwrap().unwrap();
+        match child_type.get_owns_attribute_transitive(&snapshot, &type_manager, height_type.clone()).unwrap() {
+            None => assert!(false, "child should inherit ownership of height"),
+            Some(child_owns_height) => {
+                assert_eq!(height_type, child_owns_height.attribute());
+                assert_eq!(ObjectType::Entity(person_type.clone()), child_owns_height.owner());
+            },
+        }
```
krishnangovindraj, line 183: Extra test. I could move it to a different test case and extend it to other inheritance, or let BDD do the work.

concept/type_/owns.rs
```diff
@@ -48,6 +54,25 @@ impl<'a> Owns<'a> {
         }
     }
 
+    // TODO: Should it be 'this or just 'tm on type_manager?
```
krishnangovindraj, line 57: Lifetimes question.

concept/type_/owns.rs
```diff
@@ -133,4 +156,4 @@ impl From<Annotation> for OwnsAnnotation {
             Annotation::Independent(_) => unreachable!("Independent annotation not available for Owns."),
         }
     }
-}
```
krishnangovindraj, line 136: newline needed

concept/type_/plays.rs
```diff
@@ -25,5 +34,36 @@ impl<'a> Plays<'a> {
...
 
+impl<'a> IntoCanonicalTypeEdge<'a> for Plays<'a> {
+    fn as_type_edge(&self) -> TypeEdge<'static> {
+        build_edge_plays(self.player.vertex().clone().into_owned(), self.role.vertex().clone().into_owned())
+    }
+
+    fn into_type_edge(self) -> TypeEdge<'static> {
+        build_edge_plays(self.player.vertex().clone().into_owned(), self.role.vertex().clone().into_owned())
+    }
+}
```
krishnangovindraj, line 69: I remember there's  a place in TypeReader / TypeManager where I could use these methods instead of calling the encoding method directly.

concept/type_/relation_type.rs
```diff
@@ -233,7 +239,28 @@ impl<'a> RelationType<'a> {
...
+    }
+
+    // TODO: Review. This is a bit convoluted thanks to wanting to return the relates with the declaring relation
+    pub fn get_relates_role_transitive<Snapshot: ReadableSnapshot>(
+        &self,
+        snapshot: &Snapshot,
+        type_manager: &TypeManager<Snapshot>,
+        name: &str,
+    ) -> Result<Option<Relates<'static>>, ConceptReadError> {
+        Ok(self.get_relates_transitive(snapshot, type_manager)?.get(name).map(|relates| relates.clone()))
```
krishnangovindraj, line 263: Needs a review please. It's not as convoluted anymore but still a little strange thanks.

database/transaction.rs
```diff
@@ -15,9 +16,9 @@ use super::Database;
 
 pub struct TransactionRead<D> {
     database: Arc<Database<D>>,
-    pub(crate) snapshot: ReadSnapshot<D>,
-    pub(crate) type_manager: Arc<TypeManager<ReadSnapshot<D>>>,
-    pub(crate) thing_manager: ThingManager<ReadSnapshot<D>>,
+    pub snapshot: ReadSnapshot<D>,
+    pub type_manager: Arc<TypeManager<ReadSnapshot<D>>>,
+    pub thing_manager: ThingManager<ReadSnapshot<D>>,
```
krishnangovindraj, line 21: Made these pub, replaced the getters.

dependencies/vaticle/repositories.bzl
```diff
@@ -40,8 +40,12 @@ def vaticle_typedb_protocol():
     )
 
 def vaticle_typedb_behaviour():
-    git_repository(
+    native.local_repository(
         name = "vaticle_typedb_behaviour",
-        remote = "https://github.com/vaticle/typedb-behaviour",
-        commit = "96a8bb5b49f809ff17952c079219363f663787f4" # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
+        path = "../typedb-behaviour",
```
krishnangovindraj, line 45: oops

encoding/layout/infix.rs
```diff
@@ -72,6 +73,7 @@ impl Infix {
         PropertyLabel => [0];
         PropertyValueType => [1];
         PropertyOrdering => [2];
+        PropertyOverride => [3];
```
krishnangovindraj, line 76: I guess this is fine?

encoding/layout/infix.rs
```diff
@@ -72,6 +73,7 @@ impl Infix {
...
 
        // Reserve: range 50 - 99 to store annotations with a value type - see InfixID::<CONSTANTS>
         PropertyAnnotationAbstract => [50];
```
krishnangovindraj, line 79: Question here. If we're reserving ranges, should we reserve from 50 or, say `0x40` ?
I've mostly seen these written as hex literals and printed as hex so you can tell the type from the hex code.

tests/behaviour/steps/concept/type_/relation_type.rs
```diff
@@ -0,0 +1,69 @@
...
+        debug_assert!(overridden_relates_opt.is_some(), "Ah, get_role_transitive returned None");
+        }
+
+        let overridden_relates = relation_supertype
+            .get_relates_role(&mut tx.snapshot, &tx.type_manager, overridden_label.to_typedb().name().as_str()) // MUST get_relates_role_transitive
+            .unwrap().unwrap();
+
+        // TODO: Is it ok to just set supertype here?
+        let res = relates.role().set_supertype(&mut tx.snapshot, &tx.type_manager, overridden_relates.role());
+        may_error.check(&res);
```
krishnangovindraj, line 67: The relates override question.

concept/type_/type_manager.rs
```diff
@@ -617,39 +750,74 @@ impl<Snapshot: WritableSnapshot> TypeManager<Snapshot> {
...
+        &self,
+        snapshot: &mut Snapshot,
+        plays: Plays<'static>,
+        overridden: Plays<'static>,
+    ) {
+        let property_key =
+            build_property_type_edge_override(plays.into_type_edge()).into_storage_key().into_owned_array();
+        let overridden_plays = ByteArray::copy(overridden.into_type_edge().into_bytes().bytes());
+        snapshot.put_val(property_key, overridden_plays);
+    }
```
krishnangovindraj, line 789: Do we store the edge or just the role/attribute type?

concept/type_/type_reader.rs
```diff
@@ -119,24 +172,53 @@ impl TypeReader {
...
+                    transitive_owns.insert(owns.attribute(), owns.clone());
+                    if let Some(overridden) = Self::get_owns_override(snapshot, owns.clone())? {
+                        overridden_owns.add(overridden.attribute());
+                    }
+                }
+            }
+            current_type = Self::get_supertype(snapshot, current_type.unwrap())?;
+        }
+        Ok(transitive_owns)
+    }
```
krishnangovindraj, line 216: Logic to compute inherited owns. It's a bit of code but the logic should be the same for owns/plays and even relates.

concept/type_/type_reader.rs
```diff
@@ -147,43 +229,119 @@ impl TypeReader {
...
+
+    pub(crate) fn get_plays_override(
+        snapshot: &impl ReadableSnapshot,
+        plays: Plays<'static>,
+    ) -> Result<Option<Plays<'static>>, ConceptReadError> {
+        let override_property_key = build_property_type_edge_override(plays.into_type_edge());
+        snapshot
+            .get_mapped(override_property_key.into_storage_key().as_reference(), |overridden_edge_bytes| {
+                let overridden_edge = new_edge_plays(Bytes::Reference(overridden_edge_bytes));
```
krishnangovindraj, line 267: Should this be replaced with `IntoCanonicalTypeEdge` ?
We may even be able to make the override functions generic by making Plays & Owns readable types (!!!)

krishnangovindraj (commented):
Self review of non test code.

concept/type_/entity_type.rs
```diff
@@ -239,6 +238,14 @@ impl<'a> OwnerAPI<'a> for EntityType<'a> {
         let expected_owns = Owns::new(ObjectType::Entity(self.clone().into_owned()), attribute_type);
         Ok(self.get_owns(snapshot, type_manager)?.contains(&expected_owns).then_some(expected_owns))
     }
+
+    fn get_owns_transitive<'m, Snapshot: ReadableSnapshot>(
```
flyingsilverfin, line 242: Let's make these all take an enum `Transitivity` which is `DECLARED` or `TRANSITIVE` and not have two method types throughout

concept/type_/owns.rs
flyingsilverfin: I think it'll be OK if we return 'this for the getters

concept/type_/plays.rs
flyingsilverfin: ?

concept/type_/relation_type.rs
```diff
@@ -181,14 +184,17 @@ impl<'a> RelationType<'a> {
...
+        annotation: RelationTypeAnnotation,
+    ) {
         match annotation {
             RelationTypeAnnotation::Abstract(_) => {
                 type_manager.storage_delete_annotation_abstract(snapshot, self.clone().into_owned())
             }
         }
     }
 
+    // TODO: krishnan: Make this private? get_relates_role is interface enough.
```
flyingsilverfin, line 197: Check this before merge

concept/type_/type_manager.rs
flyingsilverfin: We just need to store the overridden role type (in general, RHS of the `as`) in the property, so what you've done is good

encoding/layout/infix.rs
flyingsilverfin: I don't that idea but I just find reasoning over decimal numbers (reserved range 50 - 100) much easier than working in hex (extra mental step to think what range that is)

tests/behaviour/concept/type/BUILD
```diff
@@ -25,6 +25,16 @@ rust_test(
     data = ["@vaticle_typedb_behaviour//concept/type:entitytype.feature"],
 )
 
+rust_test(
+    name = "relationtype",
```
flyingsilverfin, line 29: can we make these with underscore, or is there a reason not to?
I'm actually thinking the BDD files should also using undescore... `entity_type.feature`

flyingsilverfin (changes requested):

PR #7066, opened 2024-05-15
Attribute and entity BDD

dmitrii-ubskii:


common/bytes/util.rs
```diff
@@ -53,3 +55,23 @@ impl fmt::Display for BytesError {
...
+                if f.alternate() && (i + 1) % BREAK == 0 {
+                    f.write_char('\n')?;
+                } else if (i + 1) % GROUP == 0 {
+                    f.write_char(' ')?;
+                }
+            }
+        }
+        Ok(())
+    }
+}
```
dmitrii-ubskii, line 77: Formats as ```
5b21 5b54 7970 6544 425d 282e 2f64 6f63
732f 6261 6e6e 6572 2e70 6e67 295d 2868
7474 7073 3a2f 2f74 7970 6564 622e 636f
```
Line breaks are only included in pretty print (`{:#?}`)

File: tests/behaviour/steps/transaction_context.rs [hunk 2]
```diff
@@ -26,7 +26,7 @@ impl fmt::Debug for ActiveTransaction {
 
 macro_rules! with_read_tx {
     ($context:ident, |$tx:ident| $expr:expr) => {
-        match $context.transaction().unwrap() {
+        match $context.active_transaction.as_ref().unwrap() {
```
dmitrii-ubskii, line 29: Splitting borrow so we can access other fields of `context` within this macro

PR #7074, opened 2024-05-29
Add development-mode for local and CI set ups

farost:
### development-mode
We add a new server's config parameter called `development-mode.enable`. This parameter can be used to separate real "release" application runs from our testing.

In the future, this parameter can influence many different sides of the server (logging settings, networking behavior?.., anything we want), but the only current purpose of this field is to **completely** turn off diagnostics and error reporting:
* sentry;
* diagnostics.

`development-mode.enable` is an optional parameter, and the absence of it in the config is equivalent to `development-mode.enable=false`. It is expected to be set to `true` for all the `config.yml` files used by local environments, tests, and snapshots.

### --//server:config=[development/release] Bazel flag
We add a new Bazel build flag that helps choose the server configuration for all the build jobs, addressing the newly added `development-mode.enable` option.
 
Example of a release configuration build:
```
bazel build --//server:config=release //server:assemble-mac-x86_64-zip
```
Example of a development configuration build:
```
bazel build //server:assemble-mac-x86_64-zip
bazel build --//server:config=development //server:assemble-mac-x86_64-zip
```

### deployment-id
We remove `deployment-id` from the `typedb` config as it makes sense only for `Cloud` servers. Suppose it was set in anyone's config before. In that case, the `deployment-id` value will be forcefully set to `server-id`, changing its previous manually set value. 

It means that all the previous versions of `TypeDB Core` diagnostics submissions with different `server-id` and `deployment-id` (not expected to be higher than 0 - 10 servers) will possibly want to change the `deployment-id` in the database to avoid misleading query results.

### development-mode
We add new `Optional` `KeyValue`s for `YAMLParser`, acting exactly like `Predefined` config options, but letting the configuration miss the key completely, setting its values to defaults inside the code. This way, declaring 
```
KeyValue.Optional<CoreConfig.DevelopmentMode> developmentMode = optional(...);
``` 
lets us set `development-mode.enable` to `true` this way:
```
vaticle-factory: # any other option
  enable: false # any value

development-mode:
  enable: true 
# end of config
```
and set `development-mode.enable` to `false` in these two ways:
```
vaticle-factory: # any other option
  enable: false # any value

development-mode:
  enable: false 
# end of config
```
```
vaticle-factory: # any other option
  enable: false # any value

# end of config
```

This flag is still available for CLI manual overrides. 

For `TypeDBCoreRunner` uses, this flag is forcefully set to `true`, not respecting the provided `config.yml`. However, it's possible to override this flag by passing it in the constructor if the runner user really wants to do it. 

### --//server:config=[development/release] Bazel flag
This flag affects all the targets that depend on the configuration file, including transitive dependencies (e.g. creating an archive of the server's library). 
It only accepts two values. Providing a wrong value leads to an explicit and specific build error.

**NOTE: This is a Bazel flag, so it needs to be specified before the `--` for application flags!**
For example, this works:
```
bazel run --//server:config=release --define version=$(cat VERSION) //:deploy-linux-x86_64-targz -- release
```
This does not:
```
bazel run --define version=$(cat VERSION) //:deploy-linux-x86_64-targz -- release --//server:config=release
```


server/parameters/CoreConfig.java
```diff
@@ -364,7 +378,7 @@ else if (!outputs.get(outputName).isFile()) {
                     output = outputs.get(outputName).asFile();
                 }
 
-                public boolean isEnabled() {
```
farost, line 367: We have `enable` in the config entries, which has a natural usage on that level (e.g. `diagnostics.reporting.enable = true`), but we had multiple variations of `getter`s in the `Java` code as `enable` is a strange name for a property, and adjectives `isEnabled` and `enabled` are more suitable forms for it. I decided to just use `enabled` everywhere.

server/parameters/util/YAMLParser.java
```diff
@@ -64,20 +69,27 @@ String description() {
 
         public abstract Help help(String path);
 
-        public static class Predefined<TYPE> extends KeyValue {
-
+        public static abstract class Static extends KeyValue {
```
farost, line 72: We already have `Dynamic` `KeyValue`s with dynamic keys, now I added `Static` `KeyValue`s with static keys (there is a strict set of acceptable keys, we can't introduce new keys without modifying the code). 

`Static` has two subclasses:
* `Predefined` (required for declaration) - already existed;
* `Optional` (not required for declaration) - a new one for `development-mode`.

BUILD
farost, line 1: There are many changes because of the reordering done by my Bazel plugin that started being pretty aggressive after switching to the new IDEA. I fought it in many places but decided that it could be useful for others to have this file formatted earlier in the end, but it may cause conflicts with merges. Feel free to ask me to reapply my small changes without this reformatting.

BUILD
```diff
@@ -31,27 +36,27 @@ native_java_libraries(
...
         # Vaticle Dependencies
         "@vaticle_typeql//common/java:common",
     ],
-    tags = ["maven_coordinates=com.vaticle.typedb:typedb:{pom_version}"],
-    visibility = ["//visibility:public"],
 )
 
 assemble_files = {
-    "//server:config": "server/conf/config.yml",
+    "//server/parameters/config": "server/conf/config.yml",
```
farost, line 48: This is the main change.

common/diagnostics/Diagnostics.java
```diff
@@ -88,40 +88,60 @@ protected Core(Metrics metrics, StatisticReporter statisticReporter, MonitoringS
...
-        protected static void initSentry(String serverID, String distributionName, String version, boolean errorReportingEnable, String errorReportingURI) {
+        protected static void initSentry(
+                boolean developmentModeEnabled,
+                String serverID,
+                String distributionName,
+                String version,
+                boolean errorReportingEnabled,
+                String errorReportingURI
+        ) {
+            boolean sentryEnabled = !developmentModeEnabled && errorReportingEnabled;
```
farost, line 141: I was thinking about not even setting usernames, etc for Sentry, just like we do for `Noop`, but decided to just combine these two flags here.

server/BUILD
farost, line 1: Same comment for auto reformatting.

server/parameters/CoreConfigParser.java
```diff
@@ -131,12 +148,12 @@ protected static class Stdout extends Compound<CoreConfig.Common.Output.Type.Std
                     );
                     private static final Predefined<Boolean> enable =
                             predefined("enable", "Enable logging to stdout.", BOOLEAN);
-                    private static final Set<Predefined<?>> parsers = set(typeParser, enable);
+                    private static final Set<Static> parsers = set(typeParser, enable);
```
farost, line 151: Static <- {Predefined, Optional} -- the names of the keys are predefined
Dynamic -- the names of the keys are NOT predefined

test/assembly/DockerTest.java
```diff
@@ -40,7 +40,7 @@ public void bootup() throws InterruptedException, TimeoutException, IOException
                 "docker", "run", "--name", "typedb",
                 "--rm", "-t", "-p", String.format("%d:%d", typeDBPort, typeDBPort),
                 "bazel:assemble-docker",
-                "/opt/typedb-all-linux-x86_64/typedb", "server", "--diagnostics.reporting.errors=false", "--diagnostics.reporting.statistics=false"
+                "/opt/typedb-all-linux-x86_64/typedb", "server"
```
farost, line 43: The server is in release mode only if this test is run with `--//server:config=release`. What if we would want to...

test/assembly/DockerTest.java
```diff
@@ -40,7 +40,7 @@ public void bootup() throws InterruptedException, TimeoutException, IOException
                 "docker", "run", "--name", "typedb",
                 "--rm", "-t", "-p", String.format("%d:%d", typeDBPort, typeDBPort),
                 "bazel:assemble-docker",
-                "/opt/typedb-all-linux-x86_64/typedb", "server", "--diagnostics.reporting.errors=false", "--diagnostics.reporting.statistics=false"
+                "/opt/typedb-all-linux-x86_64/typedb", "server", "--development-mode.enable=true"
```
farost, line 43: We run `DockerTest` and `AptTest` only with bazel, so it will be in the `development-mode` until we give it the `--//server:config=release` flag. I thought about leaving this opportunity, but I am not comfortable that somebody would like to adapt these tests to non-Bazel usage and, while the config is not explicitly set here, forget about this mode. 

In the end, I am indifferent to this flag being set inside the code for these two tests. I'm ready to remove it if you want to.

BUILD
flyingsilverfin: It's fine, no problem

farost:
I'd still think about a `correctness` job to check that the only difference between `config_release` and `config_development` are the `development-mode` options. Not sure if we want to add it here, not sure if we'll add it later if we don't right now...

tool/runner/TypeDBCoreRunner.java
```diff
@@ -32,17 +38,24 @@ public class TypeDBCoreRunner implements TypeDBRunner {
...
     }
 
-    public TypeDBCoreRunner(Map<String, String> options) throws InterruptedException, TimeoutException, IOException {
-        this.options = options;
-        port = findUnusedPorts(1).get(0);
+    public TypeDBCoreRunner(Map<String, String> userOptions) throws InterruptedException, TimeoutException, IOException {
+        this.userOptions = userOptions;
+        List<Integer> unusedPorts = findUnusedPorts(2);
+        port = unusedPorts.get(0);
+        diagnosticsMonitoringPort = unusedPorts.get(1);
```
farost, line 58: I add `diagnostics.monitoring.port` generation for parallel `TypeDBCoreRunner` runs, otherwise, they would conflict with each other just like they did in `Cloud` if we didn't turn `diagnostics.monitoring` off.

tool/runner/CoreServerOpts.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+package com.vaticle.typedb.core.tool.runner;
+
+public class CoreServerOpts {
```
farost, line 9: I migrated some of the flags from `Cloud` for better structure.

server/parameters/util/YAMLParser.java
flyingsilverfin: I think the use of `static` makes sense, good catch.

server/parameters/config/BUILD
```diff
@@ -0,0 +1,28 @@
...
+    }),
+    visibility = ["//visibility:public"],
+)
+
+genrule(
+    name = "config",
+    srcs = [":config_file"],
+    outs = ["config.yml"],
+    cmd = "cp $(SRCS) $@",
+    visibility = ["//visibility:public"],
```
flyingsilverfin, line 21: you'll need to make a `cmd_bat` addition here so that this can run on windows I think

test/assembly/DockerTest.java
flyingsilverfin: I think this is fine

flyingsilverfin (approved):

server/parameters/config/BUILD
farost: Good catch, I didn't notice that we don't use already existing similar genrules for Windows.

PR #7083, opened 2024-06-06
Implement structs

krishnangovindraj:
## Usage and product changes
We implement structs.  A struct is a named value-type defined by a number of (possibly optional) named fields, each of which has a specified value-type - including other structs.  These structs can be referenced by name when used as the value-type of an attribute. 

## Implementation
* Implement StructDefinition.
* Implement StructValue, & moves StructValue to the encoding package, where all other custom value types are also defined.
* Introduce StructIndexEntry - For looking up structs given the value of a field. These were encoded as follows:


encoding/graph/common/value_hasher.rs
krishnangovindraj, line 1: Single implementation of disambiguated hashes because I've had to do this in 3 places.

encoding/graph/thing/vertex_generator.rs
```diff
@@ -326,12 +331,13 @@ impl ThingVertexGenerator {
...
     {
-        if string.length() <= StringAttributeID::ENCODING_INLINE_CAPACITY {
+        if StringAttributeID::is_inlineable(string.as_reference()) {
             Ok(StringAttributeID::build_inline_id(string))
         } else {
             let id = StringAttributeID::build_hashed_id(type_id, string, snapshot, &self.large_value_hasher)?;
             let hash = id.get_hash_prefix_hash();
-            let lock = ByteArray::copy_concat([&type_id.bytes(), &hash]);
+            let lock =
+                ByteArray::copy_concat([&Prefix::VertexAttributeString.prefix_id().bytes(), &type_id.bytes(), &hash]);
```
krishnangovindraj, line 340: Why not just lock the whole ID?

encoding/value/value_struct.rs
krishnangovindraj, line 1: There were too many signs to move the struct value into encoding.
I would also like to move Value into encoding, because I've had to define a new FieldValue enum here.

concept/tests/test_thing.rs
```diff
@@ -655,3 +739,203 @@ fn role_player_duplicates() {
         assert_eq!(resource_1_indexed_count, 2);
     }
 }
+
```
krishnangovindraj, line 742: TODO: More test coverage.

krishnangovindraj (commented):
self review

concept/tests/test_thing.rs
```diff
@@ -655,3 +739,203 @@ fn role_player_duplicates() {
...
+        HashMap::from([("f0l".to_owned(), (ValueType::Long, false)), ("f1s".to_owned(), (ValueType::String, false))]);
+    let definition = StructDefinition::define(struct_name.to_owned(), fields);
+
+    let instance_fields = HashMap::from([
+        ("f0l".to_owned(), FieldValue::Long(123)),
+        ("f1s".to_owned(), FieldValue::String(Cow::Owned("abc".to_owned()))),
+    ]);
+    let struct_key = {
+        let mut snapshot: WriteSnapshot<WALClient> = storage.clone().open_snapshot_write();
+        let struct_key = type_manager.create_struct(&mut snapshot, definition.clone()).unwrap();
```
flyingsilverfin, line 815: Should be it be `definition.clone()` or `&definition`? Does it need ownership? Probably serialising, so it may not?

encoding/graph/definition/struct.rs
```diff
@@ -6,19 +6,56 @@
...
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
 pub struct StructDefinitionField {
-    optional: bool,
-    // value_type: ValueType, // TODO
-    index: StructFieldIDUInt,
+    pub index: StructFieldIDUInt,
+    pub optional: bool,
+    pub value_type: ValueType, // TODO
```
flyingsilverfin, line 35: Can you check for TODOs that aren't meaningful please

concept/tests/test_thing.rs
```diff
@@ -655,3 +739,203 @@ fn role_player_duplicates() {
...
+        snapshot.commit().unwrap();
+        attr_instance
+    };
+
+    {
+        let mut snapshot: WriteSnapshot<WALClient> = storage.clone().open_snapshot_write();
+        let attr_type = type_manager.get_attribute_type(&snapshot, &attr_label).unwrap().unwrap();
+        let attr_vec = thing_manager.get_attributes_in(&snapshot, attr_type.clone()).unwrap().collect_cloned();
+        let mut attr = attr_vec.get(0).unwrap().clone();
+        let value_0 = attr.get_value(&snapshot, &thing_manager).unwrap();
```
flyingsilverfin, line 852: Excellent, looks nice and consistent!

concept/tests/test_thing.rs
```diff
@@ -655,3 +739,203 @@ fn role_player_duplicates() {
...
+                .unwrap();
+            attrs.push(attr);
+        }
+
+        for (val, attr) in std::iter::zip(field_values, attrs.iter()) {
+            let mut attr_by_field_iterator = thing_manager
+                .get_attributes_by_struct_field(
+                    &snapshot,
+                    attr_type.clone(),
+                    vec![0],
```
flyingsilverfin, line 928: Can we get this ID `0` (presumably this is the field ID?) from the definition instead of hardcoding it?

concept/type_/type_manager.rs
```diff
@@ -278,6 +290,30 @@ impl<Snapshot: ReadableSnapshot> TypeManager<Snapshot> {
...
+        &self,
+        snapshot: &Snapshot,
+        name: &str,
+    ) -> Result<Option<DefinitionKey<'static>>, ConceptReadError> {
+        if let Some(cache) = &self.type_cache {
+            Ok(cache.get_struct_definition_key(name))
+        } else {
+            TypeReader::get_struct_definition_key(snapshot, name)
+        }
+    }
```
flyingsilverfin, line 303: Yess

flyingsilverfin (changes requested):

concept/type_/type_manager/type_cache/struct_definition_cache.rs
```diff
@@ -0,0 +1,37 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use bytes::Bytes;
+use encoding::graph::definition::{definition_key::DefinitionKey, r#struct::StructDefinition, DefinitionValueEncoding};
+use storage::{key_range::KeyRange, snapshot::ReadableSnapshot};
+
+pub struct StructDefinitionCache {
```
flyingsilverfin, line 11: Right now we refer to a "cache" as generally a collection of many cache entries, not the individual entries this are cached. Perhaps we can find a different name for this? `StructDefinitionCacheEntry`? 

Then the Impl block doesn't make sense anymore, but you can have a standalone function that creates the cache (Box<Option<[Entry]>>)

concept/type_/type_manager/type_writer.rs
```diff
@@ -30,12 +34,30 @@ pub struct TypeWriter<Snapshot: WritableSnapshot> {
...
+        let index_key = LabelToStructDefinitionIndex::build::<BUFFER_KEY_INLINE>(StringBytes::build_ref(
+            struct_definition.name.as_str(),
+        ));
+        snapshot.put_val(
+            index_key.into_storage_key().into_owned_array(),
+            ByteArray::copy(definition_key.clone().into_bytes().bytes()),
+        );
+        snapshot.put_val(
+            definition_key.into_storage_key().into_owned_array(),
+            struct_definition.to_bytes().unwrap().into_array(),
```
flyingsilverfin, line 51: Shouldn't this be `into_bytes()`?

concept/type_/type_manager/type_writer.rs
```diff
@@ -30,12 +34,30 @@ pub struct TypeWriter<Snapshot: WritableSnapshot> {
...
+        snapshot: &mut Snapshot,
+        definition_key: DefinitionKey<'static>,
+        struct_definition: StructDefinition,
+    ) {
+        let index_key = LabelToStructDefinitionIndex::build::<BUFFER_KEY_INLINE>(StringBytes::build_ref(
+            struct_definition.name.as_str(),
+        ));
+        snapshot.put_val(
+            index_key.into_storage_key().into_owned_array(),
+            ByteArray::copy(definition_key.clone().into_bytes().bytes()),
```
flyingsilverfin, line 47: If we're converting an already Cloned structure `into_bytes()`, then why do we need to induce another `ByteArray::copy`? Feels like one copy here is redundant.

encoding/graph/common/schema_id_allocator.rs
```diff
@@ -0,0 +1,139 @@
...
+        Typed,
+    },
+    layout::prefix::Prefix,
+    Keyable,
+};
+
+pub type TypeVertexAllocator = SchemaIDAllocator<TypeVertex<'static>>;
+pub type DefinitionKeyAllocator = SchemaIDAllocator<DefinitionKey<'static>>;
+
+pub trait AllocationHelper: Sized {
```
flyingsilverfin, line 35: Err this name is very strange!

concept/thing/attribute.rs
```diff
@@ -180,15 +183,45 @@ impl<'a> Ord for Attribute<'a> {
 }
 
 /// Attribute iterators handle hiding dependent attributes that were not deleted yet
-pub struct AttributeIterator<'a, Snapshot: ReadableSnapshot> {
+trait ExtractAttributeFromKey {
```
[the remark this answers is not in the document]
krishnangovindraj: Yes, but that's to go from an encoded attribute to a rust attribute and vice-versa.
This is to extract it from some key that contains it.

I don't find an occurrence of `Reader` that means anything like this in the codebase.

encoding/graph/common/schema_id_allocator.rs
```diff
@@ -0,0 +1,139 @@
...
+                Some(allocated_id) => {
+                    self.last_allocated_type_id.store(allocated_id.clone(), Relaxed);
+                    Ok(T::object_from_id(self.prefix, allocated_id))
+                }
+            }
+        }
+    }
+}
+
+impl AllocationHelper for TypeVertex<'static> {
```
flyingsilverfin, line 98: I think this is approaching correct, but I'm not sure it's there yet.
My first instinct is to call this `impl SchemaID for TypeVertex`. 
However, now it makes me think of the overlap with `TypeID`... is `TypeID` redundant? Somehow these are now very similar ideas.
And secondly, calling it `SchemaID` lets us move the implementation into `TypeVertex`, which is I think where this belongs (and below in DefinitionKey).

concept/thing/thing_manager.rs
```diff
@@ -358,6 +392,51 @@ impl<Snapshot: ReadableSnapshot> ThingManager<Snapshot> {
         )
     }
 
+    pub fn get_attributes_by_struct_field<'this, 'a, 'v>(
+        &'this self,
+        snapshot: &'this Snapshot,
+        attribute_type: AttributeType<'a>,
+        path_to_field: Vec<StructFieldIDUInt>,
+        value: FieldValue<'v>,
```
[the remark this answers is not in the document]
krishnangovindraj: Yeah could be. It'll just be a wrapper around `Vec<StructFieldIDUInt>` though.

encoding/graph/definition/definition_key_property.rs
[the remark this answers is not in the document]
flyingsilverfin: Delete

concept/thing/thing_manager.rs
```diff
@@ -358,6 +392,51 @@ impl<Snapshot: ReadableSnapshot> ThingManager<Snapshot> {
...
+    ) -> Result<StructIndexToAttributeIterator<'_, Snapshot>, ConceptReadError> {
+        // TODO: Should I assert that the attribute-type is a struct?
+        // Surely all that validation is done if traversal is the only way to access these.
+        debug_assert!({
+            let value_type = attribute_type.get_value_type(snapshot, &self.type_manager).unwrap().unwrap();
+            value_type.category() == ValueTypeCategory::Struct
+        });
+
+        let prefix = StructIndexEntry::build_search_key(
```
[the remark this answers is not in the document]
krishnangovindraj: The whole thing is a key. The search key is without the `StructAttributeID`

encoding/graph/thing/vertex_attribute.rs
```diff
@@ -809,19 +770,92 @@ impl StringAttributeID {
...
+    pub fn get_hash_hash(&self) -> [u8; Self::ENCODING_HASH_LENGTH] {
+        self.bytes[0..Self::ENCODING_HASH_LENGTH].try_into().unwrap()
+    }
+
+    pub fn get_hash_disambiguator(&self) -> u8 {
+        self.bytes[Self::ENCODING_STRUCT_TAIL_INDEX] & !Self::HASHID_DISAMBIGUATOR_BYTE_IS_HASH_FLAG
+    }
+}
+
+impl DisambiguatingHashedID<{ StructAttributeID::ENCODING_HASH_LENGTH + 1 }> for StructAttributeID {
```
flyingsilverfin, line 858: To be honest I think we can always just drop the `Disambiguated` part of the names, since we only have hash _and_ disambiguated combined, and I think Hashed is shorter + also telling

encoding/graph/thing/vertex_generator.rs
```diff
@@ -125,6 +126,10 @@ impl ThingVertexGenerator {
         Ok(ThingVertexGenerator { entity_ids, relation_ids, large_value_hasher })
     }
 
+    pub fn TEMP__hasher(&self) -> &impl Fn(&[u8]) -> u64 {
+        &self.large_value_hasher
+    }
+
```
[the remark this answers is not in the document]
flyingsilverfin: Yeah you can expose it through here as `pub(crate)` I think!

encoding/graph/thing/vertex_generator.rs
flyingsilverfin: HMMM I think I have done this because two concurrent transactions can end up hashing multiple values (by pure bad luck) into the same hash value, which will lead transaction 1 to use disambiguation byte = `0`, and the other one to use disambiguation byte `= 1` and we won't detect the conflict... Like this we force things that collide on the hash to fail (which should be extremely rarely anyway).

encoding/graph/type_/index.rs
```diff
@@ -59,10 +64,21 @@ impl<'a> AsBytes<'a, BUFFER_KEY_INLINE> for LabelToTypeVertexIndex<'a> {
...
         EncodingKeyspace::Schema
     }
 }
 
-impl<'a> Prefixed<'a, BUFFER_KEY_INLINE> for LabelToTypeVertexIndex<'a> {}
+impl<'a, T: IndexedType> Prefixed<'a, BUFFER_KEY_INLINE> for IdentifierToTypeIndex<'a, T> {}
+
+// Specialisations
+pub type LabelToTypeVertexIndex<'a> = IdentifierToTypeIndex<'a, TypeVertex<'static>>;
+impl<'a> IndexedType for TypeVertex<'a> {
```
flyingsilverfin, line 77: Maybe we call this `impl IdentifierIndex for TypeVertex`

encoding/layout/prefix.rs
```diff
@@ -208,9 +212,12 @@ impl Prefix {
            PropertyTypeVertex => [160], false;
            PropertyTypeEdge => [162], false;
            PropertyObjectVertex => [163], false;
+           PropertyDefinition => [164], false; // TODO: Why are properties not fixed width keys?
```
flyingsilverfin, line 215: Good point, they probably are?

resource/constants.rs
```diff
@@ -18,14 +18,15 @@ pub mod storage {
...
 
     pub const LABEL_NAME_STRING_INLINE: usize = 64;
     pub const LABEL_SCOPE_STRING_INLINE: usize = 64;
     pub const LABEL_SCOPED_NAME_STRING_INLINE: usize = LABEL_NAME_STRING_INLINE + LABEL_SCOPE_STRING_INLINE;
+    pub const DEFINITION_NAME_STRING_INLINE: usize = 64;
 
-    pub type DefinitionIDUInt = u32;
-    pub type DefinitionIDAtomicUInt = AtomicU32;
+    pub type DefinitionIDUInt = u16;
+    pub type DefinitionIDAtomicUInt = AtomicU16;
```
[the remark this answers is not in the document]
flyingsilverfin: 65k is fine :)

concept/thing/thing_manager.rs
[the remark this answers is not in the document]
flyingsilverfin: Hmm ok I'm just pointing out that the term "search" is random and new, im sure we have a term for it already?

krishnangovindraj (commented):
replies

encoding/graph/thing/vertex_generator.rs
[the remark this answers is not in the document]
krishnangovindraj: ^ This is better if `struct_value` can stay in encoding

concept/tests/test_thing.rs
```diff
@@ -655,3 +741,283 @@ fn role_player_duplicates() {
...
+                )
+                .unwrap();
+            let mut attr_by_field_iterator = thing_manager
+                .get_attributes_by_struct_field(
+                    &snapshot,
+                    attr_type.clone(),
+                    field_path,
+                    FieldValue::String(Cow::Borrowed(val)),
+                )
+                .unwrap();
```
krishnangovindraj, line 935: Marker for `resolve_struct_field` usage in a different comment.

krishnangovindraj (commented):
ready for next round.

concept/tests/test_thing.rs
```diff
@@ -655,3 +741,283 @@ fn role_player_duplicates() {
...
+            );
+            let attr = thing_manager
+                .create_attribute(&mut snapshot, attr_type.clone(), Value::Struct(Cow::Borrowed(&outer_struct_value)))
+                .unwrap();
+            attrs.push(attr);
+        }
+
+        for (val, attr) in std::iter::zip(field_values, attrs.iter()) {
+            let field_path = type_manager
+                .resolve_struct_field(
```
flyingsilverfin, line 922: I think this should not take a vec of owned slices, it should take a vec of slices (or even a slice of slices?) - conceptually, we should not need to take ownership to produce the output data type.

concept/tests/test_thing.rs
```diff
@@ -655,3 +741,283 @@ fn role_player_duplicates() {
...
+            .unwrap()
+            .unwrap();
+        assert_eq!(attr, attr_by_id);
+        assert_eq!(attr_created, attr);
+        snapshot.close_resources();
+    }
+}
+
+#[test]
+fn read_struct_by_field() {
```
flyingsilverfin, line 869: rename to to `read_attribute_struct...`

flyingsilverfin: And all these tests should contain attribute to tie them to "concept"s

PR #7094, opened 2024-07-02
Add operation time validations for schema. Finish inheritance, overrides, ordering.

farost:


concept/error.rs
```diff
@@ -107,6 +112,10 @@ pub enum ConceptWriteError {
 
     SetHasOrderedOwnsUnordered {},
     SetHasUnorderedOwnsOrdered {},
+
+    Conversion {
```
farost, line 116: I return this error from `Into<Annotation>` conversions of `PlaysAnnotation`, etc. Open for better naming. Thought about just `Internal` stuff, but `ConceptWriteError` is already `Internal` enough.

concept/tests/test_thing.rs
farost, line 1: I added checks for `Ordering`, so some of the `thing` tests already broke.

farost: There are still 2 tests failing, but it should not be because of my changes, looks like it was not just fully implemented yet (one of the failures is specific `todo!()`

concept/thing/thing_manager.rs
```diff
@@ -772,7 +772,7 @@ impl<'txn, Snapshot: WritableSnapshot> ThingManager<Snapshot> {
         {
             let owner = Object::new(ObjectVertex::new(Bytes::reference(key.bytes())));
             let owner_type = owner.type_();
-            for owns in &owner_type.get_owns(snapshot, self.type_manager())? {
+            for owns in &owner_type.get_owns_declared(snapshot, self.type_manager())? {
```
farost, line 775: Wwe previously had `get_owns` / `get_owns_transitive`

concept/type_/annotation.rs
```diff
@@ -28,6 +30,11 @@ pub enum Annotation {
     Key(AnnotationKey),
     Cardinality(AnnotationCardinality),
     Regex(AnnotationRegex),
+    Cascade(AnnotationCascade),
+    // TODO: Subkey
+    // TODO: Values
+    // TODO: Range
+    // TODO: Replace
```
farost, line 37: Might remove

concept/type_/annotation.rs
```diff
@@ -82,11 +105,118 @@ impl AnnotationRegex {
...
+    Cascade,
+    // TODO: Subkey
+    // TODO: Values
+    // TODO: Range
+    // TODO: Replace
+}
+
+impl AnnotationCategory {
+    pub fn to_default_annotation(&self) -> Annotation {
+        // TODO: fn const for regex default
```
farost, line 153: Wanted to make this function const, but could not make `AnnotationRegex::default()` const because of the `String`s things. Might go back to try again

concept/type_/annotation.rs
```diff
@@ -82,11 +105,118 @@ impl AnnotationRegex {
...
+            AnnotationCategory::Independent => Annotation::Independent(AnnotationIndependent),
+            AnnotationCategory::Unique => Annotation::Unique(AnnotationUnique),
+            AnnotationCategory::Key => Annotation::Key(AnnotationKey),
+            AnnotationCategory::Cardinality => Annotation::Cardinality(AnnotationCardinality::default()),
+            AnnotationCategory::Regex => Annotation::Regex(AnnotationRegex::default()),
+            AnnotationCategory::Cascade => Annotation::Cascade(AnnotationCascade),
+        }
+    }
+
+    pub fn compatible_to_declare_together(&self, other: AnnotationCategory) -> bool {
```
farost, line 166: Decided to move annotation checks here, as it feels strange to have **that** many annotation validations outside of annotations. Plus it complicated naming in many cases. 

This function is to check that two annotations can be declared together on one *thing*, being neighbors.

concept/type_/annotation.rs
```diff
@@ -82,11 +105,118 @@ impl AnnotationRegex {
...
+            },
+            | AnnotationCategory::Abstract
+            | AnnotationCategory::Distinct
+            | AnnotationCategory::Independent
+            | AnnotationCategory::Regex
+            | AnnotationCategory::Cascade => true,
+        }
+    }
+
+    pub fn compatible_to_transitively_add(&self, other_to_add: AnnotationCategory) -> bool {
```
farost, line 188: A more relaxed check for compatibility on different levels. Used in two cases:
* When I want to set a new annotation and need to check if it is compatible with already transitively added annotations (if we already have `Key`, we can't set `Unique` or `Cardinality` on subtypes).
* When we collect stored transitive annotations for a type starting from its declared annotations (we already have declared `Key` on a subtype, so we don't need to take `Unique` or `Cardinality` from a supertype). We consider that all the stored annotations are correct (thanks to `compatible_to_declare_together`).

I didn't come up with a better name to specify that `&self` should be a part of a collection of already processed annotations where want to add `other_to_add` to. Feel free to suggest!

concept/type_/annotation.rs
```diff
@@ -168,3 +300,16 @@ impl<'a> TypeEdgePropertyEncoding<'a> for AnnotationCardinality {
         Some(Bytes::copy(bincode::serialize(&self).unwrap().as_slice()))
     }
 }
+
+impl<'a> TypeEdgePropertyEncoding<'a> for AnnotationRegex {
+    const INFIX: Infix = Infix::PropertyAnnotationRegex;
+    fn from_value_bytes<'b>(value: ByteReference<'b>) -> Self {
+        // TODO this .unwrap() should be handled as an error
```
farost, line 307: I do feel like we have a kind of misunderstanding of when we can use `unwrap()` and where return accurate errors (based on talks with Dmitrii and Krishnan), but I don't feel like we have time to address it right now. Just keep it in mind for the future refactoring

concept/type_/attribute_type.rs
```diff
@@ -275,17 +296,48 @@ pub enum AttributeTypeAnnotation {
...
-            Annotation::Regex(annotation) => AttributeTypeAnnotation::Regex(annotation),
-
-            Annotation::Distinct(_) => unreachable!("Distinct annotation not available for Attribute type."),
-            Annotation::Unique(_) => unreachable!("Unique annotation not available for Attribute type."),
-            Annotation::Key(_) => unreachable!("Key annotation not available for Attribute type."),
-            Annotation::Cardinality(_) => unreachable!("Cardinality annotation not available for Attribute type."),
+        let into_annotation: Result<AttributeTypeAnnotation, ConversionError> = annotation.into();
+        match into_annotation {
+            Ok(into_annotation) => into_annotation,
+            Err(_) => unreachable!("Do not call this conversion from user-exposed code!"),
```
farost, line 330: The ugliest part of all these annotations conversions =) ( =( )

concept/type_/mod.rs
```diff
@@ -77,8 +77,8 @@ pub trait TypeAPI<'a>: ConceptAPI<'a> + TypeVertexEncoding<'a> + Sized + Clone +
     ) -> Result<MaybeOwns<'m, Label<'static>>, ConceptReadError>;
 }
 
-pub trait KindAPI<'a>: TypeAPI<'a> + PrefixedTypeVertexEncoding<'a> {
```
farost, line 80: Confirmed with Krishnan that `PrefixedTypeVertexEncoding` is not needed here

concept/type_/mod.rs
```diff
@@ -77,8 +77,8 @@ pub trait TypeAPI<'a>: ConceptAPI<'a> + TypeVertexEncoding<'a> + Sized + Clone +
...
-    type AnnotationType: Hash + Eq + From<Annotation>;
+pub trait KindAPI<'a>: TypeAPI<'a> {
+    type AnnotationType: Hash + Eq + Clone + From<Annotation> + Into<Annotation>;
```
farost, line 81: I do feel like we want a similar `KindAPI` for edges (relates, owns, plays), but I couldn't understand what a good term for it. So you'll see a ton of `EDGE: TypeEdgeEncoding<'static> + InterfaceImplementation<'static> + Clone` later

farost: The downside of `KindAPI::AnnotationType` is that our normal conversions of annotations return `Annotation` with a full list of possible `enum Annotation` instead of like `PlaysAnnotation`, etc. And it doesn't make the logic really smooth. But whatever

storage/snapshot/iterator.rs
```diff
@@ -92,7 +92,10 @@ impl SnapshotRangeIterator {
                         self.storage_iterator.next();
                         self.buffered_iterator.next();
                     } else {
-                        debug_assert_eq!(storage_value.bytes(), buffered_write.get_value().bytes());
+                        #[cfg(debug_assertions)]
+                        if let Write::Put { value, .. } = buffered_write {
```
farost, line 96: This assert is valid only for `Put`

storage/snapshot/iterator.rs
```diff
@@ -217,8 +222,8 @@ impl LendingIterator for SnapshotRangeIterator {
         }
         match self.ready_item_source.take() {
             Some(ReadyItemSource::Both) => {
```
farost, line 224: If we have the item both in buffer and storage, we need to get the buffered one!

tests/behaviour/concept/thing/BUILD
```diff
@@ -4,6 +4,7 @@
 
 load("@vaticle_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
 load("@rules_rust//rust:defs.bzl", "rust_test")
+# TODO: Add test_ prefix to target names
```
farost, line 7: Haven't done not to have conflicts with the guys working on things, might do now

tests/behaviour/steps/connection/transaction.rs
```diff
@@ -47,10 +47,28 @@ pub async fn transaction_has_type(context: &mut Context, tx_type: String) {
 #[apply(generic_step)]
 #[step(expr = "transaction commits{may_error}")]
 pub async fn transaction_commits(context: &mut Context, may_error: MayError) {
+    // match context.take_transaction().unwrap() {
+    //     ActiveTransaction::Read(_) => {}
+    //     ActiveTransaction::Write(tx) => may_error.check(&tx.commit()),
+    //     ActiveTransaction::Schema(tx) => may_error.check(&tx.commit()),
+    // }
+    // TODO: Temporary implementation until transaction-time validations are implemented
```
farost, line 55: It's for the next PR

concept/type_/type_manager.rs
farost, line 1: Some of the checks are kinda dirty, but it's better to be more expressive here, especially on this initial development stage, so all the bugs and changes are controlled easily. So I'm not going to point out anything specific I don't like here, just feel free to highlight what you want me to reconsider.

concept/type_/type_manager.rs
```diff
@@ -1182,136 +1577,625 @@ impl<Snapshot: WritableSnapshot> TypeManager<Snapshot> {
...
+        owns: Owns<'static>,
+    ) -> Result<(), ConceptWriteError> {
+        OperationTimeValidation::validate_owns_value_type_compatible_to_unique_annotation(
+            snapshot,
+            owns.clone(),
+            TypeReader::get_value_type_without_source(snapshot, owns.attribute().clone())?,
+        )
+        .map_err(|source| ConceptWriteError::SchemaValidation { source })?;
+
+        OperationTimeValidation::validate_set_edge_annotation_is_compatible_with_declared_annotations(
```
farost, line 1959: I call this and the next functions for all `set_annotation`, so we might want to move to a "common" check. Like `Self::set_type_annotation()` and `Self::set_edge_annotation()` with these two functions + `TypeWrite::storage_put***`. 
Decided not to overdo refactoring for now

concept/type_/annotation.rs
flyingsilverfin: It's ok let's keep it

concept/type_/annotation.rs
farost: I like `category.declarable_alongside(other_category)`, but not sure about the second one. For context:
* I use this function (`compatible_to_transitively_add`) both to check that my newly declared annotation is compatible with inherited annotations (if I have inherited `@key`, I cannot declare `@unique` or `@card`) AND to filter out annotations when traversing types from a *subtype* to the root, collecting all the transitive annotations for this *subtype* (if I already collected `@key` from *subtype* (or its parents), I don't need to add `@unique` or `@card` from their parents on top).
* We have `declared` now as a term for annotations that are directly connected to a *thing*, and using this term for transitive compatibility seems a bit confusing
* `above` and `below` do not work here either because it's kinda natural to use this function in both directions on different stages (validation and collection)

farost: I'll leave both functions as is for now to preserve symmetry in their names, will be ready to change both after we finish our disc

concept/type_/annotation.rs
```diff
@@ -59,6 +70,14 @@ impl AnnotationCardinality {
         Self { start_inclusive, end_inclusive }
     }
 
+    pub const fn default() -> Self {
+        Self::new(0, Some(1))
+    }
+
+    pub const fn key() -> Self {
```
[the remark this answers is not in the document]
farost: You're correct.
```
impl AnnotationKey {
    pub const Cardinality: AnnotationCardinality = AnnotationCardinality::new(0, Some(1));
}
```

concept/type_/annotation.rs
```diff
@@ -82,11 +113,118 @@ impl AnnotationRegex {
         Self { regex }
     }
 
+    pub fn default() -> Self {
+        Self::new(".*".to_owned())
+    }
```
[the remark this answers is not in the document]
farost:
```
#[derive(Debug, Default, Clone, Eq, PartialEq, Hash)]
pub struct AnnotationRegex {
    regex: Cow<'static, str>,
}

impl AnnotationRegex {
    pub const fn new(regex: String) -> Self {
        Self { regex: Cow::Owned(regex) }
    }

    pub const fn default() -> Self {
        Self { regex: Cow::Borrowed(".*") }
    }

    pub fn regex(&self) -> &str {
        &self.regex
    }
}
```

concept/type_/attribute_type.rs
```diff
@@ -208,35 +226,38 @@ impl<'a> AttributeType<'a> {
...
         snapshot: &mut Snapshot,
         type_manager: &TypeManager<Snapshot>,
-        annotation: AttributeTypeAnnotation,
-    ) {
-        match annotation {
+        annotation_category: AnnotationCategory,
+    ) -> Result<(), ConceptWriteError> {
+        let relation_type_annotation = AttributeTypeAnnotation::try_getting_default(annotation_category)
+            .map_err(|source| ConceptWriteError::Conversion { source })?;
+        match relation_type_annotation {
```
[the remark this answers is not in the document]
farost: Done, thanks!

concept/type_/entity_type.rs
```diff
@@ -254,16 +274,12 @@ impl<'a> OwnerAPI<'a> for EntityType<'a> {
...
-        snapshot: &Snapshot,
-        type_manager: &'m TypeManager<Snapshot>,
-    ) -> Result<MaybeOwns<'m, HashMap<AttributeType<'static>, Owns<'static>>>, ConceptReadError> {
-        type_manager.get_entity_type_owns_transitive(snapshot, self.clone().into_owned())
+        Ok(self
+            .get_owns(snapshot, type_manager)?
+            .iter()
+            .map(|(_, owns)| owns)
+            .find(|owns| owns.attribute() == attribute_type)
+            .cloned())
```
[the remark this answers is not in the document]
farost: Yep, should be fine, I overdid searches here.

concept/type_/entity_type.rs
```diff
@@ -303,16 +324,12 @@ impl<'a> PlayerAPI<'a> for EntityType<'a> {
...
-        &self,
-        snapshot: &Snapshot,
-        type_manager: &'m TypeManager<Snapshot>,
-    ) -> Result<MaybeOwns<'m, HashMap<RoleType<'static>, Plays<'static>>>, ConceptReadError> {
-        type_manager.get_entity_type_plays_transitive(snapshot, self.clone().into_owned())
+        Ok(self
+            .get_plays(snapshot, type_manager)?
+            .iter()
+            .map(|(_, plays)| plays)
+            .find(|plays| plays.role() == role_type)
```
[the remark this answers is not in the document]
farost: You're correct, thanks, fixed everything.

concept/type_/type_manager.rs
```diff
@@ -772,14 +993,26 @@ impl<Snapshot: WritableSnapshot> TypeManager<Snapshot> {
...
     }
 
+    pub fn delete_struct(
+        &self,
+        snapshot: &mut Snapshot,
+        definition_key: &DefinitionKey<'static>,
+    ) -> Result<(), ConceptWriteError> {
+        // TODO: Check that struct is not referred from attributes (value type) and other structs (fields)!
+        // OperationTimeValidation::validate_struct_is_not_used(snapshot, entity_type.clone().into_owned())
+        //     .map_err(|source| ConceptWriteError::SchemaValidation { source })?;
```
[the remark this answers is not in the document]
farost: I wanted the current tests to pass (otherwise it's hard to control that overall the tests are written correctly as they always change), but your idea makes sense for this place as it could be done just through the schema level. Will do

concept/type_/type_manager.rs
```diff
@@ -986,56 +1238,169 @@ impl<Snapshot: WritableSnapshot> TypeManager<Snapshot> {
...
+        &self,
+        snapshot: &mut Snapshot,
+        type_: RelationType<'static>,
+        label: &Label<'_>,
+    ) -> Result<(), ConceptWriteError> {
+        debug_assert!(OperationTimeValidation::validate_type_exists(snapshot, type_.clone()).is_ok());
+
+        OperationTimeValidation::validate_label_uniqueness(snapshot, &label.clone().into_owned())
+            .map_err(|source| ConceptWriteError::SchemaValidation { source })
+            .unwrap(); // TODO: Propagate error instead
```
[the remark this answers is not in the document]
farost: Because it was a copypaste from someone else's code, I should've changed the original, but missed this one =)

concept/type_/type_manager.rs
```diff
@@ -986,56 +1238,169 @@ impl<Snapshot: WritableSnapshot> TypeManager<Snapshot> {
...
+        .map_err(|source| ConceptWriteError::SchemaValidation { source })?;
+
+        OperationTimeValidation::validate_value_type_compatible_to_all_owns_annotations(
+            snapshot,
+            attribute.clone(),
+            Some(value_type.clone()),
+        )
+        .map_err(|source| ConceptWriteError::SchemaValidation { source })?;
+
+        // TODO: Maybe it should be schema validation!
```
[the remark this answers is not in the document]
farost: I mean commit time validaation, will fix the todo

concept/type_/type_manager.rs
```diff
@@ -1182,136 +1577,638 @@ impl<Snapshot: WritableSnapshot> TypeManager<Snapshot> {
...
+            None => {}
+        }
+
+        self.set_role_ordering_unconditional(snapshot, role, ordering)
     }
 
-    pub(crate) fn delete_annotation_abstract(&self, snapshot: &mut Snapshot, type_: impl TypeAPI<'static>) {
-        // TODO: Validation
-        TypeWriter::storage_delete_type_vertex_property::<AnnotationAbstract>(snapshot, type_)
+    fn set_role_ordering_unconditional(
```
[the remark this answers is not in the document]
farost: Done

concept/type_/type_manager.rs
farost: Packed everything in common functions, looks much better.

concept/type_/owns.rs
```diff
@@ -200,36 +237,59 @@ impl<'a> InterfaceImplementation<'a> for Owns<'a> {
...
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub enum OwnsAnnotation {
     Distinct(AnnotationDistinct),
     Unique(AnnotationUnique),
     Key(AnnotationKey),
     Cardinality(AnnotationCardinality),
+    Regex(AnnotationRegex),
+}
+
+impl OwnsAnnotation {
```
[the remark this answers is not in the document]
farost: I'll name it `DefaultAnnotationFromCategory` as we want to return `AnnotationError` instead of `ConversionError`.
If we need another `Default***FromCategory` in the future, we could make `ConversionError` containing `source: AnnotationError` or `***Error` for something else, renaming the trait to `DefaultFromCategory`. It is redundant now, I guess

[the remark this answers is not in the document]
farost: Done! It looks a bit more verbose but does the job =)

PR #7099, opened 2024-07-09
Traversal architecture for enumerated, counted, checked, and bound vars

flyingsilverfin:
## Usage and product changes

We implement new traversal behaviour: we enumerate all answers for named variables from a match, non-distinctly. However, when we encounter anonymous variables, we treat these as existence checks that should not be enumerated.

These requirements translate into four types of variable modes to be handled: enumerated, counted, checked, and bound.

1. **Enumerated**: these are 'selected' variables for which _all answers_ must be emitted from a step. These could be anonymous variables, introduced by the compilation step as intermediates and used in further steps, or user-written named variables that should be outputted.
Example: 
```
match $x has name $a; select $x, $a;
 // --> 
1 x [ p1, n1 ]
1 x [ p1, n2 ]
```

2. **Counted**: these are user-named, but not selected variables that can be not-emitted as an optimisation and multiplied into the anwer **Multiplicity** instead.
Example: 
```
match $x has name $a; select $x; 
// --> 
2 x [ p1 ]
```

3. **Checked**: These are anonymous variables, user-written, which should not impact the output Multiplicity or quantity:
```
match $x has name $_;
// --> 
1 x  [ p1) ]
```

Like this, we can consider introduce anonymous variables freely without worrying about affecting the user's output:
```
match $x isa person;
->
match $x isa $_0; $_0 label person;
```

4. **Bound**: these variables are bound from previous computation.

---

The overall execution algorithm is now:

1. Each `step` produces a `batch`, which a Step takes in as input from the previous Step.
2. In a `step`, one or more instruction iterators/checks agree on one output `Row`
3. Once a row is agreed, we must move the iterators forward. Here, we behave differently depending what kind of mode each Variable is in - sometimes we have to compute row multiplicity, and sometimes we have to check for a value and skip forward.
4. We record the row into the output `batch` and are ready to go back to step 2

## Implementation

Significant new APIs and changes, beyond the architecture for new Variable Modes, and answer Multiplicity:

- `next_possible()` available on Things and Types. This is a byte-wise 'increment' (big-endian as all our keys are) that basically lets us move a concept forward to the next possible value in the storage layer. This is useful for skipping forward (and in the future, seeking) to new next values in iterators.
- Standardise `RolePlayerIndex` as the shortcut index from role player to role player (5-component edge)
- `AttributeIterator` no longer depends on snapshots/managers, in order to check if a type is indepenent - we simplify this lifetime by just injecting an `Arc`ed set of independent attribute types from the cache (unless in schema txn). This set in general should be very small
- Implement a byte-wise `Hash` that is usable for all `VariableValue`s - this is just going to hash the exact content of any values as they are stored in bytes on disk. We are still susceptible to rounding errors from floating-point arithmetic, but we should never miss answers, only produce too many similar ones. Note: `Duration` are hashed as the components: `(Months, Days, Nanos)`

### Traversal implementation notes

Could not work out how to generically implement count+skip across many different types of iterators, so used Macros instead (see `iterator_advance`).

Also traded probably a bit of performance in `InstructionIterator` enum, by nesting all `HasSortedAttribute/Owner` iterators into a sub-enum. This means two layers of dispatch at runtime, but makes grouping our operations and code blocks much simpler to reason about and reuse!


traversal/executor/pattern_executor.rs
```diff
@@ -399,16 +439,16 @@ impl SortedExecutor {
...
 
                 let (containing_i, containing_max, i_index, max_index) = if current_max_index > i {
-                    let (containing_i, containing_max) = self.intersection_iterators.split_at_mut(current_max_index);
+                    let (containing_i, containing_max) = self.iterators.split_at_mut(current_max_index);
                     (containing_i, containing_max, i, 0)
                 } else {
-                    let (containing_max, containing_i) = self.intersection_iterators.split_at_mut(i);
+                    let (containing_max, containing_i) = self.iterators.split_at_mut(i);
                     (containing_i, containing_max, 0, current_max_index)
                 };
```
krishnangovindraj, line 453: What is this for? Getting two mutable references?

traversal/executor/pattern_executor.rs
```diff
@@ -399,16 +439,16 @@ impl SortedExecutor {
         loop {
```
krishnangovindraj, line 439: Am I missing something or can this be simplified to?

```
let mut failed = false; // Don't need retry.
while !failed {
 ...
}
if failed { ... }
else { /* !retry is implicitly true*/ }
```

krishnangovindraj: I have the feeling the whole loop could be simplified/optimised with a max-heap.

traversal/executor/pattern_executor.rs
flyingsilverfin: Yes

traversal/executor/pattern_executor.rs
flyingsilverfin: It may be, we could look at it monday

PR #7101, opened 2024-07-12
Implement function representation

krishnangovindraj:


ir/program/program.rs
```diff
@@ -30,18 +38,234 @@ impl Program {
...
+    pub(crate) fn functions(&self) -> &Vec<FunctionIR> {
+        &self.functions
+    }
+
+    pub(crate) fn into_parts(self) -> (FunctionalBlock, Vec<FunctionIR>) {
+        let Self { entry, functions } = self;
+        (entry, functions)
+    }
+
+    // TODO: ProgramDefinitionError?
```
krishnangovindraj, line 50: Do we want yet another level of errors?
The compilation error can come from the entry pattern.

ir/program/program.rs
```diff
@@ -4,20 +4,28 @@
...
     entry: FunctionalBlock,
-    functions: HashMap<DefinitionKey<'static>, FunctionIR>,
+    functions: Vec<FunctionIR>,
 }
 
 impl Program {
-    pub fn new(entry: FunctionalBlock, functions: HashMap<DefinitionKey<'static>, FunctionIR>) -> Self {
+    pub fn new(entry: FunctionalBlock, functions: Vec<FunctionIR>) -> Self {
         // TODO: verify exactly the required functions are provided
+        // TODO: ^ Why? I've since interpreted it as the query-local functions
```
krishnangovindraj, line 28: I interpret the functions passed to a program as the IR of the preamble functions only.
An `AnnotatedProgram` has access to all the functions in the schema as well.

function/function.rs
```diff
@@ -4,95 +4,27 @@
...
-    ValueOptional(ValueType),
-    ThingList(Type),
-    ThingListOptional(Type),
-    ValueList(ValueType),
-    ValueListOptional(ValueType),
+pub type StoredFunction = UncompiledFunction<DefinitionKey<'static>>;
+pub struct UncompiledFunction<IDType: FunctionIDTrait> {
+    pub(crate) function_id: IDType,
+    pub(crate) parsed: typeql::schema::definable::Function,
 }
```
krishnangovindraj, line 16: This no longer holds the IR.
All other information was held in the TypeQL function representation. 
The signature we needed in the IR package, so it was moved there and uses IR data-structures.
These names aren't great though.

encoding/graph/definition/definition_key_generator.rs
```diff
@@ -28,4 +32,13 @@ impl DefinitionKeyGenerator {
...
     }
+
+    pub fn create_function<Snapshot: WritableSnapshot>(
+        &self,
+        snapshot: &mut Snapshot,
+    ) -> Result<DefinitionKey<'static>, EncodingError> {
+        let definition_key = self.next_function.allocate(snapshot)?;
+        snapshot.put(definition_key.as_storage_key().into_owned_array());
+        Ok(definition_key)
+    }
```
krishnangovindraj, line 43: Do we want to split this into `StructTypeIDGenerator` and `FunctionIDGenerator`?

krishnangovindraj (commented):
First self review

krishnangovindraj (commented):
one more round

encoding/graph/definition/definition_key_generator.rs
flyingsilverfin: It's ok like this

encoding/layout/prefix.rs
```diff
@@ -114,7 +115,8 @@ pub enum Prefix {
     PropertyObjectVertex,
 
     IndexLabelToType,
-    IndexNameToDefinitionStruct, // IndexLabelToDefinitionFunction
+    IndexNameToDefinitionStruct,
+    IndexNameToDefinitionFunction,
```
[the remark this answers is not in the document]
krishnangovindraj: It seems arbitrary to put them in the same namespace.

database/transaction.rs
```diff
@@ -142,17 +144,23 @@ impl<D: DurabilityClient> TransactionSchema<D> {
         let type_manager = Arc::new(TypeManager::new(
             database.definition_key_generator.clone(),
             database.type_vertex_generator.clone(),
-            None,
+            None, // TODO: pass cache
```
flyingsilverfin, line 147: Schema txn should not have cache, so this should always be null rigth?

function/function.rs
```diff
@@ -4,95 +4,34 @@
...
-struct FunctionArgument {
-    name: String,
-    type_: FunctionValuePrototype,
-}
-
-enum FunctionReturn {
-    Stream(Vec<FunctionValuePrototype>),
-    Single(FunctionValuePrototype),
+pub type SchemaFunction = Function<DefinitionKey<'static>>;
+pub struct Function<FunctionIDType: FunctionIDTrait> {
```
flyingsilverfin, line 13: You don't need to call types `Type`, just call it `FunctionID`, and itll get used as a type :)

ir/program/program.rs
```diff
@@ -30,18 +40,238 @@ impl Program {
...
+        &self.functions
+    }
+
+    pub(crate) fn into_parts(self) -> (FunctionalBlock, Vec<FunctionIR>) {
+        let Self { entry, functions } = self;
+        (entry, functions)
+    }
+
+    pub fn compile<'index>(
+        function_index: &impl FunctionSignatureIndex,
```
krishnangovindraj, line 53: This now takes a single `FunctionSignatureIndex` instead of one having only the Schema in. It's clean, because it requires you to index even the preamble functions before calling it and separates out indexing and compilation completely.

PR #7105, opened 2024-07-19
Finish database reset function to speed up behaviour tests

farost:
## Usage and product changes
We add `connection reset database` BDD command, which works just like `create database` if the database is absent and resets the database (removing all the data from it, making it look like a new database). 

This function is expected to be used in BDDs to speed up its execution, especially while the `cucumber` crate works quite slowly with huge files. It is important to note that this function should be used in:
* production (at least with the current non-thread-friendly implementation: can we somehow hide it?)
* `connection/database` BDD tests

## Implementation
* add `Mutex` to the `Arc` containing `Server` in BDDs so we could modify it (to fix the build)
* change `server.reset***` logic from `reset else delete` to `reset else recreate` to always end up with a clean database by the end of this call
* close active transactions at the end of a scenario + have an option to clean databases


tests/behaviour/concept/type/BUILD
```diff
@@ -57,6 +57,17 @@ rust_test(
     data = ["@vaticle_typedb_behaviour//concept/type:owns.feature"],
 )
 
+rust_test(
```
farost, line 60: Split tests here https://github.com/vaticle/typedb-behaviour/pull/290

tests/behaviour/steps/lib.rs
```diff
@@ -86,20 +90,43 @@ impl Context {
...
     }
 
-    pub fn databases(&self) -> &HashMap<String, Arc<Database<WALClient>>> {
-        self.server().unwrap().databases()
+    pub fn databases(&self) -> HashMap<String, Arc<Database<WALClient>>> {
+        self.server().unwrap().lock().unwrap().databases().clone()
+    }
+
+    pub fn database_names(&self) -> HashSet<String> {
+        self.server().unwrap().lock().unwrap().databases().keys().map(|name| name.to_owned()).collect()
```
farost, line 129: I have a separate function not increase the `Arc` counter while getting `databases()` for the cleanup

tests/behaviour/steps/lib.rs
```diff
@@ -86,20 +90,43 @@ impl Context {
...
 
-    pub fn server_mut(&mut self) -> Option<&mut typedb::Server> {
+    pub fn server_mut(&mut self) -> Option<&mut Arc<Mutex<typedb::Server>>> {
         self.server.as_mut()
     }
 
-    pub fn databases(&self) -> &HashMap<String, Arc<Database<WALClient>>> {
-        self.server().unwrap().databases()
+    pub fn databases(&self) -> HashMap<String, Arc<Database<WALClient>>> {
+        self.server().unwrap().lock().unwrap().databases().clone()
```
farost, line 125: Can't pass the ref anymore as we have `lock()`. Not a bad thing considering that we `clone` these databases on the callers' sides

tests/behaviour/steps/lib.rs
```diff
@@ -51,18 +54,19 @@ mod thing_util {
...
     object_lists: HashMap<String, Vec<Object<'static>>>,
     attributes: HashMap<String, Option<Attribute<'static>>>,
     attribute_lists: HashMap<String, Vec<Attribute<'static>>>,
+
+    clean_databases_after: bool,
 }
 
 impl Context {
-    pub async fn test(glob: &'static str) -> bool {
+    pub async fn test(glob: &'static str, clean_databases_after: bool) -> bool {
```
farost, line 67: Could make 2 public functions with or without cleanup if you wanted. It's fine to have a set of boolean flags now, can become a config in the future.

tests/behaviour/steps/connection/mod.rs
```diff
@@ -4,27 +4,30 @@
...
 use test_utils::{create_tmp_dir, TempDir};
 
 use crate::{generic_step, Context};
+use crate::connection::transaction::{transaction_closes, transaction_is_open};
 
 mod database;
 mod transaction;
 
-static TYPEDB: OnceCell<(TempDir, Arc<typedb::Server>)> = OnceCell::new();
+static TYPEDB: OnceLock<(TempDir, Arc<Mutex<typedb::Server>>)> = OnceLock::new();
```
farost, line 25: Had to add Mutex to send `&mut self` for the server's methods.

farost: FUNCTIONS*

flyingsilverfin (approved):

PR #7120, opened 2024-07-31
Add TypeCache to Database; implement statistics update

dmitrii-ubskii:


database/database.rs
```diff
@@ -185,37 +213,38 @@ impl Database<WALClient> {
...
 }
 
+fn update_statistics_action(storage: Arc<MVCCStorage<WALClient>>, schema: Arc<RwLock<Schema>>) -> impl Fn() {
+    move || {
+        let mut schema_lock = schema.write().unwrap();
+        let mut thing_statistics = (*schema_lock.thing_statistics).clone();
+        thing_statistics.may_synchronise(&storage).ok();
+        schema_lock.thing_statistics = Arc::new(thing_statistics);
+    }
+}
```
flyingsilverfin, line 246: This will block read transactions from opening - if there are many records in the WAL we have to catch up, even 10-60 seconds could make this operation non-trivial.
I suggest that we two the locking in two stages:
1. obtain a read-lock, get a clone of the statistics and synchronise it
2. upgrade to a write lock, and then replace the statistics object.

nb. can we atomically swap the read lock for a write lock? I think it's probably required for safety - eg. if a schema txn opens in between, then commits, we get into a skewed state

flyingsilverfin: Also let's call this function `update_statistics_runnable` or `_fn` or something? Action is a bit odd IMO

database/transaction.rs
```diff
@@ -80,24 +83,19 @@ impl<D: DurabilityClient> TransactionWrite<D> {
...
 
-        Self {
-            snapshot,
-            type_manager,
-            thing_manager,
-            _schema: schema,
-            _schema_txn_guard: schema_txn_guard,
-            _database: database,
-        }
+        drop(schema);
```
flyingsilverfin, line 96: How come we have to explicitly drop? Won't it happen automatically on return?

dmitrii-ubskii: It's borrowing `database` making it so it can't be moved into `Self {}` at the end of the scope.

flyingsilverfin (approved):

PR #7132, opened 2024-08-12
Rename roleplayer edge to links + plan

dmitrii-ubskii:
## Usage and product changes

We rename the role player edge to be referred to as a "links" edge, to avoid confusion with the `RolePlayer` struct which refers to a player in a relation bundled with the role type tag.

### DIFF
--- a/compiler/planner/pattern_plan.rs
+++ b/compiler/planner/pattern_plan.rs
@@ -19,8 +19,11 @@ use itertools::Itertools;
 
 use crate::{
     inference::annotated_program::AnnotatedProgram,
-    instruction::constraint::instructions::{ConstraintInstruction, HasInstruction, Inputs, IsaReverseInstruction},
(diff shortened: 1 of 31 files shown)


compiler/planner/pattern_plan.rs
```diff
@@ -134,42 +151,93 @@ impl PatternPlan {
...
-                                has.owner(),
-                                vec![ConstraintInstruction::Has(HasInstruction::new(
-                                    has.clone(),
-                                    Inputs::None([]),
-                                    program.entry_annotations(),
-                                ))],
-                                &[has.owner(), has.attribute()],
-                            )
-                        } else if bound_variables.len() == 2 {
+                        if bound_variables.len() == 2 {
```
flyingsilverfin, line 199: You wanna make these branches `todo!()` so they fail if we ever hit them in future match  BDD?

flyingsilverfin (approved):

PR #7136, opened 2024-08-16
Add data (thing) validation

farost:
## Usage and product changes
We add data validation for schema migration (changes of the schema) and data modifications.

Moreover, we add a number of additional operation time validations of data for schema migration which only existed for transaction commits, aiming to maximize the synchronization of schema and data of a database. There is also a change in the cleanup behavior for relations and attributes: now, nothing is hidden for queries until the commit time, considering that we delete and hide data only on the commit time.

We refactor locking for things in `thing_manager`, making sure that not only objects exist when we add edges to them, but also their interfaces and other objects, connected to these objects through inheritance like for `@unique` and `@card` constraints. The `@card` constraint, in its turn, is now checked both in operation time for schema for all operations with a potential effect on data's cardinality and in commit time for data (in case something has been missed in the more complex operation time checks).

Changes of previously existing checks:
1. Validations for types when an attribute type implicitly loses @independent annotation or a relation type implicitly gains @cascade annotation (which can lead to data deletion) are now throwing errors only if there are existing instances for these types (including subtypes).
2. We accidentally made the `create {root} type` idempotent, which is incorrect. Now, if you try to create an existing entity type, it is an error.

We also add the whole `concept` BDD tests package to the CI as it is considered to be complete. From now on, we better aim not to break the existing tests too much with new merges.


.factory/automation.yml
```diff
@@ -59,12 +59,12 @@ build:
...
-#      dependencies: [build]
-#      command: |
-#        bazel run @vaticle_dependencies//tool/bazelinstall:remote_cache_setup.sh
-#        bazel test //test/behaviour/concept/... --test_output=streamed
+    test-behaviour-concept:
+      image: vaticle-ubuntu-22.04
+      dependencies: [build]
+      command: |
+        bazel run @vaticle_dependencies//tool/bazelinstall:remote_cache_setup.sh
+        bazel test //tests/behaviour/concept/... --test_output=streamed
```
farost, line 67: I don't want to worry about unit tests now, but I don't want to continue fixing bdds after random commits from us, so I believe it's worth adding concept bdds (as they are finished and just wait for some minor todos based on the future discussions) into CI

compiler/match_/inference/mod.rs
farost, line 1: Passing thing_manager here

compiler/match_/inference/pattern_type_inference.rs
farost, line 1: Passing thing_manager here

compiler/match_/inference/type_seeder.rs
```diff
@@ -641,7 +641,7 @@ impl BinaryConstraint for Has<Variable> {
         owner
             .get_owns(seeder.snapshot, seeder.type_manager)?
             .iter()
-            .map(|(attribute, _)| TypeAnnotation::Attribute(attribute.clone()))
```
farost, line 644: There was a "bug": get_capabilities returned HashMap<InterfaceType, Capability>, where InterfaceType is always exactly the same as Capability::interface(). However, it may be worth returning Capability as a key and ObjectType (the source of this capability from inheritance) as value. We don't use it anywhere right now, so I didn't care, but I left a todo for it (don't want to spend additional time now).

compiler/match_/inference/type_seeder.rs
```diff
@@ -752,7 +752,7 @@ impl BinaryConstraint for Isa<Variable> {
             match left_type {
                 TypeAnnotation::Attribute(attribute) => {
                     attribute
-                        .get_supertypes(seeder.snapshot, seeder.type_manager)?
+                        .get_supertypes_transitive(seeder.snapshot, seeder.type_manager)?
```
farost, line 755: Thought that in
`get_supertype`
`get_supertypes` <- can look confusing 
`get_subtypes`
`get_subtypes_transitives`

concept/BUILD
```diff
@@ -41,13 +41,17 @@ rust_library(
         "@crates//:regex",
         "@crates//:serde",
     ],
+    proc_macro_deps = [
+        "@crates//:paste",
```
farost, line 45: I use it to generate multiple functions in a single macro for "smart" constraints checking. See `operation_time_validation` for `type_`.

concept/benches/bench_thing_write.rs
```diff
@@ -88,15 +88,18 @@ fn create_schema(
 ) {
     let mut snapshot: WriteSnapshot<WALClient> = storage.clone().open_snapshot_write();
     {
+        // TODO: Should not affect much if it's Arc while it was previously Arc?
```
farost, line 91: Is it fine?

concept/thing/attribute.rs
```diff
@@ -119,35 +119,43 @@ impl<'a> ThingAPI<'a> for Attribute<'a> {
...
+                | ValueType::DateTime
+                | ValueType::DateTimeTZ
+                | ValueType::Duration => snapshot.put(self.vertex().as_storage_key().into_owned_array()),
+                // ValueTypes with expensive writes
+                | ValueType::String | ValueType::Struct(_) => {
+                    thing_manager.lock_existing_attribute(snapshot, self.as_reference())
+                }
+            },
+            // Read error or empty value type
+            _ => snapshot.put(self.vertex().as_storage_key().into_owned_array()),
```
farost, line 151: Decided not to worry about strange value types specifically for locks. These errors should be checked in the business logic, not here.

concept/thing/attribute.rs
```diff
@@ -267,7 +277,8 @@ where
                 None => self.state = State::Done,
                 Some(Ok(attribute)) => {
                     let attribute_vertex = attribute.vertex();
-                    let independent = self.independent_attribute_types.contains(&attribute.type_());
+                    // let independent = self.independent_attribute_types.contains(&attribute.type_());
+                    let independent = true; // TODO: Turn off filtering, come up with a complete solution later
```
farost, line 281: Am I right that I can just remove the independence check here?

concept/thing/thing_manager.rs
```diff
@@ -806,31 +838,226 @@ impl ThingManager {
             Err(error) => Err(ConceptReadError::SnapshotGet { source: error }),
         }
     }
+
+    pub(crate) fn relation_exists<'a>(
+        &self,
+        snapshot: &impl ReadableSnapshot,
+        relation: &Relation<'a>,
+    ) -> Result<bool, ConceptReadError> {
+        match snapshot.get::<0>(relation.vertex().as_storage_key().as_reference()) {
```
farost, line 847: I couldn't add all the needed traits impls for `ThingAPI` as it's not a struct (spent some time there, coudn't finish it up), so I just created some special functions for `relation`s alongside `object`s instead of making it accept `&impl ThingAPI`

concept/thing/thing_manager.rs
```diff
@@ -806,31 +838,226 @@ impl ThingManager {
...
+            if cardinality == AnnotationCardinality::unchecked() {
+                break;
+            }
+
+            let lock_key = create_custom_lock_key(
+                [
+                    &Infix::PropertyAnnotationCardinality.infix_id().bytes(),
+                    owner.vertex().bytes().bytes(),
+                    locked_capability.interface().vertex().bytes().bytes(),
+                    // TODO: Do we need attribute type id?
```
farost, line 983: Attribute type id was mentioned in my notes, but I do believe that we already have enough for `card` locks here:
`Infix::AnnootationCardinality` + `owner's object` + `interface_type that is involved in the card check`

concept/thing/thing_manager.rs
```diff
@@ -806,31 +838,226 @@ impl ThingManager {
...
+        owner: &Object<'a>,
+        value: Value<'a>,
+        owns: Owns<'static>,
+    ) -> Result<(), ConceptReadError> {
+        if let Some(uniqueness_source) = get_uniqueness_source(snapshot, self, owns)? {
+            let lock_key = create_custom_lock_key(
+                [
+                    &Infix::PropertyAnnotationUnique.infix_id().bytes(),
+                    uniqueness_source.attribute().vertex().bytes().bytes(),
+                    value.encode_bytes::<128>().bytes(), // TODO: Where to get 128 from?
```
farost, line 954: Do we want to encode the whole value here? If yes, I feel like it'd be cool to just place this `128` in `value.rs` as `SOMETHING_DEFAULT` as we already reference it in multiple places.

concept/thing/thing_manager.rs
```diff
@@ -866,23 +1093,59 @@ impl ThingManager {
             }
         }
 
+        let mut any_deleted = true;
```
farost, line 1096: The following code is related to relations and attributes cleanups. 
I could optimize it to iterate through all writes just once instead of copying the buffer every time just like we did (with an additional `clone` =) ) in the locks function, but it's easier to scale the code this way for now.

concept/thing/thing_manager.rs
farost, line 1: There are some OperationTimeValidations in concept files (`relation.rs`), and some checks in `thing_manager`. I had the urge to find a single place for it, but it feels as just a correct result of having more logic in the concept files compared to `thing_manager`, where concept classes are just thin wrappers around `type_manager` functions, so didn't spend time on changing the ordering and places of these validations calls.

concept/thing/thing_manager/validation/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+        Ordering,
+    },
+};
+
+pub(crate) mod commit_time_validation;
+pub(crate) mod operation_time_validation;
+pub(crate) mod validation;
+
+#[derive(Debug, Clone)]
+pub enum DataValidationError {
```
farost, line 31: Moved a bunch of ConceptWriteErrors related to validation here and added much more new errors.

concept/type_/attribute_type.rs
```diff
@@ -177,78 +224,81 @@ impl<'a> AttributeType<'a> {
...
-        type_manager: &'m TypeManager,
-    ) -> Result<MaybeOwns<'m, Vec<AttributeType<'static>>>, ConceptReadError> {
-        type_manager.get_attribute_type_supertypes(snapshot, self.clone().into_owned())
+        type_manager: &TypeManager,
+    ) -> Result<bool, ConceptReadError> {
+        type_manager.get_attribute_type_is_independent(snapshot, self.clone())
     }
 
-    pub fn get_subtypes<'m>(
+    // TODO: Might call it "get_regex", "get_range", "get_values", but "get_values" sounds like
```
farost, line 254: Unsure about the naming =)

concept/type_/constraint.rs
```diff
@@ -0,0 +1,144 @@
...
+                    $annotation_enum(annotation) => Some(annotation),
+                    _ => None,
+                })
+                .next()
+        }
+    };
+}
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub enum ConstraintValidationMode {
```
farost, line 45: My algorithm of annotations revalidation based on schema changes (you'll see it in type_manager, but it's a good place to announce it):
1. Collect all the changed types/capabilities and their annotations (set annotation, changed supertype, changed override, created a new capability for type, unset override (get default cardinality), ....). It may be just one type/capability with one annotation, it may be a list (type/capability) of lists (annotations).
2. For types, just iterate over your instances and instances of your subtypes if they are affected (annotation is inherited). For capabilities, iterate over every pair of {object_type, interface_type} affected (considering overrides, considering inheritance of capability, blablabla...)
3. Just validate the updated types/capabilities, iterating over their instances.

However, steps 2-3 can look different, as we may want to check instances of:
1. Purely one isolated `Type` (regex, values, range): grab one type/cap, iterate over it, validate. Grab its subtype/overriding, iterate over it, validate. Repeat...
2. `Type And Siblings` (cardinality). By `siblings` I mean "we have a common parent (not specifically declared) and we are related to each other in any other way (for capabilities it means being owned by a single type). Grab instances of a type/capability and its siblings and check them together.
3. `Type And Siblings And Subtypes` (unique): collect the whole family together, let them sit in one car. Start the data iteration show only when everyone is in place.

I use this enum for branching between these cases based on specific annotations updated. If I have `card`, `range`, and `regex` updated in one operation, I'll call two instances validations: `Type` and `TypeAndSiblings`. And I use a single function (with some minor preparations) for the algorithm described above, so we can reuse it everywhere (just like I did yesterday when new `card` issues appeared) and append new annotations to it just to several places without worrying about anything else!

tests/behaviour/concept/migration/migration.rs
farost, line 1: Looks like `migration` is a better name than `validation` to me, because we have a lot of different`validation`s being tested in `thing` or `type` as well. Here, it's `data_validation.feature` and basic `migration.feature` (I could call it 'migration_validation') of the `migration` package.

concept/type_/type_manager/validation/operation_time_validation.rs
```diff
@@ -70,32 +78,883 @@ macro_rules! object_type_match {
...
+            }
+
+            Ok(())
+        }
+    };
+}
+
+macro_rules! capability_or_its_overriding_capability_with_violated_new_annotation_constraints {
+    ($func_name:ident, $capability_type:ident, $object_type:ident, $interface_type:ident, $existing_instances_validation_func:path) => {
+        paste! {
```
farost, line 281: I use paste here to hide these special "collection" functions for `Type`, `TypeAndSiblings`, `TypeAndSiblingsAndSubtypes` constraint validation modes. Otherwise I'd need to expose them as separate macros with their own implementations, while they should be only used in this context.

concept/thing/thing_manager.rs
farost: I could also try to write general functions to reduce code volumes, but I didn't do it for the same reason...

concept/type_/annotation.rs
```diff
@@ -79,7 +81,12 @@ impl AnnotationCardinality {
         Self { start_inclusive, end_inclusive }
     }
 
+    // TODO: Can move to impl Default for AnnotationCardinality and other Annotations
     pub const fn default() -> Self {
+        Self::unchecked()
+    }
+
+    pub const fn unchecked() -> Self {
```
flyingsilverfin, line 89: Why is this called `unchecked`?

concept/type_/annotation.rs
```diff
@@ -583,25 +590,35 @@ macro_rules! empty_type_vertex_property_encoding {
     };
 }
 
-empty_type_vertex_property_encoding!(AnnotationAbstract, PropertyAnnotationAbstract);
-empty_type_vertex_property_encoding!(AnnotationIndependent, PropertyAnnotationIndependent);
-empty_type_vertex_property_encoding!(AnnotationDistinct, PropertyAnnotationDistinct);
+// It is a sign of poor architecture, but it lets us wrap other general places for annotations,
```
flyingsilverfin, line 593: okay, i'm not gonna look into this too deeply since it's small :)

concept/type_/attribute_type.rs
```diff
@@ -141,6 +177,15 @@ impl<'a> ThingTypeAPI<'a> for AttributeType<'a> {
 }
 
 impl<'a> AttributeType<'a> {
+    pub fn get_value_type_declared(
+        // TODO: add concept api tests
```
flyingsilverfin, line 181: TODO still relevant?

concept/type_/attribute_type.rs
flyingsilverfin: The naming is not bad at all - i would just flip them around to be `get_constraint_xxx` instead of `get_xxx_constraint` so the groupings become obvious when you're reading!

concept/type_/entity_type.rs
```diff
@@ -94,12 +94,16 @@ impl<'a> TypeAPI<'a> for EntityType<'a> {
...
 
-    fn delete(self, snapshot: &mut impl WritableSnapshot, type_manager: &TypeManager) -> Result<(), ConceptWriteError> {
-        type_manager.delete_entity_type(snapshot, self.clone().into_owned())
+    fn delete(
+        self,
+        snapshot: &mut impl WritableSnapshot,
+        type_manager: &TypeManager,
+        thing_manager: &ThingManager,
+    ) -> Result<(), ConceptWriteError> {
+        type_manager.delete_entity_type(snapshot, thing_manager, self.clone().into_owned())
```
flyingsilverfin, line 106: Not sure we need the `.clone()` on self since you have self by ownership!

concept/type_/relation_type.rs
```diff
@@ -257,21 +290,33 @@ impl<'a> RelationType<'a> {
...
         &self,
         snapshot: &impl ReadableSnapshot,
         type_manager: &TypeManager,
         role_name: &str,
     ) -> Result<Option<Relates<'static>>, ConceptReadError> {
-        let role_label = Label::build_scoped(role_name, self.get_label(snapshot, type_manager)?.name().as_str());
-        Ok(type_manager
-            .get_role_type(snapshot, &role_label)?
-            .map(|role_type| Relates::new(self.clone().into_owned(), role_type)))
+        for relates in type_manager.get_relation_type_relates(snapshot, self.clone().into_owned())?.into_iter() {
```
flyingsilverfin, line 312: Just double check, I think Krishnan or Dmitrii recently added an API to get relations with a role Name, which might not be useful here (about as expensive/difficult) but might be useful elsewhere

concept/type_/role_type.rs
```diff
@@ -143,12 +137,16 @@ impl<'a> TypeAPI<'a> for RoleType<'a> {
...
 
-    fn delete(self, snapshot: &mut impl WritableSnapshot, type_manager: &TypeManager) -> Result<(), ConceptWriteError> {
-        type_manager.delete_role_type(snapshot, self.clone().into_owned())
+    fn delete(
+        self,
+        snapshot: &mut impl WritableSnapshot,
+        type_manager: &TypeManager,
+        thing_manager: &ThingManager,
+    ) -> Result<(), ConceptWriteError> {
+        type_manager.delete_role_type(snapshot, thing_manager, self.clone().into_owned())
```
flyingsilverfin, line 149: Same redundant clone i think!

traversal/BUILD
```diff
@@ -0,0 +1,14 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+load("@vaticle_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
+
+checkstyle_test(
```
flyingsilverfin, line 7: This package doesn't exist anymore! YOu can delete this

flyingsilverfin: was probably a m merge/rebase misstep, no big deal

tests/behaviour/steps/params.rs
```diff
@@ -298,14 +303,38 @@ impl FromStr for ObjectRootLabel {
...
+            "relation" | "relations" => TypeDBTypeKind::Relation,
             _ => unreachable!(),
         };
         Ok(Self { kind })
     }
 }
 
+#[derive(Debug, Parameter)]
+#[param(name = "root_label_extended", regex = r"(attribute|entity|relation|role|object)")]
+pub(crate) enum RootLabelExtended {
```
flyingsilverfin, line 316: Can we rename this to `Kind` instead of `RootLabel`? Also probably needs to be done in BDD and implementations thereof... to keep this PR from growing more, maybe you can make a note to yourself to make a little follow up one to replace Root with Kind in BDD/place you find it in Core

tests/behaviour/steps/connection/transaction.rs
```diff
@@ -18,6 +18,7 @@ use crate::{
 #[apply(generic_step)]
 #[step(expr = "connection open {word} transaction for database: {word}")]
 pub async fn connection_open_transaction(context: &mut Context, tx_type: String, db_name: String) {
+    assert!(context.transaction().is_none(), "Have to close previous transaction");
```
flyingsilverfin, line 21: I prefer the assertion/error message style like "Previous/Existing transaction must be closed" but not any kind of issue really

flyingsilverfin (changes requested):
I've only seen 35/85 files, but I will continue tomorrow!

concept/thing/attribute.rs
flyingsilverfin: I need to have a deeper think about this one tomrrow/on monday with you, will come back to it

concept/thing/thing_manager.rs
flyingsilverfin: I think we need the attribute type ID because you can have different card annotations for different ownerships of the object:
```
define
person owns id @card(1..1), owns email @card(0..10)
```
When you update the person's ID you need to lock `[@card][person1][has][id]` and [@card][person1][has][email]` otherwise transactions that dont actually overlap could affect each other with the lock `[@card][person1][has]`

compiler/match_/inference/type_seeder.rs
flyingsilverfin: We still have a way to go from one capability to the original declarer easily right? Then we don't need this right now, i agree

tests/behaviour/steps/params.rs
farost: Will look at it

concept/type_/annotation.rs
farost: The idea was like "this card is so permissive that we don't really need to check it", can be used for optimizations (e.g. when we create locks)

concept/type_/attribute_type.rs
farost: Yep, specifically for this call. Probably just need several steps in existing tests

concept/thing/attribute.rs
flyingsilverfin, line 151: This should be an impossible state right? Impossible/unexpected states we should as a policy make the system crash - we don't want it to move along in an invalid state. This will force us to address it quickly as well.

i would just say `panic!("Attribute instances must have a value type.)`.

However, the ConceptReadError you can actually propagate up to the caller since that could be something like the database is deleted/storage shutdown/OS error (we haven't quite decided how to handle these yet, so we're sticking to propagating them back)

compiler/match_/inference/type_seeder.rs
farost: Yeah, I lagged when adding this comment tbh. We don't need a HashMap here because a capability already contains {source, interface_type}. So no more refactoring is needed

flyingsilverfin (commented):
Only thing left to check is the type_/constaint thing you've done! I will look at it a bit later :)

concept/thing/attribute.rs
```diff
@@ -171,8 +178,13 @@ impl<'a> ThingAPI<'a> for Attribute<'a> {
         snapshot: &impl ReadableSnapshot,
         type_manager: &TypeManager,
     ) -> Result<Prefix, ConceptReadError> {
-        let value_type = type_.get_value_type(snapshot, type_manager)?.unwrap();
-        Ok(Self::Vertex::value_type_category_to_prefix_type(value_type.category()))
+        let value_type = type_.get_value_type(snapshot, type_manager)?;
+        match value_type {
+            // It should be valid to try getting a prefix without a value type, so we just return
+            // the first prefix of possible. Should get 0 instances anyway
+            None => Ok(Prefix::VertexAttributeBoolean),
```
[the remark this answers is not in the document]
farost: @flyingsilverfin Well, it was my intention to be able to use it for types without value types, because it is a valid request to get instances of abstract types. I might change the caller's code not to use this method, but, if I remember correctly, the code was generic and I didn't want to change anything there. Might do.

flyingsilverfin (changes requested):
Ok last batch!

Overall, fantastic job -- I'm not sure I could have taken on so much without getting lost and going in circles myself! Excellent stuff, I have high confidence of the correctness of your implementation, even if it is a big surface area of new code 💯 

My main review feedbacks would only be to generally think twice you're using the right data structures (Set? &[]? etc.) and input/return types. Secondly, placing methods in the right domain (eg. type validations would be under a `/type/...` package. 

Lastly, I suspect there are elements of repetition, duplication, that can be eliminated and centralised (the previous feedback about code placement generally makes this easier to identify) and shorten the code you submitted. However, on this point, these elements were not very easy for me to spot large wins on during my review. Given that let's aim for a longer but easy to follow and correct implementation like you've got here, and see if it gets in the way and how it benchmarks (and it's ok if this isn't performant for now anyhow).

Nice work!

concept/thing/thing_manager.rs
farost: @flyingsilverfin Right now I use:
`[@card][person][id]` and `[@card][person][email]`. I don't think we need to specify `has` here because we don't have (and more likely won't have) situations when there is more than one capability between the same pair of `{object_type, interface_type}`.

concept/thing/thing_manager.rs
farost: @flyingsilverfin What do you think?

concept/type_/attribute_type.rs
farost: Agree, done.

concept/thing/thing_manager.rs
```diff
@@ -893,121 +1147,311 @@ impl ThingManager {
...
+            ))
+            .filter_map(|(key, _)| {
+                let bytes = Bytes::reference(key.bytes());
+                if EdgeOverride::<Owns<'static>>::is_decodable_from(bytes.clone()) {
+                    let property = TypeEdgeProperty::new(Bytes::Reference(key.byte_array().as_ref()));
+                    let edge = property.type_edge();
+                    let prefix = edge.prefix();
+                    // We always store EdgeOverride on canonical edges, but it's safe to check both here just in case.
+                    if prefix == Owns::CANONICAL_PREFIX || prefix == Owns::REVERSE_PREFIX {
+                        return Some(ObjectType::new(edge.from()).into_owned());
```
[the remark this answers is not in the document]
farost: Good point! Done

concept/thing/thing_manager.rs
```diff
@@ -806,31 +829,226 @@ impl ThingManager {
             Err(error) => Err(ConceptReadError::SnapshotGet { source: error }),
         }
     }
+
+    pub(crate) fn relation_exists<'a>(
```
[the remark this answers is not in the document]
farost: Done.

concept/thing/thing_manager.rs
```diff
@@ -893,121 +1147,311 @@ impl ThingManager {
...
+        {
+            let mut players = self.get_objects_in(snapshot, player_type);
+            while let Some(player) = players.next().transpose()? {
+                if self.object_exists(snapshot, &player)? {
+                    out_players.insert(player.into_owned());
+                }
+            }
+        }
+
+        for relation_type in snapshot
```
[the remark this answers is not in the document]
farost: Yes, that's what we agreed to. I'm going to remove these lines and see if the tests are moved in any direction.

concept/thing/thing_manager.rs
```diff
@@ -1045,115 +1529,88 @@ impl ThingManager {
...
+            Value::Double(double) => {
+                let encoded_double = DoubleBytes::build(double);
+                self.vertex_generator.create_attribute_double(
+                    attribute_type.vertex().type_id_(),
+                    encoded_double,
+                    snapshot,
+                )
+            }
+            Value::Decimal(decimal) => {
+                println!("PUT DECIMAL ATTR: {:?}", decimal);
```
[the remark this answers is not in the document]
farost: Done, mb

concept/thing/thing_manager.rs
```diff
@@ -1241,24 +1698,43 @@ impl ThingManager {
...
+            value.value_type(),
+        )?;
+
+        let owns = owner.get_type_owns(snapshot, self.type_manager(), attribute_type.clone())?.into_owned();
+
+        OperationTimeValidation::validate_has_regex_constraint(snapshot, self, owns.clone(), value.clone())
+            .map_err(|source| ConceptWriteError::DataValidation { source })?;
+
+        OperationTimeValidation::validate_has_range_constraint(snapshot, self, owns.clone(), value.clone())
+            .map_err(|source| ConceptWriteError::DataValidation { source })?;
```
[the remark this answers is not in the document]
farost: I want to change both `DataValidation` and `SchemaValidation`, but the latter can expect more context. So I'd wait until we merge with your error implementations. I'm up to cleaning up these `map_err`s on one of the weekends (I usually do some lazy refactoring like that then) after we finalize our approach

concept/thing/relation.rs
```diff
@@ -372,26 +334,20 @@ impl<'a> ThingAPI<'a> for Relation<'a> {
...
-        // TODO
-        /*
-        for attr in self
-            .get_has_ordered(snapshot, thing_manager)
-            .collect_cloned_vec(|(key, _value)| key.into_owned())
-            .map_err(|err| ConceptWriteError::ConceptRead { source: err })?
-        {
-            // TODO huh?
-            thing_manager.unset_has_ordered(snapshot, &self, attr.type_());
+        for owns in self.type_().get_owns(snapshot, thing_manager.type_manager())?.iter() {
```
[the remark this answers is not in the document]
farost: Have I understood you correctly that I don't need to change anything in my PR here and we already have a task for the mentioned optimization?

concept/thing/object.rs
```diff
@@ -231,51 +224,32 @@ pub trait ObjectAPI<'a>: for<'b> ThingAPI<'a, Vertex<'b> = ObjectVertex<'b>> + C
...
+
+        OperationTimeValidation::validate_object_type_owns_attribute_type(
+            snapshot,
+            thing_manager,
+            self.type_(),
+            attribute.type_(),
+        )
+        .map_err(|error| ConceptWriteError::DataValidation { source: error })?;
+
+        let owns = self.get_type_owns(snapshot, thing_manager.type_manager(), attribute.type_())?;
```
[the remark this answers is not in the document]
farost: Sounds good!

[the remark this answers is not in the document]
farost: Btw, it's not even my method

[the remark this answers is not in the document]
farost: But it makes everything cleaner indeed, done

concept/thing/thing_manager/validation/validation.rs
```diff
@@ -0,0 +1,75 @@
...
+    },
+    type_::{
+        annotation::{AnnotationCategory, AnnotationKey, AnnotationUnique},
+        owns::{Owns, OwnsAnnotation},
+        type_manager::TypeManager,
+        Capability, TypeAPI,
+    },
+};
+
+pub(crate) fn get_label_or_concept_read_err<'a>(
```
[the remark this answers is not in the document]
farost: `try_get_label` sounds good, I'll do it!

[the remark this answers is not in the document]
farost: I'll do it for `thing_manager`, but ignore `type_manager` and leave a todo for it, because I'll need to pass `&TypeManager` everywhere, while I'm still using `TypeReader` in most of the places for `type_manager` validations. Haven't found time to refactor it.

concept/thing/thing_manager/validation/validation.rs
```diff
@@ -0,0 +1,75 @@
...
+
+pub(crate) fn get_label_or_data_err<'a>(
+    snapshot: &impl ReadableSnapshot,
+    type_manager: &TypeManager,
+    type_: impl TypeAPI<'a>,
+) -> Result<Label<'static>, DataValidationError> {
+    get_label_or_concept_read_err(snapshot, type_manager, type_).map_err(DataValidationError::ConceptRead)
+}
+
+pub(crate) fn get_uniqueness_source(
```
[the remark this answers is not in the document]
farost: I didn't do it because:
1. I didn't want to expose it as it should be needed only for the inner logic of us
2. It kinda feels lonely and I have a desire to create multiple `get_***_source` for other annotations (and their final values), like the last `values` narrowing, but it's not needed. 

Feels better in terms of domain, so I'll move this function to `Owns`.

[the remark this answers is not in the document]
farost: Nah, I still want to panic, because `is_unique == true` && `get_uniqueness_source() == None` means broken logic!

concept/thing/thing_manager/validation/operation_time_validation.rs
```diff
@@ -0,0 +1,448 @@
...
+                snapshot,
+                &thing_manager.type_manager,
+                type_,
+            )?))
+        } else {
+            Ok(())
+        }
+    }
+
+    pub(crate) fn validate_object_type_plays_role_type(
```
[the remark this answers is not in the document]
farost: It's a `data validation` for me as it's a validation of a `data` operation. Similarly, when I check that there are no instances of a `type` when I set `abstract`, it's a `schema validation`: we validate `schema` against `data`, and it's the changed `schema` who would be in an incorrect state if we approved that operation, not the already persisted data!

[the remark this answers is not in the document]
farost: For the same reason I don't see a need to move it to `validation.rs`, it's good to stay here while it's only used here. I call `OperationTimeValidation::xxx` and `CommitTimeValidation::xxx` to highlight the intentional moment of time when it should be called (not sure if its' the correct domain specification, but it's what we've been using in `type_manager`, so it's a good symmetry).

Don't find nesting of `Schema`/`Data` validation errors useful here. I might refactor it so it becomes useful, but if will take time, and I'm not sure if it's beneficial right now. Better just look at what we have in the end and spend one-two hours on refactoring later.

concept/type_/attribute_type.rs
farost: Fixed

concept/type_/entity_type.rs
farost: Gp!

concept/type_/relation_type.rs
farost: Yes, I see `get_roles_by_name`, but it's different and is not scoped to specific relation. Thanks

concept/type_/role_type.rs
farost: I'll be more careful!

traversal/BUILD
farost: Done

tests/behaviour/steps/params.rs
farost: Created a task

tests/behaviour/steps/connection/transaction.rs
farost: Gp, fixed

concept/thing/attribute.rs
farost: Good point, fixed. Hadn't wanted to return Result from this method for some reason.

concept/type_/type_manager/type_reader.rs
```diff
@@ -277,37 +280,35 @@ impl TypeReader {
...
+    ) -> Result<HashSet<CAP>, ConceptReadError> {
         let owns_prefix = CAP::prefix_for_canonical_edges_from(CAP::ObjectType::new(owner.into_vertex()));
         snapshot
             .iterate_range(KeyRange::new_within(owns_prefix, TypeEdge::FIXED_WIDTH_ENCODING))
             .collect_cloned_hashset(|key, _| CAP::decode_canonical_edge(Bytes::Reference(key.byte_ref()).into_owned()))
             .map_err(|error| ConceptReadError::SnapshotIterate { source: error })
     }
 
-    pub(crate) fn get_capabilities<CAP>(
+    // TODO: Return HashMap<CAP, ObjectType (source)> if needed, just like we do for get_annotations
```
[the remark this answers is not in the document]
farost: Yes!

concept/type_/type_manager/type_writer.rs
```diff
@@ -70,8 +70,28 @@ impl<Snapshot: WritableSnapshot> TypeWriter<Snapshot> {
...
+    pub(crate) fn storage_unput_vertex(snapshot: &mut Snapshot, type_: impl TypeAPI<'static>) {
+        debug_assert!(snapshot.get::<0>(type_.vertex().as_storage_key().as_reference()).unwrap_or(None).is_some());
+        snapshot.unput(type_.vertex().as_storage_key().into_owned_array());
+    }
+
+    pub(crate) fn storage_unput_edge<EDGE>(snapshot: &mut Snapshot, capability: EDGE)
+    where
+        EDGE: TypeEdgeEncoding<'static> + Clone,
+    {
+        let canonical_key = capability.clone().to_canonical_type_edge().into_storage_key();
```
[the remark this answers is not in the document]
farost: Yes, you're right. Will create a separate task for this refactoring for simplicity, will work on it right after this merge.

concept/type_/type_manager/type_writer.rs
```diff
@@ -70,8 +70,28 @@ impl<Snapshot: WritableSnapshot> TypeWriter<Snapshot> {
...
+        let reverse_key = capability.to_reverse_type_edge().into_storage_key();
+        debug_assert!(snapshot.get::<0>(canonical_key.as_reference()).unwrap_or(None).is_some());
```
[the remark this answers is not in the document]
farost: Okay, I like it! Done.

encoding/graph/thing/vertex_attribute.rs
```diff
@@ -176,7 +174,7 @@ impl<'a> ThingVertex<'a> for AttributeVertex<'a> {
 
 impl<'a> Keyable<'a, BUFFER_KEY_INLINE> for AttributeVertex<'a> {
     fn keyspace(&self) -> EncodingKeyspace {
-        EncodingKeyspace::Data
+        Self::KEYSPACE
```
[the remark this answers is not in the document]
farost: Done.

concept/type_/constraint.rs
```diff
@@ -0,0 +1,144 @@
...
+    TypeAndSiblings,            // or CapabilityAndSiblings (siblings = override the same capability)
+    TypeAndSiblingsAndSubtypes, // or CapabilityAndSiblingsAndOverridingCapabilities
+}
+
+impl Constraint {
+    compute_constraint_one_to_one_annotation!(compute_abstract, Annotation::Abstract, AnnotationAbstract);
+    compute_constraint_one_to_one_annotation!(compute_independent, Annotation::Independent, AnnotationIndependent);
+    compute_constraint_one_to_one_annotation!(compute_key, Annotation::Key, AnnotationKey);
+
+    pub fn compute_distinct<'a, A: Into<Annotation> + Clone + 'a>(
```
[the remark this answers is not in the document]
farost: Good point, I've created a task for it.

concept/type_/constraint.rs
```diff
@@ -0,0 +1,144 @@
...
+            .next();
+        return if cardinality.is_some() { cardinality } else { default };
+    }
+
+    compute_constraint_one_to_one_annotation!(compute_regex, Annotation::Regex, AnnotationRegex);
+    compute_constraint_one_to_one_annotation!(compute_cascade, Annotation::Cascade, AnnotationCascade);
+    compute_constraint_one_to_one_annotation!(compute_range, Annotation::Range, AnnotationRange);
+    compute_constraint_one_to_one_annotation!(compute_values, Annotation::Values, AnnotationValues);
+
+    pub fn inherited_constraint_validation_mode(
```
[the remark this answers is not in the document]
farost: Yeah, I was choosing between a static collection and map, but wanted to support dynamic collection for keys. Refactored, good point about `pub`, totally missed it!

concept/type_/type_manager/validation/operation_time_validation.rs
```diff
@@ -1239,15 +2823,381 @@ impl OperationTimeValidation {
...
+            cardinality.is_some()
+                || distinct.is_some()
+                || unique.is_some()
+                || regex.is_some()
+                || range.is_some()
+                || values.is_some(),
+            "At least one constraint should exist otherwise we don't need to iterate"
+        );
+
+        // TODO: It is EXCEPTIONALLY memory-greedy and should be optimized before a non-alpha release!
```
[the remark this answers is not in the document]
farost: https://github.com/typedb/typedb/issues/7138

concept/thing/attribute.rs
```diff
@@ -171,8 +178,13 @@ impl<'a> ThingAPI<'a> for Attribute<'a> {
         snapshot: &impl ReadableSnapshot,
         type_manager: &TypeManager,
     ) -> Result<Prefix, ConceptReadError> {
-        let value_type = type_.get_value_type(snapshot, type_manager)?.unwrap();
-        Ok(Self::Vertex::value_type_category_to_prefix_type(value_type.category()))
+        let value_type = type_.get_value_type(snapshot, type_manager)?;
```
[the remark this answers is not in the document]
farost: Rewritten another way

concept/thing/attribute.rs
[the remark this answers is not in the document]
farost: Rewritten to error if there is no value + add filtering on the caller side

concept/thing/thing_manager.rs
flyingsilverfin: That is true, I think we can get away without it, there is never an overlap? Owns is always an attribute type, and therefore can never overlap with Plays cardinality, since that's always RoleType. The only thing that might overlap is `[card][Relation][Plays | Relates][RoleType]` right? If this is a possible overlap, I think we might want to be consistent and add the `has/plays/relates` all the time IMO

concept/thing/thing_manager.rs
flyingsilverfin: You could make a constant in `//resources` which is something like `AD_HOC_INLINE_BYTES` and set that to `128`. This is basically the size of the array allocated for short-lived encode/decode. Better if we don't go to the heap much for these unless they're very large.

concept/thing/thing_manager.rs
farost: `Relation` <-> `RoleType` cannot have both `Plays` and `Relates` with each other (we can't play what we relate), so I don't see a reason to. But if you want to super safe, I can add it.

flyingsilverfin: Actually, we can play what we relate (in a different relation instance!)
```
define
relation linked-list, relates previous, plays linked-list:previous;
```

flyingsilverfin: I'm actually not sure we have a reason to ban the same relation playing a role in itself either :P

farost: I never thought about it, what a monster :'(
Okay, I am convinced now

farost: Done

concept/thing/thing_manager.rs
farost: Done, thx!

flyingsilverfin (approved):

PR #7139, opened 2024-08-20
Implement query pipelines

krishnangovindraj:
## Usage and product changes
We implement query pipelines as two versions: Read and Write pipelines. In this iteration we centralise the IR translation, type annotation, and compilation passes into a single top-level driver package: `//query`.

We also implement some larger refactors of the BlockContext, introducing new TranslationContext, PipelineContext, and VariableRegistry types. The `//compiler` package is now split into `//compiler/match|insert|delete`, with the intent that each one has an entry point to `compile()` a block of TypeQL IR into a `Program` for each type, along with possibly a set of type inference or type checkers to go with each type of program to be compiled.

Note: lots of refactoring still to follow, merging early to unblock waiting work.


ir/program/block.rs
```diff
@@ -106,37 +119,171 @@ impl FunctionalBlockBuilder {
...
             variable_names: HashMap::new(),
-            variable_declaration: HashMap::new(),
-            variable_names_index: HashMap::new(),
             variable_id_allocator: 0,
-            scope_id_allocator: 1, // `0` is reserved for ROOT
-            scope_parents: HashMap::new(),
             variable_categories: HashMap::new(),
             variable_optionality: HashMap::new(),
         }
     }
```
krishnangovindraj, line 138: BlockContext has been split to have a reference to a pipeline level `VariableRegistry`

query/tests/pipeline.rs
krishnangovindraj, line 1: These are broken. Check the executor tests instead.

executor/tests/pipelines.rs
```diff
@@ -0,0 +1,302 @@
...
+        rows.push(row.unwrap().clone().into_owned());
+    }
+    assert_eq!(rows.len(), 7);
+    for row in rows {
+        for value in row {
+            print!("{}, ", value);
+        }
+        println!()
+    }
+}
```
krishnangovindraj, line 304: This is a good starting point. I have a match & an insert running separately.

executor/pipeline/accumulator.rs
```diff
@@ -0,0 +1,120 @@
...
+    rows: Box<[(Box<[VariableValue<'static>]>, u64)]>,
+    next_index: usize,
+}
+
+impl<Snapshot: ReadableSnapshot + 'static> AccumulatedRowIterator<Snapshot> {
+    pub(crate) fn finalise(self) -> PipelineContext<Snapshot> {
+        // TODO: Ensure we have been consumed
+        debug_assert!(self.next_index >= self.rows.len());
+        self.context
+    }
```
krishnangovindraj, line 105: This has to return a result

executor/write/insert_executor.rs
krishnangovindraj, line 1: renamed to insert.rs but git didn't pick that up.

PR #7140, opened 2024-08-20
Refactor vertices conversion to reduce clones and clean up test code. Remove RootLabel name usages

farost:
## Usage and product changes
We reduce the amount of methods that consume `self` and force us to use `clone` for their usages:
* Move `fn vertex` from `TypeAPI` to `TypeVertexEncoding`
* Make some non-`into_``TypeEdgeEncoding` methods accept `&self` instead of `self`.

Moreover, we clean the test code that used `RootLabel`, switching to `Kind` instead.

### DIFF
--- a/answer/lib.rs
+++ b/answer/lib.rs
@@ -15,11 +15,13 @@ use concept::{
     },
 };
 use encoding::{
-    graph::type_::{vertex::TypeVertex, Kind},
+    graph::type_::{
+        vertex::{TypeVertex, TypeVertexEncoding},
+        Kind,
+    },
     value::value::Value,
     AsBytes,
 };
-use encoding::graph::type_::vertex::TypeVertexEncoding;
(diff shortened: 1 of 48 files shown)


answer/lib.rs
```diff
@@ -19,6 +19,7 @@ use encoding::{
     value::value::Value,
     AsBytes,
 };
+use encoding::graph::type_::vertex::TypeVertexEncoding;
```
farost, line 22: I moved `fn vertex(&self)` into TypeVertexEncoding. We already use `encoding` in many places in `concept` even if we were interested to reduce the amount of its usages on the higher logic's level, so it doesn't affect much.
It allows me to have both `into_vertex` and `vertex` (shall I rename every `fn vertex(&self)` to `fn as_vertex(&self)`?..) in the same trait.

farost:
Will revive it after one more pr next week.

PR #7144, opened 2024-08-27
Checks in traversal executors

dmitrii-ubskii:


executor/instruction/has_reverse_executor.rs
```diff
@@ -109,25 +106,30 @@ impl HasReverseExecutor {
...
+            checker,
         })
     }
 
-    pub(crate) fn get_iterator<Snapshot: ReadableSnapshot>(
+    pub(crate) fn get_iterator(
         &self,
-        snapshot: &Snapshot,
-        thing_manager: &ThingManager,
+        snapshot: &Arc<impl ReadableSnapshot + 'static>,
```
flyingsilverfin, line 115: Do you want to just take the `Arc` directly? You're going to have to clone it into the checker anyway right?

executor/instruction/mod.rs
```diff
@@ -290,14 +284,136 @@ impl TernaryIterateMode {
...
+        snapshot: &Arc<impl ReadableSnapshot + 'static>,
+        thing_manager: &Arc<ThingManager>,
+        row: &ImmutableRow<'_>,
+    ) -> Box<FilterFn<T>> {
+        let mut filters: Vec<Box<dyn Fn(&T::HktSelf<'_>) -> Result<bool, ConceptReadError>>> =
+            Vec::with_capacity(self.checks.len());
+        for check in &self.checks {
+            match *check {
+                CheckInstruction::Comparison { lhs, rhs, comparator: _comp } => {
+                    let lhs = self.extractors[&lhs];
```
flyingsilverfin, line 358: split naming

executor/instruction/has_reverse_executor.rs
dmitrii-ubskii: No, only checks that involve edges need to clone the arc.

PR #7145, opened 2024-08-27
3.0 Define and Redefine

farost:
## Usage and product changes
We add `define` and the first version of `redefine` queries to `TypeDB 3.0`, excluding functions and aliases. 

## Implementation
Definition (and redefinition) order of types (it can be changed) :
* types
* types' alias (can be moved up if anything depends on it later, like "set type alias" + "alias set annotation ..."
* types' value types (doesn't depend on anything)
* types' annotations (can potentially depend on value types)
* types' subs (expects all types to be already declared, can depend on annotations: `@abstract`)
* relates + annotations (expects all types. Annotation is a capability of a capability, doesn't depend on anything above)
* relates overrides (expects all relates to be previously declared)
* owns + annotations (same as for relates + annotations)
* owns overrides (same as for relates overrides)
* plays + annotations (expects all relates to be declared + same as for relates + annotations)
* plays overrides (same as for relates overrides)

Redefinition follows the same logic now, but:
* overrides update call doesn't wait for all the other capabilities' updates and is called right after annotations update;
* allows only one update per query (for simplicity: redefine requires much more work on its planner to execute queries in a smart way).

`define` and `redefine` queries use two common files: `definition_resolution` and `definition_status`. 
`definition_resolution` contains common functions to search for / wrap into errors type axioms, constraints, etc. It is directly used when we only need to search for a specific *something* and only wish for success. 
`definition_status` is used to get the status of *something* we search for: `DoesNotExist` / `ExistsSame` / `ExistsDifferent`. It is a simple and convenient (for both creating and refactoring) way to describe how to act in different queries like `define`/`redefine`/`undefine` based on the status of the definition. Of course, it uses `definition_resolution` as its basis.

`Ordering` is removed from `set_owns` in Concept API. `set` operations are idempotent, and providing ordering to this operation is risky, you always need to read ordering from storage when you call for this operation. Moreover, we need to worry about ordering's change, cover it with additional tests while not necessarily using it, etc.
Better solution: `set_owns` + `set_owns_ordering`. Ordering is set to `unordered` by default, which should be the most common case, and it shouldn't be a really excessive operation.

`Ordering` is still a part of `create_relates` as: 
* it is a `create`, a non-idempotent operation. We don't expect to call it twice for the same `relates`;
* it is generally a complex creation operation of both `relates` and `role types`, and it's a bit more reasonable to have it more configurable.
* we already (not for long) have another parameter `Cardinality` in this api call, so removing `Ordering` is only a half-way cleaning.
However, I'd only support the decision to remove `Ordering` from `create`s in the future, too. It can be done in a separate PR.


.factory/automation.yml
```diff
@@ -59,12 +59,12 @@ build:
 #      command: |
 #        bazel run @vaticle_dependencies//tool/bazelinstall:remote_cache_setup.sh
 #        bazel test //test/behaviour/connection/... --test_output=streamed
-    test-behaviour-concept:
-      image: vaticle-ubuntu-22.04
-      dependencies: [build]
-      command: |
-        bazel run @vaticle_dependencies//tool/bazelinstall:remote_cache_setup.sh
-        bazel test //tests/behaviour/concept/... --test_output=streamed
+#    test-behaviour-concept: # TODO: Turn it on when concept api and its tests are refactored based on overrides deletion
```
farost, line 62: We're going to change it a lot in the next PR, so supporting the current tests is not necessary, plus the locking issue still doesn't let us control the flow smoothly. I'll take responsibility of fixing the tests later.

farost:
Build will be green once I update refs from other repos.

query/define.rs
```diff
@@ -363,151 +516,442 @@ fn define_capabilities_relates(
...
+            .map_err(|source| DefineError::DefinitionResolution { source })?;
+
+        let definition_status =
+            get_relates_status(snapshot, type_manager, relation_type.clone(), &role_label, ordering)
                 .map_err(|source| DefineError::UnexpectedConceptRead { source })?;
-            if ordering != existing_ordering {
-                Err(DefineError::CreateRelatesModifiesExistingOrdering {
+        let (defined, is_new) = match definition_status {
+            DefinitionStatus::DoesNotExist => {
+                let init_cardinality = if let Some(typeql_cardinality) = capability
```
farost, line 527: Will be removed in the next PR when the init_cardinality is removed from the API!

query/define.rs
```diff
@@ -363,151 +516,442 @@ fn define_capabilities_relates(
...
+        let converted_for_relates = capability_convert_and_validate_annotation_definition_need(
+            snapshot,
+            type_manager,
+            &relation_label,
+            relates.clone(),
+            annotation.clone(),
+        );
+        match converted_for_relates {
+            Ok(Some(relates_annotation)) => {
+                // New relates should set Cardinality on initialization
```
farost, line 604: Will be removed in the next PR when the init_cardinality is removed from the `create_relates` API!

tests/behaviour/steps/concept/type_/owns.rs
```diff
@@ -40,34 +40,6 @@ pub async fn set_owns(
...
             &tx.thing_manager,
             attr_type,
-            Ordering::Unordered,
-        );
-        may_error.check_concept_write_without_read_errors(&res);
-    });
-}
-
-#[apply(generic_step)]
-#[step(expr = "{root_label}\\({type_label}\\) set owns: {type_label}[]{may_error}")]
```
flyingsilverfin, line 50: Is removing this intentional?

query/definition_resolution.rs
flyingsilverfin, line 1: I think this file isn't `definition_resolution` but instead `schema_resolution`. Calling it with "definition" ties it too closely to 'define/redefine/undefine', which it's perfectly fine to use this in any other part of Query (if we wanted to). Additionally, you can resolve both Types or Definitions (structs or functions). So i think `schema_resolution` is a bit clearer!

query/definition_resolution.rs
```diff
@@ -0,0 +1,474 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+// TODO: not sure if these go elsewhere, or are useful in lower level packges outside //query
```
flyingsilverfin, line 7: This is a good point - maybe it should move elsewhere later! Not right now though I think!

concept/type_/mod.rs
```diff
@@ -210,6 +240,10 @@ pub trait OwnerAPI<'a>: TypeAPI<'a> {
         Ok(self.get_owns(snapshot, type_manager)?.iter().find(|owns| owns.attribute() == attribute_type).cloned())
     }
 
+    get_with_overridden! {
```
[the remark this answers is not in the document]
farost: Sure, let's not think about the next changes in this pr. It is purely isolated and correct for the old days state, which is reflected in tests.

concept/type_/annotation.rs
```diff
@@ -555,6 +555,22 @@ impl AnnotationCategory {
             | AnnotationCategory::Values => true,
         }
     }
+
+    pub fn boolean(&self) -> bool {
```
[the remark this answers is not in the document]
farost: Renamed, reversed

encoding/graph/definition/struct.rs
```diff
@@ -57,15 +57,27 @@ impl StructDefinition {
...
         }
     }
+
+    pub fn get_field(&self, field_name: &str) -> Option<&StructDefinitionField> {
+        let field_idx = if let Some(id) = self.field_names.get(field_name) {
+            id
+        } else {
+            return None;
+        };
+        self.fields.get(field_idx)
```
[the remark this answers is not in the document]
farost: Fixed, you're right

encoding/graph/definition/struct.rs
```diff
@@ -77,3 +89,9 @@ impl DefinitionValueEncoding for StructDefinition {
         Some(Bytes::copy(bincode::serialize(&self).unwrap().as_slice()))
     }
 }
+
+impl StructDefinitionField {
+    pub fn same(&self, optional: bool, value_type: ValueType) -> bool {
```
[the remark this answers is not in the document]
farost: What is a `description`? `Same` is `same`, not `eq`, that's why it's not an `eq` trait implementation!

tests/behaviour/steps/concept/type_/owns.rs
farost: Yes, concept api tests should be as much one-to-one command-to-apicall as possible. Using this shortcut is not correct anymore.

query/definition_resolution.rs
farost: I thought that `definition` is generally... any definition of anything in the schema? 
When I `define`, I create a `definition`. When I `redefine`, I update a `definition`. When I `undefine`, I remove a `definition`. 
When I `match`, I may be interested in existing `definitions` -- something that has been `define`d!

encoding/graph/definition/struct.rs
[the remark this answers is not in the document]
flyingsilverfin: `has_optionality_and_value_type`

query/definition_resolution.rs
farost: Decided to go with `definable_`

encoding/graph/definition/struct.rs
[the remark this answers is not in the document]
farost: Done

PR #7146, opened 2024-08-29
Refactoring pipelines

flyingsilverfin:
## Usage and product changes

We simplify the traits and data structures around Pipeline execution. We also optimise their execution by avoiding extra allocations and Vec<Vec<...>> constructs.

We also rename `ImmutableRow` to `MaybeOwnedRow`, which allows getting a mutable reference to the Row whenever it is actually owned. `Row` remains purely mutable reference to set of row data owned elsewhere.


executor/batch.rs
```diff
@@ -120,105 +116,91 @@ impl LendingIterator for BatchRowIterator {
 }
 
 #[derive(Debug)]
-pub struct Row<'a> {
-    row: &'a mut [VariableValue<'static>],
-    multiplicity: &'a mut u64,
+pub struct Batch {
```
flyingsilverfin, line 119: New unlimited length Batch construct

PR #7149, opened 2024-08-31
Introduce periodic fsync to guarantee commits are persisted to disk

krishnangovindraj:
## Usage and product changes
Commits guarantee that changes have been persisted to the disk (using `fsync`) before changes are visible and the commit is acknowledged. The sync is done periodically and committing threads must wait for a signal that the sync is complete. This avoids the performance penalty of each commit doing its own sync, while providing guarantees that committed data is not lost in the event of an OS crash.

## Implementation
* Introduce an `fsync` thread in the WAL.
* Introduce a `request_sync` API to the WAL, which returns a `mpsc::receiver` which will receive a single `()` when the next fsync is completed.

### DIFF
--- a/durability/wal.rs
+++ b/durability/wal.rs
@@ -88,6 +88,10 @@ impl WAL {
     pub fn previous(&self) -> DurabilitySequenceNumber {
         DurabilitySequenceNumber::from(self.next_sequence_number.load(Ordering::Relaxed) - 1)
     }
+
+    pub fn sync_all(&self) {
+        self.files.write().unwrap().sync_all()
+    }
 }
 
 impl DurabilityService for WAL {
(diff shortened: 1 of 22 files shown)


durability/wal.rs
```diff
@@ -481,6 +508,90 @@ impl<'a> Iterator for FileRecordIterator<'a> {
...
+        let context = FsyncThreadContext {
+            files,
+            shutting_down: AtomicBool::new(false),
+            signalling: [Mutex::new(Vec::new()), Mutex::new(Vec::new())],
+            current_signal: AtomicU8::new(0),
+        };
+        Self { handle: None, context: Arc::new(context) }
+    }
+
+    fn subscribe_to_next_flush(&self) -> mpsc::Receiver<()> {
```
flyingsilverfin, line 536: `subscribe_to_next_sync`?

storage/storage.rs
```diff
@@ -221,14 +222,15 @@ impl<Durability> MVCCStorage<Durability> {
...
         let validated_commit = self
             .isolation_manager
             .validate_commit(commit_sequence_number, commit_record, &self.durability_client)
             .map_err(|error| Durability { name: self.name.to_owned(), source: error })?;
-
-        match validated_commit {
+        let result = match validated_commit {
             ValidatedCommit::Write(write_batches) => {
-                // Write to the k-v storage
+                sync_notifier.recv().unwrap(); // Ensure WAL is persisted inserting to RocksDB
```
flyingsilverfin, line 232: `ensure WAL is persisted <before> inserting to storage`

flyingsilverfin (approved):

PR #7150, opened 2024-09-02
Clean up duplicate test setup and implement GRPC Query streaming pt 1

flyingsilverfin:
## Usage and product changes

Refactor tests that were all redeclaring storage and manager setup and loading into various reusable `<package>/tests/test_utils_<package>.rs`. 

We also extend the implementation of the GRPC service for concept answers, and implement the query streaming architecture:

1. spawn a dedicated thread to execute a query, which writes messages to submit into a Blocking MPSC queue with default capabity 32 (or whatever the user's transaction query prefect configuration is).
2. spawn a task to take messages from the MPSC queue and insert them into the GRPC output sender (which is another MPSC queue). This is meant to submit a batch of messages at a time, and next time the stream is polled, another tokio task is created to submit the next batch.
3. Precompute size is controlled by the size of the blocking MPSC queue.


answer/lib.rs
```diff
@@ -50,6 +56,19 @@ impl Type {
         }
     }
 
+    pub fn get_label<'a>(
```
flyingsilverfin, line 59: Unfortunately it seems like we can't use the inner as an `&dyn TypeAPI`, otherwise this would be extremely easy!
However, most of the time we know the particular type we unwrap to -- this is mostly used in error reporting.

PR #7157, opened 2024-09-10
Rework constraints and capabilities

farost:
## Motivation
Our old approach to annotations and capabilities wasn't correct. We considered types and capabilities inheriting annotations from their supertypes and overridden capabilities, drawing a parallel between subtyping and overriding. However, none of these considerations were correct:
* Thinking of annotations as constraints for types creates a conceptual mess, where one annotation can play the role of two other annotations (`@key` as `@unique` and `@card(1..1)`), where we can't understand what a default cardinality is, where we struggled to set default annotations, and where we have to introduce strange validation methods like "inheritable_together", "inheritable_below", etc.
* In reality, capabilities should not inherit anything from their overrides. They should simply comply with the constraints of the capabilities of their super interface types, and you don't need to override anything. This means that the "inheritance" of annotations/constraints for capabilities depends on each specific object type, depending on whether it has a capability for super interface type or not.

This PR addresses this general issue, additionally refactoring most of the existing validations in Concept API. 

## Implementation
1. Remove `override`
Previously, to `override` a capability meant:
* ensure there is a subtyping between these two interface types;
* "hide" the interface type from the overridden capability. 

The first step is just validation for `owns` and `plays`, so we only need a "hide" operation for them. 

At the same time, we decided that it's logical to be able to query for hidden capabilities (including `relates`, i.e. overridden role types). So we don't hide them, but make them `abstract`: the object type has this capability, but cannot create instances based on it. However, it can have capabilities for sub interface types, and these sub interface types will be affected by the constraints of the abstract capability. And there is no need to override or specialise anything, it works out of the box!

However, we still want to use `as` keyword for `relates`, because it's the only way of `role types` subtyping, so we call it `specialisation` now, which is an exclusive operation for `relates`, consisting of the two steps mentioned for `override` above, In details:
* set the specialising `role type`'s supertype;
* set an `@abstract` annotation to the specialised `relates`.

In fact, "specialisation" is a good term for "capability which complies for constraints of supercapability of a super interface type, but also specialises something (a subtype or a constraint, narrowing one of them)". Which means that:
* `*** relates father as parent` -> `father` specialises `parent`
* `supertype owns age @range(1..100); subtype owns age @range(10..50);` -> `owns(subtype, age)` specialises `owns(supertype, age)` (cannot be applied to `relates` as `relates` also creates a `role type`!)

Triggered changes:
* removed all the interfaces for `get_override` / `get_overriding` / etc.
* removed a number of validations,
* `get_owns` returns only the most "actual" `owns` for an `interface_type` (for `entity supertype owns name; entity subtype sub supertype, owns name @regex("...");`, `subtype.get_owns` will only return `Owns(subtype, name)`). However, `subtype.get_owns_with_specialises` will return all the "duplicates" if it is for some reason needed.

2. Introduce `Constraint`.
A `constraint` consists of a `ConstraintDescription` (what this constraint is about, I'm open to better names) and a `Source` (where it is declared, it is needed as we cannot think about a constraint in isolation from its source). To understand the connection between `ConstraintDescription` and `Source`, there is a enum `ConstraintScope`. Examples:
* `abstract` constraint with source `supertype` means that we cannot have instances of `supertype`, but it does not affect its subtypes. It is what `ConstraintScope::SingleInstanceOfType` is about.
* `regex` constraint with source `supertype` means that all instances of `supertype` or its subtypes should comply for this set regex. It is what `ConstraintScope::SingleInstanceOfTypeOrSubtype` is about.
* `card(0, 2)` constraint with source `owns_of_supername` means that an object of type `owner_of_supername` can only have up to 2 instances of `supername` or its subtypes `name***`. In the meantime, if `owns_of_name` has its own `card(1, 1)` constraint for `owner_of_supername`, the combination of these two constraints means "this object type can have up to 2 `supername`s, but 1 of them **has** to be a `name` (instance of `name` or its subtypes)".  However, it won't affect the second object of type `owner_of_supername`: each object of this type has its own `count`. That's what `ConstraintScope::AllInstancesOfSiblingTypeOrSubtypes` is about, where `Sibling` is "both interface types i1 and i2 are capabilities of the same capability type (owns/plays/relates) of the same object type (e.g. they are owned by the same type, they are played by the same type)";
* `unique` constraint with source `owns_of_supername` means that for each value `X`, there can only be one object of type `owner_of_supername` or its subtypes, which has a single instance type `supername` or its subtype with value `X`. It affects all `$owner isa owner_of_supername` and `$attribute isa supername`. That's what `ConstraintScope::AllInstancesOfTypeOrSubtypes` is about.

Theoretically, constraints can appear in any way we want. Right now they appear via:
* setting an annotation;
* default values based on the absence of annotations overriding these constraints and ordering.

Triggered changes:
* types and capabilities now have `annotations_declared` without `annotations`. They are not inherited, because we simply annotate a type, and this annotation can create a constraint, a trigger, a cute label, a cute nothing;
* types and capabilities now have `constraints` instead of `annotations`, which can be "inherited". For types, it's direct inheritance (I am aware of constraints of my parent if these constraints affect me (namely, if it's not `ConstraintScope::SingleInstanceOfType`)). For capabilities, they affect each other based on the interface type inheritance only if they are set on one object type;
* types constraints can only be queried for a type, and it will include all the constraints: ones from declared annotations, ones from default values, ones affecting this type from its supertypes; 
* capabilities constraints can be queried purely for a capability (capability.get_constraints()). It will include both constraints from declared annotations and default constraints! There is no need to query for declared constraints, you can use annotations for these purposes (not to spam with useless APIs, can be added in the future if there is a need);  
* capabilities constraints can be queried regarding a specific object type. More specifically, object type has constraints for each `interface type` it has capabilities for. These constraints will include "inherited" constraints that affect this capability from other capabilities of super interface types (including all specialised capabilities);
* a type or a capability can now have multiple constraints of the same `ConstraintCategory` or even `ConstraintDescription` (multiple `@card(1, 1)`), but these constraints will have different `sources`. And these constraints will need to be validated together via a beautiful loop (can be optimized if we are 100% sure that specific constraints are narrowed, skipped for now). 
* all the validations including annotations and constraints are reworked. For example, we don't check narrowing for annotations for capabilities, as they can only affect each other on specific object types (it can be done, but the reasoning for it is still unclear, and this validation takes time: there is a task for it). 

Additionally:
* Concrete object types can have capabilities for abstract interface types.
* Consume https://github.com/typedb/typedb/pull/7140 (remove `RootLabel`, refactor `to_something(self)`, consuming `self` instead of using its reference.


concept/type_/type_manager.rs
```diff
@@ -2659,144 +2348,130 @@ impl TypeManager {
...
+        if let Some(supertype) = role_type.get_supertype(snapshot, self)? {
+            self.unset_role_type_supertype(snapshot, thing_manager, role_type)?;
 
-            OperationTimeValidation::validate_interface_change_supertype_does_not_corrupt_capabilities_overrides_transitive::<Plays<'static>>(
-                snapshot,
-                relates.role(),
-                None, // new_interface_supertype
-            )
-                .map_err(|source| ConceptWriteError::SchemaValidation { source })?;
+            if supertype.get_subtypes(snapshot, self)?.is_empty() {
```
farost, line 2425: Here, if we unset `as role`, and it is not specialised by anyone, we automatically unset `@abstract` from `relates`. However, it could be set manually, so I'd prefer to leave it `@abstract` and ask the user to undefine `@abstract` manually, even if they haven't defined it, using only `as` instead. I've left this code here for reference. @flyingsilverfin thoughts?

A potential "clean" solution here would be to generate an `@abstract` constraint based on the fact that `relates` is specialised, and save it as we previously did. But it is indeed redundant (and I don't want to use `relates.role().get_supertype().is_some()`...)

compiler/expression/block_compiler.rs
```diff
@@ -138,7 +138,7 @@ fn resolve_type_for_variable<'a, Snapshot: ReadableSnapshot>(
             .iter()
             .map(|type_| match type_ {
                 Type::Attribute(attribute_type) => attribute_type
-                    .get_value_type(context.snapshot, context.type_manager)
+                    .get_value_type_without_source(context.snapshot, context.type_manager)
```
farost, line 141: I was torn apart between 
`get_value_type() -> ValueType` + `get_value_type_with_source -> (ValueType, AttributeType)`
and
`get_value_type() -> (ValueType, AttributeType)` + `get_value_type_without_source -> ValueType`

For now, I've chosen the second option, because all the other non-`_defined` interfaces return both the value and the source. Plus, we duplicate the `type_manager`'s call (`get_value_type()` -> `type_manager.get_value_type()`), and `without_source` is just an additional method. Still, not sure if it's pleasant to use.

compiler/match_/inference/type_seeder.rs
```diff
@@ -720,7 +720,7 @@ impl BinaryConstraint for Has<Variable> {
             } // It can't be another type => Do nothing and let type-inference clean it up
         };
         attribute
-            .get_owns(seeder.snapshot, seeder.type_manager)?
```
farost, line 723: We used to have 
`get_owns` and `get_owns_declared` for `AttributeType`. However, `_declared` is confusing here, because you don't fully understand what are the additional `ObjectType`s returned in `get_owns`. So it's just called `get_owner_types` now (all owner types for this attribute type), and `get_owns_declared` is now `get_owns` of this `attribute_type` (not `ObjectType`s, but `Owns`!)

concept/thing/thing_manager.rs
```diff
@@ -1113,7 +1123,7 @@ impl ThingManager {
                 }
                 let subtypes = relation_type.get_subtypes_transitive(snapshot, self.type_manager())?;
                 once(&relation_type).chain(subtypes.into_iter()).try_for_each(|type_| {
-                    let is_cascade = type_.is_cascade(snapshot, self.type_manager())?;
+                    let is_cascade = true; // TODO: Always consider cascade now, can be changed later.
```
farost, line 1126: Waiting for the new cascade =)

concept/thing/thing_manager.rs
```diff
@@ -1986,8 +1903,7 @@ impl ThingManager {
     ///
     /// TODO:
     /// Call index regenerations when cardinality changes in schema
```
farost, line 1905: Not sure if this index should work the same!

concept/type_/annotation.rs
```diff
@@ -104,14 +129,16 @@ impl AnnotationCardinality {
...
 
     pub fn narrowed_correctly_by(&self, other: &Self) -> bool {
-        self.value_satisfies_start(other.start()) && self.value_satisfies_end(other.end())
+        // We allow other.min < self.min and other.max > self.max.
+        // The only limitation is other.min <= self.max.
+        self.value_satisfies_end(Some(other.start()))
     }
 
-    fn value_satisfies_start(&self, value: u64) -> bool {
+    pub fn value_satisfies_start(&self, value: u64) -> bool {
```
farost, line 137: Will probably need to call these validations separately

concept/type_/constraint.rs
```diff
@@ -4,139 +4,544 @@
...
+}
+
+impl ConstraintDescription {
+    pub fn from_annotation(annotation: Annotation) -> HashSet<Self> {
+        static KEY_CONSTRAINTS: [ConstraintDescription; 2] = [
+            ConstraintDescription::Unique(AnnotationKey::UNIQUE),
+            ConstraintDescription::Cardinality(AnnotationKey::CARDINALITY),
+        ];
+        match annotation {
+            Annotation::Abstract(annotation) => HashSet::from([ConstraintDescription::Abstract(annotation)]),
```
farost, line 73: Cannot make it 'static =(

farost:
Respective typeql pr: https://github.com/typedb/typeql/pull/364

concept/type_/constraint.rs
```diff
@@ -4,139 +4,544 @@
...
+    Independent,
+    Unique,
+    Cardinality,
+    Regex,
+    Range,
+    Values,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum ConstraintDescription {
```
farost, line 55: Generally, I like what I've got with constraints, but passing `ConstraintDescription` around and just calling "vallidate it, please" is not statically checked by Rust (in a function to check Regex constraints, I could potentially call "get all Card constraints, validate" and it would be fine. Same for general getters of Constraints. I'm sure it could be refactored to consider static checks, but I didn't have that much motivation to do so because we're covered by BDD tests from all sides like crazy, and they should do the job. Still, something important to highlight

tests/behaviour/steps/concept/type_/relation_type.rs
```diff
@@ -176,14 +137,20 @@ pub async fn relation_role_unset_override(
...
+        let res =
+            relates.unset_specialise(Arc::get_mut(&mut tx.snapshot).unwrap(), &tx.type_manager, &tx.thing_manager);
         may_error.check_concept_write_without_read_errors(&res);
     });
 }
 
 #[apply(generic_step)]
-#[step(expr = r"relation\({type_label}\) get roles {contains_or_doesnt}:")]
-pub async fn relation_roles_contain(context: &mut Context, type_label: Label, contains: ContainsOrDoesnt, step: &Step) {
+#[step(expr = r"relation\({type_label}\) get relates {contains_or_doesnt}:")]
```
farost, line 147: If I say `get role(...)`, I believe that I am talking to the root role for my relation (`relation:role` is the label of this `role`, not `superrelation:role`!)
And for explicity of checks, you cannot say
`relation(subrelation) get role(superrole)` if `superrole`'s label is `superrelation:superrole` in Concept API tests.
Instead, you have to work with `relation(superrelation) get role(superrole)`!

At the same time, we want to check what's inside relates, so now we have `get relates contains`, etc.

concept/type_/constraint.rs
```diff
@@ -4,139 +4,544 @@
...
+    )]);
+
+    if let Some(default_distinct) = Owns::get_default_distinct(ordering) {
+        constraints.insert(CapabilityConstraint::new(ConstraintDescription::Distinct(default_distinct), source));
+    }
+
+    constraints
+}
+
+pub(crate) fn get_plays_default_constraints<CAP: Capability<'static>>(
```
farost, line 473: For simplicity, unordered `owns` and `relates` have a `distinct` constraint (so we could check "if distinct then..." without worrying about ordering). I wasn't sure what to do with plays:
* don't add `distinct` because it can't generally have ordering and `@distinct`, we just know that it is **always** distinct without constraints;
* add `distinct` and start adding `distinct` to all `Capability` (trait from `type_`) to highlight when a capability is distinct. Creates redundancy, but it will be consistent.

I can also remove default `distinct` from other capabilities, but it will create additional risks!

concept/type_/type_manager/validation/operation_time_validation.rs
```diff
@@ -1363,621 +846,643 @@ impl OperationTimeValidation {
...
-                &updated_overrides,
-                &mut validation_errors,
-            )
-            .map_err(SchemaValidationError::ConceptRead)?;
-
-            if let Some(error) = validation_errors.first() {
-                return Err(error.clone());
-            }
-        }
+    // TODO: Capabilities constraints narrowing checks are currently disabled
```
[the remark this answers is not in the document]
farost: I'm commenting out these BDD tests as well (instead of deleting them completely)
I have a task for it, and I'll prepare a plan for these validations, estimate it and we can prioritize it after the first alpha.

concept/type_/mod.rs
```diff
@@ -136,34 +129,117 @@ pub trait TypeAPI<'a>: ConceptAPI<'a> + TypeVertexEncoding<'a> + Sized + Clone +
...
+        other: Self,
+    ) -> Result<bool, ConceptReadError> {
+        Ok(self == &other || self.is_subtype_transitive_of(snapshot, type_manager, other)?)
+    }
+
+    fn inheritance_cmp<'m>(
+        &self,
+        snapshot: &impl ReadableSnapshot,
+        type_manager: &'m TypeManager,
+        other: Self,
```
flyingsilverfin, line 210: What's the expected 'inheritance_cmp' if they're not equal but in siblings? :D

concept/type_/owns.rs
```diff
@@ -51,66 +52,63 @@ impl<'a> Owns<'a> {
...
         snapshot: &impl ReadableSnapshot,
         type_manager: &TypeManager,
-    ) -> Result<bool, ConceptReadError> {
-        type_manager.get_owns_is_distinct(snapshot, self.clone())
+    ) -> Result<HashSet<CapabilityConstraint<Owns<'static>>>, ConceptReadError> {
+        type_manager.get_owns_distinct_constraints(snapshot, self.clone().into_owned())
     }
 
-    pub fn get_constraint_regex(
+    pub fn is_distinct(
```
flyingsilverfin, line 74: Instead of `is_distinct` maybe `has_constraint_distinct`? To keep it consistent?

encoding/layout/infix.rs
```diff
@@ -27,7 +27,7 @@ pub enum Infix {
     PropertyLabel,
     PropertyValueType,
     PropertyOrdering,
-    PropertyOverride,
+    PropertyIsAbstract,
```
flyingsilverfin, line 30: Why is abstract a PropertyIsAbstract and also an PropertyAnnotationAbstract below?

executor/instruction/relates_reverse_executor.rs
```diff
@@ -112,8 +112,7 @@ impl RelatesReverseExecutor {
                 let relates: Vec<_> = self
                     .role_relation_types
                     .keys()
-                    .map(|role| role.as_role_type().get_relates(&**snapshot, type_manager))
-                    .map_ok(|set| set.to_owned())
+                    .map(|role| role.as_role_type().get_relates_root(&**snapshot, type_manager))
```
flyingsilverfin, line 115: Hmm are we sure about this one? It's not making sense to me - pull me in to explain please

flyingsilverfin (changes requested):
So i've not reviewed the massive files (validation definitions), more the invocation points of them! -- will trust your code's clean-ness & tests for that detailed stuff!

concept/type_/entity_type.rs
```diff
@@ -288,12 +302,88 @@ impl<'a> OwnerAPI<'a> for EntityType<'a> {
         type_manager.get_entity_type_owns(snapshot, self.clone().into_owned())
     }
 
-    fn get_owns_overrides<'m>(
+    fn get_owns_with_specialised<'m>(
```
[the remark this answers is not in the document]
farost: British it is, then

concept/type_/constraint.rs
farost: If we introduce a constraint by default, we are not forced to use it. If you don't see a need in it, then I'm happy to go without `distinct`ness for plays.
I was just thinking that `constraints` can be a tool for describing capabilities/types characteristics, and `plays` would just be always `distinct`. It's my issue to overgeneralize, that's why I am asking =)

concept/type_/annotation.rs
```diff
@@ -226,6 +251,11 @@ impl AnnotationRegex {
     pub fn value_type_valid(value_type: Option<ValueType>) -> bool {
         matches!(value_type, Some(ValueType::String))
     }
+
+    // Can try to implement the check, but allow everything now!
+    pub fn narrowed_correctly_by(&self, _other: &Self) -> bool {
```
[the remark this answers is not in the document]
farost: I'm not sure if it's something that we're going to address, and not sure if we need to check narrowing, considering the headache caused by capabilities' annotations. I'll put a TODO, but it's something TBD

concept/type_/attribute_type.rs
```diff
@@ -89,16 +94,12 @@ impl<'a> TypeAPI<'a> for AttributeType<'a> {
         Self::from_vertex(vertex).unwrap()
     }
 
-    fn vertex(&self) -> TypeVertex<'_> {
```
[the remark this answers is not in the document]
farost: It's not a non-trait implementation. It's `impl<'a> TypeVertexEncoding<'a>`
Consumed from the forgotten https://github.com/typedb/typedb/pull/7140

concept/type_/constraint.rs
```diff
@@ -4,139 +4,548 @@
...
+pub(crate) fn get_cardinality_constraints<C: Constraint<T>, T: Hash + Eq>(
+    constraints: impl IntoIterator<Item = C>,
+) -> HashSet<C> {
+    filter_by_constraint_category!(constraints.into_iter(), Cardinality).collect()
+}
+
+pub(crate) fn get_cardinality_constraint<'a, CAP: Capability<'static>>(
+    capability: CAP,
+    constraints: impl IntoIterator<Item = &'a CapabilityConstraint<CAP>>,
+) -> Option<CapabilityConstraint<CAP>> {
```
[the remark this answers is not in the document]
farost: That's the point of this API: you have a collection of constraints belonging to different capabilities, and you search for the constraint of a specific capability! For example, it's used to filter out inherited constraints, Can also be used for any other reason.

Capabilities themselves get their constraints using these methods. 
* The whole `constraint.rs` file is for defining `Constraint` and macros/functions to work with them, filter their collections, etc.
* `type_manager.rs` gets constraints from `type_reader.rs`, filters them using `constraint.rs`
* `owns.rs`/`plays.rs`/`relates.rs`/..../`entity_type.rs` call for `type_manager.rs` methods in their concept api public methods
* managers have data and schema validations, and sometimes these methods from `constraint.rs` are useful not only for concept api calls, but for complex algorithms of finding interesting constraints.

This API doesn't need to be tight and it doesn't even try to seem tight: it returns an `Option` because of the clear reason.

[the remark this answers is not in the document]
farost: I get your idea overall though and may have similar issues somewhere else, and maybe these methods are just badly formed so you misunderstood their purpose.

concept/type_/mod.rs
farost: Don't quite get the case of 

> not equal but in siblings

"siblings" (based on my unofficial and potentially banned notation) are interface types with the same root parent and with the same object type. And I actually do not care about "siblings" in this method. 

You might be confused by `std::cmp::Ordering::Equal` in case if types are not equal, and I could return "None" in this case. However, comparators are usually simple like that, and used for ordering, not for hashing to store in buckets. Sooo we'd just leave these types somewhere together, surrounded by `Less` and `Greater`, which is the point of sorting. Might have bad side effects if misused, maybe there is a better way to express it.
It is used in `type_get_constraints_closest_source` in `constraint.rs`

concept/type_/owns.rs
farost: It is consistent, you have multiple interfaces for different purposes:

* `get_constraint_abstract` returns an `abstract` constraint if exists.
* `is_abstract` (in impl Capability, that's why you don't see it here and it may confuse you) calls this getter and checks `is_some`.

* `get_constraints_abstract` returns a collection of `distinct` constraints. `abstract` constraint is stored only once (because it's scoped to only a single instance of a single type) for a type, while you can have multiple `distinct` constraints, that why it's `Option` vs `HashSet`.
* `is_distinct` does a job similar to `is_abstract`.

You call the latter when you're interested if it's just abstract or distinct overall. You ask for a constraint to get its description and source.

encoding/layout/infix.rs
farost: Beeeecaaaauuseee I forgot to remove it! It's not needed, thanks

executor/instruction/relates_reverse_executor.rs
farost: You might be right, maybe I need to collect all the relates here, not only the root ones. Let's discuss how this thing is used

concept/type_/mod.rs
flyingsilverfin: Address

farost: Done

executor/instruction/relates_reverse_executor.rs
farost: This change doesn't affect the old logic, so we merge it this way, but we'll revisit the logic later

PR #7167, opened 2024-09-23
Fix greedy planner: variables are produced immediately

dmitrii-ubskii:

Fixes a crash on a malformed intersection plan.

## Motivation

Previously the greedy planner could produce a double intersection. E.g., when the planner was given constraints `$x has $a`, `$x isa T`, `$a isa A`, could then attempt to lump all the constraints into a single intersection iterator producing `($x, $a)`.
The lowering of the plan to the executor, however, cannot support this intersection (there is no corresponding executor, nor is one possible in the current architecture), so it would e.g. intersect `$x has $a` and `$x isa T` to produce `$x` and `$a`, then traverse `$a isa A`. As `$a` was not registered as an input to the third constraint, it is unbound. The executor then attempts to produce what is essentially a cartesian product of `$a * $a`, encounters multiple values for `$a`, and crashes.
 
## Implementation

Mark variables as produced eagerly as the constraint is registered. This does mean that the greedy planner is currently incapable of producing intersections.

### DIFF
--- a/compiler/match_/instructions/mod.rs
+++ b/compiler/match_/instructions/mod.rs
@@ -8,12 +8,15 @@
 
 use std::{collections::HashMap, ops::Deref};
 
+use answer::{variable::Variable, Type};
 use ir::pattern::{
     constraint::{Comparator, Comparison, Constraint, ExpressionBinding, FunctionCallBinding, IsaKind, SubKind},
-    IrID, Vertex,
+    IrID, ParameterID, Vertex,
 };
+use itertools::Itertools;
 
-use crate::match_::planner::pattern_plan::InstructionAPI;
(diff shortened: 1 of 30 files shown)


compiler/match_/instructions/mod.rs
```diff
@@ -268,18 +271,101 @@ impl<ID: IrID + Copy> InstructionAPI<ID> for ConstraintInstruction<ID> {
     }
 }
 
+#[derive(Debug, Clone)]
+pub enum CheckVertex<ID> {
```
dmitrii-ubskii, line 275: This will be the executor package version of Vertex so we don't have to pass the already resolved `Label`s around.

flyingsilverfin (approved):

PR #7169, opened 2024-09-24
Introduce reduce stages for pipelines

krishnangovindraj:
Introduce reduce stages for pipelines to enable grouped aggregates.

## Motivation
Introduce reduce stages for pipelines to enable grouped aggregates.

## Implementation
- Implement grouped reduce stages for pipelines.


executor/pipeline/reduce.rs
```diff
@@ -0,0 +1,205 @@
...
+    while let Some(result) = iterator.next() {
+        grouped_reducer.accept(&result?, &context)?;
+    }
+    Ok(grouped_reducer.finalise())
+}
+
+struct GroupedReducer {
+    input_group_positions: Vec<VariablePosition>,
+    grouped_aggregates: HashMap<Vec<VariableValue<'static>>, Vec<ReducerImpl>>,
+    reused_group: Vec<VariableValue<'static>>,
```
flyingsilverfin, line 75: is this `current_group`?

executor/pipeline/reduce.rs
```diff
@@ -0,0 +1,205 @@
...
+    sample_reducers: Vec<ReducerImpl>,
```
flyingsilverfin, line 76: Maybe just `reducers`?

executor/pipeline/reduce.rs
```diff
@@ -0,0 +1,205 @@
...
+    let mut grouped_reducer = GroupedReducer::new(program);
+    while let Some(result) = iterator.next() {
+        grouped_reducer.accept(&result?, &context)?;
+    }
+    Ok(grouped_reducer.finalise())
+}
+
+struct GroupedReducer {
+    input_group_positions: Vec<VariablePosition>,
+    grouped_aggregates: HashMap<Vec<VariableValue<'static>>, Vec<ReducerImpl>>,
```
flyingsilverfin, line 74: Please be careful with terminology... `aggregate` is 2.x terminology

executor/pipeline/reduce.rs
```diff
@@ -0,0 +1,205 @@
...
+        VariableValue::Value(value) => Some(value.clone().into_owned()),
+        VariableValue::Thing(Thing::Attribute(attribute)) => {
+            // As long as these are trivial, it's safe to unwrap
+            let snapshot: &Snapshot = &context.snapshot;
+            let value = attribute.get_value(snapshot, &context.thing_manager).unwrap();
+            Some(value.clone().into_owned())
+        }
+        _ => unreachable!(),
+    }
+}
```
flyingsilverfin, line 155: Does this have to be an owned value you return? This will induce extra Clones you don't really need I think?

krishnangovindraj: One is owned, the other isn't. I can wrap it in `MaybeOwns`. I can also likely make it return the primitive.

executor/pipeline/reduce.rs
krishnangovindraj: Fair. What's the word for the output of a reduction?

executor/pipeline/reduce.rs
krishnangovindraj: Unsure. I should probably re-create the reducers from the program instead but I figured a clone would be faster. This is an always empty set that can be cloned.

executor/pipeline/reduce.rs
krishnangovindraj: No, it's just a reused vec.

krishnangovindraj (commented):
replies

krishnangovindraj (commented):
Flush pending comments

executor/pipeline/reduce.rs
flyingsilverfin: reduction

executor/pipeline/reduce.rs
flyingsilverfin: Should be possible to make this not owned in any branch

executor/pipeline/reduce.rs
flyingsilverfin: `unitialised_reduce_executors`

ir/program/block.rs
```diff
@@ -234,6 +237,26 @@ impl VariableRegistry {
             VariableOptionality::Optional => true,
         }
     }
+
+    // TODO: This is out of place
```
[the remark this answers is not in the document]
flyingsilverfin: delete TODO

PR #7170, opened 2024-09-25
Add support for ISO timezones; finalise datetime & duration support

dmitrii-ubskii:


encoding/value/date_time_tz_bytes.rs
dmitrii-ubskii, line 1: Timezone handling moved to its own module

encoding/value/timezone.rs
```diff
@@ -0,0 +1,731 @@
...
+}
+
+impl TimeZone {
+    pub(crate) fn encode(self) -> [u8; 4] {
+        match self {
+            TimeZone::IANA(tz) => encode_u32(encode_tz(tz)),
+            TimeZone::Fixed(fixed) => {
+                // from UTC-23:59:59 to UTC+23:59:59 <=> -86399..=86399
+                encode_i32(fixed.local_minus_utc())
+            }
```
dmitrii-ubskii, line 57: The named timezones are given values in `0..NUM_TZS`, while fixed offsets are in the range of `0x8000_0000_0000_000 ± 86399`.

ir/translation/literal.rs
```diff
@@ -169,13 +181,54 @@ impl FromTypeQLLiteral for NaiveDateTime {
...
+                    let seconds_decimal = (time_part.seconds.as_ref())
+                        .map(|seconds| parse_primitive::<Decimal>(&seconds.value))
+                        .transpose()?
+                        .unwrap_or_default();
+
+                    nanos += seconds_decimal.integer_part() as u64 * NANOS_PER_SEC;
+
+                    let nanos_decimal = (seconds_decimal - seconds_decimal.integer_part()) * NANOS_PER_SEC;
+                    debug_assert_eq!(nanos_decimal.fractional_part(), 0);
+                    nanos += nanos_decimal.integer_part() as u64;
```
dmitrii-ubskii, line 226: It would be cleaner to do `(seconds_decimal * NANOS_PER_SEC).integer_part()`, but that can overflow `i64` while still being in range for `u64`.

server/service/answer.rs
```diff
@@ -242,6 +241,9 @@ fn encode_date_time(date_time: NaiveDateTime) -> typedb_protocol::value::Datetim
     typedb_protocol::value::Datetime { seconds: date_time.and_utc().timestamp_millis(), nanos: date_time.nanosecond() }
 }
 
-fn encode_time_zone(timezone: Tz) -> typedb_protocol::value::datetime_tz::Timezone {
-    typedb_protocol::value::datetime_tz::Timezone::Named(timezone.name().to_string())
+fn encode_time_zone(timezone: TimeZone) -> typedb_protocol::value::datetime_tz::Timezone {
+    match timezone {
+        TimeZone::IANA(tz) => typedb_protocol::value::datetime_tz::Timezone::Named(tz.name().to_string()),
+        TimeZone::Fixed(fixed) => typedb_protocol::value::datetime_tz::Timezone::Offset(fixed.local_minus_utc()),
```
dmitrii-ubskii, line 247: Note: protocol should probably specify that the offset is in seconds. @flyingsilverfin

storage/keyspace/raw_iterator.rs
```diff
@@ -66,6 +79,7 @@ impl Seekable<[u8]> for DBIterator {
                 Ordering::Less => {
                     self.item.take();
                     self.iterator.seek(key);
+                    self.item = peek_item(&self.iterator); // repopulate `item` to prevent advancing the underlying iterator
```
dmitrii-ubskii, line 82: Bug fix! When item is empty, that means that the kv pair the Rocks iterator _is currently pointing to_  has been yielded to the user, and the underlying iterator needs to be advanced before peeking. That means that after a seek, we always attempted to advance the iterator once more.

flyingsilverfin (approved):

PR #7172, opened 2024-09-26
Let match stages accept inputs

dmitrii-ubskii:

We now support `match` stages in the middle of a pipeline, which enables queries like this:
```
insert
    $org isa organisation;
match
    $p isa person, has age 10;
insert
    (group: $org, member: $p) isa membership;
```


executor/tests/compile_execute.rs
dmitrii-ubskii, line 1: Using define and insert queries in the compile-execute tests rather than manual setup.

ir/pattern/conjunction.rs
```diff
@@ -37,6 +40,21 @@ impl Conjunction {
     pub fn nested_patterns(&self) -> &[NestedPattern] {
         &self.nested_patterns
     }
+
+    pub fn captured_variables(&self, scope_context: &ScopeContext) -> impl Iterator<Item = Variable> + '_ {
+        iter::empty() // TODO
+    }
+
```
dmitrii-ubskii, line 47: This will be used to identify inputs for subpatterns in a match.

flyingsilverfin (approved):

PR #7182, opened 2024-10-03
Negation

dmitrii-ubskii:

We add support for negations in match queries:
```php
match $x isa person, has name $a; not { $a == "Jeff"; };
```

Note: unlike in 2.x, TypeDB 3.x supports nesting negations, which leads to a form of "for all" operator: 
```php
match
    $super isa set;
    $sub isa set;
    not { # all elements of $sub are also elements of $super:
        (item: $element, set: $sub) isa set-membership;  # there is no element in $sub...
        not { (item: $element, set: $super) isa set-membership; }; # ... that isn't also an element of $super
    };
```


compiler/match_/inference/type_seeder.rs
```diff
@@ -114,12 +114,9 @@ impl<'this, Snapshot: ReadableSnapshot> TypeSeeder<'this, Snapshot> {
...
     ) -> Result<(), TypeInferenceError> {
-        self.local_variables(context, tig.conjunction.scope_id()).for_each(|var| {
-            let vertex = Vertex::Variable(var);
-            if let Some(parent_annotations) = parent_vertices.get(&vertex) {
-                tig.vertices.insert(vertex, parent_annotations.clone());
-            }
-        });
+        for (vertex, parent_annotations) in parent_vertices.iter() {
+            tig.vertices.insert(vertex.clone(), parent_annotations.clone());
+        }
```
dmitrii-ubskii, line 119: `local_variables` only considers variables declared at this depth (i.e. not referenced in any scope above). We do, however, still need to be able to access the type annotations of the captured variables.

compiler/match_/planner/pattern_plan.rs
dmitrii-ubskii, line 1: The planning and lowering was moved around and refactored

flyingsilverfin (approved):

PR #7190, opened 2024-10-07
Publish multi-architecture docker images

krishnangovindraj:
Publish arm64 docker images. Previously, docker would fall-back to running the x86_64 images, which sometimes resulted in a JVM crash loop.

## Motivation
Support docker on arm64 platforms

## Implementation
Uses the approach similar to 3.0:  
1. We import two ubuntu base images, one for x86_64 and one for arm64
2. We unpack the typedb-linux-all-{x86_64, arm64} on top of the corresponding base image.
3. We push each to `vaticle/typedb:<VERSION>-<ARCH>`
4. We use docker cli to create the multi-arch `vaticle/typedb:<VERSION>` manifest from these and push it.
5. We repeat 4, but for the 'latest' tag

### DIFF
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -20,208 +20,224 @@ build:
       command: |
         bazel run @vaticle_dependencies//factory/analysis:dependency-analysis
   correctness:
-    build:
-      image: vaticle-ubuntu-22.04
-      command: |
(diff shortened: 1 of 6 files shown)


krishnangovindraj:
Will reopen from my fork

PR #7208, opened 2024-10-15
Add database name validation for database creation

farost:
We add validation of names for created databases. Now, all database names should be valid TypeQL identifiers. 

## Implementation
We reuse TypeQL `is_valid_identifier` to have a single source of truth for identifier formats.

### DIFF
--- a/tests/behaviour/steps/params.rs
+++ b/tests/behaviour/steps/params.rs
@@ -459,7 +459,7 @@ impl Value {
     ];
     const DATE_FORMAT: &'static str = "%Y-%m-%d";
 
-    const FRACTIONAL_ZEROES: usize = 18;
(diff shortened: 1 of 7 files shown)


tests/behaviour/steps/params.rs
```diff
@@ -459,7 +459,7 @@ impl Value {
     ];
     const DATE_FORMAT: &'static str = "%Y-%m-%d";
 
-    const FRACTIONAL_ZEROES: usize = 18;
+    const FRACTIONAL_ZEROES: usize = 19;
```
farost, line 462: I was stealing from here for the sake of the driver, so noticed and fixed in both places

farost:
Needs this PR to work https://github.com/typedb/typeql/pull/372

flyingsilverfin (approved):

PR #7214, opened 2024-10-17
Fix document answers streaming for fetch

farost:
We fix document answers streaming for fetch in order to complete the first version of `fetch` queries execution.

## Motivation
There were incorrect assertions that didn't allow the server to process correct requests.

### DIFF
--- a/server/service/transaction_service.rs
+++ b/server/service/transaction_service.rs
@@ -998,7 +998,6 @@ impl TransactionService {
         if pipeline.has_fetch() {
             let initial_response = StreamQueryResponse::init_ok_documents(Read);
             Self::submit_response_sync(sender, initial_response);
-
             let (mut iterator, context) =
                 unwrap_or_execute_and_return!(pipeline.into_documents_iterator(interrupt.clone()), |(err, _)| {
                     Self::submit_response_sync(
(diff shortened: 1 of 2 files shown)


server/service/transaction_service.rs
```diff
@@ -1299,19 +1298,36 @@ impl QueryStreamTransmitter {
...
+    }
+
+    async fn send_on_stream_done(
+        response_sender: &Sender<Result<typedb_protocol::transaction::Server, Status>>,
+        req_id: Uuid,
+        rows: Vec<typedb_protocol::ConceptRow>,
+        documents: Vec<typedb_protocol::ConceptDocument>,
+    ) -> ControlFlow<(), ()> {
+        debug_assert!(rows.is_empty() || documents.is_empty());
+        if !rows.is_empty() {
```
farost, line 1325: There was just
```
if !rows.is_empty() { send_rows } Continue(())
```
before the first versions of fetch, so I guess it is correct.

server/service/transaction_service.rs
```diff
@@ -1274,7 +1273,7 @@ impl QueryStreamTransmitter {
                         return Break(());
                     }
                     StreamQueryResponse::StreamDoneOk() => {
-                        if let Break(()) = Self::send_rows(response_sender, req_id, rows).await {
```
farost, line 1277: We always tried to send rows even if it was a `fetch` query processing.

dmitrii-ubskii (approved):

PR #7216, opened 2024-10-18
Basic streaming functions

krishnangovindraj:
Implement function execution for non-recursive stream functions which return streams **only**. 
**Recursive functions & Non-stream functions will throw unimplemented.**

These can currently only be used from the preamble. Sample:
```
           with
            fun get_ages($p_arg: person) -> { age }:
            match
                $p_arg has age $age_return;
            return {$age_return};

            match
                $p isa person;
                $z in get_ages($p);
```

## Motivation
Implement some form of function execution end-to-end

## Implementation
* Refactors PatternExecutor to recursively execute subpatterns
* Implements "Steps" for nested patterns & reducer stages.


executor/read/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+    let executors = step_executor::create_executors_for_match(
+        snapshot,
+        thing_manager,
+        function_registry,
+        match_executable,
+        &mut HashSet::new(),
+    )?;
+    Ok(PatternExecutor::new(executors))
+}
+
```
krishnangovindraj, line 41: Goes away when pipelines are fully consumed.

krishnangovindraj (commented):
self review skim, without looking at PatternExecutor

executor/read/pattern_executor.rs
```diff
@@ -0,0 +1,188 @@
...
+                    self.control_stack.push(ControlInstruction::ExecuteImmediate(next_instruction_index));
+                }
+                StepExecutors::Branch(_) => self
+                    .control_stack
+                    .push(ControlInstruction::MapRowBatchToRowForSubQuery(next_instruction_index, batch.into_iterator())),
+                StepExecutors::CollectingStage(collecting_stage) => {
+                    collecting_stage.prepare(batch);
+                    self.control_stack.push(ControlInstruction::CollectingStage(next_instruction_index.clone()));
+                }
+                StepExecutors::ReshapeForReturn(_) => todo!(),
```
krishnangovindraj, line 183: I may not need this because the call knows the mapping.

executor/read/pattern_executor.rs
```diff
@@ -0,0 +1,188 @@
...
+                ControlInstruction::StreamCollected(index, mut collected_iterator) => {
+                    if let Some(batch) = collected_iterator.batch_continue()? {
+                        self.control_stack.push(ControlInstruction::StreamCollected(index.clone(), collected_iterator));
+                        self.prepare_next_instruction_and_push_to_stack(context, index.next(), batch)?;
+                    }
+                }
+                ControlInstruction::Yield(batch) => {
+                    return Ok(Some(batch));
+                }
+            }
```
krishnangovindraj, line 157: I'm tempted to refactor this into a `batch_continue` which produces the `Option<Batch>` + a `transition` function based on the current `ControlInstruction`

compiler/executable/match_/planner/function_plan.rs
```diff
@@ -24,9 +35,21 @@ impl ExecutableFunctionRegistry {
...
+
+    pub fn get(&self, function_id: FunctionID) -> &ExecutableFunction {
+        match &function_id {
+            FunctionID::Schema(id) => self.schema_functions.get(id).unwrap(),
+            FunctionID::Preamble(id) => self.preamble_functions.get(id).unwrap(),
+        }
+    }
+
+    // TODO: Find all references, update them and remove this.
+    pub fn TODO__empty() -> Self {
```
flyingsilverfin, line 50: Hygene clean up

executor/read/step_executor.rs
```diff
@@ -0,0 +1,251 @@
...
+            Ok(vec![StepExecutors::CollectingStage(step)])
+        }
+        ExecutableStage::Reduce(reduce_executable) => {
+            let step = CollectingStageExecutor::new_reduce(
+                PatternExecutor::new(previous_stage_steps),
+                reduce_executable.clone(),
+            );
+            Ok(vec![StepExecutors::CollectingStage(step)])
+        }
+        ExecutableStage::Insert(_) | ExecutableStage::Delete(_) => todo!("Or unreachable?"),
```
flyingsilverfin, line 249: I think having a `todo` with a question `unreachable?` makes no sense :D Either it is a TODO or it is unreachable - I think these kinds of questions marks should be resolved before merge

flyingsilverfin: There's another one in plan.rs

executor/match_executor.rs
```diff
@@ -34,18 +33,16 @@ impl MatchExecutor {
...
+        function_registry: &ExecutableFunctionRegistry,
     ) -> Result<Self, ConceptReadError> {
-        let step_executors = match_executable
-            .steps()
-            .iter()
-            .map(|step| StepExecutor::new(step, snapshot, thing_manager))
-            .try_collect()?;
-
         Ok(Self {
+            entry: TODO_REMOVE_create_executors_for_match(
```
flyingsilverfin, line 39: Hmmm clean up?

executor/pipeline/pipeline.rs
```diff
@@ -114,6 +117,7 @@ impl<Snapshot: ReadableSnapshot + 'static> Pipeline<Snapshot, ReadPipelineStage<
         snapshot: Arc<Snapshot>,
         thing_manager: Arc<ThingManager>,
         variable_registry: &VariableRegistry,
+        executable_functions: Arc<ExecutableFunctionRegistry>,
```
flyingsilverfin, line 120: We should add to functions Write pipelines!

executor/read/step_executor.rs
```diff
@@ -0,0 +1,251 @@
...
+
+impl StepExecutors {
+    pub(crate) fn unwrap_immediate(&mut self) -> &mut ImmediateExecutor {
+        match self {
+            StepExecutors::Immediate(step) => step,
+            _ => unreachable!(),
+        }
+    }
+
+    pub(crate) fn unwrap_nested_pattern_branch(&mut self) -> &mut NestedPatternExecutor {
```
flyingsilverfin, line 50: In your naming convention here, the `unwrap` is named after the enum, so this should be `unwrap_branch`!

executor/read/step_executor.rs
```diff
@@ -0,0 +1,251 @@
...
+        match self {
+            StepExecutors::Branch(step) => step,
+            _ => unreachable!(),
+        }
+    }
+
+    pub(crate) fn unwrap_collecting_stage(&mut self) -> &mut CollectingStageExecutor {
+        match self {
+            StepExecutors::CollectingStage(step) => step,
+            _ => unreachable!(),
```
flyingsilverfin, line 60: I think the convention for `unwrap`s is to `panic` btw

executor/read/pattern_executor.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    pub(super) fn batch_continue(
+        &mut self,
+        context: &ExecutionContext<impl ReadableSnapshot + 'static>,
+        interrupt: &mut ExecutionInterrupt,
+    ) -> Result<Option<FixedBatch>, ReadExecutionError> {
+        if self.control_stack.len() == 0 {
+            debug_assert!(self.control_stack.len() > 0);
+        }
```
flyingsilverfin, line 82: Wait what does this do?

executor/read/pattern_executor.rs
```diff
@@ -0,0 +1,188 @@
...
+        while self.control_stack.last().is_some() {
+            let Self { control_stack, executors } = self;
+            // TODO: inject interrupt into Checkers that could filter out many rows without ending as well.
+            if let Some(interrupt) = interrupt.check() {
+                return Err(ReadExecutionError::Interrupted { interrupt });
+            }
+            let mut step = control_stack.pop().unwrap();
```
flyingsilverfin, line 89: unneeded mut (intellij suggestion)

executor/read/pattern_executor.rs
```diff
@@ -0,0 +1,188 @@
...
+
+            match step {
+                ControlInstruction::Start(batch) => {
+                    self.prepare_next_instruction_and_push_to_stack(context, InstructionIndex(0), batch)?;
+                }
+                ControlInstruction::ExecuteImmediate(index) => {
+                    let found = executors[index.0].unwrap_immediate().batch_continue(context, interrupt)?;
```
flyingsilverfin, line 96: Seomtimes just to avoid coming up with an arbitrary name i'll split this kind of code into two different lines:
```
let executor = executors[index.0].unwrap_immediate();
if let Some(bach) = executor.batch_continue(context, interrupt)?;
```

In this case I think it is much more readable!

executor/read/pattern_executor.rs
```diff
@@ -0,0 +1,188 @@
...
+                    if let Some(batch) = found {
+                        control_stack.push(ControlInstruction::ExecuteImmediate(index.clone()));
```
flyingsilverfin, line 98: If you make `BranchIndex` and `InstructionIndex` have `Copy`, then you don't need these redundant `.clone()`s that clutter up the code!

executor/read/pattern_executor.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+pub(super) enum ControlInstruction {
+    Start(FixedBatch),
+
+    ExecuteImmediate(InstructionIndex),
+
+    MapRowBatchToRowForSubQuery(InstructionIndex, FixedBatchRowIterator),
+    ExecuteSubQuery(InstructionIndex, BranchIndex),
```
flyingsilverfin, line 39: Nb inconsistent naming: `SubQuery` here vs `Branch` in other places

executor/read/nested_pattern_executor.rs
```diff
@@ -0,0 +1,267 @@
...
+    fn prepare_and_map_input(&mut self, input: &MaybeOwnedRow<'_>) -> MaybeOwnedRow<'static>;
+    fn map_output(&mut self, input: &MaybeOwnedRow<'_>, subquery_result: Option<FixedBatch>) -> SubQueryResult;
+}
+
+pub(super) enum SubQueryResult {
+    Retry(Option<FixedBatch>),
+    Done(Option<FixedBatch>),
+}
+
+pub(super) struct NegationMapper;
```
flyingsilverfin, line 105: It feels like NegationMapper shouldn't use the default `map_output`, which clones the input unnecessarily

executor/read/step_executor.rs
krishnangovindraj: I suppose the question is: Can you use functions in a write pipeline?
In which case this will likely have to be split out into two code paths - One for the entry pattern and one for functions.

compiler/executable/match_/planner/function_plan.rs
krishnangovindraj: Will go in the next PR when fetch have their functions.

executor/read/pattern_executor.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    CollectingStage(InstructionIndex),
+    StreamCollected(InstructionIndex, CollectedStageIterator),
```
flyingsilverfin, line 42: I think naming the second to imply the 'Done'ness of the collection a bit more precisely could help? `CollectingStageOutput` (tie it directly to `CollectingStage`)

executor/read/nested_pattern_executor.rs
```diff
@@ -0,0 +1,267 @@
...
+pub(super) enum NestedPatternExecutor {
+    Disjunction(Vec<NestedPatternBranch>, IdentityMapper),
+    Negation([NestedPatternBranch; 1], NegationMapper),
+    InlinedFunction([NestedPatternBranch; 1], InlinedFunctionMapper),
+    Offset([NestedPatternBranch; 1], OffsetMapper),
+    Limit([NestedPatternBranch; 1], LimitMapper),
+}
+
+// Bad name because I want to refactor later
+pub(super) enum SubQueryResultMapperEnumMut<'a> {
```
flyingsilverfin, line 22: Therefore to match this should be `NestedPatternResultMapper...`

PR #7222, opened 2024-10-23
Fix and add to CI concept BDD tests

farost:
We fix data commit error collection and checks and add `concept` package BDD tests to CI.

### DIFF
--- a/database/transaction.rs
+++ b/database/transaction.rs
@@ -144,11 +144,7 @@ impl<D: DurabilityClient> TransactionWrite<D> {
 
     pub fn try_commit(self) -> Result<(), DataCommitError> {
         let mut snapshot = Arc::into_inner(self.snapshot).ok_or_else(|| DataCommitError::SnapshotInUse {})?;
-        self.thing_manager.finalise(&mut snapshot).map_err(|errs| {
-            // TODO: send all the errors, not just the first
-            let error = errs.into_iter().next().unwrap();
-            DataCommitError::ConceptWriteErrorsFirst { typedb_source: error }
-        })?;
+        self.thing_manager.finalise(&mut snapshot).map_err(|source| DataCommitError::ConceptWriteErrors { source })?;
         drop(self.type_manager);
         snapshot.commit().map_err(|err| DataCommitError::SnapshotError { typedb_source: err })?;
         Ok(())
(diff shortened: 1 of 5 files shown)


farost:
Requires https://github.com/typedb/typedb-behaviour/pull/306, will not merge until I update the ref.

database/transaction.rs
```diff
@@ -144,11 +144,7 @@ impl<D: DurabilityClient> TransactionWrite<D> {
 
     pub fn try_commit(self) -> Result<(), DataCommitError> {
         let mut snapshot = Arc::into_inner(self.snapshot).ok_or_else(|| DataCommitError::SnapshotInUse {})?;
-        self.thing_manager.finalise(&mut snapshot).map_err(|errs| {
-            // TODO: send all the errors, not just the first
-            let error = errs.into_iter().next().unwrap();
-            DataCommitError::ConceptWriteErrorsFirst { typedb_source: error }
-        })?;
+        self.thing_manager.finalise(&mut snapshot).map_err(|source| DataCommitError::ConceptWriteErrors { source })?;
```
farost, line 147: @flyingsilverfin Is there anything else needed to do here? Maybe I don't get the idea of this TODO.

tests/behaviour/steps/connection/transaction.rs
```diff
@@ -122,12 +122,18 @@ pub async fn transaction_commits(context: &mut Context, may_error: params::MayEr
...
-                if let DataCommitError::ConceptWriteErrors { source: errors, .. } = error {
-                    for error in errors {
-                        may_error.check_concept_write_without_read_errors::<()>(&Err(error))
+                match error {
+                    DataCommitError::ConceptWriteErrors { source: errors, .. } => {
+                        for error in errors {
+                            may_error.check_concept_write_without_read_errors::<()>(&Err(error))
+                        }
+                    }
+                    DataCommitError::ConceptWriteErrorsFirst { typedb_source } => {
```
farost, line 131: It's not used now. Do we still want to preserve it in the enum?

database/transaction.rs
flyingsilverfin: Basically: we have this stack trace system, where if you provide a `typedb_source`, we can print it in a stack trace with useful details. However, this only works with 1 error, not many (otherwise it will be a stack trace tree) - keep the old error and TODO please!

farost: Ah... I'll extend the todo, because it just felt like we need to change the return type...

farost: Done

tests/behaviour/steps/connection/transaction.rs
farost: It is used again, nvm

flyingsilverfin (approved):

PR #7224, opened 2024-10-25
Implement tabling machinery for cyclic functions

krishnangovindraj:
Introduces machinery needed to support finding the fixed-point of cyclic function calls. Cyclic functions now run, and return an incomplete set of results followed by an error. It is possible that the planner chooses a plan.

## Motivation
Reasonable waypoint towards full support.

## Implementation
Implements state for the tabled-functions as well as executors for tabled function calls.

### DIFF
--- a/compiler/executable/fetch/executable.rs
+++ b/compiler/executable/fetch/executable.rs
@@ -105,7 +105,7 @@ fn compile_some(
             Ok(FetchSomeInstruction::SingleAttribute(*position, attribute_type))
         }
         AnnotatedFetchSome::SingleFunction(function) => {
-            let compiled = compile_function(statistics, available_functions, function)
+            let compiled = compile_function(statistics, available_functions, function, false)
                 .map_err(|err| FetchCompilationError::AnonymousFunctionCompilation { typedb_source: Box::new(err) })?;
             Ok(FetchSomeInstruction::SingleFunction(compiled))
         }
@@ -114,7 +114,7 @@ fn compile_some(
             Ok(FetchSomeInstruction::Object(Box::new(compiled)))
         }
         AnnotatedFetchSome::ListFunction(function) => {
-            let compiled = compile_function(statistics, available_functions, function)
+            let compiled = compile_function(statistics, available_functions, function, false)
                 .map_err(|err| FetchCompilationError::AnonymousFunctionCompilation { typedb_source: Box::new(err) })?;
             Ok(FetchSomeInstruction::ListFunction(compiled))
         }
--- a/compiler/executable/fetch/executable.rs
+++ b/compiler/executable/fetch/executable.rs
@@ -21,6 +21,7 @@ use crate::{
     },
     VariablePosition,
 };
+use crate::executable::function::FunctionTablingType;
 
 pub struct ExecutableFetch {
     pub object_instruction: FetchObjectInstruction,
@@ -105,7 +106,7 @@ fn compile_some(
             Ok(FetchSomeInstruction::SingleAttribute(*position, attribute_type))
         }
         AnnotatedFetchSome::SingleFunction(function) => {
-            let compiled = compile_function(statistics, available_functions, function, false)
+            let compiled = compile_function(statistics, available_functions, function, FunctionTablingType::Untabled)
                 .map_err(|err| FetchCompilationError::AnonymousFunctionCompilation { typedb_source: Box::new(err) })?;
             Ok(FetchSomeInstruction::SingleFunction(compiled))
         }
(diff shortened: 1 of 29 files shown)


compiler/executable/fetch/executable.rs
```diff
@@ -105,7 +105,7 @@ fn compile_some(
             Ok(FetchSomeInstruction::SingleAttribute(*position, attribute_type))
         }
         AnnotatedFetchSome::SingleFunction(function) => {
-            let compiled = compile_function(statistics, available_functions, function)
+            let compiled = compile_function(statistics, available_functions, function, false)
```
dmitrii-ubskii, line 108: Please no bare `bool`s

compiler/executable/function.rs
```diff
@@ -75,3 +86,97 @@ fn compile_return_operation(
...
+        if cycle_detection[function_id] == TablingRequirement::Unexplored {
+            determine_tabling_requirements_impl(functions, &mut cycle_detection, function_id)
+        }
+    }
+    cycle_detection
+        .into_iter()
+        .map(|(function_id, tabling_requirement)| match tabling_requirement {
+            TablingRequirement::KnownTabled => (function_id, true),
+            TablingRequirement::KnownUntabled => (function_id, false),
+            TablingRequirement::Unexplored | TablingRequirement::UnknownInStack => unreachable!(),
```
farost, line 105: Let's put descriptive messages into all `unreachable`s we have to share to context why it is so. Otherwise, it will be really tough to understand what went wrong when anybody else steps on it in 1 year.

executor/error.rs
```diff
@@ -17,5 +18,8 @@ typedb_error!(
         CreatingIterator(3, "Error creating iterator from {instruction_name} instruction.", instruction_name: String, ( source: ConceptReadError )),
         AdvancingIteratorTo(4, "Error moving iterator (by steps or seek) to target value.", ( source: ConceptReadError )),
         ExpressionEvaluate(5, "Error evaluating expression", ( source: ExpressionEvaluationError )),
+        TabledFunctionLockError(6, "The table function lock was poisoned", function_id: FunctionID, arguments: MaybeOwnedRow<'static>),
```
dmitrii-ubskii, line 21: When can that happen that isn't an immediate `abort`?

executor/read/pattern_executor.rs
```diff
@@ -41,10 +48,13 @@ pub(super) enum ControlInstruction {
     ExecuteImmediate(InstructionIndex),
 
     MapRowBatchToRowForNested(InstructionIndex, FixedBatchRowIterator),
-    ExecuteNested(InstructionIndex, BranchIndex, NestedPatternResultMapper),
+    ExecuteNested(InstructionIndex, BranchIndex, NestedPatternResultMapper, MaybeOwnedRow<'static>),
+
+    TabledCall(InstructionIndex), // TODO: Use a FunctionMapper
```
dmitrii-ubskii, line 53: I'd prefer if these were struct-like variants with named fields

dmitrii-ubskii (approved):

compiler/executable/match_/planner/plan.rs
```diff
@@ -444,6 +443,7 @@ impl<'a> ConjunctionPlanBuilder<'a> {
                 self.graph.variable_index[variable]
             })
             .collect();
+        // TODO: This is just a mock
```
farost, line 446: ... and what (when?) do we want to do with it in the end? 

It's a common issue with codebases to put TODOs "for the next week" and abandon them for ages. After some time, it's really hard to restore the context. I'd try to be more descriptive for other developer's sake

farost: There are better examples of your todos in other files. Let's put a direct command to action / a desired result: "Make it ***", "implement ***", "should use ***"

executor/error.rs
krishnangovindraj: Can't. May as well unwrap it then?

farost (approved):
Comments about todos, etc.

compiler/executable/function.rs
```diff
@@ -46,11 +46,17 @@ pub enum ExecutableReturn {
...
+    Tabled,
+    Untabled,
+}
+
 pub(crate) fn compile_function(
     statistics: &Statistics,
     schema_functions: &ExecutableFunctionRegistry, // Can't have preamble in them when you're compiling functions
     function: AnnotatedFunction,
-    is_tabled: bool,
+    is_tabled: FunctionTablingType,
```
farost, line 59: Feels like `is_tabled` should be called differently now, t oo

krishnangovindraj: Good point. This is actually a boolean. I might shift to `pub struct IsFunctionTabled(bool)`

farost: Choosing the right approach might actually be a good question for @dmitrii-ubskii. In C++, we used to put comments near the boolean params passed into functions and force the vertical alignment of params (blocking some of the auto-formatters to rearrange it):
```
person.yawn(
    duration,
    false // isLoud
);
```
I don't know what's the common approach in Rust.

farost: Creating structs / enums / whatever for every boolean we want to visually annotate can be annoying.

PR #7227, opened 2024-10-28
Green match BDD redux

dmitrii-ubskii:

We resolve the remaining flaky failures in match BDDs.

## Implementation

- ensure sort variable for the sorted tuple iterator always comes first,
- fix incorrect tuple positioning in some executors,
- disallow intersection over unsorted executors in planner.

### DIFF
--- a/executor/instruction/has_executor.rs
+++ b/executor/instruction/has_executor.rs
@@ -99,10 +99,9 @@ impl HasExecutor {
         let owner = has.owner().as_variable().unwrap();
         let attribute = has.attribute().as_variable().unwrap();
 
-        let output_tuple_positions = if iterate_mode.is_inverted() {
-            TuplePositions::Pair([Some(attribute), Some(owner)])
-        } else {
-            TuplePositions::Pair([Some(owner), Some(attribute)])
+        let output_tuple_positions = match iterate_mode {
+            BinaryIterateMode::Unbound => TuplePositions::Pair([Some(owner), Some(attribute)]),
+            _ => TuplePositions::Pair([Some(attribute), Some(owner)]),
         };
 
         let checker = Checker::<(Has<'_>, _)>::new(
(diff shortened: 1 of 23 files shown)


executor/read/immediate_executor.rs
```diff
@@ -395,7 +388,7 @@ impl IntersectionExecutor {
 
         let input_row = self.input.as_mut().unwrap().peek().unwrap().as_ref().map_err(|&err| err.clone())?;
         for &position in &self.outputs_selected {
-            if position.as_usize() < input_row.len() {
+            if position.as_usize() < input_row.len() && !input_row.get(position).is_empty() {
```
krishnangovindraj, line 391: Wait, is there ever a case where this check can fail?

executor/instruction/has_executor.rs
```diff
@@ -99,10 +99,9 @@ impl HasExecutor {
         let owner = has.owner().as_variable().unwrap();
         let attribute = has.attribute().as_variable().unwrap();
 
-        let output_tuple_positions = if iterate_mode.is_inverted() {
-            TuplePositions::Pair([Some(attribute), Some(owner)])
-        } else {
-            TuplePositions::Pair([Some(owner), Some(attribute)])
+        let output_tuple_positions = match iterate_mode {
+            BinaryIterateMode::Unbound => TuplePositions::Pair([Some(owner), Some(attribute)]),
+            _ => TuplePositions::Pair([Some(attribute), Some(owner)]),
```
krishnangovindraj, line 104: For the `HasExecutor` (rest. the `HasReverseExectutor), what exactly does the TuplePosition convey? 
I think I heard you say the sort variable is always in the first position? Does it tell us how to parse the two parts of the key?

executor/read/immediate_executor.rs
dmitrii-ubskii: Yes. In a disjunction, one branch can set row positions as A: 0, B: 1, C: 2, but the other branch may choose traversal such that it produces the vvariables in the order A, C, B. In that case, when we get to the last step, the input will be `0: A, 1: None, 2: C`, and the `1: B` assigned by the iterator will be (attempted to be) overwritten by the `1: None` from the input.

executor/instruction/has_executor.rs
dmitrii-ubskii: This associates the positions of the variable values in the output of the tuple iterator with their positions in the output row.

flyingsilverfin (approved):

PR #7231, opened 2024-10-31
Introduce 3.0 undefine queries

farost:
We introduce `undefine` queries with the new "targeted" syntax to enable safe and transparent schema concept undefinitions. 

If you want to undefine the whole type, you can just say:
```
undefine 
  person;
```

If you want to undefine a capability from **somewhere**, use `undefine **capability** from **somewhere**`. 
It consistently works with `owns` (`person` is preserved, only the `owns` is undefined):
```
undefine
  owns name from person;
```
annotations of `owns` (`person owns name` is preserved, only `@regex(...)` (you don't need to specify the regex's argument) is undefined):
```
undefine
  @regex from person owns name;
```
and any other capability, even specialisation:
```
undefine
  as parent from fathership relates father;
```

Want to undefine multiple concepts in one go? Worry not!
```
undefine
  person;
  relates employee from employment;
  @regex from name;
  @values from email;
```

The error messages in all definition queries (`define`, `redefine`, and `undefine`) were enhanced, and the respective `query/language` BDD tests were introduced in the CI with `concept` unit tests.

Additional fixes: 
* answers of `match` capability queries like `match $x owns $y`, `match $x relates $y`, and `$match $x plays $y` now include transitive capabilities;
* `define` lets you write multiple declarations of an undefined type, specifying its kind **anywhere**, even in the last mention of this type;
* failures in schema queries for schema transactions no longer lead to freezes of the following opened transactions;
* no more crashes on long string attributes deletion with existing `has` edges;

## Implementation

The `undefine` feature implementation follows the inherited architecture and logic of `define` and `redefine` queries: 
* traverse the parsed undefinables (in reversed order to `define` to start from "leaves") 
* for types, check their definition status (doesn't exist at all / exists as requested / exists different (different sub, different ordering) -> return an error or do undefine
* for capabilities of types, find types and error if not found -> check capabilities definition status -> return an error or do undefine
* for capabilities of capabilities, find types and error if not found -> find capabilities and error if not found or different -> check capabilities's capabilities definition status -> return an error or do undefine

For additional fixes, additional comments are left on top of the code when needed.


executor/instruction/owns_executor.rs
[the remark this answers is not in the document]
farost: repaste

farost, line 1: I refactored all capabilities executors to use a tuple of ObjectType and InterfaceType instead of capabilities.

There was an issue that we saved iterators of duplicated, say, Owns, with the same declared owner of this owns for multiple subowner sub owner and subsubowner sub subowner and subowner2 sub owner, etc. This way, we didn't returned inherited, transitive capabilities.

I don't want to change Concept API to return "not stored" constraints: it is consistent in the way of returning only declared things. And it's too much work now with no clear benefit.

We already use tuples for has, so it's not too dirty, and the approach is quite consistent here, too.

executor/instruction/relates_executor.rs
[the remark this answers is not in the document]
farost: repaste

farost, line 1: I've made relates always use transitive relation types, because we never hide it in 3.0.

The way to query for non-specialised relates should contain @abstract annotation for relates as it's a sign of specialisation.

However, maybe we need to return only the root relation type (the relation type where this relates is declared) somewhere, as it was previously implemented, and I just couldn't find an example where it worked better than with my fixes.

@dmitrii-ubskii notice: I've put all the existing random failures with `relates` into a Discord thread connected to the message with these prs

concept/tests/test_statistics.rs
```diff
@@ -289,13 +289,15 @@ fn put_has_twice() {
...
+
     let mut snapshot = storage.clone().open_snapshot_write_at(create_commit_seq);
     person.set_has_unordered(&mut snapshot, &thing_manager, name.as_reference()).unwrap();
     thing_manager.finalise(&mut snapshot).unwrap();
-    snapshot.commit().unwrap().unwrap();
+    snapshot.commit().unwrap_err(); // Can't concurrently modify the same 'has'
 
-    let mut synchronised = Statistics::new(SequenceNumber::MIN);
-    synchronised.may_synchronise(&storage).unwrap();
+    synchronised.sequence_number = synchronised.sequence_number + 1;
```
farost, line 300: There was another snapshot trying to commit, so it should be correct.

server/service/transaction_service.rs
```diff
@@ -733,6 +735,14 @@ impl TransactionService {
                 transaction_options,
             );
             self.transaction = Some(Transaction::Schema(transaction));
+
+            let message_ok_done =
```
farost, line 739: We need to return the transaction into `self` before exiting this method. Otherwise, the transaction is not closed, and the schema lock is not released....

database/database.rs
```diff
@@ -87,35 +88,39 @@ impl<D> Database<D> {
         &self.name
     }
 
-    pub(super) fn reserve_write_transaction(&self, timeout_millis: u64) {
-        // TODO: TransactionError
+    pub(super) fn reserve_write_transaction(&self, timeout_millis: u64) -> Result<(), TransactionError> {
+        // TODO: try_lock() instead of lock()? If it's TryLockError::Poisoned (and crash) or TryLockError::WouldBlock (try again later / timeout). Which timeout to use?
```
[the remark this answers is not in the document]
dmitrii-ubskii: I'd expect us to keep track of how much time we spent acquiring the lock and then use the adjusted timeout (minus time already spent) later in `recv`.

database/database.rs
```diff
@@ -87,35 +88,39 @@ impl<D> Database<D> {
...
         let (has_schema_transaction, running_write_transactions, ref mut notify_queue) = *guard;
         if has_schema_transaction || !notify_queue.is_empty() {
             let (sender, receiver) = sync_channel::<()>(0);
             notify_queue.push_back(TransactionReservationRequest::Write(sender));
             drop(guard);
-            receiver.recv().unwrap();
-            // TODO: turn into TransactionError on timeout
+            receiver
+                .recv_timeout(Duration::from_millis(timeout_millis))
+                .map_err(|source| TransactionError::Timeout { source })?;
```
dmitrii-ubskii, line 101: How does the other side of the queue handle the `SendError`s which will pop up if this timeout triggers?

database/transaction.rs
```diff
@@ -314,25 +335,20 @@ impl<D: DurabilityClient> TransactionSchema<D> {
...
-    fn source(&self) -> Option<&(dyn Error + 'static)> {
-        todo!()
+typedb_error!(
+    pub TransactionError(component = "Transaction", prefix = "TXN") {
+        Timeout(1, "Transaction timeout.", source: RecvTimeoutError),
     }
-}
-
-// TODO: TypeDB Error
+);
```
dmitrii-ubskii, line 354: Item-like macros should use `{}` rather than `()`.
`{}`, `[]`, and `()` are interchangeable in macro invocations, but the convention is `{}` for item-like (e.g. `thread_local! {}`), `()` for function-like (`print!()`), `[]` for array-like (`vec![]`).

dependencies/vaticle/repositories.bzl
```diff
@@ -40,8 +41,9 @@ def vaticle_typedb_protocol():
     )
 
 def vaticle_typedb_behaviour():
+    # TODO: Return typedb
     git_repository(
         name = "vaticle_typedb_behaviour",
-        remote = "https://github.com/typedb/typedb-behaviour",
-        commit = "bcc49a3ffd8305f844922a2b36e6f705d9a437de",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
+        remote = "https://github.com/farost/typedb-behaviour",
+        commit = "66baa5dd51240380308a97c3ca4baed9de6d59a6",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @vaticle_typedb_behaviour
```
dmitrii-ubskii, line 48: I believe both of these are approved now.

query/undefine.rs
```diff
@@ -4,22 +4,1149 @@
...
+    fn fmt(&self, _: &mut fmt::Formatter<'_>) -> fmt::Result {
+        todo!()
+    }
+}
+
+impl Error for UndefineError {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        todo!()
+    }
+}
```
dmitrii-ubskii, line 1152: `TypeDBError` doesn't require these, and has its own impl of `Display` based on provided description. These should be removed.

server/service/transaction_service.rs
```diff
@@ -428,14 +430,20 @@ impl TransactionService {
 
         let transaction = match transaction_type {
             typedb_protocol::transaction::Type::Read => {
-                Transaction::Read(TransactionRead::open(database, transaction_options))
+                Transaction::Read(TransactionRead::open(database, transaction_options).map_err(|typedb_source| {
+                    TransactionServiceError::TransactionError { typedb_source }.into_error_message().into_status()
```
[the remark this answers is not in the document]
dmitrii-ubskii: We should not use `Error` in variant names. The fact that it's an error is already communicated by the type, and the variant should clarify which kind of error occurred.

concept/tests/test_thing.rs
```diff
@@ -6,7 +6,7 @@
 
 #![deny(unused_must_use)]
 
-use std::{borrow::Cow, collections::HashMap};
+use std::{borrow::Cow, collections::HashMap, ops::Index};
```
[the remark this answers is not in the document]
farost: Yes, and it hasn't been cleaned up (which is a common issue in our code base, btw..). Thank you!

concept/tests/test_thing.rs
```diff
@@ -521,6 +516,181 @@ fn role_player_duplicates() {
...
+        list_1
+            .add_player(&mut snapshot, &thing_manager, entry_type.clone(), Object::Entity(resource_1.as_reference()))
+            .unwrap();
+        list_1
+            .add_player(&mut snapshot, &thing_manager, entry_type.clone(), Object::Entity(resource_1.as_reference()))
+            .unwrap();
+
+        let player_counts: u64 =
+            list_1.get_players(&snapshot, &thing_manager).map_static(|res| res.unwrap().1).into_iter().sum();
+        assert_eq!(player_counts, 2);
```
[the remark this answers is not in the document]
farost: True, I was also confused! I'll rename it to another collection.

database/database.rs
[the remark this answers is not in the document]
farost: Cool, I wanted to do it, but decided to wait for your thoughts. I'll implement it this way, then, now.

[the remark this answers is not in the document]
farost: Done, please review in the next commit!

query/undefine.rs
farost: Huh, I was sure I had an error when I tried removing it (and I saw that it's already implemented in the macro). Now it works, so it's good. Thanks

dependencies/vaticle/repositories.bzl
farost: They are, but I'm waiting for @flyingsilverfin 's review

database/database.rs
farost: Good point, I didn't think there would be an error in this case. There was a 
`// TODO: if get disconnect error, skip!`

So I implemented it like
`let _skipped_sync_error = notifier.send(()).ok();`
Check the latest commits, please! I might have gotten the idea wrong.

database/transaction.rs
farost: @flyingsilverfin thoughts?

concept/tests/test_statistics.rs
flyingsilverfin: does it even matter?

flyingsilverfin: I think you should move the statistics assert check to right after the new `may_synchronise()`, and then you can do the commit that fails, and then check the statistics has _not_ changed.

concept/tests/test_thing.rs
```diff
@@ -518,11 +513,373 @@ fn role_player_duplicates() {
...
+        collection_1
+            .add_player(&mut snapshot, &thing_manager, entry_type.clone(), Object::Entity(resource_1.as_reference()))
+            .unwrap();
+        collection_1
+            .add_player(&mut snapshot, &thing_manager, entry_type.clone(), Object::Entity(resource_1.as_reference()))
+            .unwrap();
+
+        let player_counts: u64 =
+            collection_1.get_players(&snapshot, &thing_manager).map_static(|res| res.unwrap().1).into_iter().sum();
+        assert_eq!(player_counts, 2);
```
flyingsilverfin, line 544: looks right now yep

concept/tests/test_thing.rs
```diff
@@ -518,11 +513,373 @@ fn role_player_duplicates() {
...
+        let collection_type = type_manager.create_relation_type(&mut snapshot, &collection_label).unwrap();
+        let entry_relates = collection_type
+            .create_relates(&mut snapshot, &type_manager, &thing_manager, entry_role_label, Ordering::Ordered)
+            .unwrap();
+        entry_relates
+            .set_annotation(
+                &mut snapshot,
+                &type_manager,
+                &thing_manager,
+                RelatesAnnotation::Cardinality(AnnotationCardinality::new(0, Some(4))), // must be small to allow index to kick in
```
flyingsilverfin, line 686: Correct!

concept/tests/test_thing.rs
```diff
@@ -518,11 +513,373 @@ fn role_player_duplicates() {
...
+        let group_1_indexed_count: u64 = group_1
+            .get_indexed_players(&snapshot, &thing_manager)
+            .map_static(|res| {
+                let (_, _, _, count) = res.unwrap();
+                count
+            })
+            .into_iter()
+            .sum();
+        assert_eq!(group_1_indexed_count, 0, "// Default card.end for ordered is INF -> indexing is OFF -> expected 0");
+        let resource_1_indexed_count: u64 = resource_1
```
flyingsilverfin, line 828: I'd be happy if this is still all working! :D nice

concept/type_/mod.rs
```diff
@@ -283,6 +283,22 @@ pub trait OwnerAPI<'a>: TypeAPI<'a> {
         attribute_type: AttributeType<'static>,
     ) -> Result<MaybeOwns<'m, HashSet<CapabilityConstraint<Owns<'static>>>>, ConceptReadError>;
 
+    fn get_owned_attribute_types_declared(
```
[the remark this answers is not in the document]
flyingsilverfin: My only thinking is that this has a memory/performance performance profile from the `get_owns_declared`, since it has to allocate a HashSet each time you call it.
1. Do you need this API?
2. Does it guarantee extra Uniqueness (not getting duplicate AttributeTypes) that is useful to you?

If yes to both - i think it might be worth adding a API for this that doesn't rely on the other `get_owns_declared`, and has to be implemented in each implementation -> can be a new or different cache entry to avoid the allocation...

On the other hand, this is the kind of premature optimisation that led us to some technical debt and probably not worth doing! How do we feel about doing a middle ground of prefixing `compute_xxxx` when it's a Types ConceptAPI that does extra allocations and dropping that prefix when it is not? Does that mess up readability too much?

concept/type_/relation_type.rs
```diff
@@ -318,6 +318,22 @@ impl<'a> RelationType<'a> {
         type_manager.get_relation_type_relates_with_specialised(snapshot, self.clone().into_owned())
     }
 
+    pub fn get_related_role_types_declared(
```
flyingsilverfin, line 321: Similar point for all these basically

database/database.rs
```diff
@@ -83,55 +86,94 @@ impl<D> fmt::Debug for Database<D> {
...
             Self::fulfill_reservation_requests(&mut guard)
         }
     }
 
     pub(super) fn release_schema_transaction(&self) {
-        let mut guard = self.schema_write_transaction_exclusivity.lock().unwrap();
+        let mut guard = self
+            .schema_write_transaction_exclusivity
+            .lock()
+            .expect("The exclusive access should already be acquired in `reserve`");
```
flyingsilverfin, line 144: I've also started using `expect` to make debugging these panics so much easier! We should keep using them.

database/database.rs
```diff
@@ -83,55 +86,94 @@ impl<D> fmt::Debug for Database<D> {
...
+        let start_time = Instant::now();
+
+        let guard = loop {
+            match self.schema_write_transaction_exclusivity.try_lock() {
+                Ok(guard) => break guard,
+                Err(TryLockError::WouldBlock) => {
+                    if start_time.elapsed() >= timeout {
+                        return Err(TransactionError::WriteExclusivityTimeout {});
+                    }
+                    std::thread::sleep(Self::TRY_LOCK_SLEEP_TIMEOUT);
```
flyingsilverfin, line 162: I'd call this `TRY_LOCK_SLEEP_INTERVAL`

query/undefine.rs
```diff
@@ -4,22 +4,1137 @@
...
+        TypeCannotHaveCapability(
+            17,
+            "Invalid undefine - the type '{type_}' of kind '{kind}', which cannot have: '{capability}'. Maybe you wanted to undefine something else?",
+            type_: Label<'static>,
+            kind: Kind,
+            capability: CapabilityType
+        ),
+        TypeSubNotDefined(
+            18,
+            "Undefining 'sub {supertype}' for type '{type_}' failed since there is no defined '{type_} sub {supertype}'.\nSource:\n{declaration}",
```
flyingsilverfin, line 1006: We should generally not have constant strings that we know exist elsewhere - for `sub` should be from a TypeQL Token, same with `as`, `relates`... etc.

database/database.rs
farost: Gp!

query/undefine.rs
farost: It kinda makes sense (ignoring the fact that you were the one who introduced these strings first :wink:), although it makes the error message reading even messier, and we won't probably put whole error messages in our bdd checks. I can refactor it as I agree it's cleaner, but it doesn't feel that good.

concept/type_/mod.rs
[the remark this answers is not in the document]
farost: I was hesitant about creating these APIs. The facts that inclined me to doing it:
* We already have similar "helper" apis with clones, and we are forced to do these clones because we have `MaybeOwns`. Even if we don't do it, we have to clone on the API user (in this case `executor`) side
* We actually do clones, collects, and other "greedy" (if applicable here)  operations for these types/capabilities multiple times in executors
* Without these apis, Dmitrii had an impression that `get_owns` returns just all owns a type has, with this type as the owner if it's an inherited owns. I didn't want to change anything globally, it would break many things (and I'm not sure if it's needed), but we agreed that it's not that bad if we at least had an api to answer the question "can I get all the attributes this type owns?". That's the answer
* The overall usage looked cleaner this way


Based on this, I'm not sure if we need to push any concept api interfaces optimization now since we don't have clear numbers, and with the limited resources, we should prioritize. It looks consistent enough in the context of other concept API calls now, although I agree with the drawbacks we have now.

We can always add these caches, it's easy, but a bit annoying work. Let's do it later and globally. I'm not against of creating a list of hypotheses to check our numbers against later!

query/undefine.rs
farost: I have pushed an example of what I can do for `sub`, please check it out!

query/undefine.rs
```diff
@@ -4,22 +4,1142 @@
...
+        ),
+        TypeSubNotDefined(
+            18,
+            "Undefining '{key} {supertype}' for type '{type_}' failed since there is no defined '{type_} {key} {supertype}'.\nSource:\n{declaration}",
+            type_: Label<'static>,
+            key: Keyword,
+            supertype: Label<'static>,
+            declaration: CapabilityType
+        ),
+        TypeSubDefinedButDifferent(
```
farost, line 1015: @flyingsilverfin Is it better now? Check `{key}`

database/transaction.rs
farost: @flyingsilverfin ping

database/database.rs
farost: Done

query/undefine.rs
flyingsilverfin: works for me!

query/undefine.rs
flyingsilverfin: Yeah i know i was rushing, I think there probably is a better strategy overall where we tie the problem back to an original  TypeQL query part. Although that isn't really always possible!

flyingsilverfin (approved):
Good stuff

dependencies/vaticle/repositories.bzl
```diff
@@ -40,8 +40,9 @@ def vaticle_typedb_protocol():
     )
 
 def vaticle_typedb_behaviour():
+    # TODO: Return typedb
```
farost, line 43: TODO: Update before merge

PR #7232, opened 2024-11-01
Bugfix: do not reuse cartesian product iterators between inputs

dmitrii-ubskii:

We can't reuse the iterators created from the previous input row, as they would iterate over different tuples. Now, when we take a new input, we reset the cartesian iterators.


executor/read/immediate_executor.rs
```diff
@@ -242,6 +242,7 @@ impl IntersectionExecutor {
                     return Ok(true);
                 } else {
                     self.iterators.clear();
+                    self.cartesian_iterator.clear();
                     let _ = self.input.as_mut().unwrap().next().unwrap().map_err(|err| err.clone());
```
dmitrii-ubskii, line 246: We can't reuse the iterators created from the previous input row, as they would iterate over different tuples, so we reset the cartesian entirely.

PR #7233, opened 2024-11-01
Optimise comparators

flyingsilverfin:

We optimise the execution of queries containing comparators (`>/</==/...`) where these can be used to restrict the data search space. This PR sets up general infrastructure for using restrictions when searching the data, but uses it only for optimising ISA lookups - follow up work will optimise HAS operations as well.

## Motivation

We can 1) make operations much smaller and 2) make query plans more roboust to a badly chosen plan, by inlining comparators into data lookup operations.

## Implementation

- Utilise Value ranges in executors to restrict the search space of lookups
- Implement approximate casting for Values
- Refactor query planning costings, which now makes a Variable's `cost` take into account the intersections generated at this variable, and update the cost models using a combination of expected counts and selectivity, where selectivity is used to incorporate the value predicates into the output costs and the branching factors.


compiler/executable/match_/planner/plan.rs
```diff
@@ -606,8 +612,9 @@ impl<'a> ConjunctionPlanBuilder<'a> {
         let cost = ordering
             .iter()
             .enumerate()
-            .map(|(i, idx)| self.graph.elements[idx].cost(&ordering[..i], &self.graph))
-            .fold(ElementCost::FREE, |acc, e| acc.chain(e));
+            // TODO: how do we pass the intersection vars back in to recompute the full cost?
+            .map(|(i, idx)| self.graph.elements[idx].cost(&ordering[..i], None, &self.graph))
+            .fold(ElementCost::FREE_BRANCH_1, |acc, e| acc.chain(e));
```
flyingsilverfin, line 617: We now pass the intersection variable into the `cost` function... this becomes a problem here since we want to re-compute the cost, but we no longer have access to the intersection variable!

compiler/executable/match_/instructions/mod.rs
```diff
@@ -64,7 +64,7 @@ impl VariableModes {
         named: &HashSet<ExecutorVariable>,
     ) -> Self {
         let mut modes = Self::new();
-        instruction.ids_foreach(|var| {
+        instruction.used_variables_foreach(|var| {
```
flyingsilverfin, line 67: new API

storage/keyspace/iterator.rs
```diff
@@ -41,67 +35,89 @@ impl KeyspaceRangeIterator {
...
 
-        let keyspace_name = keyspace.name();
-
-        let max_required_length = range
-            .end_value()
-            .map(|bytes| max(bytes.length(), range.start().length()))
-            .unwrap_or(range.start().length());
-
-        // TODO: this Box and copy for comparison shouldn't be necessary
-        let range_iterator = iterator
```
flyingsilverfin, line 64: Optimised to remove the box and copy!

executor/instruction/isa_executor.rs
```diff
@@ -165,72 +180,39 @@ impl IsaExecutor {
...
                     unreachable!("Has thing must be an entity or relation.")
                 };
                 let type_ = thing.type_();
-                let type_manager = thing_manager.type_manager();
-                let types = match self.isa.isa_kind() {
-                    IsaKind::Exact => vec![type_.clone()],
-                    IsaKind::Subtype => get_supertypes(snapshot, type_manager, &type_, SubKind::Subtype)?,
-                };
-                let as_tuples: IsaBoundedSortedType = with_types(lending_iterator::once(Ok(thing)), types)
+                let supertypes = self.instance_type_to_types.get(&type_).cloned().unwrap_or(TYPES_EMPTY.clone());
```
dmitrii-ubskii, line 214: You never need to clone constants, they are copied around by the compiler

executor/instruction/mod.rs
```diff
@@ -67,6 +71,8 @@ mod sub_reverse_executor;
 pub(crate) mod tuple;
 mod type_list_executor;
 
+pub(crate) const TYPES_EMPTY: Vec<Type> = Vec::new();
```
dmitrii-ubskii, line 74: Why is this necessary? Surely you can just use `Vec::new()` inline?

compiler/executable/match_/planner/mod.rs
```diff
@@ -332,6 +339,30 @@ impl MatchExecutableBuilder {
...
+    fn inline_as_optimisation(&mut self, variables: &[Variable], check: &CheckInstruction<ExecutorVariable>) {
+        if !matches!(check, &CheckInstruction::Comparison { .. }) {
+            // TODO: inject IID check as well
+            return;
+        }
+
+        for step in self.steps.iter_mut() {
+            // TODO: we may be able to inject into non-intersection steps as well? For now, we know intersection steps are always sorted
+            if let StepInstructionsBuilder::Intersection(intersection) = &mut step.builder {
+                for instruction in intersection.instructions.iter_mut().rev() {
```
dmitrii-ubskii, line 352: Why `.rev()`?

compiler/executable/match_/planner/mod.rs
```diff
@@ -307,16 +307,23 @@ impl MatchExecutableBuilder {
...
-                    return;
+                    instruction.add_check(check.clone());
+                    is_added = true;
                 }
             }
+            if is_added {
+                self.current.as_mut().unwrap().selected_variables = self.current_outputs.clone();
+                return;
+            }
         }
```
dmitrii-ubskii, line 327: Can this be unified with the body of `inline_as_optimisation()`? The code isn't quite the same, but it presumably does roughly the same thing?

compiler/executable/match_/planner/mod.rs
```diff
@@ -307,16 +307,23 @@ impl MatchExecutableBuilder {
     }
 
     fn push_check(&mut self, variables: &[Variable], check: CheckInstruction<ExecutorVariable>) {
+        // if it is a comparison or IID (TODO) we can inline the check into previous instructions
+        self.inline_as_optimisation(variables, &check);
```
dmitrii-ubskii, line 311: If this succeeds, we should probably short-circuit out of here.

compiler/executable/match_/planner/plan.rs
dmitrii-ubskii: It should be inferrable from the ordering itself (viz. the next variable in the ordering should be the intersection).

compiler/executable/match_/planner/vertex/constraint.rs
```diff
@@ -295,17 +345,13 @@ impl<'a> HasPlanner<'a> {
...
         Self {
             has,
             owner: variable_index[&owner.as_variable().unwrap()],
             attribute: variable_index[&attribute.as_variable().unwrap()],
-            expected_size,
-            expected_unbound_size,
-            unbound_direction,
+            unbound_typed_expected_size: expected_size,
+            unbound_typed_expected_size_canonical: unbound_forward_size,
+            unbound_typed_expected_size_reverse: unbound_backward_size,
```
dmitrii-ubskii, line 354: The locals should've been renamed too.

compiler/executable/match_/planner/vertex/variable.rs
```diff
@@ -186,25 +199,31 @@ impl TypePlanner {
...
 #[derive(Clone)]
 pub(crate) struct ThingPlanner {
     variable: Variable,
     binding: Option<PatternVertexId>,
-    expected_size: f64,
+    unrestricted_expected_size: f64,
+    unrestricted_expected_attribute_types: usize,
 
-    bound_exact: HashSet<VariableVertexId>, // IID or exact Type + Value
+    restriction_exact: HashSet<VariableVertexId>, // IID or exact Type + Value
```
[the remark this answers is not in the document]
dmitrii-ubskii: Can we then rename `restriction_value_below` to `_from_below`?

compiler/executable/match_/planner/vertex/constraint.rs
```diff
@@ -240,11 +247,52 @@ impl<'a> IsaPlanner<'a> {
...
+
+        let (per_input, per_output) = match (is_thing_bound, is_type_bound) {
+            (true, true) | (true, false) => (0.0, 0.0),
+            (false, true) | (false, false) => {
+                let per_input = OPEN_ITERATOR_RELATIVE_COST;
+                // when the type is bound or unbound, we may reject multiple things until we find a matching one, depending on the selectivity of the thing
+                let per_output = ADVANCE_ITERATOR_RELATIVE_COST / thing.selectivity();
+                (per_input, per_output)
+            }
+        };
```
dmitrii-ubskii, line 276: `is_type_bound` is unused here, this should just be `if is_thing_bound`

compiler/executable/match_/planner/vertex/variable.rs
```diff
@@ -271,82 +308,59 @@ impl ThingPlanner {
...
-}
 
-impl Costed for ThingPlanner {
-    fn cost(&self, inputs: &[VertexId], graph: &Graph<'_>) -> ElementCost {
-        let bounds = chain!(&self.bound_value_equal, &self.bound_value_above, &self.bound_value_below).collect_vec();
-        for &i in inputs {
-            let VertexId::Variable(i) = i else { continue };
-            if !bounds.contains(&&Input::Variable(i)) {
-                return ElementCost::FREE;
+    fn selectivity(&self) -> f64 {
```
dmitrii-ubskii, line 312: Why does this not take the ordered prefix? Here, if you query `match $x has id $x-id; $y has id $y-id; $x-id == $y-id;`, both `$x-id` and `$y-id` will be highly selective, but only the later one in execution can actually make use of that.

compiler/executable/match_/planner/vertex/mod.rs
```diff
@@ -432,3 +437,12 @@ impl Costed for DisjunctionPlanner<'_> {
             .fold(ElementCost::EMPTY, ElementCost::combine_parallel)
     }
 }
+
+pub(super) fn type_count(type_: &Type, statistics: &Statistics) -> u64 {
```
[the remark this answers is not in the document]
dmitrii-ubskii: `instance_count()`

compiler/executable/match_/planner/vertex/mod.rs
```diff
@@ -432,3 +437,12 @@ impl Costed for DisjunctionPlanner<'_> {
...
+    match type_ {
+        Type::Entity(entity) => *statistics.entity_counts.get(entity).unwrap_or(&0),
+        Type::Relation(relation) => *statistics.relation_counts.get(relation).unwrap_or(&0),
+        Type::Attribute(attribute) => *statistics.attribute_counts.get(attribute).unwrap_or(&0),
+        Type::RoleType(_) => 0,
```
dmitrii-ubskii, line 446: Is this even reachable? Why?

executor/instruction/has_reverse_executor.rs
```diff
@@ -116,6 +104,25 @@ impl HasReverseExecutor {
         context: &ExecutionContext<impl ReadableSnapshot + 'static>,
         row: MaybeOwnedRow<'_>,
     ) -> Result<TupleIterator, ConceptReadError> {
+        if self.iterate_mode.is_inverted() && self.attribute_cache.get().is_none() {
+            // TODO: this will only work if the range is actually not dependent on the Row... needs special API
+
+            // one-off initialisation of the cache of constants
+            let value_range =
+                self.checker.value_range_for(context, None, self.has.attribute().as_variable().unwrap())?;
```
dmitrii-ubskii, line 112: This is missing from `HasExecutor`'s `BoundFrom` case

executor/instruction/mod.rs
```diff
@@ -358,29 +365,96 @@ impl<T: Hkt> Checker<T> {
...
-                    range = intersect(range, comp_range);
+        for i in 0..self.checks.len() {
+            let check = &self.checks[i];
+            match check {
+                CheckInstruction::Comparison { lhs, rhs, comparator } => {
+                    if lhs.as_variable().is_some_and(|var| var == target_variable) {
+                        let rhs_variable_value = Self::get_vertex_value(rhs, row.as_ref(), &context.parameters);
+                        let rhs_value =
+                            Self::read_value(context.snapshot.as_ref(), &context.thing_manager, &rhs_variable_value)?;
+                        // let rhs_value = Self::read_value(context.snapshot.as_ref(), &context.thing_manager, &rhs_variable_value)?;
```
dmitrii-ubskii, line 377: delet

executor/instruction/mod.rs
```diff
@@ -358,29 +365,96 @@ impl<T: Hkt> Checker<T> {
...
+                                Comparator::GreaterOrEqual => (Bound::Unbounded, Bound::Included(lhs_value)),
+                                Comparator::Like => continue,
+                                Comparator::Contains => continue,
+                                Comparator::NotEqual => continue,
+                            };
+                            range = intersect(range, comp_range);
+                        }
+                    }
                 }
                 _ => (),
```
dmitrii-ubskii, line 413: Can we also add the `Is` check as if it were equality here? It's probably less common for attributes, but still.

executor/instruction/mod.rs
```diff
@@ -358,29 +365,96 @@ impl<T: Hkt> Checker<T> {
...
-                        Comparator::GreaterOrEqual => (Bound::Included(rhs), Bound::Unbounded),
-                        Comparator::Like => continue,
-                        Comparator::Contains => continue,
-                    };
-                    range = intersect(range, comp_range);
+        for i in 0..self.checks.len() {
+            let check = &self.checks[i];
+            match check {
+                CheckInstruction::Comparison { lhs, rhs, comparator } => {
+                    if lhs.as_variable().is_some_and(|var| var == target_variable) {
```
dmitrii-ubskii, line 373: This is a little cleaner imo:
```suggestion
                    if lhs.as_variable() == Some(target_variable) {
```

executor/instruction/mod.rs
```diff
@@ -358,29 +365,96 @@ impl<T: Hkt> Checker<T> {
...
+                ExecutorVariable::Internal(_) => {
+                    unreachable!("Comparator check variables must have been recorded in the row.")
+                }
+            },
+            CheckVertex::Type(type_) => VariableValue::Type(type_.clone()),
+            CheckVertex::Parameter(parameter_id) => {
+                VariableValue::Value(parameters.value_unchecked(*parameter_id).as_reference())
+            }
+        }
+    }
```
dmitrii-ubskii, line 439: Partially duplicated with `make_const_extractor()`

executor/instruction/mod.rs
```diff
@@ -358,29 +365,96 @@ impl<T: Hkt> Checker<T> {
...
+
+    fn read_value<'b, 'a>(
+        snapshot: &'a impl ReadableSnapshot,
+        thing_manager: &'a ThingManager,
+        variable_value: &'a VariableValue<'a>,
+    ) -> Result<Option<Value<'static>>, ConceptReadError> {
+        // TODO: is there a way to do this without cloning the value?
```
dmitrii-ubskii, line 446: Is there a reason it needs to be `<'static>` and not `<'a>`? Otherwise no.

executor/instruction/mod.rs
```diff
@@ -358,29 +365,96 @@ impl<T: Hkt> Checker<T> {
...
+            CheckVertex::Parameter(parameter_id) => {
+                VariableValue::Value(parameters.value_unchecked(*parameter_id).as_reference())
+            }
+        }
+    }
+
+    fn read_value<'b, 'a>(
+        snapshot: &'a impl ReadableSnapshot,
+        thing_manager: &'a ThingManager,
+        variable_value: &'a VariableValue<'a>,
```
dmitrii-ubskii, line 444: `'b` is unused, `VariableValue<'a>` is redundant and potentially overly restrictive. `'a` on snapshot may be redundant as well, but I can't tell without static analysis.
```suggestion
    fn read_value<'a>(
        snapshot: &'a impl ReadableSnapshot,
        thing_manager: &'a ThingManager,
        variable_value: &'a VariableValue<'_>,
```

dmitrii-ubskii (changes requested):
Reviewed `planner` and `Checker`

compiler/executable/match_/planner/mod.rs
flyingsilverfin: I think you're right - had this before and split it out again and I don't remember why... will do that

compiler/executable/match_/planner/vertex/variable.rs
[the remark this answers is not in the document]
flyingsilverfin: sure

compiler/executable/match_/planner/vertex/variable.rs
flyingsilverfin: Good point

executor/instruction/has_reverse_executor.rs
flyingsilverfin: Yeah i'm aware, thanks

executor/instruction/mod.rs
flyingsilverfin: Easier to read IMO

executor/instruction/mod.rs
flyingsilverfin: Done

executor/instruction/mod.rs
flyingsilverfin: It messees with the caller somehow - maybe you could look into it after it merges sometime

executor/instruction/mod.rs
flyingsilverfin: nice catch!

compiler/executable/match_/planner/vertex/variable.rs
flyingsilverfin: Done!

PR #7235, opened 2024-11-04
Introduce 3.0 `value` and `as` match constraints

farost:
We introduce `value` match constraint for querying for attribute types of specific value types:
```
match
  $lo value long;
```

We introduce `as` match constraint for querying for role types specialisation:
```
match
  $relation relates $role as parentship:child;
```
which is equivalent to:
```
match
  $relation relates $role;
  $role sub parentship:child;
```
but is useful for shortening your queries, making it similar to `define` definitions, and in case of anonymous variables:
```
match
  $relation relates $_ as parentship:child;
```

## Implementation
`as` uses `sub` constraint and is expected to have an `as!` alternative to `sub!` in the future.


compiler/executable/match_/planner/plan.rs
```diff
@@ -943,7 +949,9 @@ impl ConjunctionPlan<'_> {
                         Inputs::Single([rhs_var.unwrap()]),
                         self.type_annotations,
                     ))
-                } else if constraint.unbound_direction() == Direction::Canonical && Some(sort_variable) != rhs_var {
+                } else if constraint.unbound_direction() == Direction::Canonical && Some(sort_variable) != rhs_var
+                    || Some(sort_variable) == lhs_var
```
farost, line 953: When testing `as` as `sub`, I started encountering `UnboundInverted` iterate mods, which were not expected.

executor/instruction/iterator.rs
```diff
@@ -50,7 +50,7 @@ use crate::{
         relates_executor::{RelatesBoundedSortedRole, RelatesUnboundedSortedRelation},
         relates_reverse_executor::{RelatesReverseBoundedSortedRelation, RelatesReverseUnboundedSortedRole},
         sub_executor::{SubBoundedSortedSuper, SubUnboundedSortedSub},
-        sub_reverse_executor::{SubReverseBoundedSortedSuper, SubReverseUnboundedSortedSub},
+        sub_reverse_executor::{SubReverseBoundedSortedSub, SubReverseUnboundedSortedSuper},
```
farost, line 53: Looked like a typo based on the actual usage

farost:
The BDDs for these features are updated here (I didn't want to resolve stupid conflicts in bdd declarations, and my work directions are quite coupled now): https://github.com/typedb/typedb-behaviour/pull/309

compiler/annotation/type_seeder.rs
```diff
@@ -672,6 +675,47 @@ impl UnaryConstraint for RoleName<Variable> {
...
+                let attribute_value_type_opt = attribute_type
+                    .get_value_type_without_source(seeder.snapshot, seeder.type_manager)
+                    .map_err(|source| TypeInferenceError::ConceptRead { source })?;
+                if let Some(attribute_value_type) = attribute_value_type_opt {
+                    if pattern_value_type == attribute_value_type {
+                        annotations.insert(TypeAnnotation::Attribute(attribute_type));
+                    }
+                }
+                Ok(())
+            })?;
```
dmitrii-ubskii, line 712: This can be written a little simpler with less nesting.
```suggestion
        let attribute_types = seeder
            .type_manager
            .get_attribute_types(seeder.snapshot)
            .map_err(|source| TypeInferenceError::ConceptRead { source })?;
        for attribute_type in attribute_types {
            let attribute_value_type = attribute_type
                .get_value_type_without_source(seeder.snapshot, seeder.type_manager)
                .map_err(|source| TypeInferenceError::ConceptRead { source })?;
            if attribute_value_type == Some(pattern_value_type) {
                annotations.insert(TypeAnnotation::Attribute(attribute_type));
            }
        }
```

executor/instruction/iterator.rs
dmitrii-ubskii: I think I didn't update a lot of these when I adjusted tuple ordering for these iterators. Note to self to fix in my next PR (out of scope for this one).
Good catch!

dmitrii-ubskii (approved):
LGTM!

compiler/annotation/type_seeder.rs
farost: Oh wow, I didn't think about using `attribute_value_type == Some(pattern_value_type)`!! That's fire, thanks.

compiler/executable/match_/planner/plan.rs
flyingsilverfin: That sounds like a planner issue more than anything!

flyingsilverfin: So this is (direction == CANONICAL && soft_var != rhs_var) || Some(soft_var) == lhs_var ?

farost: Yeah, the code is correct. I can highlight it if you want to

flyingsilverfin (approved):

PR #7237, opened 2024-11-05
Abstract types should never be inferred for instance variables

cxdorn:

This PR makes makes the type checker more precise (and complain less), by excluding abstract types from being inferred for instance variables.

## Motivation

Abstract types should never be inferred for instance (i.e. non-type) variables.

## Implementation

At the end of generating the type seed, we filter out all abstract types of instance variables.


compiler/annotation/type_seeder.rs
```diff
@@ -140,6 +140,35 @@ impl<'this, Snapshot: ReadableSnapshot> TypeGraphSeedingContext<'this, Snapshot>
...
+                           Type::RoleType(t) => t.is_abstract(self.snapshot, self.type_manager),
+                           _ => Ok(false)
+                       }.map_err(|source| { TypeInferenceError::ConceptRead { source }})?;
+                       if annotated_type_is_abstract {
+                           to_be_removed.push(annotated_type.clone());
+                       }
+                   }
+                   for annotated_type in to_be_removed.iter() {
+                       annotations.remove(annotated_type);
+                   }
```
flyingsilverfin, line 167: I think there's something like `annotations.retain(|annotated_type| <return false for abstract types> )` that is simpler

cxdorn: Yes, @krishnangovindraj pointed that out too. It requires the closure to be a `bool` returning function, whereas best we can do here is `Result<bool,_>` (since `is_abstract()` returns a result)

(unless we unwrap?)

flyingsilverfin: Got itttt I'd probably not structure it as "removals" but "keeping" (find vec of non-abstract types, and set that annotations to that somehow if that is possible, but i suppose this is ok too!

flyingsilverfin: You could also push `is_abstract` onto `Type` if we think it is generically useful ( I think it is )

compiler/annotation/type_seeder.rs
```diff
@@ -140,6 +140,28 @@ impl<'this, Snapshot: ReadableSnapshot> TypeGraphSeedingContext<'this, Snapshot>
...
+                   let annotated_type_is_abstract = annotated_type
+                       .is_abstract(self.snapshot, self.type_manager)
+                       .map_err(|source| { TypeInferenceError::ConceptRead { source }})?;
+                   if annotated_type_is_abstract {
+                       to_be_removed.push(annotated_type.clone());
+                   }
+               }
+               for annotated_type in to_be_removed.iter() {
+                   annotations.remove(annotated_type);
+               }
```
krishnangovindraj, line 161: What we've done here is essentially a `try_retain` and I think all the 3 cases would be more readable if we defined that as a function. I think it's worth having atleast locally. It doesn't have to be generic:

```rust
fn try_retain<T: std::cmp::Ord + Clone, Err>(set: &mut BTreeSet<T>, fallible_check: impl Fn(&T) -> Result<bool, Err>) -> Result<(), Err>{
    let to_remove: Vec<T> = set.iter().filter_map(|item| {
        match fallible_check(item) {
            Err(err) => Some(Err(err)),
            Ok(false) => Some(Ok(item.clone())),
            Ok(true) => None,
        }
    }).collect::<Result<Vec<_>, _>>()?;
    to_remove.iter().for_each(|item| {set.remove(item);});
    Ok(())
}
```

Then, this would simply read:
```rust
 annotations.try_retain(|type_| {
  type_.is_abstract()).map_err(|source| { TypeInferenceError::ConceptRead { source }})?
});
```

compiler/annotation/type_seeder.rs
```diff
@@ -703,8 +739,19 @@ trait BinaryConstraint {
...
+                for right_type in right_annotations.iter() {
+                    if right_is_thing && right_type.is_abstract(seeder.snapshot, seeder.type_manager)? {
+                        continue;
+                    }
+                    left_to_right
+                        .entry(left_type.clone())
+                        .or_insert_with(BTreeSet::new)
+                        .insert(right_type.clone());
+                }
+            }
```
krishnangovindraj, line 755: If we have the `try_retain` proposed above.
```rust
self.annotate_left_to_right_for_type(seeder, left_type, &mut right_annotations)?;
if right_is_thing { 
   try_retain(right_annotations, |type_| type_.is_abstract(...))?;
}
left_to_right.insert(left_type.clone(), right_annotations);
```

compiler/annotation/type_seeder.rs
```diff
@@ -717,7 +764,19 @@ trait BinaryConstraint {
...
+                        continue;
+                    }
+                    right_to_left
+                        .entry(right_type.clone())
+                        .or_insert_with(BTreeSet::new)
+                        .insert(left_type.clone());
+                }
+            }
             right_to_left.insert(right_type.clone(), left_annotations);
         }
```
krishnangovindraj, line 782: Same comments as  left_to_right

krishnangovindraj (changes requested):

compiler/annotation/type_seeder.rs
```diff
@@ -703,8 +739,19 @@ trait BinaryConstraint {
         let mut left_to_right = BTreeMap::new();
         for left_type in left_types {
             let mut right_annotations = BTreeSet::new();
+            let (left_is_thing, right_is_thing) = self.check_for_thing_vars(seeder);
             self.annotate_left_to_right_for_type(seeder, left_type, &mut right_annotations)?;
-            left_to_right.insert(left_type.clone(), right_annotations);
+            if !(left_is_thing && left_type.is_abstract(seeder.snapshot, seeder.type_manager)?) {
```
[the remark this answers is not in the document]
krishnangovindraj: Actually this doesn't work because this function is also called in the vertex annotation stage.

flyingsilverfin (approved):

PR #7238, opened 2024-11-05
Fix outputs from earlier disjunction branches affecting later branches

krishnangovindraj:
Fixes outputs from earlier disjunction branches affecting later branches

## Implementation
Disjunctions can't just re-use `assigned_positions` as input variables. `MatchExecutableBuilder` now takes an explicit input_variables argument to support this.

### DIFF
--- a/compiler/executable/match_/planner/plan.rs
+++ b/compiler/executable/match_/planner/plan.rs
@@ -836,7 +836,7 @@ impl ConjunctionPlan<'_> {
                 let negation = negation.plan().lower(
                     match_builder.current_outputs.iter().copied(),
                     match_builder.position_mapping(),
-                    &match_builder.current_outputs,
+                    &match_builder.position_mapping().keys().copied().collect(),
                     variable_registry,
                 );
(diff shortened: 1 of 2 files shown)


compiler/executable/match_/planner/plan.rs
```diff
@@ -832,6 +836,7 @@ impl ConjunctionPlan<'_> {
                 let negation = negation.plan().lower(
                     match_builder.current_outputs.iter().copied(),
                     match_builder.position_mapping(),
+                    &match_builder.position_mapping().keys().copied().collect(),
```
krishnangovindraj, line 839: @dmitrii-ubskii : This is equivalent to what it was before the change.
 Is this ok, or should it be: 
```rust 
&negation.variables().map(|id| self.graph.index_to_variable.get(&id).unwrap().clone()).collect_vec()```

dmitrii-ubskii: I think that ends up the same semantically. I'd expect `.keys().collect()` to be a little faster though.

PR #7239, opened 2024-11-06
Ensure sorted join executor respects selected outputs

dmitrii-ubskii:

Before this change, an intersection executor (which is used for any number of sorted iterators, not necessarily a join) would write into the output row all items in the constituent executors, so long as they have a row position assigned. That would happen even if those values were not selected for future steps, which prevented deduplication of answers from being handled correctly.


farost (approved):
Lgtm!

flyingsilverfin (approved):

PR #7240, opened 2024-11-07
Commit generated Cargo.tomls + Cargo.lock

dmitrii-ubskii:

We commit the generated cargo manifests so that TypeDB can be built as a normal cargo project without using Bazel.

Integration and unit tests can be run normally, and rust-analyzer behaves correctly. Behaviour tests currently cannot be run without Bazel, as the tests expect to find the feature files in `bazel-typedb/external/typedb_behaviour`.

In addition, if during development the developer needs to depend on a local clone of a dependency (e.g. making parallel changes to typeql), the Cargo.toml would need to be temporarily manually adjusted to point to the path dependency.

## Implementation

We also propagate the rename of vaticle bazel org prefix to typedb.


.gitignore
```diff
@@ -57,7 +57,3 @@ logs/
 
 # Python cache
 __pycache__
-
-# Cargo files
-Cargo.lock
-Cargo.toml
```
flyingsilverfin, line 63: Why do we commit cargo locks here but not in Protocol or TypeQL?

dependencies/typedb/repositories.bzl.orig
```diff
@@ -0,0 +1,52 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
flyingsilverfin, line 1: Leftover files from merge!

flyingsilverfin (approved):

.gitignore
dmitrii-ubskii: TypeDB is a binary, and it's accepted practice in Rust community to provide lockfiles. Libraries cannot rely on lockfiles though.

PR #7241, opened 2024-11-07
Introduce naive retries for suspended function calls

krishnangovindraj:
Introduces retries for suspended function calls. 
Functions may suspend to break cycles of recursion. Restoring suspend points is essential to completeness of recursive functions, and thus correctness of negations which call them.

## Motivation
Towards complete fixpoint computation.

## Implementation
* Introduces a SuspendPointContext to abstract away some of the logic.
* Introduces retries at the entry point and at each negation.

### DIFF
--- a/executor/match_executor.rs
+++ b/executor/match_executor.rs
@@ -18,7 +18,7 @@ use crate::{
     error::ReadExecutionError,
     pipeline::stage::ExecutionContext,
     read::{
-        pattern_executor::PatternExecutor, tabled_functions::TabledFunctions, SuspendPoint,
+        pattern_executor::PatternExecutor, tabled_functions::TabledFunctions, QueryPatternSuspensions,
         TODO_REMOVE_create_executors_for_match,
     },
     row::MaybeOwnedRow,
@@ -29,7 +29,8 @@ pub struct MatchExecutor {
     entry: PatternExecutor,
     input: Option<MaybeOwnedRow<'static>>,
     tabled_functions: TabledFunctions,
-    suspend_points: Vec<SuspendPoint>,
+    suspensions: QueryPatternSuspensions,
+    last_seen_table_size: usize,
 }
 
 impl MatchExecutor {
@@ -49,7 +50,8 @@ impl MatchExecutor {
             )?,
             tabled_functions: TabledFunctions::new(function_registry),
             input: Some(input.into_owned()),
-            suspend_points: Vec::new(),
+            suspensions: QueryPatternSuspensions::new(),
+            last_seen_table_size: 0,
         })
     }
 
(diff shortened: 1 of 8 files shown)


executor/read/mod.rs
```diff
@@ -62,35 +62,106 @@ pub(super) fn create_executors_for_pipeline(
...
+
+#[derive(Debug)]
+pub(super) struct SuspendPointContext {
+    at_depth: usize,
+    suspended_points: Vec<SuspendPoint>,
+    restore_from: Vec<SuspendPoint>, //Peekable<std::vec::IntoIter<SuspendPoint>>,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+struct SuspendPointTrackerState(usize);
```
flyingsilverfin, line 104: So this is basically the number of suspend points? Can we just call it `SuspendPointCount`? Then when we `record_nested_pattern_entry` we get back a `SuspendPointCount` which I think makes sense?

executor/read/pattern_executor.rs
```diff
@@ -139,11 +232,11 @@ impl PatternExecutor {
                         unreachable!();
                     };
                     let branch = &mut disjunction.branches[branch_index.0];
-                    let suspend_point_len_before = suspend_point_accumulator.len();
+                    let suspend_tracker_state_before = suspend_point_context.record_nested_pattern_entry();
```
flyingsilverfin, line 235: If this type becomes `SuspendPointCount` then this is `suspend_point_count_before`, which then kinda makes sense to compare it to the count of suspend points after and see if somethiing needs to be done?

executor/read/mod.rs
```diff
@@ -62,35 +62,106 @@ pub(super) fn create_executors_for_pipeline(
...
 #[derive(Debug)]
 pub(super) struct NestedSuspension {
     pub(crate) executor_index: ExecutorIndex,
+    pub(crate) depth: usize,
     pub(crate) branch_index: BranchIndex,
     pub(crate) input_row: MaybeOwnedRow<'static>,
 }
+
+#[derive(Debug)]
+pub(super) struct SuspendPointContext {
```
flyingsilverfin, line 97: I feel like `Context` is a little too generic for something that does something extremely specific - in my local branch I played with `SuspendPointCollector` since it records and pops suspend points? `Context` is hard to relate to the other data structures and algorithms at play.

executor/read/pattern_executor.rs
```diff
@@ -382,26 +478,32 @@ impl PatternExecutor {
...
+                let suspension_state_after = function_suspend_points.record_nested_pattern_exit();
                 if let Some(batch) = batch_opt {
                     let deduplicated_batch = executor.add_batch_to_table(&function_state, batch);
                     Some(deduplicated_batch)
                 } else {
-                    if !suspend_points.is_empty() {
-                        suspend_point_accumulator.push(executor.create_suspend_point_for(index))
+                    if suspension_state_after.has_suspend_points() {
+                        // TODO: Consider retrying here. For now, just record a suspend point for ourselves.
+                        suspend_point_context.push_tabled_call(index, executor);
```
flyingsilverfin, line 506: So if i'm getting this right - every function call & the entry point has a set of suspend points (loosely, which parts of the query that is being executed were signalled to pause. If a function hits suspend points, we also record a suspend point in the parent caller.

executor/read/tabled_call_executor.rs
```diff
@@ -82,9 +89,10 @@ impl TabledCallExecutor {
         output_batch
     }
 
-    pub(crate) fn create_suspend_point_for(&self, executor_index: ExecutorIndex) -> SuspendPoint {
+    pub(crate) fn create_suspend_point_for(&self, executor_index: ExecutorIndex, depth: usize) -> SuspendPoint {
         SuspendPoint::TabledCall(TabledCallSuspension {
             executor_index,
+            depth,
```
flyingsilverfin, line 95: Is the `try_read_next_batch` below where we do recursion/cycle detection? By checking the write lock on the executor state? Is is ever possible to hold the lock in non-cyclic cases -> trigger suspend points and recomputation?

executor/match_executor.rs
```diff
@@ -70,12 +72,35 @@ impl MatchExecutor {
...
+            // TODO: We do it all the restoring here, but we should probably be doing it elsewhere. Maybe pattern_executor::execute_tabled_call
+            //  when the function returns None, AND it's possibly the head of a cycle.
+            while return_batch.is_none() && !self.suspend_points.is_empty() {
+                self.last_seen_table_size = self.tabled_functions.total_table_size();
+                for function_state in self.tabled_functions.iterate_states() {
+                    let mut guard = function_state.executor_state.try_lock().unwrap();
+                    guard.prepare_to_retry_suspend_points();
+                }
+                // And on the entry
+                self.suspend_points.swap_suspend_and_restore_points();
```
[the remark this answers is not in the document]
krishnangovindraj: `prepare_restore_from_suspend_points`

executor/read/mod.rs
```diff
@@ -62,35 +62,106 @@ pub(super) fn create_executors_for_pipeline(
...
+    at_depth: usize,
+    suspended_points: Vec<SuspendPoint>,
+    restore_from: Vec<SuspendPoint>, //Peekable<std::vec::IntoIter<SuspendPoint>>,
```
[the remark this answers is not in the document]
krishnangovindraj: I need two collections for sure. I'm restoring from one while pushing to the other.

executor/read/pattern_executor.rs
krishnangovindraj: Yeah I was wondering whether to add the depth as well then decided not to.

executor/read/mod.rs
krishnangovindraj: Used to be `SuspendPointTracker`

executor/read/pattern_executor.rs
krishnangovindraj: Yes. Because we need to restore the whole stack till that point.

executor/read/tabled_call_executor.rs
krishnangovindraj: It shouldn't be possible in a single threaded case.

executor/read/tabled_functions.rs
```diff
@@ -89,7 +108,7 @@ impl TabledFunctionState {
             table: RwLock::new(AnswerTable { answers: Vec::new(), width: answer_width }),
             executor_state: Mutex::new(TabledFunctionPatternExecutorState {
                 pattern_executor,
-                suspend_points: Vec::new(),
+                suspend_points: SuspendPointContext::new(),
                 parameters,
             }),
         }
```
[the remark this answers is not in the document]
krishnangovindraj: Ah I forget we can have empty rows. We could do this and standardise.

flyingsilverfin (approved):

PR #7242, opened 2024-11-07
Optimise has comparators and implement Executable Display

flyingsilverfin:

We inline comparators into `HasReverse` executors, as a continuation of #7233, which inlined comparators into `Isa` executors. Both optimisations reduce the search space by using the value ranges provided by constants in the query or previously executed query steps. We also improve the query planner's statistics and cost function.

## Motivation

Traversals can do extra work scanning ranges of data when 1) we know a range to search in eg, `$age > 18;` and 2) the data is sorted in an order that we can take advantage of.

## Implementation

- Push value ranges into `HasReverse` execution, which causes the iterators to skip data ranges that would never match the query
- Implement `Display` for `MatchExecutable`s, which allow much faster debugging of slow or incorrect queries, which may always be caused by an unusual query plan!
- Optimise the query planner statistics calculation, including making in-memory costs non-zero.

### DIFF
--- a/compiler/executable/match_/instructions/thing.rs
+++ b/compiler/executable/match_/instructions/thing.rs
@@ -6,11 +6,11 @@
 
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap},
+    fmt::{Display, Formatter},
     sync::Arc,
 };
-use std::fmt::{Display, Formatter};
 
-use answer::{Type, variable::Variable};
+use answer::{variable::Variable, Type};
 use ir::pattern::{
     constraint::{Has, Isa, Links},
     IrID,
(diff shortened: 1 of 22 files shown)


compiler/executable/match_/instructions/thing.rs
```diff
@@ -19,36 +20,7 @@ use crate::{
     annotation::type_annotations::TypeAnnotations,
     executable::match_::instructions::{CheckInstruction, Inputs},
 };
-
-#[derive(Debug, Clone)]
-pub struct ConstantInstruction<ID> {
```
flyingsilverfin, line 24: Unused

concept/thing/statistics.rs
```diff
@@ -143,8 +143,6 @@ impl Statistics {
                         self.update_writes(&commits, storage).map_err(|err| DataRead { source: err })?;
                     }
                 }
-            } else {
```
flyingsilverfin, line 146: This seemed like a bug that triggered sometimes, so I've removed it

PR #7245, opened 2024-11-10
Boxed large result Error variants

flyingsilverfin:

We follow some best practices to box the large Error variants of Result types, which should optimise both the stack size and the amount of memory that has to be copied between function calls. 

## Motivation

This refactor was prompted by spotting a non-trivial amount of time doing `mem-copies` in flamegraphs. Although this is only 1 place we do significant memory copies (others might be `clones` or collections that are being reallocated), we expect that this optimisation has some impact (and in fact, we measured maybe 1-2% improvement in TPCC). 

## Implementation

- Where clippy warned us of a large `Err` variant of Results that are being returned, we Box the Error type. This filters through to our error `source` fields in error enums, function signatures, and construction of errors. However, nothing is particularly interesting or fundamental.


concept/error.rs
```diff
@@ -49,7 +49,7 @@ typedb_error!(
     pub ConceptWriteError(component = "Concept write", prefix = "COW") {
         SnapshotGet(1, "Concept write failed due to a snapshot read error.", (source: SnapshotGetError)),
         SnapshotIterate(2, "Concept write failed due to a snapshot iteration error.", (source: Arc<SnapshotIteratorError>)),
-        ConceptRead(3, "Concept write failed due to a concept read error.", (source: ConceptReadError)),
+        ConceptRead(3, "Concept write failed due to a concept read error.", (source: Box<ConceptReadError>)),
```
dmitrii-ubskii, line 52: This can still be inlined, you just need to move out of the `Box` (`*` deref) in the `from()`

concept/error.rs
```diff
@@ -64,34 +64,34 @@ typedb_error!(
...
-        }
+            ConceptReadError::CorruptFoundHasWithoutOwns => ConceptWriteError::ConceptRead { source: error },
+            ConceptReadError::CorruptFoundLinksWithoutPlays => ConceptWriteError::ConceptRead { source: error },
+            ConceptReadError::CorruptFoundLinksWithoutRelates => ConceptWriteError::ConceptRead { source: error },
+            ConceptReadError::CannotGetOwnsDoesntExist(_, _) => ConceptWriteError::ConceptRead { source: error },
+            ConceptReadError::CannotGetPlaysDoesntExist(_, _) => ConceptWriteError::ConceptRead { source: error },
+            ConceptReadError::CannotGetRelatesDoesntExist(_, _) => ConceptWriteError::ConceptRead { source: error },
+            ConceptReadError::Annotation { .. } => ConceptWriteError::ConceptRead { source: error },
+            ConceptReadError::Constraint { .. } => ConceptWriteError::ConceptRead { source: error },
+            ConceptReadError::ValueTypeMismatchWithAttributeType { .. } => ConceptWriteError::ConceptRead { source: error },
```
dmitrii-ubskii, line 93:
```suggestion
            other => ConceptWriteError::ConceptRead { source: other },
```

tests/behaviour/steps/params.rs
```diff
@@ -65,11 +65,11 @@ impl MayError {
...
-    ) -> Option<ConceptWriteError> {
+        res: &Result<T, Box<ConceptWriteError>>,
+    ) -> Option<Box<ConceptWriteError>> {
         match self.check(res.as_ref().map_err(|e| e.clone())) {
             None => None,
-            Some(error) => match error {
+            Some(error) => match *error {
                 ConceptWriteError::ConceptRead { source } => {
                     panic!("Expected logic error, got ConceptRead {:?}", source)
                 }
```
dmitrii-ubskii, line 75: I'd prefer to see something more like this:
```rs
    pub fn check_concept_write_without_read_errors<'a, T: fmt::Debug, E: AsRef<ConceptWriteError>>(
        &self,
        res: &'a Result<T, E>,
    ) -> Option<&'a E> {
        match self.check(res.as_ref().map_err(E::as_ref)) {
            None => None,
            Some(error) => match error {
                ConceptWriteError::ConceptRead { source } => {
                    panic!("Expected logic error, got ConceptRead {:?}", source)
                }
```
So that you don't have to box random errors just for this check.
Also, do we need ownership of the return here at all? I assumed not.

dmitrii-ubskii (commented):
Looks good overall, left some code comments

concept/error.rs
flyingsilverfin: I didn't quite understand what this meant, please feel free to improve it afterward in your PRs

tests/behaviour/steps/params.rs
flyingsilverfin: I tried this and reverted it since i couldn't make it work with one of the call sites - also feel free to tweak it tmrw after i merge

PR #7247, opened 2024-11-11
Stabilise fetch and introduce fetch functions

farost:
We introduce function invocation in `fetch` queries. Fetch can call already existing functions:
```
match
  $p isa person;
fetch {
  "names": [ get_names($p) ],
  "age": get_age($p)
};
``` 
and also use local function blocks:
```
match
  $p isa person;
fetch {
  "names": [
    match
      $p has name $n;
      return { $n };
  ],
  "age": (
    match
      $p has age $a;
      return first $a;
  )
};
```
In the examples above, results collected in concept document lists (`"names": [ ... ]`) represent streams (purposely multiple answers), while single (optional) results (`"age": ...`) represent a single concept document leaf and do not require (although allow) any wrapping.
 

Moreover, we stabilize attribute fetching, allowing you to expect a specific concept document structure based on your schema. This way, if the attribute type is owned with default or key cardinality (`@card(0..1)` or `@card(1..1)`), meaning that there can be at most one attribute of this type, it can be fetched as a single leaf, while other cardinalities force list representation and make your system safe and consistent. For example, a query
```
match
  $p isa person;
fetch {
  $p.*
};
```
can automatically produce a document like
```
{
  "names": [ "Linus", "Torvalds" ],
  "age": 54
}
```
with
```
define
  entity person 
    owns name @card(1..), 
    owns age @card(1..1);
```

Additionally, fetch now returns attributes of all subtypes `x sub name; y sub name;` for `$p.name`, just like regular `match` queries like `match $p owns name $n`.

With this, feel free to construct your fetch statements the way you want:
```
match
  $p isa person, has name $name;
  fetch {
    "info": {
      "name": {
        "from entity": [ $p.person-name ],
        "from var": $name,
      },
      "optional age": $p.age,
      "rating": calculate_rating($p)
    }
  };
```
to expect consistent structure of the results:
```
[
  {
    "info": {
      "name": {
        "from entity": [ "Name1", "Surname1" ],
        "from var": "Name1"
      },
      "optional age": 25,
      "rating": 19.5
    }
  },
  {
    "info": {
      "name": {
        "from entity": [ "Name1", "Surname1" ],
        "from var": "Surname1"
      },
      "optional age": 25,
      "rating": 19.5
    }
  },
  {
    "info": {
      "name": {
        "from entity": [ "Bob" ],
        "from var": "Bob"
      },
      "optional age": null
      "rating": 28.783
    }
  }
]
```

## Implementation
With this PR, `fetch` should be fully implemented for the 3.0 release. The only missing parts that we could be interested in testing are:
* attribute lists (not implemented, left with specific `TODO`s);
* functions and expressions limitations, which can have bugs and failures on their side.


compiler/executable/pipeline.rs
```diff
@@ -49,7 +51,7 @@ pub enum ExecutableStage {
     Offset(Arc<OffsetExecutable>),
     Limit(Arc<LimitExecutable>),
     Require(Arc<RequireExecutable>),
-    Reduce(Arc<ReduceExecutable>),
+    Reduce(Arc<ReduceStageExecutable>),
```
farost, line 54: `ReduceExecutable`, used for return operations as well, doesn't need to contain output positions mappings. That's why we create a new `ReduceStageExecutable`, that contains `ReduceExecutable` and these output mappings.

encoding/value/decimal_value.rs
```diff
@@ -254,10 +254,8 @@ impl fmt::Display for Decimal {
                 fractional /= 10;
             }
 
-            // count number of leading 0's that have to be represented
-            let digits = fractional.ilog10() as u64; // `fractional` cannot be a power of 10 here, so ilog10 correctly gives the number of digits
-            let leading_0s = FRACTIONAL_PART_DENOMINATOR - digits - tail_0s;
-            write!(f, "{}.{:0width$}{}", self.integer, "", digits, width = leading_0s as usize)?;
+            let fractional_width = FRACTIONAL_PART_DENOMINATOR_LOG10 - tail_0s;
+            write!(f, "{}.{:0width$}", self.integer_part(), fractional, width = fractional_width as usize)?;
```
farost, line 258: Copied what I use for the drivers. This version failed with an overflow (`leading_0s`, with no surprise, contained billions...) and we never tested it before I needed it in concept documents printing.

encoding/value/value.rs
```diff
@@ -428,10 +428,13 @@ impl<'a> fmt::Display for Value<'a> {
             Value::Boolean(bool) => write!(f, "{bool}"),
             Value::Long(long) => write!(f, "{long}"),
             Value::Double(double) => write!(f, "{double}"),
-            Value::Decimal(decimal) => write!(f, "{decimal}[decimal]"),
```
farost, line 431: I do the same formatting as in the rust driver here.

executor/pipeline/fetch.rs
```diff
@@ -32,24 +36,41 @@ use lending_iterator::LendingIterator;
...
     },
+    read::{
+        pattern_executor::PatternExecutor, step_executor::create_executors_for_function,
+        tabled_functions::TabledFunctions,
+    },
     row::MaybeOwnedRow,
     ExecutionInterrupt,
 };
 
+macro_rules! execute_once_error_if_has_second {
```
farost, line 55: Useful for single rows final validation.

executor/pipeline/fetch.rs
```diff
@@ -297,16 +380,21 @@ fn execute_attribute_single<'a>(
     snapshot: Arc<impl ReadableSnapshot>,
     thing_manager: Arc<ThingManager>,
 ) -> Result<DocumentLeaf, FetchExecutionError> {
-    let mut iter = object.get_has_type_unordered(snapshot.as_ref(), thing_manager.as_ref(), attribute_type);
-    let attribute = iter.next();
-    match attribute {
-        None => Ok(DocumentLeaf::Empty),
-        Some(Err(err)) => Err(FetchExecutionError::ConceptReadError { source: err }),
-        Some(Ok((attribute, count))) => {
+    let mut iter = prepare_attribute_type_has_iterator(object, attribute_type.clone(), &snapshot, &thing_manager)?;
```
farost, line 383: We fetch all attribute `sub`s for `$p.name`

executor/pipeline/pipeline.rs
```diff
@@ -186,15 +193,14 @@ impl<Snapshot: WritableSnapshot + 'static> Pipeline<Snapshot, WritePipelineStage
         let output_variable_positions = executable_stages.last().unwrap().output_row_mapping();
         let context = ExecutionContext::new(Arc::new(snapshot), thing_manager, parameters);
         let mut last_stage = WritePipelineStage::Initial(Box::new(InitialStage::new_empty(context)));
+        // TODO: Receive as arg?
+        let executable_functions = Arc::new(ExecutableFunctionRegistry::empty());
```
farost, line 197: Not sure if we should receive it from somewhere?

ir/translation/fetch.rs
```diff
@@ -54,27 +55,40 @@ use crate::{
...
-    match &object.body {
+    match &typeql_object.body {
         TypeQLFetchObjectBody::Entries(entries) => {
             let mut object = HashMap::new();
+            let mut unique_keys: HashSet<&str> = HashSet::new();
             for entry in entries {
                 let (key, value) = (&entry.key, &entry.value);
-                let key_id = register_key(parent_context, key);
-                object.insert(key_id, translate_fetch_some(snapshot, parent_context, function_index, value)?);
+                if !unique_keys.insert(key.value.as_ref()) {
```
farost, line 80: A document is a map, which should contain unique keys.

ir/translation/fetch.rs
```diff
@@ -121,52 +143,54 @@ fn translate_fetch_list(
...
             let mut local_context = parent_context.clone();
-            let body = translate_function_block(snapshot, function_index, &mut local_context, block)
+            let body = translate_function_block(snapshot, function_index, &mut local_context, value_parameters, block)
                 .map_err(|err| FetchRepresentationError::FunctionRepresentation { declaration: block.clone() })?;
-            let args = find_function_body_arguments(parent_context, &body);
-            if !body.return_operation().is_stream() {
-                Err(Box::new(FetchRepresentationError::ExpectedStreamInlineFunction { declaration: block.clone() }))
-            } else {
-                Ok(FetchSome::SingleFunction(create_anonymous_function(local_context, args, body)))
+            if !body.return_operation.is_scalar() && !matches!(body.return_operation, ReturnOperation::ReduceReducer(_))
```
farost, line 182: The special case of `return count(), count(), count();` is allowed

query/tests/fetch.rs
```diff
@@ -85,23 +85,17 @@ fetch {
...
-#        match
-#        $x has name $name;
-#        return first $name;
-#    ),
-#    "reduce answer block": (
-#        match
-#        $x has name $name;
-#        return count($name);
-#    ),
-#    "list positional return block": [
```
farost, line 99: Is incorrect, we can't return a stream of tuples.

tests/behaviour/steps/json.rs
farost, line 1: I implement the document in a similar way we did in the Rust driver.  But I use it only for testing. That's why I allow myself a little bit of copypaste without too much thinking about some of the potentially not really required details.

tests/behaviour/steps/params.rs
```diff
@@ -264,7 +264,7 @@ impl FromStr for IsEmptyOrNot {
 }
 
 #[derive(Debug, Parameter)]
-#[param(name = "contains_or_doesnt", regex = "(contain|do not contain)")]
+#[param(name = "contains_or_doesnt", regex = "(contain|contains|do not contain|does not contain)")]
```
farost, line 267: I got tired of `*thing* contain` and `*thing* do not`......... Plus, I've written drivers documents steps with `contains` and `does not contain`, so I decided to reuse the same steps for core.

executor/read/step_executor.rs
```diff
@@ -180,10 +185,22 @@ pub(crate) fn create_executors_for_function(
...
+        }
+        ExecutableReturn::Single(selector, positions) => {
+            steps.push(StepExecutors::ReshapeForReturn(positions.clone()));
+            let step = match selector {
+                SingleSelector::First => StreamModifierExecutor::new_first(PatternExecutor::new(steps)),
+                SingleSelector::Last => StreamModifierExecutor::new_last(PatternExecutor::new(steps)),
+            };
+            Ok(vec![step.into()])
+        }
+        ExecutableReturn::Check => todo!("ExecutableReturn::Check"),
```
farost, line 198: I could leave everything unimplemented, but I really wanted to test it with fetch. I'm not that interested in `Check`, so it's left with a todo.

compiler/executable/pipeline.rs
```diff
@@ -145,12 +147,14 @@ pub fn compile_stages_and_fetch(
     (HashMap<Variable, VariablePosition>, Vec<ExecutableStage>, Option<Arc<ExecutableFetch>>),
     ExecutableCompilationError,
 > {
+    let selected_variables = variable_registry.variable_names().keys().copied().collect_vec();
```
farost, line 150: We used to use all the variables down the pipeline. However, this is incorrect because we also pass anonymous variables, and later, we fail to find positions for them.
Here we introduce selected variables, that are expected to be used later.

compiler/executable/pipeline.rs
```diff
@@ -171,14 +175,31 @@ pub(crate) fn compile_pipeline_stages(
...
     annotated_stages: Vec<AnnotatedStage>,
     input_variables: impl Iterator<Item = Variable>,
+    selected_variables: &Vec<Variable>,
 ) -> Result<(HashMap<Variable, VariablePosition>, Vec<ExecutableStage>), ExecutableCompilationError> {
     let mut executable_stages: Vec<ExecutableStage> = Vec::with_capacity(annotated_stages.len());
     let input_variable_positions =
         input_variables.enumerate().map(|(i, var)| (var, VariablePosition::new(i as u32))).collect();
+
     for stage in annotated_stages {
+        let selected_variables = selected_variables.iter().cloned().chain(stage.referenced_variables()).collect();
```
farost, line 185: On a new stage, we need not only the final selected variable, but also the variables our stages require, so we add it here.

compiler/executable/pipeline.rs
```diff
@@ -182,7 +182,13 @@ pub(crate) fn compile_pipeline_stages(
         input_variables.enumerate().map(|(i, var)| (var, VariablePosition::new(i as u32))).collect();
 
     for stage in annotated_stages {
-        let selected_variables = selected_variables.iter().cloned().chain(stage.referenced_variables()).collect();
+        // TODO: We can filter out the variables that are no longer needed in the future stages, but are carried as selected variables from the previous one
```
farost, line 185: I'm scared of breaking too many existing things if I do so, and I just want to merge my stable fetch.

executor/pipeline/fetch.rs
```diff
@@ -316,27 +404,74 @@ fn execute_attributes_list<'a>(
...
+        .get_subtypes_transitive(snapshot.as_ref(), thing_manager.type_manager())
+        .map_err(|source| FetchExecutionError::ConceptRead { source })?;
+    let attribute_types = TypeAPI::chain_types(attribute_type.clone(), subtypes.into_iter().cloned());
+
+    object
+        .get_has_types_range_unordered(snapshot.as_ref(), thing_manager.as_ref(), attribute_types)
+        .map_err(|source| FetchExecutionError::ConceptRead { source })
+}
+
+fn execute_single_function<Snapshot: ReadableSnapshot + 'static>(
```
dmitrii-ubskii, line 443: This function doesn't execute though, right? I t creates and seeds the executor.

executor/pipeline/fetch.rs
```diff
@@ -316,27 +404,74 @@ fn execute_attributes_list<'a>(
...
+    }
+    let args = MaybeOwnedRow::new_owned(args, row.multiplicity());
+
+    let step_executors = create_executors_for_function(&snapshot, &thing_manager, &functions_registry, function)
+        .map_err(|err| FetchExecutionError::ConceptRead { source: err })?;
+    let mut pattern_executor = PatternExecutor::new(step_executors);
+    pattern_executor.prepare(FixedBatch::from(args));
+
+    let execution_context =
+        Arc::new(ExecutionContext::new(snapshot.clone(), thing_manager.clone(), parameters.clone()));
```
dmitrii-ubskii, line 465: Needless `.clone()`s.
On the other hand, can this function accept an `Arc<ExecutionContext>` rather than reconstruct it?

dmitrii-ubskii (changes requested):

compiler/executable/pipeline.rs
farost: @dmitrii-ubskii With `ExecutableStage::Reduce(Arc<ReduceExecutor>)`?

executor/pipeline/fetch.rs
farost: Sure

compiler/annotation/fetch.rs
```diff
@@ -268,37 +275,41 @@ fn annotate_some(
...
+        match owner_type.kind() {
+            Kind::Entity | Kind::Relation => {}
+            kind @ (Kind::Attribute | Kind::Role) => {
+                return Err(AnnotationError::FetchSingleAttributeCannotBeOwnedByKind {
+                    var: owner.to_owned(),
+                    kind: kind.to_string(),
+                    attribute: attribute_type.get_label(snapshot, type_manager).unwrap().name().as_str().to_owned(),
+                })
+            }
+        }
```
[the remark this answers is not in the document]
farost: Ha, interesting!

compiler/annotation/mod.rs
```diff
@@ -24,20 +24,22 @@ typedb_error!(
     pub AnnotationError(component = "Query annotation", prefix = "QUA") {
         Unimplemented(0, "Unimplemented: {description}", description: String),
         TypeInference(1, "Type inference error while compiling query annotations.", ( typedb_source : TypeInferenceError )),
-        PreambleTypeInference(2, "Type inference error while compiling query premable functions.", ( typedb_source : Box<FunctionAnnotationError> )),
+        PreambleTypeInference(2, "Type inference error while compiling query preamable functions.", ( typedb_source : Box<FunctionAnnotationError> )),
```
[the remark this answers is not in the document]
farost: Lol.... I've broken it differently while rebasing

compiler/annotation/pipeline.rs
```diff
@@ -74,6 +75,26 @@ pub enum AnnotatedStage {
...
+            AnnotatedStage::Delete { block, .. } => Box::new(block.variables()),
+            AnnotatedStage::Select(select) => Box::new(select.variables.iter().cloned()),
+            AnnotatedStage::Sort(sort) => Box::new(sort.variables.iter().map(|sort_variable| sort_variable.variable())),
+            AnnotatedStage::Offset(_) => Box::new(iter::empty()),
+            AnnotatedStage::Limit(_) => Box::new(iter::empty()),
+            AnnotatedStage::Require(_) => Box::new(iter::empty()),
+            AnnotatedStage::Reduce(reduce, _) => Box::new(reduce.variables()),
+        };
+        Box::new(variables.filter(move |variable| variable_registry.get_variable_name(variable.clone()).is_some()))
+    }
```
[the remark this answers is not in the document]
farost: Interesting, didn't think that it would consider it as a single type.

compiler/executable/function.rs
```diff
@@ -40,11 +41,12 @@ pub struct ExecutableFunction {
     // pub plan_cost: f64, // TODO: Where do we fit this in?
 }
 
+#[derive(Debug)]
 pub enum ExecutableReturn {
     Stream(Vec<VariablePosition>),
     Single(SingleSelector, Vec<VariablePosition>),
     Check,
-    Reduce(Vec<ReduceInstruction<VariablePosition>>),
+    Reduce(Arc<ReduceExecutable>),
```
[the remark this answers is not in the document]
farost: They could. I made an Arc because I didn't want to clone the whole `ReduceExecutable`. But we just clone vectors of positions in many places, so I didn't really care about them before. Shall I `Arc` them?

concept/thing/thing_manager/validation/validation.rs
```diff
@@ -50,10 +50,10 @@ macro_rules! create_data_validation_type_abstractness_error_methods {
             ) -> Box<DataValidationError> {
                 debug_assert!(constraint.description().unwrap_abstract().is_ok());
                 let constraint_source = constraint.source();
-                let error_source = Box::new(ConstraintError::ViolatedAbstract);
+                let typedb_source = Box::new(ConstraintError::ViolatedAbstract {});
```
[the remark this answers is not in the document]
farost: Because it is a `typedb_error!` now

concept/type_/constraint.rs
```diff
@@ -42,11 +37,11 @@ macro_rules! unwrap_constraint_description_methods {
         fn $method_name:ident() -> $return_type:ident = $target_enum:ident;
     )*) => {
         $(
-            pub(crate) fn $method_name(self) -> Result<$return_type, Box<ConstraintError>> {
+            pub fn $method_name(self) -> Result<$return_type, Box<ConstraintError>> {
                 with_constraint_description!(
                     self,
                     $target_enum,
-                    Err(Box::new(ConstraintError::CannotUnwrapConstraint(stringify!($target_enum).to_string()))),
+                    Err(Box::new(ConstraintError::CannotUnwrap { type_: stringify!($target_enum).to_string() })),
```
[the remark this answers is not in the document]
farost: Good point!

executor/pipeline/fetch.rs
```diff
@@ -194,12 +289,14 @@ fn execute_fetch_some(
...
+                    Err(FetchExecutionError::FetchAttributesOfList {})
+                }
             }
         }
-        FetchSomeInstruction::ListAttributesFromList(_, _) => {
-            Err(FetchExecutionError::Unimplemented { description: "List attributes are not available yet.".to_string() })
-        }
+        FetchSomeInstruction::ListAttributesFromList(_, _) => Err(FetchExecutionError::Unimplemented {
+            description: "List attributes are not available yet.".to_string(),
+        }),
```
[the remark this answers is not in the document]
farost: Good point. I had not been thinking about it enough before.

executor/pipeline/fetch.rs
farost: Fooof. I'll try.

executor/pipeline/fetch.rs
```diff
@@ -95,19 +120,16 @@ fn execute_fetch_some(
     snapshot: Arc<impl ReadableSnapshot + 'static>,
     thing_manager: Arc<ThingManager>,
     parameters: Arc<ParameterRegistry>,
+    functions_registry: Arc<ExecutableFunctionRegistry>,
     row: MaybeOwnedRow<'_>,
     interrupt: ExecutionInterrupt,
 ) -> Result<DocumentNode, FetchExecutionError> {
     match fetch_some {
```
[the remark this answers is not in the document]
farost: You're right. I guess I don't feel enough "ownership" when I meet unknown components and hesitate to make more refactoring.

ir/translation/mod.rs
```diff
@@ -29,7 +29,6 @@ pub mod writes;
 pub struct TranslationContext {
     pub variable_registry: VariableRegistry, // TODO: Unpub
     visible_variables: HashMap<String, Variable>,
-    pub parameters: ParameterRegistry,
 }
```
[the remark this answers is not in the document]
farost: If I remember correctly, we thought that it's more correct to separate them as we have multiple `parent` and `sub` contexts which should contain isolated information, while parameters should be collected and saved for the whole execution, so the parent pipeline can access them while collecting the document:
```
fetch {
  "I am saved in the context": [ 
    match
      ...
      fetch {
        "I might be forgotten": ...
      };
}
};
```

ir/pipeline/function.rs
```diff
@@ -84,6 +90,15 @@ impl ReturnOperation {
         }
     }
 
+    pub(crate) fn is_scalar(&self) -> bool {
+        match self {
+            Self::Stream(vars) => vars.len() < 2,
+            Self::Single(_, vars) => vars.len() < 2,
+            Self::ReduceCheck() => true,
+            Self::ReduceReducer(reducers) => reducers.len() < 2,
```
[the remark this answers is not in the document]
farost: Not `ReduceSingle` for sure. I'll leave it for @flyingsilverfin

executor/read/pattern_executor.rs
```diff
@@ -360,6 +360,15 @@ impl PatternExecutor {
                         input: input.clone().into_owned(),
                     }));
                 }
+                StreamModifierExecutor::Last { inner } => {
```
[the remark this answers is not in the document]
farost: Just like `first`:
```
with
fun get_name($p_arg: person) -> name:
match
  $p_arg has person-name $name;
sort $name;
return last $name;
```

Actually, [the spec](https://github.com/typedb/typedb/blob/3.0/docs/spec.md#theory-function-evaluation) also contains `random`, but we only have `first` and `last` in `typeql` now.

executor/pipeline/fetch.rs
```diff
@@ -119,70 +141,143 @@ fn execute_fetch_some(
...
+                None => return Ok(DocumentNode::Leaf(DocumentLeaf::Empty)),
+            };
+
+            let mut row_iter = batch.into_iter();
+            let document = match execute_once_error_if_has_second!(
+                row_iter.next(),
+                FetchExecutionError::FetchSingleFunctionNotSingle { func_name: "func".to_string() }
+            ) {
+                Some(row) => {
+                    let mut row_iter = row.row().iter();
```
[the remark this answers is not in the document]
farost: Renamed

executor/read/immediate_executor.rs
```diff
@@ -660,11 +660,14 @@ impl AssignExecutor {
...
-                .map(|&pos| (pos, ExpressionValue::try_from_value(input_row.get(pos).to_owned(), context).unwrap()))
-                .collect();
+            let mut input_variables = HashMap::new();
+            for &pos in &self.inputs {
+                input_variables.insert(
+                    pos,
+                    ExpressionValue::try_from_value(input_row.get(pos).to_owned(), context)
+                        .map_err(|source| ReadExecutionError::ExpressionEvaluate { source })?,
+                );
+            }
```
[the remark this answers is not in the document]
farost: Your comment gave a hint to map to a result of pairs =)

executor/pipeline/fetch.rs
farost: It sounds massive, so I just removed `clone()`s, thanks!

executor/read/immediate_executor.rs
```diff
@@ -660,14 +660,16 @@ impl AssignExecutor {
...
+            let input_variables = self
+                .inputs
+                .iter()
+                .map(|&pos| {
+                    let value = input_row.get(pos).to_owned();
+                    let expression_value = ExpressionValue::try_from_value(value, context)
+                        .map_err(|source| ReadExecutionError::ExpressionEvaluate { source })?;
+                    Ok((pos, expression_value))
+                })
+                .collect::<Result<_, _>>()?;
```
dmitrii-ubskii, line 672:
```suggestion
                .try_collect()?;
```

dmitrii-ubskii (approved):

compiler/executable/pipeline.rs
dmitrii-ubskii: No. `FooExecutable`s are stages, and contain `BarExecutor`s.

farost: Okay... So my current
```
#[derive(Debug, Clone)]
pub struct ReduceStageExecutable {
    pub reduce_executable: Arc<ReduceExecutable>,
    pub output_row_mapping: HashMap<Variable, VariablePosition>, // output_row = (group_vars, reduce_outputs)
}

#[derive(Debug, Clone)]
pub struct ReduceExecutable {
    pub reductions: Vec<ReduceInstruction<VariablePosition>>,
    pub input_group_positions: Vec<VariablePosition>,
}
```
to be changed to
```
#[derive(Debug, Clone)]
pub struct ReduceExecutable {
    pub executor: Arc<ReduceExecutor>,
    pub output_row_mapping: HashMap<Variable, VariablePosition>, // output_row = (group_vars, reduce_outputs)
}

#[derive(Debug, Clone)]
pub struct ReduceExecutor {
    pub reductions: Vec<ReduceInstruction<VariablePosition>>,
    pub input_group_positions: Vec<VariablePosition>,
}
```
@flyingsilverfin does it make sense to you?

executor/read/immediate_executor.rs
farost: Thanks, will push with other fixes if there are

compiler/executable/pipeline.rs
flyingsilverfin: Err an `Executable` shouldn't have a reference to an `Executor` - the other way around however could be true.

flyingsilverfin: Eg: 
```
pub struct FetchStageExecutor<Snapshot: ReadableSnapshot> {
    executable: Arc<ExecutableFetch>,
    _phantom: PhantomData<Snapshot>,
}
```

flyingsilverfin: Our naming convention is `TypeQL -> Translated -> Annotated -> Executable -> StageExecutor(executable)`. The main case where we have an extra layer is for `match`: `MatchExecutable -> MatchStageExecutable(MatchExecutor(MatchExecutable))`, although the extra `MatchExecutor` is created within the iterator for the `MatchStageExecutable`

executor/pipeline/fetch.rs
```diff
@@ -95,125 +121,144 @@ fn execute_fetch_some(
...
+            .map_err(|err| FetchExecutionError::ReadExecution { typedb_source: Box::new(err) })?,
+        FetchExecutionError::FetchSingleFunctionNotSingle { func_name: "func".to_string() } // TODO: Can we get function name here?
+    );
+
+    let batch = match batch {
+        Some(batch) => batch,
+        None => return Ok(DocumentNode::Leaf(DocumentLeaf::Empty)),
+    };
+
+    let mut row_iter = batch.into_iter();
```
flyingsilverfin, line 227: I suspect it might be neat/nice to have another type of `BatchRowIterator` that you build that returns you a Row iterator instead of  Batch iterator -> Row iterator within a batch? So you don't have to deal with this manual stuff...

executor/pipeline/fetch.rs
```diff
@@ -222,6 +267,122 @@ fn execute_object(
...
+    while let Some(batch) = pattern_executor
+        .compute_next_batch(&execution_context, &mut interrupt, &mut tabled_functions, &mut suspend_points)
+        .map_err(|err| FetchExecutionError::ReadExecution { typedb_source: Box::new(err) })?
+    {
+        for row in batch {
+            for value in row {
+                nodes.push(variable_value_to_document(value.clone())?);
+            }
+        }
+    }
```
flyingsilverfin, line 302: would also squash this

executor/pipeline/fetch.rs
```diff
@@ -316,27 +466,71 @@ fn execute_attributes_list<'a>(
...
+fn prepare_attribute_type_has_iterator<'a>(
+    object: impl ObjectAPI<'a>,
+    attribute_type: AttributeType<'static>,
+    snapshot: &Arc<impl ReadableSnapshot>,
+    thing_manager: &Arc<ThingManager>,
+) -> Result<HasIterator, FetchExecutionError> {
+    let subtypes = attribute_type
+        .get_subtypes_transitive(snapshot.as_ref(), thing_manager.type_manager())
+        .map_err(|source| FetchExecutionError::ConceptRead { source })?;
+    let attribute_types = TypeAPI::chain_types(attribute_type.clone(), subtypes.into_iter().cloned());
```
flyingsilverfin, line 498: great thank you for fixing t his!

flyingsilverfin: Oh actually this will be incorrect - you need to filter the iterator against the attribute types again! The `Range` literally means range from Min Attributes -> Max attribute! (eg if you say attribute range `(0 to 10)` where you really want `(0, 1, 5, 10)` you'll get `(0, 1, 2, 3...10)`!

flyingsilverfin: Dmitrii was meant to fix the API: it should take a `KeyRange<AttributeType>` instead of an Iterator of attribute types! Do you think you could edit it in the ThingManager `get_has_from_thing_to_type_range_unordered` (see `get_has_reverse_by_attribute_and_owner_type_range` for an example of what it should look like)

executor/pipeline/pipeline.rs
flyingsilverfin: Yeah krishnan should be wiring this through in his PRs at some point!

tests/behaviour/query/language/fetch.rs
```diff
@@ -0,0 +1,23 @@
...
+#[tokio::test]
+async fn test() {
+    // Bazel specific path: when running the test in bazel, the external data from
+    // @vaticle_typedb_behaviour is stored in a directory that is a sibling to
+    // the working directory.
+    #[cfg(feature = "bazel")]
+    let path = "../vaticle_typedb_behaviour/query/language/fetch.feature";
+
+    #[cfg(not(feature = "bazel"))]
+    let path = "bazel-typedb/external/vaticle_typedb_behaviour/query/language/fetch.feature";
```
flyingsilverfin, line 20: Make sure you rebase and use the new format after Dmitrii's merge - this won't be right anymore

compiler/executable/pipeline.rs
farost: I'm kinda lost as I feel like some of the statements are contradicting. 

With my current naming, we have:

`executor/pipeline/reduce`:
```
pub struct ReduceStageExecutor<PreviousStage> {
    executable: Arc<ReduceStageExecutable>,
    previous: PreviousStage,
}
```

`compiler/executable/reduce`:
```
pub struct ReduceStageExecutable {
    pub reduce_executable: Arc<ReduceExecutable>,
    pub output_row_mapping: HashMap<Variable, VariablePosition>, // output_row = (group_vars, reduce_outputs)
}

pub struct ReduceExecutable {
    pub reductions: Vec<ReduceInstruction<VariablePosition>>,
    pub input_group_positions: Vec<VariablePosition>,
}
```

What's the correct naming, then?

concept/type_/annotation.rs
```diff
@@ -156,6 +156,10 @@ impl AnnotationCardinality {
         self == &Self::unchecked()
     }
 
+    pub fn is_scalar(&self) -> bool {
```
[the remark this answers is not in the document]
farost: I like this function as it makes searching for its usages in different places easier to analyze what depends on it, while `.end == Some(1)` can be written differently. I'll try `is_bounded_to_one`, as I also want to use the same term in concept api.

concept/type_/mod.rs
```diff
@@ -313,6 +313,19 @@ pub trait OwnerAPI<'a>: TypeAPI<'a> {
         attribute_type: AttributeType<'static>,
     ) -> Result<HashSet<CapabilityConstraint<Owns<'static>>>, Box<ConceptReadError>>;
 
+    fn is_owned_attribute_type_scalar(
```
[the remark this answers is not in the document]
farost: I've changed it to `bounded_to_one` for now.

executor/pipeline/fetch.rs
farost: Check the usages of this function. I just prepare the iterator here, and I additionally filter out non-suitable attributes.

farost:
```
let mut iter = prepare_attribute_type_has_iterator(object, attribute_type.clone(), &snapshot, &thing_manager)?;

    while let Some(result) = iter.next() {
        let (has, count) = result.map_err(|source| FetchExecutionError::ConceptRead { source })?;
        let attribute = has.attribute();
        let suitable = attribute
            .type_()
            .is_subtype_transitive_of_or_same(snapshot.as_ref(), thing_manager.type_manager(), attribute_type.clone())
            .map_err(|source| FetchExecutionError::ConceptRead { source })?;
        if suitable {
        .....
        
        ```

tests/behaviour/query/language/fetch.rs
farost: Sure, I was waiting for it

executor/pipeline/fetch.rs
farost: I left a todo here to move forward this week as I didn't quickly understand how to make it in a good way with the lack of understanding of the whole local codebase.

flyingsilverfin: ok

executor/pipeline/fetch.rs
flyingsilverfin: Ah ok thanks!

compiler/executable/pipeline.rs
flyingsilverfin: Why do you have separate `ReduceStageExecutable` and `ReduceExecutable`? We don't have a **StageExecutable** for any other tyoe pf stage? We only ever have a `XXXStageExcutor` that has an `XXXExecutable` inside?

farost: @flyingsilverfin I've explained it in the first comment of this thread. I could send `output_row_mapping` as an additional argument everywhere, but it doesn't really make sense, and more comfortable to wrap it together with the executable.
Specifically, I had a case in `compile_return_operation` in `compiler`, where I didn't have and didn't need `output_row_mapping` as we are working with the return statement already, where the mapping is direct and automatic. So we made this refactoring with @krishnangovindraj

flyingsilverfin (approved):

PR #7249, opened 2024-11-13
Variable position fix

cxdorn:
Variable positions in inserts were filtered before they were enumerated, which meant that variable position number were passed along incorrectly.

### DIFF
--- a/compiler/executable/pipeline.rs
+++ b/compiler/executable/pipeline.rs
@@ -59,8 +59,8 @@ impl ExecutableStage {
             ExecutableStage::Insert(executable) => executable
                 .output_row_schema
                 .iter()
-                .filter_map(|opt| opt.map(|(v, _)| v))
                 .enumerate()
+                .filter_map(|(i, opt)| opt.map(|(v, _)| (i,v)))
                 .map(|(i, v)| (v, VariablePosition::new(i as u32)))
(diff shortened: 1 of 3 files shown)


krishnangovindraj (approved):
Looks good.
The insert executor will need a larger rewrite in future where it will compact the row.

compiler/executable/pipeline.rs
```diff
@@ -59,8 +59,8 @@ impl ExecutableStage {
             ExecutableStage::Insert(executable) => executable
                 .output_row_schema
                 .iter()
-                .filter_map(|opt| opt.map(|(v, _)| v))
                 .enumerate()
+                .filter_map(|(i, opt)| opt.map(|(v, _)| (i,v)))
```
farost, line 63: `(i,v)` -> `(i, v)`, let's show good formatting examples!

Btw, it's a good practice to run (I assume that it should fix this) 
`bazel run @rules_rust//:rustfmt --@rules_rust//:rustfmt.toml=//:rustfmt.toml` before you commit, you can also set it as a pre-commit hook. We'll add the CI/CD formatting check which will bonk your head if you don't do it.

cxdorn: thanks for this!

PR #7254, opened 2024-11-14
Guarantee transaction causality

flyingsilverfin:

Under concurrent write workloads, causality can appear to be violated. For example:

```
- Open tx1, tx2
- start tx1 commit (validation begins)
- start tx2 commit (validation begins and ends, no conflict with tx1!)
- open tx3
- end tx1 commit (validation finishes)
```
When we open `tx3`, we end up with a snapshot that is actually from _before_ `tx1`, even though `tx2` has committed - because we don't know the status of `tx1` yet, and in our current simplified model, transaction are assigned a linear commit order decided at WAL write time, the read watermark remains before `tx1` until it finishes validating.

In this scenario, a client that commits a transaction can actually end up opening the next transaction snapshot _before_ the last commit that successfully returned.

After this change, when opening a transaction, TypeDB wait until the currently pending transactions all finish, guaranteeing we see the latest possible data version. The assumption is that 1) validation in general is a small amount of time and 2) is fully concurrent, so the wait time should be very small, and only occur under large concurrently committing transactions.

## Motivation

TypeDB by default should guarantee causality across transactions. In particular, when one client/driver submits and successfully commits a transaction, they should always read their own preceding writes.

## Implementation

We add a busy wait loop (with 100 microsecond sleeps) at snapshot open, that waits until the storage watermark rises to at least the latest pending sequence number.

## Analysis and cost

Short of changing TypeDB's logically-sequential transaction model, we have two alternatives to guarantee causality

1) on commit, we wait until all preceding commits finish validation and the watermark rises past the current commit sequence number.
2) on open, we take the latest committed sequence number, and wait for the watermark to rise until that point, before opening the transaction

Option 1) implies that all write transactions, even ones that don't interact at all, are essentially queued behind the slowest concurrent validating commit (let's say this takes `t` time). It also means that the network latency of responding to the user can't be used to hide this slow committer - we can't respond until `t` elapses. So let's say every write transaction takes on average `w` time, we now have write transactions that can take around `w + t`.

Option 2) implies that not just all write transactions, but also all read transactions, have to wait some amount of time before opening. Similarly to above, we could hit a case where there is a validating commit that finishes after `t` time - but this is paid on transaction open, not commit. As a result, write transactions still take `w + t` time and also read transactions (avg duratino `r`) also take `r + t` time, which is strictly worse.

One optimisation for 2), say 2a) is to have the drivers hold onto the last commit sequence number they saw, so opening a new transaction can open the current watermark as long as it is a newer version that the last commit number. A new connection would fall back to waiting for the currently last pending commit to finish, then proceed. However, this can still lead to unexpected behaviour:
```
- driver 1 opens tx1 @ 0, tx2 @0
- commit tx1 (assigned 1), which is a slow commit taking T time and doesn't return yet (watermark = 0)
- commit tx2 (assigned 2), which is a fast commit, and returns. Driver 1 receives that the latest sequence number is 2. (watermark = 0)
- open a new driver 2, open tx3 and tx4, which waits for watermark to rise to 2, and opens both (watermark = 2)
- commit tx3 (assigned 3), which is a slow commit taking T2 time, and doesn't return yet (watermark = 2)
- commit tx4 (assigned 4), which is a fast commit and returns. Driver 2 receives that the latest sequence number is 4. (watermark = 2)

- Driver 1 opens a new transaction tx5, which must open on a version higher than 2, which is fine, so opens at 2. 
```

Here the apparent causality violation comes from receiving a confirmed commit at 4 in the same program, but not seeing that data from the other driver opening a transaction.

A third option would combine some properties of both variants, option 3). Here, we allow the user to open a transaction any time, and take the currently highest committed sequence number and record this in the transaction, and respond to the driver that the transaction is open. Any subsequent request would pay the network latency again! If by the time any query arrives from the driver to the server, and the watermark has not risen yet to the recorded sequence number, we block the query until it does. This gives us the nice property that we leverage network latency to hide the time of the slowest commit validation `T` (benefit of option 1), while not sequentialising write transactions (benefit of option 2) and also never violating causality in any case (risk of option 2a).

This PR does not implement option 3), though it is likely the best compromise we can reach for the time being, only implementing option 2) until we can benchmark the consequences in a stable fashion.

### DIFF
--- a/storage/isolation_manager.rs
+++ b/storage/isolation_manager.rs
@@ -53,7 +53,7 @@ impl IsolationManager {
         IsolationManager {
             initial_sequence_number: next_sequence_number,
             timeline: Timeline::new(next_sequence_number),
-            highest_validated_sequence_number: AtomicU64::new(next_sequence_number.number() - 1)
+            highest_validated_sequence_number: AtomicU64::new(next_sequence_number.number() - 1),
         }
     }
 
@@ -246,7 +246,7 @@ impl IsolationManager {
     pub(crate) fn watermark(&self) -> SequenceNumber {
         self.timeline.watermark()
     }
-    
+
     pub(crate) fn highest_validated_sequence_number(&self) -> SequenceNumber {
         SequenceNumber::new(self.highest_validated_sequence_number.load(Ordering::SeqCst))
     }
@@ -377,10 +377,7 @@ impl Timeline {
     // The whole of the timeline uses the underlying u64
     fn new(next_sequence_number: SequenceNumber) -> Timeline {
         let windows = VecDeque::from([Arc::new(TimelineWindow::new(next_sequence_number))]);
-        Timeline {
-            windows: RwLock::new(windows),
-            watermark: AtomicU64::new(next_sequence_number.number() - 1),
-        }
+        Timeline { windows: RwLock::new(windows), watermark: AtomicU64::new(next_sequence_number.number() - 1) }
     }
 
     fn may_free_windows(&self) {
--- a/storage/isolation_manager.rs
+++ b/storage/isolation_manager.rs
@@ -39,6 +39,7 @@ use crate::{
 pub(crate) struct IsolationManager {
     initial_sequence_number: SequenceNumber,
     timeline: Timeline,
+    highest_validated_sequence_number: AtomicU64,
 }
 
 impl fmt::Display for IsolationManager {
@@ -52,6 +53,7 @@ impl IsolationManager {
         IsolationManager {
             initial_sequence_number: next_sequence_number,
             timeline: Timeline::new(next_sequence_number),
+            highest_validated_sequence_number: AtomicU64::new(next_sequence_number.number() - 1)
         }
     }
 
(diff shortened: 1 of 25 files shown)


storage/isolation_manager.rs
```diff
@@ -365,13 +370,18 @@ struct Timeline {
     // We can adjust the Window size to amortise the cost of the read-write locks to maintain the timeline
     windows: RwLock<VecDeque<Arc<TimelineWindow<TIMELINE_WINDOW_SIZE>>>>,
     watermark: AtomicU64,
+    latest_utilised_sequence_number: AtomicU64,
```
dmitrii-ubskii, line 373: What does `utilised` mean? Claimed? Committed?

storage/storage.rs
```diff
@@ -159,42 +162,51 @@ impl<Durability> MVCCStorage<Durability> {
...
-         */
-
-        let open_sequence_number = self.isolation_manager.watermark();
+        // guarantee external consistency: we always await the latest snapshots to finish
+        let possible_sequence_number = self.isolation_manager.highest_possible_watermark();
+        let mut open_sequence_number = self.snapshot_watermark();
+        while open_sequence_number < possible_sequence_number {
+            sleep(Duration::from_micros(WATERMARK_WAIT_INTERVAL_MICROSECONDS));
+            open_sequence_number = self.isolation_manager.watermark();
+        }
```
dmitrii-ubskii, line 171: This should be a function, e.g. `wait_for_watermark_to_reach(SequenceNumber) -> SequenceNumber`

dmitrii-ubskii (approved):
Makes sense, left a couple comments

PR #7256, opened 2024-11-18
Exit on panic

dmitrii-ubskii:

In Rust, whenever a panic is encountered in a background thread, the main thread is normally unaffected until it either joins the background thread, or attempts to lock a mutex that was held by the panicking thread.

We override the default behaviour to `exit` from the process with code `1`.

### DIFF
--- a/main.rs
+++ b/main.rs
@@ -13,6 +13,8 @@ use server::parameters::config::Config;
 
 #[tokio::main]
 async fn main() {
+    setup_abort_on_panic();
+
     print_ascii_logo(); // very important
     initialise_logging_global();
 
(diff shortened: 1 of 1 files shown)


main.rs
```diff
@@ -34,3 +36,13 @@ fn get_configuration() -> Config {
 fn print_ascii_logo() {
     println!("{ASCII_LOGO}");
 }
+
+fn setup_abort_on_panic() {
+    std::panic::set_hook({
+        let default_panic = std::panic::take_hook();
+        Box::new(move |info| {
+            default_panic(info);
+            std::process::exit(1);
```
dmitrii-ubskii, line 45: Note: `abort()` would bail immediately without calling any destructors or exit hooks. `exit()` is more gentle.

flyingsilverfin: Ok, thank you!

flyingsilverfin (approved):

PR #7257, opened 2024-11-18
Implement query executable cache

flyingsilverfin:

We implement the cache (somewhat arbitrarily limited to 100 entries) for compiled executable queries, along with cache maintanance when statistics change significantly or the schema updates.

Query execution without any cache hits still looks like this:
```
Parsing -> Translation (to intermediate representation) -> Annotation -> Compilation -> Execution
```
However, with a cache hit, we now have:
```
Parsing -> Translation ---Cache--> Execution
```
skipping the annotation and compilation/planning phases, which take significant time.

Note that schema transactions don't have a query executable cache, since keeping the cache in-sync when schema operations run can be error prone.

The query cache is a _structural_ cache, which means it will ignore all Parameters in the query: variable names, constants and values, and fetch document keys. Most production systems run a limited set of query structures, only varying values and terms, making a structural cache like this highly effective!

## Motivation

We introduce the structural query cache to improve query execution time: we no amortise the cost of type inference, expression compilation, and query planning. Especially for simple queries, these costs can dominate. 

## Implementation

- We introduce a Moka cache with a constant number of entries (currently 100), which uses structural hashing and equality to re-use query plans
- Convert the stateless `QueryManager` into a stateful one, which primarily contains the QueryCache
- Introduce a `total_count` value in Statistics, which represents all statistics added together -- this is used as an indicator of the total data volume, which is tracked and used to flush the statistics cache. Currently, when the data changes by 10% overall, we flush the cache and force query re-planning.

Besides the broader implementation we also:
- simplify the use of the `VariableRegistry` and the `ParameterRegistry`, passing them by reference instead of Arc (or removing them fully) in some methods
- fix a critical bug in the propagation of Concepts across stages, which led to over-writing of values produced by a later `match` stage by the inputs to the match stage


flyingsilverfin:
TODO: add `fetch` and use a real cache!

compiler/annotation/fetch.rs
```diff
@@ -349,18 +348,19 @@ fn annotate_sub_fetch(
...
-    let FetchListSubFetch { context, parameters, input_variables, stages, fetch } = sub_fetch;
+    let FetchListSubFetch { context, input_variables, stages, fetch } = sub_fetch;
     let TranslationContext { mut variable_registry, .. } = context;
     let (annotated_stages, annotated_fetch) = annotate_stages_and_fetch(
         snapshot,
         type_manager,
         schema_function_annotations,
         &mut variable_registry,
-        &parameters,
+        parameters,
```
flyingsilverfin, line 363: Unneeded - fetch parameters are now rolled into the root query paramters

compiler/executable/match_/planner/match_executable.rs
```diff
@@ -192,8 +192,8 @@ impl Display for IntersectionStep {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         write!(
             f,
-            "inputs={:?}, output_size={}, sort_by={}",
-            &self.input_variables, self.output_width, self.sort_variable
+            "bound_vars={:?}, output_size={}, sort_by={}",
```
flyingsilverfin, line 195: "Bound" variables is more accurate because these are only the variables that are mentioned in this step/instruction AND inputs (eg. bound)

concept/thing/statistics.rs
```diff
@@ -52,6 +52,8 @@ pub struct Statistics {
     encoding_version: StatisticsEncodingVersion,
     pub sequence_number: SequenceNumber,
 
+    pub total_count: u64,
```
flyingsilverfin, line 55: Total count of all things is now also kept track of, so we can know when to flush the query cache by % overall change

answer/variable.rs
```diff
@@ -6,16 +6,21 @@
 
 use std::fmt;
 
-use structural_equality::StructuralEquality;
+use structural_equality::{ordered_hash_combine, StructuralEquality};
 
 #[derive(Copy, Clone, Hash, Eq, PartialEq, Ord, PartialOrd)]
 pub struct Variable {
     id: VariableId,
+    anonymous: bool,
```
flyingsilverfin, line 14: This was the easiest way to make a non-anonymous and an anonymous variable not be structurally equal -- otherwise we end up with `match $_ has $_` and `match $x has $y;` as structurally equal.

PR #7262, opened 2024-11-22
Add more typing information into key prefixes to optimise storage lookups

flyingsilverfin:

We incorporate more typing information into the storage lookup operations, making them more longer and specific and reducing the range of data that must be retrieved for any traversal operation. This should make future RocksDB bloom filters more effective, and reduce the amount of data that is read even when bloom filters are not available for the prefix. Overall, we expect this to allow TypeDB to scale better, particularly on larger-than-memory workloads.

## Motivation

We implement a feature that is already available in TypeDB 2.x: using the destination type of a connection traversal (say, a 'has') to restrict the range of data that must be searched. This should improve performance and scalability.

## Implementation

We extend the prefixes used to do storage lookups to a consistent length, where possible, of : 
`[connection type][start concept][end concept type prefix]`
instead of
`[connection type][start concept]`. 

This should allow bloom filters to be more effective when they are added, since we are searching for more specialised prefixes that will hit fewer levels in the RocksDB LSM tree.

## Notes

- It is no long entirely clear that adding the `end concept type prefix` is beneficial when the key range we need to scan is a range of prefixes, rather than just within the one prefix: it seems that logical that Rocks bloom filters only work exactly within 1 prefix, as they are used to position iterators in the LSM tree files.
- If we want our iterators to support scanning beyond the initial prefix without missing data, we will need to manually disable prefix filters:
```
Users can only use prefix bloom filter to read inside a specific prefix. If iterator is outside a prefix, the feature needs to be disabled for specific iterators to prevent wrong results:
```

from: https://github.com/facebook/rocksdb/wiki/Prefix-Seek

Thus, in RocksDB, we will need to be careful to make sure that the key range we wish to scan in an iterator has a consistent prefix (`key_range.start()[..bloom_prefix_length] == key_range.end()[..bloom_prefix_length]`)

### DIFF
--- a/concept/thing/thing_manager.rs
+++ b/concept/thing/thing_manager.rs
@@ -225,9 +225,9 @@ impl ThingManager {
         })
     }
 
-    pub fn get_attributes<'this, Snapshot: ReadableSnapshot>(
-        &'this self,
-        snapshot: &'this Snapshot,
+    pub fn get_attributes<Snapshot: ReadableSnapshot>(
+        &self,
+        snapshot: &Snapshot,
     ) -> Result<AttributeIterator<InstanceIterator<Attribute>>, Box<ConceptReadError>> {
         let has_reverse_start = ThingEdgeHasReverse::prefix_from_prefix(Prefix::VertexAttribute);
         let range = KeyRange::new_within(has_reverse_start, Prefix::VertexAttribute.fixed_width_keys());
@@ -241,9 +241,9 @@ impl ThingManager {
         ))
     }
 
-    pub fn get_attributes_in<'this>(
-        &'this self,
-        snapshot: &'this impl ReadableSnapshot,
+    pub fn get_attributes_in(
+        &self,
+        snapshot: &impl ReadableSnapshot,
         attribute_type: AttributeType,
     ) -> Result<AttributeIterator<InstanceIterator<Attribute>>, Box<ConceptReadError>> {
         let attribute_value_type =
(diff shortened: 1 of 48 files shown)


concept/thing/thing_manager.rs
```diff
@@ -734,36 +910,15 @@ impl ThingManager {
         Ok(attributes)
     }
 
-    pub(crate) fn get_has_from_thing_to_type_range_unordered(
```
flyingsilverfin, line 737: duplicate API, removed

concept/thing/thing_manager.rs
```diff
@@ -386,61 +385,90 @@ impl ThingManager {
...
-        })?
-        else {
+        let Some((value_lower_bound, value_upper_bound)) = Self::get_value_range(attribute_value_type, range)? else {
             return Ok(AttributeIterator::new_empty());
         };
-        let key_range_start = match range_start {
-            Bound::Included(start) => RangeStart::Inclusive(start),
-            Bound::Excluded(start) => RangeStart::Exclusive(start),
+
+        let start_attribute_vertex_prefix_range = match value_lower_bound {
```
flyingsilverfin, line 392: We now take a `impl RangeBounds`, since the API is more restrictive than our `KeyRange` type. Each method here translates the RangeBounds into a slightly different KeyRange, but we consistently extend the range to be as long as possible (including using the type ID `0` when no target type is provided), which should generally hit a longer prefix.

I'm actually not entirely sure, thinking more about how RocksDB uses prefix filters, that this is actually useful. But at least now we're using a consistent prefix length of searches...

PR #7266, opened 2024-11-25
Implement system database and user authentication

lolski:
Implement a system database, a special database that the server will use to store various systems-related information. It will be used to store database server users, but in the future can be extended to store more information.

Bring user authentication and user management functionalities into TypeDB Core through the following changes:
- Implement user management functionality, handling user creation, retrieval, update, and deletion.
- Implement authentication protocol - TypeDB Core drivers are now required to supply credentials when making a connection

## Implementation
- System database:
  - implement the schema definition
  - implement the accessor, through "repository" structs
- User management API: implement the following gRPC services: user creation, retrieval, update, and deletion.
- Authentication API: supply credentials through gRPC headers.
- Driver creation API: Move the TLS setting away from the `Credential` object into a new object `ConnectionSettings`.


system/BUILD
```diff
@@ -0,0 +1,41 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+load("@vaticle_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
+load("@rules_rust//rust:defs.bzl", "rust_library")
+package(default_visibility = ["//visibility:public"])
+
+load("@rules_rust//rust:defs.bzl", "rust_library", "rust_test")
```
farost, line 9: This line is not needed

system/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+    match database_manager.database(SYSTEM_DB) {
+        Some(db) => db,
+        None => {
+            database_manager.create_database(SYSTEM_DB).unwrap();
+            let db = database_manager.database(SYSTEM_DB).unwrap();
+            let tx_mgr = TransactionUtil::new(db.clone());
+            tx_mgr.schema_transaction(|query_mgr, snapshot, type_mgr, thing_mgr, fn_mgr| {
+                let query = typeql::parse_query(SCHEMA).unwrap().into_schema();
+                query_mgr.execute_schema(snapshot, type_mgr, thing_mgr, query).unwrap();
+            }).unwrap();
```
farost, line 48: Too many unwraps in a single function, might be worth putting more `expect("What is expected instead to be unwrapped")` for readability of the crash output

system/repositories.rs
```diff
@@ -0,0 +1,112 @@
...
+    use uuid::Uuid;
+    use crate::util::answer_util::get_string;
+
+    pub fn list(tx: TransactionRead<WALClient>) -> Vec<User> {
+        let query = parse_query(
+            "match (user: $u, password: $p) isa user-password; $u has name $n;"
+        ).unwrap();
+        let (tx, result) =
+            execute_read_pipeline(tx, &query.into_pipeline());
+        let rows = result.unwrap();
```
farost, line 24: Might be same for "expect" as the read pipeline will fail with an error if something is not synced with the schema

system/repositories.rs
```diff
@@ -0,0 +1,112 @@
...
+        let mut users: Vec<User> = vec![];
```
farost, line 25: These 5 lines 25 - 30 can be rewritten as a single `rows.map(|row| ...).collect()`

system/repositories.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    pub fn get(tx: TransactionRead<WALClient>, username: &str) -> Option<(User, Credential)> {
+        let query = parse_query(
+            format!(
+                "match
+                (user: $u, password: $p) isa user-password;
+                $u has name '{username}';
+                $p has hash $h, has salt $s;",
+                username = username
```
farost, line 40: Looks like you don't even need to pass an argument if you write `{username}`

system/util.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub struct TransactionUtil {
+        database: Arc<Database<WALClient>>,
+    }
+
+    impl TransactionUtil {
+        pub fn new(database: Arc<Database<WALClient>>) -> Self {
+            TransactionUtil { database }
+        }
+
+        pub fn schema_transaction<T>(
```
farost, line 24: It really looks like `with_schema_tx!` macro from our tests. Might be cleaner to reuse these macros. Another `Util` class (which's instance is called `mgr`) feels heavily redundant, but might fine for the `system` package... 
I'm not going to fight over it considering our timeline though, but some TODOs can look good here.

system/util.rs
```diff
@@ -0,0 +1,212 @@
...
+                type_manager,
+                thing_manager,
+                function_manager,
+                database,
+                transaction_options,
+            } = TransactionSchema::open(self.database.clone(), TransactionOptions::default());
+            let mut snapshot: SchemaSnapshot<WALClient> = Arc::into_inner(snapshot).unwrap();
+            let result = fn_(QueryManager::new(), &mut snapshot, &type_manager, &thing_manager, &function_manager);
+            let tx = TransactionSchema::from(snapshot, type_manager, thing_manager, function_manager, database, transaction_options);
+            tx.commit().map(|u| result)
```
farost, line 45:
```suggestion
            tx.commit().map(|_| result)
```

system/util.rs
```diff
@@ -0,0 +1,212 @@
...
+pub mod answer_util {
+    use answer::variable_value::VariableValue;
+    use compiler::VariablePosition;
+    use executor::batch::Batch;
+    use lending_iterator::LendingIterator;
+    use std::collections::HashMap;
+    use database::transaction::TransactionRead;
+    use storage::durability_client::WALClient;
+
+    pub fn collect_answer(
```
farost, line 190: Really sounds like something possible to be outside of "utils".

user/BUILD
```diff
@@ -0,0 +1,37 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+load("@vaticle_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
+load("@rules_rust//rust:defs.bzl", "rust_library")
+package(default_visibility = ["//visibility:public"])
+
+load("@rules_rust//rust:defs.bzl", "rust_library", "rust_test")
```
farost, line 9: Also redundant

user/errors.rs
```diff
@@ -0,0 +1,13 @@
+use error::typedb_error;
+
+typedb_error!(
+    pub UserCreateError(component = "User create", prefix = "USC") {
+        Unexpected(1, "TODO: unexpected error"),
+    }
+);
+
+typedb_error!(
```
farost, line 9: `USD`? Check that there are no prefix collisions

user/errors.rs
```diff
@@ -0,0 +1,13 @@
+use error::typedb_error;
+
+typedb_error!(
+    pub UserCreateError(component = "User create", prefix = "USC") {
+        Unexpected(1, "TODO: unexpected error"),
```
farost, line 5: Let's populate more unique unexpected errors. Unexpected query result, unexpected network response, whatever

system/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+        relates password @card(1);
+";
+
+pub fn create_if_not_exists(database_manager: &DatabaseManager) -> Arc<Database<WALClient>> {
+    match database_manager.database(SYSTEM_DB) {
+        Some(db) => db,
+        None => {
+            database_manager.create_database(SYSTEM_DB).unwrap();
+            let db = database_manager.database(SYSTEM_DB).unwrap();
+            let tx_mgr = TransactionUtil::new(db.clone());
```
farost, line 44: Not sure if we use the `mgr` shortcut here. Makes things smaller, it's true. Not sure if we want to follow this path.

farost (approved):
Mostly stylistic and debugging comments, I trust that you'll address it. 
Run `rustfmt` and ask for @flyingsilverfin 's review, please, he has more ownership in this area

server/authenticator.rs
```diff
@@ -0,0 +1,42 @@
...
+                    }
+                    None => Err(Status::unauthenticated("invalid credential supplied"))
+                }
+            }
+            _ => {
+                Err(Status::unauthenticated("credential must be supplied"))
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
lolski: What's our convention for running `rustfmt`? Do we just invoke it directly or do we have a wrapper script that I can call instead?

system/BUILD
lolski: Fixed in 827697d0ab61affb6305c444d0d4147da2f310d4.

server/typedb.rs
```diff
@@ -5,38 +5,44 @@
...
-
-        Ok(Self { data_directory, typedb_service: Some(typedb_service), config })
+        let system_db = create_if_not_exists(&database_manager);
+        let user_manager = Arc::new(UserManager::new(system_db));
+        let typedb_service = TypeDBService::new(
+            &config.server.address, database_manager, user_manager.clone()
+        );
+        Ok(Self {
+            data_directory: storage_directory.to_owned(),
+            user_manager: user_manager.clone(),
```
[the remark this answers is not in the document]
lolski: Fixed in 827697d0ab61affb6305c444d0d4147da2f310d4.

system/lib.rs
lolski: Fixed in a31af758ecf7da05d190eb6684f9423b6f1e8441.

system/repositories.rs
lolski: Fixed in 8910c738519f0fbc7618e98c9dea3b9440935462.

system/repositories.rs
```diff
@@ -43,11 +43,9 @@ pub mod user_repository {
...
         let rows = result.unwrap();
-        let mut users: Vec<User> = vec![];
-        for row in rows {
-            let name  = get_string(&tx, &row, "n");
-            users.push(User::new(name))
-        }
+        let users = rows.iter().map(|row| {
+            User::new(get_string(&tx, &row, "n"))
+        }).collect();
         users
```
farost, line 49: Can be just
```
rows.iter().map(|row| User::new(get_string(&tx, &row, "n"))).collect()
```
(without `let`)

farost: (not going to push you to do so, it's just cleaner. If we want to highlight what this collection means, it should be fixed in the method's name)

server/authenticator.rs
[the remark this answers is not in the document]
lolski: Thanks! Fixed in 12cc62ffeacb542cb8e6316e14a0d05c41578431.

server/typedb.rs
```diff
@@ -46,12 +52,24 @@ impl Server {
...
         println!("Ready...");
+        let authenticator = Arc::new(Authenticator::new(self.user_manager.clone()));
         tonic::transport::Server::builder()
             .http2_keepalive_interval(Some(GRPC_CONNECTION_KEEPALIVE))
+            .layer(tonic::service::interceptor(move |req| authenticator.authenticate(req)))
             .add_service(service)
             .serve(self.config.server.address)
             .await
     }
+    fn create_storage_directory(storage_directory: &PathBuf) -> Result<(), ServerOpenError> {
```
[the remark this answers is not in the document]
lolski: Fixed in 12cc62ffeacb542cb8e6316e14a0d05c41578431.

system/util.rs
lolski: Fixed in 08f3441d77d54a2fc5d18b90a24b199c00f5aeac.

system/repositories.rs
lolski: Fixed in 08f3441d77d54a2fc5d18b90a24b199c00f5aeac.

system/util.rs
lolski: Currently it's only used by the system package, hence why I'm keeping it there. We can move it when it's needed elsewhere.

system/util.rs
lolski: Yeah agreed we could look how to generalise them. The reason I'm not using `with_schema_tx!` is because it's not quite generic and makes some assumptions about some test context.

user/BUILD
lolski: Fixed in 719a14ff1db2a1955e1c9107bf2addf970f4bd8e.

user/errors.rs
lolski: Just checked that `USC` is unique, there's no other error category that uses that three letters.

system/repositories.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub fn update(
+        snapshot: WriteSnapshot<WALClient>,
+        type_manager: &TypeManager,
+        thing_manager: Arc<ThingManager>,
+        function_manager: &FunctionManager,
+        username: &str,
+        user: &Option<User>,
+        credential: &Option<Credential>,
+    ) -> Arc<WriteSnapshot<WALClient>> {
+        if user.is_some() {
```
lolski, line 111: WIP

user/lib.rs
```diff
@@ -0,0 +1,19 @@
+use std::sync::Arc;
+
+use system::concepts::{Credential, PasswordHash, User};
+
+use crate::user_manager::UserManager;
+
+pub mod errors;
+pub mod user_manager;
+
+pub fn initialise_default_user(user_manager: &Arc<UserManager>) {
```
flyingsilverfin, line 10: You don't need `&Arc<>`, you can just use `&UserManager` here - I believe Arc<T> automatically is usable as `&T` (or sometimes you use `.as_ref()` when it gets confused I think)

user/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+    if !user_manager.contains("admin") {
```
flyingsilverfin, line 11: These constants should go in `constants.rs`

server/authenticator.rs
```diff
@@ -0,0 +1,40 @@
...
+                        Ok(req)
+                    } else {
+                        Err(Status::unauthenticated("invalid credential supplied"))
+                    }
+                }
+                None => Err(Status::unauthenticated("invalid credential supplied")),
+            },
+            _ => {
+                Ok(req)
+                // Err(Status::unauthenticated("credential must be supplied"))
```
flyingsilverfin, line 36: I think this should be the Err case right ("credential must be supplied") ? Otherwise, if you for isntance write your owner driver, which never sends the username/password, you just pass through here as if you're authenticated!

flyingsilverfin (changes requested):

system/repositories.rs
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        if credential.is_some() {
+            todo!("update credential detail")
+        }
+
+        todo!()
+    }
+
+    pub fn delete(
```
farost, line 122: What do we expect when we delete a user which is logged in? Change their password? Do we need to disconnect them with an error directly? Do we expect all the API calls to be verified with an auth error if something changed between the connection and this call?

server/authenticator.rs
```diff
@@ -0,0 +1,40 @@
...
+impl Authenticator {
+    pub(crate) fn new(user_manager: Arc<UserManager>) -> Self {
+        Self { user_manager }
+    }
+}
+
+impl Authenticator {
+    pub fn authenticate(&self, req: Request<()>) -> Result<Request<()>, Status> {
+        let metadata = req.metadata();
+        let username_metadata = metadata.get("username").map(|u| u.to_str());
```
farost, line 21: These should be constants, too

server/service/typedb_service.rs
```diff
@@ -156,6 +162,67 @@ impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
...
+        &self,
+        request: Request<typedb_protocol::user::update::Req>,
+    ) -> Result<Response<typedb_protocol::user::update::Res>, Status> {
+        users_update_req(request)
+            .and_then(|(username, user, cred)| self.user_manager.update(username.as_str(), &user, &cred))
+            .map(|_| Response::new(user_update_res()))
+            .map_err(|err| err.into_error_message().into_status())
+    }
+
+    async fn users_delete(
```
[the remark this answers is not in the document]
farost: Where is/will be the validation of your rights to delete a user, of the users available for deletion?

system/repositories.rs
lolski: Fixed in 0728efbac92ba3eb2dbcf32e2fbd6db45f827a46.

user/errors.rs
lolski: Aah I get what you meant now. Fixed in 92cda1e30049f05fbbf2a59996039a7aac04012a.

system/repositories.rs
lolski: Fixed in 12cc62ffeacb542cb8e6316e14a0d05c41578431.

server/service/typedb_service.rs
```diff
@@ -35,12 +41,12 @@ use crate::service::{
 pub(crate) struct TypeDBService {
     address: SocketAddr,
     database_manager: Arc<DatabaseManager>,
-    // map of connection ID to ConnectionService
+    user_manager: Arc<UserManager>, // map of connection ID to ConnectionService
```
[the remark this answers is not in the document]
lolski: Fixed in f77b03b9319a7cd392bb8a8fe27504654f466645.

user/errors.rs
```diff
@@ -0,0 +1,19 @@
+use error::typedb_error;
+
+typedb_error!(
+    pub UserCreateError(component = "User create", prefix = "USC") {
```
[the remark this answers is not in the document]
lolski: Fixed in 92cda1e30049f05fbbf2a59996039a7aac04012a

user/lib.rs
lolski: "Extract into method" feature in IDEA is not as good as expected turns out. Fixed in 16b0608fdba6a3ab35f078650ff0be39c34869d2.

user/lib.rs
lolski: Fixed in b9c8e80f8260a7b32a597e6374418e00cdf7d5d7

server/authenticator.rs
lolski: Fixed in https://github.com/typedb/typedb/commit/b9c8e80f8260a7b32a597e6374418e00cdf7d5d7

server/service/typedb_service.rs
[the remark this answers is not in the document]
lolski: > Can we delete any user?
Good catch. I've added protection against deleting the default user here: a710dc0564f59ba97e860d96d25b42cb2f1df12e. In the future we'll aim for a better solution since this solution isn't very flexible / user friendly.

> Where is/will be the validation of your rights to delete a user, of the users available for deletion?
This is when we implement roles

server/authenticator.rs
lolski: Fixed in fdb755434435739522d94dfda7bb07647a304544.

system/repositories.rs
lolski: When a user is deleted, subsequent connections made by them will be rejected. However, an already running transaction will stay open until it is closed. Given that transaction has a timeout, the user will not be able to maintain it indefinitely. I think this solution should be good enough for now, right @flyingsilverfin ?

farost: What about user's operations? Can they query, commit? All the permissions should be updated immediately. Otherwise we are not really ready production user management

system/repositories.rs
```diff
@@ -0,0 +1,143 @@
+// TODO: read from a file at compile time
+pub const SCHEMA: &str = "
```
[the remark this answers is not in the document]
lolski: Nice! Fixed in 2f523a16acbf2b1be69f37aa5a0b1d4c77f006c1

flyingsilverfin (approved):

system/repositories.rs
lolski: I agree with this, I think connections associated with a user that is either deleted, or has their credential updated should immediately be closed. However, in order to do that we need to keep track of which connections belong to which user, which is something the server  currently doesn't do.

Therefore, as have been [discussed](https://discord.com/channels/665254494820368395/672580553677078548/1311046300401401896) with @flyingsilverfin over Discord, we agreed that we'll postpone this issue as performance optimisation takes higher precedence.

PR #7269, opened 2024-11-27
Optimisations from typeql

krishnangovindraj:
Updates typeql for parsing optimisations.

## Implementation
* Updates typeql for parsing optimisations.
* Implements checks for ensuring reserved keywords aren't used as identifiers.

### DIFF
--- a/ir/pipeline/mod.rs
+++ b/ir/pipeline/mod.rs
@@ -104,6 +104,11 @@ typedb_error! {
             signature: Signature,
             return_: ReturnStatement
         ),
+        ReservedKeywordAsIdentifier(
+            9,
+            "A reserved keyword \"{identifier}\" was used as function name",
+            identifier: String
+        ),
     }
 }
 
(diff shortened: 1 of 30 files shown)


ir/pipeline/mod.rs
```diff
@@ -104,6 +104,11 @@ typedb_error! {
             signature: Signature,
             return_: ReturnStatement
         ),
+        ReservedKeywordAsIdentifier(
+            9,
+            "A reserved keyword \"{identifier}\" was used as function name",
```
flyingsilverfin, line 109: Maybe we make it more action oriented: `the reserved keyword ... cannot be used as a ... ` in all the error messages

flyingsilverfin (approved):

ir/pipeline/mod.rs
krishnangovindraj: Do we want to have a variant per usage (as function name , as type label, etc.) or is just one enough?
The identifier does contain the span so I could return an offset int the string, if we had the string.

krishnangovindraj (commented):
replies

PR #7272, opened 2024-11-28
Validate user input to the 'user repository' module

lolski:
Validate input to 'user repository'  methods by restricting the value such that it must be a valid TypeQL identifier.

## Motivation
The validation is done mainly to prevent TypeQL injection into the system database by a malicious attacker.

## Implementation
- Validate the supplied username into the various methods in the `user_repository` module , make sure it is a valid TypeQL identifier

### DIFF
--- a/system/repositories.rs
+++ b/system/repositories.rs
@@ -37,9 +37,9 @@ pub mod user_repository {
         users
     }
 
-    pub fn get(tx: TransactionRead<WALClient>, username: &str) -> Result<Option<(User, Credential)>, QueryError> {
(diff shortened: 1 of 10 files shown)


system/repositories.rs
```diff
@@ -132,6 +142,16 @@ pub mod user_repository {
...
-        snapshot
+        (Ok(()), snapshot)
+    }
+
+    pub fn is_valid_typeql_value(value: &str) -> bool {
+        is_valid_identifier(value)
     }
+
+    typedb_error!(
+        pub QueryError(component = "System database", prefix = "SDB") {
```
flyingsilverfin, line 153: Ooops you need a new name here! `SystemError`?

flyingsilverfin (approved):

PR #7273, opened 2024-11-28
Hide the system database

lolski:
Prevent user access to any internal databases (ie., any database whose name is prefixed with `_`) through the database manager.

## Motivation
Add the ability for the server to create databases that is not accessible by user for internal purposes, such as storing user accounts and possibly other internal configurations

## Implementation
- Prevent access to internal databases through the database manager

### DIFF
--- a/system/lib.rs
+++ b/system/lib.rs
@@ -13,21 +13,25 @@ use std::{fmt::format, sync::Arc};
 use database::{database_manager::DatabaseManager, Database};
 use storage::durability_client::WALClient;
 use typeql;
-
+use database::database_manager::INTERNAL_DATABASE_PREFIX;
 use crate::{repositories::SCHEMA, util::transaction_util::TransactionUtil};
 
-const SYSTEM_DB: &str = "system";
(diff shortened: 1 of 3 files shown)


system/lib.rs
```diff
@@ -13,42 +13,46 @@ use std::{fmt::format, sync::Arc};
...
 use storage::durability_client::WALClient;
 use typeql;
-
+use database::database_manager::INTERNAL_DATABASE_PREFIX;
 use crate::{repositories::SCHEMA, util::transaction_util::TransactionUtil};
 
-const SYSTEM_DB: &str = "system";
+fn get_system_database_name() -> String {
+    format!("{}system", INTERNAL_DATABASE_PREFIX)
+}
```
flyingsilverfin, line 21: Why don't you keep this as as `&'static str`? You can use `concat!` macro to create this string statically: https://doc.rust-lang.org/std/macro.concat.html

flyingsilverfin (approved):

flyingsilverfin (approved):

PR #7274, opened 2024-11-29
Bytes

dmitrii-ubskii:


compiler/executable/match_/planner/plan.rs
dmitrii-ubskii, line 1: Drive-by: deterministic greedy planner based on heuristics

answer/lib.rs
```diff
@@ -81,62 +81,62 @@ impl Type {
...
             Type::Attribute(_) => panic!("Attribute Type is not an object type"),
             Type::RoleType(_) => panic!("Role Type is not an Object type"),
         }
     }
 
-    pub fn as_entity_type(&self) -> EntityType<'static> {
+    pub fn as_entity_type(&self) -> EntityType {
         match self {
-            Type::Entity(entity) => entity.clone().into_owned(),
+            Type::Entity(entity) => (*entity).into_owned(),
```
dmitrii-ubskii, line 95: Some of these `into_owned()` calls are no-ops. I've cleaned up a few, but there are simply too many.

encoding/graph/thing/edge.rs
```diff
@@ -614,100 +592,76 @@ impl<'a> ThingEdgeRolePlayerIndex<'a> {
...
-        from: ObjectVertex<'_>,
-    ) -> StorageKey<'static, { ThingEdgeRolePlayerIndex::LENGTH_PREFIX_FROM }> {
+        player_from: ObjectVertex,
+        player_to: ObjectVertex,
+        relation: ObjectVertex,
+        role_id_from: TypeID,
+        role_id_to: TypeID,
+    ) -> Self {
+        Self { player_from, player_to, relation, role_id_from, role_id_to }
+    }
```
flyingsilverfin, line 613: This may or may not be a lot of work, though is just renaming - we should flip `new` and `build` now, since `new()` should now be the one that just wraps the struct for the input, while `build()` should be the one that extracts/constructs complex things before returning the struct

flyingsilverfin (changes requested):
See 2 comments

PR #7280, opened 2024-12-03
Partition data into Keyspaces and configure RocksDB

flyingsilverfin:

We split the data in TypeDB into 5 RocksDB database (internally known as Keyspaces):

```
    /// Keyspace optimised for 11 byte prefix seeks:
    /// mostly Short Attribute Vertices that are Put (12 bytes)
    /// also existence/IID checks for Object vertices (11 bytes)
    /// also schema (default KS)
    /// Ordered Object Properties prefix: [1: prefix][11: object][1: ordered property] (13 bytes)
    DefaultOptimisedPrefix11,

    /// Keyspace optimised for 15 byte prefix seeks:
    /// Links & Links Reverse prefix:  [1: prefix][11: from][3: to type]
    /// Has prefix:  [1: prefix][11: from][3: to type]
    OptimisedPrefix15,

    /// Keyspace optimised for 16 byte prefix seeks:
    /// Has Reverse prefix for Short attribute vertices: [1: prefix][12: from][3: to type]
    OptimisedPrefix16,

    /// Keyspace optimised for 17 byte prefix seeks:
    /// Long Attribute Vertices existence checks have 21 bytes [1: prefix][2: type][18: ID + category], but could still benefit from a 17 byte bloom prefix
    /// LinksIndex prefix: [1: prefix][11: player 1][2: rel type id][3: player 2 type]
    OptimisedPrefix17,


    /// Keyspace optimised for 25 byte prefix seeks:
    /// Has Reverse prefix for Long attribute vertices: [1: prefix][21: from][3: to type]
    OptimisedPrefix25,
 ```

In addition, we configure RocksDB caches, index and filter block pinning, bloom filters, compression at various levels, and prefix extractors.

We also try to squeeze a bit of performance by enable Link-Time optimisation and producing more optimal output binaries using the following cargo release profile:

```
[profile.release]
    codegen-units = 1
    lto = "fat"
    panic = "abort"
```

**Warnings**
1)  We configure each database to have a default **1GB block cache for the time being**.
2) We have **not** seen the bloom filter actually produce any hits yet (according to RocksDB statistics), and need to investigate further
3) As noted in https://github.com/typedb/typedb/pull/7262, if we get bloom filters to activate, we will also have to careful to _deactivate_ them when the scan range escapes a single constant bloom-filtered prefix.

## Motivation

We aim to settle into our final storage keyspace layout, and optimise RocksDB for optimal performance.

## Implementation

- Use `EncodingKeyspace` to generate RocksDB configuration objects, which are used at Keyspace creation time to populate the keyspace's database
- Pass `keyspace` in a easy but architecturally suboptimal manner into places where required, partially motivated by the desire to split Attribute vertices into 2 keyspaces: Short and Long, to allow using a longer bloom filter prefix for Long attributes. In particular, this should be beneficial when generating hashed String ID's, which require searching for a slot to write the string to and use a longer prefix lookup.

### CODE REVIEW

File: server/service/transaction_service.rs [hunk 1]
```diff
@@ -425,20 +425,29 @@ impl TransactionService {
 
         let transaction = match transaction_type {
             typedb_protocol::transaction::Type::Read => {
-                Transaction::Read(TransactionRead::open(database, transaction_options).map_err(|typedb_source| {
-                    TransactionServiceError::TransactionFailed { typedb_source }.into_error_message().into_status()
-                })?)
+                let transaction = spawn_blocking(move ||
```
flyingsilverfin, line 428: We sometimes have to wait for the watermark, so better to run this in spawn_blocking

storage/keyspace/keyspace.rs
```diff
@@ -39,15 +40,11 @@ pub trait KeyspaceSet: Copy {
...
-    options.create_if_missing(true);
-    options.create_missing_column_families(true);
-    // options.enable_statistics();
-    // TODO optimise per-keyspace
-    options
+    fn rocks_configuration(&self, _cache: &rocksdb::Cache) -> rocksdb::Options {
+        let mut options = Options::default();
+        options.create_if_missing(true);
+        options
+    }
```
flyingsilverfin, line 47: Default impl, useful for tests/mocks

storage/keyspace/keyspace.rs
```diff
@@ -196,28 +195,10 @@ impl Keyspace {
         Self { path, name: keyspace.name(), id: keyspace.id(), kv_storage, read_options, write_options }
     }
 
-    // TODO: we want to be able to pass new options, since Rocks can handle rebooting with new options
-    pub(crate) fn load_from_checkpoint(_path: PathBuf) {
```
flyingsilverfin, line 200: This code path is not used to load from a checkpoint

PR #7283, opened 2024-12-05
Relation index

flyingsilverfin:

We implement the application of the relation index, which is an optimised storage-layer index that allows skipping through a relation from one player to another player, avoiding one lookup operation and giving the query planner the choice to produce different sorted intersection points. This change could improve performance on traversals across relations by up to 50%.

## Motivation

We aim to optimise a common case of read queries, which are for small, pairwise relation queries: 
```
match 
$x isa person;
$y isa dog;
$r isa pet-ownership, links (owner: $y, pet: $y);
```

Rather than traversing this as `$x -links-> $r -links=> $y`, using two storage-layer operations, we can now implement this as: `$x -relation_index-> ($y, $r)`, producing the relations and pets at the same time.

## Implementation

We create a new package: `//compiler/transformation`, which is intended to be similar to LLVM's optimisation passes: each transformation will operate over a type-annotated IR structure, and produce an equivalent but structurally different query. We use this architecture to implement the `relation_index` transformation.

The relation index transformation is applied when a binary player-to-player relation query is found: `$r link ($x); $r links ($y);`. One case where the shape is found but the index is not applied is when the relation `$r` has an equality or IID identifier, which can be used instead fo generate a sorted iterator of `$r` relations, and intersected with iterators from the two role players - in general a more optimal plan. However there are casese between these two where the planner should decide whether to use the index or the `links` traversal operations**.

The storage layout of the relation index is as follows:
```
[relation index byte][relation type][start player (type + id)][end player (type + id)][relation id][start player role type][end player role type]
```

This is designed to cater to a couple of common cases:
1. There is a single, or very few, relation types we are traversing across. Each of which will require their own storage lookup, so this index will work best with smaller number of applicable relation types.
2. Wish to either generate the start player (unbound case) or the end player (bound case) in sorted order - or even the relation (start and end player are boundd)

Note that this scheme cannot support using a known relation ID as part of the prefix to cut down the search space. This is an example of where the relation index may be less optimal than using the `links` traversals.

Lastly, we introduce the `RelationIndexExecutor` and all the required downstream changes in `ThingManager` and `Object`.

### Minor Changes

- Fix bug in the usage of the QueryProfiler in the Sort stage
- Fix a bug that hasn't appeared yet, when using a Links/LinksReverse executor where the RoleType is already bound. We now additionally check that the RoleType that is bound is equal to the one found in the iterator, otherwise we eliminate it
- Simplify the `TupleIterator` variants, recognising that many of the types it contains are actually the same

## Extensions

** We plan to extend the planner's model to allow for specifiying choices: either use (links, links) operations, or use a (relation index) operation, depending on what is best according to the statistics and the query provided. This PR does not implement this choice system.


executor/instruction/links_executor.rs
```diff
@@ -162,13 +167,27 @@ impl LinksExecutor {
...
+        let row_role = if self.variable_modes.get(role_var).is_some_and(|mode| matches!(VariableMode::Input, mode)) {
+            Some(row.get(role_var.as_position().unwrap()).as_type().as_role_type().clone())
+        } else {
+            None
+        };
+
+        let filter_for_row: Box<LinksFilterMapFn> = Box::new(move |item|{
+            match filter(&item) {
+                Ok(true) => match check(&item) {
+                    Ok(true) => match verify_role(&item, row_role) {
```
flyingsilverfin, line 181: We need to check if the role == existing role, if there is an existing role. I added a test for this case in the IndexExecutor case.

PR #7288, opened 2024-12-12
Introduce 3.0 server diagnostics

farost:
### Diagnostics
We introduce a diagnostics package to collect metrics related to server usage. There are multiple potential clients of this package, a couple of them being initially implemented:
* A web endpoint for pulling data (e.g. Prometheus) is bound at port 4104 by default and exposes diagnostics formatted for prometheus (`http://localhost:4104/metrics?format=prometheus`) or as JSON (`http://localhost:4104/metrics?format=JSON`) (done);
* A push http client for sending diagnostics data to the TypeDB Diagnostics Service (done);
* PostHog (will be done in a near future).

Example monitored JSON diagnostics (the reported diagnostics are in a similar JSON format, but with a couple of changed fields) (simplified):
```
{
  "version": 1,
  "deploymentID": "IXP50DFI2J7RTQNA",
  "serverID": "IXP50DFI2J7RTQNA",
  "distribution": "TypeDB CE",
  "timestamp": "2024-12-16T18:22:48.335",
  "server": {
    "version": "3.0.0-alpha-10",
    "uptimeInSeconds": 27,
    "os": {
      "name": "Darwin",
      "arch": "arm64",
      "version": "14.2.1"
    },
    "memoryUsedInBytes": 38149668864,
    "memoryAvailableInBytes": 30569807872,
    "diskUsedInBytes": 426276663032,
    "diskAvailableInBytes": 1568941502728
  },
  "load": [
    {
      "database": "95615005977010737",
      "schema": {
        "typeCount": 8
      },
      "data": {
        "entityCount": 2,
        "relationCount": 1,
        "attributeCount": 3,
        "hasCount": 3,
        "roleCount": 2,
        "storageInBytes": 8107,
        "storageKeyCount": 71
      }
    }
  ],
  "actions": [
    {
      "name": "AUTHENTICATE",
      "attempted": 1094,
      "successful": 1094
    },
    {
      "name": "DATABASES_CREATE",
      "database": "3244421341483603138",
      "attempted": 1,
      "successful": 0
    },
    {
      "name": "TRANSACTION_QUERY",
      "database": "95615005977010737",
      "attempted": 274,
      "successful": 270
    },
    {
      "name": "DATABASES_CREATE",
      "database": "95615005977010737",
      "attempted": 88,
      "successful": 88
    },
    {
      "name": "TRANSACTION_OPEN",
      "database": "95615005977010737",
      "attempted": 153,
      "successful": 153
    },
    {
      "name": "TRANSACTION_COMMIT",
      "database": "95615005977010737",
      "attempted": 68,
      "successful": 66
    }
  ],
  "errors": [
    {
      "code": "DBC1",
      "database": "3244421341483603138",
      "count": 1
    },
    {
      "code": "TSV2",
      "database": "95615005977010737",
      "count": 1
    },
    {
      "code": "SYR1",
      "database": "95615005977010737",
      "count": 3
    }
  ]
}
```

Example of monitored Prometheus diagnostics (simplified, some `actions` from the previous example can be missed):
```
# distribution: TypeDB CE
# version: 3.0.0-alpha-10
# os: Darwin arm64 14.2.1

# TYPE server_resources_count gauge
server_resources_count{kind="memoryUsedInBytes"} 38218219520
server_resources_count{kind="memoryAvailableInBytes"} 30501257216
server_resources_count{kind="diskUsedInBytes"} 426276581112
server_resources_count{kind="diskAvailableInBytes"} 1568941584648

# TYPE typedb_schema_data_count gauge
typedb_schema_data_count{database="95615005977010737", kind="typeCount"} 8
typedb_schema_data_count{database="95615005977010737", kind="entityCount"} 2
typedb_schema_data_count{database="95615005977010737", kind="relationCount"} 1
typedb_schema_data_count{database="95615005977010737", kind="attributeCount"} 3
typedb_schema_data_count{database="95615005977010737", kind="hasCount"} 3
typedb_schema_data_count{database="95615005977010737", kind="roleCount"} 2
typedb_schema_data_count{database="95615005977010737", kind="storageInBytes"} 8107
typedb_schema_data_count{database="95615005977010737", kind="storageKeyCount"} 71

# TYPE typedb_attempted_requests_total counter
typedb_attempted_requests_total{kind="CONNECTION_OPEN"} 276
typedb_attempted_requests_total{kind="AUTHENTICATE"} 1094
typedb_attempted_requests_total{kind="USERS_ALL"} 88
typedb_attempted_requests_total{kind="DATABASES_ALL"} 295
typedb_attempted_requests_total{database="95615005977010737", kind="TRANSACTION_QUERY"} 274
typedb_attempted_requests_total{database="95615005977010737", kind="DATABASES_CREATE"} 88
typedb_attempted_requests_total{database="95615005977010737", kind="TRANSACTION_OPEN"} 153
typedb_attempted_requests_total{database="95615005977010737", kind="TRANSACTION_COMMIT"} 68
typedb_attempted_requests_total{database="95615005977010737", kind="DATABASES_DELETE"} 89

# TYPE typedb_successful_requests_total counter
typedb_successful_requests_total{kind="CONNECTION_OPEN"} 276
typedb_successful_requests_total{kind="AUTHENTICATE"} 1094
typedb_successful_requests_total{database="95615005977010737", kind="TRANSACTION_QUERY"} 270
typedb_successful_requests_total{database="95615005977010737", kind="DATABASES_CREATE"} 88
typedb_successful_requests_total{database="95615005977010737", kind="TRANSACTION_OPEN"} 153
typedb_successful_requests_total{database="95615005977010737", kind="TRANSACTION_COMMIT"} 66
typedb_successful_requests_total{database="95615005977010737", kind="DATABASES_DELETE"} 89

# TYPE typedb_error_total counter
typedb_error_total{database="3244421341483603138", code="DBC1"} 1 
typedb_error_total{database="95615005977010737", code="TSV2"} 1 
typedb_error_total{database="95615005977010737", code="SYR1"} 3
``` 

If the reporting is turned off, we report a single batch of minimal diagnostics information after 1 hour of runtime just to mark that a server has been booted up. No user information is shared.
If the reporting is turned on, the reporting happens every hour following [the rules set in 2.x](https://github.com/typedb/typedb/pull/7059).
If the development mode is active (details can be found below), no reporting is executed, and the reporting config flag is ignored.

### Configuration
We add new CLI flags and for configuration:
1. `--diagnostics.monitoring.port 4014` for configuring the monitoring server's port;
2. `--diagnostics.monitoring.enable false` for disabling and enabling the monitoring server (true by default, thus called `enable` and requires a boolean value);
3. `--diagnostics.reporting.metrics false` for disabling and enabling the collected load and usage metrics' reporting (true by default, ignored in the development mode).

An additional switch for the development mode has also been added in order to use TypeDB's release binaries in CI.   

### Development mode
We reintroduce the development mode. As in 2.x, this mode is a default mode for local bazel (& cargo) builds and snapshots. It's turned off in published releases but can be explicitly switched on through CLI, as mentioned in the `Configuration` section.
**WARNING**: the development mode is used by the TypeDB's developers and is not intended to be used in other programs. Be aware that its usage can affect the server's performance and stability. 

## Implementation
Check PR comments.


server/service/transaction_service.rs
```diff
@@ -397,139 +404,161 @@ impl TransactionService {
...
-
-            // service options
-            self.prefetch_size = options.prefetch_size.or(Some(DEFAULT_PREFETCH_SIZE));
-            self.transaction_timeout_millis =
-                options.transaction_timeout_millis.or(Some(DEFAULT_TRANSACTION_TIMEOUT_MILLIS));
-        }
-
-        let transaction_type = typedb_protocol::transaction::Type::try_from(open_req.r#type)
-            .map_err(|_| ProtocolError::UnrecognisedTransactionType { enum_variant: open_req.r#type }.into_status())?;
+        run_with_diagnostics_async(self.diagnostics_manager.clone(), ActionKind::TransactionOpen, || async {
```
farost, line 407: I wrap every network call-based logic in `run_with_diagnostics`(`_async`). Inside this function, we save information about actions (success/fail), and also extract the error codes when possible.

diagnostics/reporter.rs
```diff
@@ -0,0 +1,154 @@
...
+        let _ = Arc::get_mut(&mut Arc::clone(&self._reporting_job)).map(|opt| *opt = Some(reporting_job));
+    }
+
+    fn report_once_if_needed(&self) {
+        let disabled_reporting_file = self.data_directory.join(DISABLED_REPORTING_FILE_NAME);
+        if !disabled_reporting_file.exists() {
+            let diagnostics = self.diagnostics.clone();
+            let reporting_uri = self.reporting_uri;
+            let data_directory = self.data_directory.clone();
+            thread::spawn(move || {
```
farost, line 85: Here, I spawn an std::thread.

diagnostics/reporter.rs
```diff
@@ -0,0 +1,154 @@
...
+        Arc, Mutex,
+    },
+    thread,
+    time::{Duration, SystemTime, UNIX_EPOCH},
+};
+
+use chrono::{DateTime, Local, Timelike};
+use concurrency::IntervalRunner;
+use logger::{debug, trace};
+use reqwest::{
```
farost, line 25: A use the [reqwest](https://crates.io/crates/reqwest) crate which seems to be one of the most popular and mature crates for creating REST clients.

diagnostics/reporter.rs
```diff
@@ -0,0 +1,154 @@
...
+        } else {
+            self.report_once_if_needed();
+        }
+    }
+
+    fn schedule_reporting(&self) {
+        let diagnostics = self.diagnostics.clone();
+        let reporting_uri = self.reporting_uri;
+
+        let reporting_job = IntervalRunner::new_with_initial_delay(
```
farost, line 69: Right now, I use an IntervalRunner, which uses an `std::thread::spawn` under the hood.

diagnostics/monitoring_server.rs
```diff
@@ -0,0 +1,90 @@
...
+
+use std::{
+    convert::Infallible,
+    net::SocketAddr,
+    sync::{Arc, Mutex},
+    thread,
+};
+
+use futures::executor::block_on;
+use hyper::{
```
farost, line 15: A use the [hyper](https://crates.io/crates/hyper) crate which seems to be one of the most popular and mature crates for creating REST servers (I actually haven't tried it as a client crate for `reporter`, may pay attention to it if needed).
This crate requires using a `tokio` runtime.

diagnostics/monitoring_server.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    pub fn may_start(&self) {
+        if !self.is_enabled {
+            return;
+        }
+
+        let addr = SocketAddr::from(([0, 0, 0, 0], self.port));
+        let diagnostics = self.diagnostics.clone();
+
+        task::spawn(async move {
```
farost, line 46: A `tokio` task for the created `hyper` service.

diagnostics/diagnostics_manager.rs
```diff
@@ -0,0 +1,217 @@
...
+        deployment_id: String,
+        server_id: String,
+        distribution: String,
+        version: String,
+        data_directory: PathBuf,
+        monitoring_port: u16,
+        is_monitoring_enabled: bool,
+        is_reporting_enabled: bool,
+    ) -> Self {
+        let diagnostics = Arc::new(Diagnostics::new(
```
farost, line 61: `DiagnosticsManager` holds an `Arc` wrapping the only instance of `Diagnostics`. This `Arc` is shared between it, `MonitoringServer`, and `Reporter`.

diagnostics/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+pub mod metrics;
+mod monitoring_server;
+mod reporter;
+mod version;
+
+type DatabaseHash = u64;
+type DatabaseHashOpt = Option<u64>;
+
+#[derive(Debug)]
+pub struct Diagnostics {
```
farost, line 34: `Diagnostics` consists of a number of `Properties` and `Metrics`. The model of these metrics is a replication of what we had in 2.x, but with slightly different names. 

The synchronization is placed here at the moment. 
We are still not fully interested in preserving the balance between `load_metrics` and `action_metrics`, so it's fine if they are updated and cleared on snapshotting in parallel, these losses are minimal and not critical. 

However, for simplicity, I placed the mutexes on the whole metrics now, which means that two transactions against two different databases will share a single mutex when they report their actions and errors. It's not ideal, but I decided to start with this approach, leaving the `***Metrics` structs unaware of the cruel concurrent world. This might be a performance issue!

diagnostics/diagnostics_manager.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    diagnostics_method! {
+        pub fn submit_database_metrics(&self, database_metrics: HashSet<DatabaseMetrics>);
+        pub fn submit_error(&self, database_name: Option<&str>, error_code: String);
+        pub fn submit_action_success(&self, database_name: Option<&str>, action_kind: ActionKind);
+        pub fn submit_action_fail(&self, database_name: Option<&str>, action_kind: ActionKind);
+        pub fn increment_load_count(&self, database_name: &str, connection_: LoadKind);
+        pub fn decrement_load_count(&self, database_name: &str, connection_: LoadKind);
+    }
```
dmitrii-ubskii, line 83: Why are both `DiagnosticsManager` and `Diagnostics` responsible for this? As far as I can tell, these are only called from a handful of places.

`DiagnosticsResultTracker`, e.g., doesn't need anything manager-specific, and should just have a reference to `Diagnostics` instead. Same goes for `synchronize_database_metrics()`. Then this macro can go away.

diagnostics/diagnostics_manager.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn drop(&mut self) {
+        match &self.error_code {
+            Some(error_code) => {
+                self.diagnostics_manager.submit_action_fail(self.database_name, self.action_kind);
+                self.diagnostics_manager.submit_error(self.database_name, error_code.clone());
+            }
+            None => self.diagnostics_manager.submit_action_success(self.database_name, self.action_kind),
+        }
+    }
+}
```
dmitrii-ubskii: An interesting approach.

diagnostics/diagnostics_manager.rs
```diff
@@ -0,0 +1,217 @@
...
+
+fn run_and_track<F, T>(mut tracker: DiagnosticsResultTracker<'_>, f: F) -> Result<T, Status>
+where
+    F: FnOnce() -> Result<T, Status>,
+{
+    let result = f();
+    if let Err(ref status) = result {
+        tracker.set_error_code(get_status_error_code(status));
+    }
```
dmitrii-ubskii, line 180: I'm not convinced this needs the tracker. `ActionKind` and `Option<Database>` can be arguments to `run_and_track`, and the `drop` glue can be inlined here.

diagnostics/diagnostics_manager.rs
```diff
@@ -0,0 +1,217 @@
...
+    result
+}
+
+fn get_status_error_code(status: &Status) -> String {
+    if let Ok(details) = status.check_error_details() {
+        if let Some(error_info) = details.error_info() {
+            return error_info.reason.clone();
+        }
+    }
+    "UKNOWN".to_owned() // TODO: What to do with error codes?
```
dmitrii-ubskii, line 230: We should at least use `Status::code()` I think. Although I don't think we should ever be in that situation?

diagnostics/diagnostics_manager.rs
dmitrii-ubskii: This ctor should take a `Diagnostics` value instead of passing a bunch of parameters through to it.

diagnostics/reporter.rs
```diff
@@ -0,0 +1,154 @@
...
+        } else {
+            (scheduled_minute - current_minute) * 60
+        };
+
+        Duration::from_secs(delay_secs)
+    }
+
+    fn report(diagnostics: Arc<Diagnostics>, reporting_uri: &'static str) -> bool {
+        let diagnostics_json = diagnostics.to_reporting_json().to_string();
+        diagnostics.take_snapshot();
```
dmitrii-ubskii, line 115: Are these the wrong way around? Shouldn't we take a snapshot and report that snapshot?

diagnostics/monitoring_server.rs
dmitrii-ubskii: A little weird, but it's probably worth defensively marking `may_start` as `async` since it requires being run in a `tokio` runtime. That will force `Server::open` to be `async` as well.

server/service/transaction_service.rs
dmitrii-ubskii: These huge closures are a bit awkward. Can we keep the original methods and call them in `run_with_diagnostics_async` instead?

dmitrii-ubskii: That's less relevant for the main typedb service.

dmitrii-ubskii (changes requested):

server/service/transaction_service.rs
farost: Both methods will be private, so it will be quite easy to accidentally use the one without diagnostics. That's why I decided to just make a closure. Do you have a naming idea of how to make the intended usage clear?

diagnostics/monitoring_server.rs
farost: Which means that I'll move task::spawn for the monitoring start to `Server::open`, which is now:
```
        diagnostics_manager.may_start_monitoring();
        diagnostics_manager.may_start_reporting();
```
And we'll spawn it even if monitoring is not needed (if it returns after `!self.is_enabled`). Is it fine?

diagnostics/diagnostics_manager.rs
farost: From my point of view, it manages `Diagnostics` by itself, and it is responsible for creating `Diagnostics`. It's cleaner to avoid passing the arguments into multiple places (which can also lead to mistakes), but I'm not really convinced. 
Just like the `DatabaseManager`, which is responsible for **creating**, updating, deleting, and viewing databases. 

But I'm okay to give up if we believe that the argument passing is a strong argument, or you have something else to mention.

common/logger/logger.rs
```diff
@@ -7,7 +7,7 @@
 #![allow(unexpected_cfgs)]
 
 use tracing::{self, dispatcher::DefaultGuard, metadata::LevelFilter, Level};
-pub use tracing::{error, info, trace};
+pub use tracing::{debug, error, info, trace};
```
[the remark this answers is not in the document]
farost: Used, we need to declare it here so we can access these methods through `use logger::{debug, trace};`

diagnostics/diagnostics_manager.rs
farost: It was simpler for me to think about resource sharing in concurrency this way. `Diagnostics` are initialized by the `DiagnsoticsManager` that also controls all the `MonitoringServer`s and `Reporter`s, so it can share the ownership once on the setup. Then, if you need to work with `Diagnostics`, you need to do it through the `DiagnosticsManager`, which can, if needed, control the concurrency issues if needed. 

It's not used now as this control is moved to `Diagnostics`, and I don't have the plan to move it higher. 

In case of actions and errors, right now it can really be removed from the manager as the `ResultTracker` covers these needs. However, I still need `submit_database_metrics`, `increment_load_count`, and `decrement_load_count` which are used against the manager, and I'd like to encapsulate it instead of exposing the `Diagnostics` out of the `Manager`.

Is it worth it?

diagnostics/diagnostics_manager.rs
farost: In a positive, neutral, or negative way? =)

diagnostics/diagnostics_manager.rs
farost: Okay, I added replaced it. Wasn't sure if it's correct.

diagnostics/diagnostics_manager.rs
farost: Well, you're right, it ended up being quite simple. Cool, I'll remove the tracker.

diagnostics/diagnostics_manager.rs
farost: UPD. As I am going to remove the tracker, I think that it is worth preserving the intermediate encapsulating calls of `DiagnosticsManager`

diagnostics/diagnostics_manager.rs
dmitrii-ubskii: > Just like the `DatabaseManager`, which is responsible for creating, updating, deleting, and viewing databases.
But when we're interacting with a database, that's still done via a `Database` value, access to which is provided through `DatabaseManager::database`, so it's not as if it's invisible.

My main argument here is that it will structure the arguments rather than having just a list of strings and numbers.

common/logger/logger.rs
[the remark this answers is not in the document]
dmitrii-ubskii: Oh it's a re-export. Understood.

server/service/transaction_service.rs
farost: Fine, just for `TransactionService`. Thanks

diagnostics/monitoring_server.rs
farost: But `Server::open` is not async, how can I `.await` there without a runtime?

diagnostics/diagnostics_manager.rs
dmitrii-ubskii: I was originally thinking of exposing the diagnostics field through a method and using that when needed. There's no danger in exposing it, and it's no less clear at the call site (arguably, more clear that `submit_` methods only affect the diagnostics registry and not any services). In any case, the encapsulation is kinda meaningless here since all the same behaviour is available on the manager anyway, and no extra checking is done.

But now I'm thinking, do we even need a `DiagnosticsManager`? It's not managing anything, it just groups the `Diagnostics` with a couple of independent services.

diagnostics/diagnostics_manager.rs
farost: Fine, I'll `Diagnostics` outside.

diagnostics/diagnostics_manager.rs
dmitrii-ubskii: I was undecided until I saw where it was used :P

diagnostics/reporter.rs
dmitrii-ubskii: I think I'm still confused as to what `take_snapshot` is actually doing. It could use a better name.

Is it more like `reset` if the deltas are calculated against the snapshot, so effectively the deltas are reset to 0?

diagnostics/diagnostics_manager.rs
farost: It doesn't really help, actually. I either need to extract data from `diagnostics` for two params or:

```
impl DiagnosticsManager {
    pub fn new(
        deployment_id: String,
        diagnostics: Diagnostics,
        data_directory: PathBuf,
        monitoring_port: u16,
        is_monitoring_enabled: bool,
        is_reporting_enabled: bool,
    ) -> Self {
        let diagnostics = Arc::new(diagnostics);
        let reporter =
            Reporter::new(deployment_id, diagnostics.clone(), REPORTING_URI, data_directory, is_reporting_enabled);
        let monitoring_server = MonitoringServer::new(diagnostics.clone(), monitoring_port, is_monitoring_enabled);

        Self { diagnostics, reporter, monitoring_server }
    }
    }
    ```

File: storage/storage.rs [hunk 15]
```diff
@@ -462,11 +470,12 @@ typedb_error!(
         DurabilityClientWrite(6, "Failed to write to durability client for database '{name}'.", name: String, (typedb_source: DurabilityClientError)),
 
         KeyspaceOpen(7, "Failed to open keyspace '{name}'.", name: String, (source: KeyspaceOpenError)),
+        Keyspace(8, "Failed to operate with keyspaces.", (source: KeyspaceError)),
```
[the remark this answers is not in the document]
farost: The name is inside the `KeyspaceError`, I can spend some time to make it a typedb_error if needed.

diagnostics/diagnostics_manager.rs
farost: I do think that it's easier to have this (or any other) intermediate layer, because `Diagnostics` are aware of `Reporter`s and `MonitoringServer`s (they have special formatting for them), and the `Reporter` and the `MonitoringServer` are aware of `Diagnostics`. So this layer of managing all these connections makes it simpler, but I may be too used to this structure, thus hesitant to change anything. 

I don't feel a strong urge to deconstruct it at the moment, and I don't like the diagnostics getter... Yet.

diagnostics/reporter.rs
farost: Uhhhmmmmmm `reset` sounds scarier and makes it look like we forget the whole data! I'll start by renaming `diagnsotics.to_reporting_json` to `to_reporting_json_against_snapshot`. It should give a better hint of the workflow and what's expected inside.

database/database.rs
```diff
@@ -426,6 +427,24 @@ impl Database<WALClient> {
...
+            data: DataLoadMetrics {
+                entity_count: schema.thing_statistics.total_entity_count,
+                relation_count: schema.thing_statistics.total_relation_count,
+                attribute_count: schema.thing_statistics.total_attribute_count,
+                has_count: schema.thing_statistics.total_has_count,
+                role_count: schema.thing_statistics.total_role_count,
+                storage_in_bytes: self.storage.estimate_size_in_bytes().expect("Expected storage size in bytes"),
+                storage_key_count: self.storage.estimate_key_count().expect("Expected storage key count"),
+            },
+            is_primary_server: true, // TODO: Should be retrieved differently for Cloud
```
farost, line 445: Ignoring it for now. it will be `true` even in Cloud as far as we have only one server.

database/database_manager.rs
```diff
@@ -8,45 +8,48 @@ use std::{
...
                     source: Arc::new(error),
-                })?;
-                let database = Database::<WALClient>::open(&entry.path())?;
-                Ok((database.name().to_owned(), Arc::new(database)))
-            })
-            .try_collect()?;
+                })?
+                .path();
+
+            if entry_path.file_name().unwrap().to_string_lossy().starts_with(SYSTEM_FILE_PREFIX) {
```
farost, line 43: The stream-based `filter` made the code a total mess with `try_`s forced by `Database::open()?`, so I've rewritten it to a loop. Didn't introduce any initial `capacity` to the HashMap, not sure if we need to worry about it (I'd assume the base case to be 1 system and 1 user db, so it's a single memory allocation operation anyway).

dependencies/typedb/repositories.bzl
```diff
@@ -12,10 +12,12 @@ def typedb_bazel_distribution():
     )
 
 def typedb_dependencies():
+    # TODO: Return typedb
     git_repository(
         name = "typedb_dependencies",
-        remote = "https://github.com/typedb/typedb-dependencies",
-        commit = "bc2381be8f38a6e55d15f61739dcf04d32e67769", # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_dependencies
+        remote = "https://github.com/farost/dependencies",
```
farost, line 18: https://github.com/typedb/typedb-dependencies/pull/564

diagnostics/diagnostics_manager.rs
```diff
@@ -0,0 +1,155 @@
...
+        let monitoring_server = if is_monitoring_enabled {
+            Some(MonitoringServer::new(diagnostics.clone(), monitoring_port))
+        } else {
+            None
+        };
+
+        Self { diagnostics, reporter, monitoring_server }
+    }
+
+    diagnostics_method! {
```
farost, line 67: I agree that this is not ideal. However, I'd prefer to stick to having the `DiagnosticsManager` not exposing `Diagnostics` to the outside world directly. Let the `DiagnosticsManager` manage all the diagnostics-related stuff, understanding if it needs to create and manage `Reporter`s and `MonitoringServer`s, let it be ready to have additional logic for these `Diagnostics` methods in the future.

I don't want to get rid of `DiagnosticsManager`, so exposing `Diagnostics` to `Server`, `TransactionService`, and other components look verbose and incorrect. We could also potentially add other "Diagnositcs"-like things to the Manager.

diagnostics/diagnostics_manager.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    pub async fn may_start_monitoring(&self) {
+        if let Some(server) = &self.monitoring_server {
+            server.start_serving().await;
+        }
+    }
+}
+
+pub fn run_with_diagnostics<F, T>(
```
farost, line 89: These wrappers can be simply used when we perform a user action:
`run_with_diagnostics(...... { transaction.commit() })`
This way, we don't need to worry about counting actions, their results, and error codes. Everything is accounted.

diagnostics/diagnostics_manager.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+fn get_status_error_code(status: &Status) -> Option<String> {
+    if let Ok(details) = status.check_error_details() {
+        if let Some(error_info) = details.error_info() {
+            return Some(error_info.reason.clone());
+        }
+    }
+    None // status.code() can return too long string descriptions from external libraries
```
farost, line 146: I tried using `status.code()` as @dmitrii-ubskii suggested, but it doesn't seem correct. For example, in Authentication, we use these status codes, which only have the `Display` impls with very-very long strings. I'd assume that `error metrics` are exactly the metrics for our server's status codes, and the authentication issues can be tracked through `action metrics` and the amount of total/successful calls without error details. Didn't want to bother with this refactoring.

However, if you're okay with sending long descriptions of the errors in such cases, I can add them.

diagnostics/diagnostics_manager.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+fn is_diagnostics_needed(database_name: Option<impl AsRef<str> + Hash>) -> bool {
+    // TODO: Would be good to reuse DatabaseManager's is_user_database() instead
```
farost, line 150: If `TODO`s are not addressed, it means that I left them for some future. Not sure what's correct here.

diagnostics/metrics.rs
```diff
@@ -0,0 +1,729 @@
...
+
+    pub fn server_id(&self) -> &str {
+        &self.server_id
+    }
+
+    pub fn is_reporting_enabled(&self) -> bool {
+        self.is_reporting_enabled
+    }
+
+    pub fn to_reporting_json(&self) -> JSONValue {
```
farost, line 47: I was thinking about having trait(s), but it ended up being quite messy, so I gave up and implemented these methods in a straightforward way. No "overrefactoring".

diagnostics/metrics.rs
```diff
@@ -0,0 +1,729 @@
...
+}
+
+#[derive(Debug, Hash, Copy, Clone, PartialEq, Eq)]
+pub enum LoadKind {
+    SchemaTransactions,
+    ReadTransactions,
+    WriteTransactions,
+}
+
+// TODO: Would be great to have a static protection that every kind is covered here
```
farost, line 645: idk

diagnostics/monitoring_server.rs
farost: Done!

diagnostics/reporter.rs
```diff
@@ -0,0 +1,159 @@
...
+};
+use resource::constants::{
+    common::SECONDS_IN_MINUTE,
+    diagnostics::{DISABLED_REPORTING_FILE_NAME, REPORT_INTERVAL, REPORT_ONCE_DELAY},
+};
+
+use crate::{hash_string_consistently, Diagnostics};
+
+#[derive(Debug)]
+pub struct Reporter {
```
farost, line 37: I'm not yet sure if I want to use this class purely for the diagnostics service or for both the diagnostics service and PostHog. Tight coupling can become painful in the future if we want to configure them separately. I may refactor this class in the next PR with the PostHog introduction.

diagnostics/version.rs
```diff
@@ -0,0 +1,7 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+pub const JSON_API_VERSION: u64 = 1;
```
farost, line 7: Couldn't come up with a better way of configuring it. Maybe it's better to place it to constants, but I like how it's placed in this package.

main.rs
```diff
@@ -25,9 +25,10 @@ async fn main() {
     print_ascii_logo(); // very important
     initialise_logging_global();
 
+    let deployment_id = None;
```
farost, line 28: I exposed the deployment id outside of the server so it can be simply overridden by TypeDB Cloud. Plus, it's not that bad conceptually: `Server` generated its server id, but it is actually a part of a deployment, and something else should prepare the information about this deployment. Thus, if we don't have the deployment information, we are a Core server without deployment and can put out `server_id` as the `deployment_id`.

server/BUILD
```diff
@@ -51,7 +76,11 @@ rust_library(
         "@crates//:tracing",
         "@crates//:uuid",
         "@crates//:pwhash",
-    ]
+    ],
+    crate_features = select({
+        ":development_mode": [],
+        ":published_mode": ["published"],
```
farost, line 82: Cargo sync is fine. 
`bazel run //:typedb_server_bin --//server:mode=published` runs the server in the published mode (not "release" so we don't get confused by the code performance without optimizations)
`bazel run //:typedb_server_bin --//server:mode=development` and `bazel run //:typedb_server_bin` run the server in the development mode (no `published` crate feature)

The cargo sync tool doesn't add this feature by default. Thus, `cargo build / run` work with development mode by default for the local builds, which is what we needed. However, we can either add the `published` feature to our `Cargo.toml`s or just use the CLI switch for the development mode if needed.

server/parameters/cli.rs
```diff
@@ -29,4 +30,22 @@ pub struct CLIArgs {
...
+
+    /// Port on which to expose the diagnostics monitoring endpoint
+    #[arg(long = "diagnostics.monitoring.port", default_value_t = MONITORING_DEFAULT_PORT)]
+    pub diagnostics_monitoring_port: u16,
+
+    /// Enable development mode for testing setups. Note that running TypeDB in development mode
+    /// may result in error reporting limitations (obstructing maintenance and support), additional
+    /// logging, restricted functionalities, and reduced performance
+    #[arg(long = "development-mode.enabled", hide = true)]
+    pub development_mode_enabled: bool,
```
farost, line 50: Unfortunately, we have to have this config flag as we want to run our TypeDB artifacts in tests and CI (say, `typedb-driver/tool` run scripts). So we'll need to carefully pass these switches to all these test scripts again (just like in 2.x). 

It's hidden from the help menu. Plus, the forced development mode (without the crate `published` feature) cannot be unset by this switch as we can only say 
`bazel run //:typedb_server_bin --//server:mode=published -- --development-mode.enabled`
or
`bazel run //:typedb_server_bin -- --development-mode.enabled`
but not
`bazel run //:typedb_server_bin -- --development-mode.enabled false`

server/parameters/config.rs
```diff
@@ -80,15 +105,40 @@ pub struct EncryptionConfig {
...
+            is_monitoring_enabled: true,
+            monitoring_port: MONITORING_DEFAULT_PORT,
+            is_service_reporting_enabled: true,
+        }
+    }
+}
+
+impl Default for DiagnosticsConfig {
+    fn default() -> Self {
+        Self::enabled()
```
farost, line 142: Everything is enabled by default.

server/parameters/cli.rs
```diff
@@ -29,4 +30,22 @@ pub struct CLIArgs {
     /// Path to the data directory
     #[arg(long = "storage.data", value_name = "DIR")]
     pub storage_data: Option<String>,
+
+    /// Enable usage metrics reporting
+    #[arg(long = "diagnostics.reporting.metrics", default_value_t = true, action=ArgAction::Set)]
+    pub diagnostics_reporting_metrics: bool, // used to be `statistics` in 2.x
```
farost, line 36: Called it `metrics` as we will have `errors` (for sentry). At the same time, `PostHog` should not be configured by a separate flag, so something general should go fine. 
Diagnostics generally consist of metrics in our model, so it's... bearable.

.circleci/config.yml
```diff
@@ -285,22 +285,22 @@ workflows:
       - test-deploy-snapshot-linux-x86_64:
           filters:
             branches:
-              only: [ "3.0" ]
+              only: [ "3.0-diagnostics" ]
```
farost, line 288: Temporary for benchmarks, will remove before merge

.factory/automation.yml
```diff
@@ -9,6 +9,7 @@ config:
     typeql: [build, release]
     typedb-protocol: [build, release]
     typedb-behaviour: [build]
+todo unbreak me
```
farost, line 12: Will remove soon

tests/behaviour/steps/connection/mod.rs
```diff
@@ -7,30 +7,35 @@
...
+use tokio::{runtime::Runtime, sync::OnceCell};
 
 use crate::{generic_step, Context};
 
 mod database;
 mod transaction;
 
-static TYPEDB: OnceLock<(TempDir, Arc<Mutex<typedb::Server>>)> = OnceLock::new();
+const DISTRIBUTION: &str = "TypeDB TEST";
+static TYPEDB: OnceCell<(TempDir, Arc<Mutex<typedb::Server>>)> = OnceCell::const_new();
```
farost, line 24: `Server::open` is async now, so I need to pass an `async` closure into `get_or_init`.

.circleci/config.yml
```diff
@@ -111,7 +111,7 @@ commands:
           export DEPLOY_ARTIFACT_USERNAME=$REPO_TYPEDB_USERNAME
           export DEPLOY_ARTIFACT_PASSWORD=$REPO_TYPEDB_PASSWORD
           bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
-          bazel run --jobs=8 --define version=$(cat VERSION) //:deploy-typedb-server --compilation_mode=opt -- release
+          bazel run --jobs=8 --define version=$(cat VERSION) //:deploy-typedb-server --compilation_mode=opt --//server:mode=published -- release
```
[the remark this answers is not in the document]
farost: You're correct! I've missed it because it didn't contain the "release" word...

resource/constants.rs
```diff
@@ -4,11 +4,21 @@
...
 pub mod server {
     use std::time::Duration;
 
+    use crate::constants::common::{SECONDS_IN_HOUR, SECONDS_IN_MINUTE};
+
     pub const ASCII_LOGO: &str = include_str!("typedb-ascii.txt");
 
+    pub const VERSION: &str = include_str!("../VERSION");
+
     pub const GRPC_CONNECTION_KEEPALIVE: Duration = Duration::from_secs(2 * 60 * 60);
```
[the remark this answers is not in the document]
farost: Idk how I ignored it, I really spent 5 seconds observing if there is anything to refactor!

diagnostics/monitoring_server.rs
```diff
@@ -0,0 +1,77 @@
...
+                    }))
+                }
+            });
+
+            let server = Server::bind(&addr).serve(make_svc);
+            if let Err(e) = server.await {
+                eprintln!("Monitoring server error: {}", e);
+            }
+        });
+    }
```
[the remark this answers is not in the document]
farost: Well, it does shut down. What can be a sign of an incorrect shutdown?

server/BUILD
flyingsilverfin: that's fine!

diagnostics/diagnostics_manager.rs
flyingsilverfin: yeah i think it's ok without the super long strings, we can imrpove this in the future if not illustrative enough!

diagnostics/metrics.rs
flyingsilverfin: Agree, maybe dmitrii has an idea, but I'm outta ideas

diagnostics/monitoring_server.rs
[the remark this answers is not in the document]
flyingsilverfin: hanging!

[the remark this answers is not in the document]
farost: Well, it does shut down on pressing Ctrl + C like always. Not sure if there is another way

flyingsilverfin (approved):

PR #7292, opened 2024-12-17
Improved query profile formatting and cleaner beam search

cxdorn:
### Printing and log tracing
* Improved Display and Debug prints
* Log tracing for query profiler now prints variables (instead of positions)
* Added log tracing for query planning

### Query planning
* Support for Relation indexes
* Hash checks during planning
   * trivial permutations of the "same" plan are ignored.
   * "same" means produces the same variables with the same cost
* Make planner more deterministic
   * trivial plan elements are now decided upon determinstically, freeing the planner to consider only the more "heavy-weight" choices.
* Add heuristic weighting that penalizing larger intermediate results more than before, reward producing results early
* Various bug fixes (joins in lowering)
   
## Future work
* Improve focus on "difficult bits" in planner
* Integrate dynamic planning approach (see postgres) + allow for non-linear plans and execution (DAGs are all the hype now)


compiler/executable/match_/planner/plan.rs
```diff
@@ -577,48 +603,74 @@ impl<'a> ConjunctionPlanBuilder<'a> {
...
+        let complete_plan = best_plan.into_complete_plan(&self.graph);
         (complete_plan.vertex_ordering, complete_plan.pattern_metadata)
     }
 
     // Execute plans
     pub(super) fn plan(self) -> ConjunctionPlan<'a> {
         // Beam plan
+        let duration = Instant::now();
         let (ordering, metadata) = self.beam_search_plan();
+        println!("Planning took {:?}", duration.elapsed());
```
flyingsilverfin, line 665: some random printlns that if you want to keep should use event! with a LEVEL and we can enable/disable with the logging configuration

compiler/executable/match_/planner/mod.rs
```diff
@@ -60,7 +60,12 @@ pub fn compile(
         statistics,
     )
     .lower(input_variables.keys().copied(), selected_variables.to_vec(), &assigned_identities, variable_registry)
-    .finish(variable_registry)
+    .finish(variable_registry);
+
+    // println!("Variable names:\n {:#?}", variable_registry.variable_names());
+    // println!("Variable positions:\n {:#?}", plan.variable_positions());
```
flyingsilverfin, line 66: cleanup

PR #7293, opened 2024-12-17
Extensible user credential representation in system database

lolski:
Create an extensible representation of user credential, so that it can be extended by concrete types down the line. Currently, password is the only concrete type, but later can be extended into token, SSH keypair, and so on.

## Motivation
To create an extensible credential representation as there's a high chance we may introduce new types down the line.

## Implementation
1. Update system database schema definition:
    - create a credential supertype that can later be extended into multiple kinds of credentials
    - add credential ID in order to be able to refer to a specific credential
3. Update system database "repository" classes according to the new schema definition


.circleci/config.yml
```diff
@@ -285,27 +285,27 @@ workflows:
       - test-deploy-snapshot-linux-x86_64:
           filters:
             branches:
-              only: [ "3.0" ]
+              only: [ "3.0", "authentication-schema" ]
```
farost, line 288: Don't forget to remove it before merging

system/repositories.rs
```diff
@@ -74,21 +74,23 @@ pub mod user_repository {
         function_manager: &FunctionManager,
         query_manager: &QueryManager,
         user: &User,
-        credential: &Credential,
+        credentials: &Credential,
```
farost, line 77: Should not the struct be `Credentials` instead of `Credential` as well?

system/repositories.rs
```diff
@@ -153,9 +155,9 @@ pub mod user_repository {
         let unexpected_error_msg = "An unexpected error occurred when attempting to delete a user";
         let query = parse_query(
             format!(
-                "match $up isa user-password, links (user: $u, password: $p);
+                "match $uc isa user-credentials, links (user: $u, credentials: $p);
```
farost, line 158: `credentials: $p` -> `credentials: $c` would be a little cleaner
with `delete $u; $p; $uc;",` -> `delete $u; $c; $uc;",`

system/schema.tql
```diff
@@ -10,12 +10,15 @@ define
...
 
-    relation user-password,
+    entity password, sub credentials,
+        owns hash @card(1);
+
+    relation user-credentials,
         relates user @card(1),
-        relates password @card(1);
\ No newline at end of file
+        relates credentials @card(1..);
```
farost, line 24: Let's add an empty line at the end of the file while we are here

farost (approved):
Looks good! Formatting comments

system/repositories.rs
lolski: Yes, I'm planning to do it in a follow up PR after we've finished with the release since there are many places where the rename should take place.

As you've seen from this PR and also the PRs in driver, I've only introduced the new name `Credentials` on places that are public facing and should not change after release, namely, the TypeDB Driver API and this database schema.

system/schema.tql
lolski: Fixed in f4dca852654fa7d18d662be1ccb685a02d3b8a4d.

flyingsilverfin (approved):
As georgii said - delete the branch filter

PR #7296, opened 2024-12-19
Introduce PostHog diagnostics and correct server shutdown

farost:
### Diagnostics
We add PostHog diagnostics events reporting for gathering usage data about users' journeys through TypeDB and enhance the user experience after connecting it with other metrics collected in PostHog. 
The reporting configuration remains the same: the `--diagnostics.reporting.metrics` flag controls both the old (Service diagnostics) and the new (PostHog) reporting, letting a user disable everything in one action.

There are two groups of events sent:
* server usage: hourly heartbeats with deployment/server IDs and the version of the server, can optionally contain information about actions performed without connection to a specific database (opened connection, created user, etc.);
* database usage:  optional event (can be not sent if there is nothing to send), similar to the optional server usage action metrics, but regarding a specific database (created database, opened transaction, executed query). 

The format of the events is the following:
* If nothing happened for the past hour:
```
{
  "batch": [
    {
      "event": "server_usage",
      "properties": {
        "distinct_id": "ARUTFWVJ3MIOJJ50",
        "server_id": "ARUTFWVJ3MIOJJ50",
        "version": "3.0.0-alpha-10"
      }
    }
  ]
}
```
* If anything happened for the past hour for multiple databases:
```
{
  "batch": [
    {
      "event": "database_usage",
      "properties": {
        "distinct_id": "ARUTFWVJ3MIOJJ50",
        "server_id": "ARUTFWVJ3MIOJJ50",
        "version": "3.0.0-alpha-10",
        "databases_presence_checked": 80,
        "databases_created": 81,
        "queries_executed": 237,
        "databases_deleted": 80,
        "transactions_opened": 133,
        "transactions_committed": 57,
        "database": 95615005977010740
      }
    },
    {
      "event": "server_usage",
      "properties": {
        "distinct_id": "ARUTFWVJ3MIOJJ50",
        "server_id": "ARUTFWVJ3MIOJJ50",
        "version": "3.0.0-alpha-10",
        "all_users_retrieved": 79,
        "all_databases_retrieved": 271,
        "connections_opened": 248,
        "authentications_performed": 988
      }
    },
    {
      "event": "database_usage",
      "properties": {
        "distinct_id": "ARUTFWVJ3MIOJJ50",
        "server_id": "ARUTFWVJ3MIOJJ50",
        "version": "3.0.0-alpha-10",
        "databases_deleted": 1,
        "transactions_opened": 2,
        "transactions_committed": 1,
        "databases_presence_checked": 4,
        "databases_created": 1,
        "database": 3619243786487231000
      }
    },
    {
      "event": "database_usage",
      "properties": {
        "distinct_id": "ARUTFWVJ3MIOJJ50",
        "server_id": "ARUTFWVJ3MIOJJ50",
        "version": "3.0.0-alpha-10",
        "databases_created": 1,
        "database": 3244421341483603000
      }
    },
    {
      "event": "database_usage",
      "properties": {
        "distinct_id": "ARUTFWVJ3MIOJJ50",
        "server_id": "ARUTFWVJ3MIOJJ50",
        "version": "3.0.0-alpha-10",
        "databases_presence_checked": 4,
        "database": 2857950145998399500
      }
    },
    {
      "event": "database_usage",
      "properties": {
        "distinct_id": "ARUTFWVJ3MIOJJ50",
        "server_id": "ARUTFWVJ3MIOJJ50",
        "version": "3.0.0-alpha-10",
        "databases_deleted": 1,
        "databases_presence_checked": 1,
        "databases_created": 1,
        "database": 5191865013595434000
      }
    }
  ]
}
```

### Server shutdown
Now, the server is correctly shut down when using `CTRL-C`signal, awaiting for all the background tasks and printing pretty and informative messages:
```
Running TypeDB CE 3.0.0-alpha-10.
Ready!
^C
Received a CTRL-C signal. Shutting down...
Exited.
```
Additionally, we add a feature of sending diagnostics (both types) on server shutdown (if enabled).

## Motivation
Integration with PostHog for management's queries aimed at exploring the quality of the user experience.

### Diagnostics
We introduce two `ReportingEndpoint`s:
* Service / DiagnosticsService (the old one);
* PostHog (the new one).

PostHog reporting is an extension to the already existing `diagnostics/Reporter`. 
The metrics we are interested in are partially `server properties`, `server metrics` (version), and `action metrics`.
Its reporting is controlled by the same mechanisms from the main `report` method.

However, we want this data to be a little more precise, sustainable, and consistent through time (it's acceptable to lose some as it's just diagnostics). So the main differences from diagnostics are:
* Small retries for the event reporting (3 attempts with small increasing timeouts between them);
* These retries introduce a higher delay between creating a diagnostics report and making a snapshot (to take action metrics' diffs every hour). Considering that we'd like to make it less lossy, we use a special, a little more memory-greedy approach of snapshotting: get a json -> take a snapshot -> try sending it -> if not successful, give up and restore the snapshot from its backup. 

With this approach, we achieve the same rule of "if could not send in this hour, combine this data with the data from the next hour on the next attempt", but with an additional retry logic with better precision. 
**This approach is implemented only for PostHog reporting**. Can be extended to the service diagnostics in the future if we like it. 

The shutdown diagnostics reporting does not collide with the regular hourly reporting. The key points of the current naive implementations are:
* If we sent a snapshot of diagnostics a second before the shutdown, we still send an "empty" batch. Is not considered a serious problem for simplicity.
* If we are in the process of waiting for reporting timeouts (currently, its' 500ms, 1s, 4s -> 5.5s in total), we will wait for them to finish as they ignore the drop processing. Again, it doesn't sound too bad now, but it's something to be fixed in the future.

### Shutting down
We discovered (by trying the shut down feature for diagnostics) that we don't actually trigger any `Drop`s because we don't currently shut down our Server's tonic/tokio tasks. Now, it's fixed by using `tokio::signal::ctrl_c()`.

### DIFF
--- a/diagnostics/lib.rs
+++ b/diagnostics/lib.rs
@@ -15,7 +15,7 @@ use std::{
 };
 
 use error::TypeDBError;
-use serde_json::Value as JSONValue;
+use serde_json::{json, Value as JSONValue};
 use xxhash_rust::xxh3::Xxh3;
 
 use crate::metrics::{
@@ -120,14 +120,14 @@ impl Diagnostics {
         self.lock_error_metrics_write().values_mut().for_each(|metrics| metrics.take_snapshot());
     }
 
-    pub fn to_reporting_json_against_snapshot(&self) -> JSONValue {
+    pub fn to_service_reporting_json_against_snapshot(&self) -> JSONValue {
         match self.is_full_reporting {
-            true => self.to_full_reporting_json(),
-            false => self.to_minimal_reporting_json(),
+            true => self.to_full_service_reporting_json(),
+            false => self.to_minimal_service_reporting_json(),
         }
     }
 
-    fn to_full_reporting_json(&self) -> JSONValue {
+    fn to_full_service_reporting_json(&self) -> JSONValue {
         let mut diagnostics = self.server_properties.to_reporting_json();
 
         diagnostics["server"] = self.server_metrics.to_json();
@@ -144,7 +144,7 @@ impl Diagnostics {
         let actions = self
             .lock_action_metrics_read()
             .iter()
-            .map(|(database_hash, metrics)| metrics.to_reporting_json(database_hash))
+            .map(|(database_hash, metrics)| metrics.to_service_reporting_json(database_hash))
             .flatten()
             .collect();
         diagnostics["actions"] = JSONValue::Array(actions);
@@ -160,7 +160,7 @@ impl Diagnostics {
         diagnostics
     }
 
-    fn to_minimal_reporting_json(&self) -> JSONValue {
+    fn to_minimal_service_reporting_json(&self) -> JSONValue {
         let mut diagnostics = self.server_properties.to_reporting_json();
         diagnostics["server"] = self.server_metrics.to_reporting_minimal_json();
         diagnostics
@@ -247,6 +247,19 @@ impl Diagnostics {
             if user_errors_data.len() > user_errors_data_header_length { user_errors_data } else { String::new() },
         ]
         .join("\n")
+    }
+
+    pub fn to_posthog_reporting_json_against_snapshot(&self, api_key: &str) -> JSONValue {
+        let deployment_id = self.server_properties.deployment_id();
+        let batch: Vec<_> = self
+            .lock_action_metrics_read()
+            .iter()
+            .map(|(database_hash, metrics)| metrics.to_posthog_reporting_json(deployment_id, database_hash))
+            .collect();
+        json!({
+            "api_key": api_key,
+            "batch": batch,
+        })
     }
 
     fn hash_database(database_name: impl AsRef<str> + Hash) -> DatabaseHash {
--- a/diagnostics/lib.rs
+++ b/diagnostics/lib.rs
@@ -114,10 +114,18 @@ impl Diagnostics {
         errors.get(&database_hash).expect("Expected database in errors").submit(error_code);
     }
 
-    pub fn take_snapshot(&self) {
+    pub fn take_service_snapshot(&self) {
         self.lock_load_metrics_read().values().for_each(|metrics| metrics.take_snapshot());
-        self.lock_action_metrics_write().values_mut().for_each(|metrics| metrics.take_snapshot());
+        self.lock_action_metrics_write().values_mut().for_each(|metrics| metrics.take_service_snapshot());
         self.lock_error_metrics_write().values_mut().for_each(|metrics| metrics.take_snapshot());
+    }
+
+    pub fn take_posthog_snapshot(&self) {
+        self.lock_action_metrics_write().values_mut().for_each(|metrics| metrics.take_posthog_snapshot());
+    }
+
+    pub fn restore_posthog_snapshot(&self) {
+        self.lock_action_metrics_write().values_mut().for_each(|metrics| metrics.restore_posthog_snapshot());
     }
 
     pub fn to_service_reporting_json_against_snapshot(&self) -> JSONValue {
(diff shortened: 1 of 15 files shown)


diagnostics/lib.rs
```diff
@@ -249,6 +257,27 @@ impl Diagnostics {
...
+
+        let mut batch: Vec<_> = self
+            .lock_action_metrics_read()
+            .iter()
+            .filter_map(|(database_hash, metrics)| {
+                metrics.to_posthog_reporting_json(database_hash, common_properties.clone())
+            })
+            .collect();
+        if batch.is_empty() {
+            batch.push(ActionMetrics::empty_posthog_reporting_json(common_properties));
```
farost, line 272: The heartbeat (server state) is mandatory

diagnostics/reporter.rs
```diff
@@ -92,67 +88,107 @@ impl Reporter {
...
         }
     }
 
-    fn calculate_initial_delay(&self) -> Duration {
-        let report_interval_secs = REPORT_INTERVAL.as_secs();
-        assert!(report_interval_secs == 3600, "Modify the algorithm if you change the interval!");
+    async fn report(diagnostics: Arc<Diagnostics>) -> bool {
+        let service_task = Self::report_diagnostics_service(diagnostics.clone());
+        let posthog_task = Self::report_posthog(diagnostics.clone());
+        let (is_service_reported, is_posthog_reported) = tokio::join!(service_task, posthog_task);
```
farost, line 105: Execute both concurrently

resource/constants.rs
```diff
@@ -17,8 +17,6 @@ pub mod server {
 
     pub const ASCII_LOGO: &str = include_str!("typedb-ascii.txt");
 
-    pub const VERSION: &str = include_str!("../VERSION");
```
farost, line 20: We want to substitute the `VERSION` file in `cloud`.

resource/constants.rs
```diff
@@ -108,8 +106,17 @@ pub mod diagnostics {
 
     pub const UNKNOWN_STR: &'static str = "Unknown";
 
-    pub const REPORTING_URI: &str = "https://diagnostics.typedb.com/";
+    pub const SERVICE_REPORTING_URI: &str = "https://diagnostics.typedb.com/";
+    pub const POSTHOG_BATCH_REPORTING_URI: &str = "https://us.i.posthog.com/batch/";
+    // The key is write-only and safe to expose
+    pub const POSTHOG_API_KEY: &str = "phc_pYoyROZCtNDL8obeJfLZ8cP0UKzIAxmd0JcQQ03i07T";
```
farost, line 112: Alex said that it's a common approach to just use these public keys for pushes, and PostHog also mentions it on their settings page.

diagnostics/lib.rs
```diff
@@ -114,23 +114,31 @@ impl Diagnostics {
...
-    pub fn take_snapshot(&self) {
+    pub fn take_service_snapshot(&self) {
         self.lock_load_metrics_read().values().for_each(|metrics| metrics.take_snapshot());
-        self.lock_action_metrics_write().values_mut().for_each(|metrics| metrics.take_snapshot());
+        self.lock_action_metrics_write().values_mut().for_each(|metrics| metrics.take_service_snapshot());
         self.lock_error_metrics_write().values_mut().for_each(|metrics| metrics.take_snapshot());
     }
 
-    pub fn to_reporting_json_against_snapshot(&self) -> JSONValue {
+    pub fn take_posthog_snapshot(&self) {
```
farost, line 123: I was thinking about having just a single snapshot for everything, but the goals are different, and it was a huge headache. So it's both more correct and more comfortable to have separate snapshots. It also aligns with the existing monitoring, which does not use snapshots!

diagnostics/metrics.rs
```diff
@@ -787,6 +904,32 @@ impl ActionKind {
             (Self::TransactionQuery, ActionInfo::default()),
         ])
     }
+
+    pub fn to_posthog_event_name(&self) -> &'static str {
+        match self {
+            ActionKind::ConnectionOpen => "connection_opens",
```
farost, line 910: Reviewed with Alex

flyingsilverfin (approved):

PR #7297, opened 2024-12-20
Add concept document streaming from write queries

farost:
We add concept document streaming from write query pipelines. Previously, we eagerly unwrapped the answers as concept rows, not letting users use `fetch` queries in write pipelines.

### DIFF
--- a/server/parameters/config.rs
+++ b/server/parameters/config.rs
@@ -132,11 +132,7 @@ pub struct DiagnosticsConfig {
 
 impl DiagnosticsConfig {
     pub fn enabled() -> Self {
-        Self {
-            is_monitoring_enabled: true,
-            monitoring_port: MONITORING_DEFAULT_PORT,
-            is_reporting_enabled: true,
-        }
+        Self { is_monitoring_enabled: true, monitoring_port: MONITORING_DEFAULT_PORT, is_reporting_enabled: true }
     }
 }
 
--- a/server/service/transaction_service.rs
+++ b/server/service/transaction_service.rs
@@ -29,6 +29,7 @@ use diagnostics::{
 use error::typedb_error;
 use executor::{
     batch::Batch,
+    document::ConceptDocument,
     pipeline::{
         pipeline::Pipeline,
         stage::{ExecutionContext, ReadPipelineStage, StageIterator},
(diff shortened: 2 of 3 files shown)


server/parameters/config.rs
```diff
@@ -132,11 +132,7 @@ pub struct DiagnosticsConfig {
 
 impl DiagnosticsConfig {
     pub fn enabled() -> Self {
-        Self {
-            is_monitoring_enabled: true,
-            monitoring_port: MONITORING_DEFAULT_PORT,
-            is_reporting_enabled: true,
-        }
+        Self { is_monitoring_enabled: true, monitoring_port: MONITORING_DEFAULT_PORT, is_reporting_enabled: true }
```
farost, line 135: Rustfmt

server/service/transaction_service.rs
```diff
@@ -131,8 +133,10 @@ pub(crate) struct TransactionService {
     transaction: Option<Transaction>,
     request_queue: VecDeque<(Uuid, typeql::query::Pipeline)>,
     responders: HashMap<Uuid, (JoinHandle<()>, QueryStreamTransmitter)>,
-    running_write_query:
-        Option<(Uuid, JoinHandle<(Transaction, Result<(StreamQueryOutputDescriptor, Batch), Box<QueryError>>)>)>,
+    running_write_query: Option<(
+        Uuid,
+        JoinHandle<(Transaction, Result<Either<WriteQueryBatchAnswer, WriteQueryDocumentsAnswer>, Box<QueryError>>)>,
```
farost, line 138: I wanted to highlight that there is an `Either`, so I don't create a `type ... = Either<>`

farost:
Tested with a driver bdd:
```

  Scenario: Driver processes concept row query answers correctly
    Given connection open schema transaction for database: typedb
    Given typeql schema query
      """
      define entity person owns name @key; attribute name, value string;
      """
    Given transaction commits
    When connection open write transaction for database: typedb
    When get answers of typeql write query
      """
      insert $p isa person, has name "John";
      fetch {
        "name": $p.name,
        "sub fetch": {
          "all attributes": { $p.* },
        }
      };
      """
    Then answer type is: concept documents
    Then answer type is not: ok
    Then answer type is not: concept rows
    Then answer query type is: write
    Then answer query type is not: schema
    Then answer query type is not: read
    Then answer size is: 1
    Then answer contains document:
    """
    {
      "name": "John",
      "sub fetch": {
        "all attributes": {
          "name": "John"
        }
      }
    }
    """

```

dmitrii-ubskii (approved):

PR #7300, opened 2024-12-23
Fix rollback and remove dependency on typedb-common

farost:
We fix the `rollback` feature, which used to hang the transaction operated from an external driver, not allowing the operations to proceed correctly.

Additionally, we remove an outdated Bazel dependency on `typedb-common`, which is no longer needed for TypeDB 3.0.

## Implementation
There were two issues:
* We didn't return the transaction into the member of the service, so it got lost for the future operations;
* We could not receive any more operations for the transaction stream as the driver did not receive any response on this operation, while the `ControlFlow` was hanging on `Continue`.


server/service/transaction_service.rs
```diff
@@ -586,14 +587,27 @@ impl TransactionService {
         }
 
         match self.transaction.take().unwrap() {
-            Transaction::Read(_) => {
+            Transaction::Read(transaction) => {
+                self.transaction = Some(Transaction::Read(transaction));
```
flyingsilverfin, line 591: Oof good catch!

flyingsilverfin (approved):

PR #7305, opened 2025-01-06
Improve TypeDB error macro

dmitrii-ubskii:

We refactor the `typedb_error!` macro to avoid the use of meaningful parentheses and position-dependence.

## Motivation

Previously, when defining a new typedb error variant with a source, that source had to come at the end of the field list, and be wrapped in parentheses to help distinguish it from the rest of the fields. We refactor the macro by creating internal rules that perform incremental token tree munching to lift those restrictions.

### DIFF
--- a/common/error/error.rs
+++ b/common/error/error.rs
@@ -121,7 +121,7 @@ macro_rules! typedb_error {
         impl $crate::TypeDBError for $name {
             fn variant_name(&self) -> &'static str {
                 match self {
-                    $(Self::$variant { .. } => &stringify!($variant),)*
+                    $(Self::$variant { .. } => stringify!($variant),)*
                 }
             }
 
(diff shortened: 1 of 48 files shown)


common/error/error.rs
```diff
@@ -104,41 +104,24 @@ impl<T: TypeDBError> TypeDBError for Box<T> {
...
-                    )*
-                    _ => (),
-               }
-           };
-        }
+        const _: () = {
+            // fail to compile if any Numbers are the same
+            trait Assert {}
+            $(impl Assert for [(); $number ] {})*
+        };
```
dmitrii-ubskii, line 119: This didn't actually do anything. Replaced with a trait assert trick.

common/error/error.rs
```diff
@@ -160,51 +141,81 @@ macro_rules! typedb_error {
...
+                    $(Self::$variant { .. } => $number,)*
                 }
             }
 
             fn format_description(&self) -> String {
                 match self {
-                    $(
-                        Self::$variant { $( $payload_name, )* .. } => format!($description),
-                    )*
+                    $(typedb_error!(@args $variant { $($($arg)*)? }) => format!($description),)*
```
dmitrii-ubskii, line 150: I'd prefer to verify here that all non-source fields are used in the formatting, but that would make the macro a lot more complex.

flyingsilverfin (approved):

PR #7308, opened 2025-01-08
Add Sentry for critical error reporting and fix panics on diagnostics initialization

farost:
We add integration with Sentry to allow critical errors (e.g. `panic!`s) reporting. This information will help us see and eliminate unexpected TypeDB Server crashes.
Please use the `-diagnostics.reporting.errors` boolean option to disable this feature if its work is undesirable (note that it will reduce the efficiency of our maintenance).

Additionally, an overflow subtraction bug that sometimes affects diagnostics initialization and leads to crashes is fixed.
 
## Motivation
The goal of the Sentry integration is to control unexpected errors, which most likely indicate bugs or other issues introduced by developers. It intentionally skips handled user errors as they are reported to the diagnostics service as "user-related errors".

## Implementation
Based on [the official Sentry Rust setup page](https://docs.sentry.io/platforms/rust/), the simplest way to initialize Sentry correctly and not worry about threads and "sentry hubs" is to leave Sentry in the `main` function instead of integrating it inside of the `diagnostics` package (which was desired). 
From the architecture point of view, it's not the cleanest solution, but it is a decent approach from the perspective of how the Sentry crate works in Rust. More comments are left in the PR diff.


diagnostics/reporter.rs
```diff
@@ -207,7 +207,7 @@ impl Reporter {
             hash_string_consistently(&self.deployment_id) % (report_interval_secs / SECONDS_IN_MINUTE);
 
         let delay_secs = if current_minute > scheduled_minute {
-            report_interval_secs - (current_minute + scheduled_minute) * SECONDS_IN_MINUTE
+            report_interval_secs - current_minute * SECONDS_IN_MINUTE + scheduled_minute * SECONDS_IN_MINUTE
```
farost, line 210: An amazing bug (and really sad that I haven't caught it before) where I acted like I was a compiler's optimizer. The former string might panic because of the unsigned values (`3600 - (55 + 10) * 60`).

main.rs
```diff
@@ -11,33 +11,38 @@ use std::{path::PathBuf, str::FromStr};
...
-    config::{Config, DiagnosticsConfig, EncryptionConfig},
+    config::{Config, DiagnosticsConfig, EncryptionConfig, ServerConfig},
 };
 
 const DISTRIBUTION: &str = "TypeDB CE";
 const VERSION: &str = include_str!("VERSION");
 
-#[tokio::main]
-async fn main() {
+fn main() {
```
farost, line 24: I tested Sentry with `#[tokio:main]` instead of this runtime refactoring and everything was fine. I also read additional documentation and their github (sadly, could not find complex examples of late initialization and stuff) and only understood that their "Hub"s can have random issues if you work with them in different threads, so I gave up on my idea to encapsulate Sentry in `diagnostics`, following their simplest example by:
1. Making `main` sync and creating an explicit server async block after the CLI processing;
2. Initializing Sentry in this `main`

Not a fan of this decision because:
1. It makes code less compact 
2. It forces us to copypaste it in `typedb-cloud`

But it seems safer and more "correct" as we have the scope of the whole program after the CLI parsing, which is covered by the `_error_reporting_guard`.

Meh.

main.rs
```diff
@@ -11,33 +11,38 @@ use std::{path::PathBuf, str::FromStr};
...
+    initialise_logging_global();
 
+    let cli_args = server::parameters::cli::CLIArgs::parse();
     print_ascii_logo(); // very important
-    initialise_logging_global();
 
     let deployment_id = None;
+    let is_error_reporting_enabled = cli_args.diagnostics_reporting_errors;
     let config = get_configuration(cli_args);
+    let _error_reporting_guard = setup_error_reporting(config.get_server_config(), is_error_reporting_enabled);
```
farost, line 34: I decided that I wanted to reuse the development mode from the config (which is the cleanest way). This forces me to initialize Sentry only after the config initialization (so we won't see exceptions before this point). We did ignore even more initialization errors in 2.x, so it might be fine.

main.rs
```diff
@@ -75,3 +80,14 @@ fn setup_abort_on_panic() {
         })
     });
 }
+
+fn setup_error_reporting(server_config: &ServerConfig, is_error_reporting_enabled: bool) -> Option<SentryGuard> {
```
farost, line 84: This is the only place where I work with `Sentry`. I could also add user error reporting to filter it out later on the Sentry side, but we already agreed that we have `TypeDB Diagnostics` for these purposes. This way, Sentry will work with only critical errors, which will reduce noise and clear endpoint responsibilities.

server/parameters/cli.rs
```diff
@@ -40,6 +40,10 @@ pub struct CLIArgs {
     #[arg(long = "diagnostics.reporting.metrics", default_value_t = true, action=ArgAction::Set)]
     pub diagnostics_reporting_metrics: bool, // used to be `statistics` in 2.x
 
+    /// Enable critical errors reporting
```
farost, line 43: Probably `error reporting`? If yes, I will fix the PR description as well. Please teach me English

dependencies/typedb/repositories.bzl
```diff
@@ -12,11 +12,17 @@ def typedb_bazel_distribution():
     )
 
 def typedb_dependencies():
+    # TODO: Return typedb after merge
     git_repository(
         name = "typedb_dependencies",
-        remote = "https://github.com/typedb/typedb-dependencies",
-        commit = "52191031bb1231c714a2f2017571d61c81b4a890",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_dependencies
+        remote = "https://github.com/farost/dependencies",
+        commit = "fc508101efb5c61a63262c174c02649c8a0c7795",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_dependencies
```
farost, line 19: https://github.com/typedb/typedb-dependencies/pull/565

server/parameters/cli.rs
dmitrii-ubskii: "Fatal error reporting" maybe? Since we're not going to report syntax errors and such to Sentry.

server/parameters/config.rs
```diff
@@ -80,15 +80,19 @@ impl Config {
             storage: StorageConfig { data: data_directory.to_owned() },
         }
     }
+
+    pub fn get_server_config(&self) -> &ServerConfig {
+        &self.server
+    }
```
dmitrii-ubskii, line 86: We should unlearn the `get_` prefixes now
```suggestion
    pub fn server_config(&self) -> &ServerConfig {
        &self.server
    }
```

main.rs
dmitrii-ubskii: Can there even be exceptions above this line? I'm not seeing any `Result`s or `unwrap()`s

main.rs
dmitrii-ubskii: 1. I don't think we've lost that much terseness here.
2. What is actually all that's duplicated? Isn't it just the contents of `main()`? That's not that much.

dmitrii-ubskii (approved):

main.rs
farost: @dmitrii-ubskii Still a need to duplicate it and extend both sources all the time. Good if you're okay with it!

main.rs
farost: There should not be, but who knows what can happen inside either of the calls above in the future. What if we eagerly unwrap something while creating a config. So it's a blank spot

server/parameters/cli.rs
farost: I've used `critical` in case we want to extend this set in the future =) @flyingsilverfin thoughts?

server/parameters/config.rs
farost: For some reason, I thought that the usage of `get_` is actually the preferred naming! Thanks

main.rs
flyingsilverfin: This is ok i think as well!

server/parameters/cli.rs
flyingsilverfin: I'd just call it `Enable error reporting` :)

flyingsilverfin (approved):
Approved w/ one recommendation

server/parameters/cli.rs
farost: Discussed in person, decided to leave `critical` in place!

PR #7312, opened 2025-01-14
Apply typedb error macro to top-level packages

flyingsilverfin:

TypeDB now reports extended error stack traces for all error types in the compiler and the intermediate representation builder, improving debuggability and ease-of-use.

## Motivation

Improving UX for users of TypeDB.

## Implementation

We convert the following errors into `TypeDBError` implementations:
- `ExpressionCompilationError`
- `ExpressionExecutionError`
- `WritecompilationError`
- `AnnotationError`
- `LiteralParseError`
- `ExpressionRepresentationError`
- `FunctionReadError`

This PR does not yet convert all the errors in `//concept`, `//encoding`, or `//storage`


compiler/annotation/expression/instructions/mod.rs
```diff
@@ -69,3 +70,11 @@ impl Error for ExpressionEvaluationError {
         }
     }
 }
+
+typedb_error!(
```
flyingsilverfin, line 74: WIP replacing this one here, I normally make a duplicate in the macro with suffix `2`, then delete the original and the `2` -- this leads to all the imports working everywhere automatically, then fix all the broken builds because the invokation is wrong!

PR #7313, opened 2025-01-15
Cleanup todos

krishnangovindraj:
Replaces usages of the todo macro with either errors which will be returned, or custom macros if the line is unreachable. This reduces server crashes due to panics when execution reaches the todo.

## Motivation
Avoid server crashes due to unimplemented code-paths.

## Implementation
Replaces usages of the todo macro (which would cause server crashes when hit) with either errors which will be returned, or custom macros if the line is unreachable. The custom macros (defined in `error.rs`) use an enum `UnimplementedFeature` which helps track unimplemented code by feature. This should be useful in the future when we're getting around to implementing them since we can just delete the variant and see what breaks.


server/service/document.rs
```diff
@@ -173,8 +174,17 @@ fn encode_leaf(
                 }
             }),
         }),
-        DocumentLeaf::Kind(_) => {
-            todo!()
-        }
+        DocumentLeaf::Kind(kind) => Ok(typedb_protocol::concept_document::node::Leaf {
+            leaf: Some(typedb_protocol::concept_document::node::leaf::Leaf::Kind(encode_kind(kind).into())),
+        }),
```
krishnangovindraj, line 179: Please check

storage/isolation_manager.rs
```diff
@@ -178,8 +178,10 @@ impl IsolationManager {
                 if let Some(conflict) = handle_dependency(commit_dependency) {
                     return Ok(Some(conflict));
                 }
+            } else if let Err(err) = commit_status_result {
+                return Err(err);
             } else {
-                todo!()
+                unreachable!()
```
krishnangovindraj, line 184: Please review.

common/error/error.rs
```diff
@@ -219,3 +219,59 @@ macro_rules! typedb_error {
         None
     };
 }
+
```
krishnangovindraj, line 222: General strategy:
We fail translation with a custom Error (with a variant name `UnimplementedFeature`, or something similar).
Any place which is unreachable thanks to this error, we use a `unimplemented_feature!` macro.

krishnangovindraj (commented):
Self review

dmitrii-ubskii (approved):
Approved with a couple small change requests.

ir/pattern/disjunction.rs
```diff
@@ -56,7 +50,12 @@ impl StructuralEquality for Disjunction {
 
 impl fmt::Display for Disjunction {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        todo!()
+        debug_assert!(self.conjunctions.len() > 0);
+        write!(f, "{}", self.conjunctions[0])?;
+        for i in 1..self.conjunctions.len() {
+            write!(f, " or {}", self.conjunctions[i])?;
+        }
+        Ok(())
```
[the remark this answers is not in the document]
krishnangovindraj: Unsure. Doesn't seem to be done for other constraints either.

PR #7318, opened 2025-01-17
Fix TypeDB reporting crashes when CA certificates are not found

farost:
We fix TypeDB reporting crashes when CA certificates are not found on the host. An additional Sentry (our crash reports endpoint) warning will be reported, but it no longer affects the server's availability. 

## Implementation
If reporting is not successful for either endpoints, a warning/error (based on the cause) to Sentry will be reported, and the reporting for this endpoint will stop. For simplicity, the job will be still activated every 1 hour, but it will quit after a boolean flag check.

The next real attempt to report anything will happen only after a restart.

The logic described above works for two endpoints separately. This way, if only one endpoint fails (which should not be the case now), the other one continues working as expected, reporting data every hour.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -888,6 +888,7 @@ dependencies = [
  "hyper-rustls",
  "logger",
  "resource",
+ "sentry",
  "serde_json",
  "sysinfo",
  "tokio",
@@ -4164,7 +4165,7 @@ checksum = "42ff0bf0c66b8238c6f3b578df37d0b7848e55df8577b3f74f92a69acceeb825"
 [[package]]
 name = "typeql"
 version = "0.0.0"
-source = "git+https://github.com/typedb/typeql?rev=059c0332e33a892e6a04bbf83df5634d3d5df249#059c0332e33a892e6a04bbf83df5634d3d5df249"
+source = "git+https://github.com/typedb/typeql?tag=3.0.1#bb23f987091855a6ac56081206b9d3a736a1a9d5"
 dependencies = [
  "chrono",
  "itertools 0.10.5",
--- a/RELEASE_TEMPLATE.md
+++ b/RELEASE_TEMPLATE.md
@@ -4,6 +4,6 @@
 
 **Pull the Docker image:**
 
-```docker pull vaticle/typedb:{version}```
+```docker pull typedb/typedb:{version}```
 
 { release notes }
--- a/diagnostics/reporter.rs
+++ b/diagnostics/reporter.rs
@@ -270,13 +270,12 @@ impl Reporter {
     }
 
     fn report_inner_error(error: DiagnosticsReporterError) {
-        let error_string = <dyn TypeDBError>::to_string(&error).as_str();
-        match error.as_ref() {
+        match error {
             DiagnosticsReporterError::HttpsClientBuilding { .. } => {
-                Self::report_inner_error_message_warning(error_string)
+                Self::report_inner_error_message_warning(<dyn TypeDBError>::to_string(&error).as_str())
             }
             DiagnosticsReporterError::HttpRequestBuilding { .. } => {
-                Self::report_inner_error_message_critical(error_string)
+                Self::report_inner_error_message_critical(<dyn TypeDBError>::to_string(&error).as_str())
             }
         }
     }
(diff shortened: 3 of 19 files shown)


RELEASE_TEMPLATE.md
```diff
@@ -4,6 +4,6 @@
 
 **Pull the Docker image:**
 
-```docker pull vaticle/typedb:{version}```
+```docker pull typedb/typedb:{version}```
```
farost, line 7: !

Cargo.lock
farost, line 1: Updated dependencies to the most recent ones because I could. Decided to push it as the PR is small, and it's not meshed with anything. But can stop doing it if you don't like it! Please let me know

diagnostics/reporter.rs
```diff
@@ -213,6 +268,25 @@ impl Reporter {
         };
         Duration::from_secs(delay_secs)
     }
+
+    fn report_inner_error(error: DiagnosticsReporterError) {
+        match error {
+            DiagnosticsReporterError::HttpsClientBuilding { .. } => {
+                Self::report_inner_error_message_warning(<dyn TypeDBError>::to_string(&error).as_str())
```
farost, line 270: It is a warning because it is caused by user's environment.

diagnostics/reporter.rs
```diff
@@ -213,6 +268,25 @@ impl Reporter {
...
+            }
+            DiagnosticsReporterError::HttpRequestBuilding { .. } => {
+                Self::report_inner_error_message_critical(<dyn TypeDBError>::to_string(&error).as_str())
```
farost, line 273: It is an error because we don't want to crash the server because of a reporter bug, but we want to see this in reports. We don't test reporters in CI (which is a problem: should fix it by some cool tests somehow), so we at least have this.

Cargo.lock
farost: P.s. I think this should've been done as a part of the release. As we want to support cargo builds, we better update dependencies and run sync locally as a release person

diagnostics/reporter.rs
```diff
@@ -213,6 +263,25 @@ impl Reporter {
...
+        }
+    }
+
+    fn report_inner_error_message_critical(message: &str) {
+        sentry::capture_message(message, sentry::Level::Error);
+    }
+
+    fn report_inner_error_message_warning(message: &str) {
+        sentry::capture_message(message, sentry::Level::Warning);
+    }
```
dmitrii-ubskii, line 284: Is this indirection necessary?

diagnostics/reporter.rs
```diff
@@ -229,3 +298,10 @@ impl ReportingEndpoint {
         }
     }
 }
+
+typedb_error!(
+    pub DiagnosticsReporterError(component = "DiagnosticsReporter", prefix = "DIR") {
+        HttpsClientBuilding(1, "Error while building a diagnostics reporting https client.", source: Arc<std::io::Error>),
+        HttpRequestBuilding(2, "Error while building a diagnostics reporting http request.", source: Arc<http::Error>),
```
dmitrii-ubskii, line 305: Why the `Arc`s?

diagnostics/reporter.rs
```diff
@@ -145,17 +195,17 @@ impl Reporter {
...
+    async fn send_request(body: String, uri: &'static str) -> Result<bool, DiagnosticsReporterError> {
         let request = Request::post(uri)
             .header(CONTENT_TYPE, "application/json")
             .header(CONNECTION, "close")
             .body(Body::from(body))
-            .expect("Failed to construct the request");
+            .map_err(|source| DiagnosticsReporterError::HttpRequestBuilding { source: Arc::new(source) })?;
 
-        match Self::new_https_client().request(request).await {
+        Ok(match Self::new_https_client()?.request(request).await {
```
dmitrii-ubskii, line 208: I'd much rather see this `Ok()` lowered to the actual return values.

diagnostics/reporter.rs
dmitrii-ubskii: Can we ever get an HTTP error? Both our endpoints are HTTPS.

dmitrii-ubskii (approved):
Approved with some comments.

diagnostics/reporter.rs
farost: Well, by `Request` I meant a general HTTP entity that contains a body, metadata, etc. So it's called in general `Http`, not `-s`

diagnostics/reporter.rs
farost: It's better to hide the internals of `sentry` from my point of view. Like how's `critical` represented there, how's `warning` represented there. Not necessary, but I'd leave it as it's nice.

diagnostics/reporter.rs
farost: To have `Clone` for `typedb_error!`. Discussed in person

diagnostics/reporter.rs
farost: Okay, done

PR #7320, opened 2025-01-20
Re-enable behaviour tests

krishnangovindraj:
Enables tests for various read & write behaviour.  

## Implementation
* Introduces test files for each disabled feature, runs them, and either
  - fixes bugs that fall out; or
  - ignores cases for bigger known bugs / features that aren't supported yet. 
* Enables these BDD tests in CI

### DIFF
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -133,7 +133,7 @@ build:
         bazel test //tests/behaviour/query/language:test_insert --test_output=streamed
 #        bazel test //tests/behaviour/query/language:test_delete --test_output=streamed
 #        bazel test //tests/behaviour/query/language:test_update --test_output=streamed
-#        bazel test //tests/behaviour/query/language:test_pipelines --test_output=streamed
+        bazel test //tests/behaviour/query/language:test_pipelines --test_output=streamed
 
     test-behaviour-query-definable:
       image: typedb-ubuntu-22.04
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -132,7 +132,7 @@ build:
         bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
         bazel test //tests/behaviour/query/language:test_insert --test_output=streamed
         bazel test //tests/behaviour/query/language:test_delete --test_output=streamed
-#        bazel test //tests/behaviour/query/language:test_update --test_output=streamed
+        bazel test //tests/behaviour/query/language:test_update --test_output=streamed
         bazel test //tests/behaviour/query/language:test_pipelines --test_output=streamed
 
     test-behaviour-query-definable:
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -121,7 +121,7 @@ build:
         bazel test //tests/behaviour/query/language:test_match --test_output=streamed
         bazel test //tests/behaviour/query/language:test_fetch --test_output=streamed
         bazel test //tests/behaviour/query/language:test_reduce --test_output=streamed
-#        bazel test //tests/behaviour/query/language:test_expressions --test_output=streamed
+        bazel test //tests/behaviour/query/language:test_expressions --test_output=streamed
 #        bazel test //tests/behaviour/query/language:test_modifiers --test_output=streamed
 #        bazel test //tests/behaviour/query/language:test_get --test_output=streamed
 
@@ -131,7 +131,7 @@ build:
       command: |
         bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
         bazel test //tests/behaviour/query/language:test_insert --test_output=streamed
-#        bazel test //tests/behaviour/query/language:test_delete --test_output=streamed
+        bazel test //tests/behaviour/query/language:test_delete --test_output=streamed
 #        bazel test //tests/behaviour/query/language:test_update --test_output=streamed
         bazel test //tests/behaviour/query/language:test_pipelines --test_output=streamed
 
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -122,7 +122,7 @@ build:
         bazel test //tests/behaviour/query/language:test_fetch --test_output=streamed
         bazel test //tests/behaviour/query/language:test_reduce --test_output=streamed
         bazel test //tests/behaviour/query/language:test_expressions --test_output=streamed
-#        bazel test //tests/behaviour/query/language:test_modifiers --test_output=streamed
+        bazel test //tests/behaviour/query/language:test_modifiers --test_output=streamed
 #        bazel test //tests/behaviour/query/language:test_get --test_output=streamed
 
     test-behaviour-query-write:
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -125,12 +125,12 @@ build:
 #        bazel test //tests/behaviour/query/language:test_modifiers --test_output=streamed
 #        bazel test //tests/behaviour/query/language:test_get --test_output=streamed
 
-#    test-behaviour-query-write:
-#      image: typedb-ubuntu-22.04
-#      dependencies: [build]
-#      command: |
-#        bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
-#        bazel test //tests/behaviour/query/language:test_insert --test_output=streamed
+    test-behaviour-query-write:
+      image: typedb-ubuntu-22.04
+      dependencies: [build]
+      command: |
+        bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
+        bazel test //tests/behaviour/query/language:test_insert --test_output=streamed
 #        bazel test //tests/behaviour/query/language:test_delete --test_output=streamed
 #        bazel test //tests/behaviour/query/language:test_update --test_output=streamed
 #        bazel test //tests/behaviour/query/language:test_pipelines --test_output=streamed
--- a/.factory/automation.yml
+++ b/.factory/automation.yml
@@ -123,7 +123,6 @@ build:
         bazel test //tests/behaviour/query/language:test_reduce --test_output=streamed
         bazel test //tests/behaviour/query/language:test_expressions --test_output=streamed
         bazel test //tests/behaviour/query/language:test_modifiers --test_output=streamed
-#        bazel test //tests/behaviour/query/language:test_get --test_output=streamed
 
     test-behaviour-query-write:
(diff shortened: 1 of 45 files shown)


compiler/annotation/pipeline.rs
```diff
@@ -290,6 +290,18 @@ fn annotate_stage(
...
+                        if !running_variable_annotations.contains_key(&variable)
+                            && insert_annotations.vertex_annotations_of(links.role_type()).is_some()
+                        {
+                            running_variable_annotations.insert(
+                                variable,
+                                insert_annotations.vertex_annotations_of(links.role_type()).unwrap().clone(),
+                            );
+                        }
+                    }
+                }
```
krishnangovindraj, line 304: Vertices corresponding to the role may also be freshly added by the insert clause.

.factory/automation.yml
```diff
@@ -121,19 +121,19 @@ build:
         bazel test //tests/behaviour/query/language:test_match --test_output=streamed
         bazel test //tests/behaviour/query/language:test_fetch --test_output=streamed
         bazel test //tests/behaviour/query/language:test_reduce --test_output=streamed
-#        bazel test //tests/behaviour/query/language:test_expressions --test_output=streamed
-#        bazel test //tests/behaviour/query/language:test_modifiers --test_output=streamed
+        bazel test //tests/behaviour/query/language:test_expressions --test_output=streamed
+        bazel test //tests/behaviour/query/language:test_modifiers --test_output=streamed
 #        bazel test //tests/behaviour/query/language:test_get --test_output=streamed
```
farost, line 126: What are we going to do with the get tests?

.factory/automation.yml
```diff
@@ -121,19 +121,19 @@ build:
...
-#    test-behaviour-query-write:
-#      image: typedb-ubuntu-22.04
-#      dependencies: [build]
-#      command: |
-#        bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
-#        bazel test //tests/behaviour/query/language:test_insert --test_output=streamed
-#        bazel test //tests/behaviour/query/language:test_delete --test_output=streamed
-#        bazel test //tests/behaviour/query/language:test_update --test_output=streamed
-#        bazel test //tests/behaviour/query/language:test_pipelines --test_output=streamed
+    test-behaviour-query-write:
```
farost, line 128: Ooofff so satisfying

compiler/annotation/expression/expression_compiler.rs
```diff
@@ -234,6 +234,25 @@ impl<'this> ExpressionCompilationContext<'this> {
...
+                }
+                Operator::Subtract => {
+                    CastLeftIntegerToDecimal::validate_and_append(self)?;
+                    operators::OpDecimalSubtractDecimal::validate_and_append(self)?;
+                }
+                Operator::Multiply => {
+                    CastLeftIntegerToDecimal::validate_and_append(self)?;
+                    operators::OpDecimalMultiplyDecimal::validate_and_append(self)?;
+                }
+                other_op => {
```
farost, line 250: Because we consider these other operations lossy enough and don't actually worry? Or why are other operations so special?

compiler/annotation/mod.rs
```diff
@@ -168,6 +170,11 @@ typedb_error!(
             label: String
         ),
         ValueTypeNotFound(8, "Value type '{name}' was not found.", name: String),
+        AnnotationsUnavailableForVariableInInsert(9,
+            "The annotations for the variable '{variable}' in the inserted constraint {constraint} were not available. Check if the variable is available from a previous stage or is inserted in this stage.",
```
farost, line 174: I wonder if these inner messages can freak the users out by usage of **annotation** and **constraint** in a totally different field D: (:D)
Not a reason to change anything (now..)

compiler/executable/insert/type_check.rs
```diff
@@ -125,11 +135,29 @@ fn validate_links_insertable(
...
     left_right_filtered: &LinksAnnotations,
 ) -> Result<(), TypeInferenceError> {
+    // TODO: Should we check uniqueness of inferred role-types here instead of at compilation?
     // TODO: Improve. This is extremely coarse and likely to rule out many valid combinations
     // Esp when doing queries using type variables.
-    let input_relation_types = input_annotations_variables.get(&links.relation().as_variable().unwrap()).unwrap();
-    let input_player_types = input_annotations_variables.get(&links.player().as_variable().unwrap()).unwrap();
-    let input_role_types = input_annotations_variables.get(&links.role_type().as_variable().unwrap()).unwrap();
+
+    // TODO: These should be unreachable
```
farost, line 142: Wdym?

ir/lib.rs
```diff
@@ -193,6 +193,10 @@ typedb_error! {
             lhs_category: VariableCategory,
             rhs_category: VariableCategory,
         ),
+        DeleteVariableUnavailable(40,
+            "The variable '{variable}' referenced in the delete stage is unavailable:",
```
farost, line 197: Why is there a colon `:` at the end?

ir/tests/structural_equality.rs
```diff
@@ -251,7 +251,6 @@ match
   $o links (customer: $c), isa ORDER, has O_ID 0, has O_NEW_ORDER $o_new_order;
 delete
   has $o_new_order of $o;
-  has $o_carrier_id of $o;
```
farost, line 254: Why?

ir/translation/constraints.rs
```diff
@@ -400,7 +400,6 @@ fn add_typeql_isa(
                 constraints.add_comparison(Vertex::Variable(thing), assigned_to, Comparator::Equal)?;
             }
             IsaInstanceConstraint::Comparison(comparison) => {
-                add_typeql_isa(function_index, constraints, thing, isa)?;
```
farost, line 403: Does it solve a specific bug?

tests/behaviour/steps/params.rs
```diff
@@ -505,10 +505,15 @@ impl Value {
             TypeDBValueType::Integer => TypeDBValue::Integer(self.raw_value.parse().unwrap()),
             TypeDBValueType::Double => TypeDBValue::Double(self.raw_value.parse().unwrap()),
             TypeDBValueType::Decimal => {
-                let (integer, fractional) = if let Some(split) = self.raw_value.split_once(".") {
+                let trimmed_value = if self.raw_value.ends_with("dec") {
```
farost, line 508: Good

farost (commented):
Some questions!

krishnangovindraj (commented):
Forgot to publish self-review comments

ir/translation/constraints.rs
krishnangovindraj: Infinite recursion. This is just a call to the function itself.

.factory/automation.yml
krishnangovindraj: Good flag again.  I dissolved them here: https://github.com/typedb/typedb-behaviour/commit/8224557260e681845f2a1fea6854ef77e80241eb

compiler/annotation/expression/expression_compiler.rs
krishnangovindraj: They're the only ones @dmitrii-ubskii has implemented for Decimals >.<
But yes. Our integer division results in a double, so I figured this makes sense too. It'll be hard to keep the decimal-level precision for those operations (except maybe mod?)

compiler/executable/insert/type_check.rs
krishnangovindraj: Updated. I should have said "These should be safe to unwrap"

ir/tests/structural_equality.rs
krishnangovindraj: Unreferenced in the match, so the parsing failed.

compiler/annotation/mod.rs
farost: I know that we use constraints and annotations in many places in compilers, so it's fine, but maybe there is a keyword we could drop in this sentence to isolate the context from user-facing annotations. I wouldn't totally rename it unless we understand what's going on when users send us this error

ir/translation/constraints.rs
farost: 3.inf

compiler/executable/insert/type_check.rs
```diff
@@ -125,11 +135,29 @@ fn validate_links_insertable(
...
+            variable: links.player().as_variable().unwrap(),
+            constraint: Constraint::Links(links.clone()),
+        });
+    };
+    let Some(input_role_types) = input_annotations_variables.get(&links.role_type().as_variable().unwrap()) else {
+        return Err(TypeInferenceError::AnnotationsUnavailableForVariableInInsert {
+            variable: links.role_type().as_variable().unwrap(),
+            constraint: Constraint::Links(links.clone()),
+        });
+    };
```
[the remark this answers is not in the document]
farost: If you are sure that we should not appear here, you can unwrap / expect. If you are not, then an error is always better, especially since we already have this infrastructure

I'd also think about a user error classification, so users can differentiate their mistakes from our non-critical bugs like "I can't perform this operation because... unexpected ... Please report to us"

compiler/executable/insert/type_check.rs
krishnangovindraj: Yeah, but we get that wrong sometimes when something changes and someone violates a precondition.

compiler/annotation/mod.rs
```diff
@@ -168,6 +170,11 @@ typedb_error!(
             label: String
         ),
         ValueTypeNotFound(8, "Value type '{name}' was not found.", name: String),
+        AnnotationsUnavailableForVariableInInsert(9,
```
flyingsilverfin, line 173: @krishnangovindraj  notice how every message here doesn't use `Annotations` but instead uses `Types` - `TypesUnavailableForVariableInInsert` would be more consistent

compiler/executable/insert/type_check.rs
```diff
@@ -82,8 +82,19 @@ fn validate_has_insertable(
 ) -> Result<(), TypeInferenceError> {
     // TODO: Improve. This is extremely coarse and likely to rule out many valid combinations
     // Esp when doing queries using type variables.
-    let input_owner_types = input_annotations_variables.get(&has.owner().as_variable().unwrap()).unwrap();
-    let input_attr_types = input_annotations_variables.get(&has.attribute().as_variable().unwrap()).unwrap();
+    // TODO: These should be safe to unwrap
```
flyingsilverfin, line 85: Why is this a TODO?

PR #7327, opened 2025-01-24
Enhance error messages from compiler

farost:
Add error stack traces for write compilation errors. Fix undefine error messages for inherited traits for better readability.

## Implementation
Updated only a limited amount of errors so as not to conflict with https://github.com/typedb/typedb/pull/7312. I've seen users sharing the write compilation errors with us without stack traces, and it makes no sense without them. The undefine errors are just the ones that I've encountered now, it was certainly broken.


compiler/executable/insert/mod.rs
```diff
@@ -57,28 +58,20 @@ pub(crate) fn get_kinds_from_annotations(annotations: &BTreeSet<Type>) -> HashSe
...
-    IllegalInsertForRole { isa: Isa<Variable> },
-    DeletedThingWasNotInInput { variable: Variable },
-}
-
-impl fmt::Display for WriteCompilationError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        error::todo_display_for_error!(f, self)
+typedb_error! {
+    pub WriteCompilationError(component = "Write Compilation", prefix = "WRC") {
+        // TODO: Update error message to something more understandable
```
farost, line 63: I don't really understand these errors. But it's something.

krishnangovindraj: `IsaStatementForInputVariable`: It's basically the re-insert problem.

farost: I see now. Will change the error message to highlight it. Thanks

farost:
Will become a part of another PR (I wanted to push it to the closest release, didn't manage to)

PR #7329, opened 2025-01-27
Resolve ambiguity in duration arithmetic

dmitrii-ubskii:

We stabilize duration arithmetic with datetimes with time zones in cases where the resulting datetime is ambiguous or does not exist in the target time zone.

## Motivation

When adding a date duration to a datetime-tz (i.e. years, months, weeks, or days), the resulting combination of date and time may not exist in the target time zone. When that happened, the arithmetic operation would panic and the server would crash.

## Implementation

We adopt the Java standard library approach:
1. perform date arithmetic on local date time without taking into account time zone rules,
2. if the datetime exists in the time zone and is unambiguous, skip to step 5,
3. if the datetime is ambiguous in the time zone, take the earliest possible datetime and skip to step 5,
4. if the datetime does not exist, i.e. there's a gap in the time zone, advance it by the length of that gap (generally 1 hour for DST),
5. perform duration arithmetic on the datetime-tz using the time component of the duration (this can never fail as it modifies the underlying UTC timestamp directly).

### DIFF
--- a/encoding/value/duration_value.rs
+++ b/encoding/value/duration_value.rs
@@ -10,7 +10,7 @@ use std::{
     str::FromStr,
 };
 
-use chrono::{DateTime, Days, Months, NaiveDateTime, TimeDelta, TimeZone};
+use chrono::{DateTime, Days, MappedLocalTime, Months, NaiveDateTime, Offset, TimeDelta, TimeZone};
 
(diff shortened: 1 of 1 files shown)


flyingsilverfin (approved):

flyingsilverfin:
Looks good, great!! Love java spec, so thorough :D

PR #7330, opened 2025-01-27
Fix relates double specialization

farost:
Fix the behavior of `relates` specialization, featuring:
* Unblocked double specialization:
```
define 
  relation family-relation relates member @abstract;
  relation parentship sub family relation, relates parent as member, relates child as member; # Good!
```  
* Fixed validations for multi-layered specializations:
```
define
  relation family-relation relates member @abstract, relates pet @abstract;
  relation parentship sub family relation, relates parent as member;
  relation fathership sub parentship, relates father as member; # Bad!
  relation fathership sub parentship, relates fathers-dog as pet; # Good!
```
* Better definition resolution and error messaging.
* Change the inner terminology for generated `relates` for specialization: "root" and "non specializing" are substituted by "explicit", and "specializing" is substituted by "implicit".

## Motivation
We fix https://github.com/typedb/typedb/issues/7322 and connected issues.

## Implementation
Two steps:
**1.** 
The definable resolution was not correctly implemented for capabilities, specifically for specialized relates. The new logic is the following:
* While we work with concept API, we **almost** always want to get all capabilities, and the specialized capabilities should be returned with the `@abstract` annotation. Thus, we do trust regular Concept API methods everywhere.
* However, (re/un/)definitions are different. They need to find the actually defined capabilities, not the ones being generated implicitly. This way, a new set of methods is needed for these `non_specialised` searches (decided to be explicit with the naming). Moreover, they need to be straightforward, searching for a definition that a user would think about. 

With this simplified (the previous logic was quite complicated, at least its reasoning, thus error-prone) search, the first part of the problem has been solved. Now, define and redefine always check if they have a **non-specialized** relates declared for them. Undefine uses "transitive" search and searches for the closest non-specialized relates when a role type is mentioned. 

**2.**
The validations for specializations were not quite correct. They could still allow some invalid usages of Concept API. Now, we check 2 conditions in multiple steps:
* The specialized `relates` should be had by my supertype (it was a similar check, but generally for a role type, so it didn't quite work).
* The specialized `relates` should not be specializing (as before).

P.S. Sadly, the current state of Concept BDDs with a huge amount of tests is not enough for multiple reasons (the first one is simply the additional logic in definitions that can be buggy + a need to produce more and more concept apis). I have some thoughts in my head which could help us bring the old idea of auto-conversion to queries in reality (probably only for schema queries: it will already be enough), but it will take at least a weak to refactor and implement it. So we'll see... In a year... Or more...


.factory/automation.yml
```diff
@@ -131,7 +132,7 @@ build:
         bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
         bazel test //tests/behaviour/query/language:test_insert --test_output=streamed
         bazel test //tests/behaviour/query/language:test_delete --test_output=streamed
-        bazel test //tests/behaviour/query/language:test_update --test_output=streamed
+        # TODO: Place when fixed => bazel test //tests/behaviour/query/language:test_update --test_output=streamed
```
farost, line 135: Rerunning tests takes too much time.

query/undefine.rs
```diff
@@ -654,13 +654,17 @@ fn undefine_type_capability_relates(
         return Err(err_unsupported_capability(type_label, type_.kind(), capability_undefinable));
     };
 
+    // Not Transitive: when deleting owns or plays, we "unset" it, and the TypeManager knows how to
```
farost, line 657: I could potentially make all capabilities be searched through `Declared`, but I like the descriptive and multi-layered error that we get from Concept API (which we can't replicate here because we don't unset `relates` but delete `role`)

dependencies/typedb/repositories.bzl
```diff
@@ -36,5 +36,5 @@ def typedb_behaviour():
     git_repository(
         name = "typedb_behaviour",
         remote = "https://github.com/typedb/typedb-behaviour",
-        commit = "1cef4b5af6599861fb240b722a66dc4383846554",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        commit = "032d0ee88c3737b7d7ed9f921908422dd585c6e8",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
```
farost, line 40: Updated BDDs: https://github.com/typedb/typedb-behaviour/pull/327

concept/type_/type_manager/validation/mod.rs
```diff
@@ -32,13 +32,13 @@ typedb_error!(
...
         ChangingAttributeTypeSupertypeWillLeadToConflictingValueTypes(9, "Changing supertype of attribute type '{attribute}' will lead to conflicting value types between '{value_type:?}' and '{new_super_value_type}'.", attribute: Label, value_type: Option<ValueType>, new_super_value_type: ValueType),
         CannotUnsetAbstractnessOfAttributeTypeAsItHasSubtypes(10, "Cannot unset the abstractness of attribute type '{attribute}' as it has subtypes.", attribute: Label),
         CannotDeleteTypeWithExistingSubtypes(11, "Cannot delete type '{label}' as it has existing subtypes.", label: Label),
-        RelatesNotInherited(12, "Relation type '{relation}' does not inherit role '{role}'.", relation: Label, role: Label),
-        RelatesIsAlreadySpecialisedByASupertype(13, "Relation type '{relation}' is already specialised by a supertype for '{role}'.", relation: Label, role: Label),
+        RelatesNotInheritedForSpecialisation(12, "Role type '{role}' should be inherited without specialization by supertypes to be specialized by relation type '{relation}'.", relation: Label, role: Label),
```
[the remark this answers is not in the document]
dmitrii-ubskii: "role is already specialized by a supertype" is a separate error we already have. This should just be:
```suggestion
        RelatesNotInheritedForSpecialisation(12, "Role type '{role}' must be inherited to be specialized by relation type '{relation}'.", relation: Label, role: Label),
```
Ideally we'd also mention which relation type should be the supertype for the role to be inherited.

concept/type_/type_manager/validation/mod.rs
```diff
@@ -62,9 +62,9 @@ typedb_error!(
...
-        CannotUnsetInheritedOwns(37, "Cannot unset inherited 'owns' constraint from '{supertype}' for '{supertype}'.", supertype: Label, subtype: Label),
-        CannotUnsetInheritedPlays(38, "Cannot unset inherited 'plays' constraint from '{supertype}' for '{subtype}'.", supertype: Label, subtype: Label),
-        CannotUnsetInheritedValueType(39, "Cannot unset inherited value type '{value_type}' for '{subtype}'.", value_type: ValueType, subtype: Label),
+        CannotUnsetInheritedOwns(37, "Cannot unset 'owns {attribute_type}' constraint for '{subtype}' inherited from '{supertype}'.", subtype: Label, supertype: Label, attribute_type: Label),
+        CannotUnsetInheritedPlays(38, "Cannot unset 'plays {role_type}' constraint for '{subtype}' inherited from '{supertype}'.", subtype: Label, supertype: Label, role_type: Label),
+        CannotUnsetInheritedValueType(39, "Cannot unset value type '{value_type}' for '{subtype}' inherited from '{supertype}'.", subtype: Label, supertype: Label, value_type: ValueType),
```
dmitrii-ubskii, line 67: 👍

dmitrii-ubskii (approved):
Minor comments

concept/type_/type_manager/validation/mod.rs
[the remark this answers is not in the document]
farost: That's what I meant in my comment. We already have the specialization-related error, but this error is more common, and without the context of specialization it is not that clear. That's why I introduced the mention of it here. Otherwise, a user could think "but the type already inherits this role, it's just specialized from above! Strange!"

concept/type_/relation_type.rs
```diff
@@ -455,9 +454,66 @@ impl RelationType {
...
+    ) -> Result<Option<Relates>, Box<ConceptReadError>> {
+        if let Some(relates) = self.get_relates_role_declared(snapshot, type_manager, role_type.clone())? {
+            if !relates.is_implicit(snapshot, type_manager)? {
+                return Ok(Some(relates));
+            }
+        }
+        Ok(None)
+    }
+
+    pub fn get_relates_role_explicit(
```
farost, line 471: I still have the non-declared explicit versions for other places in definition resolution. E.g., when we define
```
define
parentship relates parent;
fathership relates father as parent;
```

When we say
```
fathership relates parent;
```
^ this should return an error. However, if we don't search for **explicit** relates here, we'd find an implicit relates, its definition status would be "just like the user defined!", and the query would do nothing, saying "OK!". And the user won't even understand that his request is invalid.

Forgot to mention it in person. However, I do believe that the usage of implicit / explicit already makes everything much cleaner and streamlined.

I just don't like the duplications for "_role_" and "_role_name_", would be cool to find a way to generalize it (if needed).

flyingsilverfin (approved):

PR #7331, opened 2025-01-29
Remove cardinalities operation time validation

farost:
Remove cardinalities operation time validation. Now, both schema and data modifications lead to cardinalities revalidations only on commits.
This fix solves https://github.com/typedb/typedb/issues/7317.

## Motivation
Previously, cardinalities were checked on operation time for schema modifications, while the data modifications only verified it on commit time. This inconsistency led to less smooth user perception of expected validations and possible actions, as well as blocked a couple of complicated and rare cases of schema modifications. 

The main issue discovered was a combination of the `define` query logic and the need to revalidate every Concept API call affecting cardinalities. When we declared a new `owns name @card(X..Y)`, two cardinality validations happened on operation time: a validation of the default cardinality for `owns name` and a validation of the `@card(X..Y)` cardinality. However, the restrictive default cardinality (which `@card(0..1)` for owns) would block this schema mutation if a subtype instance of this type already had multiple instances of `name`s (or subnames). 

The issue described above could be resolved by skipping the default validation if a cardinality is defined. However, this would require us to consider multiple changes:
* Changing the architecture of `define`, allowing it to look at the annotations while declaring capabilities, and considering queries like `define person owns name; person owns name @card(...); person owns name; person owns name;` -- quite a complication
* Passing the cardinality into the `set_owns` API call, making it less granular.
* Or introducing another signal and, thus, state to Concept API to call for cardinality validations

Overall, combining it with the asymmetry of schema/data validations and its possible confusion for the users, we decided to disable the operation time validations for cardinalities, which is both easier and more natural for the end user.

## Implementation
Introduce a new parameter "is operation time validated" for constraints and add the respective filtering for constraints operation time validations in Type Manager.
Extend the list of features to look at on Thing Manager's commit validations when constructing a list of affected instances for cardinality checks. Now, we control all objects subtyping changes, interfaces subtyping changes, traits changes, and annotations changes. See code comments for more information.


concept/type_/constraint.rs
```diff
@@ -146,11 +146,19 @@ impl ConstraintDescription {
...
-            ConstraintDescription::Cardinality(cardinality) => cardinality == &AnnotationCardinality::unchecked(),
-            ConstraintDescription::Independent(_) => true,
-            _ => false,
+            ConstraintDescription::Cardinality(cardinality) => cardinality.is_checked(),
+            ConstraintDescription::Independent(_) => false,
+            _ => true,
+        }
+    }
+
+    pub(crate) fn is_operation_time_checked(&self) -> bool {
```
farost, line 157: Constraints are tightly connected with type / thing managers and validations, so I decided that introducing the "operation time" term here is not a bad idea. It even lets us check what's the logic for each specific constraint right in a single model file.

dependencies/typedb/repositories.bzl
```diff
@@ -33,8 +33,9 @@ def typedb_protocol():
     )
 
 def typedb_behaviour():
+    # TODO: Return typedb
```
farost, line 36: https://github.com/typedb/typedb-behaviour/pull/330

common/iterator/mod.rs
```diff
@@ -39,3 +39,14 @@ impl<T: Hash + Eq + PartialEq> Collector<T> for HashSet<T> {
         self.insert(element);
     }
 }
+
+#[macro_export]
+macro_rules! minmax_or {
```
farost, line 44: A small refactoring for reusability.

concept/type_/type_manager/validation/operation_time_validation.rs
```diff
@@ -250,7 +250,7 @@ macro_rules! new_acquired_capability_instances_validation {
                 &affected_object_types,
                 &affected_interface_types,
                 &HashMap::new(), // all interface supertypes are read from storage
-                &get_checked_constraints(default_constraints.into_iter()),
+                &get_operation_time_checked_constraints(default_constraints.into_iter()),
```
farost, line 253: One of a couple of places where I needed to add filtering to disable operation-time validation. With this, other constraints are still validated at operation time (including the default values, say, for `distinct`!). So no other places require changes. At the same time, if no constraints are affected, we won't spend time iterating over instances here.

concept/thing/thing_manager.rs
```diff
@@ -184,6 +217,66 @@ impl ThingManager {
...
+        self.get_thing_in_range_unordered(
+            snapshot,
+            object_type_range,
+            Prefix::object_type_range_inclusive(),
+            <Object as ThingAPI>::Vertex::FIXED_WIDTH_ENCODING,
+            <Object as ThingAPI>::Vertex::KEYSPACE,
+        )
+    }
+
+    fn get_thing_in_range_unordered<T: ThingAPI>(
```
farost, line 234: Replica of `get_has_from_thing_to_type_unordered`, but for instances

concept/thing/thing_manager.rs
```diff
@@ -184,6 +217,66 @@ impl ThingManager {
...
+        &self,
+        snapshot: &impl ReadableSnapshot,
+        type_range: &impl RangeBounds<T::TypeAPI>,
+        unbounded_prefix_range: (Prefix, Prefix),
+        fixed_width_encoding: bool, // TODO: Can be generalized for ThingAPI?
```
farost, line 239: Haven't touched anything globally regarding constants for impls. Can do if we agree that these two constants should be initialized for all `ThingAPI` impls.

concept/thing/thing_manager.rs
```diff
@@ -1773,6 +1866,42 @@ impl ThingManager {
     }
 
     fn validate(&self, snapshot: &mut impl WritableSnapshot) -> Result<(), Vec<ConceptWriteError>> {
+        /*
+        The cardinalities validation flow is the following:
```
farost, line 1870: I spent quite some time reconstructing the whole validation pipeline and got stuck with some complicated cases for half a day, so I decided that it is be easier if I paste these "docs" just here: the expected understanding speed up is x20 lol

I kinda like my current architecture of validations, so I haven't refactored it but highlighted its strong sides (it goes up and then down the inheritance line) and weaknesses (it requires careful affected data preparation). So I put the logic behind this here to hint at why the affected data is prepared in such a scary way.

concept/thing/thing_manager.rs
```diff
@@ -1786,10 +1915,23 @@ impl ThingManager {
...
-            self.collect_modified_links(snapshot, &mut modified_relations_role_types, &mut modified_objects_role_types);
+
+        res = self.collect_modified_links_objects(
+            snapshot,
+            &mut modified_relations_role_types,
+            &mut modified_objects_role_types,
+        );
+        collect_errors!(errors, res, |source| DataValidationError::ConceptRead { source });
+
+        res = self.collect_modified_schema_capability_cardinalities_objects(
```
farost, line 1929: The new subset of affected objects and their capabilities: modified schema.

concept/thing/thing_manager.rs
```diff
@@ -1918,6 +2060,312 @@ impl ThingManager {
...
+        &self,
+        snapshot: &impl WritableSnapshot,
+        out_object_attribute_types: &mut HashMap<Object, HashSet<AttributeType>>,
+        out_object_role_types: &mut HashMap<Object, HashSet<RoleType>>,
+        out_relation_role_types: &mut HashMap<Relation, HashSet<RoleType>>,
+    ) -> Result<(), Box<ConceptReadError>> {
+        let mut modified_owns = HashMap::new();
+        let mut modified_plays = HashMap::new();
+        let mut modified_relates = HashMap::new();
+        self.collect_modified_schema_capability_cardinalities(
```
farost, line 2073: Collect all affected capabilities here. Inside, there is a bunch of scary code working with snapshots.

concept/thing/thing_manager.rs
```diff
@@ -1918,6 +2060,312 @@ impl ThingManager {
...
+            snapshot,
+            &mut modified_owns,
+            &mut modified_plays,
+            &mut modified_relates,
+        )?;
+
+        for (relation_type, role_types) in modified_relates {
```
farost, line 2080: It's more readable if I leave it this way rather than create yet another scary macro. At least, I thought so.

concept/thing/thing_manager.rs
```diff
@@ -1918,6 +2060,312 @@ impl ThingManager {
...
+    }
+
+    fn collect_modified_schema_capability_cardinalities(
+        &self,
+        snapshot: &impl WritableSnapshot,
+        modified_owns: &mut HashMap<ObjectType, HashSet<AttributeType>>,
+        modified_plays: &mut HashMap<ObjectType, HashSet<RoleType>>,
+        modified_relates: &mut HashMap<RelationType, HashSet<RoleType>>,
+    ) -> Result<(), Box<ConceptReadError>> {
+        // New / deleted capabilities
```
farost, line 2135: This method is bloated and is quite scary. Not sure what I can do to make it better.
I thought about moving all these snapshot-related operations (old and new ones) into a separate file in  `thing_manager/validation` just to hide these internals, but not sure if it is worth it. 
Didn't find creating a bunch of separate methods useful either.

concept/thing/thing_manager.rs
```diff
@@ -1918,6 +2060,312 @@ impl ThingManager {
...
+        for (key, write) in snapshot.iterate_writes_range(&KeyRange::new_within(
+            TypeEdge::build_prefix(Prefix::EdgeOwns),
+            TypeEdge::FIXED_WIDTH_ENCODING,
+        )) {
+            let edge = TypeEdge::decode(Bytes::reference(key.bytes()));
+            let attribute_type = AttributeType::new(edge.to());
+            let updated_attribute_types = modified_owns.entry(ObjectType::new(edge.from())).or_default();
+            match write {
+                Write::Insert { .. } | Write::Put { .. } => {
+                    updated_attribute_types.insert(attribute_type);
```
farost, line 2146: This is a small optimization to avoid chaining all the supertypes of interface types every time I've mentioned in the long "docs" comment in the `validate` method.

If `owns` is created, it is marked as affected for all the objects of the owner types. Then, this interface type will be affected by the inherited constraints of its supertypes, and it will be checked with all the other siblings. Thus, adding supertypes to the affected interfaces list is not needed here.

However, if it's deleted, we have to mark the supertypes as affected. When we remove a capability, it is no more counted for the super interface types' cardinalities, and if it is, say, `card(1..)` and abstract, we should return an error.

concept/thing/thing_manager.rs
farost: Is AttributeVertex's `FIXED_WIDTH_ENCODING` true? It's not specified nor used anywhere for these vertices generally (only for impl `HashedID` of `StructAttributeID` and ``StringAttributeID`), so I left it this way.

concept/type_/constraint.rs
```diff
@@ -146,11 +146,19 @@ impl ConstraintDescription {
         }
     }
 
-    pub(crate) fn unchecked(&self) -> bool {
+    pub(crate) fn is_checked(&self) -> bool {
```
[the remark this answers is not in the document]
farost: Sure

concept/type_/constraint.rs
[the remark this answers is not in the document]
farost: Nah, it's `requires_operation_time_validation`

concept/thing/thing_manager.rs
```diff
@@ -1918,6 +2060,312 @@ impl ThingManager {
...
+                        }
+                        Prefix::EdgeRelatesReverse => debug_assert!(false, "Unexpected property on reverse relates"),
+                        _ => {}
+                    }
+                }
+                _ => {}
+            }
+        }
+
+        Ok(())
```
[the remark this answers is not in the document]
farost: Hmm, thanks for the hint. I had a look and can say that the logic drastically different. There are some nice methods like `ObjectVertex::is_relation_vertex`, but it will probably make the readability worse, now everything is at least consistent haha

concept/thing/thing_manager.rs
flyingsilverfin: AttributeVertex's fixed width encoding is false. It gets used indirectly here:
```
// ThingManager::get_instances_in
        let prefix = <T as ThingAPI>::prefix_for_type(thing_type);
        let storage_key_prefix =
            <T as ThingAPI>::Vertex::build_prefix_type(prefix, thing_type.vertex().type_id_(), keyspace);
        let snapshot_iterator =
            snapshot.iterate_range(&KeyRange::new_within(storage_key_prefix, prefix.fixed_width_keys()));
            ```

flyingsilverfin (approved):

concept/thing/thing_manager.rs
farost: Ahhh I never paid attention to the boolean flag near the digits in `make_prefix_enum`. Makes sense, thanks!

PR #7334, opened 2025-01-29
Bugfix: direction of indexed relation instruction in lowering

cxdorn:

We fix the logic in the lowering of query plans to executables that determines the direction of indexed relation instructions.

## Motivation

The previous logic was likely wrong.

## Implementation

Trivial.


dmitrii-ubskii (approved):
Looks reasonable. Approved.

PR #7336, opened 2025-01-31
Checkpoint databases at regular intervals

dmitrii-ubskii:

We checkpoint the database every minute, if changes to it were made.

### DIFF
--- a/common/concurrency/interval_runner.rs
+++ b/common/concurrency/interval_runner.rs
@@ -29,9 +29,16 @@ impl IntervalRunner {
     ) -> Self {
         let (shutdown_sender, shutdown_receiver) = sync_channel::<SyncSender<()>>(1);
         thread::spawn(move || {
-            if !initial_delay.is_zero() {
-                thread::sleep(initial_delay);
+            match shutdown_receiver.recv_timeout(initial_delay) {
+                Ok(done_sender) => {
+                    drop(action);
+                    done_sender.send(()).unwrap();
+                    return;
+                }
+                Err(RecvTimeoutError::Timeout) => (),
+                Err(RecvTimeoutError::Disconnected) => return, // TODO log?
             }
+
             loop {
                 action();
                 match shutdown_receiver.recv_timeout(interval) {
--- a/database/database.rs
+++ b/database/database.rs
@@ -260,7 +260,7 @@ impl Database<WALClient> {
         let query_cache = Arc::new(QueryCache::new(0));
         let update_statistics =
             make_update_statistics_fn(storage.clone(), schema.clone(), schema_txn_lock.clone(), query_cache.clone());
-        let checkpoint_fn = make_checkpoint_fn(path.to_owned(), storage.clone());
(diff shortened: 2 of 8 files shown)


database/database.rs
```diff
@@ -201,6 +202,7 @@ impl<D> Database<D> {
 
 impl Database<WALClient> {
     const STATISTICS_UPDATE_INTERVAL: Duration = Duration::from_millis(100);
+    const CHECKPOINT_INTERVAL: Duration = Duration::from_secs(60);
```
dmitrii-ubskii, line 205: Every minute might be too often? Though you can load a lot of data in that time.

common/concurrency/interval_runner.rs
```diff
@@ -29,9 +29,16 @@ impl IntervalRunner {
...
-                thread::sleep(initial_delay);
+            match shutdown_receiver.recv_timeout(initial_delay) {
+                Ok(done_sender) => {
+                    drop(action);
+                    done_sender.send(()).unwrap();
+                    return;
+                }
+                Err(RecvTimeoutError::Timeout) => (),
+                Err(RecvTimeoutError::Disconnected) => return, // TODO log?
             }
```
dmitrii-ubskii, line 40: Without this change, `^C` hangs until this delay is up. This is because the `Drop` of `IntervalRunner` waits for acknowledgement from the background thread and will not complete until then.

database/database.rs
```diff
@@ -447,6 +459,22 @@ impl Database<WALClient> {
...
+    storage: Arc<MVCCStorage<WALClient>>,
+) -> impl FnMut() {
+    move || {
+        let watermark = storage.snapshot_watermark();
+        if prev_checkpoint < watermark {
+            let checkpoint = Checkpoint::new(&path).unwrap();
+            storage.checkpoint(&checkpoint).unwrap();
+            checkpoint.finish().unwrap();
+            prev_checkpoint = watermark;
+        }
```
dmitrii-ubskii, line 474: This can result in tearing if the checkpoint is created _right as_ data is written to the DB, though unlikely. Ideally, `Checkpoint` would lock the storage for the duration.

executor/tests/execute_comparison_check.rs
dmitrii-ubskii, line 1: Purely fmt changes

database/database.rs
flyingsilverfin: Oh can we move both of these to `constants.rs` please!

database/database_manager.rs
```diff
@@ -40,6 +40,11 @@ impl DatabaseManager {
                 })?
                 .path();
 
+            if !entry_path.is_dir() {
+                debug!("Not attempting to load database @ {:?}: not a directory", entry_path);
+                continue;
+            }
```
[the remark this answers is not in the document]
flyingsilverfin: saw this too - thanks

flyingsilverfin (approved):

PR #7338, opened 2025-02-03
Add flag to type-seeder for write stages

### LINKED ISSUE #7333: Redundant type inference checks while inserting a relation with a concrete super role type
## Description
When a role type has a specializing subtype, inserting instances for this role type is not possible even if it is done using a concrete explicit relation type.

## Environment

1. TypeDB distribution: Core
2. TypeDB version: 3.0.3
3. Environment: Mac

## Reproducible Steps

1. Set up
Schema:
```
define 
  relation parentship relates parent; 
  relation fathership sub parentship, relates father as parent;
  entity person plays parentship:parent; 
  entity subperson sub person, plays fathership:father;
```

2. Execute
Write query:
```
insert (parent: $p) isa parentship; $p isa person;
```

3. Unexpected result
```
[INF4] Left type 'parentship' across constraint 'links' is not compatible with right type 'fathership:father'.
Caused: [QUA1] Type inference error while compiling query annotations.
Caused: [QEX8] Error analysing query.
```

## Expected result
Success.

## Additional information

krishnangovindraj:
Adds a flag to type-seeder to indicate the stage is a write stage. This ensures variables constrained by `isa`, or labelled roles are seeded with the exact type, and not (transitive) subtypes.

## Motivation
Fixes #7333 

## Implementation
Adds a flag to type-seeder to indicate the stage is a write stage. This ensures variables constrained by `isa`, or labelled roles are seeded with the exact type, and not (transitive) subtypes.
Updates the TypeSeeder `UnaryConstraint` for Isa and RoleName


compiler/annotation/pipeline.rs
```diff
@@ -333,6 +335,7 @@ fn annotate_stage(
                 type_manager,
                 running_variable_annotations,
                 annotated_function_signatures,
+                true,
```
farost, line 338: Does it even matter here? If not, we could mark it with a comment (or go crazy with an `Option<bool>`) All the variables should be initialized in the previous read/write stages, and they should use transitivity in inference.

farost: It definitely does not insert (the flag is called `is_insert`), and I don't fully understand how it works for deletes. Do we ever reach your checks?

farost (commented):
Don't we want to write tests?

compiler/annotation/pipeline.rs
krishnangovindraj: I should rename the flags to `is_write` I guess.
We might reach it for the role name when delete a links. It's a good idea to add this to the BDD as well.

PR #7339, opened 2025-02-03
Include deleted concepts in structural equality for delete stage

### LINKED ISSUE #7321: Delete pipeline stages return deleted variables
## Description
When performing delete operations, we expect that deleted concept variables are not returned in the result. However, it works inconsistently. 

For example:

✅ Expected (we deleted the variable):
```
 match $u isa user, has username "Alice"; delete $u;
                  
No columns to show
```

✅ Expected (we delete only the `has`):
```
match $u isa user, has username "Alice", has email $email; delete has $email of $u;

   ------------
    $email | alice@typedb.com isa email
    $u     | iid 0x1e00030000000000000002 isa user
   ------------
```

❌ Unexpected (we deleted both the `has` and the `$u`, it should've returned... `$email`?):
```
match $u isa user, has username "Alice", has email $email; 
delete 
has $email of $u; 
$u;                  

   ------------
    $email | alice@typedb.com isa email
    $u     | iid 0x1e00030000000000000002 isa user
   ------------
```

❌ Unexpected (we deleted both the `has` and the `$u`, it should've returned empty columns!):
```
match $u isa user, has username "Alice", has email $email; delete has $email of $u; $u; $email;

   ------------
    $email | alice@typedb.com isa email
    $u     | iid 0x1e00030000000000000002 isa user
   ------------

```

## Environment

1. TypeDB distribution: Core
2. TypeDB version: 3.0.3
3. Environment: Mac
4. Client and version: custom 3.0.2 (with "No columns to show" - it will be a "------------" in 3.0.2)
5. Other details: 

## Reproducible Steps

1. Set up
Schema:
```
define entity user, owns email, owns username; attribute username value string; attribute email value string;
```

Data:
```
insert $a isa user, has username "Alice", has email "alice@typedb.com";
```

2. Execute
Queries from description.

3. Unexpected result
See description.

## Expected result
See description.

krishnangovindraj:
Include deleted concepts in structural equality for delete stage. This avoids a bug where the query cache picks the wrong cached query and runs it - allowing the execution of a delete stage with a totally different set of deleted concepts.

## Motivation
Fixes wrong deleted behaviour due to query-caching; Fixes #7321 

## Implementation
Include deleted concepts in structural equality for delete stage

### DIFF
--- a/ir/translation/pipeline.rs
+++ b/ir/translation/pipeline.rs
@@ -4,7 +4,11 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::{iter::empty, mem};
+use std::{
+    hash::{DefaultHasher, Hasher},
+    iter::empty,
+    mem,
+};
 
 use answer::variable::Variable;
 use primitive::either::Either;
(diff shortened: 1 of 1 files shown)


farost:
What does it mean from the user's perspective? It fixes the bug, but in what direction: what are the answers from the referenced issue?
Don't we need to extend tests?

krishnangovindraj:
>  It fixes the bug, but in what direction:
In the direction of the expected behaviour. But good point. Updated.

> Don't we need to extend tests?
It's a bit of a niche bug.

farost (approved):

PR #7343, opened 2025-02-04
Activate diagnostics reporting for the Docker image

farost:
The TypeDB Docker image receives enough minimal dependencies to report diagnostics and error data for maintenance.

## Motivation
Enable diagnostics reporting from the Docker image. Previously, it could not build HTTPS clients for reporting.

## Implementation
The issue was in a lack of CA certificates on the base Ubuntu image we use for TypeDB CE. It is not relevant to TypeDB (Cluster) as it uses another image with much more preinstalled dependencies (thus weighs much more).

We decided to add another image layer with certificates preinstalled on it.
To achieve it, we use the `container_run_and_commit` rule to create another layer for Docker, which is equivalent to the `RUN` command from Dockerfile. This rule requires a docker binary on the system, but it is already required to at least authenticate on our CD agents, so it's not a limitation.
*NOTE: if docker is installed after bazel, bazel may have issues finding the binary to run this rule, and a `bazel clear --expunge` is needed.*

**Alternatively,** we could use the image used in TypeDB (Cluster), [prepared in dependencies](https://github.com/typedb/typedb-dependencies/tree/development/images/docker/ubuntu). However, it depends on OpenJDK for TypeDB Console support and has a bigger size we don't need for the CE version. 

Whenever we decide to bundle TypeDB CE with TypeDB Console or TypeDB Console is rewritten in Rust, so the Cluster's image does not require OpenJDK anymore, we can consider using the same lighter image in both versions of TypeDB.

We expect the current image to have a limited usage only for TypeDB CE, so we don't move it to the dependencies repo, preparing these image layers directly in TypeDB CE's build files.

### DIFF
--- a/docker/BUILD
+++ b/docker/BUILD
@@ -3,6 +3,7 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 package(default_visibility = ["//visibility:public"])
+load("@io_bazel_rules_docker//docker/util:run.bzl", docker_container_run_and_commit = "container_run_and_commit")
 load("@typedb_bazel_distribution//platform:constraints.bzl", "constraint_linux_arm64", "constraint_linux_x86_64")
 load("@typedb_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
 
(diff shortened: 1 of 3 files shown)


flyingsilverfin (approved):

docker/BUILD
```diff
@@ -33,6 +33,26 @@ genrule(
   cmd = "VERSION=$$(cat $(location //:VERSION)); echo \"$$VERSION-x86_64\" > $@"
 )
 
+docker_image_prepare_commands = [
+    "apt-get update",
+    "apt-get install -y ca-certificates",
+    "rm -rf /var/lib/apt/lists/*",
```
krishnangovindraj, line 39: Do we need to cleanup the lists?

farost: We don't have to. I haven't tested the size without it, but we don't need the whole update, so I clean it up as usually suggested

PR #7344, opened 2025-02-04
Let attribute supertypes be not abstract and fix independent attributes behavior

farost:
Remove the constraint that attribute types can be subtyped only if they are abstract. Now, it is possible to build schemas like:
```
define
  attribute name;
  attribute first-name sub name;
  attribute surname sub name;

  entity dog owns name;
  entity passport owns first-name, owns surname;
```

Or be less restrictive than before with: 
```
define
  entity person owns name, owns first-name, owns surname;
```
Be aware that annotations of `name` will still be applied to all instances of `first-name` and `surname`. Thus, the schema above will let a `person` have either one `name`, one `first-name`, or one `surname`. Use the `@card` annotation to override the default cardinality values for ownerships.

Additionally, a couple of rare bugs related to `has` deletion and dependent attribute retrieval were fixed. This includes:
* Dependent attributes without owners can no longer appear after the query affecting these attributes is run. While it can be a part of the query result (e.g. the returned attribute variable when deleting a `has`), it will no longer be accessible for further queries before and after the commit, thus deleted from the system.
* Abstract attribute supertypes are now consistently returned as results for valid queries.

## Motivation
The removed restriction caused more limitations that required schema hacking and duplications rather than protection. TypeDB 3.0 aims to give its users more flexible instruments for creativity when safe and possible. 

While testing this new feature, multiple old bugs were fixed.

## Implementation
* Removed Concept API checks. 
* Removed filtering for attribute types without value types from executors.
* Fixed bugs in thing_manager (more in code comments).


concept/type_/type_manager/validation/mod.rs
```diff
@@ -35,12 +35,10 @@ typedb_error!(
         RoleNameShouldBeUniqueForRelationTypeHierarchy(7, "Role name of '{role}' should be unique in relation type hierarchy of '{relation}'.", role: Label, relation: Label),
         CycleFoundInTypeHierarchy(8, "A cycle was found in the type hierarchy between '{start}' and '{end}'.", start: Label, end: Label),
         ChangingAttributeTypeSupertypeWillLeadToConflictingValueTypes(9, "Changing supertype of attribute type '{attribute}' will lead to conflicting value types between '{value_type:?}' and '{new_super_value_type}'.", attribute: Label, value_type: Option<ValueType>, new_super_value_type: ValueType),
-        CannotUnsetAbstractnessOfAttributeTypeAsItHasSubtypes(10, "Cannot unset the abstractness of attribute type '{attribute}' as it has subtypes.", attribute: Label),
```
farost, line 38: What's our approach to error enumeration? If I remove an error, I shall leave holes like `... 8, 9, 11, ...`, right?

tests/behaviour/concept/migration/function.rs
farost, line 1: https://github.com/typedb/typedb-behaviour/pull/333/files#r1942712163

concept/thing/thing_manager.rs
```diff
@@ -2658,41 +2656,56 @@ impl ThingManager {
...
+        attribute: &Attribute,
+    ) -> Result<(), Box<ConceptWriteError>> {
         let owner_status = owner.get_status(snapshot, self);
-        let has = ThingEdgeHas::new(owner.vertex(), attribute.vertex()).into_storage_key().into_owned_array();
-        let has_reverse =
+        let has = ThingEdgeHas::new(owner.vertex(), attribute.vertex()).into_storage_key();
+        let has_array = has.clone().into_owned_array();
+        let has_reverse_array =
             ThingEdgeHasReverse::new(attribute.vertex(), owner.vertex()).into_storage_key().into_owned_array();
-        match owner_status {
```
farost, line 2683: As already mentioned, I can't simply put a `delete` write and expect it to act like "just make sure it doesn't exist". If a record is not in storage, the statistics are corrupted, and the server crashes. 

There are two simplest ways I can make the unset method correctly:

First. I don't implement the direct storage lookup: 
1. Always check if a has put write exists. Unput if needed
2. If the owner is just inserted into the buffer, finish. If it's persisted... Check if the has exists generally: check the buffer (redundant as we already unput), check the storage. If exists, we need to write a delete.
**Result:** 
Best case: 2 buffer reads
Worst case: 3 buffer reads, 1 storage read

Second. 
1. If the owner is just inserted into the buffer, unput if the has put in the buffer (if put). Finish.
2. If the owner is not just inserted, read the storage directly. If a has exists, write a delete.
**Result:**
Best case: 2 buffer reads
Worst case: 2 buffer reads, 1 storage read

I implemented the first option because I didn't want to create a separate interface for `snapshot` to access the storage directly. It's not a good practice, although we already read the storage directly in the Concept layer (see `attribute` and how it filters out dependent attributes...). We believe that one additional buffer lookup in the worst case is not critical.

concept/thing/thing_manager.rs
```diff
@@ -2658,41 +2656,56 @@ impl ThingManager {
...
+                    snapshot.unput_val(has_reverse_array.clone(), value);
+                }
+                Write::Delete => {}
+                Write::Insert { .. } => {
+                    unreachable!("Encountered an `insert` has. Owner: {owner:?}, attribute: {attribute:?}.")
+                }
+            }
         }
+
+        if !matches!(owner_status, ConceptStatus::Inserted) {
```
farost, line 2697: Could make a `match` will all options just to add an `unreachable`, idk if we really need it... Quite annoying to read.

concept/thing/thing_manager.rs
```diff
@@ -2658,41 +2656,56 @@ impl ThingManager {
...
+            if self
+                .has_attribute(snapshot, owner, attribute)
```
farost, line 2699: This call reads both the buffer and the storage (if not found in the buffer)

concept/thing/thing_manager.rs
```diff
@@ -2777,13 +2790,12 @@ impl ThingManager {
         role_type: RoleType,
         count: u64,
     ) -> Result<(), Box<ConceptWriteError>> {
-        let links = ThingEdgeLinks::build_links(relation.vertex(), player.vertex(), role_type.vertex());
-        let links_reverse = ThingEdgeLinks::build_links_reverse(player.vertex(), relation.vertex(), role_type.vertex());
-
         if count == 0 {
-            snapshot.delete(links.into_storage_key().into_owned_array());
-            snapshot.delete(links_reverse.into_storage_key().into_owned_array());
+            self.unset_links(snapshot, relation, player, role_type)
```
farost, line 2793: I replicate the same logic for links on the thing_manager layer. However, the Concept API interfaces are different from `has`: it's `remove_player_single` instead of `unset_has_unordered` and it uses a different flow with more attention to counts. Following the discussion in Discord (https://discord.com/channels/665254494820368395/672580553677078548/1339540049464791112), we want to make the interfaces similar, so I've created a refactoring task: https://github.com/typedb/typedb/issues/7358

We should pay attention to this task when we work on lists.

concept/thing/thing_manager.rs
```diff
@@ -2658,41 +2656,56 @@ impl ThingManager {
...
-                snapshot.delete(has);
-                snapshot.delete(has_reverse);
-            }
-            ConceptStatus::Put => unreachable!("Encountered a `put` attribute owner: {owner:?}."),
-            ConceptStatus::Deleted => unreachable!("Attempting to unset attribute ownership on a deleted owner."),
+
+        if let Some(has_write) = snapshot.get_write(has.as_reference()).cloned() {
+            match has_write {
+                Write::Put { value, .. } => {
+                    snapshot.unput_val(has_array.clone(), value.clone());
```
farost, line 2687: Unput the put value instead of 1.

executor/instruction/isa_executor.rs
```diff
@@ -217,13 +217,8 @@ pub(super) fn instances_of_all_types_chained(
     // Since the object types are sorted, and instance ordering follows matches type ordering, we have instance-sorting here
     let object_iter: MultipleTypeIsaObjectIterator = object_iters.into_iter().flatten();
 
-    let type_manager = thing_manager.type_manager();
     let attribute_iters: Vec<_> = attribute_types
         .into_iter()
-        // TODO: we shouldn't really filter out errors here, but presumably a ConceptReadError will crop up elsewhere too if it happens here
-        .filter(|(type_, _)| {
-            type_.as_attribute_type().get_value_type(snapshot, type_manager).is_ok_and(|vt| vt.is_some())
```
farost, line 225: We should not really filter out attribute types without value types here. It is a valid result (abstract types generally is another question, this is handled by the type inference).

tests/behaviour/concept/migration/BUILD
```diff
@@ -5,16 +5,6 @@
 load("@typedb_dependencies//tool/checkstyle:rules.bzl", "checkstyle_test")
 load("@rules_rust//rust:defs.bzl", "rust_test")
 
-rust_test(
```
farost, line 8: Removed as it's unused. However, we may still want to introduce such tests if they are not written anywhere. @krishnangovindraj ?

farost: Resolved here https://github.com/typedb/typedb-behaviour/pull/333/files

dmitrii-ubskii (changes requested):
The logic looks good, but there is some work to be done on the implementation side.

concept/thing/thing_manager.rs
farost: Good p, fixed

storage/snapshot/write.rs
```diff
@@ -93,4 +93,35 @@ impl Write {
...
+        }
+    }
+
+    pub fn is_delete(&self) -> bool {
+        match self {
+            WriteCategory::Insert | WriteCategory::Put => false,
+            WriteCategory::Delete => true,
+        }
+    }
 }
```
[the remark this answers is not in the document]
farost: I was not sure what's the safest way to implement this. We already have two versions of "new": `Insert` and `Put`. What if we have `Delete` and somewhat `Remove`? At least we'll catch this match in compile time and refactor this stuff. That's the idea.

concept/type_/type_manager/validation/mod.rs
```diff
@@ -35,13 +35,11 @@ typedb_error!(
...
         CannotDeleteTypeWithExistingSubtypes(11, "Cannot delete type '{label}' as it has existing subtypes.", label: Label),
         RelatesNotInheritedForSpecialisation(12, "Role type '{role}' must be inherited without specialization by supertypes to be specialized by relation type '{relation}'.", relation: Label, role: Label),
         RelatesIsAlreadySpecialisedByASupertype(13, "Role type '{role}' is already specialised by a supertype of relation type '{relation}'.", relation: Label, role: Label),
         CannotManageAnnotationsForSpecialisingRelates(14, "Cannot manage annotations for the specialising relates role '{role}'.", role: Label),
-        AttributeTypeMustBeAbstractToHaveSubtypes(15, "Attribute type '{attribute}' must be abstract to have subtypes.", attribute: Label),
-        AbstractTypesSupertypeHasToBeAbstract(16, "Abstract type '{attribute}' must have an abstract supertype, but it has '{super_attribute}'.", attribute: Label, super_attribute: Label),
+        AbstractTypesSupertypeHasToBeAbstract(16, "Abstract type '{type_}' must have an abstract supertype, but it has '{supertype}'.", type_: Label, supertype: Label),
```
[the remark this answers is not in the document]
farost: Okay, I hear you... I swear I'll at least try to adjust my habits

concept/thing/attribute.rs
```diff
@@ -260,25 +270,30 @@ where
...
-                                        attribute_vertex,
-                                    ) {
-                                        Ok(has_owner) => {
-                                            if has_owner {
-                                                self.state = State::ItemReady
-                                            } else {
-                                                advance_attribute = true
-                                            }
+                                    unreachable!("Not new and not delete write category");
+                                }
```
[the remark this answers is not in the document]
farost: I kinda didn't want to match here, but it's probably just easier. I'll get rid of the methods for now, you're right.

concept/thing/attribute.rs
```diff
@@ -260,25 +270,30 @@ where
...
+                                match Self::has_owner(
+                                    self.has_reverse_iterator_storage.as_mut().unwrap(),
+                                    attribute_vertex,
+                                ) {
+                                    Ok(has_owner) => {
+                                        if has_owner {
+                                            self.state = State::ItemReady
+                                        } else {
+                                            advance_attribute = true
                                         }
```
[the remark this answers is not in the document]
farost: I had the same urge after the refactor above, thank!

concept/thing/attribute.rs
```diff
@@ -299,10 +314,13 @@ where
     ) -> Result<bool, Box<ConceptReadError>> {
         let has_reverse_prefix = ThingEdgeHasReverse::prefix_from_attribute(attribute_vertex);
         has_reverse_iterator.seek(has_reverse_prefix.as_reference());
-        match has_reverse_iterator.peek() {
+        match has_reverse_iterator
+            .peek()
+            .transpose()
+            .map_err(|source| Box::new(ConceptReadError::SnapshotIterate { source }))?
+        {
```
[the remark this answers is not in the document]
farost: I agree that big blocks in scrutinees are not great, and I hate this verbosity, but I did really not like the added complexity of match branches. It makes me spend more time on reading the code to understand the logic of the method (just like `Option<bool>`). But I trust your experience, so I'll revert this

concept/thing/has.rs
```diff
@@ -6,10 +6,17 @@
...
+    Keyable,
+};
 use lending_iterator::higher_order::Hkt;
+use storage::snapshot::ReadableSnapshot;
 
-use crate::thing::{attribute::Attribute, object::Object, ThingAPI};
+use crate::{
+    thing::{attribute::Attribute, object::Object, thing_manager::ThingManager, ThingAPI},
+    ConceptStatus,
+};
```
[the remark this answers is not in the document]
farost: I forgot that rustfmt does not clean up unused imports ='(

concept/thing/thing_manager.rs
```diff
@@ -2658,41 +2655,56 @@ impl ThingManager {
...
+        &self,
+        snapshot: &mut impl WritableSnapshot,
+        owner: impl ObjectAPI,
+        attribute: &Attribute,
+    ) -> Result<(), Box<ConceptWriteError>> {
         let owner_status = owner.get_status(snapshot, self);
-        let has = ThingEdgeHas::new(owner.vertex(), attribute.vertex()).into_storage_key().into_owned_array();
-        let has_reverse =
+        let has = ThingEdgeHas::new(owner.vertex(), attribute.vertex()).into_storage_key();
+        let has_array = has.clone().into_owned_array();
```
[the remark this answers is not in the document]
farost: We need a `StorageKeyArray` in `delete` and `unput_val` methods, and array does not have an `as_reference`, I could not find an alternative.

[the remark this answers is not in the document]
dmitrii-ubskii: Seems like an oversight, but okay. Noted for later.

dmitrii-ubskii (approved):

concept/type_/type_manager/validation/mod.rs
flyingsilverfin: Feel free to renumber for now

flyingsilverfin (approved):
Good finds, tricky stuff!!

storage/snapshot/write.rs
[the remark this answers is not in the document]
flyingsilverfin: I agree keeping a match is good for future extensions

PR #7345, opened 2025-02-05
Database debug tools

dmitrii-ubskii:

We introduce the database/tools package, hosting small CLI tools to help investigate misbehaving or corrupted databases.

Tools included in this commit:
- read-wal: read and dump the contents of the WAL in the database at specified sequence numbers;
- replay-wal: replay the records from one WAL to another, filtering by record type and sequence numbers.


encoding/layout/prefix.rs
```diff
@@ -72,7 +72,7 @@ impl Prefix {
         if Prefix::VertexEntityType.prefix_id().byte < Prefix::VertexRelationType.prefix_id().byte {
             Prefix::VertexRelationType
         } else {
-            Prefix::VertexRelationType
+            Prefix::VertexEntityType
         }
     }
```
dmitrii-ubskii, line 77: Drive-by: `max` didn't actually return the higher prefix.

encoding/layout/prefix.rs
```diff
@@ -82,42 +82,42 @@ impl Prefix {
...
-    VertexRoleType => 15, true;
-    DefinitionStruct => 20, true;
-    DefinitionFunction => 21, true;
+    // Reserved: 0x00-0x09
+    VertexEntityType => 0x0A, true;
+    VertexRelationType => 0x0B, true;
+    VertexAttributeType => 0x0C, true;
+    VertexRoleType => 0x0F, true;
+    DefinitionStruct => 0x14, true;
+    DefinitionFunction => 0x15, true;
```
dmitrii-ubskii, line 91: I usually only look here when reading a raw dump of a key or a commit, which is in hex.

The values were replaced automatically with `printf("0x%02X", num)`, with no changes.

encoding/graph/type_/vertex.rs
```diff
@@ -69,8 +75,9 @@ impl Prefixed<BUFFER_KEY_INLINE> for TypeVertex {}
 
 impl Typed<BUFFER_KEY_INLINE> for TypeVertex {
     fn type_id_(&self) -> TypeID {
-        // TODO: is this the best way to get the TypeID out?
-        TypeID::decode(self.clone().to_bytes()[Self::RANGE_TYPE_ID].try_into().unwrap())
+        debug_assert_eq!(Self::LENGTH, 3);
+        debug_assert_eq!(Self::RANGE_TYPE_ID, 1..3);
+        TypeID { value: (self.value & 0xFFFF) as u16 }
```
dmitrii-ubskii, line 80: No.

dmitrii-ubskii: Technically `& 0xFFFF` is unnecessary, `as` is defined to truncate, but I decided to be explicit.

common/bytes/byte_array.rs
```diff
@@ -110,7 +110,7 @@ impl<const INLINE_BYTES: usize> ByteArray<INLINE_BYTES> {
 
 impl<const INLINE_BYTES: usize> fmt::Debug for ByteArray<INLINE_BYTES> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct(std::any::type_name_of_val(self)).field("data", &HexBytesFormatter::borrowed(self)).finish()
+        write!(f, "{}", HexBytesFormatter::borrowed(self))
     }
 }
```
dmitrii-ubskii, line 115: I never want to know if the bytes are in a Reference or an Array in the debug prints. Reduce noise.

concept/thing/statistics.rs
```diff
@@ -468,6 +469,80 @@ struct CommittedWrites {
     operations: OperationsBuffer,
 }
 
+impl fmt::Debug for Statistics {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
dmitrii-ubskii, line 473: A little more compact than the default `Debug` impl.

database/tools/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+		default-features = false
+
+[[bin]]
+	path = "read-wal.rs"
+	name = "read-wal"
+
+[[bin]]
+	path = "replay-wal.rs"
+	name = "replay-wal"
+
```
dmitrii-ubskii, line 41: Note: Cargo sync tool can't actually handle two binaries in the same package! This entry was added manually.

database/tools/replay-wal.rs
dmitrii-ubskii, line 1: Run as `cargo run -p database-tools --bin replay-wal -- <args>`, or build first and run directly as `target/debug/replay-wal <args>`

database/tools/BUILD
```diff
@@ -0,0 +1,32 @@
...
+    ],
+    deps = [
+        "//concept",
+        "//durability",
+        "//storage",
+        "@crates//:clap",
+    ],
+)
+
+rust_binary(
```
farost, line 20: As it's still declared as a binary, we can run it through bazel, right? You just need to pass the args after `--`. 

If (for some reason) not, let's move your PR's comment about how to run it right in the BUILD file. We can also just mark that it can be run only through cargo due to the sync tool.

dmitrii-ubskii: We can, bazel command is as expected, `bazel run //database/tools:read-wal` e.g.

I've specifically called out cargo because the `--package/-p` and `--bin` flags will be new to many.

flyingsilverfin (approved):

PR #7346, opened 2025-02-05
Introduce multi-staged server termination

farost:
Make server termination through `CTRL-C` interrupt opened transactions to proceed with the resource deallocation.
Introduce two stages of termination:
* When `CTRL-C` is pressed, a graceful termination will be initiated, ensuring that all resources are freed securely and all open transactions finish preparing current batches of streamed answers (this might take some time to calculate answers for the whole batch, but will not wait until all the answers are sent).
* When `CTRL-C` is pressed while a graceful termination is active, the server's process will be forcefully immediately terminated. This prevents the server from hanging because of long-running queries or other issues. 

## Motivation
Current graceful termination does not initiate the closure of existing transactions, and there is no way to use `CTRL-C` for forced process termination. With this PR, the server's binary matches the standard of applications running through a terminal.

## Implementation
Create a channel for a shutdown signal. This channel can be used for multiple components of the system initialized in a `Server` instance. The sender of the signal is the `Server` itself, while current listeners are `TransactionService`s.  

The following sections are a breakdown of the current server's behavior for different situations regarding process termination and transaction deletion.

### Database deletion while a transaction is open
`Cannot delete database since it is in use.` -> can be forcefully shut down, then this command will reconnect the console and successfully delete the db. Is fine

### Transaction closure
Correctly closes the transaction awaiting current answer batches, etc.

### CTRL+C without opened transaction 
```
Running TypeDB CE 3.0.5 in development mode.
Ready!
^C
Received CTRL-C. Initiating graceful shutdown...
Exited.
```
Closes through a graceful shutdown just like transaction closure, finishes immediately. It waits for 2 seconds if it's run in a published mode and already sent a diagnostics reporting batch once (it will send another batch on shutdown).

### CTRL+C with opened transaction
Same as the previous option, with little branches: 

#### No queries are running
The server is terminated instantly. TypeDB Console does not react on it. What's happening with the Console based on multiple commands:

```
> transaction a schema
a::schema> match $p isa person;
           
[CXN06] The transaction is closed and no further operation is allowed.
```

```
> transaction a schema
a::schema> close
Transaction closed without committing changes
```

```
> transaction a schema
a::schema> commit
[CXN06] The transaction is closed and no further operation is allowed.
```

#### There is an active query streaming answers
There are multiple possible options. What errors I've seen:

When the answers are not yet streamed:
```
a::schema> match $p isa person; not {$v isa person; $v is $p;};
           
Finished validation and compilation...
TSV12 
[TSV12] Execution interrupted by to a concurrent transaction close.
```

When answer streaming is interrupted:
```
…..
   --------
    $p | iid 0x1e00000000000000002866 isa person
    $v | iid 0x1e00000000000000012866 isa person
   --------
[CXN06] The transaction is closed and no further operation is allowed.
```

#### Somewhere in between
**!!!!!!!**
Probably, the driver waited for the initial response but got Close. It's clearly a driver's mistake, we can also cover it on the driver's side:
```
a::schema> match $p isa person; insert $v isa person;
           
[INT4] Unexpected response type for remote procedure call: Close. This is either a version compatibility issue or a bug.
```

### Double CTRL+C anywhere
Terminates immediately without the "Exited" keyword. It is reproducible if we send diagnostics data on shutdown (this way, you have a moment to actually press the button the second time. In my other cases, everything works just too fast without many operations going in parallel):
```
Running TypeDB CE 3.0.5.
Ready!
Reported? true and true
^C
Received CTRL-C. Initiating graceful shutdown...
^C
Received CTRL-C. Forcing shutdown...
λ ~/work/typedb/
```

### DIFF
--- a/server/service/transaction_service.rs
+++ b/server/service/transaction_service.rs
@@ -52,6 +52,7 @@ use tokio::{
     sync::{
         broadcast,
         mpsc::{channel, Receiver, Sender},
+        watch,
     },
     task::{spawn_blocking, JoinHandle},
 };
@@ -60,7 +61,7 @@ use tonic::{Status, Streaming};
 use tracing::{event, Level};
 use typedb_protocol::{
     query::Type::{Read, Write},
-    transaction::{stream_signal::Req, Server},
+    transaction::{stream_signal::Req, Server as ProtocolServer},
 };
 use typeql::{
     parse_query,
@@ -120,9 +121,10 @@ pub(crate) struct TransactionService {
     diagnostics_manager: Arc<DiagnosticsManager>,
 
     request_stream: Streaming<typedb_protocol::transaction::Client>,
-    response_sender: Sender<Result<typedb_protocol::transaction::Server, Status>>,
+    response_sender: Sender<Result<ProtocolServer, Status>>,
     query_interrupt_sender: broadcast::Sender<InterruptType>,
     query_interrupt_receiver: ExecutionInterrupt,
+    shutdown_receiver: watch::Receiver<()>,
 
     transaction_timeout_millis: Option<u64>,
     schema_lock_acquire_timeout_millis: Option<u64>,
@@ -242,9 +244,10 @@ impl StreamingCondition {
 impl TransactionService {
     pub(crate) fn new(
         request_stream: Streaming<typedb_protocol::transaction::Client>,
-        response_sender: Sender<Result<typedb_protocol::transaction::Server, Status>>,
+        response_sender: Sender<Result<ProtocolServer, Status>>,
         database_manager: Arc<DatabaseManager>,
         diagnostics_manager: Arc<DiagnosticsManager>,
+        shutdown_receiver: watch::Receiver<()>,
     ) -> Self {
         let (query_interrupt_sender, query_interrupt_receiver) = broadcast::channel(1);
 
@@ -256,6 +259,7 @@ impl TransactionService {
             response_sender,
             query_interrupt_sender,
             query_interrupt_receiver: ExecutionInterrupt::new(query_interrupt_receiver),
+            shutdown_receiver,
 
             transaction_timeout_millis: None,
             schema_lock_acquire_timeout_millis: None,
@@ -275,6 +279,11 @@ impl TransactionService {
             let result = if self.running_write_query.is_some() {
                 let (req_id, write_query_worker) = self.running_write_query.as_mut().unwrap();
                 tokio::select! { biased;
+                    _ = self.shutdown_receiver.changed() => {
+                        event!(Level::TRACE, "Shutdown signal received, closing transaction service.");
+                        self.do_close().await;
+                        return;
+                    }
                     write_query_result = write_query_worker => {
                         let req_id = *req_id;
                         self.running_write_query = None;
@@ -292,8 +301,16 @@ impl TransactionService {
                     }
                 }
             } else {
-                let next = self.request_stream.next().await;
-                self.handle_next(next).await
+                tokio::select! { biased;
+                    _ = self.shutdown_receiver.changed() => {
+                        event!(Level::TRACE, "Shutdown signal received, closing transaction service.");
+                        self.do_close().await;
+                        return;
+                    }
+                    next = self.request_stream.next() => {
+                        self.handle_next(next).await
+                    }
+                }
             };
 
             match result {
@@ -433,7 +450,7 @@ impl TransactionService {
     }
 
     async fn respond_query_response(
-        response_sender: &Sender<Result<typedb_protocol::transaction::Server, Status>>,
+        response_sender: &Sender<Result<ProtocolServer, Status>>,
         req_id: Uuid,
         immediate_query_response: ImmediateQueryResponse,
     ) -> ControlFlow<(), ()> {
@@ -1438,7 +1455,7 @@ impl TransactionService {
 
 #[derive(Debug)]
 struct QueryStreamTransmitter {
-    response_sender: Sender<Result<Server, Status>>,
+    response_sender: Sender<Result<ProtocolServer, Status>>,
     req_id: Uuid,
     prefetch_size: usize,
     network_latency_millis: usize,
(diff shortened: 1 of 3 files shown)


server/service/transaction_service.rs
```diff
@@ -275,6 +279,11 @@ impl TransactionService {
             let result = if self.running_write_query.is_some() {
                 let (req_id, write_query_worker) = self.running_write_query.as_mut().unwrap();
                 tokio::select! { biased;
+                    _ = self.shutdown_receiver.changed() => {
```
farost, line 282: As we use `biased`, the shutdown signal can be prioritized. I copypaste the same small piece of code for both `if` branches here as we need to `return` anyway.

server/typedb.rs
```diff
@@ -122,12 +130,13 @@ impl Server {
...
 
         Self::print_hello(self.distribution, self.version, self.config.server.is_development_mode);
+
         Self::create_tonic_server(&self.config.server.encryption)?
             .layer(&authenticator)
             .add_service(service)
             .serve_with_shutdown(self.address, async {
-                tokio::signal::ctrl_c().await.expect("Failed to listen for a CTRL-C signal");
-                println!("\nReceived a CTRL-C signal. Shutting down...");
+                Self::shutdown_signal().await;
```
farost, line 138: Wait for a signal. Send the signal to the listeners.

server/typedb.rs
```diff
@@ -56,6 +60,8 @@ impl Server {
         version: &'static str,
         deployment_id: Option<String>,
     ) -> Result<Self, ServerOpenError> {
+        let (shutdown_sender, shutdown_receiver) = tokio::sync::watch::channel(());
```
farost, line 63: Using `watch` for one producer and multiple receivers of the last (for us, only) message.

server/typedb.rs
```diff
@@ -251,6 +260,21 @@ impl Server {
         }
         println!("Ready!");
     }
+
+    async fn shutdown_signal() {
+        Self::listen_ctrl_c().await;
+        println!("\nReceived CTRL-C. Initiating graceful shutdown...");
+
+        tokio::spawn(async {
```
farost, line 268: When the button is pressed first, we tell the outside world: "Okay, let's shut down". However, we also prepare the ground for the forced shutdown, which will just terminate the process with an exit code.

Btw, not sure if we want to use different exit codes.

server/typedb.rs
```diff
@@ -251,6 +260,21 @@ impl Server {
         }
         println!("Ready!");
     }
+
+    async fn shutdown_signal() {
```
flyingsilverfin, line 264: I would call this `spawn_force_shutdown_listener` or something

farost: It's not a force shutdown listener, just `spawn_shutdown_listener`

server/typedb.rs
flyingsilverfin, line 138: You should print the 'initiating graceful shutdown' after spawning the forceful shutdown in that method. The printout belongs in here IMO.

Why don't you:
1. print "initiating graceful shutdown"
2. sent the shutdown sender signal
3. initiate the force shutdown listener
?

server/typedb.rs
farost: The forced listener is spawned in a task inside after the first signal is received

server/typedb.rs
farost: Current idea and the flow:

The tonic server is initialized with a shutdown serving. The shutdown logic is initialized inside the closure. Its responsibility: wait for a shutdown signal, do something as the Server when it's received.

Working with the channel clearly belongs here to me. 
Then, all the other stuff is quite optional.

My approach is that I wrap all the logic about "working" with signals, i.e. waiting for them and communicating with the user to confirm that a signal is being processed, in a separate function.

Inside the function, we wait for the first Ctrl+C signal, and only after that we spawn an additional background task for the terminating signal "just in case", which is our last resort. But we generally care about the initial terminating signal.

I don't mind just moving the content of `shutdown_signal()` here. I thought that it's not really interesting to see in this initialization method inside the closure, thus I hid it.

server/typedb.rs
```diff
@@ -122,12 +130,13 @@ impl Server {
...
 
         Self::print_hello(self.distribution, self.version, self.config.server.is_development_mode);
+
         Self::create_tonic_server(&self.config.server.encryption)?
             .layer(&authenticator)
             .add_service(service)
             .serve_with_shutdown(self.address, async {
-                tokio::signal::ctrl_c().await.expect("Failed to listen for a CTRL-C signal");
-                println!("\nReceived a CTRL-C signal. Shutting down...");
+                // The tonic server starts a shutdown process when this closure execution finishes
```
farost, line 138: Refactored

server/typedb.rs
```diff
@@ -251,6 +260,24 @@ impl Server {
...
+
+    async fn forced_shutdown_handler() {
+        Self::block_and_listen_ctrl_c().await;
+        println!("\nReceived CTRL-C. Forcing shutdown...");
+        std::process::exit(1);
+    }
+
+    async fn block_and_listen_ctrl_c() {
+        tokio::signal::ctrl_c().await.expect("Failed to listen for CTRL-C signal");
+    }
```
dmitrii-ubskii, line 280: "Block" has a very specific meaning in `async`. This could just be called `ctrl_c()`, the use as `Self::ctrl_c().await` reads pretty well I think.
```suggestion
    async fn ctrl_c() {
        tokio::signal::ctrl_c().await.expect("Failed to listen for CTRL-C signal");
    }
```

server/service/transaction_service.rs
```diff
@@ -275,6 +279,11 @@ impl TransactionService {
             let result = if self.running_write_query.is_some() {
                 let (req_id, write_query_worker) = self.running_write_query.as_mut().unwrap();
```
dmitrii-ubskii, line 280:
```suggestion
            let result = if let Some((req_id, write_query_worker)) = &mut self.running_write_query {
```

dmitrii-ubskii (approved):
LGTM, fairly straightforward.

server/typedb.rs
farost: @flyingsilverfin `ctrl_c` or `listen_ctrl_c`? How readable is it for you?

flyingsilverfin: `await_ctlr_c` or `listen_ctlr_c`

flyingsilverfin: yeah ok i get dmitrii's argument, but i still find `listen_ctlr_c` easier to read without looking at the invocation.

farost: Done

PR #7347, opened 2025-02-06
Avoid persisting empty statistics deltas

dmitrii-ubskii:

We resolve the issue in which statistics would be written to the WAL more often than necessary, which could cause WAL (not the data) to become corrupted.

## Motivation

If a schema commit was performed within 8 commits before the statistics was last updated, the sequence number stamp in the statistics would roll back, but not the counts. Any intermittent commit deltas would be reapplied repeastedly, in some cases causing underflows and crashes.

In addition, statistics would be unconditionally persisted to the WAL when said schema commit was encountered, adding needless repeated records _and_ persisting the corrupted statistics records, causing startup crashes.

## Implementation

We now keep track of when the last write was performed, as well as fix the issue with rolling back the sequence number.

### DIFF
--- a/concept/thing/statistics.rs
+++ b/concept/thing/statistics.rs
@@ -182,14 +182,11 @@ impl Statistics {
         commits: &BTreeMap<SequenceNumber, CommittedWrites>,
         storage: &MVCCStorage<D>,
     ) -> Result<(), MVCCReadError> {
-        let mut total_delta = 0;
         for (sequence_number, writes) in commits.range(self.sequence_number.next()..) {
-            total_delta += self.update_write(*sequence_number, writes, commits, storage)?;
+            let delta = self.update_write(*sequence_number, writes, commits, storage)?;
+            self.total_count = self.total_count.checked_add_signed(delta).unwrap();
+            self.sequence_number = *sequence_number;
         }
-        if let Some((&last_sequence_number, _)) = commits.last_key_value() {
-            self.sequence_number = last_sequence_number;
-        }
-        self.total_count = self.total_count.checked_add_signed(total_delta).unwrap();
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


concept/thing/statistics.rs
```diff
@@ -933,7 +934,8 @@ mod serialise {
                         .collect();
                     Ok(Statistics {
                         encoding_version: statistics_version,
-                        sequence_number: open_sequence_number,
+                        sequence_number,
+                        last_durable_write_sequence_number: sequence_number,
```
dmitrii-ubskii, line 938: Assuming `last_durable_write_sequence_number` would be the one we deserialize from, so that encoding doesn't have to change.

concept/thing/statistics.rs
```diff
@@ -130,27 +133,25 @@ impl Statistics {
...
                         data_commits.clear();
-
-                        let writes = CommittedWrites {
-                            open_sequence_number: record.open_sequence_number(),
-                            operations: record.into_operations(),
-                        };
-                        let mut commits = BTreeMap::new();
-                        commits.insert(seq, writes);
-                        self.update_writes(&commits, storage).map_err(|err| DataRead { source: err })?;
                     }
```
dmitrii-ubskii, line 155: Original flow when encountering a schema commit:
1. Apply deltas from data commits read so far.
2. Persist the statistics.
3. Apply any deltas from the schema commit.
4. Reset data commits and move on.

What went wrong:
- In step 1, we set the sequence number to whichever was the last in the applied commits, _even if it's earlier than where the statistics are,_ which can happen if the schema commit is a part of the prefetched context.
- In step 2, if the last time we've written to disk was just before the schema commit, we duplicate the record. We do that every time `may_synchronise` is triggered, which is to say every 100 ms so long as there's a schema commit record in context.

Instead, the new flow is:
1. If the schema commit was already applied, go to 5.
2. If the last durable write was performed on the commit immediately preceding this one, go to 4.
3. Apply data commit deltas and persist the statistics.
4. Apply the schema commit deltas.
4. Reset data commits and move on.

flyingsilverfin (approved):

PR #7356, opened 2025-02-10
Match disjunction variables as optional

dmitrii-ubskii:

We treat variables that are only used in a branch of a disjunction as optionally matched. If the branch is not used by a given row, the column is left empty.

## Implementation

All named variables in a nested pattern are considered to be visible from every other subpattern. Previously, any variable referenced in a subpattern of a conjunction (viz. in a disjunction branch) would only be an output if it was also referenced _outside_ of the subpattern.

Negations are the exception: if a variable is only introduced and used in a negation, its value does not escape the negation's scope. If a value were to be successfully bound to this variable, the negation constraint would be violated, and the value not returned in the first place.

### DIFF
--- a/ir/translation/function.rs
+++ b/ir/translation/function.rs
@@ -123,7 +123,7 @@ pub(crate) fn translate_function_block(
                 typedb_source: err,
             })?;
 
-    let mut illegal_stages = stages.iter().filter(|stage| match stage {
+    let has_illegal_stages = stages.iter().any(|stage| match stage {
         TranslatedStage::Insert { .. } | TranslatedStage::Update { .. } | TranslatedStage::Delete { .. } => true,
         TranslatedStage::Match { .. }
         | TranslatedStage::Distinct(_)


ir/pipeline/block.rs
```diff
@@ -220,22 +220,21 @@ impl BlockContext {
...
     }
 
-    pub fn is_visible_child(&self, child: ScopeId, candidate: ScopeId) -> bool {
-        self.is_transparent(child)
-            && self
-                .scope_parents
-                .get(&child)
-                .is_some_and(|&parent| candidate == parent || self.is_visible_child(parent, candidate))
+    pub fn is_visible_child(&self, child: ScopeId, ancestor: ScopeId) -> bool {
+        if !self.is_transparent(child) {
```
flyingsilverfin, line 233: Can we change `is_transparent` to `is_negation`, and change `ScopeTransparency` to `enum ScopeType { Negation, NonNegation }` - does that make sense? Do we only have `Root` and `Negation` scopes now? Maybe it's `Negation/Root`?

flyingsilverfin (approved):
Approved with 1 comment

ir/pipeline/block.rs
dmitrii-ubskii: My issue with that naimng is that it tells me nothing. Why are we interested in whether a scope is a negation — because the variables declared in a negation are not visible from outer scopes. Transparent/opaque at least somewhat carries the intent.

PR #7359, opened 2025-02-13
Unsatisfiable type constraints do not cause type-inference errors; Enable more BDD tests

krishnangovindraj:
Queries containing unsatisfiable type-constraints no longer fail type-inference. Instead the (sub)pattern returns 0 answers.  We also enable BDD tests which were failing due to overly strict debug_asserts

## Motivation
More BDD coverage, more intuitive behaviour on type-queries.

## Implementation
* Overly strict debug asserts are removed.
* Type-inference now errors only if any **thing** variable has an empty set of type-annotations **in any subpattern**.
* Conjunctions containing unsatisfiable type-constraints are optimised away by replacing their constraints with a single `OptimisedAway` constraint. All the nested patterns are cleared too. This is done in `apply_transformations`


executor/instruction/type_list_executor.rs
```diff
@@ -49,7 +49,7 @@ impl TypeListExecutor {
     ) -> Self {
         debug_assert!(!variable_modes.all_inputs());
         let types = type_.types().iter().cloned().sorted().collect_vec();
-        debug_assert!(!types.is_empty());
+        // This instruction that needs to supports empty type-lists
```
dmitrii-ubskii, line 52:
```suggestion
        // This instruction needs to supports empty type-lists
```

executor/Cargo.lock
dmitrii-ubskii, line 1: This file should not exist. There's supposed to be one `Cargo.lock` at the root of the workspace.

dependencies/typedb/repositories.bzl
```diff
@@ -36,5 +36,5 @@ def typedb_behaviour():
     git_repository(
         name = "typedb_behaviour",
         remote = "https://github.com/typedb/typedb-behaviour",
-        commit = "692c099bdb9ac418707be276c120a144219df816",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        commit = "c31a439fcffd7628aebe60ce260ed7164cbf926b",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
```
dmitrii-ubskii, line 39: Is this merged?

executor/tests/execute_function.rs
```diff
@@ -4,7 +4,10 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::HashMap,
+    sync::Arc,
+};
```
dmitrii-ubskii, line 10: Weird reformat

ir/pattern/disjunction.rs
```diff
@@ -36,6 +36,10 @@ impl Disjunction {
     pub fn referenced_variables(&self) -> impl Iterator<Item = Variable> + '_ {
         self.conjunctions().iter().flat_map(|conjunction| conjunction.referenced_variables())
     }
+
+    pub fn optimise_away_failing_branches(&mut self, scope_ids: Vec<ScopeId>) {
+        self.conjunctions.retain(|v| !scope_ids.contains(&v.scope_id()))
+    }
```
dmitrii-ubskii, line 42:
```suggestion
    pub fn optimise_away_failing_branches(&mut self, unsatisfiable: Vec<ScopeId>) {
        self.conjunctions.retain(|v| !unsatisfiable.contains(&v.scope_id()))
    }
```

ir/pattern/constraint.rs
```diff
@@ -2696,3 +2710,62 @@ impl<ID: IrID> fmt::Display for LinksDeduplication<ID> {
         write!(f, "LinksDeduplication({}, {}))", self.links1, self.links2)
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct OptimisedToUnsatisfiable {
+    conjunction: Conjunction,
+}
```
dmitrii-ubskii, line 2717: Can this just hold the scope id? It doesn't look like the conjunction is used anywhere.

executor/tests/compile_execute.rs
```diff
@@ -15,6 +15,7 @@ use compiler::{
         match_inference::infer_types,
     },
     executable::function::ExecutableFunctionRegistry,
+    transformation::transform::apply_transformations,
```
dmitrii-ubskii, line 18: Unused?

ir/pattern/conjunction.rs
```diff
@@ -55,6 +55,18 @@ impl Conjunction {
...
+    pub fn optimise_away(&mut self) {
+        let mut swapped_conjunction = Self::new(self.scope_id);
+        std::mem::swap(&mut self.nested_patterns, &mut swapped_conjunction.nested_patterns);
+        std::mem::swap(
+            self.constraints_mut().constraints_mut(),
+            swapped_conjunction.constraints_mut().constraints_mut(),
+        );
+        self.constraints_mut()
+            .constraints_mut()
+            .push(Constraint::OptimisedToUnsatisfiable(OptimisedToUnsatisfiable::new(swapped_conjunction)))
```
dmitrii-ubskii, line 67: You should be able to just assign to `*self`:
```suggestion
        *self = Self::new(self.scope_id);
        self.constraints
            .constraints_mut()
            .push(Constraint::OptimisedToUnsatisfiable(OptimisedToUnsatisfiable::new(swapped_conjunction)))
```

compiler/executable/match_/planner/vertex/mod.rs
```diff
@@ -510,6 +517,41 @@ impl Costed for ComparisonPlanner<'_> {
...
+    ) -> Self {
+        Self { optimised_unsatisfiable }
+    }
+
+    fn is_valid(&self, ordered: &[VertexId], _graph: &Graph<'_>) -> bool {
+        true
+    }
+
+    fn variables(&self) -> impl Iterator<Item = VariableVertexId> {
+        [].into_iter()
```
dmitrii-ubskii, line 540:
```suggestion
        iter::empty()
```

compiler/transformation/transform.rs
```diff
@@ -61,3 +62,60 @@ fn prune_redundant_roleplayer_deduplication(conjunction: &mut Conjunction, block
...
 }
+
+pub fn optimize_away_statically_unsatisfiable_conjunctions(
+    conjunction: &mut Conjunction,
+    block_annotations: &TypeAnnotations,
+) {
+    optimize_away_statically_unsatisfiable_conjunctions_impl(conjunction, block_annotations);
+}
+
+fn optimize_away_statically_unsatisfiable_conjunctions_impl(
```
dmitrii-ubskii, line 73: ?

compiler/transformation/transform.rs
```diff
@@ -61,3 +62,60 @@ fn prune_redundant_roleplayer_deduplication(conjunction: &mut Conjunction, block
...
+        match nested {
+            NestedPattern::Disjunction(disjunction) => {
+                let mut optimised_unsatisfiable_branch_ids = Vec::new();
+                for branch in disjunction.conjunctions_mut().iter_mut() {
+                    if optimize_away_statically_unsatisfiable_conjunctions_impl(branch, block_annotations) {
+                        optimised_unsatisfiable_branch_ids.push(branch.scope_id())
+                    }
+                }
+                disjunction.optimise_away_failing_branches(optimised_unsatisfiable_branch_ids);
+                must_optimise_away = must_optimise_away || disjunction.conjunctions().is_empty();
```
dmitrii-ubskii, line 88: We should break out here if `true`.

compiler/transformation/transform.rs
```diff
@@ -61,3 +62,60 @@ fn prune_redundant_roleplayer_deduplication(conjunction: &mut Conjunction, block
...
+                optimize_away_statically_unsatisfiable_conjunctions_impl(negation.conjunction_mut(), block_annotations);
+            }
+            NestedPattern::Optional(optional) => {
+                optimize_away_statically_unsatisfiable_conjunctions_impl(optional.conjunction_mut(), block_annotations);
+            }
+        }
+    }
+    let must_optimise_away = must_optimise_away
+        || conjunction.constraints().iter().any(|constraint| {
+            if let Some(constraint_annotation) = block_annotations.constraint_annotations_of(constraint.clone()) {
```
dmitrii-ubskii, line 100: If constraints are `Copy`, we should do this:
```suggestion
        || conjunction.constraints().iter().any(|&constraint| {
            if let Some(constraint_annotation) = block_annotations.constraint_annotations_of(constraint) {
```
If they are not, I question why `constraint_annotations_of` does not accept a reference.

dmitrii-ubskii (approved):
Logic seems sound, left some housekeeping comments.

ir/pattern/constraint.rs
krishnangovindraj: ^ We're deprecating + this might be useful for debugging?

compiler/transformation/transform.rs
krishnangovindraj: We could.
I didn't so that all unsatisfiable sub-patterns would be replaced with an `OptimisedToUnsatisfiable`.

dependencies/typedb/repositories.bzl
krishnangovindraj: Sadly, yes.

ir/pattern/conjunction.rs
krishnangovindraj: I do want the constraints and the nested patterns to be swapped, so I guess you mean `std::mem::swap(self, &mut swapped_conjunction)`?

compiler/transformation/transform.rs
krishnangovindraj: They're not copy. 
I question that and have a TODO comment on top that I need to get to in a different PR.

krishnangovindraj (commented):
replies

ir/pattern/conjunction.rs
dmitrii-ubskii: I thought the `swapped_conjunction` was dropped at the end of the function, I missed that it was in `OptimisedToUnsatisfiable::new`.
In that case, you can swap it with `self`, which I can see you've done in a recent commit.

ir/pattern/conjunction.rs
```diff
@@ -55,6 +55,14 @@ impl Conjunction {
         &mut self.nested_patterns
     }
 
+    pub fn optimise_away(&mut self) {
```
flyingsilverfin, line 58: `set_unsatisfiable` the fact it replaces itself with a different constraint, or sets a boolean or whatever, is implementation detail.

ir/pattern/constraint.rs
```diff
@@ -510,6 +511,7 @@ pub enum Constraint<ID> {
     Plays(Plays<ID>),
     Value(Value<ID>),
     LinksDeduplication(LinksDeduplication<ID>),
+    OptimisedToUnsatisfiable(OptimisedToUnsatisfiable),
```
flyingsilverfin, line 514: `Constraint::Unsatisfiable` Optimisation irrelevant to this data structure. The constraint can also just be `Unsatisfiable`

ir/pattern/constraint.rs
```diff
@@ -2696,3 +2710,62 @@ impl<ID: IrID> fmt::Display for LinksDeduplication<ID> {
...
+    }
+
+    pub fn map<ID: IrID, T: Clone>(self, mapping: &HashMap<ID, T>) -> OptimisedToUnsatisfiable {
+        self
+    }
+}
+
+impl PartialEq<Self> for OptimisedToUnsatisfiable {
+    fn eq(&self, other: &Self) -> bool {
+        self.conjunction.scope_id() == other.conjunction.scope_id()
```
flyingsilverfin, line 2745: I dont think the scope ID should be in here at all - any Unsatisfiable constraint is equal to any other.  Notice how none of the other constraints care about their context/scope when doing equivalence checks.

ir/pattern/constraint.rs
```diff
@@ -2696,3 +2710,62 @@ impl<ID: IrID> fmt::Display for LinksDeduplication<ID> {
...
+
+impl StructuralEquality for OptimisedToUnsatisfiable {
+    fn hash(&self) -> u64 {
+        self.conjunction.scope_id().id as u64
+    }
+
+    fn equals(&self, other: &Self) -> bool {
+        self.conjunction.scope_id() == other.conjunction.scope_id()
+    }
+}
```
flyingsilverfin, line 2765: same thinking here.

compiler/transformation/transform.rs
```diff
@@ -61,3 +62,53 @@ fn prune_redundant_roleplayer_deduplication(conjunction: &mut Conjunction, block
...
+                    }
+                }
+            } else {
+                false
+            }
+        });
+    if must_optimise_away {
+        conjunction.optimise_away();
+    }
+    must_optimise_away
```
flyingsilverfin, line 113: You're returning a boolean when your function implies you're doing a stateful operation on the conjunction - which you are! As a result, after your function runs you now have two sources of truth: the conjunction, and this boolean. This is at risk of going out of sync at some point.

A tighter API is to make the caller determine whether the transformation was applied themselves: here we do `conjunction.set_unsatisfiable()`, and then the caller, if they must know, can check `conjunction.is_unsatisifable()`.

This is also safer for the future, if there are _other_ ways to achieve unsatisifable conjunctions! If a conjunction arrives at this transformation unsatisifiable, it'll still work nicely and combine that information with what it computes here.

flyingsilverfin: Ganesh, and then I, call this making "tight" APIs (you can't misuse them).

compiler/transformation/transform.rs
```diff
@@ -61,3 +62,53 @@ fn prune_redundant_roleplayer_deduplication(conjunction: &mut Conjunction, block
         }
     });
 }
+
+pub fn optimize_away_statically_unsatisfiable_conjunctions(
```
flyingsilverfin, line 66: Our current architecture has each transformation in an independent file, so let's give this one its own called, maybe `unsatisfiability.rs`?

flyingsilverfin (changes requested):

compiler/transformation/transform.rs
krishnangovindraj: > This is at risk of going out of sync at some point.
Sure. I could also not mutate the inner conjunctions and just optimise away the outer-most unsatisfiable conjunction.

compiler/transformation/transform.rs
krishnangovindraj: I broke that pattern two weeks ago. We're going to have a bunch of files containing a 20 line function this way.

PR #7361, opened 2025-02-14
Introduce update queries

farost:
We introduce update queries as a shortcut for `delete + insert`. Using `update`, it is possible to replace an existing `has` or `links` edge of a specific attribute or role type by a new value in one step.

If an old value exists, it is deleted. Then, the new value specified in the query is set. 

This feature is only available for `has` and `links`. Moreover, to make sure that no excessive data is unexpectedly deleted, it only works with types with `owns` and `relates` of cardinalities not exceeding 1: `card(0..1)` (the default value) or `card(1..1)` (card of `@key`).

Usage examples:

```
match
  $p isa person, has name "Bob";
update
  $p has balance 1234.56789dec;
```

```
match
  $alice isa person, has name "Alice";
  $bob isa person, has name "Bob";
  $e isa employment, links (employer: $alice);
update
  $e links (employee: $bob);
```

## Motivation
We want to introduce a shorter query for faster data updates instead of using `delete` and `insert`.

## Implementation
We extend IR, annotation, compilation, and execution with a new stage and a new set of instructions. As update queries behave similarly to insert (they insert new data), many insert-related methods are reused, with additional logic and validation on top. More in the PR comments.


dependencies/typedb/repositories.bzl
```diff
@@ -33,8 +33,13 @@ def typedb_protocol():
     )
 
 def typedb_behaviour():
+    # TODO: Return typedb
```
farost, line 36: https://github.com/typedb/typedb-behaviour/pull/338

compiler/annotation/mod.rs
```diff
@@ -163,35 +163,44 @@ typedb_error!(
             name: String,
             source_span: Option<Span>,
         ),
-        IllegalInsertTypes(
+        IllegalInsertableTypes(
```
farost, line 166: I use some of the methods of `insert`s in `update`s because `update` actually inserts new things with some differences. When reusing `insert`-related things, I call the objects "insertables" for the inner reference. It should not face the users.

compiler/executable/insert/executable.rs
```diff
@@ -258,35 +251,31 @@ fn add_role_players(
...
-            }
-            (Some(_), Some(_)) => unreachable!(),
-        };
+        let role = resolve_links_role(type_annotations, input_variables, variable_registry, &named_role_types, links)?;
         instructions.push(ConnectionInstruction::Links(Links { relation, player, role }));
     }
     Ok(())
 }
 
+pub(crate) fn get_thing_input_position(
```
farost, line 260: Moved from `mod.rs`. It only relates to executables.

compiler/executable/insert/mod.rs
```diff
@@ -43,100 +34,3 @@ pub enum ValueSource {
...
-        })),
-    }
-}
-
-pub(crate) fn get_kinds_from_types(types: &BTreeSet<Type>) -> HashSet<Kind> {
-    types.iter().map(Type::kind).collect()
-}
-
-typedb_error! {
-    pub WriteCompilationError(component = "Write compiler", prefix = "WCP") {
```
farost, line 71: Although I reuse some logical methods right from `insert::executable`, I decided to move the error declaration one layer up because it's more common than just an insert error.

compiler/executable/update/executable.rs
```diff
@@ -0,0 +1,126 @@
...
+        WriteCompilationError,
+    },
+    filter_variants, VariablePosition,
+};
+
+#[derive(Debug)]
+pub struct UpdateExecutable {
+    pub executable_id: u64,
+    // Reuse the insert's concept instruction for attributes. Other isas should be validated earlier
+    pub concept_instructions: Vec<ConceptInstruction>,
```
farost, line 36: `ConnectionInstruction` is different because it is an actually different instruction with a different implementation. However, when inserting new instances of attributes through `update $p has name "Name"`, it's literally the same instruction as for `insert`.

compiler/executable/update/type_check.rs
```diff
@@ -0,0 +1,243 @@
...
+                    snapshot,
+                    type_manager,
+                    links,
+                    input_annotations_variables,
+                    input_annotations_constraints,
+                    insert_annotations.constraint_annotations_of(constraint.clone()).unwrap().as_links(),
+                )?;
+            }
+
+            Constraint::Isa(_)
```
farost, line 59: I'm not sure what data constraints can appear here implicitly, so I just left it this way.

compiler/executable/update/type_check.rs
```diff
@@ -0,0 +1,243 @@
...
+
+fn validate_has_updatable(
+    snapshot: &impl ReadableSnapshot,
+    type_manager: &TypeManager,
+    has: &Has<Variable>,
+    input_annotations_variables: &BTreeMap<Variable, Arc<BTreeSet<answer::Type>>>,
+    input_annotations_constraints: &HashMap<Constraint<Variable>, ConstraintTypeAnnotations>, // Future use
+    left_right: &LeftRightAnnotations,
+) -> Result<(), TypeInferenceError> {
+    insert::type_check::validate_has_insertable(
```
farost, line 89: I was thinking about adding an additional validation stage into the original `insert`'s method I reuse here, so it is validated in one go. However, it would become a mess, please I would not be able to extend the error message.

This way, I decided to have a second run through all the available types and validate the card separately, not affecting the original method.

ir/translation/writes.rs
```diff
@@ -98,3 +93,118 @@ pub fn translate_delete(
     }
     Ok((builder.finish()?, deleted_concepts))
 }
+
+fn validate_update_statements_and_variables(
```
farost, line 97: I try to reject all the invalid expressions in IR.

compiler/executable/update/instructions.rs
```diff
@@ -0,0 +1,37 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use std::fmt;
+
+use crate::executable::insert::{ThingPosition, TypeSource};
+
+#[derive(Debug)]
```
flyingsilverfin, line 11: We not use the general `write_instructions`? They're the same, no?

flyingsilverfin (changes requested):

compiler/executable/update/instructions.rs
flyingsilverfin: I saw your comment earlier, but if they're the same data structure then we can keep them the same IMO, just process them differently

compiler/executable/update/executable.rs
```diff
@@ -0,0 +1,126 @@
...
+) -> Result<UpdateExecutable, Box<WriteCompilationError>> {
+    let mut attributes_inserts = Vec::with_capacity(constraints.len());
+    let variables = add_inserted_concepts(
+        constraints,
+        input_variables,
+        type_annotations,
+        variable_registry,
+        &mut attributes_inserts,
+        source_span,
+    )?;
```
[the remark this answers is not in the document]
farost: We do insert new concepts. Check out my comment on this page:

> However, when inserting new instances of attributes through update $p has name "Name", it's literally the same instruction as for insert.

It happens, for example, for this test: https://github.com/typedb/typedb-behaviour/blob/c9ade0e31ff79ae1d77cde66a87f573e5ecb3cbc/query/language/update.feature#L245

Only attributes can be created in update queries. And they can be created only by doing this. We do not support explicit `isa` statements (basically do not support new variables declaration), only generate `isa` constraints for attributes specified through values.

compiler/executable/update/executable.rs
```diff
@@ -0,0 +1,126 @@
...
+            has.attribute().as_variable().unwrap(),
+            variable_registry,
+            has.source_span(),
+        )?;
+        instructions.push(ConnectionInstruction::Has(Has { owner, attribute }));
+        Ok(())
+    })
+}
+
+fn add_role_players(
```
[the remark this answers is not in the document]
farost: We use `role_players` instead of `links` in many places. Do we want to incline to `links`? I do support it, although it's really mixed now.

compiler/executable/update/instructions.rs
flyingsilverfin: RIght ok - let's leave it at that!

compiler/executable/update/executable.rs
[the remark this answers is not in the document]
flyingsilverfin: Ok that's what I thought, thanks

compiler/executable/update/executable.rs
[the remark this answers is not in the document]
flyingsilverfin: All the new stuff can use links, role_players is legacy and should be renamed generally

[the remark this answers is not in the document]
farost: Done

PR #7365, opened 2025-02-14
Enable functions in write pipelines

### LINKED ISSUE #7363: Enable functions in write pipelines
## Problem to Solve
Enable functions in write pipelines

## Current Workaround
None.

## Proposed Solution
Enable them, but reset function tables after every write stage - ensuring consistency.
Also, add a test which ensures the functions are reset.
```
...
insert
     $x isa number 1;
match
    let $sum1 = some_cyclic_function_accessing_numbers();
insert
     $y isa number == $sum1;
match
     $sum2 = some_cyclic_function_accessing_numbers();
 select $sum2;

```

krishnangovindraj:
Enable functions in write pipelines

## Motivation
Enable functions in write pipelines.
fixes #7363 
Requires typedb/typedb-behaviour#337

## Implementation
* Passes the function registry to the write pipeline.
* Fixes a bug in type_seeder where we didn't recurse into negations within disjunctions.

### DIFF
--- a/function/function_manager.rs
+++ b/function/function_manager.rs
@@ -26,6 +26,7 @@ use encoding::{
 use ir::{
     pattern::{conjunction::Conjunction, constraint::Constraint, nested_pattern::NestedPattern},
     pipeline::{
+        function::ReturnOperation,
         function_signature::{
             FunctionID, FunctionIDAPI, FunctionSignature, FunctionSignatureIndex, HashMapFunctionSignatureIndex,
         },
@@ -44,7 +45,6 @@ use storage::{
     snapshot::{ReadableSnapshot, WritableSnapshot},
 };
 use typeql::common::Spanned;
-use ir::pipeline::function::ReturnOperation;
 
 use crate::{function::SchemaFunction, function_cache::FunctionCache, FunctionError};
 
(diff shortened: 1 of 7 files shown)


function/function_manager.rs
```diff
@@ -338,7 +339,15 @@ fn validate_no_cycles_impl<ID: FunctionIDAPI + Ord + Eq>(
                 | TranslatedStage::Reduce(_)
         )
     });
-    let unnegated_stratum = if this_stage_has_operator { current_stratum.add(1, 1) } else { current_stratum.add(0, 1) };
+    let has_aggregate_return = match function.function_body.return_operation {
```
farost, line 342: This could be a method in `ReturnOperation` (it could probably be useful somewhere else, idk)

farost (approved):
Looks good to me, left some comments

flyingsilverfin (approved):

flyingsilverfin:
It looks right, but where do we do the function table clearing that you had discussed might be needed?

krishnangovindraj:
We don't need to clear because each match stage has its own MatchExecutor that owns the state. So the tables aren't shared in the first place.

PR #7366, opened 2025-02-16
Improve query plan eviction

flyingsilverfin:

We improve the query plan eviction policy from a flat 1% change in total count of the data, to instead be a 25% change of _any individual_ statistic, such as a specific entity type's instances, or the count of `has` between two specific types, etc. This means the query plans will be much more accurate in rapidly changing data, in particular in the common case of building an initial dataset or benchmarking.

Before, we'd get tpcc results for the PAYMENT action that look like this:
```

Execution Results after 63 seconds
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
                  Complete        Time (µs)         minLatMs        p50             p75             p90             p95             p99             maxLatMs        Aborts                   
...
  PAYMENT         20                 61.618         7.90            12.91           1244.86         15993.48        24360.62        24360.62        24360.62        0                    
...
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

This shows that during a 60s execution time, a few pathological query executions would blow up the entire runtime! However, sometimes the query also executed quickly. This was fundamentally caused by the initial queries using the query plan based on old statistics, which would only get updated later in the benchmark when the total statistics updated sufficiently.

## Motivation

We move towards the goal of being able to reliably execute TPCC benchmarks without needing to force query plans to replan manually.

We now reliably get TPCC results for PAYMENT that look like this instead:

```
Execution Results after 60 seconds
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
                  Complete        Time (µs)         minLatMs        p50             p75             p90             p95             p99             maxLatMs        Aborts                
...
  PAYMENT         428                 43.54         7.00            14.05           26.01          184.51          249.36          281.22          675.73          0               
...
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

## Implementation

The query plan cache now retains the last full set of statistics it was flushed against, which allows calculating a much finer diff. We then take the maximum percentage that any individual statistic has changed as the signal of whether to flush the cache. For now, we increase that threshold to 25% of the minum of the old or new statistics, whichever is smaller.

### DIFF
--- a/common/logger/logger.rs
+++ b/common/logger/logger.rs
@@ -6,9 +6,11 @@
 
 #![allow(unexpected_cfgs)]
 
-use tracing::{self, dispatcher::DefaultGuard, metadata::LevelFilter, Level};
+use std::io::stdout;
+
+use tracing::{self, dispatcher::DefaultGuard, Level, metadata::LevelFilter};
 pub use tracing::{debug, error, info, trace};
-use tracing_subscriber::{fmt::SubscriberBuilder, EnvFilter};
+use tracing_subscriber::{EnvFilter, fmt::SubscriberBuilder};
 
 pub mod result;
 
@@ -25,7 +27,21 @@ pub fn initialise_logging_global() {
         // .add_directive("tonic=trace".parse().unwrap());
     ;
 
-    let subscriber = SubscriberBuilder::default().with_max_level(Level::TRACE).with_env_filter(filter).finish();
+    // Create a file appender
+    // let file_appender = File::create("output.log")
+    //     .expect("Failed to create log file");
+
+    let subscriber = SubscriberBuilder::default()
+        .with_max_level(Level::TRACE)
+        .with_env_filter(filter)
+        .with_writer(stdout)
+        // .with_writer(file_appender)
+        .with_ansi(false) // Disable ANSI colors in file output
+        .with_thread_ids(true)
+        .with_file(true)
+        .with_line_number(true)
+        .finish();
+
     tracing::subscriber::set_global_default(subscriber).unwrap()
 }
 
--- a/common/logger/logger.rs
+++ b/common/logger/logger.rs
@@ -8,9 +8,9 @@
 
 use std::io::stdout;
 
-use tracing::{self, dispatcher::DefaultGuard, Level, metadata::LevelFilter};
+use tracing::{self, dispatcher::DefaultGuard, metadata::LevelFilter, Level};
 pub use tracing::{debug, error, info, trace};
-use tracing_subscriber::{EnvFilter, fmt::SubscriberBuilder};
+use tracing_subscriber::{fmt::SubscriberBuilder, EnvFilter};
 
 pub mod result;
 
--- a/concept/thing/statistics.rs
+++ b/concept/thing/statistics.rs
@@ -5,42 +5,41 @@
  */
 
 use std::{
+    cmp::min,
     collections::{BTreeMap, HashMap},
     fmt,
+    hash::Hash,
     ops::Bound,
+    time::Instant,
 };
-use std::cmp::{min};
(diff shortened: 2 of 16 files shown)


common/logger/logger.rs
```diff
@@ -25,7 +27,21 @@ pub fn initialise_logging_global() {
         // .add_directive("tonic=trace".parse().unwrap());
     ;
 
-    let subscriber = SubscriberBuilder::default().with_max_level(Level::TRACE).with_env_filter(filter).finish();
+    // Create a file appender
+    // let file_appender = File::create("output.log")
+    //     .expect("Failed to create log file");
```
flyingsilverfin, line 32: We can add file output soon (TM) so the user actually has log files...

concept/thing/statistics.rs
```diff
@@ -167,6 +175,14 @@ impl Statistics {
...
         }
 
+        let millis = Instant::now().duration_since(start).as_millis();
+        event!(
+            Level::TRACE,
+            "Statistics sync finished in {} ms. Storage watermark was initially: {}. Current statistics sequence is from: {}",
+            millis,
+            storage_watermark,
+            start_seq_nr
+        );
```
flyingsilverfin, line 185: Really useful when investigating whether statistics is able to keep up with the current 100ms update interval @dmitrii-ubskii  @cxdorn

docker/BUILD
```diff
@@ -44,13 +44,15 @@ docker_container_run_and_commit(
     image = "@ubuntu-22.04-x86_64//image",
     commands = docker_image_prepare_commands,
     target_compatible_with = constraint_linux_x86_64,
+    tags = ["manual"],
```
flyingsilverfin, line 47: Docker builds may not work on local machines now?

query/query_cache.rs
```diff
@@ -7,33 +7,39 @@
...
     perf_counters::QUERY_CACHE_FLUSH,
 };
+use storage::sequence_number::SequenceNumber;
 use structural_equality::StructuralEquality;
 
 #[derive(Debug)]
 pub struct QueryCache {
     cache: Cache<IRQuery, ExecutablePipeline>,
-    statistics_size: AtomicU64,
+    last_statistics: Mutex<Statistics>,
```
flyingsilverfin, line 36: We have to keep the last used statistics the query cache is operating against, so we can calculate a long-term diff. This isn't really much  more expensive than what we were doing before, since we cloned the entire Statistics on every update anyway.

resource/constants.rs
```diff
@@ -61,7 +61,7 @@ pub mod server {
 pub mod database {
     use std::time::Duration;
 
-    pub const QUERY_PLAN_CACHE_FLUSH_STATISTICS_CHANGE_PERCENT: f64 = 0.05;
+    pub const QUERY_PLAN_CACHE_FLUSH_ANY_STATISTIC_CHANGE_FRACTION: f64 = 0.25;
```
flyingsilverfin, line 64: Raised this to 25% since it now applies to _any_ individual entity/relation/attribute/has/relation->role/role->player statistic! 

In general what we want to ensure is that if we have a big increase in "cost" of planning, we replan. Critically, this means we really need to replan when we go from say 1 -> 2 elements (since a plan which includes 5 stacked 1 element iteration steps is going to behave quite different to 5 stacked 2 element iteration steps = 2x2x2x2x2 = 32 elements! The bigger the diff between the old plan and the new number of elements, the worse can plan can be, obviously!)

docker/BUILD
farost: `bazel //...` worked for me on Mac though, `target_compatible_with` was enough. I don't really understand why you commented out it below.

PR #7367, opened 2025-02-17
Improve performance using storage seeks and enhance profiling tools

flyingsilverfin:

We propagate "seekable" iterators throughout the query executors, which allows us to trade off doing more seeks for doing more advances through the rocksdb iterators. We also extend the ability to profile performance of various operations, such as query execution and compiling, and transaction commits.

## Motivation

We improve TypeDB's performance by doing two main things:
1. extending the architecture of the execution paths to support 'seeks', which makes intersections more efficient
2. incorporate more information into the prefixes used to look up data on disk, which means less data is scanned for each operation

## Implementation

- We implement seek operations for all data iterators used by the `//executor` package
- We add more information into the tuple iterators to allow better seeking back through their inner iterators
- We extend prefix information with lower bound and upper bound information of values and types, where they were missing before
- We enhance operation profiling dramatically:
  - introduce StorageCounters throughout the code base to track low-level operations, and allowing us to test how much work is being induced
  - introduce CompileProfile to record how much time is taken in the translation, annotation, and compilation (eg. planning) phases
  - introduce the CommitProfile to record the time of operations in the commit phase

To enable query profiles, set the `//query` package to `TRACE`, and to enable Comit profiles, set the `//database` package to TRACE.

### Bug fixes

- We fix one bootup performance bug, by only reading from the last WAL file, instead of from the beginning of time
- We fix one critical statistics bug, which manifested as some records not being applied to statistics at all. This could lead to misaligned statistics and bad query plans!


common/logger/logger.rs
```diff
@@ -6,36 +6,40 @@
...
         // useful for debugging what tonic is doing:
         // .add_directive("tonic=trace".parse().unwrap());
     ;
 
+    // TODO: remove
     // Create a file appender
-    // let file_appender = File::create("output.log")
-    //     .expect("Failed to create log file");
+    let file_appender = File::create("output.log")
+        .expect("Failed to create log file");
```
flyingsilverfin, line 36: This needs to be removed before merging!

database/transaction.rs
```diff
@@ -72,6 +76,7 @@ impl<D: DurabilityClient> TransactionRead<D> {
             query_manager,
             database,
             transaction_options,
+            profile: TransactionProfile::new(tracing::enabled!(Level::TRACE))
```
flyingsilverfin, line 79: To enable transaction profiles we have to set //database to TRACE. Now only containing Commit, but we could also add Open (causality await is there, for example, and that could be expensive in concurrent scenarios... and we'd want to find that out)

executor/tests/efficiency.rs
flyingsilverfin, line 1: This is where the work started!! Pretty good but nuanced test IMO

query/query_manager.rs
```diff
@@ -80,6 +80,9 @@ impl QueryManager {
         source_query: &str,
     ) -> Result<Pipeline<Snapshot, ReadPipelineStage<Snapshot>>, Box<QueryError>> {
         event!(Level::TRACE, "Running read query:\n{}", query);
+        let mut query_profile = QueryProfile::new(tracing::enabled!(Level::TRACE));
```
flyingsilverfin, line 83: Query profiles are currently detatched from transcation profiles, though they could be linked!

resource/profile.rs
```diff
@@ -0,0 +1,767 @@
...
+        Arc, RwLock,
+    },
+    time::{Duration, Instant},
+};
+use std::fmt::{Display, Formatter};
+
+use itertools::Itertools;
+
+#[derive(Debug)]
+pub struct TransactionProfile {
```
flyingsilverfin, line 21: I would like to extend this with an 'OpenProfile' as well, since we have to do the causality check + wait there

resource/profile.rs
```diff
@@ -0,0 +1,767 @@
...
+                    - data.snapshot_put_statuses_check_nanos
+                    - data.snapshot_commit_record_create_nanos
+                    - data.snapshot_durable_write_data_submit_nanos
+                    - data.snapshot_isolation_validate_nanos
+                    - data.snapshot_durable_write_data_confirm_nanos
+                    - data.snapshot_storage_write_nanos
+                    - data.snapshot_isolation_manager_notify_nanos
+            },
+        }
+    }
```
flyingsilverfin, line 271: This is a very unsophisticated approach, and will break immediately if we change the order of commits. If we can find something as simply/fast/roboust to skipping individual steps, that would work for me too!

resource/profile.rs
```diff
@@ -0,0 +1,767 @@
...
+            schema_update_caches_update_nanos: 0,
+            schema_update_statistics_update_nanos: 0,
+            total_nanos: 0,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct QueryProfile {
+    compile_profile: CompileProfile,
```
flyingsilverfin, line 381: New: compileProfile

resource/profile.rs
```diff
@@ -0,0 +1,742 @@
...
+    translation: Duration,
+    validation: Duration,
+    annotation: Duration,
+    compilation: Duration,
+}
+
+#[derive(Debug)]
+pub struct EncodingProfile {
+    storage_counters: StorageCounters,
+}
```
dmitrii-ubskii, line 494: Profiling how many storage operations it took to encode responses.
Name options were between `ResponseProfile` and `EncodingProfile`, and the latter felt more precise.

concept/type_/type_manager/type_reader.rs
```diff
@@ -55,6 +58,7 @@ use crate::{
 // non-instantiable
 pub enum TypeReader {}
 
+// TODO: this file needs to use StorageCounters correctly
```
dmitrii-ubskii, line 61: I've left this TODO hanging as pulling the storage counters here would be a _massive_ change, and most of these are very cheap, so not worth the time investment.

executor/read/collecting_stage_executor.rs
```diff
@@ -215,7 +216,10 @@ impl CollectorTrait for SortCollector {
     fn into_iterator(self, context: &ExecutionContext<impl ReadableSnapshot>) -> CollectedStageIterator {
         let Self { sort_on, collector } = self;
         let unsorted = collector.unwrap();
-        let sorted_indices = unsorted.indices_sorted_by(context, &sort_on).into_iter().peekable();
+        let profile = context.profile.profile_stage(|| String::from("Sort"), 0); // TODO executable id
```
dmitrii-ubskii, line 219: There's a question of what to use as an executable id (or if we even need one) for stages not associated with an executable. There are a few other places this comes up, notably in Reduce and schema queries (which don't have stages).

query/query_manager.rs
```diff
@@ -48,14 +51,35 @@ impl QueryManager {
         source_query: &str,
     ) -> Result<(), Box<QueryError>> {
         event!(Level::TRACE, "Running schema query:\n{}", query);
-        match query {
+        let query_profile = QueryProfile::new(tracing::enabled!(Level::TRACE));
```
dmitrii-ubskii, line 54: Validation of new / changed schema constraints gets expensive fast.

resource/profile.rs
flyingsilverfin: Agree, looks good to me

concept/type_/type_manager/type_reader.rs
flyingsilverfin: Agree as well!

executor/read/collecting_stage_executor.rs
flyingsilverfin: I guess we don't need one, just use 0 for now i spose

query/query_manager.rs
flyingsilverfin: For sure lol

common/logger/logger.rs
```diff
@@ -17,25 +17,27 @@ pub mod result;
...
-        // .add_directive("database=trace".parse().unwrap())
-        // .add_directive("server=trace".parse().unwrap())
+        .add_directive("database=trace".parse().unwrap())
+        // .add_directive("durability=trace".parse().unwrap())
+        .add_directive("server=trace".parse().unwrap())
         // .add_directive("storage=trace".parse().unwrap())
+        // .add_directive("concept=trace".parse().unwrap())
         // .add_directive("executor=trace".parse().unwrap())
-        // .add_directive("query=trace".parse().unwrap())
+        .add_directive("query=trace".parse().unwrap())
```
flyingsilverfin, line 26: We can set these back and disable the File appender for the the merge

flyingsilverfin (commented):
Looks ok to me!

PR #7368, opened 2025-02-19
Introduce docker snapshot jobs

farost:
Introduce docker snapshot jobs.

## Motivation
To be able to verify docker builds before releasing.

## Implementation
We use the same approach as for release jobs. Following the common practice for snapshot artifacts, commit shas are used for versions.
The implementation uses some extent of copypaste we could shorten by introducing new rules in the bazel-distribution repo (which would also help structure the rules better), but it would require a noticeable amount of effort now.

[Example build](https://app.circleci.com/pipelines/github/typedb/typedb/1749/workflows/afa07099-778a-4f50-998b-116cb40c56da) (you can actually pull & run this snapshot through `typedb/typedb-snapshot:*commit*` or `typedb/typedb-snapshot:latest`)


.circleci/config.yml
```diff
@@ -121,7 +122,32 @@ commands:
...
+          # TODO after beta: bazel run //:deploy-docker-snapshot-overwrite-latest-tag
+
+  # TODO: Migrate to bazel sh_binary
+  deploy-docker-snapshot-multiarch:
+    steps:
+      - run: |
+          docker login -u $REPO_DOCKER_USERNAME -p $REPO_DOCKER_PASSWORD
+          export VERSION=$(git rev-parse HEAD)
+          export DOCKER_ORG="typedb"
+          export DOCKER_REPO="typedb-snapshot"
```
farost, line 145: I could suggest moving it into CircleCI vars.

.circleci/config.yml
```diff
@@ -121,7 +122,32 @@ commands:
...
+          docker manifest create $DOCKER_ORG/$DOCKER_REPO:$VERSION --amend $DOCKER_ORG/$DOCKER_REPO:$VERSION-x86_64 --amend $DOCKER_ORG/$DOCKER_REPO:$VERSION-arm64
+          docker manifest push $DOCKER_ORG/$DOCKER_REPO:$VERSION
+          # TODO: After beta: docker manifest push $DOCKER_ORG/$DOCKER_REPO:latest
```
farost, line 148: @krishnangovindraj what's this about? (it's a copypaste from the release job)

docker/BUILD
```diff
@@ -51,7 +51,7 @@ docker_container_run_and_commit(
     name = "typedb-ubuntu-22.04-arm64",
     image = "@ubuntu-22.04-arm64//image",
     commands = docker_image_prepare_commands,
-#    target_compatible_with = constraint_linux_arm64,
+    target_compatible_with = constraint_linux_arm64,
```
farost, line 54: Not sure why Joshua has commented it out. Maybe he struggled without the manual tag and forgot to remove this comment.

.circleci/config.yml
```diff
@@ -121,7 +122,33 @@ commands:
...
+      image-arch:
+        type: string
+    steps:
+      - run: |
+          docker login -u $REPO_DOCKER_USERNAME -p $REPO_DOCKER_PASSWORD
+          export VERSION=$(git rev-parse HEAD)
+          bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
+          bazel run --jobs=8 --@io_bazel_rules_docker//transitions:enable=false --platforms=//docker:linux-<<parameters.image-arch>> \
+            --define container-version="$VERSION-<<parameters.image-arch>>" //:deploy-docker-snapshot-<<parameters.image-arch>>
+          # TODO after beta: bazel run //:deploy-docker-snapshot-overwrite-latest-tag
```
farost, line 136: We don't have a rule for it, and I decided to skip this step. I only upload the latest tag for the combined image.

BUILD
```diff
@@ -311,6 +311,37 @@ docker_container_image(
...
+    name = "deploy-docker-snapshot-x86_64",
+    format = "Docker",
+    image = ":assemble-docker-x86_64",
+    registry = deployment_docker["docker.index"],
+    repository = "{}/{}".format(
+        deployment_docker["docker.organisation"],
+        deployment_docker["docker.snapshot.repository"],
+    ),
+    # using $(version) propagates to `assemble-docker` and breaks the image so it doesn't boot
+    tag = "$(container-version)",
```
krishnangovindraj, line 326: How does this work? Won't the two jobs overwrite each other's output?

deployment.bzl
```diff
@@ -5,6 +5,7 @@
 deployment = {
   'docker.index': 'index.docker.io',
   'docker.organisation': 'typedb',
+  'docker.snapshot.repository': 'typedb-snapshot',
```
krishnangovindraj, line 8: We should figure out whether we want to push snapshots to dockerhub or cloudsmith.

krishnangovindraj (commented):
more comments

BUILD
farost: They cannot overwrite the output because:
1. They can be run only on different machines suitable for the arch
2. The images are different

Also, it's a tag, not a tag file. Maybe I don't understand the question

deployment.bzl
farost: Good point. I agree that we should push it to cloudsmith, so I'll prepare the change. If Joshua disagrees, I'll roll back.

.circleci/config.yml
farost: Okay, I'll have a look!

krishnangovindraj: Anywho, we don't care about the latest tag in snapshots.

.circleci/config.yml
```diff
@@ -143,7 +170,8 @@ commands:
           export DOCKER_REPO="typedb"
           docker manifest create $DOCKER_ORG/$DOCKER_REPO:$VERSION --amend $DOCKER_ORG/$DOCKER_REPO:$VERSION-x86_64 --amend $DOCKER_ORG/$DOCKER_REPO:$VERSION-arm64
           docker manifest push $DOCKER_ORG/$DOCKER_REPO:$VERSION
-          # TODO: After beta: docker manifest push $DOCKER_ORG/$DOCKER_REPO:latest
+          docker manifest create $DOCKER_ORG/$DOCKER_REPO:latest --amend $DOCKER_ORG/$DOCKER_REPO:$VERSION-x86_64 --amend $DOCKER_ORG/$DOCKER_REPO:$VERSION-arm64
+          docker manifest push $DOCKER_ORG/$DOCKER_REPO:latest
```
krishnangovindraj, line 174: Nice thanks

deployment.bzl
farost: Struggle with it for now, may drop the idea https://discord.com/channels/665254494820368395/672580553677078548/1342086641862312037

We may need to preserve the PR if we are not satisfied with pushes to docker hub. It may blow up fast, and we can only manually clean the images ourselves through `prune` or whatever... Which is annoying.

.circleci/config.yml
flyingsilverfin: Either you should remove the commented line though, or implement it, based on the decision you guys take.

.circleci/config.yml
```diff
@@ -379,6 +436,24 @@ workflows:
             branches:
               only: [ "3.0" ]
 
+      - deploy-docker-snapshot-x86_64:
+          filters:
+            branches:
+              only: [ "3.0" ]
```
flyingsilverfin, line 442: you can add `development` as well (and to the other ones that are `3.0` only right now - I will switch then soon (!))

docker/BUILD
flyingsilverfin: Oh yeah my bad, thanks!

flyingsilverfin (approved):
Approved with comments

.circleci/config.yml
farost: We'd better update this completely, removing 3.0 from this list. The switch should be complete

.circleci/config.yml
farost: I'll remove it, whatever

PR #7371, opened 2025-02-20
Avoid writing tombstones for temporary concepts

dmitrii-ubskii:

When deleting an entity or an attribute that was inserted in the same transaction (i.e. never committed), we remove the key from the write buffer rather than create a tombstone. This logic was already present for relations.

## Motivation

Inserting and deleting an entity within a transaction would write a tombstone without a corresponding insert. Statistics would read that tombstone as a delta of -1 and, at best, undercount the population of its type, and at worst underflow the count and crash.

## Implementation

We've pushed this logic to thing manager, and removed `uninsert_{entity,relation}` and made `unput_attribute` private to reduce room for mistakes.

### DIFF
--- a/concept/iterator.rs
+++ b/concept/iterator.rs
@@ -129,6 +129,7 @@ macro_rules! edge_iterator {
         impl Iterator for $name {
             type Item = Result<$mapped_type, Box<$crate::error::ConceptReadError>>;
             fn next(&mut self) -> Option<Self::Item> {
+                use ::lending_iterator::LendingIterator;
                 use $crate::error::ConceptReadError::SnapshotIterate;
                 self.snapshot_iterator.as_mut()?.next().map(|result| {
                     result
--- a/concept/thing/thing_manager.rs
+++ b/concept/thing/thing_manager.rs
@@ -2559,10 +2559,15 @@ impl ThingManager {
         snapshot: &mut impl WritableSnapshot,
         attribute: Attribute,
     ) -> Result<(), Box<ConceptWriteError>> {
-        let key = attribute.vertex().into_storage_key().into_owned_array();
-        match attribute.get_status(snapshot, self) {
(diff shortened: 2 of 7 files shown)


concept/iterator.rs
```diff
@@ -129,6 +129,7 @@ macro_rules! edge_iterator {
         impl Iterator for $name {
             type Item = Result<$mapped_type, Box<$crate::error::ConceptReadError>>;
             fn next(&mut self) -> Option<Self::Item> {
+                use ::lending_iterator::LendingIterator;
```
farost, line 132: ?

concept/thing/thing_manager.rs
```diff
@@ -2555,23 +2560,16 @@ impl ThingManager {
         attribute: Attribute,
     ) -> Result<(), Box<ConceptWriteError>> {
         let key = attribute.vertex().into_storage_key().into_owned_array();
-        snapshot.delete(key);
```
farost, line 2558: Interestingly, we have
```
    pub(crate) fn lock_existing_attribute(&self, snapshot: &mut impl WritableSnapshot, attribute: &Attribute) {
        snapshot.unmodifiable_lock_add(attribute.vertex().into_storage_key().into_owned_array())
    }
```
which is called for `set_required` on strings and structs
But we don't ever remove it for attributes. If I understand correctly, it could wrongly fail another transaction getting rid of this attribute, right? A very-very small chance of this happening, but still.

Your PR does not affect this though, so I'm good. Let's have a look / an issue created though

farost (approved):
Looks good, asked related questions

concept/iterator.rs
dmitrii-ubskii: The macro uses this trait, so the macro should be the one importing it where needed.

PR #7373, opened 2025-02-21
Implement like & contains comparators

krishnangovindraj:
Implement like & contains comparators for strings.
- The 'like' operator expects the right operand to be a string literal which is a valid rust regex. The left operand may be a string literal or a variable holding a string value.
- The 'contains' operator first performs [case-folding](https://www.w3.org/International/wiki/Case_folding) on either operand, then checks whether the (folded) left operand contains the (folded) right operand. Both operands may be string literals or variables holding string values.

## Motivation
Feature completeness
requires: https://github.com/typedb/typedb-dependencies/pull/568
requires: https://github.com/typedb/typedb-behaviour/pull/340 

## Implementation
Implement like & contains comparators for strings


executor/instruction/mod.rs
```diff
@@ -875,8 +875,14 @@ impl<T> Checker<T> {
...
                         Comparator::GreaterOrEqual => |a, b| a >= b,
-                        Comparator::Like => unimplemented_feature!(ComparatorLike),
-                        Comparator::Contains => unimplemented_feature!(ComparatorContains), // |a,b| a.unwrap_string_ref().contains(b.unwrap_string_ref()),
+                        Comparator::Like => |a, b| {
+                            regex::Regex::new(b.unwrap_string_ref())
+                                .expect("Invalid regex should have been caught at compile time")
+                                .is_match(a.unwrap_string_ref())
+                        },
+                        Comparator::Contains => |a, b| {
+                            a.unwrap_string_ref().to_lowercase().contains(b.unwrap_string_ref().to_lowercase().as_str())
```
dmitrii-ubskii, line 884: We should be using something like https://docs.rs/unicase/latest/unicase/ rather than `.to_lowercase()`

executor/instruction/mod.rs
```diff
@@ -875,8 +875,14 @@ impl<T> Checker<T> {
                         Comparator::Greater => |a, b| a > b,
                         Comparator::LessOrEqual => |a, b| a <= b,
                         Comparator::GreaterOrEqual => |a, b| a >= b,
-                        Comparator::Like => unimplemented_feature!(ComparatorLike),
-                        Comparator::Contains => unimplemented_feature!(ComparatorContains), // |a,b| a.unwrap_string_ref().contains(b.unwrap_string_ref()),
+                        Comparator::Like => |a, b| {
+                            regex::Regex::new(b.unwrap_string_ref())
+                                .expect("Invalid regex should have been caught at compile time")
+                                .is_match(a.unwrap_string_ref())
```
dmitrii-ubskii, line 881: We're recompiling the regex every time we check? Could we cache the `Regex` during translation?

compiler/executable/match_/planner/plan.rs
```diff
@@ -575,8 +575,8 @@ impl<'a> ConjunctionPlanBuilder<'a> {
                 Comparator::NotEqual => (), // no tangible impact on traversal costs
                 Comparator::Less | Comparator::LessOrEqual => rhs.add_upper_bound(lhs),
                 Comparator::Greater | Comparator::GreaterOrEqual => rhs.add_lower_bound(lhs),
-                Comparator::Like => unimplemented_feature!(ComparatorLike),
-                Comparator::Contains => unimplemented_feature!(ComparatorContains),
+                Comparator::Like => (),
+                Comparator::Contains => (),
```
dmitrii-ubskii, line 579: `like` and `contains` don't reduce the total cost, but they impact the `io_ratio` by filtering out rows. The planner should take that into account.
This is probably out of scope here, but warrants at least a TODO comment.

dmitrii-ubskii (approved):

krishnangovindraj (commented):
Forgot to publish self-review

executor/instruction/mod.rs
krishnangovindraj: Done

executor/instruction/mod.rs
krishnangovindraj: Tried but it smelt wrong throughout, raising the question as to whether we want to treat `like` as a comparator between two `String` values or some different operator class.

flyingsilverfin (approved):

PR #7374, opened 2025-02-21
Avoid writing tombstones for temporary relation indices

farost:
Fix relation index eager deletes to avoid writing tombstones for temporary relation indices and corrupting statistics.

## Motivation
Creating and removing role players within a transaction would write a tombstone of a relation index without a corresponding put. Statistics would read that tombstone as a delta of -1 and, at best, undercount the population of its type, and at worst underflow the count and crash.

## Implementation
We replicate the logic of `unset_has` and `unset_links` for the relation index deletions (which we missed before). 
First, we clean up the snapshot in case there are new index updates. Then, if the `links` record had been persisted, we also record a delete for the index. We do not read the storage to search for the index for optimization purposes.

### DIFF
--- a/concept/thing/thing_manager.rs
+++ b/concept/thing/thing_manager.rs
@@ -60,7 +60,7 @@ use resource::constants::{
 };
 use storage::{
     key_range::{KeyRange, RangeEnd, RangeStart},
-    key_value::{StorageKey, StorageKeyArray},
+    key_value::{StorageKey, StorageKeyArray, StorageKeyReference},
     snapshot::{lock::create_custom_lock_key, write::Write, ReadableSnapshot, WritableSnapshot},
 };
 
@@ -2680,17 +2680,10 @@ impl ThingManager {
         let has_reverse_array =
             ThingEdgeHasReverse::new(attribute.vertex(), owner.vertex()).into_storage_key().into_owned_array();
 
-        if let Some(has_write) = snapshot.get_write(has.as_reference()).cloned() {
-            match has_write {
-                Write::Put { value, .. } => {
-                    snapshot.unput_val(has_array.clone(), value.clone());
-                    snapshot.unput_val(has_reverse_array.clone(), value);
-                }
-                Write::Delete => {}
-                Write::Insert { .. } => {
-                    unreachable!("Encountered an `insert` has. Owner: {owner:?}, attribute: {attribute:?}.")
-                }
-            }
+        let snapshot_value_opt = Self::get_snapshot_put_value(snapshot, has.as_reference());
+        if let Some(snapshot_value) = snapshot_value_opt {
+            snapshot.unput_val(has_array.clone(), snapshot_value.clone());
+            snapshot.unput_val(has_reverse_array.clone(), snapshot_value);
         }
 
         if owner_status != ConceptStatus::Inserted {
@@ -2826,17 +2819,12 @@ impl ThingManager {
                 .into_storage_key()
                 .into_owned_array();
 
-        if let Some(links_write) = snapshot.get_write(links.as_reference()).cloned() {
-            match links_write {
-                Write::Put { value, .. } => {
-                    snapshot.unput_val(links_array.clone(), value.clone());
-                    snapshot.unput_val(links_reverse_array.clone(), value);
-                }
-                Write::Delete => {}
-                Write::Insert { .. } => {
-                    unreachable!("Encountered an `insert` links. Relation: {relation:?}, player: {player:?}.")
-                }
-            }
+        let mut was_persisted = false;
+
+        let snapshot_value_opt = Self::get_snapshot_put_value(snapshot, links.as_reference());
+        if let Some(snapshot_value) = snapshot_value_opt {
+            snapshot.unput_val(links_array.clone(), snapshot_value.clone());
+            snapshot.unput_val(links_reverse_array.clone(), snapshot_value);
         }
 
         if relation_status != ConceptStatus::Inserted {
@@ -2844,6 +2832,7 @@ impl ThingManager {
                 .has_role_player(snapshot, relation, player, role_type)
                 .map_err(|typedb_source| ConceptWriteError::ConceptRead { typedb_source })?
             {
+                was_persisted = true;
                 snapshot.delete(links_array);
                 snapshot.delete(links_reverse_array);
             }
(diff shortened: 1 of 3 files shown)


concept/thing/thing_manager.rs
```diff
@@ -2931,35 +2920,57 @@ impl ThingManager {
...
         player: impl ObjectAPI,
         role_type: RoleType,
+        is_index_persisted: bool, // An optimization to avoid another storage lookup
     ) -> Result<(), Box<ConceptWriteError>> {
         let players = relation
             .get_players(snapshot, self)
             .map_ok(|(roleplayer, _count)| (roleplayer.player(), roleplayer.role_type()));
         for rp in players {
             let (rp_player, rp_role_type) = rp?;
-            debug_assert!(!(rp_player == Object::new(player.vertex()) && role_type == rp_role_type));
```
farost, line 2946: Looks like we just expected this to fail once we introduce ordering so we can implement it differently?..

concept/thing/thing_manager.rs
```diff
@@ -2931,35 +2920,57 @@ impl ThingManager {
...
                 rp_role_type.vertex().type_id_(),
                 role_type.vertex().type_id_(),
-            );
-            snapshot.delete(index_reverse.into_storage_key().into_owned_array());
+            )
+            .into_storage_key();
+            let snapshot_index_reverse_value_opt = Self::get_snapshot_put_value(snapshot, index_reverse.as_reference());
+            let index_reverse_array = index_reverse.into_owned_array();
+            if let Some(snapshot_index_reverse_value) = snapshot_index_reverse_value_opt {
+                snapshot.unput_val(index_reverse_array.clone(), snapshot_index_reverse_value);
```
farost, line 2962: The reverse index has another value... See the method below

concept/thing/thing_manager.rs
```diff
@@ -3024,4 +3035,20 @@ impl ThingManager {
...
+    pub(crate) fn get_snapshot_put_value(
+        snapshot: &mut impl WritableSnapshot,
+        key: StorageKeyReference<'_>,
+    ) -> Option<ByteArray<BUFFER_VALUE_INLINE>> {
+        match snapshot.get_write(key).cloned() {
+            Some(index_write) => match index_write {
+                Write::Put { value, .. } => Some(value),
+                Write::Delete => None,
+                Write::Insert { .. } => {
+                    unreachable!("Encountered an `insert` while a `put` was expected in the snapshot.")
```
farost, line 3048: Less descriptive, but I don't really expect such crashes. And we already have enough concept tests to show it if it's wrong.

farost:
20 successful runs, I think it should be fine now

concept/thing/thing_manager.rs
flyingsilverfin: not sure!! I think possibly lol you can have multiple players of the same role type - this needs a count tbh. looks good to me.

flyingsilverfin (approved):

concept/thing/thing_manager.rs
```diff
@@ -3024,4 +3035,20 @@ impl ThingManager {
...
+                }
+            },
+            None => None,
+        }
```
dmitrii-ubskii, line 3052:
```suggestion
        match snapshot.get_write(key).cloned()? {
            Write::Put { value, .. } => Some(value),
            Write::Delete => None,
            Write::Insert { .. } => {
                unreachable!("Encountered an `insert` while a `put` was expected in the snapshot.")
            }
        }
```

farost: TIL O_O

concept/thing/thing_manager.rs
```diff
@@ -2931,35 +2920,57 @@ impl ThingManager {
...
     // * Clean up all parts of a relation index to do with a specific role player
     //   after the player has been deleted.
-    pub(crate) fn relation_index_player_deleted(
+    pub(crate) fn relation_index_player_links_unset(
         &self,
         snapshot: &mut impl WritableSnapshot,
         relation: Relation,
         player: impl ObjectAPI,
         role_type: RoleType,
+        is_index_persisted: bool, // An optimization to avoid another storage lookup
```
dmitrii-ubskii, line 2929: This only signifies that `links` was persisted. `index` may or may not have been.

concept/thing/thing_manager.rs
```diff
@@ -2931,35 +2920,57 @@ impl ThingManager {
...
+
+            if is_index_persisted {
+                let is_same_rp = rp_player == Object::new(player.vertex()) && rp_role_type == role_type;
+                if is_same_rp {
+                    snapshot.delete(index_array);
+                } else {
+                    snapshot.delete(index_array);
+                    snapshot.delete(index_reverse_array);
+                }
+            }
```
dmitrii-ubskii, line 2973: Say we have some relation `R` relating entity `A` in the snapshot.
We add player `B` to it, then in a separate query delete player `A`.
`R links (A)` was persisted, so `is_index_persisted` is true. Index `A-B-R-r-r` wasn't though, it was only buffered and will now be replaced by a tombstone.
I'm pretty sure we have to check the disk regardless of if `links` was persisted.

dmitrii-ubskii (changes requested):
I don't think this logic is correct.

concept/thing/thing_manager.rs
farost: Faaaiiiirrrrr I will check the storage as well

concept/thing/thing_manager.rs
```diff
@@ -2943,23 +2929,47 @@ impl ThingManager {
...
+            let snapshot_index_reverse_value_opt = Self::get_snapshot_put_value(snapshot, index_reverse.as_reference());
+            let index_reverse_array = index_reverse.into_owned_array();
+            if let Some(snapshot_index_reverse_value) = snapshot_index_reverse_value_opt {
+                snapshot.unput_val(index_reverse_array.clone(), snapshot_index_reverse_value);
+            }
+
+            let is_persisted = snapshot
+                .get_mapped(index.as_reference(), |_| true)
+                .map_err(|err| Box::new(ConceptReadError::SnapshotGet { source: err }))?
+                .unwrap_or(false);
```
dmitrii-ubskii, line 2964: Small nit, I think this reads better with `.is_some()`
```suggestion
            let is_persisted = snapshot
                .get_mapped(index.as_reference(), |_| ())
                .map_err(|err| Box::new(ConceptReadError::SnapshotGet { source: err }))?
                .is_some();
```

dmitrii-ubskii (approved):
LGTM!

concept/thing/thing_manager.rs
farost: Done + covered with a test

concept/thing/thing_manager.rs
farost: True, it was a little copypaste

PR #7375, opened 2025-02-24
Fix erroneously selected named variables

dmitrii-ubskii:

We adjust the computation of bound variables after each stage in query execution and keep better track of selected variables.

## Motivation

Before this change, all named variables referenced in a match block would be marked as being output at the end of the stage. That was incorrect and led to anything from extra empty columns in the output rows to spurious crashes.

## Implementation

We introduce computation of a pattern's dependency on a variable: when a pattern is referencing a variable, it may be in a way that:
- requires it to be bound by another pattern (`Required`),
- may bind the variable to a new value if previously unbound (`Producing`),
- neither requires it to be bound nor may bind it (`Referencing`).

An example of the latter is a disjunction that references a variable in just one branch: `{ $x has $a; } or { $y has $a; };`. This disjunction can produce `$a`, but only references `$x` and `$y` which may either be fully internal to either branch or bound in the enclosing conjunction.

Only input variables and named variables marked as produced by the root conjunction of a match are selected at the end of the block. (NB: anonymous variables may be selected by an anonymous function in fetch).

To simplify the planning portion, we also revert disjunctions optionally binding variables referenced in a subset of branches.

### DIFF
--- a/compiler/executable/match_/planner/plan.rs
+++ b/compiler/executable/match_/planner/plan.rs
@@ -133,6 +133,14 @@ fn make_builder<'a>(
                         )
                     })
                     .collect::<Result<Vec<_>, _>>()?,
+                disjunction
+                    .required_inputs()
+                    .chain(
+                        disjunction
+                            .optional_outputs()
+                            .filter(|out| conjunction.variable_dependency_modes()[out].is_produced()),
+                    )
+                    .collect(),
             )),
             NestedPattern::Negation(negation) => negation_subplans.push(
                 make_builder(
@@ -1928,11 +1936,12 @@ impl ConjunctionPlan<'_> {
 #[derive(Clone, Debug)]
 pub(super) struct DisjunctionPlanBuilder<'a> {
     branches: Vec<ConjunctionPlanBuilder<'a>>,
+    required_inputs: Vec<Variable>,
 }
 
 impl<'a> DisjunctionPlanBuilder<'a> {
-    pub(super) fn new(branches: Vec<ConjunctionPlanBuilder<'a>>) -> Self {
-        Self { branches }
+    fn new(branches: Vec<ConjunctionPlanBuilder<'a>>, required_inputs: Vec<Variable>) -> Self {
+        Self { branches, required_inputs }
     }
 
     pub(super) fn branches(&self) -> &[ConjunctionPlanBuilder<'a>] {
@@ -1950,6 +1959,10 @@ impl<'a> DisjunctionPlanBuilder<'a> {
             .collect::<Result<Vec<_>, _>>()?;
         let cost = branches.iter().map(ConjunctionPlan::cost).fold(Cost::EMPTY, Cost::combine_parallel);
         Ok(DisjunctionPlan { branches, _cost: cost })
+    }
+
+    pub(crate) fn required_inputs(&self) -> &[Variable] {
+        &self.required_inputs
     }
 }
 
--- a/compiler/executable/match_/planner/plan.rs
+++ b/compiler/executable/match_/planner/plan.rs
@@ -133,14 +133,7 @@ fn make_builder<'a>(
                         )
                     })
                     .collect::<Result<Vec<_>, _>>()?,
-                disjunction
-                    .required_inputs(block_context)
-                    .chain(
-                        disjunction
-                            .optional_outputs(block_context)
-                            .filter(|out| conjunction.variable_dependency_modes(block_context)[out].is_produced()),
-                    )
-                    .collect(),
+                disjunction.required_inputs(block_context).collect(),
             )),
             NestedPattern::Negation(negation) => negation_subplans.push(
                 make_builder(
@@ -153,7 +146,7 @@ fn make_builder<'a>(
                     statistics,
                     call_cost_provider,
                 )?
-                .with_inputs(negation.conjunction().captured_variables(block_context))
+                .with_inputs(negation.required_inputs(block_context))
                 .plan()?,
             ),
             NestedPattern::Optional(_) => unimplemented_feature!(Optionals),
--- a/compiler/executable/match_/planner/plan.rs
+++ b/compiler/executable/match_/planner/plan.rs
@@ -1486,6 +1486,7 @@ impl ConjunctionPlan<'_> {
     ) -> Result<(), QueryPlanningError> {
         match &self.graph.elements()[&VertexId::Pattern(pattern)] {
             PlannerVertex::Variable(_) => unreachable!("encountered variable @ pattern id {pattern:?}"),
+
             PlannerVertex::FunctionCall(call_planner) => {
                 // We push exactly the same as if it weren't a check.
                 let call_binding = call_planner.call_binding;
@@ -1515,6 +1516,7 @@ impl ConjunctionPlan<'_> {
                 });
                 match_builder.push_step(&HashMap::new(), step_builder.into());
             }
+
             PlannerVertex::Negation(negation) => {
                 let negation = negation.plan().lower(
                     match_builder.current_outputs.iter().copied(),
@@ -1532,12 +1534,14 @@ impl ConjunctionPlan<'_> {
                     StepInstructionsBuilder::Negation(NegationBuilder::new(negation)).into(),
                 )
             }
+
             PlannerVertex::Is(is) => {
                 let lhs = is.is().lhs().as_variable().unwrap();
                 let rhs = is.is().rhs().as_variable().unwrap();
                 let check = CheckInstruction::Is { lhs, rhs }.map(match_builder.position_mapping());
                 match_builder.push_check(&[lhs, rhs], check)
             }
+
             PlannerVertex::LinksDeduplication(deduplication) => {
                 let role1 = deduplication.links_deduplication().links1().role_type().as_variable().unwrap();
                 let player1 = deduplication.links_deduplication().links1().player().as_variable().unwrap();
@@ -1547,6 +1551,7 @@ impl ConjunctionPlan<'_> {
                     .map(match_builder.position_mapping());
                 match_builder.push_check(&[role1, player1, role2, player2], check)
             }
+
             PlannerVertex::Comparison(comparison) => {
                 let comparison = comparison.comparison();
                 let lhs = comparison.lhs();
@@ -1579,11 +1584,15 @@ impl ConjunctionPlan<'_> {
                 let vars = [lhs_var, rhs_var].into_iter().flatten().collect_vec();
                 match_builder.push_check(&vars, check)
             }
+
             PlannerVertex::Constraint(constraint) => self.lower_constraint_check(match_builder, constraint),
+
             PlannerVertex::Unsatisfiable(_) => match_builder.push_check(&Vec::new(), CheckInstruction::Unsatisfiable),
+
             PlannerVertex::Expression(_) => {
                 unreachable!("Would require multiple assignments to the same variable and be flagged")
             }
+
             PlannerVertex::Disjunction(disjunction) => {
                 let step_builder = disjunction
                     .builder()
--- a/compiler/executable/match_/planner/plan.rs
+++ b/compiler/executable/match_/planner/plan.rs
@@ -153,7 +153,7 @@ fn make_builder<'a>(
     }
 
     let mut plan_builder = ConjunctionPlanBuilder::new(
-        Vec::from_iter(conjunction.captured_required_variables(block_context)),
+        conjunction.captured_required_variables(block_context).collect(),
         type_annotations,
         statistics,
     );
--- a/compiler/executable/match_/planner/plan.rs
+++ b/compiler/executable/match_/planner/plan.rs
@@ -1555,7 +1555,7 @@ impl ConjunctionPlan<'_> {
 
                 let lhs_var = lhs.as_variable();
                 let rhs_var = rhs.as_variable();
-                let num_input_variables = [lhs_var, rhs_var].into_iter().filter(|x| x.is_some()).count();
+                let num_input_variables = [lhs_var, rhs_var].into_iter().flatten().dedup().count();
                 assert!(num_input_variables > 0);
 
                 let order = self.element_to_order[&VertexId::Pattern(pattern)];
--- a/compiler/executable/match_/planner/plan.rs
+++ b/compiler/executable/match_/planner/plan.rs
@@ -134,11 +134,11 @@ fn make_builder<'a>(
                     })
                     .collect::<Result<Vec<_>, _>>()?,
                 disjunction
-                    .required_inputs()
+                    .required_inputs(block_context)
                     .chain(
                         disjunction
-                            .optional_outputs()
-                            .filter(|out| conjunction.variable_dependency_modes()[out].is_produced()),
+                            .optional_outputs(block_context)
+                            .filter(|out| conjunction.variable_dependency_modes(block_context)[out].is_produced()),
                     )
                     .collect(),
             )),
--- a/compiler/executable/match_/planner/plan.rs
+++ b/compiler/executable/match_/planner/plan.rs
@@ -83,6 +83,7 @@ pub(crate) fn plan_conjunction<'a>(
     conjunction: &'a Conjunction,
     block_context: &BlockContext,
     variable_positions: &HashMap<Variable, VariablePosition>,
+    shared_variables: &HashSet<Variable>,
     type_annotations: &'a TypeAnnotations,
     variable_registry: &VariableRegistry,
     expressions: &'a HashMap<ExpressionBinding<Variable>, ExecutableExpression<Variable>>,
@@ -93,6 +94,7 @@ pub(crate) fn plan_conjunction<'a>(
         conjunction,
         block_context,
         variable_positions,
+        shared_variables,
         type_annotations,
         variable_registry,
         expressions,
@@ -106,6 +108,7 @@ fn make_builder<'a>(
     conjunction: &'a Conjunction,
     block_context: &BlockContext,
     variable_positions: &HashMap<Variable, VariablePosition>,
+    shared_variables: &HashSet<Variable>,
     type_annotations: &'a TypeAnnotations,
     variable_registry: &VariableRegistry,
     expressions: &'a HashMap<ExpressionBinding<Variable>, ExecutableExpression<Variable>>,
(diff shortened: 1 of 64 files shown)


krishnangovindraj (commented):
First skim. Just some questions to help me understand the code.
I still need to do the meat of it in ir.

executor/instruction/iterator.rs
```diff
@@ -149,6 +152,10 @@ impl {
 }
 
 impl TupleIterator {
+    pub(crate) fn empty() -> Self {
```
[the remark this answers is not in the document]
krishnangovindraj: I thought this happens only in an invalid DNF disjunct.

ir/pattern/disjunction.rs
```diff
@@ -40,6 +45,52 @@ impl Disjunction {
...
+    ) -> HashMap<Variable, DependencyMode<'_>> {
+        if self.conjunctions.is_empty() {
+            return HashMap::new();
+        }
+        let mut data_modes = self.conjunctions[0].variable_dependency_modes(block_context);
+        for branch in &self.conjunctions[1..] {
+            for (var, mode) in branch.variable_dependency_modes(block_context) {
+                data_modes.entry(var).or_insert(DependencyMode::Optional).or(mode)
+            }
+        }
```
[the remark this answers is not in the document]
dmitrii-ubskii: No. Good catch.

ir/pattern/constraint.rs
```diff
@@ -2045,6 +2135,10 @@ impl<ID: IrID> FunctionCallBinding<ID> {
         self.ids_assigned().chain(self.function_call.argument_ids())
     }
 
+    pub fn ids_accepted(&self) -> impl Iterator<Item = ID> + '_ {
```
flyingsilverfin, line 2138: why not just `argument_ids` ( you could rename `ids_assigned` to `assigne_ids` as well to make it match if you wanted, but not required imo)

ir/pattern/constraint.rs
```diff
@@ -1937,6 +2023,10 @@ impl<ID: IrID> ExpressionBinding<ID> {
         [&self.left].into_iter()
     }
 
+    pub fn ids_accepted(&self) -> impl Iterator<Item = ID> + '_ {
+        self.expression.variables()
+    }
```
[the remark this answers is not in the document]
flyingsilverfin: Indeed i also think it could be either `ids_required`? Or maybe `ids_input`?

ir/pattern/constraint.rs
```diff
@@ -578,16 +640,40 @@ impl<ID: IrID> Constraint<ID> {
...
             Constraint::Relates(relates) => Box::new(relates.ids()),
             Constraint::Plays(plays) => Box::new(plays.ids()),
             Constraint::Value(value) => Box::new(value.ids()),
-            Constraint::LinksDeduplication(dedup) => Box::new(dedup.ids()),
+            Constraint::LinksDeduplication(_) => Box::new(iter::empty()),
             Constraint::Unsatisfiable(inner) => Box::new(inner.ids()),
         }
     }
 
+    pub fn required_ids(&self) -> Box<dyn Iterator<Item = ID> + '_> {
```
flyingsilverfin, line 653: Oh you have `required_ids` here, so the below in Function and Expression could be `required_ids`?

ir/pipeline/block.rs
```diff
@@ -240,6 +259,30 @@ impl BlockContext {
...
         self.variable_declaration.iter().map(|(&var, &scope)| (var, scope))
     }
+
+    pub fn visible_variables(&self, root: ScopeId) -> impl Iterator<Item = Variable> + '_ {
+        self.get_variable_scopes().filter_map(move |(var, scope)| {
+            (scope == ScopeId::INPUT || scope == root || self.is_visible_child(scope, root)).then_some(var)
+        })
+    }
+
+    pub fn variable_status_in_scope(&self, var: Variable, scope: ScopeId) -> VariableStatus {
```
flyingsilverfin, line 269: Maybe we should call this `VariableLocality` instead of `status`?

```
fn variable_locality_in_scope(var, scope) {
  if declared in this scope: Local
  if declared in a child: Child
  else None
}
```

```
VariableLocality {
  Local,
  Child, 
  None
}
```

ir/pattern/mod.rs
```diff
@@ -348,3 +349,77 @@ impl fmt::Display for ValueType {
...
+        // we only really need to preserve one assignment across branches for diagnostic purposes
+        // or the branch with multiple assignments
+        if matches!((&self, &other), (Self::Assigned(_), Self::MultipleAssignments(_))) {
+            *self = other
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum DependencyMode<'a> {
```
flyingsilverfin, line 388: I'm not finding `AssignmentMode` and `DependencyMode` great at conveying what they are... they're both about _variables_ right?

`AssignmentMode` is about recording which constraints hold assignment for a variable? Why not just have `VariableAssignment` as the name?

ir/pattern/conjunction.rs
```diff
@@ -130,6 +135,44 @@ impl Conjunction {
...
+                match assignment_modes.entry(var) {
+                    hash_map::Entry::Occupied(mut entry) => entry.get_mut().and(mode),
+                    hash_map::Entry::Vacant(vacant_entry) => {
+                        vacant_entry.insert(mode);
+                    }
+                }
+            }
+        }
+        assignment_modes
+    }
```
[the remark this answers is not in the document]
flyingsilverfin: Interesting, what do you mean by that?

ir/pattern/conjunction.rs
```diff
@@ -130,6 +135,44 @@ impl Conjunction {
...
+    pub fn required_inputs<'a>(&'a self, block_context: &BlockContext) -> impl Iterator<Item = Variable> + 'a {
+        self.variable_dependency_modes(block_context)
+            .into_iter()
+            .filter_map(|(v, mode)| mode.is_required().then_some(v))
+    }
+
+    pub fn variable_dependency_modes(&self, block_context: &BlockContext) -> HashMap<Variable, DependencyMode<'_>> {
+        let mut data_modes = self.constraints.variable_dependency_modes();
+        for nested in self.nested_patterns.iter() {
+            let nested_pattern_data_modes = nested.variable_dependency_modes(block_context);
```
flyingsilverfin, line 148: `data_modes` or `dependency_modes`?

ir/pattern/mod.rs
flyingsilverfin: Same with `DependencyMode` should be `VariableDependency`?

My natural question is: do we need to have two different enums for this? Really, `assigned` is related to `produced` and `required` in two different ways... but is there a deterministic relationship at all? How/when can they co-occur?

Does it ever make sense to group on two different axes, such as `VariableStatus = Produced(Vec<Assigning constraint>) | Required(Vec<Requiring constraint>)` and then otherwise Optionality?

I didn't take too long to look into, but it's similar to an exercise I did with @krishnangovindraj which pushed him to structure the `captured_variables /declared_variables / referenced_variables / producible variables` in a a way that is hierarchically related. 

My main concern is introducing new terms which may overlap or conflict with things elsewhere.

executor/instruction/has_executor.rs
```diff
@@ -166,8 +162,15 @@ impl HasExecutor {
...
-            BinaryIterateMode::Unbound => {
+        let attribute = self.has.attribute().as_variable().unwrap().as_position();
+        let owner = self.has.owner().as_variable().unwrap().as_position();
+
+        let attribute =
+            attribute.and_then(|pos| (row.len() > pos.as_usize()).then(|| row.get(pos))).filter(|v| !v.is_empty());
+        let owner = owner.and_then(|pos| (row.len() > pos.as_usize()).then(|| row.get(pos))).filter(|v| !v.is_empty());
+
+        match (owner, attribute) {
+            (None, None) => {
```
krishnangovindraj, line 173: When the planner was planning this - A `has` instruction in a disjunction branch with it's owner `$owner` optionally bound - what would it have considered as the binding state of `$owner`?  Does it consider it unbound?
Else, Wouldn't it pick a fully unbound though there may have been a "slightly more expensive" but always bound plan starting from a different side of the pattern?

executor/instruction/mod.rs
```diff
@@ -503,404 +515,43 @@ impl<T> Checker<T> {
...
         let mut filters: Vec<Box<dyn Fn(&T) -> Result<bool, Box<ConceptReadError>>>> =
             Vec::with_capacity(self.checks.len());
 
         for check in &self.checks {
-            match check {
-                &CheckInstruction::Iid { var, iid } => {
-                    let maybe_var_extractor = self.extractors.get(&var);
-                    let var: BoxExtractor<T> = match maybe_var_extractor {
-                        Some(&subtype) => Box::new(subtype),
-                        None => make_const_extractor(&CheckVertex::Variable(var), row, context),
```
krishnangovindraj, line 516: General question:
I see row being passed to `make_const_extractor`. 
Does that actually need the row?
And in that case, does this function need the row? And does it have to be called in `*Executor::get_iterator` or can it be called in `*Executor::new`

executor/instruction/has_executor.rs
```diff
@@ -166,8 +162,15 @@ impl HasExecutor {
         let snapshot = &**context.snapshot();
```
krishnangovindraj, line 162: > ```
 let filter_for_row: Box<HasFilterMapFn> = Box::new(move |item| match filter(&item) {
            Ok(true) => match check(&item) {
                Ok(true) | Err(_) => Some(item),
                Ok(false) => None,
            },
            Ok(false) => None,
            Err(_) => Some(item),
        });
```
What does this do, just above this line?

File: executor/instruction/has_executor.rs (hunk 10 above)
Comment (krishnangovindraj) [reply executor/instruction/has_executor.rs:173]: Yes, it considers it to be unbound.

File: executor/instruction/has_executor.rs [hunk 13]
```diff
@@ -182,52 +185,10 @@ impl HasExecutor {
                     &self.variable_modes,
                 )))
             }
-            BinaryIterateMode::UnboundInverted => {
```
krishnangovindraj, line 185: Does unbound inverted go away? Or does it have to be re-introduced?

executor/instruction/mod.rs
```diff
@@ -344,15 +344,18 @@ fn type_from_row_or_annotations<'a>(
...
-) -> Type {
+) -> Option<Type> {
     match vertex {
         &Vertex::Variable(ExecutorVariable::RowPosition(var)) => {
             debug_assert!(row.len() > var.as_usize());
-            let VariableValue::Type(type_) = row.get(var).to_owned() else { unreachable!("Supertype must be a type") };
-            type_
+            match row.get(var).to_owned() {
+                VariableValue::Type(type_) => Some(type_),
+                VariableValue::Empty => None,
```
krishnangovindraj, line 353: I explicitly avoided making this return optional when I introduced the `Unsatisfiable` constraint.
Do we need to do it now? Should we not be generating answers if it's not set?

executor/instruction/iterator.rs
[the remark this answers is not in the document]
krishnangovindraj: @flyingsilverfin , I'd like to answer this. Having one might be useful but I do not understand the current usage.
I wonder whether it precedes the current view of optionals (i.e. from when unwrapping an optional would cause the search to "fail".

executor/instruction/mod.rs
krishnangovindraj: Answering myself again:
With dynamic modes it's likely we'll only ever call this with the value bound. So I'll just do an unwrap outside for now and see if all calls here are followed by an unwrap

executor/instruction/mod.rs
krishnangovindraj: Isn't it strange to call something that comes from the row a constant? Or is it like some value with a type attached?

executor/instruction/has_executor.rs
```diff
@@ -237,13 +198,26 @@ impl HasExecutor {
...
+                Ok(TupleIterator::Check(SortedTupleIterator::new(
+                    owner_obj
+                        .as_object()
+                        .has_attribute(snapshot, thing_manager, attr)?
+                        .then(|| Ok(Tuple::Pair([owner.to_owned(), attribute.to_owned()])))
+                        .into_iter(),
+                    self.tuple_positions.clone(),
+                    &self.variable_modes,
+                )))
+            }
```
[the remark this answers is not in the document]
krishnangovindraj: Oh I can answer this myself again:
The unbound mode only has the prefix and at most the from-type. So it needs to filter the to type.

ir/pattern/optional.rs
```diff
@@ -40,6 +41,22 @@ impl Optional {
     pub fn conjunction_mut(&mut self) -> &mut Conjunction {
         &mut self.conjunction
     }
+
+    pub(crate) fn variable_dependency_modes(
+        &self,
+        block_context: &BlockContext,
+    ) -> HashMap<Variable, DependencyMode<'_>> {
+        // DependencyMode::Produced means "(can be) produced in all branches"
```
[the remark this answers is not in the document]
krishnangovindraj: What does this mean? Why are we only returning required variables?

ir/pattern/mod.rs
```diff
@@ -348,3 +349,84 @@ impl fmt::Display for ValueType {
...
+            (Self::Referencing, _) | (_, Self::Referencing) => *self = Self::Referencing,
+            (Self::Producing, Self::Producing) => (),
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct VariableDependency<'a> {
+    mode: VariableDependencyMode,
+    places: Vec<&'a Constraint<Variable>>,
```
krishnangovindraj, line 383: Do we want to call places `references`?
Just to clarify: It's all places that a variable is referenced regardless of what the mode there was?

ir/pattern/conjunction.rs
```diff
@@ -106,29 +97,38 @@ impl Conjunction {
...
+            for (var, mode) in nested_pattern_data_modes {
+                match dependencies.entry(var) {
+                    hash_map::Entry::Occupied(mut entry) => *entry.get_mut() &= mode,
+                    hash_map::Entry::Vacant(vacant_entry) => {
+                        vacant_entry.insert(mode);
+                    }
+                }
+            }
+        }
+        dependencies
```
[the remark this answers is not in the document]
krishnangovindraj: Oh, we are validating it there

ir/pipeline/block.rs
```diff
@@ -185,17 +197,23 @@ impl BlockContext {
...
+            *self.variable_declaration.get_mut(&var).unwrap() = scope;
         } else {
-            Err(Box::new(RepresentationError::DisjointVariableReuse { name: var_name.to_string(), source_span }))
+            let ancestor = common_ancestor(&self.scope_parents, recorded_scope, scope);
+            if !self.is_visible_child(scope, ancestor) || !self.is_visible_child(recorded_scope, ancestor) {
+                return Err(Box::new(RepresentationError::DisjointVariableReuse {
+                    name: var_name.to_owned(),
+                    source_span,
+                }));
+            }
```
krishnangovindraj, line 213: Will this firing or not depend on which nested pattern is processed first? Do we just want to process negations at the end?
```
not { $x > 25; }; { $x isa age 5; } or {$x isa age 10; };
```

ir/pipeline/block.rs
```diff
@@ -185,17 +197,23 @@ impl BlockContext {
...
+            *self.variable_declaration.get_mut(&var).unwrap() = ancestor;
```
krishnangovindraj, line 214: Nice

krishnangovindraj (commented):
Looks good. Just a few questions for me to keep track of what's happening.

ir/pipeline/block.rs
dmitrii-ubskii: Good point. I wonder if this error check can be removed altogether, surely the variable dependency check handles this already.

ir/pattern/disjunction.rs
```diff
@@ -40,6 +45,45 @@ impl Disjunction {
...
+            for (var, dependency) in &mut dependencies {
+                if !branch_dependency_modes.contains_key(var) && dependency.is_producing() {
+                    dependency.set_referencing()
+                }
+            }
+            for (var, mut dependency) in branch_dependency_modes {
+                let entry = dependencies.entry(var);
+                match entry {
+                    hash_map::Entry::Occupied(mut entry) => {
+                        *entry.get_mut() |= dependency;
```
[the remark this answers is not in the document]
flyingsilverfin: Not a huge fan but it does look nice lol

ir/pattern/mod.rs
```diff
@@ -348,3 +349,88 @@ impl fmt::Display for ValueType {
         }
     }
 }
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+enum VariableDependencyMode {
+    Required,
+    Producing,
+    Referencing,
+}
```
flyingsilverfin, line 358: When i looked at this before i also kind of struggled to figure out how to name/interpret these. Presumably this is within some fixed context, correct? (Pretty sure it's a Pattern)
---> 
In some contexts, the variable is Required
In some contexts, the variable is Produced
In some contexts, the variable is Referenced

Intuitively knowing TypeQL, 'required' probably means it's a a variable used in an function arg, expression, or comparison (non-binding positions).

A Producing variable should be anything that is in a binding position in a context, so kind of the opposite of Required?

But then I wonder, how can any variable be 'Referencing'? What's the relationship between Referencing vs Producing and Referencing vs Required?

ir/pattern/mod.rs
flyingsilverfin: Yeah let's use `referencing_constraints`

ir/pattern/optional.rs
```diff
@@ -40,6 +41,24 @@ impl Optional {
...
+    pub(crate) fn variable_dependency(
+        &self,
+        block_context: &BlockContext,
+    ) -> HashMap<Variable, VariableDependency<'_>> {
+        self.conjunction
+            .variable_dependency(block_context)
+            .into_iter()
+            .map(|(var, mut mode)| {
+                // VariableDependency::Producing means "producing in all branches".
+                // A try {} block never produces.
```
flyingsilverfin, line 54: Is this a fact?

ir/pattern/optional.rs
```diff
@@ -40,6 +41,24 @@ impl Optional {
...
+
+    pub(crate) fn variable_dependency(
+        &self,
+        block_context: &BlockContext,
+    ) -> HashMap<Variable, VariableDependency<'_>> {
+        self.conjunction
+            .variable_dependency(block_context)
+            .into_iter()
+            .map(|(var, mut mode)| {
+                // VariableDependency::Producing means "producing in all branches".
```
flyingsilverfin, line 53: Is this even relevant anymore, we don't treat optionals as branches i thought

ir/pipeline/block.rs
```diff
@@ -132,6 +131,20 @@ fn validate_conjunction(
         let rhs_category = variable_registry.get_variable_category(is.rhs().as_variable().unwrap()).unwrap();
         lhs_category.narrowest(rhs_category).is_none()
     });
+
+    for (var, mode) in conjunction.variable_dependency(block_context) {
+        // TODO error on var that's in some but not all branches of a disj
```
flyingsilverfin, line 136: Implement?

ir/pattern/mod.rs
dmitrii-ubskii: They're all degrees of referencing. A referenced variable may have to be bound to a value by another pattern (`Required`), or may be able to be bound to a value by the pattern in question (`Producing`). `Referencing` is when it is neither, i.e. the pattern is referencing the variable but doesn't do anything special with it.

A disjunction `{ $x has $a; } or { $y has $a; }` references `$a` and can produce a value for it, so the dependency mode is `Producing`. It also references `$x`, but can't produce it and does not require it to be bound by another pattern, so the mode for `$x` is `Referencing`.

ir/pattern/optional.rs
dmitrii-ubskii: "In all code paths" if you prefer.

ir/pattern/optional.rs
dmitrii-ubskii: Yep. An optional variablet is not `Produced` for the purposes of resolving `Require`ments such as in an expression or a function argument.

ir/pattern/mod.rs
flyingsilverfin: I don't like these and don't think we should use them, but is it accurate?
```
VariableUsage {
  Input,
  InputOrOutput,
  InputOrHidden,
}
```

dmitrii-ubskii: Basically.

compiler/executable/match_/planner/plan.rs
```diff
@@ -1555,7 +1560,7 @@ impl ConjunctionPlan<'_> {
 
                 let lhs_var = lhs.as_variable();
                 let rhs_var = rhs.as_variable();
-                let num_input_variables = [lhs_var, rhs_var].into_iter().filter(|x| x.is_some()).count();
+                let num_input_variables = [lhs_var, rhs_var].into_iter().flatten().dedup().count();
```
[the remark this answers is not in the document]
flyingsilverfin: (ie flatten is ok imo)

ir/pattern/optional.rs
flyingsilverfin: But why do you have a comment on what VariableDependency means here of all places? I didn't see something similar in disjunctions or negations

dmitrii-ubskii: Because here we're unconditionally overriding all "producing" modes and I figured that the first question would be "why?".

flyingsilverfin: Right, maybe answer that directly then? `A try block never produces output, and can only have local or input variables`?

Is that we've agreed on for now? I think we will revisit this anyway

flyingsilverfin: Good idea to leave a comment tho

ir/pattern/mod.rs
flyingsilverfin: Ok i think we can keep your terms since i'm reaally struggling to find something better!

flyingsilverfin (approved):

ir/pipeline/block.rs
krishnangovindraj: Yeah, I wouldn't mind that. I think the scope being set to the common ancestor is still correct.

ir/pattern/disjunction.rs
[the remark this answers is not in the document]
krishnangovindraj: I didn't mind the `and` and `or` either. 
Whichever one we use, I like the usage of 'and' and 'or' as the name of the operations.

PR #7377, opened 2025-02-26
Allow multiple expression assignments in different branches

krishnangovindraj:
Allows multiple expression assignments for the same variable in different branches - essential for meaningful recursive functions.
* Assignments in different branches of the same disjunction are allowed
* Assignments in different constraints/nested-patterns of the same conjunction are not allowed.
* Assignments across branches must have the same value type.

Example isage:
```
  with
  fun factorial($i: integer) -> { integer }:
  match
      { $i <= 1; let $factorial = 1; } or
      { $i > 1; let $factorial = $i * factorial($i - 1); }; 
  return { $factorial };

  match
      let $f_5 in factorial(5);
```
## Motivation
Allow multiple expression assignments in different branches - essential for meaningful recursive functions.
requires: typedb/typedb-behaviour#343

## Implementation
* Rework the multiple assignments checks to allow multiple assignments in different branches of the same conjunction.
* Refactor compiled_expressions datastructures to index by constraint instead of assigned variable, since these are no longer 1:1.

### DIFF
--- a/compiler/annotation/expression/block_compiler.rs
+++ b/compiler/annotation/expression/block_compiler.rs
@@ -151,11 +151,14 @@ fn index_expressions_disjunction<'block, Snapshot: ReadableSnapshot>(
         .into_iter()
         .flat_map(|branch_index| branch_index.into_iter())
         .for_each(|(var, expressions)| combined_indices.entry(var).or_default().extend(expressions));
-    combined_indices.into_iter().try_for_each(|(var, expressions)| match index.insert(var, expressions) {
-        Some(_) => Err(ExpressionCompileError::MultipleAssignmentsForVariable {
-            variable: context.variable_name(&var),
-            source_span: None,
-        }),
+    combined_indices.into_iter().try_for_each(|(var, expressions)| match index.insert(var.clone(), expressions) {
+        Some(_) => {
+            debug_assert!(index.get(&var).unwrap().len() > 0);
+            Err(ExpressionCompileError::MultipleAssignmentsForVariable {
+                variable: context.variable_name(&var),
+                source_span: index.get(&var).unwrap().first().unwrap().source_span(),
+            })
+        }
         None => Ok(()),
     })?;
     Ok(())
@@ -166,18 +169,18 @@ fn resolve_type_for_variable<'a, Snapshot: ReadableSnapshot>(
     context: &mut BlockExpressionsCompilationContext<'a, Snapshot>,
     variable: Variable,
     expression_assignments: &HashMap<Variable, Vec<&'a ExpressionBinding<Variable>>>,
-    source_span: Option<Span>,
+    assignment_span: Option<Span>,
 ) -> Result<ExpressionValueType, Box<ExpressionCompileError>> {
     if let Some(value) = context.variable_value_types.get(&variable) {
         Ok(value.clone())
     } else if let Some(value) = try_value_type_from_assignments(context, variable, expression_assignments)? {
         Ok(value)
-    } else if let Some(value) = try_value_type_from_type_annotations(context, variable)? {
+    } else if let Some(value) = try_value_type_from_type_annotations(context, variable, assignment_span)? {
         Ok(value)
     } else {
         Err(Box::new(ExpressionCompileError::CouldNotDetermineValueTypeForVariable {
             variable: context.variable_name(&variable),
-            source_span, // TODO: this can be improved?
+            source_span: assignment_span,
         }))
     }
 }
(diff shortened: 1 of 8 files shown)


farost:
I'd add some TypeQL examples in the RN

compiler/annotation/expression/block_compiler.rs
```diff
@@ -66,173 +78,192 @@ pub fn compile_expressions<'block, Snapshot: ReadableSnapshot>(
...
+    };
+    let variable_category = context.variable_registry.get_variable_category(variable).unwrap();
+    let is_list = match variable_category {
+        VariableCategory::Value | VariableCategory::Attribute | VariableCategory::Thing => false,
+        VariableCategory::ValueList | VariableCategory::AttributeList | VariableCategory::ThingList => true,
+        _ => {
+            return Err(Box::new(ExpressionCompileError::VariableMustBeValueOrAttribute {
+                variable: context.variable_name(&variable),
+                category: variable_category,
+                source_span: None, // TODO: this can be improved?
```
farost, line 241: Let's try to think about this todo in this PR

compiler/annotation/expression/block_compiler.rs
```diff
@@ -66,173 +78,192 @@ pub fn compile_expressions<'block, Snapshot: ReadableSnapshot>(
...
-                    .variable_registry
-                    .variable_names()
-                    .get(&variable)
-                    .cloned()
-                    .unwrap_or_else(|| VariableRegistry::UNNAMED_VARIABLE_DISPLAY_NAME.to_string()),
-                value_types: value_types.iter().join(", "),
-                source_span: None, // TODO: this can be improved?
-            }))
+        let mut return_types = HashSet::new();
+        for assignment in expression_assignments.get(&variable).unwrap() {
```
farost, line 198: This is the kind of place I'd be happier to see an `expect` in instead of the `unwrap`

farost (approved):
Lgtm, left comments

compiler/annotation/expression/block_compiler.rs
krishnangovindraj: Slightly misleading if there are multiple assignments, but atleast it's a start now.

compiler/annotation/expression/block_compiler.rs
krishnangovindraj: managed to replace the `if contains_key` with a `if let Some() = get`

compiler/annotation/expression/block_compiler.rs
```diff
@@ -66,173 +78,195 @@ pub fn compile_expressions<'block, Snapshot: ReadableSnapshot>(
...
-        ExpressionCompilationContext::compile(expression, &context.variable_value_types, context.parameters)?;
-    context.variable_value_types.insert(assigned_variable, compiled.return_type.clone());
-    context.compiled_expressions.insert(assigned_variable, compiled);
+    let mut combined_indices = HashMap::<Variable, Vec<&'block ExpressionBinding<Variable>>>::new();
+    let mut branch_indices = disjunction
+        .conjunctions()
+        .iter()
+        .map(|branch| {
+            let mut branch_index = HashMap::new();
+            index_expressions_conjunction(context, branch, &mut branch_index).map(|_| branch_index)
```
flyingsilverfin, line 147: Can you add 1 test that makes sure this works across two levels of nesting?
```
match
{ let $x = ... } 
or
{ 
  $y isa person;
  { let $x = ... }
  or 
  { let $x = ... };
};
```
Seems like it should work but is the kind of thing that would easily go wrong

flyingsilverfin (changes requested):

compiler/annotation/expression/block_compiler.rs
```diff
@@ -65,174 +77,198 @@ pub fn compile_expressions<'block, Snapshot: ReadableSnapshot>(
...
-        compile_expressions_recursive(&mut context, variable, &expression_index)?
+    let mut expression_index = index_expressions_conjunction(&context, block.conjunction())?;
+    if let Some(var) = expression_index.keys().find(|var| input_value_type_annotations.contains_key(var)) {
+        return Err(Box::new(ExpressionCompileError::ReassigningValueVariable {
+            variable: context.variable_name(var),
+        }));
+    }
+    for (var, _) in &expression_index {
+        if !context.visited_expressions.contains(var) {
+            let _value_type = try_value_type_from_expression_assignments(&mut context, *var, &expression_index)?;
```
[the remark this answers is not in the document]
krishnangovindraj: If it does, we'll have a terrible crash and a wonderful rust optimiser bug.
I imagine `debug_assert`s are applied around the pre-processor step like in C++, so it's gone well before inline optimisations. 

Tangentially, I wonder whether there is a difference between the two versions.
`let x = f(); x.g();` and `f().g()` in the IR.

compiler/executable/match_/planner/plan.rs
```diff
@@ -507,11 +507,11 @@ impl<'a> ConjunctionPlanBuilder<'a> {
     fn register_expression_binding(
         &mut self,
         expression: &ExpressionBinding<Variable>,
-        expressions: &'a HashMap<Variable, ExecutableExpression<Variable>>,
+        expressions: &'a HashMap<ExpressionBinding<Variable>, ExecutableExpression<Variable>>,
     ) {
         let variable = expression.left().as_variable().unwrap();
         let output = self.graph.variable_index[&variable];
-        let expression = &expressions[&variable];
+        let expression = &expressions[expression];
```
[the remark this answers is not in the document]
krishnangovindraj: This is called executable_expressions throughout the planner. I can change it everywhere if needed but just renaming  arg to `binding` also makes sense.

flyingsilverfin (approved):
I hope this works with @dmitrii-ubskii 's changes!

flyingsilverfin:
Oh bdd is red!

PR #7378, opened 2025-02-26
Allow abstract relation types having zero role types.

farost:
Allow abstract relation types having zero role types. It is no longer a requirement for a relation type to declare or inherit role types if it's abstract. 

Non-abstract relation types are still required to have at least one role type. 
Notice that it can be a single abstract role type. However, such a schema does not allow preserving any data related to this relation type as its instances will get cleaned up on commits due to the absence of role players.

Additionally, a bug in relation type deletion has been fixed. In some cases, deletion of a relation type could affect specialized role types of its supertype.

## Motivation
This change makes schemas less restrictive and more flexible.

### Validations
Commit time validation:
* ✓ If a relation type is not abstract, but does not have any role types (abstract or not abstract, whatever), we error

Operation time validation (if not implemented, explained why):
* ✘ New relation type created: it can have 0 role types until we commit it
* ✘ Relation supertype changes to a type without role types: it could just be created, and it would be strange to error. It is the responsibility of the supertype to have at least one role (which will be inherited)
* ✘ Role type gets deleted: I initially wanted to check that it's not the last role type of a type. However, it blocks the migration of form `undefine relates rol from rel; rel;` <- it would not let to delete the last role `rol`, which is logically executed first. This query feels reasonable for "total cleanup of types". Maybe, if we want to keep it in sync with the lower checks, we should check it.
* ✓ Abstractness is unset for a relation type without role types: at least something. The error would be a little strange in case of this series of queries: `define relation rel; ... define rel @abstract; ... undefine @abstract from rel`, but we already have similar cases in other constraints checks, if I remember correctly, so it's better to at least limit this action. The only issue here is `undefine @abstract from rel; rel;` <- same as for role types, but the intent of this query feels different, so I'd error.
* ✓ Supertype is unset (`undefine sub rel2 from rel`), and there are no more role types declared or inherited, while the subtype is not abstract. Same as before: it feels reasonable, but `undefine sub rel2 from rel; rel;` <- will get a little unreasonable error. 

### Relation type deletion and its effect on role types
While I thought that I misused specialized relates only in undefine queries, it appeared that we had a bug in TypeManager. When deleting a relation type, it cleaned up its role types, but it used to get these role types based on declared relates... Which include both explicit and implicit relates. Now it's fixed, and an additional Concept API interface for Concept BDD (and other possible use) has been introduced, similarly to name-based search methods for relates.

### DIFF
--- a/concept/type_/relation_type.rs
+++ b/concept/type_/relation_type.rs
@@ -19,6 +19,7 @@ use encoding::{
     value::label::Label,
     Prefixed,
 };
+use itertools::Itertools;
 use lending_iterator::higher_order::Hkt;
 use primitive::maybe_owns::MaybeOwns;
 use resource::constants::snapshot::BUFFER_KEY_INLINE;
@@ -293,12 +294,54 @@ impl RelationType {
         type_manager.get_relation_type_relates_declared(snapshot, *self)
     }
 
+    pub fn get_relates_explicit_declared(
+        &self,
+        snapshot: &impl ReadableSnapshot,
+        type_manager: &TypeManager,
+    ) -> Result<HashSet<Relates>, Box<ConceptReadError>> {
+        self.get_relates_declared(snapshot, type_manager)?
+            .iter()
+            .filter_map(|relates| {
+                let is_explicit = match relates.is_implicit(snapshot, type_manager) {
+                    Ok(is_implicit) => !is_implicit,
+                    Err(err) => return Some(Err(err)),
+                };
+                if is_explicit {
+                    Some(Ok(*relates))
+                } else {
+                    None
+                }
+            })
+            .try_collect()
+    }
+
     pub fn get_relates<'m>(
         &self,
         snapshot: &impl ReadableSnapshot,
         type_manager: &'m TypeManager,
     ) -> Result<MaybeOwns<'m, HashSet<Relates>>, Box<ConceptReadError>> {
         type_manager.get_relation_type_relates(snapshot, *self)
+    }
+
+    pub fn get_relates_explicit(
+        &self,
+        snapshot: &impl ReadableSnapshot,
+        type_manager: &TypeManager,
+    ) -> Result<HashSet<Relates>, Box<ConceptReadError>> {
+        self.get_relates(snapshot, type_manager)?
+            .iter()
+            .filter_map(|relates| {
+                let is_explicit = match relates.is_implicit(snapshot, type_manager) {
+                    Ok(is_implicit) => !is_implicit,
+                    Err(err) => return Some(Err(err)),
+                };
+                if is_explicit {
+                    Some(Ok(*relates))
+                } else {
+                    None
+                }
+            })
+            .try_collect()
     }
 
     pub fn get_relates_with_specialised<'m>(
@@ -314,7 +357,7 @@ impl RelationType {
         snapshot: &impl ReadableSnapshot,
         type_manager: &TypeManager,
     ) -> Result<HashSet<RoleType>, Box<ConceptReadError>> {
-        Ok(self.get_relates_declared(snapshot, type_manager)?.iter().map(|relates| relates.role()).collect())
+        Ok(self.get_relates_explicit_declared(snapshot, type_manager)?.iter().map(|relates| relates.role()).collect())
     }
 
     pub fn get_related_role_types(
--- a/concept/type_/relation_type.rs
+++ b/concept/type_/relation_type.rs
@@ -302,14 +302,10 @@ impl RelationType {
         self.get_relates_declared(snapshot, type_manager)?
             .iter()
             .filter_map(|relates| {
-                let is_explicit = match relates.is_implicit(snapshot, type_manager) {
-                    Ok(is_implicit) => !is_implicit,
+                match relates.is_implicit(snapshot, type_manager) {
+                    Ok(false) => Some(Ok(*relates)),
+                    Ok(true) => None,
                     Err(err) => return Some(Err(err)),
-                };
-                if is_explicit {
-                    Some(Ok(*relates))
-                } else {
-                    None
                 }
             })
             .try_collect()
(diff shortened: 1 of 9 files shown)


concept/type_/type_manager.rs
```diff
@@ -1298,11 +1298,6 @@ impl TypeManager {
         let exists = relation_type.get_relates(snapshot, self)?.contains(&relates);
 
         if !exists && !is_implicit {
-            OperationTimeValidation::validate_cardinality_of_inheritance_line_with_updated_capabilities(
```
farost, line 1301: We are not going to implement these complicated checks any time soon, we decided that it's a user's problem if their cardinalities are not synced. Too complicated.
Not worth leaving these unwanted non-working code in the codebase: it's better to reimplement it from scratch once it's really needed.

farost:
There is a test I changed in the undefine query BDDs:
```
      undefine
      as parent from fathership relates father;
      @card from fathership relates father;
      relates father from fathership;
      # @abstract from fathership; # fathership does not have roles at this point. we may want to allow it or error...
      sub parentship from fathership;
      fathership;
```
I've written this test myself, so I don't mind leaving it with the abstract undefinition commented out. But it feels fragile. At the same time, we want to check the maximum available options at operation time. Please share your thoughts.

concept/type_/relation_type.rs
```diff
@@ -293,6 +294,27 @@ impl RelationType {
         type_manager.get_relation_type_relates_declared(snapshot, *self)
     }
 
+    pub fn get_relates_explicit_declared(
```
farost, line 297: Similar to `get_role_type_explicit_...`. I didn't think I needed it, but we actually do while working in TypeManager.

concept/type_/type_manager.rs
```diff
@@ -1388,10 +1383,10 @@ impl TypeManager {
     ) -> Result<(), Box<ConceptWriteError>> {
         self.validate_delete_type(snapshot, thing_manager, relation_type)?;
 
-        let declared_relates = TypeReader::get_capabilities_declared::<Relates>(snapshot, relation_type)?;
-        for relates in declared_relates.iter() {
-            self.delete_role_type(snapshot, thing_manager, relates.role())?;
+        for explicit_relates in relation_type.get_relates_explicit_declared(snapshot, self)?.iter() {
```
farost, line 1386: This is crucial. That's the moment I understood I actually need the advanced logic from Concept API, so I substituted all capability-related TypeReader calls with Concept API calls here, to be more precise. 
Here, we used to delete superroles. Yes.... Sometimes it returned an error, but sometimes (when we are really unlucky) we could manage to actually delete it. Depends on the actual query.

concept/type_/type_manager.rs
```diff
@@ -1437,21 +1427,25 @@ impl TypeManager {
...
+        for plays in relates.role().get_plays(snapshot, self)?.iter() {
+            self.unset_plays_unchecked(snapshot, *plays)?;
         }
 
         self.unset_relates_unchecked(snapshot, relates)?;
         TypeWriter::storage_delete_type_vertex_property::<Ordering>(snapshot, relates.role());
-        self.delete_type(snapshot, relates.role())
+        self.delete_type(snapshot, relates.role())?;
+
+        if let Some(old_supertype) = old_supertype {
```
farost, line 1440: When the role type is deleted, don't forget to delete its supertype's implicit relates of its relation type. It should always exist if the supertype exists.
I did it when specialization changed, but totally forgot about type deletions.

concept/type_/type_manager.rs
```diff
@@ -2472,22 +2444,29 @@ impl TypeManager {
...
     }
 
-    fn unset_specialising_relates_when_unset_specialise_if_no_subtypes(
+    fn unset_specialising_relates_if_no_sub_role_types_from_relation_type(
         &self,
         snapshot: &mut impl WritableSnapshot,
         relation_type: RelationType,
         specialised_role_type: RoleType,
     ) -> Result<(), Box<ConceptWriteError>> {
-        if specialised_role_type.get_subtypes(snapshot, self)?.len() == 0 {
```
farost, line 2481: It was incorrect. We should check the count against the specific relation type, not any relation type. Covered with tests.

concept/type_/type_manager.rs
```diff
@@ -3391,26 +3364,6 @@ impl TypeManager {
...
-    fn validate_updated_capability_cardinality_against_schema(
-        &self,
-        snapshot: &mut impl WritableSnapshot,
-        capability: impl Capability,
-        cardinality: AnnotationCardinality,
-    ) -> Result<(), Box<ConceptWriteError>> {
-        OperationTimeValidation::validate_cardinality_arguments(cardinality)
-            .map_err(|typedb_source| ConceptWriteError::SchemaValidation { typedb_source })?;
-
-        OperationTimeValidation::validate_cardinality_of_inheritance_line_with_updated_cardinality(
```
farost, line 3403: Deleted

concept/type_/type_manager.rs
```diff
@@ -3391,26 +3364,6 @@ impl TypeManager {
         Ok(())
     }
 
-    fn validate_updated_capability_cardinality_against_schema(
-        &self,
-        snapshot: &mut impl WritableSnapshot,
-        capability: impl Capability,
-        cardinality: AnnotationCardinality,
-    ) -> Result<(), Box<ConceptWriteError>> {
-        OperationTimeValidation::validate_cardinality_arguments(cardinality)
```
farost, line 3400: Call it explicitly instead of `validate_updated_capability_cardinality_against_schema`

dependencies/typedb/repositories.bzl
```diff
@@ -33,8 +33,19 @@ def typedb_protocol():
     )
 
 def typedb_behaviour():
+    # TODO: Return typedb
+#    native.local_repository(
+#        name = "typedb_behaviour",
+#        path = "../typedb-behaviour",
+#    )
     git_repository(
         name = "typedb_behaviour",
```
farost, line 42: https://github.com/typedb/typedb-behaviour/pull/344

concept/type_/type_manager.rs
farost: Nope, it's exclusive to relates because of specialization.

concept/type_/relation_type.rs
```diff
@@ -293,6 +294,27 @@ impl RelationType {
...
+            .filter_map(|relates| {
+                let is_explicit = match relates.is_implicit(snapshot, type_manager) {
+                    Ok(is_implicit) => !is_implicit,
+                    Err(err) => return Some(Err(err)),
+                };
+                if is_explicit {
+                    Some(Ok(*relates))
+                } else {
+                    None
+                }
```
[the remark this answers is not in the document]
farost: A good notice! Thanks

PR #7381, opened 2025-03-03
Implement put stage

krishnangovindraj:
Implements the `put` stage. 
A `put <pattern>;`will perform an insert if no answers matched `<pattern>`; else, it returns all answers matched by the pattern.
**Note:** The put matches the entire pattern or inserts the entire pattern. Partial matching and partial inserts are not performed.

**Examples**
```
insert $p isa person, has name "Alice"; # Create some initial data.
```

```
put $p1 isa person, has name "Alice"; # Does nothing
```
```
put $p2 isa person, has name "Bob"; # Creates a person with name Bob
```

```
put $p3 isa person, has name "Alice", has age 10; # Creates a new person with name "Alice" & age 10.
```
The above statement (for $p3) creates a new person because the pattern fails to match, as the existing person with name "Alice" does not have age 10.
To achieve the desired result of not re-creating the person if only the age attribute is missing, `put` stages can be pipelined.
```
put $p4 isa person, has name "Bob"; # Will match the person with name "bob", if it exists.
put $p4 isa person, has age 11; # Adds an attribute to the same person, since we use $p4 again.
```

## Motivation
Feature 'Put'
requires: typedb/typedb-behaviour#346

## Implementation
* Introduces a `PutExecutable` containing a `MatchExecutable` and `InsertExecutable`
* Updates `InsertExecutable` to be able to optionally accept the desired output row format - this allows the match & put to share the same output row format.
* Implements `PutStageExecutor`, whose `into_iterator` method holds the `if not match { insert; } else { match; }` logic.
  - The rows from the previous stage are streamed in, and collected in a batch. A `must_insert` vector is maintained alongside, indicating whether a collected row must perform an insert.
  - For each row, we run the MatchExecutor. If any rows are returned, these are appended to the collector batch and marked as not needing insert.
  - If no rows are returned, we append a new empty row with spaces for the inserted concept.
  - Once all rows are collected, all inserts are performed and then the collected rows are streamed out.

### DIFF
--- a/compiler/executable/insert/executable.rs
+++ b/compiler/executable/insert/executable.rs
@@ -60,8 +60,10 @@ pub fn compile(
     desired_output_variable_positions: Option<HashMap<Variable, VariablePosition>>,
     source_span: Option<Span>,
 ) -> Result<InsertExecutable, Box<WriteCompilationError>> {
-    let mut variable_positions = input_variables.clone();
-    compile_error!("Figure out how to accept desire output positions");
+    debug_assert!(desired_output_variable_positions.clone().map(|positions| {
+            input_variables.iter().all(|(k,v)| positions.get(k).unwrap() == v)
+        }).unwrap_or(true));
+    let mut variable_positions = desired_output_variable_positions.unwrap_or_else(|| input_variables.clone());
     let concept_inserts = add_inserted_concepts(
         constraints,
         type_annotations,
@@ -91,7 +93,7 @@ pub(crate) fn add_inserted_concepts(
     output_variables: &mut HashMap<Variable, VariablePosition>,
     stage_source_span: Option<Span>,
 ) -> Result<Vec<ConceptInstruction>, Box<WriteCompilationError>> {
-    let first_inserted_variable_position =
+    let mut next_insert_position =
         output_variables.values().map(|pos| pos.position + 1).max().unwrap_or(0) as usize;
     let type_bindings = collect_type_bindings(constraints, type_annotations)?;
     let value_bindings = collect_value_bindings(constraints)?;
@@ -165,10 +167,12 @@ pub(crate) fn add_inserted_concepts(
                     }));
                 } // else let the original be
             } else {
-                let write_to =
-                    VariablePosition::new((first_inserted_variable_position + concept_instructions.len()) as u32);
-                output_variables.insert(thing, write_to);
-                let instruction = ConceptInstruction::PutObject(PutObject { type_, write_to: ThingPosition(write_to) });
+                if !output_variables.contains_key(&thing) {
+                    output_variables.insert(thing, VariablePosition::new(next_insert_position as u32));
+                    next_insert_position += 1;
+                };
+                let write_to = ThingPosition(*output_variables.get(&thing).unwrap());
+                let instruction = ConceptInstruction::PutObject(PutObject { type_, write_to });
                 concept_instructions.insert(thing, instruction);
             }
         } else {
@@ -216,11 +220,13 @@ pub(crate) fn add_inserted_concepts(
                     }));
                 } // else let the original be
             } else {
-                let write_to =
-                    VariablePosition::new((first_inserted_variable_position + concept_instructions.len()) as u32);
-                output_variables.insert(thing, write_to);
+                if !output_variables.contains_key(&thing) {
+                    output_variables.insert(thing, VariablePosition::new(next_insert_position as u32));
+                    next_insert_position += 1;
+                };
+                let write_to = ThingPosition(*output_variables.get(&thing).unwrap());
                 let instruction =
-                    ConceptInstruction::PutAttribute(PutAttribute { type_, value, write_to: ThingPosition(write_to) });
+                    ConceptInstruction::PutAttribute(PutAttribute { type_, value, write_to });
                 concept_instructions.insert(thing, instruction);
             }
         };
@@ -382,11 +388,11 @@ pub(crate) fn get_kinds_from_types(types: &BTreeSet<Type>) -> HashSet<Kind> {
 }
 
 pub(crate) fn prepare_output_row_schema(
-    input_variables: &HashMap<Variable, VariablePosition>,
+    variable_positions: &HashMap<Variable, VariablePosition>,
 ) -> Vec<Option<(Variable, VariableSource)>> {
-    let output_width = input_variables.values().map(|i| i.position + 1).max().unwrap_or(0);
+    let output_width = variable_positions.values().map(|i| i.position + 1).max().unwrap_or(0);
     let mut output_row_schema = vec![None; output_width as usize];
-    input_variables.iter().map(|(v, i)| (i, v)).for_each(|(&i, &v)| {
+    variable_positions.iter().map(|(v, i)| (i, v)).for_each(|(&i, &v)| {
         output_row_schema[i.position as usize] = Some((v, VariableSource::InputVariable(i)));
     });
     output_row_schema
--- a/compiler/executable/insert/executable.rs
+++ b/compiler/executable/insert/executable.rs
@@ -57,9 +57,11 @@ pub fn compile(
     input_variables: &HashMap<Variable, VariablePosition>,
     type_annotations: &TypeAnnotations,
     variable_registry: &VariableRegistry,
+    desired_output_variable_positions: Option<HashMap<Variable, VariablePosition>>,
     source_span: Option<Span>,
 ) -> Result<InsertExecutable, Box<WriteCompilationError>> {
     let mut variable_positions = input_variables.clone();
+    compile_error!("Figure out how to accept desire output positions");
     let concept_inserts = add_inserted_concepts(
         constraints,
         type_annotations,
--- a/compiler/executable/insert/executable.rs
+++ b/compiler/executable/insert/executable.rs
@@ -13,6 +13,7 @@ use ir::{
         constraint,
         constraint::{Comparator, Constraint},
         expression::Expression,
+        variable_category::VariableCategory,
         ParameterID, Vertex,
     },
     pipeline::VariableRegistry,
@@ -65,6 +66,7 @@ pub fn compile(
         .map(|positions| { input_variables.iter().all(|(k, v)| positions.get(k).unwrap() == v) })
         .unwrap_or(true));
     let mut variable_positions = desired_output_variable_positions.unwrap_or_else(|| input_variables.clone());
+
     let concept_inserts = add_inserted_concepts(
         constraints,
         type_annotations,
@@ -74,15 +76,27 @@ pub fn compile(
         source_span,
     )?;
 
+    #[cfg(debug_assertions)]
+    let mut variable_positions_tmp = variable_positions
+        .iter()
+        .filter(|(var, _)| {
+            !(variable_registry.get_variable_name(**var).is_none()
+                && variable_registry.get_variable_category(**var).unwrap() == VariableCategory::RoleType)
+        })
+        .map(|(k, v)| (k.clone(), v.clone()))
+        .collect::<HashMap<_, _>>();
+    #[cfg(not(debug_assertions))]
+    compile_error!("Remove the above");
+
     let mut connection_inserts = Vec::with_capacity(constraints.len());
     add_has(constraints, &variable_positions, variable_registry, &mut connection_inserts)?;
-    add_links(constraints, type_annotations, &variable_positions, variable_registry, &mut connection_inserts)?;
+    add_links(constraints, type_annotations, &variable_positions_tmp, variable_registry, &mut connection_inserts)?;
 
     Ok(InsertExecutable {
         executable_id: next_executable_id(),
         concept_instructions: concept_inserts,
         connection_instructions: connection_inserts,
-        output_row_schema: prepare_output_row_schema(&variable_positions),
+        output_row_schema: prepare_output_row_schema(input_variables, &variable_positions),
     })
 }
 
@@ -237,19 +251,19 @@ pub(crate) fn add_inserted_concepts(
 
 fn add_has(
     constraints: &[Constraint<Variable>],
-    input_variables: &HashMap<Variable, VariablePosition>,
+    variable_positions: &HashMap<Variable, VariablePosition>,
     variable_registry: &VariableRegistry,
     instructions: &mut Vec<ConnectionInstruction>,
 ) -> Result<(), Box<WriteCompilationError>> {
     filter_variants!(Constraint::Has: constraints).try_for_each(|has| {
-        let owner = get_thing_input_position(
-            input_variables,
+        let owner = get_thing_position(
+            variable_positions,
             has.owner().as_variable().unwrap(),
             variable_registry,
             has.source_span(),
         )?;
-        let attribute = get_thing_input_position(
-            input_variables,
+        let attribute = get_thing_position(
+            variable_positions,
             has.attribute().as_variable().unwrap(),
             variable_registry,
             has.source_span(),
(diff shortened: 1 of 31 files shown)


compiler/executable/insert/executable.rs
```diff
@@ -57,9 +58,15 @@ pub fn compile(
...
     type_annotations: &TypeAnnotations,
     variable_registry: &VariableRegistry,
+    desired_output_variable_positions: Option<HashMap<Variable, VariablePosition>>,
     source_span: Option<Span>,
 ) -> Result<InsertExecutable, Box<WriteCompilationError>> {
-    let mut variable_positions = input_variables.clone();
+    debug_assert!(desired_output_variable_positions
+        .clone()
+        .map(|positions| { input_variables.iter().all(|(k, v)| positions.get(k).unwrap() == v) })
+        .unwrap_or(true));
```
krishnangovindraj, line 67: This assumption is made by pure `insert` stages. Though we pass `None` to them now.
Violating this assumption in a `put` stage should be fine.

compiler/executable/insert/executable.rs
```diff
@@ -69,15 +76,27 @@ pub fn compile(
...
+        .iter()
+        .filter(|(var, _)| {
+            !(variable_registry.get_variable_name(**var).is_none()
+                && variable_registry.get_variable_category(**var).unwrap() == VariableCategory::RoleType)
+        })
+        .map(|(k, v)| (k.clone(), v.clone()))
+        .collect::<HashMap<_, _>>();
+    #[cfg(not(debug_assertions))]
+    compile_error!("Remove the above");
+
```
krishnangovindraj, line 90: TODO: Rebase on Dmitrii's changes to see if anonymous role-types are excluded from the row returned by a match.

compiler/executable/insert/executable.rs
```diff
@@ -89,8 +108,7 @@ pub(crate) fn add_inserted_concepts(
     output_variables: &mut HashMap<Variable, VariablePosition>,
     stage_source_span: Option<Span>,
 ) -> Result<Vec<ConceptInstruction>, Box<WriteCompilationError>> {
-    let first_inserted_variable_position =
-        output_variables.values().map(|pos| pos.position + 1).max().unwrap_or(0) as usize;
+    let mut next_insert_position = output_variables.values().map(|pos| pos.position + 1).max().unwrap_or(0) as usize;
```
krishnangovindraj, line 111: rework to accept desired output positions for inserted variables

compiler/executable/insert/executable.rs
```diff
@@ -380,13 +402,18 @@ pub(crate) fn get_kinds_from_types(types: &BTreeSet<Type>) -> HashSet<Kind> {
...
+    let output_width = output_positions.values().map(|i| i.position + 1).max().unwrap_or(0);
     let mut output_row_schema = vec![None; output_width as usize];
-    input_variables.iter().map(|(v, i)| (i, v)).for_each(|(&i, &v)| {
-        output_row_schema[i.position as usize] = Some((v, VariableSource::InputVariable(i)));
+    output_positions.iter().for_each(|(var, pos)| {
+        let source =
+            input_positions.get(var).map(|pos| VariableSource::Input(*pos)).unwrap_or(VariableSource::Inserted);
+        output_row_schema[pos.position as usize] = Some((*var, source));
     });
+    // This assumption is made in prepare_output_rows, though it's likely to be violated only for put which doesn't make that assumption.
```
krishnangovindraj, line 415: Actually, why make that assumption. I'll just make the insert stage also copy-mapped

dependencies/typedb/repositories.bzl
```diff
@@ -35,6 +35,6 @@ def typedb_protocol():
 def typedb_behaviour():
     git_repository(
         name = "typedb_behaviour",
-        remote = "https://github.com/typedb/typedb-behaviour",
-        commit = "17488d012682574b8969e503a2f4ea5618491413",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        remote = "https://github.com/krishnangovindraj/typedb-behaviour",
+        commit = "6b8d1205b4b53434e1a557fdc9b1d6fdaa570b1e",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
```
krishnangovindraj, line 39: needs merge

executor/pipeline/put.rs
```diff
@@ -0,0 +1,160 @@
...
+                &context.profile,
+            )
+            .map_err(|err| Box::new(PipelineExecutionError::InitialisingMatchIterator { typedb_source: err }));
+            let mut match_iterator = match executor {
+                Ok(executor) => crate::pipeline::match_::unique_rows(crate::pipeline::match_::as_owned_rows(
+                    executor.into_iterator(context.clone(), interrupt.clone()),
+                ))
+                .peekable(),
+                Err(err) => return Err((err, context)),
+            };
```
krishnangovindraj, line 102: Just as we do in the match iterator

executor/pipeline/put.rs
```diff
@@ -0,0 +1,160 @@
...
+                match row_result {
+                    Ok(row) => {
+                        output_batch.append(row);
+                        must_insert.push(false);
+                    }
+                    Err(typedb_source) => {
+                        return Err((Box::new(PipelineExecutionError::ReadPatternExecution { typedb_source }), context))
+                    }
+                }
+            }
```
krishnangovindraj, line 117: Insert any rows produced by matching

executor/pipeline/put.rs
```diff
@@ -0,0 +1,160 @@
...
+                for _ in 0..multiplicity {
+                    output_batch.append(MaybeOwnedRow::new_borrowed(&empty_output_row, &1)); // Insert an empty row
+                    output_batch.get_row_mut(output_batch.len() - 1).copy_mapped(
+                        // Copy over input_row
+                        input_row.as_reference(),
+                        input_output_mapping.iter().map(|(s, d)| (s.clone(), d.clone())),
+                    );
+                    must_insert.push(true);
+                }
+            }
```
krishnangovindraj, line 130: Else insert rows without fresh-variables and mark for insert.

executor/pipeline/put.rs
```diff
@@ -0,0 +1,160 @@
...
+                ) {
+                    return Err((Box::new(PipelineExecutionError::WriteError { typedb_source }), context));
+                }
+            }
+            if index % 100 == 0 {
+                if let Some(interrupt) = interrupt.check() {
+                    return Err((Box::new(PipelineExecutionError::Interrupted { interrupt }), context));
+                }
+            }
+        }
```
krishnangovindraj, line 156: Iterate over all rows and insert anything that's marked.

executor/pipeline/put.rs
```diff
@@ -0,0 +1,160 @@
...
+                snapshot,
+                thing_manager,
+                input_row.clone(),
+                function_registry.clone(),
+                &context.profile,
+            )
+            .map_err(|err| Box::new(PipelineExecutionError::InitialisingMatchIterator { typedb_source: err }));
+            let mut match_iterator = match executor {
+                Ok(executor) => crate::pipeline::match_::unique_rows(crate::pipeline::match_::as_owned_rows(
+                    executor.into_iterator(context.clone(), interrupt.clone()),
```
krishnangovindraj, line 98: This should take care of interrupts during the match

krishnangovindraj (commented):
self review

compiler/executable/insert/mod.rs
```diff
@@ -8,6 +8,7 @@ use std::error::Error;
 
 use ir::pattern::ParameterID;
 use itertools::Itertools;
+use typeql::match_;
```
krishnangovindraj, line 11: unused import probably

compiler/executable/pipeline.rs
```diff
@@ -262,6 +266,29 @@ fn compile_stage(
...
+                statistics,
+                call_cost_provider,
+            )
+            .map_err(|source| ExecutableCompilationError::PutMatchCompilation { typedb_source: source })?;
+            let insert_plan = crate::executable::insert::executable::compile(
+                block.conjunction().constraints(),
+                input_variables,
+                insert_annotations,
+                variable_registry,
+                Some(match_plan.variable_positions().clone()),
```
krishnangovindraj, line 286: Pass match output mapping so they both produce the same rows.

executor/pipeline/insert.rs
```diff
@@ -61,6 +59,13 @@ where
 
         let profile = context.profile.profile_stage(|| String::from("Insert"), executable.executable_id);
 
+        // prepare_output_rows copies unmapped
+        debug_assert!(executable.output_row_schema.iter().enumerate().all(|(i, source_opt)| {
+            match source_opt {
+                Some((_, VariableSource::Input(position))) => position.as_usize() == i,
+                None | Some((_, VariableSource::Inserted)) => true,
+            }
+        }));
```
krishnangovindraj, line 68: Should I just bite and make prepare_rows accept a non-trivial mapping?

farost:
Regarding the release notes:

1. I'd rewrite it a little (removed `performs will perform` and updated the last sentence).

> Implements the put stage.
A put <pattern>; will perform an insert if no answers matched <pattern>; else, it returns all answers matched by the pattern.
Note: The put matches the entire pattern or inserts the entire pattern. Partial matching and partial inserts are not performed.

2. The examples are a little confusing. The first example reads like a set of separate queries (they are not connected logically, they have separate comments). However, the second example introduces a pipeline, and its intent is not clear. I'd a least added comments to the lines: the first line inserts a new owner, and the second line updates this owner. 

3. I'd extend the examples list to show more power of puts. For example, putting an attribute for an existing instance (match + put). At least mentioning it explicitly without writing the code. Maybe it's something to cover on the docs stage (can be stolen from here for simplicity).

compiler/executable/mod.rs
```diff
@@ -37,6 +38,8 @@ typedb_error! {
         DeleteExecutableCompilation(3, "Error compiling delete stage into executable.", typedb_source: Box<WriteCompilationError>),
         FetchCompilation(4, "Error compiling fetch stage into executable.", typedb_source: FetchCompilationError),
         MatchCompilation(5, "Error compiling match stage into executable.", typedb_source: MatchCompilationError),
+        PutMatchCompilation(6, "Error compiling the match version of put stage into executable.", typedb_source: MatchCompilationError),
```
farost, line 41: A little strange to use the article `the` for the version and continue skipping articles for other nouns. Not sure what's the standard, but the mix feels off.

compiler/executable/pipeline.rs
```diff
@@ -262,6 +266,29 @@ fn compile_stage(
...
+                selected_variables,
+                match_annotations,
+                variable_registry,
+                &HashMap::new(),
+                statistics,
+                call_cost_provider,
+            )
+            .map_err(|source| ExecutableCompilationError::PutMatchCompilation { typedb_source: source })?;
+            let insert_plan = crate::executable::insert::executable::compile(
+                block.conjunction().constraints(),
```
farost, line 282: Can you give me a dumb explanation why we pass this instead of the `block` here?

executor/pipeline/put.rs
```diff
@@ -0,0 +1,160 @@
...
+    fn into_iterator(
+        self,
+        mut interrupt: ExecutionInterrupt,
+    ) -> Result<
+        (Self::OutputIterator, ExecutionContext<Snapshot>),
+        (Box<PipelineExecutionError>, ExecutionContext<Snapshot>),
+    > {
+        let Self { previous: previous_stage, executable, function_registry } = self;
+
+        let mut output_batch = Batch::new(executable.output_width() as u32, 10); // We don't know how many ahead of time
```
farost, line 64: `10` looks like a constant var

farost: And we could at least check the previous_stage if it exists. Like `previous_stage.map(|stage| stage.output_len()).unwrap_or(MIN_LEN_CONST)` or `max(previous_stage.map(|stage| stage.output_len()).unwrap_or(MIN_LEN_CONST), MIN_LEN_CONST)` (pseudocode)

farost (approved):
Approved with comments and questions

compiler/executable/pipeline.rs
krishnangovindraj: If i passed a block, there's questions on whether there are nested patterns in the block that have to be compiled, etc. I just wanted to keep it tight. As we add features in the future, we might get an `InsertBlock`.

executor/pipeline/put.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+impl<Snapshot, PreviousStage> StageAPI<Snapshot> for PutStageExecutor<PreviousStage>
+where
+    Snapshot: WritableSnapshot + 'static,
+    PreviousStage: StageAPI<Snapshot>,
+{
+    type OutputIterator = WrittenRowsIterator;
+
+    fn into_iterator(
```
flyingsilverfin, line 55: I feel like we don't normally have _so_ much logic in the `into_iterator` of a stage executor? This feels like an anti-pattern/missing abstraction/not matching existing patterns?

krishnangovindraj: I think it has the same responsibilities as the other `into_iterator`s. 
I also don't know if we're going to keep the `PipelineStage`s or just push everything into the `PatternExecutor`, since we've already got most of the read stages in there anyway.

executor/batch.rs
```diff
@@ -166,6 +166,7 @@ pub struct Batch {
 }
 
 impl Batch {
+    pub(crate) const DEFAULT_LENGTH: usize = 10;
```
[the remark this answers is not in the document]
farost: I agree that it's own crate is a better place for such a constant.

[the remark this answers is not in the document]
flyingsilverfin: We have such 'defaults' for crate-specific constants all centralised as well right now :) For example, 'QUERY_PLAN_CACHE_SIZE'. The reason i want to do that is to have one central registry that could in the future be used for 'hyperparameter' tuning for TypeDB, and perhaps even have them become individually configurably by the user.

executor/pipeline/put.rs
flyingsilverfin: Hmm ok maybe you can modularise this into some sub-functions so it's not so monolithic and huge? eg the way `InsertStageExecutor` does it?

executor/batch.rs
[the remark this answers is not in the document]
krishnangovindraj: Yes. My point is this is not a useful hyperparameter to tune and just adds noise to that file.

executor/batch.rs
```diff
@@ -203,6 +204,16 @@ impl Batch {
...
+    pub(crate) fn append_mapped(
+        &mut self,
+        row: MaybeOwnedRow<'_>,
+        mapping: impl Iterator<Item = (VariablePosition, VariablePosition)>,
+    ) {
+        let mut destination_row = self.row_internal_mut(self.entries);
+        destination_row.copy_mapped(row, mapping);
+        self.entries += 1;
+    }
+
```
flyingsilverfin, line 216: Don't you need to append `1` to the multiplicities too?

flyingsilverfin: In both Append methods? Also, when you create variations in the APIs you'd have one canonical one and redirect the other one to the canonical one (append() should call append_mapped with a identity mapping)

krishnangovindraj: > Don't you need to append 1 to the multiplicities too?
I haven't applied the [append changes](https://github.com/krishnangovindraj/typedb/pull/7) to this branch.

> append() should call append_mapped with a identity mapping
That makes a lot of sense and works quite nicely, but I think `copy_from_row` might have scope for compiler optimisation, whereas `copy_mapped` definitely won't. I can still do it, but there's the general question of when it's worth doing these things

executor/batch.rs
```diff
@@ -180,7 +180,7 @@ impl Batch {
     }
 
     pub(crate) fn get_multiplicities(&self) -> &[u64] {
-        self.multiplicities.as_ref()
+        &self.multiplicities[0..self.entries]
```
krishnangovindraj, line 183: Comment to reference in new PR

PR #7388, opened 2025-03-18
Fix concurrent schema and write transactions locking issues

farost:
Fix concurrent transactions locking issues, allowing multiple users to open schema and write transactions in parallel with guaranteed safety and availability whenever it's possible.
* When a transaction opening failure originated from the schema lock being held by an existing schema transaction, a deadlock situation appeared, preventing new transactions from being opened after the schema lock release. This does not happen anymore. 
* Schema transaction opening requests could be ignored and rejected on timeout because of preceding write transactions in the request queue, even if they were closed in time. The retry mechanism has been corrected to prevent any request from being ignored. 

## Implementation
See comments.

### DIFF
--- a/concept/Cargo.toml
+++ b/concept/Cargo.toml
@@ -128,7 +128,7 @@ features = {}
 
 	[dependencies.typeql]
 		features = []
-		rev = "5ccc1da728b381286b66cf059df010586eb60b48"
+		rev = "d1ef6785b27681bc2296c00e61db793d41f7f018"
 		git = "https://github.com/typedb/typeql"
 		default-features = false
 
--- a/concept/type_/relation_type.rs
+++ b/concept/type_/relation_type.rs
@@ -301,12 +301,10 @@ impl RelationType {
     ) -> Result<HashSet<Relates>, Box<ConceptReadError>> {
         self.get_relates_declared(snapshot, type_manager)?
             .iter()
-            .filter_map(|relates| {
-                match relates.is_implicit(snapshot, type_manager) {
-                    Ok(false) => Some(Ok(*relates)),
-                    Ok(true) => None,
-                    Err(err) => return Some(Err(err)),
-                }
+            .filter_map(|relates| match relates.is_implicit(snapshot, type_manager) {
+                Ok(false) => Some(Ok(*relates)),
+                Ok(true) => None,
+                Err(err) => return Some(Err(err)),
             })
             .try_collect()
     }
@@ -326,12 +324,10 @@ impl RelationType {
     ) -> Result<HashSet<Relates>, Box<ConceptReadError>> {
         self.get_relates(snapshot, type_manager)?
             .iter()
-            .filter_map(|relates| {
-                match relates.is_implicit(snapshot, type_manager) {
-                    Ok(false) => Some(Ok(*relates)),
-                    Ok(true) => None,
-                    Err(err) => return Some(Err(err)),
-                }
+            .filter_map(|relates| match relates.is_implicit(snapshot, type_manager) {
+                Ok(false) => Some(Ok(*relates)),
+                Ok(true) => None,
+                Err(err) => return Some(Err(err)),
             })
             .try_collect()
     }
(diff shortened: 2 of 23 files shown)


concept/Cargo.toml
```diff
@@ -128,7 +128,7 @@ features = {}
 
 	[dependencies.typeql]
 		features = []
-		rev = "5ccc1da728b381286b66cf059df010586eb60b48"
+		rev = "d1ef6785b27681bc2296c00e61db793d41f7f018"
```
farost, line 131: We don't sync >:(

concept/type_/relation_type.rs
```diff
@@ -301,12 +301,10 @@ impl RelationType {
     ) -> Result<HashSet<Relates>, Box<ConceptReadError>> {
         self.get_relates_declared(snapshot, type_manager)?
             .iter()
-            .filter_map(|relates| {
-                match relates.is_implicit(snapshot, type_manager) {
-                    Ok(false) => Some(Ok(*relates)),
-                    Ok(true) => None,
-                    Err(err) => return Some(Err(err)),
-                }
+            .filter_map(|relates| match relates.is_implicit(snapshot, type_manager) {
```
farost, line 304: Rustfmt (idk why it had not been triggered before)

database/database.rs
```diff
@@ -183,19 +183,47 @@ impl<D> Database<D> {
...
-            return;
-        }
-        let head = notify_queue.pop_front().unwrap();
-        if let TransactionReservationRequest::Schema(notifier) = head {
-            // fulfill exactly 1 schema request
-            *has_schema_transaction = true;
-            let _skipped_sync_error = notifier.send(()).ok();
-        } else {
-            // fulfill as many write requests as possible
-            while let Some(TransactionReservationRequest::Write(notifier)) = notify_queue.pop_front() {
```
farost, line 196: This could consume a Schema transaction and ignore it if there were Write transactions run before.

database/database.rs
```diff
@@ -183,19 +183,47 @@ impl<D> Database<D> {
...
+                else {
+                    panic!("Expected the next schema request: the queue cannot be changed")
+                };
+                match notifier.send(()).ok() {
+                    Some(_) => {
+                        // fulfill exactly 1 awaiting schema request
+                        *has_schema_transaction = true;
+                        break;
+                    }
+                    None => {}
```
farost, line 210: If `send()` is unsuccessful, it means that the transaction already exited on timeout and no longer waits for the notification. We should continue processing the queue instead of hanging.

database/tests/transaction.rs
```diff
@@ -0,0 +1,845 @@
...
+                let _tx_read = open_read(database_clone);
+            });
+
+            tokio::try_join!(task1, task2)
+        })
+        .unwrap();
+}
+
+////////////////////////////////////
+// COMPLICATED TRANSACTION QUEUES //
```
farost, line 628: I'd like to test complicated timeouts, but it's complicated to code them consistently. So whatever

database/tests/transaction.rs
```diff
@@ -0,0 +1,845 @@
...
+                let _tx_read = open_read(database_clone);
+            });
+
+            tokio::try_join!(task1, task2)
+        })
+        .unwrap();
+}
+
+#[test]
+fn schema_transaction_can_be_opened_after_prior_timeout_error() {
```
farost, line 208: This test checks the main fix of the PR.

database/tests/transaction.rs
```diff
@@ -0,0 +1,845 @@
...
+                let _tx_schema = open_schema(database);
+                notify_transaction1_ready.notify_one();
+            });
+
+            let task2 = tokio::spawn(async move {
+                notify_transaction1_ready_clone.notified().await;
+                let _tx_read = open_read(database_clone);
+            });
+
+            tokio::try_join!(task1, task2)
```
farost, line 202: `unwrap()` below will fail the test if any of the tasks panics, and it can happen inside `open_schema` or `open_write` that `expect` a `Result::Ok`

database/database.rs
flyingsilverfin: ahhh ok!

flyingsilverfin (approved):

PR #7391, opened 2025-03-20
Remove server domain-leak in the main function

lolski:
- The main and server module is refactored for better maintainability

## Motivation
- The `main` function contain functionalities that is supposed to be in the `server` module

## Implementation
- Minimise domain leak by moving functionalities into the `Server` struct appropriately. The `main` function is made to only contain code that is not server-specific such as panic handler, logging, and Sentry error reporting. Diagnostics initialisation and printing TypeDB logo is now moved into Server
- Tidy up naming and method placement in Server, such that methods that belongs to a particular domain sits next to one another


server/server.rs
```diff
@@ -262,56 +284,24 @@ impl Server {
...
-
-async fn resolve_address(address: String) -> SocketAddr {
-    lookup_host(address.clone())
-        .await
-        .unwrap()
-        .next()
-        .unwrap_or_else(|| panic!("Unable to map address '{}' to any IP addresses", address))
-}
 
-typedb_error! {
```
flyingsilverfin, line 301: works for me!

main.rs
```diff
@@ -7,68 +7,36 @@
...
-};
+use resource::constants::server::{DISTRIBUTION, SENTRY_REPORTING_URI, VERSION};
+use server::parameters::config::Config;
+use server::parameters::cli::CLIArgs;
+use server::server::Server;
+use tokio::runtime::Runtime;
 
 fn main() {
-    setup_abort_on_panic();
+    let config = CLIArgs::parse().to_config();
```
flyingsilverfin, line 19: I thought you were going to move all this to Server?

lolski: the config object required by the `may_initialise_error_reporting` on the `main` function unfortunately!

flyingsilverfin: you could move that into server too though?

flyingsilverfin (approved):

main.rs
flyingsilverfin: Works for me this way too tho, approved

PR #7393, opened 2025-03-21
Introduce REST API HTTP endpoint

farost:
Introduce REST API HTTP endpoint.


server/server.rs
```diff
@@ -215,62 +255,128 @@ impl Server {
...
+    ) -> Result<(), ServerOpenError> {
+        let authenticator =
+            http::authenticator::Authenticator::new(credential_verifier, token_manager, diagnostics_manager);
+
+        let encryption_config = http::encryption::prepare_tls_config(encryption_config)?;
+        let http_service = Arc::new(service);
+        let router_service = http::typedb_service::TypeDBService::create_protected_router(http_service.clone())
+            .layer(authenticator)
+            .merge(http::typedb_service::TypeDBService::create_unprotected_router(http_service))
+            .into_make_service();
```
krishnangovindraj, line 361: Can we add this please. It lets my browser make requests from the web-app
```suggestion
        let router_service = http::typedb_service::TypeDBService::create_protected_router(http_service.clone())
            .layer(authenticator)
            .merge(http::typedb_service::TypeDBService::create_unprotected_router(http_service))
            .layer(tower_http::cors::CorsLayer::permissive())
            .into_make_service();
```

krishnangovindraj (commented):
request cors_layer so I don't forget

server/server.rs
alexjpwalker: I think we can actually do slightly better here - the official builds of Studio will only be deployed to `studio.typedb.com` and `studio.typedb.dev`, and of course locally run Studio typically ends up on `localhost`.

So if our CORS policy allows precisely these three domains, we get both a functioning Studio and the security benefits of CORS.

resource/constants.rs
```diff
@@ -40,14 +40,22 @@ pub mod server {
...
+    pub const HTTP_PASSWORD_FIELD: &str = "password";
+    pub const HTTP_AUTHORIZATION_FIELD: &str = "authorization";
+    pub const HTTP_BEARER_PREFIX: &str = "Bearer ";
+    pub const DEFAULT_AUTHENTICATION_TOKEN_TTL_SECONDS: u64 = 4 * SECONDS_IN_HOUR;
+    pub const DEFAULT_AUTHENTICATION_TOKEN_TTL: Duration =
+        Duration::from_secs(DEFAULT_AUTHENTICATION_TOKEN_TTL_SECONDS);
+
+    pub const CONTENT_TYPE_FIELD: &str = "content-type";
+    pub const CONTENT_TYPE_GRPC: &str = "application/grpc";
+    pub const CONTENT_TYPE_JSON: &str = "application/json";
```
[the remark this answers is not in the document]
farost: I agree, I was going to work on the constants placement. I'm not yet sure what's the best approach in terms of placing constants, it already looks quite bloated

server/server.rs
krishnangovindraj: Was thinking about this. Should we make it configurable? How likely is it that the user wants to connect DIRECTLY to the database from their website like we do in studio?

alexjpwalker: Hmmm, that's a fair point - ultimately the server belongs to the user, so we might be overstepping our boundaries by imposing fixed restrictions on it

dependencies/typedb/repositories.bzl
```diff
@@ -12,29 +12,52 @@ def typedb_bazel_distribution():
...
+        remote = "https://github.com/farost/typedb-protocol",
+        commit = "7df6a6bfbbbd29940558ae7940b2065e1e2ba0b1",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_protocol
     )
+#    git_repository(
+#        name = "typedb_protocol",
+#        remote = "https://github.com/typedb/typedb-protocol",
+#        tag = "3.0.0",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_protocol
+#    )
 
 def typedb_behaviour():
```
farost, line 52: https://github.com/typedb/typedb-behaviour/pull/360

dependencies/typedb/repositories.bzl
```diff
@@ -12,29 +12,52 @@ def typedb_bazel_distribution():
     )
 
 def typedb_dependencies():
```
farost, line 14: https://github.com/typedb/typedb-dependencies/pull/569

dependencies/typedb/repositories.bzl
```diff
@@ -12,29 +12,52 @@ def typedb_bazel_distribution():
...
+        remote = "https://github.com/farost/typedb-dependencies",
+        commit = "a86993d232a9f050d9fa63b93a1164f4ea31d549",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_dependencies
     )
+#    git_repository(
+#        name = "typedb_dependencies",
+#        remote = "https://github.com/typedb/typedb-dependencies",
+#        commit = "3364a72f9ba384865e3b6ac950012d427726c8f0",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_dependencies
+#    )
 
 def typeql():
```
farost, line 27: https://github.com/typedb/typeql/pull/397

dependencies/typedb/repositories.bzl
farost, line 1: Dependent: https://github.com/typedb/typedb-driver/pull/749

server/service/http/message/authentication.rs
```diff
@@ -0,0 +1,44 @@
...
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct SigninPayload {
+    pub username: String,
+    pub password: String,
+}
+
+#[async_trait]
+impl<S> FromRequestParts<S> for Accessor
```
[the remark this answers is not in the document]
farost: This specific code allows me to require `accessor: Accessor` as an argument to my REST API method, and `axum` will automatically parse it. It's very cool!!! I'll show more below.

server/service/http/message/authentication.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
+        Accessor::from_extensions(&parts.extensions)
+            .map_err(|typedb_source| HTTPServiceError::Authentication { typedb_source }.into_response())
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TokenResponse {
```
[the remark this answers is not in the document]
farost: I added `Deserialize` for tests, so I use the same structs in my BDDs. Potentially, it can be somehow exposed as a library for other Rust users...

server/service/http/message/query/concept.rs
```diff
@@ -0,0 +1,317 @@
...
+    },
+};
+use encoding::value::{value::Value, value_type::ValueType, ValueEncodable};
+use error::unimplemented_feature;
+use executor::write::WriteError::ConceptRead;
+use serde::{ser::SerializeStruct, Serialize, Serializer};
+use serde_json::json;
+use storage::snapshot::ReadableSnapshot;
+
+// TODO: Should probably be merged with JSON from behaviour/steps/query_answer_context.rs.
```
farost, line 26: Too lazy now... But I haven't finished documents checks, so TBD

server/service/http/message/query/concept.rs
```diff
@@ -0,0 +1,317 @@
...
+// Now, it's easier to have symmetry between two services, and we don't have the capacity to merge
+// these (BDDs will check if this code is correct)
+
+macro_rules! count_fields {
+    () => { 0 };
+    ($head:ident $(, $tail:ident)*) => { 1 + count_fields!($($tail),*) };
+}
+
+macro_rules! serializable_response {
```
farost, line 35: Declaration + custom serialization to include kinds without storing them

server/service/http/message/transaction.rs
```diff
@@ -0,0 +1,115 @@
...
+where
+    S: Send + Sync,
+{
+    type Rejection = Response;
+
+    async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Self::Rejection> {
+        let params: Path<HashMap<String, String>> = parts.extract().await.map_err(IntoResponse::into_response)?;
+        let id = params
+            .get("transaction_id")
+            .ok_or_else(|| (StatusCode::NOT_FOUND, "transaction_id param missing").into_response())?;
```
farost, line 53: I should probably clean up these error mesages
Will add a TODO

tests/behaviour/service/http/BUILD
farost, line 1: I'm not sure about my design of the `bevahiour` package now. What I have:
* I have common `steps` I want to use as a library in both common BDDs and http-related BDDs, probably other service-related BDDs (e.g. testing TypeDB with gRPC to finally catch its bugs without typedb-driver!!!)
* I want to introduce my local `steps` for `http`
* Separating `service` bdds from other bbds is weird (including the common steps), so I just created a `service` subpackage and added an `http` subpackage inside of it.
Check out the new packaging and share your thoughts if needed.

tests/behaviour/service/http/debug/debug.feature
```diff
@@ -0,0 +1,105 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+Feature: Debugging Space
+
+  # Paste any scenarios below for debugging.
+  # Do not commit any changes to this file.
+
+  # TODO: Remove this
```
farost, line 10: Will do before merging

tests/behaviour/service/http/http_steps/message.rs
```diff
@@ -0,0 +1,365 @@
...
+        QueryAnswerResponse,
+    },
+    transaction::TransactionResponse,
+    user::{UserResponse, UsersResponse},
+};
+use url::form_urlencoded;
+
+use crate::{Context, HttpBehaviourTestError, HttpContext};
+
+async fn send_request(
```
farost, line 29: @flyingsilverfin This is how you use the HTTP endpoint as a Rust client

server/service/http/typedb_service.rs
```diff
@@ -0,0 +1,602 @@
...
+        //
+        // cors
+    }
+
+    async fn health() -> impl IntoResponse {
+        StatusCode::NO_CONTENT
+    }
+
+    async fn signin(
+        _version: ProtocolVersion,
```
farost, line 267: I did not do this in the current version, but we can just add `signin_v1` / `signin_v2` impls and have
`async fn singin(version: ProtocolVersion ....) -> ... { match version { ...... } }` everywhere.

server/service/http/typedb_service.rs
```diff
@@ -0,0 +1,602 @@
...
+        // }
+        //
+        // cors
+    }
+
+    async fn health() -> impl IntoResponse {
+        StatusCode::NO_CONTENT
+    }
+
+    async fn signin(
```
farost, line 266: By implementing different traits for `JsonBody` and other `***Response`/`***Payload`/`***Path`, I manage to request specific fields from `axum`, and it "automatically" understands where to search for this information: whether it's path, header, or body. 
`State` here is the actual service instead of `self`

server/service/http/transaction_service.rs
```diff
@@ -0,0 +1,1091 @@
...
+            QueryAnswer::ResDocuments((_, _, comment)) => match comment {
+                None => StatusCode::OK,
+                Some(comment) => comment.status_code(),
+            },
+        }
+    }
+}
+
+#[derive(Debug)]
+pub(crate) enum QueryAnswerComment {
```
farost, line 225: A very direct solution for "I can't error, I want to return some kind of success, but I want to warn the user"

server/service/http/transaction_service.rs
```diff
@@ -0,0 +1,1091 @@
...
+
+    // TODO: any method using `Result<ControlFlow<(), ()>, Status>` should really be `ControlFlow<Result<(), Status>, ()>`
+    async fn handle_next(&mut self, next: Option<(TransactionRequest, TransactionResponder)>) -> ControlFlow<(), ()> {
+        match next {
+            None => Break(()),
+            Some((request, response_sender)) => {
+                match request {
+                    TransactionRequest::Query(query_options, query) => {
+                        self.handle_query(query_options, query, response_sender).await
+                        // run_with_diagnostics_async(
```
farost, line 383: Not done yet

server/service/http/transaction_service.rs
```diff
@@ -0,0 +1,1091 @@
...
+            }
+        }
+    }
+
+    async fn handle_commit(&mut self, responder: TransactionResponder) -> ControlFlow<(), ()> {
+        // finish any running write query, interrupt running queries, clear all running/queued reads, finish all writes
+        //   note: if any write query errors, the whole transaction errors
+        // finish any active write query
+        if let Break(()) = self.finish_running_write_query_no_transmit(InterruptType::TransactionCommitted).await {
+            return Break(()); // TODO: We should really return state errors just to indicate there is an error to return from function, but identifying that the repsonde is already sent
```
farost, line 438: I don't remember what I wanted to say here, but I was not happy. Will return to it later

main.rs
```diff
@@ -22,7 +22,7 @@ fn main() {
     initialise_logging_global();
     may_initialise_error_reporting(&config);
     create_tokio_runtime().block_on(async {
-        let server = Server::create(config, ASCII_LOGO, DISTRIBUTION, VERSION, None).await.unwrap();
+        let server = Server::new(config, ASCII_LOGO, DISTRIBUTION, VERSION, None).await.unwrap();
```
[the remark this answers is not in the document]
flyingsilverfin: ok for me

resource/constants.rs
```diff
@@ -24,10 +24,15 @@ pub mod server {
...
-    pub const DEFAULT_TRANSACTION_TIMEOUT_MILLIS: u64 = Duration::from_secs(5 * SECONDS_IN_MINUTE).as_millis() as u64;
+    // TODO: Maybe we start moving these options to separate crates?
     pub const DEFAULT_PREFETCH_SIZE: u64 = 32;
     pub const DEFAULT_SCHEMA_LOCK_ACQUIRE_TIMEOUT_MILLIS: u64 = Duration::from_secs(10).as_millis() as u64;
+    pub const DEFAULT_TRANSACTION_TIMEOUT_MILLIS: u64 = Duration::from_secs(5 * SECONDS_IN_MINUTE).as_millis() as u64;
     pub const DEFAULT_TRANSACTION_PARALLEL: bool = true;
+    pub const DEFAULT_INCLUDE_INSTANCE_TYPES: bool = true;
+    pub const DEFAULT_INCLUDE_INSTANCE_TYPES_FETCH: bool = false;
+    pub const DEFAULT_ANSWER_COUNT_LIMIT_GRPC: Option<usize> = None;
+    pub const DEFAULT_ANSWER_COUNT_LIMIT_HTTP: Option<usize> = Some(1_000_000); // TODO: Smaller?
```
flyingsilverfin, line 35: definiiteeellly smaller than that! Maybe 10k?

server/authentication/token_manager.rs
```diff
@@ -0,0 +1,127 @@
...
+                let token_owners = token_owners_clone.clone();
+                let secret_key = secret_key_clone.clone();
+                async move {
+                    Self::cleanup_expired_tokens(secret_key.as_ref(), token_owners).await;
+                }
+            },
+            tokens_cleanup_interval,
+            false,
+        ));
+        // TODO: limit tokens_expiration_time by (min, max)? What values? 1 second and 100 years?..
```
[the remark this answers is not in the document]
flyingsilverfin: Let's use 1 year I think as the absolute max length of tokens, and say 1 second for the minimum, sure!

server/authentication/token_manager.rs
```diff
@@ -0,0 +1,127 @@
...
+    fn random_key() -> String {
+        rand::thread_rng().sample_iter(&rand::distributions::Alphanumeric).take(128).map(char::from).collect()
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
+pub(crate) struct Claims {
+    sub: String,
+    exp: u64,
+    iat: u64,
```
flyingsilverfin, line 126: These names are not very useful to the reader IMO!!

server/parameters/cli.rs
```diff
@@ -4,21 +4,36 @@
...
 pub struct CLIArgs {
-    /// Server host and port, eg., 0.0.0.0:1729
-    #[arg(long = "server.address")]
-    pub server_address: Option<String>,
+    /// Server host and port (e.g., 0.0.0.0:1729)
+    #[arg(long = "server.address", default_value_t = DEFAULT_ADDRESS.to_string())]
+    pub server_address: String,
+
+    /// Enable/disable HTTP endpoint
+    #[arg(long = "server.http.enable", default_value_t = true, action=ArgAction::Set)]
```
flyingsilverfin, line 26: we use this as `--server.http.enable=false` to disable right?

server/server.rs
```diff
@@ -40,23 +40,47 @@ pub struct Server {
...
+            distribution,
+            version,
+            deployment_id,
+            shutdown_sender,
+            shutdown_receiver,
+        )
+        .await
+    }
+
+    pub async fn new_with_external_shutdown(
```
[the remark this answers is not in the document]
flyingsilverfin: That's pretty cool, nice!

server/server.rs
```diff
@@ -218,62 +260,130 @@ impl Server {
     }
 
     pub async fn serve(mut self) -> Result<(), ServerOpenError> {
-        let service = typedb_protocol::type_db_server::TypeDbServer::new(self.typedb_service);
-        let authenticator = Authenticator::new(
-            self.user_manager.clone(),
-            self.authenticator_cache.clone(),
+        Self::print_hello(ASCII_LOGO, self.distribution, self.version, self.config.is_development_mode);
+
+        // TODO: It has been introduced by the rustls dependency (for some reason, we didn't need it before).
```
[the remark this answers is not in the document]
flyingsilverfin: Werid, ok - thanks. Must be a version change somehow.

server/server.rs
```diff
@@ -218,62 +260,130 @@ impl Server {
...
-            .add_service(service)
-            .serve_with_shutdown(self.address, async {
-                // The tonic server starts a shutdown process when this closure execution finishes
-                Self::shutdown_handler(self.shutdown_sender).await;
-            })
-            .await
-            .map_err(|source| ServerOpenError::Serve { address: self.address, source: Arc::new(source) })
+        Self::print_serving_information(grpc_address, http_address);
+
+        if let Some(http_server) = http_server {
```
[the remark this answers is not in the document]
flyingsilverfin: not TOO ugly for me :D maybe dmitrii could do something prettier but not broad scope enough to bother him with

server/server.rs
```diff
@@ -218,62 +260,130 @@ impl Server {
...
-            return Ok(server);
+        mut shutdown_receiver: tokio::sync::watch::Receiver<()>,
+        service: grpc::typedb_service::TypeDBService,
+    ) -> Result<(), ServerOpenError> {
+        let mut grpc_server =
+            tonic::transport::Server::builder().http2_keepalive_interval(Some(GRPC_CONNECTION_KEEPALIVE));
+        if let Some(tls_config) = grpc::encryption::prepare_tls_config(encryption_config)? {
+            grpc_server = grpc_server
+                .tls_config(tls_config)
+                .map_err(|source| ServerOpenError::GRPCTLSFailedConfiguration { source: Arc::new(source) })?;
```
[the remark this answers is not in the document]
flyingsilverfin: Yeah we can use that, maybe just update the error variables in //server for now though so it doesn't scope creep

server/server.rs
```diff
@@ -218,62 +260,130 @@ impl Server {
...
+    ) -> Result<(), ServerOpenError> {
+        let authenticator =
+            http::authenticator::Authenticator::new(credential_verifier, token_manager, diagnostics_manager);
+
+        let encryption_config = http::encryption::prepare_tls_config(encryption_config)?;
+        let http_service = Arc::new(service);
+        let router_service = http::typedb_service::TypeDBService::create_protected_router(http_service.clone())
+            .layer(authenticator)
+            .merge(http::typedb_service::TypeDBService::create_unprotected_router(http_service))
+            // TODO: Allow CORS configuration
```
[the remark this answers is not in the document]
flyingsilverfin: discussed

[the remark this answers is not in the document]
flyingsilverfin: Further, you could check what surreal/mongo/other databases with http endpoints are doing

server/server.rs
```diff
@@ -286,20 +396,30 @@ impl Server {
...
         }
-        println!("Ready!");
     }
 
-    async fn shutdown_handler(shutdown_signal_sender: tokio::sync::watch::Sender<()>) {
-        Self::listen_to_ctrl_c_signal().await;
-        println!("\nReceived CTRL-C. Initiating shutdown...");
-        shutdown_signal_sender.send(()).expect("Expected a successful shutdown signal");
+    fn print_serving_information(grpc_address: SocketAddr, http_address: Option<SocketAddr>) {
+        print!("Serving gRPC on {grpc_address}");
```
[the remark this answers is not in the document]
flyingsilverfin: I think it's ok for now, that's what docs are for :)

server/service/grpc/concept.rs
```diff
@@ -23,27 +23,27 @@ pub(crate) fn encode_thing_concept(
     snapshot: &impl ReadableSnapshot,
     type_manager: &TypeManager,
     thing_manager: &ThingManager,
-    include_thing_types: bool,
+    include_instance_types: bool,
```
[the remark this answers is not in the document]
flyingsilverfin: thanks!

server/service/grpc/error.rs
```diff
@@ -115,6 +108,7 @@ impl IntoGRPCStatus for typedb_protocol::Error {
     fn into_status(self) -> Status {
         let mut details = ErrorDetails::with_error_info(self.error_code, self.domain, HashMap::new());
         details.set_debug_info(self.stack_trace, "");
+        // TODO: Should probably extend these Status codes somehow
```
flyingsilverfin, line 111: Im not sure we can to be honest... it's part of google's built-in grpc messages

server/service/grpc/transaction_service.rs
flyingsilverfin, line 1: Is anything here really new or are you just moving?

server/service/grpc/typedb_service.rs
```diff
@@ -77,9 +90,34 @@ impl TypeDBService {
...
+    ) -> Result<String, AuthenticationError> {
+        let Some(typedb_protocol::authentication::token::create::req::Credentials::Password(password_credentials)) =
+            request.credentials
+        else {
+            return Err(AuthenticationError::InvalidCredential {});
+        };
+
+        self.credential_verifier.verify_password(&password_credentials.username, &password_credentials.password)?;
+
+        Ok(self.token_manager.new_token(password_credentials.username).await)
```
flyingsilverfin, line 113: Looks good!

server/service/grpc/typedb_service.rs
```diff
@@ -88,34 +126,67 @@ impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
...
+                    return Err(err.into_status());
+                } else {
+                    let Some(authentication) = message.authentication else {
+                        return Err(ProtocolError::MissingField {
+                            name: "authentication",
+                            description: "Connection message must contain authentication information.",
+                        }
+                        .into_status());
+                    };
+                    let authentication_result = self
```
flyingsilverfin, line 153: This is guaranteed to be a token right? the error case is handled by the `?` so you can just call it `token`, which makes the `Ok(Response::new(... ))` a bit less confusing

server/service/grpc/typedb_service.rs
```diff
@@ -88,34 +126,67 @@ impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
...
+        )
+        .await
+    }
+
+    async fn authentication_token_create(
+        &self,
+        request: Request<typedb_protocol::authentication::token::create::Req>,
+    ) -> Result<Response<typedb_protocol::authentication::token::create::Res>, Status> {
+        let message = request.into_inner();
+        println!("Got sign in request: {message:?}"); // TODO: Remove
```
flyingsilverfin, line 182: hanging TODO

server/service/grpc/typedb_service.rs
```diff
@@ -326,7 +394,7 @@ impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
         request: Request<Streaming<Client>>,
     ) -> Result<Response<Self::transactionStream>, Status> {
         let request_stream = request.into_inner();
-        let (response_sender, response_receiver) = channel(10);
```
flyingsilverfin, line 329: good catch!

server/service/http/error.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+);
+
+impl HTTPServiceError {
+    pub(crate) fn source(&self) -> &(dyn TypeDBError + Sync + '_) {
+        self.source_typedb_error().unwrap_or(self)
+    }
+
+    pub(crate) fn format_source_trace(&self) -> String {
+        // TODO: Should be reversed? Currently, it's like:
```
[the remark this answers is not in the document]
flyingsilverfin: I think the "code" should be the most detailed error message - most of the "top" ones are going to be something generic like QueryError.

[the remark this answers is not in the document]
flyingsilverfin: To that point I think people are currently enjoying seeing the most specific thing at the 'top' of the stack right? I think in on place we have it that way around, and in another place we have it reversed lol. Perhaps driver vs core or something

server/service/http/message/body.rs
```diff
@@ -0,0 +1,50 @@
...
+    response::{IntoResponse, Response},
+    Json,
+};
+use serde::{de::DeserializeOwned, Serialize};
+
+use crate::service::http::error::HTTPServiceError;
+
+pub(crate) struct JsonBody<T>(pub T);
+
+impl<T> From<Json<T>> for JsonBody<T> {
```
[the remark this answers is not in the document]
flyingsilverfin: That is nice!

server/service/http/message/query/concept.rs
```diff
@@ -0,0 +1,369 @@
...
+
+#[derive(Serialize, Deserialize, Debug)]
+// #[serde(rename_all = "camelCase")] // TODO: driver-based documents don't use camelCase?
+pub struct TypeDocument {
+    kind: String,
+    label: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+// #[serde(rename_all = "camelCase")] // TODO: driver-based documents don't use camelCase?
```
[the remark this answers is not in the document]
flyingsilverfin: ouu interesting i wonder if we can sneak in a change without anyone noticing since I doubt anyone is using that field right now...

server/service/http/message/query/concept.rs
```diff
@@ -0,0 +1,369 @@
...
+        TypeDocument { kind: "relation".to_string(), label: self.label }
+    }
+}
+
+impl Into<AttributeTypeDocument> for AttributeTypeResponse {
+    fn into(self) -> AttributeTypeDocument {
+        AttributeTypeDocument {
+            kind: "attribute".to_string(),
+            label: self.label,
+            value_type: self.value_type.unwrap_or_else(|| "none".to_string()),
```
flyingsilverfin, line 167: doesn't json use `null` for non-existent fields?

server/service/http/message/query/concept.rs
```diff
@@ -0,0 +1,369 @@
...
+        } else {
+            None
+        },
+    })
+}
+
+// TODO: Fix, can give invalid answerss
+fn encode_iid<const ARRAY_INLINE_SIZE: usize>(iid: Bytes<'_, ARRAY_INLINE_SIZE>) -> String {
+    HexBytesFormatter::owned(Vec::from(iid)).format_iid()
+}
```
flyingsilverfin, line 260: why is this able to give invalid answers?

server/service/http/message/query/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct TransactionQueryPayload {
+    pub query_options: Option<QueryOptionsPayload>,
+    pub query: String,
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct QueryPayload {
```
flyingsilverfin, line 52: Why not use the `Request/Response` terminology we use for grpc as well?

common/bytes/util.rs
```diff
@@ -75,6 +75,15 @@ impl<'a> HexBytesFormatter<'a> {
     pub fn borrowed(bytes: &'a [u8]) -> Self {
         Self(Cow::Borrowed(bytes))
     }
+
+    // TODO: Not sure where to place it yet.
```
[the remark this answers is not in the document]
flyingsilverfin: I don't mind having it here to be honest, good idea to have it

common/error/error.rs
```diff
@@ -71,6 +71,21 @@ pub trait TypeDBError {
     fn bottom_source_span(&self) -> Option<::typeql::common::Span> {
         self.source_typedb_error().and_then(|err| err.bottom_source_span()).or_else(|| self.source_span())
     }
+
+    fn stack_trace(&self) -> Vec<String>
```
flyingsilverfin, line 75: nice

common/options/options.rs
```diff
@@ -4,19 +4,46 @@
...
             schema_lock_acquire_timeout_millis: DEFAULT_SCHEMA_LOCK_ACQUIRE_TIMEOUT_MILLIS,
+            transaction_timeout_millis: DEFAULT_TRANSACTION_TIMEOUT_MILLIS,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct QueryOptions {
+    pub include_instance_types: bool,
+    pub answer_count_limit: Option<usize>,
```
flyingsilverfin, line 32: I hav a slight concern that people will be confused that there are _two_ limits. One is based on TypeQL's `limit` clause, and one in the query options.

But I don't see a good way around it, besides making a bold point in the HTTP server documentation about it, and noting in the QueryOptions docs that this is _separate_ from the query `limit` clause and both could be active.

resource/constants.rs
```diff
@@ -24,10 +24,12 @@ pub mod server {
 
     pub const GRPC_CONNECTION_KEEPALIVE: Duration = Duration::from_secs(2 * SECONDS_IN_HOUR);
```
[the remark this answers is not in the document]
flyingsilverfin: Refactoring is out of scope, let's keep accumulating for now

flyingsilverfin (changes requested):
I've done an initial pass over everything except the http messages & services, and http BDD you introduced :) good stuff so far! Will continue tomorrow.

server/authentication/token_manager.rs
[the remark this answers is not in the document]
farost: Done

server/server.rs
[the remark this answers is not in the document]
farost: Gonna rename `TLS` -> `Tls`, `GRPC` -> `Grpc`, `HTTP` -> `Http` in enums

server/server.rs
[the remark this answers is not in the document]
farost: As far as I understand, they don't usually set up CORS through the database configuration, so we should probably leave it as is without a TODO, I agree. But I'll leave it as a separate method

server/service/http/error.rs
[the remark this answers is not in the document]
farost: Okay, I'll try getting the bottom error code then.

server/service/http/message/query/concept.rs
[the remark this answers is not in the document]
farost: I don't actually mind, I would prefer everything to be consistent. Would you like me to update the drivers?

resource/constants.rs
farost: Done!

server/authentication/token_manager.rs
farost: Yes they are not, but it's the JWT convention, and these names are encoded/decoded if needed. See https://jwt.io/

server/parameters/cli.rs
farost: Yes, it's a setter

server/service/grpc/error.rs
farost: I don't remember if I wanted it crucially, so I'll remove the todo. Thanks

server/service/grpc/transaction_service.rs
farost: Moving + reusing common functions from `server/service/transaction_service.rs` + new code for transaction/query options. You can probably check the main `listen` loop and the usage of `timeout_at`

farost: I have a general struggle with interrupt signals and timeouts at the moment, let's discuss it here. 

Currently, we have an `interrupt` which is checked mostly in query executions. This interrupt can be set, for example, on transaction close.
A transaction can be closed based on multiple signals: shutdown, transaction timeout, user credentials update, and probably something else. 
We still need to check these interrupting conditions in blocking calls, because they block the thread, and `listen` does no longer work. At the same time, we are interested in catching it in `listen()` beforehand.
It all slowly becomes spaghetti.

In an ideal world, I'd like to develop a multiplexer receiving many shutdown/interrupt/... signals, and sending a single signal / switching a boolean to notify the whole component that it should stop, including stopping its active blocking queries. However, I could not come up with a good implementation for now, and trying small steps led to crashes like https://github.com/typedb/typedb/issues/7427 (I hope we remove this issue beforehand and continue working in this direction).

For now, as I am a little blocked, I am thinking about leaving this as a separate feature request for the future, disabling some of the signals if they lead to crashes. Do you generally agree with my idea, or do you have other thoughts?

server/service/grpc/typedb_service.rs
farost: Right, I think it's a fossil left forgotten after refactoring

server/service/grpc/typedb_service.rs
farost: I will remove them, but thanks for highlighting =) I have a little bug in drivers/console I will need to investigate

server/service/http/message/query/concept.rs
farost: I understood that I got confused by the way attribute iids were represented:

```
"j": {
        "kind": "attribute",
        "iid": "0x200000084a6f686e00000000000000000000000004",
        "value": {
          "kind": "value",
          "value": "John"
        },
        "type": {
          "kind": "attributeType",
          "label": "name",
          "valueType": "string"
        }
      },
      "p": {
        "kind": "entity",
        "iid": "0x1e00000000000000000000",
        "type": {
          "kind": "entityType",
          "label": "person"
        }
      }
```
We stopped sending attribute IIDs, so it's not seen anymore.

server/service/http/message/query/mod.rs
farost: Because an HTTP request consists of many components we break it down before receiving it in a Rust callback method:
1. Path vars (`/databases/:database-name`) - parsed through, for example, `from_request_parts_impl`
2. Headers (Authorization, content type, etc) 
3. Bodies - we only accept jsons
I don't like the `Body` name here, so I use `Payload`. Multiple path variables for a single request are parsed into a single struct for better structure, so they are called `Path`. 

`Response` is used, and it matches `axum`'s HTTP response terminology.

common/options/options.rs
farost: Do you think this feature will be useful if we make the answer returning more efficient in HTTP in the future? Would you expose this feature to GRPC?
If it's a temporary feature, I think that it's fine to do as you say. If it's something to be used for years, maybe we could think of a better design / naming

krishnangovindraj (commented):
BDD looks ok. 
It feels like there's a lot of logic duplicated between the grpc & http services. I can't say I truly understand either of them, but is there a way to deduplicate?

tests/behaviour/steps/params/lib.rs
```diff
@@ -28,31 +31,29 @@ use encoding::{
...
 #[param(name = "may_error", regex = "(|; fails|; fails with a message containing: \".*\")")]
-pub(crate) enum MayError {
+pub enum MayError {
     False,
     True(Option<String>),
 }
 
 impl MayError {
-    pub fn check<T: fmt::Debug, E: fmt::Debug>(&self, res: Result<T, E>) -> Option<E> {
+    pub fn check<T: fmt::Debug, E: fmt::Debug>(&self, res: Result<T, E>) -> Option<Either<T, E>> {
```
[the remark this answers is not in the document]
farost: Fair! Will do

tests/behaviour/steps/params/lib.rs
```diff
@@ -4,6 +4,9 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+#![deny(unused_must_use)]
+#![deny(elided_lifetimes_in_paths)]
+
```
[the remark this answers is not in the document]
farost: Why? We are in tests.

tests/behaviour/service/http/http_steps/query.rs
```diff
@@ -0,0 +1,906 @@
...
+
+#[apply(generic_step)]
+#[step(expr = r"answer get row\({int}\) query type {is_or_not}: {query_type}")]
+pub async fn answer_get_row_query_type_is(
+    context: &mut Context,
+    index: usize,
+    is_or_not: IsOrNot,
+    query_type: crate::params::QueryType,
+) {
+    // no-op
```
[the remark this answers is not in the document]
farost: For drivers.

tests/behaviour/service/http/http_steps/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+typedb_error! {
+    pub HttpBehaviourTestError(component = "HTTP Behaviour Test", prefix = "HTB") {
+        InvalidConceptConversion(1, "Invalid concept conversion."),
+        InvalidValueRetrieval(2, "Could not retrieve a '{type_}' value.", type_: String),
+        HttpError(3, "Http error.", source: Arc<http::Error>),
+        HyperError(4, "Hyper error.", source: Arc<hyper::Error>),
+        StatusError(5, "Status Error {code}: {message}", code: StatusCode, message: String),
+        UnavailableRowVariable(6, "Cannot get concept from a concept row by variable '{variable}'.", variable: String),
+    }
+}
```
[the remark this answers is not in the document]
farost: Note that these steps implement driver features, and there are operations not supported in HTTP due to its API specifics. These errors "emulate" some of the checked errors. 
Ideally, we could create a separate file for HTTP, but I don't see a need for it, I better use these tricks where needed.

tests/behaviour/service/http/http_steps/query.rs
```diff
@@ -0,0 +1,906 @@
...
+    context: &mut Context,
+    index: usize,
+    var_kind: ConceptKind,
+    is_by_var_index: IsByVarIndex,
+    var: Var,
+    is_date: params::Boolean,
+) {
+    if matches!(is_by_var_index, IsByVarIndex::Is) {
+        return; // http does not have indices
+    }
```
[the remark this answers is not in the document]
farost: Return. Http does not have indices. It's no-op.

server/authentication/token_manager.rs
flyingsilverfin: hmm ok thanks

server/service/http/message/query/mod.rs
flyingsilverfin: ok

server/service/grpc/transaction_service.rs
flyingsilverfin: So how I remember it/just skimmed it:

- within a transaction service, there's a simple Interrupt signal that should reliably be able to terminate all running queries. On QueryInterrupt signal, they can be ended and removed.
- the transaction service also tracks running (eg, running in a background blocking thread) versus those that are queued (not run yet) or transmitting (async task sending batches of answers). These can be removed without the interrupt.

So that means there is a mechanism for a transaction to be closed reliably, if you can get access to the close():
transactions in the general path "close themselves", since they have an async `listen()` loop, that will eventually get a close/commit signal from the user, trigger query interrupt, clean up state, and then exit the listen() loop. Tokio will then remove that Task & therefore the entire TransactionService from the event loop!

To shutdown a server, we also have a shutdown signal that is passed in that the listen loop listens on - at which point, we should interrupt queries, close resources, and exit the listen loop as well.

I agree it's not ideal to "keep adding more listeners", such as another one that is a channel that delivers the usernames that have been invalidated and that could lead to services closing themselves... 

One medium-sized idea is that we can in the TypeDBService extract a kind of "TransactionServiceHandle" object that contains some information about the transaction service (maybe the copy username, and a source/sink to close the transaction?). We could then keep the whole `listen` loop, but instead of listening to a shared `shutdown sink`, we listen to a `close sink`, which we can trigger during a shutdown (shutdown -> close typedb service -> trigger all TransactionServiceHandle::close). We could also use the same triggers on auth change, etc... it's probably abit out of scope for now? I also worry that the collection that holds these TransactionServiceHandles will have to be behind a 'mutex', and eventually under big contention cause a bottleneck. That being said, probably transactions + queries are way slower than a map/vec of transaction handles being held somewhere...)

tests/behaviour/service/http/http_steps/message.rs
flyingsilverfin: Looks good! Thanks!

tests/behaviour/service/http/http_steps/message.rs
```diff
@@ -0,0 +1,377 @@
...
+    context: &HttpContext,
+    transaction_id: &str,
+    query: &str,
+) -> Result<QueryAnswerResponse, HttpBehaviourTestError> {
+    let url = format!("{}/transactions/{}/query", Context::default_versioned_endpoint(), transaction_id);
+    let json_body = json!({
+        "query": query,
+    });
+    let response = send_request(context, Method::POST, &url, Some(json_body.to_string().as_str())).await?;
+    println!("Response: {response}");
```
flyingsilverfin: printout

server/service/http/transaction_service.rs
```diff
@@ -0,0 +1,1091 @@
...
+    prefetch_size: Option<u64>,
+
+    is_open: bool,
+    transaction: Option<Transaction>,
+    query_queue: VecDeque<(TransactionResponder, QueryOptions, typeql::query::Pipeline, String)>,
+    running_write_query: Option<(TransactionResponder, JoinHandle<(Transaction, WriteQueryResult)>)>,
+}
+
+#[derive(Debug)]
+pub(crate) enum TransactionServiceResponse {
```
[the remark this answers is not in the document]
farost: No, `ResponsePart` is more misleading. It's not a header/path/anyothermetadata part of an HTTP Response. I think I'll leave it as is as it's declared inside the service, not in the `message` subpackage

resource/constants.rs
```diff
@@ -40,14 +45,19 @@ pub mod server {
         'W', 'X', 'Y', 'Z', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
     ];
 
-    pub const AUTHENTICATOR_USERNAME_FIELD: &str = "username";
-    pub const AUTHENTICATOR_PASSWORD_FIELD: &str = "password";
-    pub const AUTHENTICATOR_CACHE_TTL: Duration = Duration::from_secs(3 * SECONDS_IN_HOUR);
-    pub const AUTHENTICATOR_CACHE_TTI: Duration = Duration::from_secs(1 * SECONDS_IN_HOUR);
+    // TODO: Maybe we start moving these options to separate crates?
+    pub const HTTP_USERNAME_FIELD: &str = "username";
+    pub const HTTP_PASSWORD_FIELD: &str = "password";
```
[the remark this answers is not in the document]
farost: It's not needed, good catch. My IDE was not showing it correctly. I'll remove the consts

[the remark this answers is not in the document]
farost: I also moved the other HTTP consts into the single file they were requested from

server/server.rs
```diff
@@ -218,62 +259,130 @@ impl Server {
...
 
     pub async fn serve(mut self) -> Result<(), ServerOpenError> {
-        let service = typedb_protocol::type_db_server::TypeDbServer::new(self.typedb_service);
-        let authenticator = Authenticator::new(
-            self.user_manager.clone(),
-            self.authenticator_cache.clone(),
+        Self::print_hello(ASCII_LOGO, self.distribution, self.version, self.config.is_development_mode);
+
+        // TODO: It has been introduced by the rustls dependency (for some reason, we didn't need it before).
+        // Leave this call for now, but be aware that it can cause issues on different machines.
```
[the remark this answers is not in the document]
farost: I'm just afraid of these certificates after the docker issues. We'll see...

server/service/http/authenticator.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    fn poll_ready(&mut self, cx: &mut std::task::Context<'_>) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, request: Request<Body>) -> Self::Future {
+        let authenticator = self.authenticator.clone();
+        let mut inner = self.inner.clone();
+        Box::pin(async move {
+            let req = tokio::task::spawn(async move { authenticator.authenticate(request).await }).await.unwrap();
```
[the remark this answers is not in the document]
farost: Probably not, it's been in the grpc service before me, so I just copypasted. Will remove

server/service/http/transaction_service.rs
```diff
@@ -0,0 +1,1132 @@
...
+        };
+        self.diagnostics_manager.increment_load_count(&database_name, transaction.to_load_kind());
+        self.transaction = Some(transaction);
+        self.timeout_at = init_transaction_timeout(Some(transaction_timeout_millis));
+
+        let processing_time_millis = Instant::now().duration_since(receive_time).as_millis() as u64;
+        Ok(processing_time_millis)
+    }
+
+    pub(crate) async fn listen(&mut self) {
```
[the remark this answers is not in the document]
farost: I intentionally decided to let it be a separate thing without trait generalization and stuff for the first version, because something may change. The rewrite will not be fast, I think, and the context will remain the same, so it's better to leave for the future if we are confident that the approach will remain the same.

server/service/http/typedb_service.rs
```diff
@@ -0,0 +1,584 @@
...
+
+    async fn databases_get(
+        _version: ProtocolVersion,
+        State(service): State<Arc<TypeDBService>>,
+        database_path: DatabasePath,
+    ) -> impl IntoResponse {
+        run_with_diagnostics(
+            &service.diagnostics_manager,
+            Some(&database_path.database_name),
+            ActionKind::DatabasesContains,
```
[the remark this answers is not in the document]
farost: I wanted to use the existing names in GRPC and HTTP where possible. But I can avoid it.

diagnostics/lib.rs
farost, line 1: @flyingsilverfin I refactored diagnostics:

First. I understood that metrics are a mess. We have different potential endpoints (posthog / anything else), we have different formats / meaning (diffs and states for posthog / prometheus / monitoring), and bla bla bla. While I wanted to preserve it "efficient" and use a single set of `metrics` structs, it was very painful to maintain and understand.

A step into a better direction: now we have multiple structs for each specific purpose!
1. `metrics` to store metrics. The format is the optimal format for the server to record data without slowing down actual operations.
2. `reports/mod.rs` with general reports. They describe general data fetched from `metrics`.
3. `reports/***.rs` specific reports for different endpoints. Does PostHog expect the fields to be written in snake_case, while the standard json format is camelCase? Does PostHog expect a set of mandatory fields to get specified? Do we want to format data differently for PostHog? Do we want to split data by endpoints for PostHog, but record a sum for all endpoints for Prometheus or Monitoring? No problem, you have separate reports for these purposes!

The logic is still not awesome sometimes, and some more generalizing could be used, but it's already out of scope. I'm quite happy with how it looks and works, currently, and I hope that it will make the maintenance more pleasant.

And yes, we removed the TypeDB Diagnostics Service reporting.

diagnostics/reports/prometheus_monitoring.rs
```diff
@@ -0,0 +1,132 @@
...
+use crate::{
+    reports::json_monitoring::{to_monitoring_report, JsonMonitoringReport},
+    Diagnostics,
+};
+
+pub fn to_monitoring_prometheus(diagnostics: &Diagnostics) -> String {
+    to_prometheus(to_monitoring_report(diagnostics))
+}
+
+pub fn to_prometheus(report: JsonMonitoringReport) -> String {
```
farost, line 15: It's much better in a single function.

BUILD
```diff
@@ -399,8 +399,7 @@ label_flag(
 )
 
 # apt
-# TODO: Remove when console no longer requires jre
```
farost, line 402: The time has come (an RC release is incoming, so we'll see how well it works)

binary/typedb
```diff
@@ -17,8 +17,7 @@ fi
 # ================================================
 
 print_usage() {
-    # TODO: Add [--help] back to server
-    echo "  Server:          typedb server"
+    echo "  Server:          typedb server [--help]"
```
farost, line 20: The time has come

config/brew/typedb.rb
```diff
@@ -19,14 +19,8 @@ class Typedb < Formula
...
   license "MPL-2.0"
 
-  # TODO: Remove when console no longer requires jre
-  depends_on "openjdk"
-
-
   def install
     libexec.install Dir["*"]
     bin.install libexec / "typedb"
-    # TODO: Remove when console no longer requires jre
```
farost, line 29: The time has come (an RC release is incoming, so we'll see how well it works)

diagnostics/lib.rs
flyingsilverfin: Sounds like a good decision, thank you!

.factory/automation.yml
```diff
@@ -156,6 +156,13 @@ build:
         bazel test //tests/behaviour/query/functions:test_basic --test_output=streamed
         bazel test //tests/behaviour/query/functions:test_recursion --test_output=streamed
 
+    test-behaviour-service:
+      image: typedb-ubuntu-22.04
+      dependencies: [build]
+      command: |
+        bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
+        bazel test //tests/behaviour/service/... --test_output=streamed --jobs=1
```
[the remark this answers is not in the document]
flyingsilverfin: Fascinating and weird, thanks, let's move on

flyingsilverfin (approved):

.factory/automation.yml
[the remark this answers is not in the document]
farost: I removed an unnecessary timeout between steps, it became:

//tests/behaviour/service/http:test_http_driver                          PASSED in 108.2s
//tests/behaviour/service/http:test_http_user                            PASSED in 35.9s
//tests/behaviour/service/http/connection:test_http_database             PASSED in 15.0s
//tests/behaviour/service/http/connection:test_http_transaction          PASSED in 36.8s

Already better: almost all tests are slightly slower. Will see if there is anything else affecting the performance.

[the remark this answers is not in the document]
farost: Without server starts/stops:
//tests/behaviour/service/http/connection:test_http_database    PASSED in 12.8s
//tests/behaviour/service/http:test_http_driver                          PASSED in 106.4s
//tests/behaviour/service/http:test_http_user                            PASSED in 32.8s
//tests/behaviour/service/http/connection:test_http_transaction          PASSED in 33.2s

We don't wait for the server start, so other steps generally take less time. Queries in `http_driver` probably make the difference.
Still room for investigation, I'll leave it at the previous results.

PR #7402, opened 2025-03-25
Introduce branch provenance tracking

krishnangovindraj:
Introduce branch provenance tracking. This allows us to determine which branch of a disjunction a given answer originates from. Though meaningless in the logical semantics, this is helpful for other applications like visualisations.

## Motivation
Provenance & query-structure tracking for visualisations.  

## Implementation
- Introduces `branch_id` in IR disjunctions
- Introduces one `u64` per row for branch provenance tracking.
- `DisjunctionExecutor::map_output` sets the bit corresponding to the `branch_id` on the row's provenance.
- Introduces the `QueryStructure` structure with the information to reconstruct the query IR from the cached executable pipeline - so we may send it to the client side.

### DIFF
--- a/compiler/executable/match_/planner/mod.rs
+++ b/compiler/executable/match_/planner/mod.rs
@@ -13,7 +13,7 @@ use answer::variable::Variable;
 use concept::thing::statistics::Statistics;
 use error::typedb_error;
 use ir::{
-    pattern::{constraint::ExpressionBinding, Vertex},
+    pattern::{constraint::ExpressionBinding, Vertex, BranchID},
     pipeline::{block::Block, function_signature::FunctionID, VariableRegistry},
 };
 use itertools::Itertools;
@@ -85,6 +85,7 @@ pub fn compile(
         selected_variables.iter().copied(),
         &assigned_identities,
         variable_registry,
+        None,
     )
     .map_err(|source| MatchCompilationError::PlanningError { typedb_source: source })?
     .finish(variable_registry);
@@ -131,12 +132,13 @@ impl NegationBuilder {
 
 #[derive(Debug)]
 struct DisjunctionBuilder {
+    branch_ids: Vec<BranchID>,
     branches: Vec<MatchExecutableBuilder>,
 }
 
 impl DisjunctionBuilder {
-    fn new(branches: Vec<MatchExecutableBuilder>) -> Self {
-        Self { branches }
+    fn new(branch_ids: Vec<BranchID>, branches: Vec<MatchExecutableBuilder>) -> Self {
+        Self { branch_ids, branches }
     }
 }
 
@@ -247,9 +249,9 @@ impl StepBuilder {
             StepInstructionsBuilder::Negation(NegationBuilder { negation }) => ExecutionStep::Negation(
                 NegationStep::new(negation.finish(variable_registry), selected_variables, output_width),
             ),
-
-            StepInstructionsBuilder::Disjunction(DisjunctionBuilder { branches }) => {
+            StepInstructionsBuilder::Disjunction(DisjunctionBuilder { branch_ids, branches }) => {
                 ExecutionStep::Disjunction(DisjunctionStep::new(
+                    branch_ids,
                     branches.into_iter().map(|builder| builder.finish(variable_registry)).collect(),
                     selected_variables,
                     output_width,
@@ -288,10 +290,12 @@ struct MatchExecutableBuilder {
     next_output: VariablePosition,
 
     planner_statistics: PlannerStatistics,
+    branch_id: Option<BranchID>,
 }
 
 impl MatchExecutableBuilder {
     fn new(
+        branch_id: Option<BranchID>,
         assigned_positions: &HashMap<Variable, ExecutorVariable>,
         selected_variables: Vec<Variable>,
         input_variables: Vec<Variable>,
@@ -309,6 +313,7 @@ impl MatchExecutableBuilder {
             .unwrap_or(0);
         let next_output = VariablePosition::new(next_position);
         Self {
+            branch_id,
             selected_variables,
             input_variables,
             current_outputs,
--- a/compiler/executable/match_/planner/mod.rs
+++ b/compiler/executable/match_/planner/mod.rs
@@ -13,7 +13,7 @@ use answer::variable::Variable;
 use concept::thing::statistics::Statistics;
 use error::typedb_error;
 use ir::{
-    pattern::{constraint::ExpressionBinding, Vertex, BranchID},
+    pattern::{constraint::ExpressionBinding, BranchID, Vertex},
     pipeline::{block::Block, function_signature::FunctionID, VariableRegistry},
 };
 use itertools::Itertools;
--- a/compiler/executable/match_/planner/plan.rs
+++ b/compiler/executable/match_/planner/plan.rs
@@ -25,7 +25,7 @@ use ir::{
         },
         nested_pattern::NestedPattern,
         variable_category::VariableCategory,
-        Scope, Vertex,
+        BranchID, Scope, ScopeId, Vertex,
     },
     pipeline::{block::BlockContext, VariableRegistry},
 };
@@ -124,6 +124,7 @@ fn make_builder<'a>(
         match pattern {
             NestedPattern::Disjunction(disjunction) => {
                 let planner = DisjunctionPlanBuilder::new(
+                    disjunction.branch_ids().iter().copied().collect(),
                     disjunction
                         .conjunctions()
                         .iter()
@@ -1324,8 +1325,10 @@ impl ConjunctionPlan<'_> {
         selected_variables: impl IntoIterator<Item = Variable> + Clone,
         already_assigned_positions: &HashMap<Variable, ExecutorVariable>,
         variable_registry: &VariableRegistry,
+        branch_id: Option<BranchID>,
     ) -> Result<MatchExecutableBuilder, QueryPlanningError> {
         let mut match_builder = MatchExecutableBuilder::new(
+            branch_id,
             already_assigned_positions,
             selected_variables.clone().into_iter().collect(),
             input_variables.clone().into_iter().collect(),
@@ -1548,6 +1551,7 @@ impl ConjunctionPlan<'_> {
                     match_builder.selected_variables.iter().copied(),
                     match_builder.position_mapping(),
                     variable_registry,
+                    None,
                 )?;
                 let variable_positions: HashMap<Variable, ExecutorVariable> = negation
                     .index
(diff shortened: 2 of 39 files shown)


executor/read/nested_pattern_executor.rs
```diff
@@ -8,39 +8,43 @@ use std::sync::Arc;
...
         self.branches.iter_mut().for_each(|branch| branch.reset())
     }
 
-    pub(crate) fn map_output(&self, unmapped: FixedBatch) -> FixedBatch {
+    pub(crate) fn map_output(&self, branch_index: BranchIndex, unmapped: FixedBatch) -> FixedBatch {
         let mut uniform_batch = FixedBatch::new(self.output_width);
         unmapped.into_iter().for_each(|row| {
             uniform_batch.append(|mut output_row| {
                 output_row.copy_mapped(row, self.selected_variables.iter().map(|&pos| (pos, pos)));
+                output_row.set_branch_id_in_provenance(self.branch_ids[*branch_index]);
```
krishnangovindraj, line 47: `DisjunctionExecutor` sets the `branch_id` on `map_output`.

krishnangovindraj (commented):
self review

compiler/executable/match_/planner/mod.rs
```diff
@@ -131,12 +132,13 @@ impl NegationBuilder {
 
 #[derive(Debug)]
 struct DisjunctionBuilder {
+    branch_ids: Vec<BranchID>,
```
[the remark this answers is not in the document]
flyingsilverfin: I think keeping them "together" makes sense. Pairs work, but why not a Map? BranchID -> ExecutableBuilder? Not a critical code path, so we dont _have_ to use Vecs

Maybe that can be used elsewhere too where it's not a critical code path.

compiler/executable/match_/planner/plan.rs
```diff
@@ -1987,13 +1991,18 @@ impl ConjunctionPlan<'_> {
...
     branches: Vec<ConjunctionPlanBuilder<'a>>,
     required_inputs: Vec<Variable>,
 }
 
 impl<'a> DisjunctionPlanBuilder<'a> {
-    fn new(branches: Vec<ConjunctionPlanBuilder<'a>>, required_inputs: Vec<Variable>) -> Self {
-        Self { branches, required_inputs }
+    fn new(
+        branch_ids: Vec<BranchID>,
+        branches: Vec<ConjunctionPlanBuilder<'a>>,
```
flyingsilverfin, line 2002: Could also be maps IMO

compiler/executable/pipeline.rs
```diff
@@ -534,3 +553,108 @@ fn find_referenced_functions_in_fetch(
         })
     }
 }
+
```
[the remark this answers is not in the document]
flyingsilverfin: Yeah i think so, it's a top-level idea and the result of some complexity

compiler/executable/pipeline.rs
```diff
@@ -534,3 +552,113 @@ fn find_referenced_functions_in_fetch(
...
+#[derive(Debug, Clone)]
+pub struct ParametrisedQueryStructure {
+    pub branches: [Option<Vec<Constraint<Variable>>>; 64],
```
flyingsilverfin, line 558: are we limited to 64 branches throughout the entire query now? Can we relax this?

executor/batch.rs
```diff
@@ -33,15 +34,28 @@ pub struct FixedBatch {
...
+        multiplicities: FixedBatch::INIT_MULTIPLICITIES,
+        provenance: FixedBatch::INIT_PROVENANCE,
+    };
+
+    pub(crate) const EMPTY: FixedBatch = FixedBatch {
+        width: 0,
+        entries: 0,
+        data: Vec::new(),
+        multiplicities: FixedBatch::INIT_MULTIPLICITIES,
+        provenance: FixedBatch::INIT_PROVENANCE,
```
flyingsilverfin, line 57: Either we can make both plural or both singular!

flyingsilverfin: including the constants

executor/lib.rs
```diff
@@ -106,3 +107,21 @@ impl Clone for ExecutionInterrupt {
...
+        if id.0 < 64 {
+            self.0 |= 1 << id.0
+        }
+    }
+
+    pub(crate) fn branch_ids(&self) -> impl Iterator<Item = BranchID> {
+        let provenance = self.0;
+        (0..64).filter(move |id| 0 != provenance & (1 << id)).map(|id| BranchID(id))
+    }
+}
```
flyingsilverfin, line 127: Ah i see we've really stuck into the 64 branches in the entire query here?

flyingsilverfin: or rather, throughout?

executor/pipeline/fetch.rs
```diff
@@ -376,11 +378,12 @@ fn execute_list_subfetch(
             initial_row[local_row_position.as_usize()] =
                 row[parent_row_position.as_usize()].as_reference().into_owned();
         });
-        let initial_row = MaybeOwnedRow::new_owned(initial_row, row.multiplicity());
+        let initial_row = MaybeOwnedRow::new_owned(initial_row, row.multiplicity(), Provenance::INITIAL);
         Pipeline::build_read_pipeline(
             snapshot,
             thing_manager,
             variable_registry.variable_names(),
+            Arc::new(ParametrisedQueryStructure::empty()), // TODO:?
```
flyingsilverfin, line 386: todo?

executor/read/nested_pattern_executor.rs
flyingsilverfin: ok

executor/read/nested_pattern_executor.rs
```diff
@@ -8,39 +8,43 @@ use std::sync::Arc;
...
+        debug_assert!(branch_ids.len() == branches.len());
+        Self { branches, branch_ids, selected_variables, output_width }
     }
 
     pub(crate) fn reset(&mut self) {
         self.branches.iter_mut().for_each(|branch| branch.reset())
     }
 
-    pub(crate) fn map_output(&self, unmapped: FixedBatch) -> FixedBatch {
+    pub(crate) fn map_output(&self, branch_index: BranchIndex, unmapped: FixedBatch) -> FixedBatch {
```
flyingsilverfin, line 42: maybe 'branch_index' can be more descriptive as 'source branch index'?

executor/row.rs
```diff
@@ -93,29 +116,38 @@ impl fmt::Display for Row<'_> {
 pub struct MaybeOwnedRow<'a> {
     row: Cow<'a, [VariableValue<'static>]>,
     multiplicity: Cow<'a, u64>,
+    provenance: Cow<'a, Provenance>, // TODO: Review: are we not better off without the Cow?
```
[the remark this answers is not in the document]
flyingsilverfin: Good question, not sure! If in doubt it's probably not harmful to have it for now?

ir/pattern/mod.rs
```diff
@@ -32,6 +32,9 @@ pub mod expression;
 pub mod function_call;
 pub mod nested_pattern;
 
+#[derive(Debug, Copy, Clone, Hash, Eq, PartialEq)]
+pub struct BranchID(pub u16);
```
flyingsilverfin, line 36: It's not clear to me from the code or the PR description yet what the relationship is between Provenance (and that 64 limit I saw), and this BranchID? Both of them seem to impose some limit on query structure we can represent, but which one does what?

ir/pipeline/mod.rs
```diff
@@ -118,13 +119,19 @@ impl VariableRegistry {
...
             variable_categories: HashMap::new(),
             variable_optionality: HashMap::new(),
             variable_source_spans: HashMap::new(),
         }
     }
+    pub(crate) fn next_branch_id(&mut self) -> BranchID {
+        let branch_id = self.branch_id_allocator;
+        self.branch_id_allocator += 1;
+        BranchID(branch_id)
+    }
```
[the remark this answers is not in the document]
flyingsilverfin: works for me !

server/service/query_structure.rs
```diff
@@ -0,0 +1,115 @@
...
+ */
+
+use std::collections::HashMap;
+
+use answer::variable::Variable;
+use compiler::executable::pipeline::QueryStructure;
+use ir::pattern::{constraint::Constraint, Vertex};
+use itertools::Itertools;
+
+const JSON_NONE: &str = "{}";
```
flyingsilverfin, line 14: Is there no way we can encode direcrtly into a JSON type or a Map -> Serialize? 
Is this temporary to enable your testing and this will be refined later?

server/service/transaction_service.rs
```diff
@@ -1342,6 +1342,9 @@ impl TransactionService {
             let named_outputs = pipeline.rows_positions().unwrap();
             let descriptor: StreamQueryOutputDescriptor = named_outputs.clone().into_iter().sorted().collect();
             let initial_response = StreamQueryResponse::init_ok_rows(&descriptor, Read);
+            #[cfg(debug_assertions)]
+            dump_query_structure(pipeline.query_structure());
```
flyingsilverfin, line 1346: This should be on `trace!` + the debug flag you have if you want to check it into master IMO

flyingsilverfin:
Hmm ok reading your Pr description again, i'm assuming we can do max of u16::MAX branches in 1 query/pipeline. But then, doesn't a single row originate from potentially multiple leaf branches that combine? How do we fit that into a u64 value in the row? I think that would be good to write into the PR description.

krishnangovindraj:
@flyingsilverfin , We can't do `u16::MAX` branches. We can only do 64.

ir/pattern/mod.rs
krishnangovindraj: Ah I see where the confusion is from. This could easily be u8 (or technically u6).

executor/pipeline/fetch.rs
krishnangovindraj: I don't know if we want query structures within `fetch`. May as well not have them.

server/service/query_structure.rs
krishnangovindraj: Yes. It's already gone in my branch on http.

compiler/executable/match_/planner/plan.rs
krishnangovindraj: Best seems to be to introduce a new `struct Branch`, but it doesn't seem worth it at the moment.

executor/pipeline/fetch.rs
flyingsilverfin: yeah no need

flyingsilverfin:
Can we bump it to a u128 and support 128 branches? Feels more comfortable to me somehow - 64 is quite low esp if you have a moderately long pipeline?

And then, if your query has more branches than that I think we can make the QueryStructure optional, and not record anything if you have too many branches, rather than failing the query? All it means we can't provide the graph structure to the user right?

krishnangovindraj:
I've made QueryStructure optional when the total branch count exceeds 64. 
I don't think we should make it a u128 unless we're sure every client language will have a convenient way to handle u128s.
It's only 2x more branches, which isn't much. We should just replace it with an arbitrary number of bytes when we can.

64 bits gives you 32 binary disjunctions. I don't think that's too few since disjunctions seem quite rare when polymorphism is done properly. It's also unlikely that you can make much sense of a visualisation that size.

flyingsilverfin:
Ok it being optional makes sense to me. Let's go!

flyingsilverfin (approved):

executor/lib.rs
```diff
@@ -106,3 +107,21 @@ impl Clone for ExecutionInterrupt {
...
+pub struct Provenance(u64);
+
+impl Provenance {
+    pub const INITIAL: Provenance = Provenance(0);
+
+    pub(crate) fn set_branch_id(&mut self, id: BranchID) {
+        if id.0 < 64 {
+            self.0 |= 1 << id.0
+        }
+    }
```
krishnangovindraj, line 121: (To link to)

PR #7403, opened 2025-03-26
Fix write transactions left open and blocking other transactions after receiving errors

farost:
The gRPC service no longer leaves any of the transactions open after receiving transaction or query errors. This eliminates scenarios where schema transactions cannot open due to other hanging schema or write transactions and where read transactions can crash the server using incorrect operations (commits, rollbacks, ...). Additionally, it ensures the correctness of the load diagnostics data.

## Motivation
Lost hanging transactions cause temporary transaction deadlocks, can lead to server crashes on `unwrap`s, and pollute the diagnostics data. 

## Implementation
For every `transaction.take()` call, we always have to set the transaction back again. This approach still introduces the risk that we'll eventually forget to reassign the taken transaction, so it is not the best, but all the (5) cases are covered now, so it should be fine. Without it, `do_close()` called on the final `Break`s does nothing with the transaction.
I already had an initial step in this direction [in January](https://github.com/typedb/typedb/pull/7300/files#diff-65d0af60103fa2b39e70ecdca66437a8cff5876b62207d98d57f8fabe83ecbd6R591), but I was sure that we are not supposed to return transactions on errors as their closing is additionally handled. Too bad, it was not.

Additional behaviour scenarios will be written to verify this fix through both the Server's tests and the Driver's tests.

### DIFF
--- a/server/service/transaction_service.rs
+++ b/server/service/transaction_service.rs
@@ -566,7 +566,8 @@ impl TransactionService {
 
         let diagnostics_manager = self.diagnostics_manager.clone();
         match self.transaction.take().unwrap() {
-            Transaction::Read(_) => {
+            Transaction::Read(transaction) => {
+                self.transaction = Some(Transaction::Read(transaction));
                 Err(TransactionServiceError::CannotCommitReadTransaction {}.into_error_message().into_status())
             }
             Transaction::Write(transaction) => spawn_blocking(move || {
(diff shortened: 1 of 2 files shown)


tests/behaviour/steps/connection/mod.rs
```diff
@@ -9,9 +9,9 @@ use std::{
     fmt,
     sync::{Arc, Mutex},
 };
-use resource::constants::server::ASCII_LOGO;
 
 use macro_rules_attribute::apply;
+use resource::constants::server::ASCII_LOGO;
```
farost, line 14: rustfmt

server/service/transaction_service.rs
```diff
@@ -838,54 +839,59 @@ impl TransactionService {
...
+                    let message_ok_done =
+                        result.map(|_| query_res_ok_done(typedb_protocol::query::Type::Schema)).map_err(|err| {
+                            TransactionServiceError::TxnAbortSchemaQueryFailed { typedb_source: *err }
+                                .into_error_message()
+                                .into_status()
+                        })?;
+
+                    return Ok(ImmediateQueryResponse::ok(message_ok_done));
+                }
+                transaction @ _ => self.transaction = Some(transaction),
```
dmitrii-ubskii, line 890:
```suggestion
                transaction => self.transaction = Some(transaction),
```

dmitrii-ubskii (approved):

flyingsilverfin (approved):

PR #7406, opened 2025-03-27
Flag pattern as unsatisfiable if any vertex has no empty annotations

krishnangovindraj:
Patterns are now flagged as unsatisfiable if any vertex has empty annotations. Previously, we checked only variables for empty annotations.

## Motivation
Cover missing cases.

## Implementation
Update check to not filter out variables.

### DIFF
--- a/compiler/annotation/match_inference.rs
+++ b/compiler/annotation/match_inference.rs
@@ -254,16 +254,11 @@ impl TypeInferenceGraph<'_> {
         &self,
         variable_registry: &VariableRegistry,
     ) -> Result<(), TypeInferenceError> {
-        let mut thing_variable_present = false;
-        let mut any_variable_empty = false;
-        self.vertices.annotations.iter().filter_map(|(var, types)| var.as_variable().map(|v| (v, types))).for_each(
-            |(var, types)| {
-                thing_variable_present =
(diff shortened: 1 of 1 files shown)


compiler/annotation/match_inference.rs
```diff
@@ -254,16 +254,11 @@ impl TypeInferenceGraph<'_> {
...
-            },
-        );
-        if any_variable_empty && thing_variable_present {
+        let thing_variable_present = self.vertices.annotations.iter().filter_map(|(var, _)| var.as_variable())
+            .any(|var| variable_registry.get_variable_category(var).unwrap().is_category_thing());
+
+        let any_vertex_empty = self.vertices.annotations.iter().any(|(_, types)| types.is_empty());
+        if any_vertex_empty && thing_variable_present {
             return Err(TypeInferenceError::DetectedUnsatisfiablePattern {});
         }
```
dmitrii-ubskii, line 263: Will this still break if we get a schema query with no annotations for a label vertex?

krishnangovindraj: No, because we'll actually replace the whole constraint with "Unsatisfiable".
It does feel a little brittle when you put it that way, so maybe we should update executor (later) to be resilient to empty annotations.

krishnangovindraj:
Closing in favour of a combined PR: https://github.com/typedb/typedb/pull/7410

PR #7407, opened 2025-03-27
Function call arguments are now stored in a Vec

krishnangovindraj:
Function call arguments are now stored in a Vec according to argument index, instead of a mapping from variable to index.

## Motivation
Avoids a bug introduced by assuming `argument_ids()` would return the variables according to their argument indexes.

## Implementation
Replace the `call_variable_mapping: BTreeMap<Variable, usize>` field in `FunctionCall` with `arguments: Vec<Variable>` & propagate.


krishnangovindraj:
Closing in favour of a combined PR: https://github.com/typedb/typedb/pull/7410

PR #7409, opened 2025-03-28
Optimize unique constraints operation time checks for write queries

farost:
Modify the algorithm of attribute uniqueness validation at operation time, performed for each inserted `has` instance of ownerships with `@unique` and `@key` annotations. 
These changes boost the performance of insert operations for constrained `has`es by up to 35 times, resulting in an almost indiscernible effect compared to the operations without uniqueness checks. 

## Motivation
Uniqueness validations for inserts had a disastrous effect on inserts' performance.

## Implementation
The old algorithm iterated over all objects of the suitable object types (with `owns` under the constraint verified). Then, each object was verified as not having an attribute with the value verified. This straightforward search took more and more time with the increased number of objects in the database, even if they didn't have common attributes: it was bound only to the schema.

The new algorithm performs the reversed sequence of actions. First, it searches for attributes with the value verified. Then, these attributes (they can be of different attribute types) are checked on the existence of `has` edges to objects of the verified set of types. This way, we limit the number of elements and storage lookups processed, making better use of the data specifics and the iterator optimizations.

Using a shortened version of [a small benchmark](https://discord.com/channels/665254494820368395/672580553677078548/1354875257382637830), which performs simple insert operations for objects with randomized attributes, we found that a single average insert operation took (*only use this data for relative comparison: these results were gathered in a local environment*):
Non-optimal compilation mode (x35 times better than before):
* 35.65 ms in the old version with uniqueness validations 
* 1.07 ms without uniqueness validations
* 1.09 ms in the new version with uniqueness validations

Optimal compilation mode (x25 times better than before):
* 3.77 ms in the old version with uniqueness validations
* 0.15 ms without uniqueness validations
* 0.15 ms in the new version with uniqueness validations

### DIFF
--- a/concept/thing/attribute.rs
+++ b/concept/thing/attribute.rs
@@ -98,15 +98,6 @@ impl Attribute {
         thing_manager.get_owners_by_type(snapshot, self, owner_type)
     }
 
-    pub fn get_has_by_owner_types(
-        &self,
-        snapshot: &impl ReadableSnapshot,
-        thing_manager: &ThingManager,
-        owner_type_range: &impl RangeBounds<ObjectType>,
-    ) -> HasReverseIterator {
-        thing_manager.get_has_reverse_by_attribute_and_owner_type_range(snapshot, self, owner_type_range)
-    }
-
     pub fn next_possible(&self) -> Attribute {
         let mut bytes = self.vertex.to_bytes().into_array();
         bytes.increment().unwrap();
--- a/concept/thing/attribute.rs
+++ b/concept/thing/attribute.rs
@@ -10,7 +10,6 @@ use std::{
     fmt,
     hash::{Hash, Hasher},
     iter,
-    ops::RangeBounds,
     sync::{Arc, OnceLock},
 };
 
@@ -41,11 +40,11 @@ use crate::{
     edge_iterator,
     error::{ConceptReadError, ConceptWriteError},
     thing::{
-        object::{HasReverseIterator, Object},
+        object::Object,
         thing_manager::ThingManager,
         HKInstance, ThingAPI,
     },
-    type_::{attribute_type::AttributeType, object_type::ObjectType, ObjectTypeAPI},
+    type_::{attribute_type::AttributeType, ObjectTypeAPI},
     ConceptAPI, ConceptStatus,
 };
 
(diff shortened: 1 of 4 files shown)


concept/thing/thing_manager/validation/operation_time_validation.rs
```diff
@@ -418,39 +419,38 @@ impl OperationTimeValidation {
...
-                    if object == owner {
-                        continue;
-                    }
-
-                    for checked_attribute_type in &attribute_and_subtypes {
-                        if object
-                            .has_attribute_with_value(snapshot, thing_manager, *checked_attribute_type, value.clone())
-                            .map_err(|source| Box::new(DataValidationError::ConceptRead { typedb_source: source }))?
-                        {
+                    let mut has_iterator = attribute.get_has_by_owner_types(snapshot, thing_manager, &owner_type_range);
```
flyingsilverfin, line 446: So there's a difference between a Type Range, and the actual types in your subtypes -- I didn't read all the details of the logic, just be sure you're not relying on the thing manager to do anything other than constraint the search to the indicated type range -- for example it won't filter out "gaps":


```
get_has_by_owner_types([ObjectType10, ObjectType40])

-->
has(object type 10 instance 1, attr)
has(object type 11 instance 1, attr)
has(object type 12 instance 1, attr)
has(object type 13 instance 1, attr)
```

but you might not want to check for instances of type 12, 13

flyingsilverfin: This is probably already encompassed elsewhere, just wanted to make sure we are on the same page

farost: It is a good point. I remembered about this thing, but had a quick thought that it's fine. However, it is not: there is a possibility that this random type also has an attribute, but it does not affect the constraint.

The sad thing is that I can't really consistently verify it through a test, but I can at least write one for "random failures" if it's incorrect. Good catch, thank you

concept/thing/attribute.rs
farost, line 1: Will clean up

flyingsilverfin (approved):
Looks good!

concept/thing/thing_manager/validation/operation_time_validation.rs
farost: Done, tested!

PR #7410, opened 2025-03-28
Various fixes around variable usage

krishnangovindraj:
* Extend the check for empty type-annotations to include label vertices
* Store function call arguments in a Vec, so they are ordered by argument index. 
* Flag assignments to variables which were already bound in previous stages
* Flag anonymous variables in unexpected pipeline stages instead of assuming named & unwrapping.

## Motivation
Stability

### DIFF
--- a/compiler/annotation/match_inference.rs
+++ b/compiler/annotation/match_inference.rs
@@ -254,7 +254,11 @@ impl TypeInferenceGraph<'_> {
         &self,
         variable_registry: &VariableRegistry,
     ) -> Result<(), TypeInferenceError> {
-        let thing_variable_present = self.vertices.annotations.iter().filter_map(|(var, _)| var.as_variable())
+        let thing_variable_present = self
+            .vertices
+            .annotations
+            .iter()
+            .filter_map(|(var, _)| var.as_variable())
             .any(|var| variable_registry.get_variable_category(var).unwrap().is_category_thing());
 
         let any_vertex_empty = self.vertices.annotations.iter().any(|(_, types)| types.is_empty());
--- a/compiler/annotation/match_inference.rs
+++ b/compiler/annotation/match_inference.rs
@@ -254,16 +254,11 @@ impl TypeInferenceGraph<'_> {
         &self,
         variable_registry: &VariableRegistry,
     ) -> Result<(), TypeInferenceError> {
-        let mut thing_variable_present = false;
-        let mut any_variable_empty = false;
-        self.vertices.annotations.iter().filter_map(|(var, types)| var.as_variable().map(|v| (v, types))).for_each(
-            |(var, types)| {
-                thing_variable_present =
-                    thing_variable_present || variable_registry.get_variable_category(var).unwrap().is_category_thing();
-                any_variable_empty = any_variable_empty || types.is_empty();
-            },
-        );
-        if any_variable_empty && thing_variable_present {
+        let thing_variable_present = self.vertices.annotations.iter().filter_map(|(var, _)| var.as_variable())
+            .any(|var| variable_registry.get_variable_category(var).unwrap().is_category_thing());
+
+        let any_vertex_empty = self.vertices.annotations.iter().any(|(_, types)| types.is_empty());
+        if any_vertex_empty && thing_variable_present {
             return Err(TypeInferenceError::DetectedUnsatisfiablePattern {});
         }
         self.nested_disjunctions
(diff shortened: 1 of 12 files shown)


compiler/annotation/match_inference.rs
```diff
@@ -254,16 +254,15 @@ impl TypeInferenceGraph<'_> {
...
-        );
-        if any_variable_empty && thing_variable_present {
+        let thing_variable_present = self
+            .vertices
+            .annotations
+            .iter()
+            .filter_map(|(var, _)| var.as_variable())
+            .any(|var| variable_registry.get_variable_category(var).unwrap().is_category_thing());
+
+        let any_vertex_empty = self.vertices.annotations.iter().any(|(_, types)| types.is_empty());
```
krishnangovindraj, line 264: Extend to include label vertices.

ir/pipeline/mod.rs
krishnangovindraj, line 1: Update function errors to use `source_span`

ir/translation/mod.rs
```diff
@@ -89,3 +90,19 @@ impl TranslationContext {
...
+        ) {
+            Some(translated) => Ok(translated),
+            None => Err(Box::new(RepresentationError::$error {
+                variable: $var.name().unwrap().to_owned(),
+                source_span: $var.span(),
+            })),
+        }
+    };
+}
+pub(super) use verify_variable_available;
```
krishnangovindraj, line 108: Moves Georgii's great macro from `writes.rs` so I can use it everywhere

krishnangovindraj (commented):
comments

ir/pattern/constraint.rs
```diff
@@ -335,16 +337,25 @@ impl<'cx, 'reg> ConstraintsBuilder<'cx, 'reg> {
         function_name: &str,
         source_span: Option<Span>,
     ) -> Result<&FunctionCallBinding<Variable>, Box<RepresentationError>> {
+        if let Some(variable) = assigned.iter().find(|var| self.context.is_variable_input(**var)) {
+            let variable = self
+                .context
+                .get_variable_name(*variable)
+                .cloned()
+                .unwrap_or_else(|| VariableRegistry::UNNAMED_VARIABLE_DISPLAY_NAME.to_string());
+            return Err(Box::new(RepresentationError::AssigningToInputVariable { variable, source_span }));
```
farost, line 346: Do we only check this twice? If more, I'd create a separate method to wrap the whole `iter().find().....`

ir/pattern/function_call.rs
```diff
@@ -18,13 +18,12 @@ use crate::{pattern::IrID, pipeline::function_signature::FunctionID};
...
     function_id: FunctionID,
-    // map call variable to index of argument
-    call_variable_mapping: BTreeMap<ID, usize>,
+    arguments: Vec<ID>,
 }
 
 impl<ID> FunctionCall<ID> {
-    pub fn new(function_id: FunctionID, call_variable_mapping: BTreeMap<ID, usize>) -> Self {
-        Self { function_id, call_variable_mapping }
+    pub fn new(function_id: FunctionID, arguments: Vec<ID>) -> Self {
```
farost, line 25: Why was the BTreeMap a problem if it had `usize` as values? Anyway, it's fine since the inputs are small

ir/translation/function.rs
```diff
@@ -30,6 +30,20 @@ use crate::{
     },
 };
 
+macro_rules! verify_variable_available {
```
farost, line 33: I feel like the naming is not perfect (I was probably the one introducing it). Is it really a `verify` macro since it gets and returns some kind of translated result? Maybe it could get somehow enhanced

ir/translation/mod.rs
farost: As you had to duplicate it, looks like not everywhere, even if you make it `pub`, right?

ir/translation/modifiers.rs
```diff
@@ -21,13 +21,7 @@ pub fn translate_select(
     let selected_variables = typeql_select
         .variables
         .iter()
-        .map(|typeql_var| match context.get_variable(typeql_var.name().unwrap()) {
```
farost, line 24: It's very satisfying.

farost (approved):
Approved with comments, everything looks reasonable!

krishnangovindraj (commented):
replies

PR #7412, opened 2025-03-29
Variable categories restrict applicable types throughout the query

krishnangovindraj:
This change causes unintuitive behaviour (which currently shows up only in loosely typed queries). It should guarantee safe casts during execution, but there's a bigger problem to be solved wrt annotations varying in across nested patterns.

Variable categories now restrict applicable types throughout the query. This can lead to unintuitive behaviour for loosely typed queries.
` $x has name $n; ` will return entities with names as well.
` $x has name $n; not { $x links ($y);` will not return entities.

The benefit is simpler code as we can guarantee all casts done during execution are safe.

## Motivation
Prevent bad casts during execution.

## Implementation
* TypeSeeder prunes VertexAnnotations which have mismatched categories.
* Introduces `VariableCategory::{Relation, RelationType, ObjectType}`

### DIFF
--- a/compiler/annotation/type_inference.rs
+++ b/compiler/annotation/type_inference.rs
@@ -53,7 +53,7 @@ pub mod tests {
     };
 
     use answer::{variable::Variable, Type};
-    use concept::type_::{entity_type::EntityType, relation_type::RelationType, role_type::RoleType};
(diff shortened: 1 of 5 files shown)


krishnangovindraj:
Abandon, as we all agree this is undesirable, and we should just go straight for the more complete solution, such as: https://github.com/typedb/typedb/pull/7420

PR #7415, opened 2025-03-31
Better type-inference errors by pre-checking constraints for empty type annotations

krishnangovindraj:
Add a pre-check to see if type-seeding could not annotate some edge. Since such an error is localised to a pattern rather than an edge, it lets us have more descriptive errors. Such trivial cases are likely to account for most type-inference errors.
Sample:
```
mydb::read> match $c isa cat; $n isa dog-name; $c has $n;
           
INF11 
[INF11] Type-inference was unable to find compatible types for the pair of variables 'c' & 'n' across the constraint '$0 has $1'. Types were:
- c: [cat]
- n: [dog-name]
Caused: [QUA1] Type inference error while compiling query annotations.
Caused: [QEX8] Error analysing query.
Near 1:39
-----
--> match $c isa cat; $n isa dog-name; $c has $n;
                                          ^
-----
```

Since type-seeding is non-deterministic when annotating variables without explicit labels, the edge flagged may vary.
```
INF11 
[INF11] Type-inference was unable to find compatible types for the pair of variables 'n' & 'dog-name' across the constraint '$1 isa dog-name'. Types were:
- n: [name, cat-name]
- dog-name: [dog-name]
Caused: [QUA1] Type inference error while compiling query annotations.
Caused: [QEX8] Error analysing query.
Near 1:23
-----
--> match $c isa cat, has dog-name $n;
                          ^
-----
```

In cases where every edge (in isolation) can be annotated, the existing & less-descriptive "unsatisfiable pattern" error will still be returned. 

## Motivation
Better error messages for trivial cases of type-inference errors.

## Implementation
Given a set of valid vertex annotations for the left & right vertices, `TypeSeeder::seed_edges` annotates the edge with valid combinations of `left_right` & `right_left`. When there are no valid combinations, these sets are empty. 
A simple pre-check can flag such edges and return a descriptive error.

### DIFF
--- a/compiler/annotation/mod.rs
+++ b/compiler/annotation/mod.rs
@@ -205,8 +205,7 @@ typedb_error!(
         ),
         DetectedUnsatisfiableEdge(
             11,
-            "Type-inference was unable to find compatible types for the pair of variables '{left_variable}' & '{right_variable}' across the constraint '{constraint}'. Types were:\n- {left_variable}: [{left_types}]\n- {right_variable}: [{right_types}]",
-            constraint: ir::pattern::constraint::Constraint<Variable>,
+            "Type-inference was unable to find compatible types for the pair of variables '{left_variable}' & '{right_variable}' across a constraint. Types were:\n- {left_variable}: [{left_types}]\n- {right_variable}: [{right_types}]",
             left_variable: String,
             right_variable: String,
             left_types: String,
--- a/compiler/annotation/mod.rs
+++ b/compiler/annotation/mod.rs
@@ -205,7 +205,7 @@ typedb_error!(
         ),
         DetectedUnsatisfiableEdge(
             11,
-            "Type-inference was unable to find compatible types for the pair of variables '{left_variable}' & '{right_variable}' across the constraint. Types were:\n- {left_variable}: [{left_types}]\n- {right_variable}: [{right_types}]",
+            "Type-inference was unable to find compatible types for the pair of variables '{left_variable}' & '{right_variable}' across the constraint '{constraint}'. Types were:\n- {left_variable}: [{left_types}]\n- {right_variable}: [{right_types}]",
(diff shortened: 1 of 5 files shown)


compiler/annotation/mod.rs
```diff
@@ -203,6 +203,16 @@ typedb_error!(
             variable: Variable,
             source_span: Option<Span>,
         ),
+        DetectedUnsatisfiableEdge(
```
flyingsilverfin, line 206: Quick questions
- does this work with links with player/relation/role player 3 way component edge? How does it look?
- I'm not a huge fan of the user seeing '$0 has $1` like you did in the PR description. I would probably just delete this part and rely on the pointer into the query that indicates where it failed to be honest! Less redundancy.

flyingsilverfin (changes requested):
Would like to know one thing before approving!

compiler/annotation/mod.rs
krishnangovindraj: They're all binary edges in TypeInferenceGraph.
```
INF11 
[INF11] Type-inference was unable to find compatible types for the pair of variables 'p' & '_anonymous' across a constraint. Types were:
- p: [person]
- _anonymous: [friendship:friend]
Caused: [QUA1] Type inference error while compiling query annotations.
Caused: [QEX8] Error analysing query.
Near 1:48
-----
--> match $p isa person; $f isa friendship, links (friend: $p);
                                                   ^
-----
```

krishnangovindraj: Actually, I think I can fix that `_anonymous` to be a label if Links can be refactored to take a Label as role_type vertex

flyingsilverfin: ok _anonymous is better than $0, thanks

flyingsilverfin (approved):

PR #7418, opened 2025-03-31
Compile only the functions referenced in the query

krishnangovindraj:
Updates the compiler to only compile those functions referenced by the query.

## Motivation
Save effort.

## Implementation
Recursively collects all `FunctionID`s from function calls in:
1. Query pipeline, 
2. Query Fetch,
3. Any function referenced by the query.
4. Any function referenced in 3 or 4.

Note, that we don't compile unused preamble functions either. So any errors that would occur at compile time are not raised till the function is used.

### DIFF
--- a/compiler/executable/pipeline.rs
+++ b/compiler/executable/pipeline.rs
@@ -12,7 +12,10 @@ use std::{
 
 use answer::variable::Variable;
 use concept::thing::statistics::Statistics;
+use ir::pattern::conjunction::Conjunction;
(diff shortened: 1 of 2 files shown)


compiler/executable/pipeline.rs
```diff
@@ -361,3 +380,142 @@ fn compile_stage(
         }
     }
 }
+
```
krishnangovindraj, line 383: Should recursively collect.

flyingsilverfin (approved):

PR #7422, opened 2025-04-02
Release Windows in published mode and introduce declared features for Cargo sync

farost:
Fix a bug when Windows release binaries are published in the development mode.

## Motivation
Windows release binaries were running in the development mode.

## Implementation
Windows binaries are prepared using `cargo build` instead of `bazel build` due to Windows' path length limitations, which get exceeded easily with Bazel's generated directory names. This way, we need to use the server's crate `published` feature in cargo. However, we used to expose this feature only through Bazel, and the `toml` files did not contain any information about this feature. 

To fix this, a new tag was introduced to the [Cargo sync tool](https://github.com/typedb/typedb-dependencies/pull/573), which allows declaring features without setting their values for the crate or its users (the binary in our case). Other ways of propagating features through `cargo_features` would make the binary set the server's feature automatically.  

This fix has been tested on snapshots (on the commit before [this](https://github.com/typedb/typedb/commit/f6bc2e14a39985192b3ed1484718619ccfdf53f2)):

image


.circleci/windows/deploy_release.bat
```diff
@@ -8,12 +8,12 @@ REM by Chocolatey in prepare.bat is accessible
...
-cargo build --profile=release
+cargo build --profile=release --features server/published
 copy target\release\typedb_server_bin.exe  .\
 git apply .circleci\windows\git.patch
 
 SET DEPLOY_ARTIFACT_USERNAME=%REPO_TYPEDB_USERNAME%
 SET DEPLOY_ARTIFACT_PASSWORD=%REPO_TYPEDB_PASSWORD%
 set /p VER=<VERSION
-bazel --windows_enable_symlinks run --define version=%VER%  --enable_runfiles //:deploy-typedb-server -- release
+bazel --windows_enable_symlinks run --define version=%VER% --//server:mode=published --enable_runfiles //:deploy-typedb-server -- release
```
farost, line 18: Left just in case. What if something else will depend on this flag in the future, and we forget.

dependencies/typedb/repositories.bzl
```diff
@@ -12,10 +12,11 @@ def typedb_bazel_distribution():
     )
 
 def typedb_dependencies():
+    # TODO: Return typedb
     git_repository(
         name = "typedb_dependencies",
-        remote = "https://github.com/typedb/typedb-dependencies",
-        commit = "5c2d4dcc25493be9d35356075372229a6d2ba2a9",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_dependencies
+        remote = "https://github.com/farost/typedb-dependencies",
+        commit = "bb93bf52852526bd98c2a1771408f92297ef660f",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_dependencies
```
farost, line 19: https://github.com/typedb/typedb-dependencies/pull/573

ir/Cargo.toml
```diff
@@ -41,13 +41,13 @@ features = {}
 		features = []
 		default-features = false
 
-	[dependencies.durability]
-		path = "../durability"
+	[dependencies.concept]
```
farost, line 44: It's funny, I don't know why it happened

dmitrii-ubskii: This is why we can't check cargo sync in CI at the moment.

server/BUILD
```diff
@@ -81,6 +81,9 @@ rust_library(
         ":development_mode": [],
         ":published_mode": ["published"],
     }),
+    tags = [
+        "declared-features=published",
+    ]
```
dmitrii-ubskii, line 86: Not super happy with `declared-features` as a tag name, but I can't come up with anything better right now. I can see how `features=published` here would be confusing.

dmitrii-ubskii (approved):

PR #7423, opened 2025-04-03
Fix disjoint variable validation logic

dmitrii-ubskii:

Fix the condition for the disjoint variable reuse detection (`Variable 'var' is re-used across different branches of the query`).

## Implementation

Previous (incorrect) logic:
A variable is used disjointly if it is referenced by a disjunction, scoped to the containing conjunction, and appears in more than one branch of said disjunction (but not all branches).

This fails if: 
1. the disjunction is referencing a variable produced by a different pattern (false positive):
```php
match { $e isa person; } or { $e isa company; } or { $a isa name; }; $e has $a;
```
2. or if two or more disjunctions are referencing the variable without producing it (false negative):
```php
match { $e isa person; } or { $t sub $t; }; { $e isa company; } or { $u sub $u; };
```

Corrected logic:
A variable is used disjointly if it is referenced (not produced or required) by the conjunction it is scoped to.

### DIFF
--- a/ir/pattern/conjunction.rs
+++ b/ir/pattern/conjunction.rs
@@ -100,18 +100,18 @@ impl Conjunction {
     }
 
     fn producible_variables(&self, block_context: &BlockContext) -> impl Iterator<Item = Variable> + '_ {
-        self.variable_dependency(block_context).into_iter().filter_map(|(v, mode)| mode.is_producing().then_some(v))
+        self.variable_dependency(block_context).into_iter().filter_map(|(v, dep)| dep.is_producing().then_some(v))
     }
 
     pub fn required_inputs(&self, block_context: &BlockContext) -> impl Iterator<Item = Variable> + '_ {
-        self.variable_dependency(block_context).into_iter().filter_map(|(v, mode)| mode.is_required().then_some(v))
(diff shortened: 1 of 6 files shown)


ir/pattern/conjunction.rs
```diff
@@ -124,13 +124,15 @@ impl Conjunction {
     }
 
     pub(crate) fn find_disjoint(&self, block_context: &BlockContext) -> ControlFlow<(Variable, Option<Span>)> {
```
flyingsilverfin, line 126: What is the API doing? It returns a single Continue<(Variable, Span)>? isn't it possible to have more than 1 disjoint variable?

flyingsilverfin (approved):
Left one question, but looks right wrt to your description of what you wanted to do!
Don't forget to do bdd before merge

ir/pattern/conjunction.rs
dmitrii-ubskii: The `ControlFlow<B>` `B` generic is the `Break` type. It returns `Break` with the first found pair, or `Continue` if nothing found.
We could return all disjoint variables here too, though we can only report one to the user anyway.

PR #7424, opened 2025-04-06
Implement schema retrieval

flyingsilverfin:

We implement the (pre-existing) endpoint to retrieve the schema in two variations. These used to return an 'unimplemented' error.

- get the define query that specificies the Types (on the driver, this is available via a retrieved database's `database.type_schema()`)
- get the define query that specifies the Types and the Functions (on the driver, this is available via a retrieved database's `database.schema()`)

This API is useful to export and save an existing database schema, and also in particular useful for Agentic systems which use the schema definition to generate read or write queries for a database.

## Implementation

We implement the schema retrieval endpoint for the database service, down to the Database and the TypeManager. This required a new trait, `TypeQLSyntax`, which is implemented by each `Kind`, and various other internal structs that must be converted into TypeQL syntax on export.

Each part of the exported schema (types, capabilities, and annotations) is written sorted, so it is deterministically produced and tested.


tool/rust/sync.sh
farost, line 1: Nice

server/service/typedb_service.rs
```diff
@@ -177,10 +184,26 @@ impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
...
+        {
+            None => Err(ServiceError::DatabaseDoesNotExist { name: message.name.clone() }
                 .into_error_message()
-                .into_status())
+                .into_status()),
+            Some(database) => {
+                let transaction = TransactionRead::open(database, TransactionOptions::default()).map_err(|err| {
+                    ServiceError::FailedToOpenPrerequisiteTransaction {}.into_error_message().into_status()
+                })?;
+                let mut define_syntax = String::from("define\n");
```
farost, line 198: So the result for an empty database will just be 
```
define

```
? Don't we want to return an empty string instead?

farost (approved):
Looks good, left a couple of comments

server/service/typedb_service.rs
```diff
@@ -189,10 +212,23 @@ impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
         request: Request<typedb_protocol::database::type_schema::Req>,
     ) -> Result<Response<typedb_protocol::database::type_schema::Res>, Status> {
         let message = request.into_inner();
-        run_with_diagnostics(&self.diagnostics_manager, Some(&message.name), ActionKind::DatabaseTypeSchema, || {
-            Err(ServiceError::Unimplemented { description: "Database schema (types only) retrieval.".to_string() }
+        run_with_diagnostics(&self.diagnostics_manager, Some(&message.name), ActionKind::DatabaseSchema, || match self
```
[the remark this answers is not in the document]
flyingsilverfin: Thanks!

server/service/typedb_service.rs
flyingsilverfin: yess good point, thanks!

concept/type_/relation_type.rs
```diff
@@ -552,6 +564,50 @@ impl RelationType {
...
+
+    fn relates_syntax(
+        &self,
+        f: &mut impl std::fmt::Write,
+        snapshot: &impl ReadableSnapshot,
+        type_manager: &TypeManager,
+    ) -> Result<(), Box<ConceptReadError>> {
+        let mut declared_relates = HashSet::new();
+        let mut super_roles = HashSet::new();
+        for relates in self.get_relates_declared(snapshot, type_manager)?.iter() {
```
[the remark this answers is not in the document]
flyingsilverfin: I'm not quite following, is there a better way to do this :)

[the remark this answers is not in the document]
farost: It's not a change request, just a comment. When you call `get_relates_declared`, you also get specialized relates, but it looks like you filter them out. 
You could go in a reversed order: get `get_relates_root`, then add specialization through `relates.role().get_supertype()`
It's fine already

encoding/value/decimal_value.rs
```diff
@@ -247,16 +247,24 @@ impl fmt::Display for Decimal {
         if self.fractional == 0 {
             write!(f, "{}.0dec", self.integer_part())?;
         } else {
+            let (is_negative, int, frac) = if self.integer_part() < 0 {
+                let frac = FRACTIONAL_PART_DENOMINATOR - self.fractional_part();
+                (true, (self.integer_part() + 1).abs(), frac)
+            } else {
+                (false, self.integer_part(), self.fractional_part())
+            };
```
flyingsilverfin, line 255: Bug fix in representing -0.1 in dec to string

encoding/value/value.rs
```diff
@@ -458,7 +458,13 @@ impl fmt::Display for Value<'_> {
         match self {
             Value::Boolean(bool) => write!(f, "{bool}"),
             Value::Integer(integer) => write!(f, "{integer}"),
-            Value::Double(double) => write!(f, "{double}"),
+            Value::Double(double) => {
+                if double.fract() == 0.0 {
+                    write!(f, "{double}.0")
+                } else {
+                    write!(f, "{double}")
```
flyingsilverfin, line 465: Our doubles in typeql always must contain a decimal value to be re-parse-able. Technically we should probably implement `TypeQLSyntax` instead of the default `Display` being TypeQL compatible, but I think this is fine (we already were doing this a bit with the Decimal display implementation anyway, which included the required `dec` suffix!)

.factory/automation.yml
```diff
@@ -156,6 +156,13 @@ build:
         bazel test //tests/behaviour/query/functions:test_basic --test_output=streamed
         bazel test //tests/behaviour/query/functions:test_recursion --test_output=streamed
 
+    test-assembly:
```
flyingsilverfin, line 159: added back

.factory/automation.yml
```diff
@@ -168,17 +175,17 @@ release:
...
-#    validate-release-notes:
-#      image: typedb-ubuntu-22.04
-#      command: |
-#        export NOTES_VALIDATE_TOKEN=$REPO_GITHUB_TOKEN
-#        bazel run @typedb_dependencies//tool/release/notes:validate --test_output=streamed -- $FACTORY_OWNER $FACTORY_REPO RELEASE_NOTES_LATEST.md
+    validate-release-notes:
+      image: typedb-ubuntu-22.04
+      command: |
+        export NOTES_VALIDATE_TOKEN=$REPO_GITHUB_TOKEN
+        bazel run @typedb_dependencies//tool/release/notes:validate --test_output=streamed -- $FACTORY_OWNER $FACTORY_REPO RELEASE_NOTES_LATEST.md
```
flyingsilverfin, line 188: Re-enabled!

PR #7428, opened 2025-04-10
Add branch provenance  & query structure to HTTP response

krishnangovindraj:
Each answer to a query now includes a `provenance` field, telling us which branches were taken to derive the answer. 
The query response also includes a `queryStructure` field which describes the constraints in the query in terms of "edges" corresponding to the atomic TypeDB constraints. Conceptually, An edge is: 
```rust
struct QueryStructureEdge {
    type_: StructureEdgeType,
    from: StructureVertex,
    to: StructureVertex,
}
```
E.g. the statement `$x has $y`  would be 
```rust
QueryStructureEdge {
  type_: StructureEdgeType::Has,
  from: StructureVertex::Variable("x"),
  to: StructureVertex::Variable("y"),
}
```
By substituting the variables in each row back into the structure, we can reconstruct the "Answer graph"

**Note:** We use a 8-bytes per answer to track the branch-provenance. If the number of disjunction branches in a query exceeds 63, the query-structure will not be returned as the branch-provenance is no longer accurate. 

## Motivation
Support visualisation & other query-aware computation on clients

## Implementation
* Encode query structure in conceptRows response.
* Include provenance in each row of the answer.


server/service/http/transaction_service.rs
```diff
@@ -1132,7 +1132,7 @@ impl TransactionService {
             }
             respond_else_return_break!(
                 responder,
-                TransactionServiceResponse::Query(QueryAnswer::ResRows((QueryType::Read, result, warning)))
+                TransactionServiceResponse::Query(QueryAnswer::ResRows((QueryType::Read, result, encoded_query_structure, warning)))
```
farost, line 1135: We should probably encode just when we encode the `result`, and return only the `pipeline.query_structure()` here. It's cleaner.

server/service/http/message/query/query_structure.rs
```diff
@@ -0,0 +1,227 @@
...
+        Vertex::Variable(variable) => {
+            let name = query_structure.get_variable_name(variable).unwrap_or_else(|| variable.to_string());
+            if query_structure.available_variables.contains(variable) {
+                EncodedQueryStructureVertex::Variable { variable: name }
+            } else {
+                EncodedQueryStructureVertex::UnavailableVariable { variable: name }
+            }
+        }
+        Vertex::Label(label) => {
+            // TODO: Encode as in rows
```
farost, line 197: When are you going to cover these TODOs?

server/service/http/message/query/query_structure.rs
```diff
@@ -0,0 +1,227 @@
...
+
+        Constraint::IndexedRelation(indexed) => {
+            let role_type_1 = query_structure_role_type_as_vertex(query_structure, indexed.role_type_1());
+            let role_type_2 = query_structure_role_type_as_vertex(query_structure, indexed.role_type_2());
+            push_edge!(edges, query_structure, indexed.relation(), indexed.player_1(), Links(role_type_1));
+            push_edge!(edges, query_structure, indexed.relation(), indexed.player_2(), Links(role_type_2));
+        }
+        Constraint::ExpressionBinding(expr) => {
+            let expr_vertex =
+                { EncodedQueryStructureVertex::Expression { repr: format!("Expression#{index}").to_owned() } }; // TODO
```
farost, line 138: TODO?

server/service/http/message/query/mod.rs
```diff
@@ -64,36 +69,38 @@ pub struct QueryAnswerResponse {
...
+    QueryAnswerResponse { query_type, answer_type: AnswerType::ConceptRows, answers: Some(rows), query_structure: query_structure, warning }
 }
 
 pub(crate) fn encode_query_documents_answer(
     query_type: QueryType,
     documents: Vec<serde_json::Value>,
     warning: Option<String>,
 ) -> QueryAnswerResponse {
-    QueryAnswerResponse { query_type, answer_type: AnswerType::ConceptDocuments, answers: Some(documents), warning }
+    QueryAnswerResponse { query_type, answer_type: AnswerType::ConceptDocuments, answers: Some(documents),query_structure: None, warning }
```
farost, line 94: Will need to run `rustfmt`

server/server.rs
```diff
@@ -23,6 +23,8 @@ use resource::constants::server::{
 };
 use system::initialise_system_database;
 use tokio::net::lookup_host;
+use tonic::transport::{Certificate, Identity, ServerTlsConfig};
+use tower::ServiceExt;
```
flyingsilverfin, line 27: unneeded?

server/service/http/message/query/mod.rs
flyingsilverfin, line 94: fmt

server/service/http/message/query/mod.rs
```diff
@@ -64,36 +69,38 @@ pub struct QueryAnswerResponse {
...
 }
 
 impl IntoResponse for QueryAnswer {
     fn into_response(self) -> Response {
         let code = self.status_code();
         let body = match self {
             QueryAnswer::ResOk(query_type) => JsonBody(encode_query_ok_answer(query_type)),
-            QueryAnswer::ResRows((query_type, rows, warning)) => {
-                JsonBody(encode_query_rows_answer(query_type, rows, warning.map(|warning| warning.to_string())))
+            QueryAnswer::ResRows((query_type, rows, query_structure, warning)) => {
```
flyingsilverfin, line 102: Ou i forgot we had warnings - what's this @farost ?

server/service/http/message/query/query_structure.rs
```diff
@@ -0,0 +1,227 @@
...
+    from: EncodedQueryStructureVertex,
+    to: EncodedQueryStructureVertex,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase", tag = "kind", content = "value")]
+pub enum EncodedQueryStructureVertex {
+    Variable { variable: String },
+    Label { kind: String, label: String },
+    Value { value_type: String, value: String }, // TODO: Try to re-use the standard encoding for values in rows.
```
flyingsilverfin, line 61: Indeed would be good to re-use our existing value encoding IMO

server/service/http/message/query/query_structure.rs
```diff
@@ -0,0 +1,227 @@
...
+            });
+            function_call.required_ids().for_each(|variable| {
+                let argument = query_structure_vertex(query_structure, &Vertex::Variable(variable));
+                let arg_name = query_structure.get_variable_name(&variable).unwrap_or_else(|| variable.to_string());
+                let edge_type = EncodedQueryStructureEdgeType::Argument(arg_name);
+                edges.push(EncodedQueryStructureEdge { r#type: edge_type, from: argument, to: func_vertex.clone() });
+            });
+        }
+        | Constraint::Comparison(_) => {
+            // unimplemented_feature!(GraphViz);
```
flyingsilverfin, line 171: We haven't decided how to represent this basically right?

krishnangovindraj: Yeah. Or we can just not.

flyingsilverfin: happy to leave this as not for the time being for sure

server/service/http/message/query/mod.rs
farost: If a read query is stopped because of the query answers limit count, the answers are returned, but they are marked with "it's not all, you've hit the limit"

server/service/http/transaction_service.rs
krishnangovindraj: It's not really encoded, just translated. I'll update naming accordingly.

krishnangovindraj: Actually, the rows are already encoded here, so why not encode the query_structure here as well?

farost: Yeah, I forgot... It's fine

server/service/http/message/query/mod.rs
```diff
@@ -64,36 +69,38 @@ pub struct QueryAnswerResponse {
...
+                JsonBody(encode_query_rows_answer(query_type, rows, query_structure, warning.map(|warning| warning.to_string())))
```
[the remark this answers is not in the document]
krishnangovindraj: Does this still hold, given I've renamed `EncodedQueryStructure` to be `QueryStructureResponse`

compiler/query_structure.rs
```diff
@@ -131,3 +109,36 @@ fn extract_query_structure_from_branch(
...
+        match (vertex.as_label(), type_.iter().exactly_one()) {
+            (Some(label), Ok(type_)) => Some((label.clone(), type_.clone())),
+            _ => None,
+        }
+    }));
+}
+
+fn extend_function_text_from(
+    conjunction: &Conjunction,
+    function_texts: &mut HashMap<Constraint<Variable>, String>,
```
flyingsilverfin, line 127: Can we specify these are the `function_calls_syntax` perhaps? Im writing the schema retrieval code and been using `syntax` for the TypeQL representations. It won't help but you could use `call_syntax` instead here or?

server/service/http/message/query/query_structure.rs
flyingsilverfin, line 1: Thie one's for Georgii!

server/service/http/message/query/row.rs
```diff
@@ -10,14 +10,22 @@ use answer::variable_value::VariableValue;
...
 use storage::snapshot::ReadableSnapshot;
 
 use crate::service::http::message::query::concept::{encode_thing_concept, encode_type_concept, encode_value};
 
-pub fn encode_row(
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct EncodedRow<'a> {
+    data: HashMap<&'a str, serde_json::Value>,
+    provenance: u64,
```
flyingsilverfin, line 23: I wonder if we should encode this as a set of Integers that are indexes into the Branches in the QueryStructure - right now we're coupling our server-side compact encoding of Provenance onto the client-side, which is not ideal IMO.

The cost is going to be some more space... but i think this might be ok to avoid this coupling.

flyingsilverfin:
Please write a more roboust and useful PR!

flyingsilverfin:
Also, what will happen to the executor Row provenances when you have more than 64 branches? Will that panic?

krishnangovindraj:
Setting a branch_id more than 63 does nothing: https://github.com/typedb/typedb/pull/7402/files#r2049418258

server/service/http/message/query/row.rs
```diff
@@ -29,9 +37,9 @@ pub fn encode_row(
         let variable_value = row.get(*position);
         let row_entry =
             encode_row_entry(variable_value, snapshot, type_manager, thing_manager, include_instance_types)?;
-        encoded_row.insert(variable, row_entry);
+        encoded_row.insert(variable.as_str(), row_entry);
     }
-    Ok(json!(encoded_row))
+    Ok(json!(EncodedRow { data: encoded_row, provenance: row.provenance().0.to_le_bytes() }))
```
krishnangovindraj, line 42: Now encoded as le_bytes()

krishnangovindraj (commented):
Setting a branch_id more than 63 does nothing: https://github.com/typedb/typedb/pull/7402/files#r2049418258

compiler/query_structure.rs
```diff
@@ -131,3 +109,36 @@ fn extract_query_structure_from_branch(
...
+) {
+    resolved_labels.extend(vertex_annotations.filter_map(|(vertex, type_)| {
+        match (vertex.as_label(), type_.iter().exactly_one()) {
+            (Some(label), Ok(type_)) => Some((label.clone(), type_.clone())),
+            _ => None,
+        }
+    }));
+}
+
+fn extend_function_text_from(
```
flyingsilverfin, line 125: When you do renames just also do a search for the replaced term - the completeness of the rename matters hygene! You've got 'text' in a couple other places in this file (the intent was to avoid introducing two new terms - text and syntax, not specifically about this instance!)

server/service/http/message/query/row.rs
flyingsilverfin: Sick - I think you might as well throw this into `Provenance::to_array` right?

flyingsilverfin (approved):
Approved w 2 comments!

server/service/http/message/query/row.rs
flyingsilverfin: Hmm really what you want is a 64-bit array right? I'm actually not convinced the 8x8 bit array is much better than 1 u64... what if we call the network field `provenance_bit_array: u64` instead?

krishnangovindraj: We could, but that's pretty verbose. We could also encode it as a hex-string.

flyingsilverfin: I'm happy with calling it (only when we encode it into the network) as `provenance_bit_array` and not change the rest of the server :)

PR #7429, opened 2025-04-11
Fix crashing race condition after write query

### LINKED ISSUE #7427: Server crashes on `Arc::into_inner(snapshot).unwrap()`
## Description
Snapshot sharing in blocking transaction service's calls is not controlled enough, leading to multiple potential crashes and at least one reproducible crash.

I ran a number of big write queries, and my server failed on:
```
thread 'tokio-runtime-worker' panicked at server/service/transaction_service.rs:975:47:
```
in `spawn_blocking_execute_write_query`:
```
let (snapshot, result) = Self::execute_write_query_in(
    Arc::into_inner(snapshot).unwrap(),
    &type_manager,
    thing_manager.clone(),
    &function_manager,
    &query_manager,
    &pipeline,
    &source_query,
    interrupt,
);
```

We probably don't clean up other owners of `Arc<Snapshot>` clones correctly and in time.

## Environment

1. TypeDB distribution: Core
2. TypeDB version: 3.1.0
3. Environment: Mac
4. Client and version: Rust driver 3.1.0
5. Other details:

## Reproducible Steps

1. Set up
Boot up a 3.0.1 server.
Open `typedb-driver` and modify `rust/tests/integration/example.rs`:
```

// EXAMPLE START MARKER
use futures::{StreamExt, TryStreamExt};
// EXAMPLE END MARKER
use serial_test::serial;
// EXAMPLE START MARKER
use typedb_driver::{
    answer::{
        concept_document::{Leaf, Node},
        ConceptRow, QueryAnswer,
    },
    concept::{Concept, ValueType},
    Credentials, DriverOptions, Error, TransactionType, TypeDBDriver,
};

// EXAMPLE END MARKER

async fn cleanup() {
    let driver = TypeDBDriver::new(
        TypeDBDriver::DEFAULT_ADDRESS,
        Credentials::new("admin", "password"),
        DriverOptions::new(false, None).unwrap(),
    )
        .await
        .unwrap();
    if driver.databases().contains("typedb").await.unwrap() {
        driver.databases().get("typedb").await.unwrap().delete().await.unwrap();
    }
}

#[test]
#[serial]
// EXAMPLE START MARKER
fn example() {
    async_std::task::block_on(async {
        // EXAMPLE END MARKER
        cleanup().await;
        // EXAMPLE START MARKER
        // Open a driver connection. Specify your parameters if needed
        let driver = TypeDBDriver::new(
            TypeDBDriver::DEFAULT_ADDRESS,
            Credentials::new("admin", "password"),
            DriverOptions::new(false, None).unwrap(),
        )
            .await
            .unwrap();

        // Create a database
        driver.databases().create("typedb").await.unwrap();
        let database = driver.databases().get("typedb").await.unwrap();
        assert_eq!(database.name(), "typedb");

        let transaction = driver.transaction(database.name(), TransactionType::Schema).await.unwrap();

        transaction.query("define entity person, owns name; attribute name, value string;").await.unwrap();
        transaction.query("insert $p isa person, has name 'John';").await.unwrap();
        for i in 0..20 {
            println!("Start! {i}");
            let query = format!("match $p isa person; insert $p2 isa person, has name 'John{i}';");
            let _: Vec<ConceptRow> = transaction.query(&query).await.unwrap().into_rows().map(|result| result.unwrap()).collect().await;
            println!("End!");
        }
        // for i in 0..15 {
        //     println!("Start! {i}");
        //     let _: Vec<ConceptRow> = transaction.query("match $p isa person, has name $n; $p2 isa person; not { $p2 is $p; };").await.unwrap()
        //         .into_rows().map(|result| result.unwrap()).collect().await;
        //     println!("End!");
        // }
        // EXAMPLE END MARKER
    });

    async_std::task::block_on(async {
        cleanup().await;
        // EXAMPLE START MARKER
        println!("More examples can be found in the API reference and the documentation.\nWelcome to TypeDB!");
    })
}
``` 

2. Execute
Run `bazel test //rust/tests/integration:test_example --test_output=streamed --test_arg=--nocapture`, potentially multiple times.
 
3. Unexpected result
The server crashes.

## Expected result
The server does not crash.

## Additional information

dmitrii-ubskii:

Resolve issue where a large write query (~100k keys) would sometimes cause the server to crash at the end.

## Implementation

Each write query expects exclusive ownership over the snapshot. We ensure that by checking and, if necessary, interrupting outgoing answer streams that read from the snapshot before handing the snapshot over to the write query executor.

This check created a race condition when the stream has ended, but the reference to the snapshot is not yet released. This PR adds an explicit check for the end of the worker task.

Resolves #7427.

### DIFF
--- a/server/service/transaction_service.rs
+++ b/server/service/transaction_service.rs
@@ -8,7 +8,7 @@ use std::{
     collections::{HashMap, VecDeque},
     ops::ControlFlow::{self, Break, Continue},
     sync::Arc,
-    time::Instant,
+    time::{Duration, Instant},
 };
 
 use compiler::VariablePosition;
@@ -1419,7 +1419,12 @@ impl TransactionService {
         let responder = self.responders.get_mut(&request_id);
         if let Some((worker_handle, stream_transmitter)) = responder {
             if stream_transmitter.check_finished_else_queue_continue().await {
+                const ALLOWED_CLEANUP_TIME: Duration = Duration::from_secs(60);
+                let start = Instant::now();
                 while !worker_handle.is_finished() {
+                    if start.elapsed() > ALLOWED_CLEANUP_TIME {
+                        panic!("Query stream {request_id:?} ended but has not responede in over {ALLOWED_CLEANUP_TIME:?}, aborting. (This is a bug!)");
+                    }
                     // The worker _must_ be wrapping up and freeing resources here. We check on it every tick
                     // as we can't proceed until it's done.
                     yield_now().await;
--- a/server/service/transaction_service.rs
+++ b/server/service/transaction_service.rs
@@ -46,14 +46,13 @@ use storage::{
     snapshot::{ReadableSnapshot, WritableSnapshot},
 };
 use tokio::{
-    select,
     sync::{
         broadcast,
         mpsc::{channel, Receiver, Sender},
         watch,
     },
     task::{spawn_blocking, JoinHandle},
-    time::sleep,
+    time::timeout,
 };
 use tokio_stream::StreamExt;
 use tonic::{Status, Streaming};
@@ -1422,11 +1421,8 @@ impl TransactionService {
         if let Some((worker_handle, stream_transmitter)) = responder {
             if stream_transmitter.check_finished_else_queue_continue().await {
                 const ALLOWED_CLEANUP_TIME: Duration = Duration::from_secs(60);
-                select! {
-                    _ = sleep(ALLOWED_CLEANUP_TIME) => {
-                        panic!("Query stream {request_id:?} ended but has not responede in over {ALLOWED_CLEANUP_TIME:?}, aborting. (This is a bug!)");
-                    }
-                    _ = worker_handle => (),
+                if timeout(ALLOWED_CLEANUP_TIME, worker_handle).await.is_err() {
+                    panic!("Query stream {request_id:?} ended but has not responded in over {ALLOWED_CLEANUP_TIME:?}, aborting. (This is a bug!)");
                 }
                 self.responders.remove(&request_id);
             }
--- a/server/service/transaction_service.rs
+++ b/server/service/transaction_service.rs
@@ -6,10 +6,7 @@
 
 use std::{
     collections::{HashMap, VecDeque},
-    ops::{
-        ControlFlow,
-        ControlFlow::{Break, Continue},
-    },
+    ops::ControlFlow::{self, Break, Continue},
     sync::Arc,
     time::Instant,
 };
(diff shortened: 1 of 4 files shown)


server/service/transaction_service.rs
```diff
@@ -1423,6 +1419,11 @@ impl TransactionService {
         let responder = self.responders.get_mut(&request_id);
         if let Some((worker_handle, stream_transmitter)) = responder {
             if stream_transmitter.check_finished_else_queue_continue().await {
+                while !worker_handle.is_finished() {
+                    // The worker _must_ be wrapping up and freeing resources here. We check on it every tick
+                    // as we can't proceed until it's done.
+                    yield_now().await;
+                }
```
dmitrii-ubskii, line 1426: The actual fix is this.
We enter this `if` when the worker has ended the stream, at which point it exits. However, the future is yet to be freed, and we cannot proceed until they are (e.g. we expect unique ownership over the `Arc<Snapshot>`).
`yield_now()` tells the tokio runtime to suspend this future and put it at the end of the queue. It's effectively an async spin loop. In testing I've not seen this loop take more than 2 ms to complete if it is even entered.

server/service/transaction_service.rs
```diff
@@ -1681,47 +1682,6 @@ impl QueryStreamTransmitter {
     }
 }
 
-trait ControlFlowExt<B, C> {
```
dmitrii-ubskii, line 1684: These are unused and easily implemented if needed.

server/service/transaction_service.rs
farost: Can we crash or return an error or report a sentry error `sentry::capture_message(message, sentry::Level::Error);` if it takes too many iterations and hangs? I'd prefer a crash instead of a server hanging indefinitely

server/service/transaction_service.rs
farost: It will get overridden in my branch soon, but whatever, yeah.

server/service/transaction_service.rs
dmitrii-ubskii: The problem with this is that nothing in the rest of the transaction service expects a responder to be able to close the transaction. 
I've put a big time limit on this with a sensible panic message.

farost: Good, thanks. Does it work well for a combination of read/write queries now? Should probably do, but just in case.

farost: I'd test it with multiple huge read queries running after inserting a huge amount of instances

farost (changes requested):
Will approve after hearing the results of tests with read queries!

server/service/transaction_service.rs
```diff
@@ -1423,7 +1420,10 @@ impl TransactionService {
         let responder = self.responders.get_mut(&request_id);
         if let Some((worker_handle, stream_transmitter)) = responder {
             if stream_transmitter.check_finished_else_queue_continue().await {
-                debug_assert!(worker_handle.is_finished());
+                const ALLOWED_CLEANUP_TIME: Duration = Duration::from_secs(60);
+                if timeout(ALLOWED_CLEANUP_TIME, worker_handle).await.is_err() {
+                    panic!("Query stream {request_id:?} ended but has not responded in over {ALLOWED_CLEANUP_TIME:?}, aborting. (This is a bug!)");
+                }
```
dmitrii-ubskii, line 1426: The better approach to waiting for the task to finish: just await with timeout.

farost (approved):
It was interesting to see multiple approaches. Tokio does really have some cool things implemented.

It looks good!

PR #7438, opened 2025-04-22
Add span to returned query-structure

krishnangovindraj:
Adds the span of each constraint to the returned query-structure.

## Motivation
Allow mapping the query-structure edges to the statements in the TypeQL query.

## Implementation
Add `span` to `QueryStructureEdgeResponse`

### DIFF
--- a/server/service/http/message/query/query_structure.rs
+++ b/server/service/http/message/query/query_structure.rs
@@ -83,12 +83,20 @@ pub enum QueryStructureEdgeTypeResponse {
     Argument(String),
 }
 
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct QueryStructureEdgeSpan {
+    begin: usize,
+    end: usize,
+}
+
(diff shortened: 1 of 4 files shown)


server/service/http/message/query/query_structure.rs
```diff
@@ -83,12 +83,20 @@ pub enum QueryStructureEdgeTypeResponse {
     Argument(String),
 }
 
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct QueryStructureEdgeSpan {
```
flyingsilverfin, line 88: Looks ok, though only thing i wonder is if the end is inclusive or exclusive

flyingsilverfin (approved):
Approved w 1 requeest

server/service/http/message/query/query_structure.rs
krishnangovindraj: inclusive, as in TypeQL. Which is quite confusing.

flyingsilverfin: ok

PR #7439, opened 2025-04-24
Enable GRPC transaction and query options

farost:
Enable options provided by the TypeDB GRPC protocol.

**Enable transaction options:**
* `transaction_timeout`: If set, specifies a timeout for killing transactions automatically, preventing memory leaks in unclosed transactions,
* `schema_lock_acquire_timeout`: If set, specifies how long the driver should wait if opening a transaction is blocked by an exclusive schema write lock. 

**Enable query options:**
* `include_instance_types`: If set, specifies if types should be included in instance structs returned in ConceptRow answers,
* `prefetch_size`: If set, specifies the number of extra query responses sent before the client side has to re-request more responses. Increasing this may increase performance for queries with a huge number of answers, as it can reduce the number of network round-trips at the cost of more resources on the server side.

## Motivation
Allow default transaction and query behaviors to be overridden by GRPC clients.

## Implementation
While the logic has already been propagated to the whole GRPC transaction service, there were desired refinements of the protocol structure's processing. Additionally, to ensure that the BDD tests updates are in sync with the server, the relevant Bazel references are updated.

### DIFF
--- a/server/service/grpc/transaction_service.rs
+++ b/server/service/grpc/transaction_service.rs
@@ -64,6 +64,7 @@ use crate::service::{
         diagnostics::run_with_diagnostics_async,
         document::encode_document,
         error::{IntoGrpcStatus, IntoProtocolErrorMessage, ProtocolError},
+        options::{query_options_from_proto, transaction_options_from_proto},
         response_builders::transaction::{
             query_initial_res_from_error, query_initial_res_from_query_res_ok,
             query_initial_res_ok_from_query_res_ok_ok, query_res_ok_concept_document_stream,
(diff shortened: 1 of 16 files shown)


server/service/grpc/transaction_service.rs
```diff
@@ -449,23 +450,9 @@ impl TransactionService {
...
-            if let Some(transaction_timeout_millis) = options.transaction_timeout_millis {
-                transaction_options.transaction_timeout_millis = transaction_timeout_millis
-            }
-
-            // service options
-            self.prefetch_size = options.prefetch_size.or(Some(DEFAULT_PREFETCH_SIZE));
-        }
-
+        let transaction_options = transaction_options_from_proto(open_req.options);
+        // self.prefetch_size = options.prefetch_size.or(Some(DEFAULT_PREFETCH_SIZE)); // TODO get from protocol?
```
flyingsilverfin, line 454: you can delete this and the todo tbh

flyingsilverfin (approved):
APproved w 1 comment

flyingsilverfin (approved):
APproved w 1 comment

server/service/grpc/transaction_service.rs
```diff
@@ -776,7 +759,14 @@ impl TransactionService {
         req_id: Uuid,
         query_req: typedb_protocol::query::Req,
     ) -> Result<ControlFlow<(), ()>, Status> {
-        let query_options = QueryOptions::default_grpc();
+        let query_options = query_options_from_proto(query_req.options);
+        if query_options.prefetch_size < 1 {
```
farost, line 763: In 2.x, we had this check:
```
if (prefetchSize < 1) throw TypeDBException.of(RPC_PREFETCH_SIZE_TOO_SMALL, prefetchSize);
```

For some reason, it was also called without checking the `prefetch` boolean flag. As 0s did not work without smart modifications here, either, I decided to restrict users from disabling `prefetch_size` completely.

flyingsilverfin (approved):

PR #7441, opened 2025-04-25
Combine multiple BDDs into single target to reduce build times

krishnangovindraj:


.factory/automation.yml
```diff
@@ -104,57 +104,42 @@ build:
       dependencies: [build]
       command: |
         bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
-        bazel test //tests/behaviour/connection/... --test_output=streamed
+        bazel test //tests/behaviour/connection:test_connection --test_output=streamed  --test_arg="--test-threads=1"
```
krishnangovindraj, line 107: We might actually be able to parallelise if the features use different databases.

.factory/automation.yml
```diff
@@ -104,57 +104,42 @@ build:
...
       dependencies: [build]
       command: |
         bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
-        bazel test //tests/behaviour/query/functions:test_definition --test_output=streamed
-        bazel test //tests/behaviour/query/functions:test_signature --test_output=streamed
-        bazel test //tests/behaviour/query/functions:test_structure --test_output=streamed
-        bazel test //tests/behaviour/query/functions:test_usage --test_output=streamed
-        bazel test //tests/behaviour/query/functions:test_basic --test_output=streamed
-        bazel test //tests/behaviour/query/functions:test_recursion --test_output=streamed
+        bazel test //tests/behaviour/query:test_query --test_output=streamed --test_arg="--test-threads=1" --test_arg="functions::"
```
flyingsilverfin, line 142: We're relying on the cache having built test_query once before here right? In which case, I think with cache on, our builds should be faster than what you tested with cache of?

Cargo.toml
```diff
@@ -123,64 +123,12 @@ features = {}
...
-
-[[test]]
-	path = "tests/behaviour/concept/migration/data_validation.rs"
-	name = "test_data_validation"
-
-[[test]]
-	path = "tests/behaviour/concept/migration/migration.rs"
-	name = "test_migration"
+	path = "tests/behaviour/concept/main.rs"
+	name = "test_concept"
```
flyingsilverfin, line 131: Good groupings!

tests/behaviour/debug/debug.rs
```diff
@@ -7,6 +7,11 @@
 use steps::Context;
 
 #[tokio::test]
-async fn test() {
+async fn test_a() {
+    assert!(Context::test("tests/behaviour/debug/debug.feature", true).await);
+}
+
+#[tokio::test]
+async fn test_z() {
```
flyingsilverfin, line 15: cleanup?

flyingsilverfin (approved):
Approved with a set of changes to make first

.factory/automation.yml
krishnangovindraj: Yes.

PR #7442, opened 2025-04-28
Execute deletes in two passes - first constraints, then concepts

krishnangovindraj:
If a delete stage which deletes both constraints and concepts, a certain row may delete a concept referenced by a connection in a subsequent row. We avoid this problem by first deleting constraints for every row in an initial pass, followed by concepts of every row in a second pass.

### DIFF
--- a/executor/pipeline/delete.rs
+++ b/executor/pipeline/delete.rs
@@ -9,6 +9,7 @@ use std::sync::Arc;
 use compiler::executable::delete::{executable::DeleteExecutable, instructions::ConnectionInstruction};
 use concept::thing::thing_manager::ThingManager;
 use ir::pipeline::ParameterRegistry;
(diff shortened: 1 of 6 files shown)


executor/pipeline/delete.rs
```diff
@@ -59,10 +59,31 @@ where
...
+                snapshot_mut,
+                &context.thing_manager,
+                &context.parameters,
+                &mut row,
+                &profile,
+            ) {
+                return Err((Box::new(PipelineExecutionError::WriteError { typedb_source }), context));
+            }
+
+            if index % 100 == 0 {
```
farost, line 76: Let's move this `100` to a const

farost (approved):
1 comment

flyingsilverfin (approved):

PR #7447, opened 2025-05-01
Add yaml configuration file & logging to file

krishnangovindraj:
A yaml file is now required as the source of configuration settings. 
For documentation of the exposed settings, check the file at `server/config.yml`

## Implementation
* Uses `serde_yaml2` to automatically deserialize the yaml into the config structure.
* `CLIArgs` now override the read config.
  - They used to have default values, now they're all optional with default None.
* Introduced a logging.directory config option and added rotating logging to file using `tracing-appender`.


server/parameters/cli.rs
```diff
@@ -11,37 +11,41 @@ use resource::constants::server::{
...
-    #[arg(long = "server.address", default_value_t = DEFAULT_ADDRESS.to_string())]
-    pub server_address: String,
+    #[arg(long = "server.address")]
+    pub server_address: Option<String>,
 
     /// Enable/disable HTTP endpoint
-    #[arg(long = "server.http.enable", default_value_t = true, action=ArgAction::Set)]
-    pub server_http_enable: bool,
+    #[arg(long = "server.http.enabled")]
+    pub server_http_enabled: Option<bool>,
```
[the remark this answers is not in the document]
krishnangovindraj: Remove `ArgAction::Set` because we now need a way to override a config file enable with a CLI disable.
In the past the default was `false` so we could add or omit the flag to set true or false.

dependencies/typedb/repositories.bzl
```diff
@@ -14,8 +14,8 @@ def typedb_bazel_distribution():
 def typedb_dependencies():
     git_repository(
         name = "typedb_dependencies",
-        remote = "https://github.com/typedb/typedb-dependencies",
-        commit = "b92f47300913cc8555ef18580eeaa5b1b1ecd2a1",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_dependencies
+        remote = "https://github.com/krishnangovindraj/dependencies",
+        commit = "b53291cbdc907880776d19f9d7430ab4b0ff4e7c",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_dependencies
```
krishnangovindraj, line 18: todo

server/typedb.yml
```diff
@@ -0,0 +1,19 @@
...
+    http_address: 0.0.0.0:8000
+
+    authentication:
+        token_expiration_seconds: 5000
+
+    encryption:
+        enabled: false
+        # cert: =
+        # cert_key: =
+        # root_ca: =
```
krishnangovindraj, line 13: Not a fan of these fields, but they're what the keys for the CLI were.

krishnangovindraj (commented):
self review

dmitrii-ubskii (approved):

server/parameters/config.rs
```diff
@@ -31,6 +40,46 @@ impl Config {
...
+    fn resolve_path_from_executable(path: &PathBuf) -> PathBuf {
+        let typedb_dir_or_current = std::env::current_exe()
+            .map(|path| path.parent().expect("Expected parent directory of: {path}").to_path_buf())
+            .unwrap_or(std::env::current_dir().expect("Expected access to the current directory"));
+        // if path is absolute, join will just return path
+        typedb_dir_or_current.join(path)
+    }
+
+    fn is_development_mode_default() -> bool {
+        false
```
flyingsilverfin, line 81: maybe i'd call this `default_is_development_mode`?

flyingsilverfin: I thought our default is development mode, and we set mode =release explicitly?

server/typedb.yml
flyingsilverfin: is there a way we can keep the 2.x model where all the keys exist, but are empty (then everyone can see all the possible options)

i think that's what we had right?

server/typedb.yml
```diff
@@ -0,0 +1,19 @@
+server:
```
flyingsilverfin, line 1: needs license no?

server/typedb.yml
```diff
@@ -0,0 +1,19 @@
...
+        token_expiration_seconds: 5000
+
+    encryption:
+        enabled: false
+        # cert: =
+        # cert_key: =
+        # root_ca: =
+
+storage:
+    data: "data"
```
flyingsilverfin, line 16: should we call this 'directory' too?

server/typedb.yml
```diff
@@ -0,0 +1,19 @@
...
+
+logging:
+    directory: "logs"
```
flyingsilverfin, line 19: What about diagnostics config?

flyingsilverfin (changes requested):

server/parameters/cli.rs
```diff
@@ -59,59 +63,84 @@ pub struct CLIArgs {
...
-    pub diagnostics_monitoring_port: u16,
+    #[arg(long = "diagnostics.monitoring.port")]
+    pub diagnostics_monitoring_port: Option<u16>,
 
     /// Enable development mode for testing setups. Note that running TypeDB in development mode
     /// may result in error reporting limitations (obstructing maintenance and support), additional
     /// logging, restricted functionalities, and reduced performance
     #[arg(long = "development-mode.enabled", hide = true)]
-    pub development_mode_enabled: bool,
+    pub development_mode_enabled: Option<bool>,
```
krishnangovindraj, line 90: (Anchor to link to)

server/parameters/config.rs
krishnangovindraj: It's the default value for `is_development_mode`. 
Also, no: https://github.com/typedb/typedb/pull/7447/files#r2071189384

server/typedb.yml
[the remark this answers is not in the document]
krishnangovindraj: Or are they self-explanatory?

server/typedb.yml
krishnangovindraj: `storage.directory` doesn't say much to me. You could be storing logs there

server/typedb.yml
krishnangovindraj: Can add them, can omit them.

BUILD
```diff
@@ -71,8 +71,11 @@ alias(
 # The directory structure for distribution
 pkg_files(
     name = "package-layout-server",
-    srcs = ["//:typedb_server_bin", "//binary:typedb"],
-    renames = {"//:typedb_server_bin" : "server/typedb_server_bin"},
+    srcs = ["//:typedb_server_bin", "//binary:typedb", "//server:typedb.yml"],
+    renames = {
+        "//:typedb_server_bin" : "server/typedb_server_bin",
+        "//server:typedb.yml" : "server/typedb.yml",
```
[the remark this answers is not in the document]
krishnangovindraj: I don't love the name `config.yml`.
It feels lazy. It's like calling our binary `server` instead of `typedb`.
I guess it'll always be namespaced into a folder like `/etc/typedb/config.yml`

server/parameters/cli.rs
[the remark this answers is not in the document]
farost: The naming was consistent. There was a difference between switches (define to enable) and flags (set a value while defining). 
Don't we have switched anymore?

BUILD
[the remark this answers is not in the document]
farost: If we don't plan to place any more configs under this directory, then `config.yml` is fine. However, if we plan to place logging and other configs here, as well, I'd make it `config/typedb.yml`.

server/parameters/config.rs
farost: Forced development mode doesn't depend on the config, so it's a different field. It's fine

server/parameters/cli.rs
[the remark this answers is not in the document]
krishnangovindraj: No. No define to enable anymore, because of the reason above.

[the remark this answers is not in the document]
farost: That's what I am talking about. We had enable**d** and enable. One for switches, one for flags. And I thought it was quite consistent.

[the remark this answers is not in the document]
krishnangovindraj: Ah. That is indeed consistent.

common/logger/logger.rs
```diff
@@ -28,14 +32,14 @@ pub fn initialise_logging_global() {
     ;
 
     // Create a file appender
-    // let file_appender = File::create("output.log")
-    //     .expect("Failed to create log file");
-
+    let logfile_path = logdir.join("typedb.log");
```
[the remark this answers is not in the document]
flyingsilverfin: i would make it depend on constants and put the constant in the constants.rs!

BUILD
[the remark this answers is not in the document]
flyingsilverfin: well by that logic @krishnangovindraj  we'll end up with with "typedb.exe", "typedb.yml", "typedb.log", ... all around the same place! Super confusing :D

but yes - we can call it `/server/config.yml`. Another version that we can do is `/server/config/typedb.yml`, then it's scoped. The main thing is that at some point we may have a logging config file right? Is there an XML for that @krishnangovindraj like in java ? where will that be and what's the naming convention?

server/typedb.yml
flyingsilverfin: in the ideal state, we'll have 3 paths:
1. database storage
2. WAL storage
3. Replication storage

i think we should move towards that structure cc @lolski ?

server/typedb.yml
krishnangovindraj: If you say so. the first thing I do in a config is disable all diagnostics

krishnangovindraj: The deserializer doesn't recognise that 
```
a:
  b:
     c: value
```
is the same as 
```
a.b.c: value
```

I've tried serde_query as well but that didn't work either.

krishnangovindraj: ^ TODO: 
We only want to support the nested (a:b:c) notation in config files. So I'll have to unflatten Diagnostics struct in core.

server/typedb.yml
krishnangovindraj: Will we have a WAL separate from the raft log?

BUILD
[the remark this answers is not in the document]
krishnangovindraj: Looks like there's this: https://crates.io/crates/tracing-logger-config 
But do we want to directly depend on this? Any change they make would change our log-file formats too.

server/typedb.yml
flyingsilverfin: Eventually, but let's push that to later:
```
storage:
  database-directory: <path>
  // future - replication-directory: <path>
```

server/config.yml
```diff
@@ -17,7 +17,7 @@ server:
         ca-certificate:
 
 storage:
-    data-directory: "data"
+    database-directory: "data"
```
flyingsilverfin, line 20: Hmm actually i think 'data' directory is better - aligns with users will think of when they see "data" (their own data), instead of what we consider data (anything we store)

flyingsilverfin (approved):
Don't forget to change the deps branch!

PR #7448, opened 2025-05-01
Don't fetch validation data until required

dmitrii-ubskii:

We delay reading the data necessary for cardinality constraint validation until we have determined that we need to perform any validation.

## Implementation

We also add commit profile logging, enabled by setting log level of `database` to `trace`.


concept/thing/thing_manager/validation/commit_time_validation.rs
```diff
@@ -81,6 +84,7 @@ macro_rules! validate_capability_cardinality_constraint {
                     continue;
                 }
 
+                let Ok(counts) = &*counts else { return Err(counts.clone().unwrap_err()) };
```
dmitrii-ubskii, line 87: `LazyCell::into_inner` would have been better, but it's nightly :\

flyingsilverfin (approved):

flyingsilverfin:
Gud

PR #7450, opened 2025-05-02
Add HTTP schema retrieval, root and version endpoints

farost:
TypeDB HTTP endpoint acquired:
* `GET /schema` and `GET /type-schema` methods for database schema retrieval. Similarly to GRPC, the `GET /schema` method returns both type and function schemas.
* `GET /version` method returning the server's version and distribution in JSON.
* redirection from `GET /` to `GET /version` with the `307` response code. 

## Implementation
Schema retrieval reuses the code written for GRPC. To generalize this approach, a new common `typedb_service` file has been introduced, similar to `transaction_service`.

Redirection is performed through `axum` wrappers.


.factory/automation.yml
```diff
@@ -148,13 +148,12 @@ build:
...
-#      dependencies: [build]
-#      command: |
-#        bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
-#        bazel test //tests/behaviour/service/... --test_output=streamed --jobs=1
+    test-behaviour-service:
+      image: typedb-ubuntu-22.04
+      dependencies: [build]
+      command: |
+        bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
+        bazel test //tests/behaviour/service/... --test_output=streamed --jobs=1
```
flyingsilverfin, line 156: is this for http service? That would be good to reflect from the naming here

farost: Well, it tests everything in the service package. But I can write `/service/http/...` for explicitness

flyingsilverfin: Do we have more than http under service? I spose we can keep it the way it is right now

flyingsilverfin (approved):

.factory/automation.yml
farost: No

farost: I'll leave it this way to test all the service-related things at once.

PR #7454, opened 2025-05-12
Prune branch ids when pruning away unsatisfiable disjunction branches

### LINKED ISSUE #7452: Schema disjunctions can lead to assertion failures
## Description
While roleplayers without existing relation types and role types in the schema, the server fails on assertion:
`assertion failed: branch_ids.len() == branches.len()`. 
**It does not affect the release version of TypeDB: a correct result is returned without this check**. 

Also, it works correctly if there are relation types and role types in the database.

We need to verify that removing this assertion is a correct solution, and that we don't miss any corner cases.

## Environment

1. TypeDB distribution: Core
2. TypeDB version: 3.2.0
3. Environment: Mac
4. Client and version: 3.2.0
5. Other details:

## Reproducible Steps

1. Set up
Create a new database
`database create a` (console)

Open a new transaction
`transaction schema a` (console)

2. Execute
```
define
entity person owns name;
attribute name value string;
```

3. Test/Query
Run:
```
match 
{ $t sub $other;  $role is $other; } or
{ $t owns $other;  $role is $other; } or
{ $t relates $role; $role is $other;} or
{ $t plays $role;  $role is $other; };
```

4. Unexpected result (in debug, e.g., local builds)
Server crashes with:
```
thread 'tokio-runtime-worker' panicked at executor/read/nested_pattern_executor.rs:34:9:
assertion failed: branch_ids.len() == branches.len()
```

## Expected result (like in a release version)
```
Finished read query validation and compilation...
Streaming rows...
   ------------
    $other | name
    $role  | name
    $t     | person
   ------------
    $other | person
    $role  | person
    $t     | person
   ------------
    $other | name
    $role  | name
    $t     | name
   ------------
Finished. Total answers: 3

```

krishnangovindraj:
This fixes a bug where the branch_ids go out of sync with the branches.

## Motivation
fixes #7452

### DIFF
--- a/ir/pattern/disjunction.rs
+++ b/ir/pattern/disjunction.rs
@@ -57,7 +57,11 @@ impl Disjunction {
     }
 
     pub fn optimise_away_unsatisfiable_branches(&mut self, unsatisfiable: Vec<ScopeId>) {
-        self.conjunctions.retain(|v| !unsatisfiable.contains(&v.scope_id()))
+        let unsatisfiable_branch_ids = self.conjunctions.iter().zip(self.branch_ids.iter())
+            .filter_map(|(conj, branch_id)| unsatisfiable.contains(&conj.scope_id()).then_some(*branch_id))
+            .collect::<Vec<_>>();
+        self.branch_ids.retain(|branch_id| !unsatisfiable_branch_ids.contains(branch_id));
+        self.conjunctions.retain(|conj| !unsatisfiable.contains(&conj.scope_id()))
     }
 
     pub fn required_inputs(&self, block_context: &BlockContext) -> impl Iterator<Item = Variable> + '_ {
--- a/server/service/http/message/query/query_structure.rs
+++ b/server/service/http/message/query/query_structure.rs
@@ -198,6 +198,18 @@ fn query_structure_edge(
             push_edge!(edges, context, span, relates.relation(), relates.role_type(), Relates)
         }
         Constraint::Plays(plays) => push_edge!(edges, context, span, plays.player(), plays.role_type(), Plays),
+        Constraint::IndexedRelation(indexed) => {
+            let role_type_1 = query_structure_role_type_as_vertex(context, indexed.role_type_1())?;
+            let role_type_2 = query_structure_role_type_as_vertex(context, indexed.role_type_2())?;
+            let span_1 = indexed
+                .source_span_1()
+                .map(|span| QueryStructureEdgeSpan { begin: span.begin_offset, end: span.end_offset });
+            let span_2 = indexed
+                .source_span_2()
+                .map(|span| QueryStructureEdgeSpan { begin: span.begin_offset, end: span.end_offset });
+            push_edge!(edges, context, span_1, indexed.relation(), indexed.player_1(), Links(role_type_1));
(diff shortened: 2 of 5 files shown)


ir/pattern/disjunction.rs
```diff
@@ -57,7 +57,11 @@ impl Disjunction {
     }
 
     pub fn optimise_away_unsatisfiable_branches(&mut self, unsatisfiable: Vec<ScopeId>) {
-        self.conjunctions.retain(|v| !unsatisfiable.contains(&v.scope_id()))
+        let unsatisfiable_branch_ids = self.conjunctions.iter().zip(self.branch_ids.iter())
+            .filter_map(|(conj, branch_id)| unsatisfiable.contains(&conj.scope_id()).then_some(*branch_id))
+            .collect::<Vec<_>>();
+        self.branch_ids.retain(|branch_id| !unsatisfiable_branch_ids.contains(branch_id));
+        self.conjunctions.retain(|conj| !unsatisfiable.contains(&conj.scope_id()))
```
krishnangovindraj, line 64: Yes, the type-system would have caught this if I had used `Vec<(BranchId, Conjunction)>` instead of two separate vecs. But that change was big when I tried it in the morning.

krishnangovindraj: BranchIDs are allocated from a u16 that keeps going up. The way it's implemented now, they will indeed be sequential.
When a DisjunctionExecutor executes a branch with BranchID `x` it will do `output_row.provenance |= 1 << x;`

dmitrii-ubskii (approved):

ir/pattern/disjunction.rs
krishnangovindraj: Yes. Unsatisfiable branches will remain 0.

Do we return them in the header?
In the HTTP encoding, no. We only return those indices which are 1. So a provenance of `0..1101` will be encoded as `[0,2,3]`
In the GRPC encoding, I intend to send them over as a little endian byte-array, where one would have to do:
`branch_was_satisfied = byte_array[index / 8] & (1 << index % 8)`

krishnangovindraj: ohh unsatisfiable. Sorry, I read unsatisfied for some reason.
That's a good question. We should but idk if the pruning optimisation is taken into account. I'll check.

flyingsilverfin:
Was the line of questions here resolved @dmitrii-ubskii ?

ir/pattern/disjunction.rs
krishnangovindraj: I've moved the `extract_query_structure` call to be before we apply transformations, so we get the query-structure as is. Unsatisfiable branches should now be encoded in the response.

server/service/http/message/query/query_structure.rs
```diff
@@ -278,6 +265,9 @@ fn query_structure_edge(
         | Constraint::Value(_)
         | Constraint::Is(_)
         | Constraint::Iid(_) => {}
+        Constraint::IndexedRelation(indexed) => {
+            unreachable!("This is unreachable since we extract query-structure before we apply transformations");
+        }
```
dmitrii-ubskii, line 270: I think we should keep the handling as it was. It was already implemented, and I can totally see us pushing the shortcut edges to initial translation at some point.

PR #7459, opened 2025-05-14
Revise query structure output to be IR instead of edges

krishnangovindraj:
Revise query structure output to be IR instead of edges

## Implementation
Also various changes on the json structure of the `queryStructure` field

### DIFF
--- a/server/service/http/message/query/query_structure.rs
+++ b/server/service/http/message/query/query_structure.rs
@@ -55,23 +55,21 @@ impl<'a, Snapshot: ReadableSnapshot> QueryStructureContext<'a, Snapshot> {
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct QueryStructureResponse {
-    branches: Vec<QueryStructureBranchResponse>,
+    blocks: Vec<QueryStructureBlockResponse>,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct QueryStructureBranchResponse {
+pub struct QueryStructureBlockResponse {
     constraints: Vec<QueryStructureConstraintResponse>,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase", tag = "tag")]
 pub enum QueryStructureConstraint {
-    Isa {
-        instance: QueryStructureVertexResponse,
-        r#type: QueryStructureVertexResponse,
-        exactness: QueryStructureConstraintExactness,
-    },
+    Isa(QueryConstraintIsaBase),
+    #[serde(rename = "isa!")]
+    IsaExact(QueryConstraintIsaBase),
     Has {
         owner: QueryStructureVertexResponse,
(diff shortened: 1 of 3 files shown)


server/service/http/message/query/query_structure.rs
```diff
@@ -55,67 +55,125 @@ impl<'a, Snapshot: ReadableSnapshot> QueryStructureContext<'a, Snapshot> {
...
+
 pub(crate) fn encode_query_structure(
     snapshot: &impl ReadableSnapshot,
     type_manager: &TypeManager,
     query_structure: &QueryStructure,
 ) -> Result<QueryStructureResponse, Box<ConceptReadError>> {
-    let branches = query_structure
+    let blocks = query_structure
         .parametrised_structure
         .branches
```
krishnangovindraj, line 178: I can propagate this all the way if needed, but maybe after the first round of review.

server/service/http/message/query/query_structure.rs
```diff
@@ -55,67 +55,125 @@ impl<'a, Snapshot: ReadableSnapshot> QueryStructureContext<'a, Snapshot> {
...
 #[derive(Debug, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase", tag = "kind", content = "param")]
-pub enum QueryStructureEdgeTypeResponse {
-    Isa,
-    Has,
-    Links(QueryStructureVertexResponse),
+#[serde(rename_all = "camelCase", tag = "tag")]
+pub enum QueryStructureConstraint {
+    Isa(QueryConstraintIsaBase),
+    #[serde(rename = "isa!")]
```
flyingsilverfin, line 71: nice

server/service/http/message/query/query_structure.rs
```diff
@@ -55,67 +55,125 @@ impl<'a, Snapshot: ReadableSnapshot> QueryStructureContext<'a, Snapshot> {
...
+struct QueryConstraintSubBase {
+    subtype: QueryStructureVertexResponse,
+    supertype: QueryStructureVertexResponse,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum QueryStructureConstraintExactness {
+    Exact,
+    Subtypes,
```
flyingsilverfin, line 134: Can we use 'direct/transitive' for this? Is this even used anymore, if it's rolled into the variant tag??

server/service/http/message/query/query_structure.rs
flyingsilverfin: Yes let's do that in the next round

server/service/http/message/query/query_structure.rs
flyingsilverfin: Perhaps we can call it "Transitivity::Direct/Transitive"

flyingsilverfin (approved):
Approved, but please make requested changes

server/service/http/message/query/query_structure.rs
krishnangovindraj: Removed. Snuck through the `git` ing.

PR #7463, opened 2025-05-15
Fix variable visibility during translation

### LINKED ISSUE #7462: Server crashes and problems with disjunctions and fetch
## Description
Disjunctions work pretty badly with `fetch` and frequently lead to crashes. This issue compiles a selection of problematic cases for examination.

Generally, all fetch cases are covered in [behavior scenarios](https://github.com/typedb/typedb-behaviour/blob/master/query/language/fetch.feature) (they are separated by `######`). I'd recommend writing new tests with disjunctions for every fetch operation we have as the errors may be different (as shown below).

## Environment

1. TypeDB distribution: Core
2. TypeDB version: 3.3.0
3. Environment: Mac
4. Client and version: Console 3.3.0
5. Other details:

## Reproducible steps

Setup:
```
define entity person owns name; attribute name value string;
```
```
insert $p isa person, has name "John";
```

Then, proceed with the queries below.

### CRASH: Fetching an entity's attribute with an optional variable
```
match {$p isa person;} or {$k isa person;}; fetch {"p": $p.name};

Error executing command: 'match {$p isa person;} or {$k isa person;}; fetch {"p": $p.name};'
h2 protocol error: error reading a body from connection
```
Server:
```
thread 'tokio-runtime-worker' panicked at compiler/annotation/fetch.rs:178:69:
// let owner_types = input_type_annotations.get(&variable).unwrap();
```

### ANOTHER CRASH: Fetching a function call's result with an optional variable

```
 with fun name($p: person) -> name: 
match $p has name $n;
return first $n;
 
match {$p isa person;} or {$k isa person;}; fetch {"name": name($p)};

Error executing command: 'with fun name($p: person) -> name: 
match $p has name $n;
return first $n;
 
match {$p isa person;} or {$k isa person;}; fetch {"name": name($p)};'
h2 protocol error: error reading a body from connection
```
Server:
```
thread 'tokio-runtime-worker' panicked at compiler/annotation/function.rs:232:13:
internal error: entered unreachable code: The type annotations for the argument in the function call should be known by now
```

### Bad error message: Fetching all attributes with an optional variable
The error message could contain the actual variable name (`$p` instead of `$0`):

```
match {$p isa person;} or {$k isa person;}; fetch {$p.*};'

[FCP1] Internal compilation error - fetch variable $0 not found in expected inputs
Caused: [ECP4] Error compiling fetch stage into executable.
Caused: [QEX10] Error compiling query.
```

krishnangovindraj:
Fixes variable visibility tracking so that variables which are either: (1) disjunction local (2) deleted, or (3) reduced away; are not visible to subsequent stages. 

## Implementation
`TranslationContext.visible_variables` is updated as follows:
* After translating a block, only variables visible in the root conjunction of the block are retained.
* After translating a delete, deleted variables are removed.
* After translating a reduce, only output & groupby variables are retained.

Also introduces `validation.feature` with tests covering these cases.

* fixes #7462


farost:
Does this fix https://github.com/typedb/typedb/issues/7462?

farost:
If it does, then let's link it

farost (approved):
Looks good to me, will wait for new tests in the BDD pr

flyingsilverfin (approved):

PR #7464, opened 2025-05-15
Overriding config file location via CLI is now relative to working directory

krishnangovindraj:
The config file override specified via '--config' is now relative to the working directory. It was previously relative to the server binary.
Makes it easier to run typedb through bazel by copying the config file to the default location, avoiding the need for a CLI override.

## Implementation
Uses a genrule to copy `server/config.yml` to `config.yml`

What works now?
`bazel run //:typedb_server_bin`
`cargo run -- --config server/config.yml`

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -27,6 +27,13 @@ exports_files(
     ["VERSION", "deployment.bzl", "LICENSE", "README.md"],
 )
 
+genrule(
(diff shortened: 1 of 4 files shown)


flyingsilverfin (approved):
Good idea to have it relative to executable

PR #7465, opened 2025-05-15
Filter anonymous variables from insert stage output

krishnangovindraj:
Consistent with the behaviour of match stages, we no longer include anonymous variables in the output of an insert, update or put stage.


krishnangovindraj:
test_write_pipelines luckily broke.
Anonymous variables were written into the row, and the next match stage would just re-use the position. 
So when it tried to write a different value to that location, it failed.

PR #7466, opened 2025-05-15
Refactor HTTP API output to include the structure of the pipeline

krishnangovindraj:
Adds the structure of the pipeline to the query structure returned by the HTTP API. Previously, we would just return a flat list of blocks without any information of how these blocks were connected (through stages and connectives such as or/not)

## Implementation

Having a fixed-width provenance removes the need for executors to be aware of the provenance width. Though this can usually be inferred from input rows, tracking it is a lot of moving parts. So we'd like to stick to fixed-width.

### DIFF
--- a/compiler/query_structure.rs
+++ b/compiler/query_structure.rs
@@ -17,11 +17,51 @@ use ir::{
     pipeline::{ParameterRegistry, VariableRegistry},
 };
 use itertools::Itertools;
+use serde::{Deserialize, Serialize};
 
-use crate::annotation::pipeline::AnnotatedStage;
+use crate::annotation::{
+    pipeline::AnnotatedStage,
+    type_annotations::{BlockAnnotations, TypeAnnotations},
+};
+
+#[derive(Debug, Clone)]
+pub struct QueryStructureBlock {
+    pub constraints: Vec<Constraint<Variable>>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct QueryStructureBlockID(pub u16);
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct QueryStructureConjunction {
+    conjunction: Vec<QueryStructureConjunct>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum QueryStructureConjunct {
+    Block(QueryStructureBlockID),
+    Or { branches: Vec<QueryStructureConjunction> },
+    Not(QueryStructureConjunction),
+    Try(QueryStructureConjunction),
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum QueryStructureStage {
+    Match(QueryStructureConjunction),
+    Insert { block: QueryStructureBlockID },
+    Put { block: QueryStructureBlockID },
+    Update { block: QueryStructureBlockID },
+    // Select { variables: Vec<Variable> },
+    // TODO...
+}
+
 #[derive(Debug, Clone)]
 pub struct ParametrisedQueryStructure {
-    pub branches: [Option<Vec<Constraint<Variable>>>; 64],
+    pub stages: Vec<QueryStructureStage>,
+    pub blocks: Vec<QueryStructureBlock>,
     pub resolved_labels: HashMap<Label, answer::Type>,
     pub calls_syntax: HashMap<Constraint<Variable>, String>,
 }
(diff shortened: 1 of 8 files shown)


compiler/query_structure.rs
```diff
@@ -84,67 +170,81 @@ pub fn extract_query_structure_from(
...
+        let branch_id = if let Some(BranchID(id)) = existing_branch_id {
+            debug_assert!((id as usize) < self.inner.blocks.len());
+            self.inner.blocks[id as usize] = QueryStructureBlock { constraints: Vec::from(constraints) };
+            QueryStructureBlockID(id)
+        } else {
+            self.inner.blocks.push(QueryStructureBlock { constraints: Vec::from(constraints) });
+            QueryStructureBlockID(self.inner.blocks.len() as u16 - 1)
+        };
+        branch_id
+    }
```
krishnangovindraj, line 226: For blocks which already have a BranchID (i.e. branches of a disjunction & optional blocks), we assign BlockID = BranchID. This avoids having to translate provenance to BlockIDs.

server/service/http/message/query/row.rs
```diff
@@ -48,7 +49,12 @@ pub fn encode_row<'a>(
         )?;
         encoded_row.insert(variable.as_str(), row_entry);
     }
-    let involved_blocks = row.provenance().branch_ids().map(|b| b.0 as u8).collect();
+    let involved_blocks = row
+        .provenance()
+        .branch_ids()
+        .map(|b| b.0)
+        .chain(always_taken_blocks.unwrap_or(&vec![]).iter().map(|b| b.0))
+        .collect();
```
krishnangovindraj, line 57: Append always taken blocks (root blocks of stages) to the BlockIDs from provenance.

compiler/query_structure.rs
krishnangovindraj, line 1: There's actually no good reason for this to be in compiler. I'd ideally have it in query, but that's a bigger refactor that can come later.

compiler/query_structure.rs
```diff
@@ -35,45 +64,102 @@ impl ParametrisedQueryStructure {
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum QueryStructureStage {
+    Match(QueryStructureConjunction),
+    Insert { block: QueryStructureBlockID },
+    Put { block: QueryStructureBlockID },
+    Update { block: QueryStructureBlockID },
+    // Select { variables: Vec<Variable> },
+    // TODO...
```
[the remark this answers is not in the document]
flyingsilverfin: What' this TODO for?

compiler/query_structure.rs
```diff
@@ -31,124 +63,332 @@ impl ParametrisedQueryStructure {
...
+    conjunction: Vec<QueryStructureConjunct>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", tag = "tag")]
+pub enum QueryStructureConjunct {
+    Block { index: QueryStructureBlockID },
+    Or { branches: Vec<QueryStructureConjunction> },
+    Not(QueryStructureConjunction),
+    Try(QueryStructureConjunction),
```
flyingsilverfin, line 139: Oh fun I didn't know you could mix these syntaxes

compiler/query_structure.rs
```diff
@@ -31,124 +63,332 @@ impl ParametrisedQueryStructure {
...
+            .keys()
+            .filter(|v| !v.is_anonymous())
+            .map(|v| StructureVariableId::from(v))
+            .collect::<Vec<_>>();
         available_variables.sort();
+        let variable_names = variable_names.iter().map(|(var, name)| (var.into(), name.clone())).collect();
         QueryStructure { parametrised_structure: self, parameters, variable_names, available_variables }
     }
+
+    pub fn always_taken_blocks(&self) -> Vec<QueryStructureBlockID> {
```
flyingsilverfin, line 79: What is this/why do we have it?

server/service/http/message/query/query_structure.rs
```diff
@@ -40,7 +40,11 @@ impl<'a, Snapshot: ReadableSnapshot> QueryStructureContext<'a, Snapshot> {
         self.query_structure.parameters.iid(*param).map(|iid| iid.as_ref())
     }
 
-    pub fn get_variable_name(&self, variable: &Variable) -> Option<String> {
+    pub fn get_variable_name(&self, variable: &StructureVariableId) -> Option<String> {
+        self.get_structure_variable_name(variable)
+    }
+
+    pub fn get_structure_variable_name(&self, variable: &StructureVariableId) -> Option<String> {
```
flyingsilverfin: why do we need both? signatures look the same

flyingsilverfin (commented):
Mostly some questions
also please update the PR description with the expected HTTP output, this is definitely something we'd want to see to check our own implementation quickly in the future.
+ good for users to see immediately.

compiler/query_structure.rs
krishnangovindraj: The main trunk of the query (across all pipelines) has BranchID=0  (BranchIDs are used upto the executor to track provenance). 
Since we're now representing pipeline stages separately, we need the trunk from each pipeline stage to be discernible. So we introduce a distinct concept called a `BlockID`. These are those BlockIDs and we add them to the provenance of each row.
Unifying them will likely require extending BranchID to well beyond 64 bits. 

The ParametrisedQueryStructureBuilder allocates these. I've avoided mapping BranchID to BlockID by making BlockID = BranchID wherever we have a BranchID - i.e. disjunction branches & optionals.
I think the root no longer has an explicitly assigned BranchID.

server/service/http/message/query/query_structure.rs
krishnangovindraj: Cleaned up. Must have been me merging two structures into one.

compiler/query_structure.rs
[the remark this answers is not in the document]
krishnangovindraj: To implement the remaining ones. Which I have since done so I've removed it now.

compiler/query_structure.rs
```diff
@@ -31,124 +63,331 @@ impl ParametrisedQueryStructure {
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct QueryStructureConjunction {
+    conjunction: Vec<QueryStructureConjunct>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", tag = "tag")]
+pub enum QueryStructureConjunct {
```
flyingsilverfin: I think this is equivalent to "Pattern" in our existing terminology

compiler/query_structure.rs
```diff
@@ -31,124 +63,331 @@ impl ParametrisedQueryStructure {
...
-fn extract_query_structure_from_branch(
-    branches: &mut [Option<Vec<Constraint<Variable>>>; 64],
-    branch_id: BranchID,
-    conjunction: &Conjunction,
-) {
-    if branches[branch_id.0 as usize].is_none() {
-        branches[branch_id.0 as usize] = Some(Vec::new());
+    fn add_block(
+        &mut self,
+        existing_branch_id: Option<BranchID>,
```
flyingsilverfin, line 222: We always assume our existing branch IDs exist I thought? Since those are u16's, that's pretty safe

compiler/query_structure.rs
```diff
@@ -31,124 +63,331 @@ impl ParametrisedQueryStructure {
...
+            conjunction.constraints(),
+            block_annotations.type_annotations_of(conjunction).unwrap(),
+        );
+        conjuncts.push(QueryStructureConjunct::Block { index: block_id });
+        conjunction.nested_patterns().iter().for_each(|nested| match nested {
+            NestedPattern::Disjunction(disjunction) => {
+                let branches = disjunction
+                    .branch_ids()
+                    .iter()
+                    .zip(disjunction.conjunctions().iter())
```
flyingsilverfin: feels like this should be an API on disjunction directly IMO, the fact that branch IDs and conjunctions are stored as paralle vec's is implementation detail and shouldn't be relied on externally.

compiler/query_structure.rs
flyingsilverfin: Correct me if I'm wrong but:
```
match X;
match Y:
match { 1 } or { 2 }; // branch IDs
```
Won't this end up with the existing branch ID 1 and existing branch ID 2 overwriting the data X and Y inserted?

```
// 1. first step
[QueryStructureBlock(X)]
return StructureBlockId(0)

// 2. second step
[QueryStructureBlock(X), QueryStructureBlock(Y)]
return StructureBlockId(1)

// 3. third step, first disjunction term with existing branch ID 1
[QueryStructureBlock(X), QueryStructureBlock( DISJUNCTION(1) ) ]
return StructureBlockId(1) // returned again?
```

flyingsilverfin: Ah I see you've preallocated the vec to start with i see

compiler/query_structure.rs
```diff
@@ -31,124 +63,331 @@ impl ParametrisedQueryStructure {
...
+
+#[derive(Debug, Clone)]
+pub struct ParametrisedQueryStructureBuilder<'a> {
+    query_structure: ParametrisedQueryStructure,
+    source_query: &'a str,
+}
+
+impl<'a> ParametrisedQueryStructureBuilder<'a> {
+    fn new(source_query: &'a str, branch_ids_allocated: u16) -> Self {
+        // Pre-allocate for the optional branches. They come first
```
flyingsilverfin: Update comment `// Pre-allocated for query branches that have already been allocated branch ids` (the fact that only optionals have that is a separate fact - we might change conjunctions to be assigned branch IDs at some point and we don't want this to be inaccurate

compiler/query_structure.rs
krishnangovindraj: Will update. It feels a bit ambiguous, but it's definitely consistent.

compiler/query_structure.rs
krishnangovindraj: They may not have been allocated a branch_id

PR #7468, opened 2025-05-16
Fix variable category check for 'is' constraints

krishnangovindraj:
Fix the variable category check for 'is' constraints which would always fail, &  makes it recursive to catch more cases.

## Implementation
Replace faulty `find_or_first` with `find`.


farost:
The fix looks reasonable, I'd like to write a test for it

farost (approved):

flyingsilverfin (approved):

PR #7472, opened 2025-05-19
Add port 8000 to list of ports on docker image

krishnangovindraj:
Add port 8000 to the docker image so the HTTP API is accessible.


farost (approved):
Good catch

PR #7474, opened 2025-05-21
Fix forcing development mode in the config

krishnangovindraj:
Fix the compilation time forcing of development mode for local and snapshot builds

## Implementation
Wraps `ConfigBuilder` around `Config` as we do with Blocks

### DIFF
--- a/server/parameters/config.rs
+++ b/server/parameters/config.rs
@@ -164,6 +164,7 @@ impl ConfigBuilder {
             .map_err(|source| ConfigError::ErrorParsingYaml { source })
             .map(|config| Self { config })
     }
+
     pub fn override_with_cliargs(&mut self, cliargs: CLIArgs) {
         let CLIArgs {
             config_file_override: _,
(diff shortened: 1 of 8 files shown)


server/parameters/config.rs
```diff
@@ -200,8 +196,8 @@ pub struct ConfigBuilder {
     is_development_mode: Option<bool>,
 }
 
-impl ConfigBuilder {
-    fn server_address(mut self, address: impl Into<String>) -> Self {
+impl ConfigBuilderForTests {
```
farost, line 199: `TestConfigBuilder`?
And I don't really see a need to rename it... If we think about it, it might be useful when we expose TypeDB as a crate, so it's not necessary to use config files all the way around. However, we need to be careful with:
1. Not forgetting about the development mode flag anymore, now we have two places to think about
2. Maintaining the builder. Do we want to do it? Is it really safe if it's not used globally? It can have its own bugs we don't find.

krishnangovindraj: I can't really think of another way to do it apart from moving it into tests altogether.

farost:
I'd also enhance the RN description to at least: "Fix the compilation time forcing of development mode for local and snapshot builds for development purposes", so the users are less confused

PR #7477, opened 2025-05-23
Introduce file-based database export and import

farost:
Add database export and database import operations. Unlike in TypeDB 2.x, these operations are run through TypeDB **GRPC clients** such as the **Rust driver** or **TypeDB Console**, which solves a number of issues with networking and encryption, especially relevant for users of TypeDB Enterprise. With this, it becomes an official part of the [TypeDB GPRC protocol](https://github.com/typedb/typedb-protocol/pull/224). 
Both operations are performed through the network, but the server and the client can be used on the same host.

Each TypeDB database can be represented as two files:
1. A text file with its TypeQL schema description: a complete `define` query for the whole schema.
2. A binary file with its data.

This format is an extension to the TypeDB 2.x's export format. See more details below for version compatibility information.

### Database export
Database export allows a client to download database schema and data files from a TypeDB server for future import to the same or higher TypeDB version.

The files are created on the client side. While the database data is being exported, parallel queries are allowed, but none of them will affect the exported data thanks to TypeDB's transactionality.
However, the database will not be available for such operations as deletion.

**Exported TypeDB 3.x databases cannot be imported into servers of older versions.**

### Database import
Database import allows a client to upload previously exported database schema and data files to a TypeDB server. It is possible to assign any new name to the imported database.

While the database data is being imported, it is not recommended to perform parallel operations against it.
Interfering actions may lead to import errors or database corruption.

**Import supports all exported TypeDB 2.x and TypeDB 3.x databases.** It can be used to migrate between TypeDB versions with breaking changes. Please visit [our docs](https://typedb.com/docs/manual/migration/2_to_3) for more information.

## Implementation
Implement [the new protocol](https://github.com/typedb/typedb-protocol/pull/224).

The two operations are implemented as two separate services running in `tokio` tasks, similar to transaction services. 

### Database export 
This operation is simple. After establishing a stream, we just export the database's schema (the same operation as schema retrieval) and then send a number of items containing the header, the database's data (encoded concepts), and data checksums at the end.

### Database import
This operation is more tricky. It is executed in steps, containing a couple of "states":
1. The database's name and schema are expected. Without it, we can't continue. Ignoring this step leads to an error, signaling that the client is probably implemented incorrectly.
2. After the schema is received, it is executed and committed as is, to check that the provided schema is correct. Otherwise, a user error is returned, so they can rewrite the schema and try again.
3. After the schema is persisted, it is relaxed. This contains: a) substituting default cardinalities and card/key annotations with `@card(0..)` for all capabilities in the schema b) making all attributes independent (`@independent`) so they are not cleaned up between transactions without owners if they are not yet received c) making all relations independent (a new system `system independent` property, not exposed to users) so they are not cleaned up between transactions without role players. All errors at this stage are considered internal errors and bugs.
4. After the schema is prepared, we receive items, decode them into concepts, and persist in the database. There is a transaction buffer size we use to execute commits from time to time, reducing memory consumption and final commit time. This optimization requires 3b and 3c. One of the expected and required items is checksums. Another optional item is the header (we don't really use it outside of logs, so it's not required. Not sure if we need to force it).
5. After the `Done` message is received, signaling that the stream of items is completed, we perform the final data commit and unrelax the schema (undo 3).
6. If everything is good, a verification `Done` response is sent.
At any point, an error can be returned. To make it possible, the protocol introduces streaming from the server, although it remains silent until the end unless there are errors in the provided messages.

While being imported, databases are not accessible through `database_manager`: they are owned by `database_importer`. If a server crashes, the uncompleted databases will be cleaned up on the next bootup.

To avoid overflowing memory, `InstanceIDMapping` uses `SpilloverCache`, a new component combining `HashMap` and RocksDB to spill over the excessive data which is too much to fit in memory.


server/state.rs
```diff
@@ -227,85 +232,66 @@ impl ServerState {
...
         let transaction = TransactionRead::open(database, TransactionOptions::default())
-            .map_err(|err| StateError::FailedToOpenPrerequisiteTransaction {})?;
-        let types_syntax = Self::get_types_syntax(&transaction)?;
-
-        let type_schema = match types_syntax.is_empty() {
-            true => String::new(),
-            false => format!("{}\n{}", typeql::token::Clause::Define, types_syntax),
-        };
+            .map_err(|err| ServerStateError::FailedToOpenPrerequisiteTransaction {})?;
+        // TODO: Close transactions on errors?
```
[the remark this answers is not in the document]
farost: https://github.com/farost/typedb/pull/3

farost:
assembly tests will be green once we finish updating protocol

server/service/grpc/migration/import_service.rs
```diff
@@ -0,0 +1,1167 @@
...
+    database_info: Option<DatabaseInfo>,
+    is_done: bool,
+}
+
+impl DatabaseImportService {
+    const OPTIONS_PARALLEL: bool = true;
+    const OPTIONS_SCHEMA_LOCK_ACQUIRE_TIMEOUT_MILLIS: u64 = Duration::from_secs(10).as_millis() as u64;
+    const OPTIONS_TRANSACTION_TIMEOUT_MILLIS: u64 = Duration::from_secs(6 * SECONDS_IN_HOUR).as_millis() as u64;
+
+    const COMMIT_BATCH_SIZE: u64 = 10_000;
```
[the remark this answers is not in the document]
flyingsilverfin: Probably fine since we're not doing anything concurrent atm

server/service/grpc/migration/export_service.rs
```diff
@@ -0,0 +1,303 @@
...
+    checksums: Checksums,
+    shutdown_receiver: watch::Receiver<()>,
+}
+
+impl DatabaseExportService {
+    const ITEM_BATCH_SIZE: usize = 250;
+
+    const OPTIONS_PARALLEL: bool = true;
+    const OPTIONS_SCHEMA_LOCK_ACQUIRE_TIMEOUT_MILLIS: u64 = Duration::from_secs(10).as_millis() as u64;
+    const OPTIONS_TRANSACTION_TIMEOUT_MILLIS: u64 = Duration::from_secs(6 * SECONDS_IN_HOUR).as_millis() as u64;
```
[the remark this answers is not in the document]
flyingsilverfin: I'd make it a day just in case imo

server/service/grpc/migration/import_service.rs
```diff
@@ -0,0 +1,1191 @@
...
+            }
+            None => Err(DatabaseImportError::NoClientChecksumsOnDone {}),
+        }
+    }
+}
+
+#[derive(Debug)]
+struct InstanceBuffer<T: ThingAPI> {
+    original_to_buffered_ids: HashMap<String, IID>,
+    buffered_ids_to_instances: HashMap<IID, T>,
```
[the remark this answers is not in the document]
flyingsilverfin: Why do we need the 'buffered_ids_to_instances'? Surely we can just ask the Concept API to get the concept by the IID, and that's already saved in the DB?

server/service/grpc/migration/import_service.rs
```diff
@@ -0,0 +1,1167 @@
...
+    shutdown_receiver: watch::Receiver<()>,
+
+    database_info: Option<DatabaseInfo>,
+    is_done: bool,
+}
+
+impl DatabaseImportService {
+    const OPTIONS_PARALLEL: bool = true;
+    const OPTIONS_SCHEMA_LOCK_ACQUIRE_TIMEOUT_MILLIS: u64 = Duration::from_secs(10).as_millis() as u64;
+    const OPTIONS_TRANSACTION_TIMEOUT_MILLIS: u64 = Duration::from_secs(6 * SECONDS_IN_HOUR).as_millis() as u64;
```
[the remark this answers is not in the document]
flyingsilverfin: let's do a day too

server/service/grpc/migration/export_service.rs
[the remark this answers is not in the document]
farost: One day it is

common/bytes/util.rs
```diff
@@ -83,6 +85,10 @@ impl<'a> HexBytesFormatter<'a> {
         self.0.iter().for_each(|byte| write!(result, "{byte:02x}").expect("Expected IID formatting"));
         result
     }
+
+    pub fn to_base64(&self) -> String {
```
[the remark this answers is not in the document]
farost: Fair

encoding/layout/infix.rs
```diff
@@ -78,6 +79,7 @@ impl Infix {
         PropertyLabel => [0];
         PropertyValueType => [1];
         PropertyOrdering => [2];
+        PropertyIndependency => [3];
```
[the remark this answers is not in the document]
farost: Should not the struct be `RelationTypeIndependent` then?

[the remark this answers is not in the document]
farost: That's the reason why I didn't call it `Independent`, but I also didn't really want to add "Internal" or smth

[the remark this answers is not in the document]
farost: I will call everything like you said for now, but let's look at it once more

server/service/export_service.rs
```diff
@@ -0,0 +1,57 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+use concept::error::ConceptReadError;
+use database::transaction::{TransactionError, TransactionRead};
+use error::typedb_error;
+use ir::pipeline::FunctionReadError;
+use storage::durability_client::DurabilityClient;
+
+pub(crate) fn get_transaction_schema<D: DurabilityClient>(
```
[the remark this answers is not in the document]
farost: It's for the outside code as it imports these functions, to draw a line between `get_database_schema` and `get_transaction_schema.` Maybe you're right

server/service/grpc/migration/import_service.rs
```diff
@@ -0,0 +1,1191 @@
...
+    }
+}
+
+type IID = Bytes<'static, BUFFER_KEY_INLINE>;
+
+#[derive(Debug)]
+struct DataInfo {
+    objects: ObjectsInfo,
+    attributes: AttributesInfo,
+    self_checksums: Checksums,
```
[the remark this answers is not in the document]
farost: Done
I saw the used terminology in 2.x and kinda liked the idea of avoiding "expected" here because any party can be wrong. But it's easier to reason with the names you proposed

server/service/grpc/migration/import_service.rs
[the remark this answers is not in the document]
farost: 1. Regarding RocksDB, I decided to leave it the simple way until I tested the huge loads. The implementation should not be complicated. As discussed in Discord, I'll add an additional Rocksdb layer for cases when the number of items exceeds a threshold set.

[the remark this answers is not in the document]
farost: 2. Regarding the additional cache, yeah, it's an overoptimization, I should just use the concept api.

server/service/grpc/migration/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    fn transaction(&self) -> &TransactionRead<WALClient> {
+        self.0.as_ref().expect("Expected transaction")
+    }
+}
+
+impl Drop for TransactionHolder {
+    fn drop(&mut self) {
+        if let Some(transaction) = self.0.take() {
+            transaction.close();
```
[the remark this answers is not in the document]
farost: It was a temporary thing. This struct is a mess by itself.

I've implemented the general Drop for transactions now: https://github.com/farost/typedb/pull/3 
You better review it, too

As it's a general behavior now, we can't put an assert, I believe. Not sure. 
But it's anyway better than leaving transactions live without releasing their resources.

concept/type_/mod.rs
```diff
@@ -775,6 +775,27 @@ impl TypeEdgePropertyEncoding for Ordering {
     }
 }
 
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, Default, Eq, PartialEq, Hash)]
+pub struct Independency;
```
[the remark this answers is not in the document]
farost: I'll rename it the way you proposed, but I've left another response to one of your other comments below seeking for consistency

server/service/grpc/migration/import_service.rs
```diff
@@ -0,0 +1,1191 @@
...
+    }
+
+    fn fulfill_awaiting_ownerships(
+        snapshot: &mut impl WritableSnapshot,
+        thing_manager: &ThingManager,
+        original_id: &str,
+        attribute: &Attribute,
+        data_info: &mut DataInfo,
+    ) -> Result<(), DatabaseImportError> {
+        if let Some(awaiting_objects) = data_info.attributes.awaited_for_ownerships.remove(original_id) {
```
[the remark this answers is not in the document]
farost: I'll check it out

common/cache/BUILD
farost, line 1: Overall cache file, can contain different caches potentially

database/database.rs
```diff
@@ -399,13 +406,64 @@ impl Database<WALClient> {
         Ok(database)
     }
 
+    fn verify_files_consistency(path: &Path) -> Result<(), DatabaseOpenError> {
```
farost: Handle database import: if the database is still marked as "importing" on a bootup, it's not ready, and DatabaseManager can decide if it's fine or not. 
In the meantime, all cache files should be removed.

database/database.rs
```diff
@@ -399,13 +406,64 @@ impl Database<WALClient> {
...
+    }
+
     fn checkpoint(&self) -> Result<(), CheckpointCreateError> {
         let checkpoint = Checkpoint::new(&self.path)?;
         self.storage.checkpoint(&checkpoint)?;
         checkpoint.finish()?;
         Ok(())
     }
 
+    pub(super) fn mark_imported(&self) -> Result<(), DatabaseCreateError> {
```
farost: As the consistency checked here, I also give the ownership over "importing" files to `Database`.

database/database_manager.rs
```diff
@@ -45,36 +47,49 @@ impl DatabaseManager {
...
-            let database = Database::<WALClient>::open(&entry_path)?;
-            assert!(!databases.contains_key(database.name()));
-            databases.insert(database.name().to_owned(), Arc::new(database));
+            match Database::<WALClient>::open(&entry_path) {
+                Ok(database) => {
+                    assert!(!databases.contains_key(database.name()));
+                    databases.insert(database.name().to_owned(), Arc::new(database));
+                }
+                Err(DatabaseOpenError::IncompleteDatabaseImport {}) => {
+                    event!(Level::WARN, "Database {database_name} is in an incomplete state after an interrupted import operation. It will be deleted.");
```
farost: Remove incomplete databases (at least for now)

database/database_manager.rs
```diff
@@ -104,6 +119,63 @@ impl DatabaseManager {
         Ok(())
     }
 
+    pub(crate) fn prepare_imported_database(&self, name: String) -> Result<Database<WALClient>, DatabaseCreateError> {
```
farost, line 184: The database manager prepares a database, records that this name is reserved, and gives the ownership to the caller (the caller should be inside the `crate`).

database/database_manager.rs
```diff
@@ -104,6 +119,63 @@ impl DatabaseManager {
...
+        }
+        imported_databases.insert(name.to_string());
+
+        let database = Database::<WALClient>::open(&self.data_directory.join(name))
+            .map_err(|typedb_source| DatabaseCreateError::DatabaseOpen { typedb_source })?;
+        database.mark_imported()?;
+        Ok(database)
+    }
+
+    pub(crate) fn finalise_imported_database(&self, database: Database<WALClient>) -> Result<(), DatabaseCreateError> {
```
farost, line 205: Take the ownership of the database and remove all the "importing" markings

database/database_manager.rs
```diff
@@ -104,6 +119,63 @@ impl DatabaseManager {
...
+        if databases.contains_key(&name) {
+            return Err(DatabaseCreateError::AlreadyExists { name });
+        } else {
+            database.unmark_imported()?;
+            databases.insert(name, Arc::new(database));
+            Ok(())
+        }
+    }
+
+    pub(crate) fn cancel_database_import(&self, database: Database<WALClient>) -> Result<(), DatabaseDeleteError> {
```
farost, line 226: If the import is aborted, we should clean up our caches and the database itself.

I wonder if it would be easier to just put the database inside `self.databases`, but mark it as incomplete. The idea of ownership transition is cool though, and I don't see much difference between these two approaches, at least now.

database/database_manager.rs
```diff
@@ -104,6 +119,63 @@ impl DatabaseManager {
...
+        let database = Database::<WALClient>::open(&self.data_directory.join(name))
+            .map_err(|typedb_source| DatabaseCreateError::DatabaseOpen { typedb_source })?;
+        database.mark_imported()?;
+        Ok(database)
+    }
+
+    pub(crate) fn finalise_imported_database(&self, database: Database<WALClient>) -> Result<(), DatabaseCreateError> {
+        let name = database.name().to_string();
+
+        let mut imported_databases =
```
farost: Intentionally hold the lock until the end of the method so all the import-related methods stop in the beginning and wait until the other finishes its job

database/lib.rs
```diff
@@ -11,4 +11,6 @@ pub use self::database::{Database, DatabaseDeleteError, DatabaseOpenError, Datab
 
 pub mod database;
 pub mod database_manager;
+pub mod migration;
+pub mod query;
```
farost, line 15: Move some utils for querying within transactions into `service/transaction_service.rs`. It knows about high-level transactions, so it belongs to this crate.

database/migration/database_importer.rs
farost, line 1: UPD. This new file is the upgraded version of `DatabaseInfo` from the previous iteration. This importer owns the imported database and performs all the database-related operations based on the data received from the importer service.

database/migration/database_importer.rs
```diff
@@ -0,0 +1,1034 @@
...
+            .expect("Expected imported database finalization. Cannot safely progress without it.");
+        Ok(())
+    }
+
+    fn take_owned_database(&mut self) -> Option<Database<WALClient>> {
+        let Some(database) = self.database.take() else {
+            return None;
+        };
+        Some(
+            Arc::into_inner(database)
```
farost, line 439: We intentionally close write transactions before calling it.

Uncommitted schema transactions (because of errors or whatever) will be automatically cleaned up when they go out of scope on drop when I merge this: https://github.com/farost/typedb/pull/3
So we **have** to have full ownership of the database at this point..

database/migration/database_importer.rs
```diff
@@ -0,0 +1,1034 @@
...
+    fn new() -> Self {
+        Self { instance_id_mapping: InstanceIDMapping::new(), awaited_for_ownerships: HashMap::new() }
+    }
+}
+
+#[derive(Debug)]
+pub struct DatabaseImporter {
+    database_manager: Arc<DatabaseManager>,
+    database_name: String,
+    database: Option<Arc<Database<WALClient>>>, // owned by the importer!
```
farost, line 301: It's `Option` to use `take()` on `&mut self` in `Drop` to clean up the database because it should be consumed.

The idea: we initialize the database in the constructor and have it until `done`

server/lib.rs
```diff
@@ -8,9 +8,260 @@
...
 pub mod authentication;
 pub mod error;
 pub mod parameters;
-pub mod server;
 pub mod service;
-pub mod state;
+pub(crate) mod state;
+
+#[derive(Default)]
+pub struct ServerBuilder {
```
farost, line 38: Move from `server/server.rs` to `server/lib.rs`

server/service/grpc/migration/import_service.rs
```diff
@@ -0,0 +1,392 @@
...
+        let database_importer = match self.database_importer.as_mut() {
+            Some(database_importer) => database_importer,
+            None => return Err(DatabaseImportServiceError::DatabaseNotFoundForItems {}),
+        };
+
+        for item in items {
+            Self::process_item(item, database_importer).await?;
+
+            let total_items = database_importer.total_item_count();
+            if total_items != 0 && total_items % ITEMS_LOG_INTERVAL == 0 {
```
farost, line 213: Log every million items to say that the server is alive.
We may want to send this through protocol so it's printed by the client... But I did not want to overcomplicate the protocol, really.

server/service/grpc/migration/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+use database::transaction::TransactionRead;
+use storage::durability_client::WALClient;
+
+pub(crate) mod export_service;
+pub(crate) mod import_service;
+pub(crate) mod item;
+
+// TODO: Temp, remove after merges
```
farost: https://github.com/farost/typedb/pull/3

database/migration/database_importer.rs
```diff
@@ -0,0 +1,1046 @@
...
+                }
+            }
+            None => Err(DatabaseImportError::NoClientChecksumsOnDone {}),
+        }
+    }
+}
+
+#[derive(Debug)]
+struct InstanceIDMapping<T: ThingAPI> {
+    original_to_buffered_ids: SpilloverCache<IID>,
```
farost, line 224: HashMap + Rocksdb

database/migration/database_importer.rs
```diff
@@ -0,0 +1,1046 @@
...
+impl ObjectsInfo {
+    fn new(data_directory: &PathBuf) -> Self {
+        Self { instance_id_mapping: InstanceIDMapping::new(data_directory), awaited_for_roles: HashMap::new() }
+    }
+}
+
+#[derive(Debug)]
+struct AttributesInfo {
+    pub instance_id_mapping: InstanceIDMapping<Attribute>,
+    pub awaited_for_ownerships: HashMap<String, HashSet<Object>>,
```
farost, line 285: We probably need to use SpilloverCaches here (and in `awaited_for_roles` as well. But encoding and decoding these keys is a little... Ehhh... Well, I'll probably do it and retest before merging next week :/

database/migration/database_importer.rs
```diff
@@ -0,0 +1,1046 @@
...
+#[derive(Debug)]
+struct InstanceIDMapping<T: ThingAPI> {
+    original_to_buffered_ids: SpilloverCache<IID>,
+    _phantom_data: PhantomData<T>,
+}
+
+impl<T: ThingAPI> InstanceIDMapping<T> {
+    const CACHE_SPILLOVER_THRESHOLD: usize = 300_000;
+
+    fn new(data_directory: &PathBuf) -> Self {
```
farost: Rocksdb is created in the database's data directory. Once it's done, it will remove its cache. If not (because of a crash or a bug), on bootup, the database will clean it up itself.

server/service/grpc/migration/import_service.rs
flyingsilverfin: I Agree

server/service/grpc/migration/import_service.rs
```diff
@@ -0,0 +1,1191 @@
...
+    schema_info: SchemaInfo,
+    data_info: DataInfo,
+
+    data_transaction: Option<TransactionWrite<WALClient>>,
+    transaction_item_count: u64,
+}
+
+impl DatabaseInfo {
+    fn new(database: Arc<Database<WALClient>>) -> Self {
+        Self {
```
flyingsilverfin: Note: in the future Cluster where the WAL is distributed we will make this generic over D: DurabilityService instead

common/cache/cache.rs
flyingsilverfin, line 1: I'd call this module `spillover_cache`

concept/thing/thing_manager.rs
```diff
@@ -349,8 +349,8 @@ impl ThingManager {
             AttributeID::Double(id) => Ok(Value::Double(id.read().as_f64())),
             AttributeID::Decimal(id) => Ok(Value::Decimal(id.read().as_decimal())),
             AttributeID::Date(id) => Ok(Value::Date(id.read().as_naive_date())),
-            AttributeID::DateTime(id) => Ok(Value::DateTime(id.read().as_naive_date_time())),
-            AttributeID::DateTimeTZ(id) => Ok(Value::DateTimeTZ(id.read().as_date_time())),
+            AttributeID::DateTime(id) => Ok(Value::Datetime(id.read().as_naive_date_time())),
```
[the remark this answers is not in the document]
flyingsilverfin: did you do this?

database/query.rs
flyingsilverfin, line 1: nice, like this a lot

concept/thing/thing_manager.rs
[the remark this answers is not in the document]
farost: I thought that it's enough diff for this PR, I want to do it with a better control

dependencies/typedb/repositories.bzl
```diff
@@ -12,10 +12,11 @@ def typedb_bazel_distribution():
     )
 
 def typedb_dependencies():
+    # TODO: Return typedb
     git_repository(
         name = "typedb_dependencies",
-        remote = "https://github.com/typedb/dependencies",
-        commit = "b0ebf3489638774277dd3af35862c09895a08dd0",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_dependencies
+        remote = "https://github.com/farost/typedb-dependencies",
```
flyingsilverfin: don't forget to bump these back now they're merged

server/service/grpc/typedb_service.rs
```diff
@@ -211,6 +214,27 @@ impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
         )
     }
 
+    type databases_importStream = Pin<Box<ReceiverStream<Result<DatabasesImportServerProto, Status>>>>;
```
flyingsilverfin, line 217: types are normally caps: `DatabasesImportStream` no?

server/service/http/transaction_service.rs
```diff
@@ -443,8 +445,8 @@ impl TransactionService {
                 Break(())
             })
             .await
-            .expect("Expected write transaction execution completion"),
-            Transaction::Schema(transaction) => {
+            .expect("Expected write transaction commit completion"),
+            Transaction::Schema(transaction) => spawn_blocking(move || {
```
flyingsilverfin, line 449: good catch

database/database_manager.rs
```diff
@@ -45,36 +47,49 @@ impl DatabaseManager {
...
-            return Err(DatabaseCreateError::InternalDatabaseCreationProhibited {});
-        }
-        if !typeql::common::identifier::is_valid_identifier(name) {
-            return Err(DatabaseCreateError::InvalidName { name: name.to_owned() });
-        }
+        Self::validate_database_name(name.as_ref())?;
         self.create_database_unrestricted(name)
     }
 
     pub fn create_database_unrestricted(&self, name: impl AsRef<str>) -> Result<(), DatabaseCreateError> {
```
flyingsilverfin: this should be called `put_database_unrestricted` if it allows it to pre-exist

database/database_manager.rs
```diff
@@ -5,27 +5,29 @@
...
 use storage::durability_client::WALClient;
+use tracing::{event, Level};
 
 use crate::{database::DatabaseCreateError, Database, DatabaseDeleteError, DatabaseOpenError, DatabaseResetError};
 
 #[derive(Debug)]
 pub struct DatabaseManager {
     data_directory: PathBuf,
     databases: RwLock<HashMap<String, Arc<Database<WALClient>>>>,
+    imported_databases: RwLock<HashSet<String>>,
```
flyingsilverfin: This is really interesting/tricky!!
I'm not sure if you feel like "something" isn't quite right here, I sure do.

Here's what I think it is after spending significant time thinking about it: there's too much arbitrary/specialised state and flags created. This leads to specific points where things could get strange (eg, the server crashes/reboots before you set specific flags), and it's not as clean, robust, and extensible as we'd normally like.

Here's what I think the solution is:
The `DatabaseManager` should be the one initialising the Importer, rather than the `Importer` using the `DatabaseManager` to start the import.

This is more interesting and impactful than it looks, because it basically means the "control" over importing and import state should be known by the `DatabaseManager`, and not centralised in the `Importer` - this is why you ended up with a strange `Importer` -> `Database.setImported` pattern arising, which generates an error you catch in the `DatabaseManager` later. 
This also feels 'wrong' because your import `Databases` are fundamentally still a working database, even if the import stops halfway - it's just _semantically_ not complete from the point of view of the _importer_! 

Another way of seeing this as 'strange' is that the idea of "Import" status is known in 3 places: `DatabaseManager` (`imported_databases` set), the `DatabaseImporter`, and the `Database` itself. This feels unnecessary right?

-----

Here's the architecture I would use:


1. `DatabaseManager.createDatabaseImporter(name)` 
  a. This creates a Database using the regular `Database::open` HOWEVER, you do this within a new directory called `data/import` - we may need to add logic to exclude this directory when we load databases on bootup
  --> This is important because on reboot, we know that anything under `data/import` is an incomplete database import and we can for example ask the `DatabaseImporter` if it was actually completed, or decide what else to do with it
  --> Now, the `DatabaseManager` semantically knows which databases are in the "Importing" state and shouldn't be exposed to the user yet!
  --> This is an extensible pattern, if we ever wanted to hide other databases while they're initialising in another way we just use a different folder with the same pattern
2. In the `DatabaseImporter::new(Database)`, everything continues as usual
3. On finish of the `DatabaseImporter`, close the database, and you ask `DatabaseManager::include_database(path)`, which will _move_ the `data/import/<db name>` up one level to `data/<db name>` (make sure it's closed first!)

In addition, the `DatabaseImporter` should manage all the special pieces of state (eg. the IMPORTING flag and the cache) that live inside the `data/import/<db name>`, rather than the `Database`. By doing this, import logic is completely removed from the `Database`, and only orchestrated by the `DatabaseManager` on request or on bootup, and otherwise all completely encapsulated in the `DatabaseImporter`


The one catch is that there's a piece of validation you're protecting against specifically - the 'name taken' check... and this can also be done under a race condition! I would just solve this by taking the 'databases' lock on the hashmap, even though you won't use the hashmap. We basically centralise all database creation under this lock, even if we don't put them into the map:

```
create_database(self, directory, name, record=true/false):
 - take databases lock
 - list and check the `data` directory for any valid databases with the name
 - list and check the `data/import` directory for any valid databases with the name
 - let db = Database::open(directory.join(name))
 - if record: insert db into databases hashmap & return  a copy of the arc
 - else: return an arc (though that's the only owner at this time)
```

This is a bit ugly but i'd rather have 1 ugly method than scatter specialised logic around.

If we want to be even more sophisticated, we can generalise this pattern to be `data/internal` and use `data/internal/import`, and scan this internal directory's children for valid databases. Incidentally, at some point I would put the system database in there too! But i think this might be one step more general than we need for now...

database/migration/database_importer.rs
flyingsilverfin: Yeah let's visit that separately

farost: I'll put a todo

server/service/grpc/migration/import_service.rs
farost: Sure

common/cache/cache.rs
farost: I would not want to make the library `SpilloverCache`, it feels too much for this component, but I'm not sure. At the same time, I will have to leave the file `cache.rs` if the library is still `cache`

farost: @dmitrii-ubskii wdyt?

dependencies/typedb/repositories.bzl
farost: Sure

server/service/grpc/typedb_service.rs
farost: It's generated by tonic like `type transactionStream`

database/database_manager.rs
farost: Well the exposed method is also called `create_database` and it allows it to pre-exist. I was not sure if we wanted to change the overall API. It is an old problem.

database/database_manager.rs
farost: Yeah, it sounds better from the design perspective. But the point about uniqueness (your ugly proposal) and other small details showed me that it's not that great, and I preferred the idea of the transitioned ownership. I will try playing with your design and see how far I can get with it.

The first issue I see is that I don't understand what actually DatabaseImporter owns here. In my model, it owns the database, so it controls the cleanup, it's its responsibility to govern the import consistency (or at least call all the needed parts to do their job). It lets me call the cleanup on `Drop` if anything happens, addressing `DatabaseManager` to deal with directories and inner HashMaps.

In your model, `DatabaseManager` creates the Importer. Then, I can't store the `Arc` to the `DatabaseManager` in the `DatabaseImporter`. Then, I can't call finalization things through it. Then, I move this cleanup responsibility from the Importer to the ImportService (GRPC ONLY for now).
If the service controls it, then what does it do for each of the messages?
* name + schema -> call DatabaseManager to create an Importer
* data items -> call Importer directly as it's not stored anywhere else?
* done -> call Importer, make sure it's finalized, then call DatabaseManager as we can't do it from the Importer?
* Additionally handle everything in `do_close` if we are not finished, right?

What does Importer do then? Just a layer of abstraction for methods without owning anything in reality. At least it should not be responsible for closing the database before asking the DatabaseManager to move the directory, we can move the database out of it, handing the ownership, so it will be safer. But, instead of spreading "importing" logic between `DatabaseImporter + DatabaseManager + Database` we tweaked the `DatabaseImporter` the way it became a weird zombie, and the `ImportService` ended up without a strict domain itself, too. It works with user requests, it understands when to work with the importer, but when to work with the manager as well. Otherwise, everything breaks. 

Maybe it's fine, but it is the thing that I find as "something is not right", and one of the reasons I've chosen the current design.

I'll start with finding a middle ground between my approach and your main idea. I think that the truth is somewhere in that area.

farost: There is also a problem of ownership in your design. You say that DatabaseImporter should close the database, but you also return an Arc from the DatabaseManager when you create a new database even for the Importer, which means that Importer does not know if it exclusively owns the database or not. And it cannot be dropped because we want (probably we still want) to call the cleanup logic on `Drop for DatabaseImporter`.

farost: But it's fine, just less generalization and all is good. Another issue with moving directories now is that the SpilloverCache cannot control its temporary directories on drop: currently, we create them in the database's directory. However, if we move the directory, the path will change. I will try changing it so it points to the import directory instead, and just adds the database name as a prefix... mehh

database/migration/database_importer.rs
```diff
@@ -163,10 +163,10 @@ struct DataInfo {
 }
 
 impl DataInfo {
-    fn new(data_directory: &PathBuf) -> Self {
+    fn new(data_directory: &PathBuf, database_name: &str) -> Self {
```
farost: This should use import_directory instead, it will be exposed as `pub(crate)` for DatabaseImporter to get

database/database_manager.rs
```diff
@@ -8,86 +8,140 @@ use std::{
...
 
-        Ok(Arc::new(Self {
-            data_directory: data_directory.to_owned(),
-            databases: RwLock::new(databases),
-            imported_databases: RwLock::new(HashSet::new()),
-        }))
+        Ok(())
     }
 
-    pub fn create_database(&self, name: impl AsRef<str>) -> Result<(), DatabaseCreateError> {
```
farost, line 60: I gave up. It's really hard to reason about database creations when it's called `create`. But the protocol API and the API in `State` will remain `create`, probably....

database/database_manager.rs
```diff
@@ -236,6 +287,53 @@ impl DatabaseManager {
         name.starts_with(INTERNAL_DATABASE_PREFIX)
     }
 
+    // TODO: Remove if not needed
```
farost: Will be needed, as I said above, I'll use it for the SpilloverCache target dir

flyingsilverfin (approved):

database/database_manager.rs
farost: I've stopped at a middle ground: we don't share the exporting info with the database, but we still preserve the ownership on the DatabaseImporter side. The Importer calls DatabaseManager when needed. DatabaseManager has an extra directory for imported databases, and that's the only way to differentiate between real and imported databases. We use a single write lock to ensure the uniqueness of database names, but, just in case, have a final check before actually creating a new database after its import.

PR #7479, opened 2025-05-27
Make server state object extensible

lolski:

### DIFF
--- a/server/server.rs
+++ b/server/server.rs
@@ -6,6 +6,7 @@
 
 use std::{net::SocketAddr, sync::Arc};
 
+use crate::state::ServerState;
 use crate::{
     error::ServerOpenError,
     parameters::config::{Config, EncryptionConfig},
@@ -20,7 +21,6 @@ use tokio::{
     net::lookup_host,
     sync::watch::{Receiver, Sender},
 };
-use crate::state::ServerState;
 
 pub struct Server {
     server_info: ServerInfo,
--- a/server/server.rs
+++ b/server/server.rs
@@ -25,7 +25,7 @@ use tokio::{
 pub struct Server {
     server_info: ServerInfo,
     config: Config,
-    server_state: Arc<ServerState>,
+    server_state: Arc<Box<ServerState>>,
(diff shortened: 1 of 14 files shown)


server/service/http/typedb_service.rs
```diff
@@ -358,7 +361,7 @@ impl TypeDBService {
         State(service): State<Arc<TypeDBService>>,
         accessor: Accessor,
     ) -> impl IntoResponse {
-        run_with_diagnostics(&service.server_state.diagnostics_manager, None::<&str>, ActionKind::UsersAll, || {
+        run_with_diagnostics(&service.server_state.diagnostics_manager(), None::<&str>, ActionKind::UsersAll, || {
```
farost, line 364: You introduced redundancy by `fetch_add`ing the ref counter and then taking its reference without a need (with a `clone` underneath the method)

server/server.rs
```diff
@@ -18,44 +21,63 @@ use crate::{
...
+    shutdown_channel: Option<(Sender<()>, Receiver<()>)>,
+}
+
+impl ServerBuilder {
+    pub fn server_info(mut self, server_info: ServerInfo) -> Self {
+        self.server_info = Some(server_info);
+        self
+    }
+
+    pub async fn server_state(mut self, server_state: BoxServerState) -> Self {
```
farost, line 40: Why is it async?

farost (approved):

server/state.rs
```diff
@@ -46,21 +48,67 @@ use crate::{
     parameters::config::{Config, DiagnosticsConfig},
 };
 
+pub type BoxServerState = Box<dyn ServerState + Send + Sync>;
+
+#[async_trait]
+pub trait ServerState: Debug {
+    fn databases_all(&self) -> Vec<String>;
```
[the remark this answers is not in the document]
flyingsilverfin: Yeah it looks like most of these will go over the network? At least all that do writes to the state right?

 @lolski  did we decide that all reads would also be put into the log and replicated as action intents for simplicity/ensuring consistency in the end?

PR #7492, opened 2025-06-12
Ignore importing databases with empty schemas

farost:
Ignore empty imported schemas to allow importing databases without schemas. This also unlocks the importing of empty exported databases, ensuring the symmetry of export and import operations.

## Implementation
Trim the schema string in case there are some excessive symbols to be more welcoming.

### DIFF
--- a/dependencies/typedb/repositories.bzl
+++ b/dependencies/typedb/repositories.bzl
@@ -33,9 +33,8 @@ def typedb_protocol():
     )
 
 def typedb_behaviour():
-    # TODO: Return typedb
     git_repository(
         name = "typedb_behaviour",
-        remote = "https://github.com/farost/typedb-behaviour",
-        commit = "cdba20c65814f9677aa7974fdb373055d584b831",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        remote = "https://github.com/typedb/typedb-behaviour",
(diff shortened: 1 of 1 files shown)


database/migration/database_importer.rs
```diff
@@ -333,6 +333,9 @@ impl DatabaseImporter {
     }
 
     pub async fn import_schema(&mut self, schema: String) -> Result<(), DatabaseImportError> {
+        if schema.trim().is_empty() {
```
farost, line 336: Tested against cdba20c65814f9677aa7974fdb373055d584b831

dependencies/typedb/repositories.bzl
```diff
@@ -33,8 +33,9 @@ def typedb_protocol():
     )
 
 def typedb_behaviour():
+    # TODO: Return typedb
     git_repository(
         name = "typedb_behaviour",
-        remote = "https://github.com/typedb/typedb-behaviour",
-        commit = "65258a4a3ad80be5918f33b74b1b08e25ee6fd7b",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        remote = "https://github.com/farost/typedb-behaviour",
```
farost: This does not mean anything. I will need to merge this PR and then update `typedb-driver` with a snapshot from this server.

I've tested the BDDs locally, and the change is simple, so I think you can trust me with the next steps. If there is a typo, I will find it after updating `typedb-driver` and fix it additionally.

PR #7493, opened 2025-06-13
Implement check operation in function return

krishnangovindraj:
Implement check operation in function return.

## Implementation
Introduce `StreamModifierExecutor::Check`.

### DIFF
--- a/dependencies/typedb/repositories.bzl
+++ b/dependencies/typedb/repositories.bzl
@@ -35,6 +35,6 @@ def typedb_protocol():
 def typedb_behaviour():
     git_repository(
         name = "typedb_behaviour",
-        remote = "https://github.com/krishnangovindraj/typedb-behaviour",
-        commit = "f12b9194f8a328aa5cb54108a7c5118d16fd0f9d",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        remote = "https://github.com/typedb/typedb-behaviour",
+        commit = "c94f0245084e5b5d4e0fcc9fb5a447c05b01330c",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
     )


executor/read/stream_modifier.rs
```diff
@@ -264,3 +273,47 @@ impl StreamModifierResultMapperTrait for LastMapper {
...
+impl CheckMapper {
+    pub(crate) fn new() -> Self {
+        Self { returned: None }
+    }
+}
+
+impl StreamModifierResultMapperTrait for CheckMapper {
+    fn map_output(&mut self, subquery_result: Option<FixedBatch>) -> Option<FixedBatch> {
+        if self.returned.is_some() {
+            return None;
```
farost, line 291: Why?

krishnangovindraj: Check has to return an answer once and fail everytime afterwards. 
The PatternExecutor will actually pop it off the stack when it returns None the first time so this should be unreachable, but I think I've implemented all the other mappers to behave correctly regardless of whether the PatternExecutor does it's job.

farost: Cool, I thought so. Thanks for clarifying
I don't mind **not** putting a debug assert here if we don't generally care

farost (approved):

flyingsilverfin (approved):
Thanks!

flyingsilverfin:
Is red for some reason!

executor/read/stream_modifier.rs
krishnangovindraj: I made a mistake @farost , This is actually legit logic since I need to return `Some(Value::Boolean(_))` once and `None` in the second time PatternExecutor calls it. After that the PatternExecutor will never call it again.
So the debug_assert has to go away.

farost: It's great that we understood the execution better thanks to it! =) Thanks for trying it out, although I said I don't mind

PR #7494, opened 2025-06-13
Reduce usages of assert! that don't need to be assert!

krishnangovindraj:
Allow `assert!` only in unrecoverable cases where we must crash the server.

## Implementation
To facilitate searching (Ctrl+F) for these cases, we replace usages of `assert!` in unit tests to `assert_true!` (via `use assert as assert_true`)

### DIFF
--- a/compiler/annotation/type_inference.rs
+++ b/compiler/annotation/type_inference.rs
@@ -95,6 +95,7 @@ pub mod tests {
         type_seeder::TypeGraphSeedingContext,
         TypeInferenceError,
     };
+    use assert as assert_true;
 
     #[test]
     fn test_functions() {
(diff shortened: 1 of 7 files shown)


compiler/annotation/type_inference.rs
```diff
@@ -95,6 +95,7 @@ pub mod tests {
         type_seeder::TypeGraphSeedingContext,
         TypeInferenceError,
     };
+    use assert as assert_true;
```
farost, line 98: I'm not sure `assert_true`'s intent to be used in unit tests is clear. Have you considered more explicit `test_assert` if you want this specific definition?

farost: Overall, I'm afraid that this design decision can degrade easily once we forget about it and start writing new unit tests without copypasting the older ones

krishnangovindraj: I don't mind `test_assert` if you think that's clearer.

farost: I do. But I'm still not sure if it's a sustainable solution

dmitrii-ubskii: How about `expect!`?

PR #7495, opened 2025-06-16
Simplify PartialCostHash and beam search implementation

krishnangovindraj:
We simplify the hash used to detect redundant states during the plan-search, loosening the definition of equivalent plans for more aggressive pruning.

## Implementation
We update the `PartialCostHash` to have a stronger definition of 
Since we want to choose the plan with the lowest cost amongst all plans corresponding to the same state, we convert `new_plans_heap` to a min_heap so the first one we see is the best one.

### DIFF
--- a/compiler/executable/match_/planner/plan.rs
+++ b/compiler/executable/match_/planner/plan.rs
@@ -1208,11 +1208,9 @@ impl PartialCostPlan {
 
     fn hash(&self) -> PartialCostHash {
         PartialCostHash {
-            all_vars: self.remaining_patterns.clone(),
-            ongoing_vars: self.ongoing_step.clone(),
-            approx_io: (self.cumulative_cost.io_ratio * self.ongoing_step_cost.io_ratio) as u64, // TODO: improve rounding/hashing (make relative)
-            approx_cost: self.cumulative_cost.chain(self.ongoing_step_cost).cost as u64, // TODO: improve rounding/hashing (make relative)
-            ongoing_join: self.ongoing_step_join_var.is_some(),
+            n_remaining_patterns: self.remaining_patterns.len() as u32,
+            planned_vertices: self.vertex_ordering.iter().filter_map(|v| v.as_pattern_id()).collect::<BTreeSet<_>>(),
+            ongoing_non_trivial_patterns: self.ongoing_step.iter().copied().collect::<BTreeSet<_>>(),
         }
     }
 }
@@ -1231,35 +1229,11 @@ impl Ord for PartialCostPlan {
     }
 }
 
-#[derive(Clone, PartialEq, Eq)]
+#[derive(Clone, Hash, PartialEq, Eq)]
 pub(super) struct PartialCostHash {
-    all_vars: HashSet<PatternVertexId>,
-    ongoing_vars: HashSet<PatternVertexId>,
-    approx_io: u64,
-    approx_cost: u64,
-    ongoing_join: bool,
-}
-
-impl Hash for PartialCostHash {
-    fn hash<H: Hasher>(&self, state: &mut H) {
-        let mut acc = 0;
-        for pattern in &self.all_vars {
-            let mut hasher = DefaultHasher::new();
-            pattern.hash(&mut hasher);
-            acc ^= hasher.finish();
-        }
-        acc.hash(state);
-        let mut acc = 0;
-        for pattern in &self.ongoing_vars {
-            let mut hasher = DefaultHasher::new();
-            pattern.hash(&mut hasher);
-            acc ^= hasher.finish();
-        }
-        acc.hash(state);
-        self.approx_io.hash(state);
-        self.approx_cost.hash(state);
-        self.ongoing_join.hash(state);
-    }
+    n_remaining_patterns: u32, // Needed for continuous search (A*), but not step-based (beam)
+    planned_vertices: BTreeSet<PatternVertexId>,
+    ongoing_non_trivial_patterns: BTreeSet<PatternVertexId>,
 }
 
 #[derive(Clone, PartialEq, Debug)]
--- a/compiler/executable/match_/planner/plan.rs
+++ b/compiler/executable/match_/planner/plan.rs
@@ -12,6 +12,7 @@ use std::{
     hash::{DefaultHasher, Hash, Hasher},
     sync::Arc,
 };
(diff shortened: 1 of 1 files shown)


compiler/executable/match_/planner/plan.rs
```diff
@@ -810,6 +758,28 @@ impl<'a> ConjunctionPlanBuilder<'a> {
...
+        self.heap.pop()
+    }
+}
+
+impl<'a, T: Ord> Drop for DrainSorted<'a, T> {
+    fn drop(&mut self) {
+        self.heap.clear();
+    }
+}
+
```
krishnangovindraj, line 781: [Nightly](https://doc.rust-lang.org/std/collections/binary_heap/struct.DrainSorted.html) but i wanted it because it makes things nice and readable.

compiler/executable/match_/planner/plan.rs
```diff
@@ -693,87 +696,32 @@ impl<'a> ConjunctionPlanBuilder<'a> {
...
                         break;
-                    }
-
-                    if extension_heap.len() < extension_width {
-                        extension_heap.push(extension);
-                    } else if let Some(top) = extension_heap.peek() {
-                        if extension < *top {
-                            extension_heap.pop();
-                            extension_heap.push(extension);
-                        }
```
krishnangovindraj, line 727: I could keep this way of doing things, but figured using two unbounded size min-heaps is less confusing than one unbounded-min and one bounded-max.
The old approach is strictly more efficient, because populating this would be `w * log(w)` and the heap is `n*log(n)`. 
w (width) is set to be `num_patterns/2`, whereas the number of extensions is likely to be `num_patterns`.

compiler/executable/match_/planner/plan.rs
```diff
@@ -1208,11 +1218,9 @@ impl PartialCostPlan {
...
     fn hash(&self) -> PartialCostHash {
         PartialCostHash {
-            all_vars: self.remaining_patterns.clone(),
-            ongoing_vars: self.ongoing_step.clone(),
-            approx_io: (self.cumulative_cost.io_ratio * self.ongoing_step_cost.io_ratio) as u64, // TODO: improve rounding/hashing (make relative)
-            approx_cost: self.cumulative_cost.chain(self.ongoing_step_cost).cost as u64, // TODO: improve rounding/hashing (make relative)
-            ongoing_join: self.ongoing_step_join_var.is_some(),
+            n_remaining_patterns: self.remaining_patterns.len() as u32,
+            planned_vertices: self.vertex_ordering.iter().filter_map(|v| v.as_pattern_id()).collect::<BTreeSet<_>>(),
+            ongoing_non_trivial_patterns: self.ongoing_step.iter().copied().collect::<BTreeSet<_>>(),
```
krishnangovindraj, line 1222: This needs to be checked for correctness.

compiler/executable/match_/planner/plan.rs
```diff
@@ -669,18 +669,20 @@ impl<'a> ConjunctionPlanBuilder<'a> {
...
 
         let mut extension_heap = BinaryHeap::with_capacity(extension_width); // reused
+        let mut new_plans_heap = BinaryHeap::with_capacity(beam_width);
+        let mut new_plans_hashset = HashSet::with_capacity(beam_width);
         for i in 0..num_patterns {
             event!(Level::TRACE, "{INDENT:4}PLANNER STEP {}", i);
-            let mut new_plans_heap = BinaryHeap::with_capacity(beam_width);
-            let mut new_plans_hashset = HashSet::with_capacity(beam_width);
 
+            // TODO: Do we need this?
```
dmitrii-ubskii, line 677: Not necessarily. In fact probably not, if we don't need to artificially restrict the search space near the end of planning to speed up.

compiler/executable/match_/planner/plan.rs
dmitrii-ubskii: The fact that the extension heap grows without bounds makes me a little uneasy. The number of extensions _per prefix_ is ~ N, and the beam width is tied to N (although capped), so we're looking at an O(N^2) space extension heap here.

compiler/executable/match_/planner/plan.rs
dmitrii-ubskii: Is `n_remaining_patterns` meaningful? Surely `planned_patterns` + `ongoing_non_trivial_patterns` already account for all the information, and the stash doesn't really do anything by definition.
I also wonder if we can just lump them all together into `planned_patterns`. `ongoing_step_join_var` may disambiguate enough on its own.

dmitrii-ubskii (commented):
Overall this looks sound. Left a couple comments.

compiler/executable/match_/planner/plan.rs
krishnangovindraj: In a beam search, it's irrelevant because we reset the hash-set. 
In a search that doesn't proceed in steps, we need it so we can actually add things to the stash without it being instantly ruled out by the hash.
I think an A* will be helpful to evaluate our heuristic so Idm the num_patterns being in there.

compiler/executable/match_/planner/plan.rs
krishnangovindraj: Good point. I can revert this if space is a concern. I don't think we have much of a time-cost because we generate the extensions regardless.

compiler/executable/match_/planner/plan.rs
krishnangovindraj: Yeah I think I'll remove it.
It's true that the most plans are in the middle of the search and there are less at the end, but I don't think this makes a big difference to the bigger picture

compiler/executable/match_/planner/plan.rs
krishnangovindraj: @dmitrii-ubskii To clarify: We drain the extension heap after each plan. So it's still limited to `O(N)` space.
The new plans heap will have a size of `O(beam_width * extension_width)`  overhead. This is indeed `O(N^2)`. 

I've also realised why I don't want to use the old approach of capping the size of `new_plans_heap` to `beam_width` at all times because:
1. We can't check the hash before inserting to `new_plans_heap` as they're not added in order of cost.
2. Since we check the hash when draining the `new_plans_heap`, we may rule out candidates and be left with less than `beam_width` candidates in the next step.

PR #7497, opened 2025-06-17
Update the lower bound for output size of a step to be a small fraction instead of 1

krishnangovindraj:
Updates the lower bound that the planner uses for estimating the (average) output size of a step to be a small fraction instead of 1 - allowing it to better model highly selective constraints. Clamping the value to 1.0 prevented the planner from using its otherwise accurate estimates .

## Implementation
Update `MIN_SCAN_SIZE` to `0.000000001`, the value used for `Cost::MIN_IO_RATIO`.

### DIFF
--- a/compiler/executable/match_/planner/vertex/constraint.rs
+++ b/compiler/executable/match_/planner/vertex/constraint.rs
@@ -31,7 +31,7 @@ use crate::{
     },
 };
 
-const MIN_SCAN_SIZE: f64 = 1.0;
+const MIN_SCAN_SIZE: f64 = 0.000000001;
 const MAX_SCAN_SIZE: f64 = 10e15;
 
 #[derive(Clone, Debug)]


krishnangovindraj:
This change commits to the assumption that data is randomly distributed. The downfall of not clamping the MIN_SCAN_SIZE  to 1 is that we may decay the `io_ratio` of the resulting partial plan to a very small number, wrongly discounting the costs of any future patterns. Dmitrii provides a great demonstrative example
```
match $x has $n; // Assume each $x has a $n, this *produces* the n edges
match $x has $n; // This is now a check step, which makes a random distribution assumption, and decides it has an io_ratio of 1/N [= N / (N * N) ]
```

imo (and purely my opinion), assuming data is randomly distributed is not a bad assumption in itself, but a more realistic/safer assumption may be that data is randomly distributed within communities and there are multiple communities within the global dataset.

compiler/executable/match_/planner/vertex/constraint.rs
```diff
@@ -31,7 +31,7 @@ use crate::{
     },
 };
 
-const MIN_SCAN_SIZE: f64 = 1.0;
+const MIN_SCAN_SIZE: f64 = 0.000000001;
```
dmitrii-ubskii, line 34:
```suggestion
const MIN_SCAN_SIZE: f64 = 1e-9;
```

dmitrii-ubskii: This is very small. Have you run TPC-C with this change?

krishnangovindraj: It's copied verbatim from `MIN_IO_RATIO`. The fact that it's `1e-9` is pretty irrelevant. I just need it to be small.
Idk if I want to update it to scientific notation without updating the other one too

krishnangovindraj: Also, it looks like there's a difference in the running time of `STOCK-LEVEL`. I might have to look into it

krishnangovindraj:
Reverting this to draft since there is a performance drop in the Stock level query in TPCC.

krishnangovindraj:
P sure Dmitrii found a better fix for this.

PR #7498, opened 2025-06-19
Fix how we decide the join variable of consecutive steps during planner

krishnangovindraj:
Fixes how the join variable (if any) between consecutive steps in a candidate plan is determined.

## Implementation
Simplifies the `determine_joinability` and `join_from_direction_and_inputs` functions

### DIFF
--- a/compiler/executable/match_/planner/plan.rs
+++ b/compiler/executable/match_/planner/plan.rs
@@ -984,33 +984,24 @@ impl PartialCostPlan {
     }
 
     fn determine_joinability(&self, graph: &Graph<'_>, pattern: PatternVertexId) -> Option<VariableVertexId> {
+        // If there is no ongoing step, there is no join
         let &prev_pattern = self.ongoing_step.iter().next()?;
+
         // We only join constraint patterns, so let's extract constraints
-        let prev_planner = &graph.elements[&VertexId::Pattern(prev_pattern)];
-        let PlannerVertex::Constraint(prev_constraint) = prev_planner else { return None };
-        let planner = &graph.elements[&VertexId::Pattern(pattern)];
-        let PlannerVertex::Constraint(constraint) = planner else { return None };
-        // Determine whether there are any candidate join variables:
-        let candidate_join_var = constraint
-            .variables()
-            .filter(|var| self.ongoing_step_produced_vars.contains(var) && constraint.can_join_on(*var))
-            .exactly_one()
-            .ok()?;
-        // Only direct-able patterns are join-able:
+        let prev_constraint = &graph.elements[&VertexId::Pattern(prev_pattern)].as_constraint()?;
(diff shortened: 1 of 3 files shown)


compiler/executable/match_/planner/plan.rs
```diff
@@ -984,33 +984,24 @@ impl PartialCostPlan {
...
+        let possible_join_var = match &self.ongoing_step_join_var {
+            Some(existing_join_var) => *existing_join_var,
+            None => prev_constraint.join_from_direction_and_inputs(
+                prev_dir,
+                &self.ongoing_step_produced_vars,
+                &self.all_produced_vars,
+            )?
+        };
+        (constraint.variables().contains(&possible_join_var) && constraint.can_join_on(possible_join_var))
+        .then(|| possible_join_var)
```
krishnangovindraj, line 1004: This seems like an equivalent, simpler formulation. Please verify

compiler/executable/match_/planner/vertex/constraint.rs
```diff
@@ -92,7 +92,7 @@ impl ConstraintVertex<'_> {
             Self::Links(_) | Self::Has(_) | Self::IndexedRelation(_) => {
                 let unbound_join_variables: Vec<VariableVertexId> = self
                     .variables()
-                    .filter(|&var| self.can_join_on(var) && (!exclude.contains(&var) || include.contains(&var)))
+                    .filter(|&var| self.can_join_on(var) && (include.contains(&var) && !exclude.contains(&var)))
```
krishnangovindraj, line 95: Was this the right condition?

krishnangovindraj (commented):
There's some loss of information and weird re-derivation of the iterate mode from the input variables, join variables and CostMetadata. ~~Refactoring all of them into one informative struct feels like the ideal solution~~

compiler/executable/match_/planner/vertex/constraint.rs
```diff
@@ -133,7 +133,7 @@ impl ConstraintVertex<'_> {
             Self::Links(_) | Self::Has(_) | Self::IndexedRelation(_) => {
                 let unbound_join_variables: Vec<VariableVertexId> = self
                     .variables()
-                    .filter(|&var| self.can_join_on(var) && (!exclude.contains(&var) || include.contains(&var)))
+                    .filter(|&var| self.can_join_on(var) && (!exclude.contains(&var) && include.contains(&var)))
```
krishnangovindraj, line 136: Replaced the `||` with the `&&` because that felt more correct, but now that I think about it, it's probably not the case. The || Will additionally include steps that are not in the vertex_ordering regardless of whether they are in the ongoing step.

Another question I have is: Does this work for more than 2 constraints being joined?

compiler/executable/match_/planner/plan.rs
krishnangovindraj: I'm going to revert this file

PR #7499, opened 2025-06-24
Implement optionality

flyingsilverfin:


compiler/annotation/fetch.rs
```diff
@@ -355,7 +355,7 @@ fn annotate_sub_fetch(
     input_value_type_annotations: &BTreeMap<Variable, ExpressionValueType>,
 ) -> Result<AnnotatedFetchListSubFetch, AnnotationError> {
     let FetchListSubFetch { context, input_variables, stages, fetch } = sub_fetch;
-    let TranslationContext { mut variable_registry, .. } = context;
+    let PipelineTranslationContext { mut variable_registry, .. } = context;
```
flyingsilverfin, line 358: Rename - this is specifically for translating a pipeline in order of the stages!

compiler/executable/match_/instructions/mod.rs
```diff
@@ -620,6 +620,9 @@ pub enum CheckInstruction<ID> {
         role2: ID,
         player2: ID,
     },
+    VariablesNotNone {
+        variables: Vec<ID>,
```
flyingsilverfin: I wonder if there's a better way to represent this check. Should we also have each 'not none' check contain only a single variable?

compiler/executable/match_/planner/mod.rs
```diff
@@ -276,9 +300,11 @@ impl StepBuilder {
 }
 
 #[derive(Debug)]
-struct MatchExecutableBuilder {
+struct ConjunctionExecutableBuilder {
     selected_variables: Vec<Variable>,
     input_variables: Vec<Variable>,
+    constraint_variables: HashSet<Variable>,
```
flyingsilverfin, line 306: new field used for optional filtering when lowering: all variables mentioned in constraints in this conjunction, excluding the ones nested in nested patterns and inputs

compiler/executable/match_/planner/plan.rs
```diff
@@ -97,7 +99,7 @@ pub(crate) fn plan_conjunction<'a>(
         conjunction,
         block_context,
         variable_positions,
-        shared_variables,
+        selected_variables,
```
flyingsilverfin, line 102: consistent renaming

compiler/executable/match_/planner/plan.rs
```diff
@@ -151,27 +154,57 @@ fn make_builder<'a>(
...
-                shared_variables.extend(negation.required_inputs(block_context));
-                shared_variables =
-                    shared_variables.intersection(&negation.referenced_variables().collect()).copied().collect();
+                // let mut negation_selected_vars = selected_variables.clone();
+                // negation_selected_vars.extend(negation.referenced_variables(block_context));
+                // negation_selected_vars =
+                //     negation_selected_vars.intersection(&negation.referenced_variables().collect()).copied().collect();
+                let negation_selected_vars = negation.conjunction().referenced_variables()
+                    .filter(|var| block_context.is_in_scope_or_parent(conjunction.scope_id(), *var))
+                    .collect();
```
flyingsilverfin: After renaming to 'selected variables', i realised that starting from only selected variables, and then extending to required inputs (eg, arguments or expression variables in the negation) can still miss certain inputs - such as those bound above but aren't required but also aren't selected? (such as a disjunction branch's local variables?)

compiler/executable/match_/planner/vertex/mod.rs
```diff
@@ -585,6 +590,45 @@ impl Costed for NegationPlanner<'_> {
     }
 }
 
+#[derive(Clone, Debug)]
```
krishnangovindraj, line 593: I think optionals can always be scheduled at the end. Maybe we should just do that?
If refactoring the planner is imminent, I guess it can wait and just live in a comment.

compiler/annotation/match_inference.rs
```diff
@@ -281,6 +282,9 @@ impl TypeInferenceGraph<'_> {
         for nested_graph in &mut self.nested_disjunctions {
             nested_graph.prune_self_from_vertices(&self.vertices)
         }
+        for nested_graph in &mut self.nested_optionals {
+            nested_graph.prune_constraints_from_vertices()
+        }
```
krishnangovindraj, line 287: Optionals shouldn't be involved here since they can't affect the annotations of the parent.
They should be the in the (aptly named) `prune_types_for_nested_negations_and_optionals`

compiler/annotation/match_inference.rs
```diff
@@ -291,12 +295,14 @@ impl TypeInferenceGraph<'_> {
         for nested_graph in &mut self.nested_disjunctions {
             is_modified |= nested_graph.prune_vertices_from_self(&mut self.vertices);
         }
+        for nested_graph in &mut self.nested_optionals {
+            is_modified |= nested_graph.prune_vertices_from_constraints();
+        }
```
krishnangovindraj, line 300: Same as above. This shouldn't be here. Just in `prune_types_for_nested_negations_and_optionals`

compiler/annotation/type_seeder.rs
```diff
@@ -190,7 +190,7 @@ impl<'this, Snapshot: ReadableSnapshot> TypeGraphSeedingContext<'this, Snapshot>
             disjunction.conjunctions().iter().map(|conj| self.build_recursive(context, conj)).collect_vec();
         let shared_variables = disjunction
             .referenced_variables()
-            .filter(|var| context.is_variable_available(parent_conjunction.scope_id(), *var))
+            .filter(|var| context.is_in_scope_or_parent(parent_conjunction.scope_id(), *var))
```
krishnangovindraj, line 193: I don't think these kind of renames help. Yes, sure it's checking whether the variable is in the current scope or parent. If I wanted to know what this line is doing, I could always get that information from clicking through. What I'm not getting from the new name is why it's doing it, and I'm left guessing. 
I do actually have trouble following the logic when we rename things like this. It's exposing the implementation details and lowering the level of abstraction.
I think it's true that a variable is available iff it's declaration is considered to be in this or a parent scope, so I prefer the old name. 
I felt the same about Dmitrii's use of scope "transparency". It makes for an easier read because it conveys the relevance of the function, rather than summarising it's implementation.

compiler/annotation/type_seeder.rs
```diff
@@ -317,6 +320,9 @@ impl<'this, Snapshot: ReadableSnapshot> TypeGraphSeedingContext<'this, Snapshot>
                     any |= self.annotate_some_unannotated_vertex(nested_graph, context)?;
                 }
             }
+            for optional in &mut graph.nested_optionals {
+                any |= self.annotate_some_unannotated_vertex(optional, context)?;
+            }
```
krishnangovindraj, line 325: This should not be here. 
Optionals cannot affect the parent's annotations, and should be handled as negations are in `seed_types_in_nested_negations_and_optionals`.

compiler/annotation/type_seeder.rs
```diff
@@ -368,6 +374,11 @@ impl<'this, Snapshot: ReadableSnapshot> TypeGraphSeedingContext<'this, Snapshot>
             is_modified |= self.reconcile_nested_disjunction(nested, &mut graph.vertices)?;
         }
 
+        // Propagate from nested disjunctions
+        for nested in &mut graph.nested_optionals {
+            is_modified |= self.reconcile_nested_optional(nested, &mut graph.vertices)?;
+        }
+
```
krishnangovindraj, line 381: Same as above. use `seed_types_in_nested_negations_and_optionals` instead.

compiler/annotation/type_seeder.rs
```diff
@@ -485,6 +496,32 @@ impl<'this, Snapshot: ReadableSnapshot> TypeGraphSeedingContext<'this, Snapshot>
         Ok(something_changed)
     }
 
+    fn reconcile_nested_optional(
```
krishnangovindraj, line 499: Should go away. Parent does not need to be reconciled.

krishnangovindraj (changes requested):
Comments on type-inference.

ir/pattern/mod.rs
```diff
@@ -354,84 +354,91 @@ impl fmt::Display for ValueType {
...
+impl BitAndAssign for BindingMode {
     fn bitand_assign(&mut self, rhs: Self) {
         match (*self, rhs) {
-            (Self::Producing, _) | (_, Self::Producing) => *self = Self::Producing,
-            (Self::Required, _) | (_, Self::Required) => *self = Self::Required,
-            (Self::Referencing, Self::Referencing) => (),
+            (Self::AlwaysBinding, _) | (_, Self::AlwaysBinding) => *self = Self::AlwaysBinding,
+            (Self::NonBinding, _) | (_, Self::NonBinding) => *self = Self::NonBinding,
+            (Self::LocallyBindingInChild, _) | (_, Self::LocallyBindingInChild) => *self = Self::LocallyBindingInChild,
+            (Self::OptionallyBinding, Self::OptionallyBinding) => (),
```
krishnangovindraj, line 370: NonBinding isn't very descriptive. Isn't it always `Required` if it's non-binding? 
Do we need the `OptionallyBinding` mode? In the `None` world, doesn't an optional block always bind a value to either or to `None` or some other `VariableValue`?

In the current setting, Do we need `LocallyBindingInChild`? Is there a case that isn't covered by it being `AlwaysBinding` in the child and either 
* `AlwaysBinding` in the parent, if it's `AlwaysBinding` in all the children; or
*  `Required` in the parent if it's not present in all the children?

ir/translation/match_.rs
```diff
@@ -35,12 +36,7 @@ pub(crate) fn add_patterns(
         typeql::Pattern::Conjunction(nested) => add_patterns(function_index, conjunction, &nested.patterns),
         typeql::Pattern::Disjunction(disjunction) => add_disjunction(function_index, conjunction, disjunction),
         typeql::Pattern::Negation(negation) => add_negation(function_index, conjunction, negation),
-        typeql::Pattern::Optional(optional) => {
-            // add_optional(function_index, conjunction, optional)
-            Err(Box::new(RepresentationError::UnimplementedLanguageFeature {
-                feature: error::UnimplementedFeature::Optionals,
```
krishnangovindraj, line 41: You should delete this Enum variant now. That'll give you compile errors for all the usages so you're sure not to have missed any.

ir/translation/match_.rs
```diff
@@ -70,9 +66,21 @@ fn add_negation(
...
+    add_patterns(function_index, &mut optional_builder, &optional.patterns)?;
+    let ConjunctionBuilder {
+        conjunction,
+        context,
+    } = optional_builder;
+    for var in conjunction.referenced_variables() {
+        // if the variable is available in the parent scope, it's bound externally and passed in so not optional
+        if !context.is_variable_in_scope_or_parent(parent_scope, var) {
+            context.set_variable_optionality(var, true)
+        }
```
krishnangovindraj, line 83: Is this sound for cases where you first write the optional block and then the usage?

executor/batch.rs
```diff
@@ -60,7 +60,18 @@ impl FixedBatch {
...
             width,
-            data: vec![VariableValue::Empty; size as usize],
+            data: vec![VariableValue::None; size as usize],
+            entries: 0,
+            multiplicities: FixedBatch::INIT_MULTIPLICITIES,
+            provenance: FixedBatch::INIT_PROVENANCES,
+        }
+    }
+
+    pub(crate) fn new_with_rows(width: u32, rows: u32) -> Self {
```
krishnangovindraj, line 70: Unused + redundant because we can just use new?

executor/read/pattern_executor.rs
```diff
@@ -266,7 +287,13 @@ impl PatternExecutor {
                     )
                 }
             }
+            StepExecutors::Optional(OptionalExecutor { inner, .. }) => {
+                inner.prepare(FixedBatch::from(input.as_reference()));
+                self.control_stack.push(ExecuteOptional { index, input: input.into_owned(), any_found: false }.into());
+            }
             StepExecutors::Negation(NegationExecutor { inner }) => {
+                // TODO: question: why do we send the input row into the Inner, but also stash it in the ExecuteNegation control instruction?
+                //     Is it so we can restore the executor to the same state somehow?
```
krishnangovindraj, line 296: A negation succeeds when the `inner` PatternExecutor finds no answer. `inner` does not track the input row, so we need it to be accessible when the PatternExecutor finds no answer

compiler/executable/match_/planner/mod.rs
```diff
@@ -477,24 +505,51 @@ impl MatchExecutableBuilder {
...
+                selected_variables: Vec::from_iter(self.current_outputs.iter().copied()),
+                builder: StepInstructionsBuilder::Check(CheckBuilder::default()),
+            };
+            builder.builder.as_check_mut().unwrap().instructions.push(
+                CheckInstruction::NotNone { variables: optional_inputs_in_constraints },
+            );
+            steps.push(builder.finish(&self.index, &named_variables, variable_registry));
+        }
+
+        steps.extend(self
```
krishnangovindraj, line 529: Can't we avoid the second Vec by doing self.steps.insert(0, check_step)

Also, could we somehow combine this and may_make_input_check_step in plan.rs?
They're both pre-checks to ensure input variables are valid (here, that they're non-optional and there that they're valid types)

compiler/executable/match_/planner/plan.rs
```diff
@@ -633,12 +650,20 @@ impl<'a> ConjunctionPlanBuilder<'a> {
         }
     }
 
+    // TODO: this is quite confusing: why do we have a fully formed NegationPlan, and then wrap it back into a NegationPlanner??
```
krishnangovindraj, line 653: The `register_*` method probably exists for consistency. The (parent) `ConjunctionPlanBuilder` still has to schedule the negation somewhere. The reason it's a fully formed plan is probably because (1)  the cost of the negation is needed for the parent ConjunctionPlanBuilder to schedule it and (2) it is known beforehand, 

Dmitrii's suggestion of renaming the `*Planner` to `*Vertex` is probably the best solution to this.

compiler/executable/match_/planner/plan.rs
```diff
@@ -1352,29 +1378,29 @@ impl ConjunctionPlan<'_> {
...
                         }
                     }
                     for output in self.outputs_of_pattern(pattern) {
                         let is_selected =
-                            || match_builder.selected_variables.contains(&self.graph.index_to_variable[&output]);
-                        let has_consumers = || self.consumers_of_var(output).next().is_some();
-                        if is_selected() || has_consumers() {
-                            match_builder.finish_one();
-                            match_builder.register_output(self.graph.index_to_variable[&output]);
+                             conjunction_builder.selected_variables.contains(&self.graph.index_to_variable[&output]);
```
krishnangovindraj, line 1387: I don't mind Dmitrii's approach of making them functions. It does save a call when the `or` short-circuits. I don't think the compiler can do the same if it can't guarantee that the `has_consumers` call has no side-effects.

ir/pipeline/block.rs
```diff
@@ -314,9 +335,12 @@ pub enum VariableLocality {
...
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub(crate) enum ScopeTransparency {
-    Transparent,
-    Opaque,
+pub(crate) enum ScopeType {
+    Input,
+    Conjunction,
+    Disjunction,
+    Negation,
+    Optional,
```
krishnangovindraj, line 343: These don't seem to behave in more ways than just `Opaque` and `Transparent`? If not, I don't think the rename is justified. Transparency is a perfectly easy concept to grasp and this is just adding 5 concepts where we need 2. And the new names don't convey on what this is used for. 
So unless we have a functional reason to make this change, I'd revert.

compiler/annotation/match_inference.rs
flyingsilverfin: Thanks, was going to ask questions about all these - removed from here

ir/translation/match_.rs
flyingsilverfin: Thanks!

compiler/executable/match_/planner/plan.rs
flyingsilverfin: I just prefer not having them as function for debugging. This isn't performance critical code so that's fine

compiler/annotation/match_inference.rs
flyingsilverfin: Where is the step where we prune inside of an Optional if there is no Parent?
```
match try { $x, $y, $z..... $a};
```
?

compiler/executable/match_/planner/vertex/mod.rs
flyingsilverfin: Yeah that's what i was thinking

compiler/annotation/match_inference.rs
flyingsilverfin: Ok well for some reason when i delete all the things you noted I shouldn't need, i get 
```
expected thing variable to have been annotated with types
```
So i'm just going to keep them in for now, perhaps you can look into that when you get to it the writes

ir/pipeline/block.rs
flyingsilverfin: Funny enough now 'Transparency' has gone away as a concept entirely

compiler/annotation/type_seeder.rs
flyingsilverfin: Yeah i've actually moved it back to approximately the old name at the end of my refactoring

compiler/executable/match_/planner/plan.rs
flyingsilverfin: Yeah i've addressed the naming inconsistencies elsewhere

ir/pattern/mod.rs
flyingsilverfin: The need for `LocallyBindingInChild` came from the updated 'disjointVariable` check:
```
    pub(crate) fn find_disjoint_variable(&self, block_context: &BlockContext) -> ControlFlow<(Variable, Option<Span>)> {
        for (var, mode) in self.variable_binding_modes() {
            let scope = block_context.get_declaring_scope(&var).unwrap();
            // variables present in sibling scopes are "declared" in their common ancestor
            // variables are only considered locally binding in child if the variable is locally bound in all children it is present in
            //   because locally-binding loses to both non-binding and binding modes
            // therefore: fail if the variable is only bound in >=1 child, but "declared" here
            if scope == self.scope_id && (mode.is_locally_binding_in_child() || mode.is_optionally_binding()) {
                return ControlFlow::Break((var, mode.referencing_constraints().first().and_then(|c| c.source_span())));
            }
        }
        for nested in &self.nested_patterns {
            nested.find_disjoint_variable(block_context)?;
        }
        ControlFlow::Continue(())
    }
    ```

ir/translation/match_.rs
flyingsilverfin: Good point! Lifted up!

compiler/executable/match_/planner/mod.rs
flyingsilverfin: Excellent yes i didn't know we had that - will add it there.

ir/translation/match_.rs
flyingsilverfin: So this made me realise that I haven't done function return optionals and assignment optionals, but im ok to leave that for now as a different enum name

flyingsilverfin:
@krishnangovindraj  we can keep commenting on this one, though I will now close it, and continue with the other PR that includes the moderate planner refactoring

compiler/annotation/match_inference.rs
krishnangovindraj: No pls. I'd rather fix it in your branch than risk missing it. It should be in  `prune_types_for_nested_negations_and_optionals`. It does it for its own negations and optionals, followed by the negations & optionals in itself.

Now that I think about it, there are indeed values returned from optional blocks that need to be available in the parent. I'd just add it in `prune_types_for_nested_negations_and_optionals`. Copy out the annotations of just the optional variables in the optional block.

ir/pattern/mod.rs
krishnangovindraj: Ah nice. This might also allow us to detect unplannable patterns now, rather than failing planning.

ir/pipeline/block.rs
krishnangovindraj: Interesting. Do we need scopes at all then?

krishnangovindraj:
@flyingsilverfin , Can you try to split them into two PRs? If the refactor can happen independent of optionals, it would be a lot easier to review + debug if needed in the future.

PR #7501, opened 2025-06-27
Refactor planner to track join variable in cost metadata

krishnangovindraj:
Updates the metadata tracked by the planner to include the variable which each pattern is sorted on, and can be joined on. Though this can be derived from the direction & bound variables, the code is much simpler when we store it and avoids the planner & lowering independently re-deriving it.

## Implementation
* Updates CostMetaData::Direction to include the join variable alongside the direction.
* Simplifies the code to determine whether two steps can be joined.

### DIFF
--- a/compiler/executable/match_/planner/plan.rs
+++ b/compiler/executable/match_/planner/plan.rs
@@ -918,14 +918,13 @@ impl PartialCostPlan {
                 }
             })
             .flat_map(move |&extension| match self.determine_joinability(graph, extension) {
-                None => vec![(extension, None)].into_iter(),
-                Some(var) => vec![(extension, None), (extension, Some(var))].into_iter(),
+                None => vec![(extension, None)],
+                Some(var) => vec![(extension, None), (extension, Some(var))],
             })
             .map(move |(extension, join_var_opt)| {
                 let planner_vertex = &graph.elements[&VertexId::Pattern(extension)];
                 let (cost_before_extension, cost_of_extension, metadata) = if let Some(join_var) = join_var_opt {
                     // Step continues
-                    debug_assert!(matches!(planner_vertex, PlannerVertex::Constraint(_)));
                     let PlannerVertex::Constraint(constraint) = planner_vertex else {
                         unreachable!("Cannot join unless constraint");
                     };
@@ -1667,9 +1666,10 @@ impl ConjunctionPlan<'_> {
                 };
 
                 let direction = if matches!(inputs, Inputs::None([])) {
-                    let CostMetaData::Direction(unbound_direction, _) = metadata else {
+                    let CostMetaData::Direction(unbound_direction, _sort_var) = metadata else {
                         unreachable!("expected metadata for constraint")
                     };
+                    debug_assert!(sort_variable.is_none() || sort_variable.as_ref().map(|v| self.graph.variable_index[v]) == _sort_var);
                     unbound_direction
                 } else if rhs_var.is_some_and(|rhs| inputs.contains(rhs)) {
                     Direction::Reverse
--- a/compiler/executable/match_/planner/plan.rs
+++ b/compiler/executable/match_/planner/plan.rs
@@ -633,7 +633,8 @@ impl<'a> ConjunctionPlanBuilder<'a> {
 
     pub(super) fn plan(self) -> Result<ConjunctionPlan<'a>, QueryPlanningError> {
         let search_patterns: HashSet<_> = self.graph.pattern_to_variable.keys().copied().collect();
-        let input_variables = self.graph
+        let input_variables = self
+            .graph
             .variable_index
             .values()
             .copied()
@@ -655,7 +656,6 @@ impl<'a> ConjunctionPlanBuilder<'a> {
             metadata
         );
         let element_to_order = ordering.iter().copied().enumerate().map(|(order, index)| (index, order)).collect();
-
         let Self { graph, local_annotations: type_annotations, mut planner_statistics, .. } = self;
 
         planner_statistics.finalize(cost);
@@ -679,7 +679,8 @@ impl<'a> ConjunctionPlanBuilder<'a> {
     // We record directionality information for each pattern in the plan, indicating which prefix index to use for pattern retrieval
 
     fn beam_search_plan(
-        graph: &Graph<'_>, initial_empty_plan: PartialCostPlan
+        graph: &Graph<'_>,
+        initial_empty_plan: PartialCostPlan,
     ) -> Result<(Vec<VertexId>, HashMap<PatternVertexId, CostMetaData>, Cost), QueryPlanningError> {
         const INDENT: &str = "";
         const BEAM_REDUCTION_CYCLE: usize = 2;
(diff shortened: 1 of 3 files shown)


flyingsilverfin (approved):
Looks ok to me but definitely let dmitrii look at it before merging!

compiler/executable/match_/planner/plan.rs
```diff
@@ -941,42 +949,47 @@ impl PartialCostPlan {
...
-                let join_var = self.determine_joinability(graph, extension);
-
-                if join_var.is_none() {
-                    vec![(extension, join_var)].into_iter()
-                } else {
-                    vec![(extension, None), (extension, join_var)].into_iter()
-                }
+            .flat_map(move |&extension| match self.determine_joinability(graph, extension) {
+                None => vec![(extension, None)].into_iter(),
+                Some(var) => vec![(extension, None), (extension, Some(var))].into_iter(),
```
dmitrii-ubskii: One more step: `.flat_map` accepts `IntoIterator`
```suggestion
                None => vec![(extension, None)],
                Some(var) => vec![(extension, None), (extension, Some(var))],
```

compiler/executable/match_/planner/plan.rs
```diff
@@ -941,42 +949,47 @@ impl PartialCostPlan {
...
-                let meta_data: CostMetaData;
-
-                if join_var.is_none() {
-                    (added_cost, meta_data) =
-                        self.compute_added_cost(graph, extension, &all_available_vars, join_var)?;
+            .map(move |(extension, join_var_opt)| {
+                let planner_vertex = &graph.elements[&VertexId::Pattern(extension)];
+                let (cost_before_extension, cost_of_extension, metadata) = if let Some(join_var) = join_var_opt {
+                    // Step continues
+                    debug_assert!(matches!(planner_vertex, PlannerVertex::Constraint(_)));
```
dmitrii-ubskii: This assert doubles the check below with no additional information. Either remove or add a message with the failing constraint.

compiler/executable/match_/planner/plan.rs
```diff
@@ -1676,7 +1653,7 @@ impl ConjunctionPlan<'_> {
                 };
 
                 let direction = if matches!(inputs, Inputs::None([])) {
-                    let CostMetaData::Direction(unbound_direction) = metadata else {
+                    let CostMetaData::Direction(unbound_direction, _) = metadata else {
```
dmitrii-ubskii: We could verify here that the metadata sort var agrees with the input `sort_variable`.

compiler/executable/match_/planner/vertex/constraint.rs
```diff
@@ -124,32 +124,34 @@ impl ConstraintVertex<'_> {
 
     pub(crate) fn direction_from_join_var(
         &self,
-        var: VariableVertexId,
-        include: &HashSet<VariableVertexId>,
-        exclude: &HashSet<VariableVertexId>,
-    ) -> Option<Direction> {
+        join_var: VariableVertexId,
+        _ongoing_step_produced: &HashSet<VariableVertexId>,
```
dmitrii-ubskii, line 128: This is clearly used.
```suggestion
        ongoing_step_produced: &HashSet<VariableVertexId>,
```

compiler/executable/match_/planner/vertex/constraint.rs
```diff
@@ -124,32 +124,34 @@ impl ConstraintVertex<'_> {
...
-        }
+        let (canonical_from, canonical_to) = match self {
+            Self::Links(links) => (links.relation, links.player),
+            Self::Has(has) => (has.owner, has.attribute),
+            Self::IndexedRelation(indexed) => (indexed.player_1, indexed.player_2),
+            _ => return Err(QueryPlanningError::UnimplementedJoinForConstraint {} ),
+        };
+        // It can't be a join var if it's being newly produced in this step
+        let direction = if join_var == canonical_from {
+            debug_assert!(all_produced.contains(&canonical_to) || !_ongoing_step_produced.contains(&canonical_to));
```
dmitrii-ubskii, line 147: If `all_produced` is a superset of `ongoing_step_produced`, this is a tautology. Should that have been `&&` as a negation of the filter above?

compiler/executable/match_/planner/vertex/constraint.rs
```diff
@@ -124,32 +124,34 @@ impl ConstraintVertex<'_> {
...
+            if all_produced.contains(&canonical_to) { Direction::Reverse } else { Direction::Canonical }
+        } else {
+            debug_assert!(join_var == canonical_to);
+            debug_assert!(all_produced.contains(&canonical_from) || !_ongoing_step_produced.contains(&canonical_from));
```
dmitrii-ubskii, line 151: Same here.

compiler/executable/match_/planner/vertex/constraint.rs
```diff
@@ -1194,7 +1204,7 @@ impl Costed for SubPlanner<'_> {
         _: Option<Direction>,
         _: &Graph<'_>,
     ) -> Result<(Cost, CostMetaData), QueryPlanningError> {
-        Ok((Cost::in_mem_complex_with_ratio(1.0), CostMetaData::Direction(Direction::Reverse)))
+        Ok((Cost::in_mem_complex_with_ratio(1.0), CostMetaData::Direction(Direction::Reverse, None)))
```
dmitrii-ubskii, line 1208: The "TODO" comment from `isa` should be replicated in all these.

compiler/executable/match_/planner/vertex/constraint.rs
```diff
@@ -1311,6 +1321,36 @@ impl Costed for PlaysPlanner<'_> {
...
+            Direction::Reverse => {
+                if is_to_bound {
+                    from_var
+                } else {
+                    to_var
+                }
+            }
+        };
+        Some(join_var)
     }
```
dmitrii-ubskii, line 1347: Optionally reduce rightward drift:
```suggestion
    if is_from_bound && is_to_bound {
        return None;
    }
    match direction {
        Direction::Canonical if is_from_bound => Some(to_var),
        Direction::Canonical => Some(from_var),
        Direction::Reverse if is_to_bound => Some(from_var),
        Direction::Reverse => Some(to_var),
    }
```

dmitrii-ubskii (approved):
Left some mostly cosmetic comments

compiler/executable/match_/planner/vertex/constraint.rs
krishnangovindraj: It's only used in debug_asserts. Does that count if we have them configured off?

compiler/executable/match_/planner/vertex/constraint.rs
krishnangovindraj: Aren't all asserts meant to be tautologies? I'm updating the comment. The comment I've left is wrong though. I'm updating it - could you check if it makes sense?

compiler/executable/match_/planner/plan.rs
krishnangovindraj: Done. Though the sort_variable in the argument is really a join_variable because there's an `is_join.then(var)` in the caller. Shall I rename that?

compiler/executable/match_/planner/vertex/constraint.rs
dmitrii-ubskii: Is this assert meant to verify that `ongoing_step_produced` _is_ a subset of `all_produced`? There's a better way to phrase that.

Otherwise, this is saying that either `canonical_to` is in `all_produced`, or not in `ongoing_step_produced` (same as "not in `all_produced`").

compiler/executable/match_/planner/vertex/constraint.rs
dmitrii-ubskii: Oh I didn't notice the `filter` bit was also moved into a `debug_assert`. Never mind.

compiler/executable/match_/planner/plan.rs
dmitrii-ubskii: Might as well, if you think "join of size one" isn't confusing. I'd say it's not.

PR #7505, opened 2025-06-30
Enable RocksDB Prefix bloom filters

krishnangovindraj:
Enable [RocksDB bloom filters](https://github.com/facebook/rocksdb/wiki/RocksDB-Bloom-Filter), reducing avoidable disk-reads in large databases.

## Implementation
* Splits the IteratorPool into two: One for iterators which use a prefix, and one for those which do not .
  - The prefix-iterator sets `options.set_total_order_seek(false)`
  - If the limits of the range being iterated have the same prefix (of the length specified by the keyspace), then we use the prefix iterators

To verify that the filters are working, turn on statistics and check the `rocksdb.[non.]last.level.seek.filtered COUNT`

### DIFF
--- a/encoding/encoding.rs
+++ b/encoding/encoding.rs
@@ -102,6 +102,7 @@ impl KeyspaceSet for EncodingKeyspace {
         options.set_max_write_buffer_size_to_maintain(0);
         options.set_max_write_buffer_number(2);
         options.set_memtable_whole_key_filtering(false);
+        options.set_optimize_filters_for_hits(false); // true => don't build bloom filters for the last level
         options.set_compression_per_level(&[
             DBCompressionType::None,
             DBCompressionType::None,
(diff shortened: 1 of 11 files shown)


encoding/encoding.rs
```diff
@@ -102,6 +102,7 @@ impl KeyspaceSet for EncodingKeyspace {
         options.set_max_write_buffer_size_to_maintain(0);
         options.set_max_write_buffer_number(2);
         options.set_memtable_whole_key_filtering(false);
+        options.set_optimize_filters_for_hits(false); // true => don't build bloom filters for the last level
```
krishnangovindraj, line 105: Do we want bloom filters on the last level? The rationale behind this option is that there's a very high chance (in large databases) that the last level contains a key chosen at random.

storage/iterator.rs
```diff
@@ -9,6 +9,7 @@ use std::{cmp::Ordering, error::Error, fmt, sync::Arc};
 use bytes::byte_array::ByteArray;
 use lending_iterator::{LendingIterator, Peekable, Seekable};
 use resource::profile::StorageCounters;
+use tracing::log::Log;
```
flyingsilverfin: unused?

storage/keyspace/iterator.rs
```diff
@@ -97,6 +94,29 @@ impl KeyspaceRangeIterator {
...
+        if start.length() < prefix_length {
+            return false;
+        };
+        match range.end() {
+            RangeEnd::WithinStartAsPrefix => true,
+            RangeEnd::EndPrefixInclusive(end) => {
+                end.length() >= prefix_length && end[0..prefix_length] == start[0..prefix_length]
+            }
+            RangeEnd::EndPrefixExclusive(end) => {
+                end.length() >= prefix_length && end[0..prefix_length] == start[0..prefix_length]
```
flyingsilverfin, line 116: these can be `RangeEnd::EndInclusive(end) | RangeEnd::EndExclusive(end)` right so we dont duplicate the comparision logic?

flyingsilverfin (approved):

storage/keyspace/iterator.rs
krishnangovindraj: I mean to flag this in a self-review but forgot to publish.
Yes, but I don't know how we use EndPrefixExclusive and whether I should decrement the end by one.

storage/keyspace/iterator.rs
```diff
@@ -106,25 +127,34 @@ impl LendingIterator for KeyspaceRangeIterator {
...
+            self.is_finished = true;
         }
+        item
     }
 }
 
 impl Seekable<[u8]> for KeyspaceRangeIterator {
     fn seek(&mut self, key: &[u8]) {
-        self.iterator.seek(key);
+        if !self.is_finished {
```
krishnangovindraj, line 155: Consistency with the `DBIterator`.

Issue #7475, opened 2025-05-21
Schema query crashes server "assertion failed: !self.intersection_value.is_empty()"


krishnangovindraj:
## Description
Simple schema query causes server crash:
> thread 'tokio-runtime-worker' panicked at executor/read/immediate_executor.rs:461:9:
assertion failed: !self.intersection_value.is_empty()

## Environment

1. TypeDB distribution: Core
2. TypeDB version: 3.3
3. Environment: Mac
4. Client and version: Studio 3.3
5. Other details: 

## Reproducible Steps

1. Set up
Load up the bookstore schema.

3. Test/Query
```
match 
{ $t sub! $supertype; } or {$t sub $supertype; $t is $supertype; };
{ $t owns $attr; not { $t sub! $sown; $sown owns $attr; }; } or { $_ owns $attr ; };
```

4. Unexpected result
```
thread 'tokio-runtime-worker' panicked at executor/read/immediate_executor.rs:461:9:
assertion failed: !self.intersection_value.is_empty()
```

## Expected result
A beautiful graph

krishnangovindraj:
```
match 
{ $t sub! $supertype; } or {$t sub $supertype; $t is $supertype; };
{ $t owns $attr; not { $t sub! $sown;  }; } or { $_ owns $attr ; };
```
yields the plan:
```
Match executable plan [selected=[p3, p1, p0], output_width=4]:
  0: Disjunction [selected=[p1, p0], output_size=2]
      --- Start branch ---Match executable plan [selected=[p1, p0], output_width=2]:
  0: Sorted Iterator Intersection [bound_vars=[], selected=[p1, p0], output_size=2, sort_by=p1]
      [p1 owns p0] filter [] with (outputs=p0, p1, )
  1: Negation
      --- Start negation ---
 Match executable plan [selected=[p0, p1], output_width=2]:
  0: Check [selected=[p0, p1], output_width=2]
      Check[p1 type ([EntityType:[0001]], [EntityType:[0002]], [EntityType:[0003]], [EntityType:[0006]], [EntityType:[0007]], [EntityType:[0010]], [EntityType:[0011]], [EntityType:[0012]], )] 
  1: Sorted Iterator Intersection [bound_vars=[p1], selected=[p0, p1], output_size=2, sort_by=__$3__]
      [p1 sub __$3__] filter [] with (inputs=p1, counts=__$3__, )
      --- End negation ---
      --- End branch ---
[...]

PR #7508, opened 2025-06-30
Implement optionality in read queries

flyingsilverfin:

We implement a huge new feature: optionality in read queries! This means you can now do partial matches. 

### Basics 
In the following query, all `person` instances will be returned, and if any name exists, the name will be added to the answer. However, _not_ having an answer will _not_ cause that person instance to be eliminated:
```
# Find any person, and if they have a name add that into the answer
match 
  $p isa person;
  try { $p has name $name; };
```
The answers will conceptually conform to a type like `[ $p=person, $name=Option<name>]`.

Optionals only return when the _full_ pattern inside the optional matches:
```
# Return any people, and use the ages only of people with name "john", else do not match anything additional
match 
  $p isa person;
  try { $p has name "john"; $p has age $age; };
```

Multiple `try` clauses are matched to the maximum degree at simultaneously for each non-optional answer component:
```
# If every person can have at most 1 name and 1 age: return every person, plus their name if it exists, plus their age if it exists
match
  $p isa person;
  try { $p has name $name; };
  try { $p has age $age; };
```
Note that this will produce 1 answer per person, not a permutation of each person, with and without their names and ages. If there are higher cardinalities of name or age, you get permutations of each possible name and age filled but neither filled.

Optional variables cannot be re-used across sibling clauses due to ambiguous outcomes:
```
### ILLEGAL - optional $name cannot be re-used across sibling try clauses
match 
  $p isa person;
  try { $p has name $name; }; 
  try { $x isa person, has name $name; };
```

### Nesting and Pipelining

Optionals can also not be nested within disjunctions (`or`) or negations (`not`) - in negations, these operations would be no-ops, and disjunctions are currently limited to only returning variables that are guaranteed to be returned from any branch, ie. they cannot currently return optional values at all.

Optional variables _can_ be used across stages of a query pipeline. Any pattern in a query using an optional variable that is not set will "fail" immediately (return 0 answers). However, by nesting a pattern containing an unset optional variable inside try` clause, this behaviour is absorbed since only the `try` clause will fail:

```
match
  try { $name isa name "John"; };
match
  try { $p isa person, has name $name; };

```
This will return exactly `[$name=None, $p=None ]` or `[$name=Some(name), $p=Some(person) ]` or `[$name=Some(name), $p=None]`

In contrast, by using the optional `$name` _outside_ any optional scope, the following query will eliminate any answers that don't have `$name` set and only return answers containing `[$name=name, $person=person]`
```
match
  try { $name isa name "John"; };
match
  $p isa person, has name $name;
```

## Implementation

We've made several architectural changes to how we think about the structures present in the `compiler` package.

### Centralising "scoping semantics" into the BlockContext and Patterns

Attempted to move towards the `BlockContext` having enough information on its own in order to have all the information required to determine variable visibility, binding, and legality (eg. no sharing variables between sibling optionals!)

This was not entirely completed, though we made two changes that contribute towards this.

* `Transparency` to `ScopeType`
`Transparency` was used to help determine if a variable was "visible" through a nested scope or not - which depended on whether it was a Disjunction, Negation, or Conjunction. However, this was not general enough to consider Optionals as well.

Instead, we now just record the type of scope directly (Disjunction, Negation, Conjunction, or Optional) into the BlockContext, and let the user of the Context figure out the visibility rules based on this. In theory, since the BlockContext now has full information about scope parent-child combinations, and also the type of the scope, we are almost at the state at which we could determine all semantic information about variables, visibility, and binding requirements. What is missing is the information to determine if a variable must be an input (eg. bound elsewhere) or if it can be bound in the pattern scope. 

This information is currently only visible directly in Patterns, which takes us the the second change:

* `VariableDependency` to `VariableBindingMode`
Dependency used to indicate if a variable was producing, required input, or referencing. This strategy did not really extend to optionals and was generally confusing. It's been replaced with a trait that produces a "Binding mode" for each variable in a Pattern - including recursively defined or referenced variables in child patterns! These are 'context free', in that information about the scopes that are parents to a scope are irrelevant.

The Binding mode provides full information about each variable having to be `PreBound` (eg same as required input before - expression arguments, function arguments, etc.), `AlwaysBinding` (say bound in every branch of a disjunction, or if this is a Conjunction), `OptionallyBinding` (for optionals that have local variables that are also visible to the parents), and `LocallyBindingInChild` (for variables in less than all branches of a disjunction, or variables locally used only in a negation).

A variable's binding mode can be different when queried at any level of a pattern - in a `Negation`, which just contains a `Conjunction`, all variables are probably `AlwaysBinding`. However, at the Negation's level they are probably mostly `LocallyBindingInChild`, etc. 

To help use this information, we have a new trait on all `NestedPattern` which is called `Pattern`, that exposes the raw context-free `VariableBindingMode` for each variable, but also a set of APIs that are not context-free and utilise the BlockContext to see what parents are available from the parents - for example, `required_inputs(blockContext)`, `visible_binding_variables(blockContext)`, and `named_visible_binding_variables(blockContext)`.

### Simplifications to Planning

We eliminate some redundancy in the planner functions by noticing that scoped pattern planning graphs are build recursively, and already have all the required information about what is a required input to a particular scope's planner. There previously was confusion generated by setting variables as "inputs" when they were already given as "shared variables" during construction as well.

We also streamling naming throughout the planner, and settle on calling the representations of "sub-plans" in a planning graph "xxxVertex", which is how they are used and perceived by the parent planner.

#### Futuer Improvements
One future improvement of the planner should be achieved by noting that the "Graph" terminology is not very explicit. There is the idea of a "Vertex", but no idea of an "edge" - if we had adopted this naming, we'd notice that there is redundancy in a Graph, which stores both VariableVertex -> PatternVertex edges and PatternVertex -> VariableVertex edges, but at the same time each Pattern can also generate the set of `VariableVertex` again.

A simpler model would also explicitly write out the bipartite graph:
```
Graph {
  variables:
  patterns:
  variables_to_patterns:
  patterns_to_variables:
}
```
And then possible decide to either keep `patterns_to_variables` as the only way to traverse from patterns to variables, or only keep a `PatternVertex.variables()` as the only way to traverse from patterns to variables.


flyingsilverfin:
@krishnangovindraj  you can continue here if you want, but still want to comment on the PR and finish my self-review anyhow

krishnangovindraj:
General description comments:
> Note that this will produce 1 answer per person, not a permutation of each person, with and without their names and ages.

This can be more precise, because it will produce `m * n` if there are m & n answers for name & age.

---
> Optionals can also not be nested within disjunctions (or) or negations (not) as these operations would be no-ops.

For negations, sure. For disjunctions they can return optional values, can't they? I'd give this as motivating example, but maybe there's another way to write this query:
```
$member isa person; $insti isa institute;
{ $emp isa employment, links ($member, $insti); try {  $emp has last-date $end;  }; } or 
{ $stship isa studentship, links ($member, $insti); try {  $stship has graduation-date $end;  }; };
```

-- 
>  Any pattern in a query using an optional variable will "fail" immediately (return 0 answers).

Seems to contradict the example where 
> the following query will eliminate any answers that don't have $name set

compiler/executable/match_/planner/vertex/mod.rs
```diff
@@ -550,31 +559,32 @@ impl Costed for UnsatisfiablePlanner<'_> {
 }
 
 #[derive(Clone, Debug)]
-pub(super) struct NegationPlanner<'a> {
+pub(super) struct NestedNegationPlan<'a> {
```
krishnangovindraj: Nested doesn't add much information to the name honestly. Our top-level is always a conjunction so these are always nested. It might fit your way of seeing things better, but I think the original naming fits the better way of seeing things better.

krishnangovindraj: This isn't a plan. It just happens to contain a plan.
it's called `NegationPlanner` because it belongs to the same class of concepts as all the other `*Planner`s. 
Dmitrii's suggestion was to rename all the `*Planner` to `*Vertex`. 

I'd actually like to see this encoded in the type system by having every vertex implement a `PlannerVertex` trait.

compiler/executable/match_/planner/vertex/mod.rs
```diff
@@ -550,31 +559,32 @@ impl Costed for UnsatisfiablePlanner<'_> {
...
-    fn is_valid(&self, ordered: &[VertexId], _graph: &Graph<'_>) -> bool {
-        self.variables().all(|var| ordered.contains(&VertexId::Variable(var)))
+    fn is_valid(&self, ordered: &[VertexId]) -> bool {
+        self.referenced_parent_vertex_ids.iter().all(|var| ordered.contains(&VertexId::Variable(*var)))
     }
 
-    pub(crate) fn variables(&self) -> impl Iterator<Item = VariableVertexId> + '_ {
-        self.shared_variables.iter().copied()
+    pub(crate) fn referenced_parent_vertex_ids(&self) -> impl Iterator<Item = VariableVertexId> + '_ {
+        self.referenced_parent_vertex_ids.iter().copied()
```
krishnangovindraj, line 584: I don't think we need to rename this honestly. The structs is called `*Planner` and the method is called `variables` because you're supposed to think about them the same way. This might be the more obvious name to someone who reads the `ir` package and passes through the compiler package but it's detrimental to someone who has to understand how the planner works.

krishnangovindraj (changes requested):
Publish a first set of feedback on the planner renamings.

executor/read/nested_pattern_executor.rs
```diff
@@ -51,6 +51,50 @@ impl DisjunctionExecutor {
...
+    }
+
+    pub(crate) fn map_as_failed_output(&self, unmapped_input: MaybeOwnedRow<'_>) -> FixedBatch {
+        let mut output = FixedBatch::new(self.output_width);
+        output.append(|mut output_row| {
+            output_row
+                .copy_mapped(unmapped_input.as_reference(), self.selected_variables.iter().map(|&pos| (pos, pos)));
+            output_row.set_provenance(unmapped_input.provenance()); // Pass through old provenance
+        });
+        output
```
flyingsilverfin, line 94: This this right?

flyingsilverfin:
Responded to your comments improving the PR description @krishnangovindraj

flyingsilverfin:
Annoyingly these still fail: https://github.com/typedb/typedb/issues/7475#issuecomment-2970293515

krishnangovindraj:
> Annoyingly these still fail: https://github.com/typedb/typedb/issues/7475#issuecomment-2970293515
Good point. The `LocallyBindingInChild` mode should probably be considered a required input to the `DIsjunctionPlanner`. This is an assumption we currently make but could revise. The catch is doing so will need VariableValue::(Empty and None) to be distinct.

All these interacting changes makes me wonder whether this is indeed worth the effort of splitting this into multiple PRs.

The changes are all a step in the right direction and "feel" correct, but these are things we've struggled with and changed 
Apart from optionals the other two components are parts that (I'm familiar with, but) you're the authority on. We know the planner, scoping & variable dependency/binding modes have suffered from having to adapt to us expanding the language so any work that goes into making adapting them better to the current state of the language is welcome. I think the VariableDependencyMode change is a good step in that direction, but I think there's some interaction between them + VariableValue::(Empty -> None) rename that might need a bigger language discussion. 
As such, I'm a bit uncomfortable with merging all these changes as one fat PR. Majorly because it makes it difficult to isolate them should we change our assumptions.

1. `is_valid` of the planner depends on the new variable modes.
- This means I may have to review them for correctness, which we already struggle with.
- Though we know the existing ones are not correct ([7475](https://github.com/typedb/typedb/issues/7475))
2. Are we happy to commit to the idea that we will never want to schedule a disjunction where only some of the branches bind a variable before a constraint that surely binds it. If not, I think we'll need `VariableValue::None` and `VariableValue::Empty` to be distinct. Idk whether it's cleaner to unify them now and split them up when we need to, or to keep them separate for now.


I'd be a lot more comfortable if these were separate PRs which could be reviewed and committed separately.

compiler/executable/match_/planner/plan.rs
```diff
@@ -110,34 +109,31 @@ pub(crate) fn plan_conjunction<'a>(
...
     statistics: &'a Statistics,
     call_cost_provider: &impl FunctionCallCostProvider,
 ) -> Result<ConjunctionPlanBuilder<'a>, QueryPlanningError> {
     let mut negation_subplans = Vec::new();
+    let mut optional_subplans = Vec::new();
     let mut disjunction_planners = Vec::new();
     for pattern in conjunction.nested_patterns() {
         match pattern {
             NestedPattern::Disjunction(disjunction) => {
+                let shared_variables = disjunction.named_always_binding_variables(block_context).collect();
```
krishnangovindraj: Since this has changed and ends up never being read, maybe delete?

compiler/executable/match_/planner/vertex/mod.rs
```diff
@@ -586,37 +596,96 @@ impl Costed for NegationPlanner<'_> {
...
+            .branches()
+            .iter()
+            .flat_map(|branch| branch.referenced_variables().filter_map(|v| parent_variable_index.get(&v).copied()))
+            .collect();
+        let required_parent_vertex_ids = builder
+            .branches()
+            .iter()
+            .flat_map(|branch| branch.required_input_variables())
+            .map(|v| parent_variable_index[&v])
+            .collect();
```
krishnangovindraj, line 665: For [#7475](https://github.com/typedb/typedb/issues/7475), This would have to include the variables which are only referenced in some of the branches.

```
let required_inputs = disjunction.variable_binding_modes().iter()
    .filter(|(v,mode)| mode.is_locally_binding_in_child() && block_context.is_variable_available_in(parent_conjunction.scope_id(), **v)).map(|(v,_)| *v)
    .chain(disjunction.required_inputs(block_context)).collect::<HashSet<_>>();

```

executor/conjunction_executor.rs
```diff
@@ -16,32 +18,34 @@ use crate::{
...
     row::MaybeOwnedRow,
     ExecutionInterrupt,
 };
 
-pub struct MatchExecutor {
+pub struct ConjunctionExecutor {
     entry: PatternExecutor,
     input: Option<MaybeOwnedRow<'static>>,
     tabled_functions: TabledFunctions,
 }
```
krishnangovindraj: This can actually remain a `MatchExecutor`, because that's what it is. PatternExecutor is what executes a conjunction. Or it can be absorbed completely into the `MatchStageExecutor`

ir/pattern/expression.rs
```diff
@@ -63,7 +63,7 @@ impl<ID: IrID> ExpressionTree<ID> {
         self.preorder_tree.len() - 1
     }
 
-    pub fn variables(&self) -> impl Iterator<Item = ID> + '_ {
+    pub fn ids(&self) -> impl Iterator<Item = ID> + '_ {
```
krishnangovindraj: Consider `argument_ids` ?

compiler/executable/match_/planner/plan.rs
flyingsilverfin: Done!

compiler/executable/match_/planner/vertex/mod.rs
flyingsilverfin: Do you think you can roll this into your takeover of the work?

krishnangovindraj (commented):
Publish pending comments

executor/conjunction_executor.rs
flyingsilverfin: nie thanks

krishnangovindraj:
Can we update the implementation section of the description to outline the changes made to Scope, DependencyMode and the motivation in refactoring the planner?

compiler/executable/match_/planner/vertex/mod.rs
```diff
@@ -588,37 +601,96 @@ impl Costed for NegationPlanner<'_> {
 }
 
 #[derive(Clone, Debug)]
-pub(super) struct DisjunctionPlanner<'a> {
-    input_variables: Vec<VariableVertexId>,
-    shared_variables: HashSet<VariableVertexId>,
+pub(super) struct OptionalVertex<'a> {
+    pub(super) plan: OptionalPlan<'a>,
+    pub(super) referenced_parent_vertex_ids: HashSet<VariableVertexId>,
+    pub(super) optional_vertex_ids: HashSet<VariableVertexId>,
```
krishnangovindraj, line 607: I don't think this field is needed. There's no reason for the parent to be aware of these variables.

compiler/executable/match_/planner/vertex/mod.rs
```diff
@@ -588,37 +601,96 @@ impl Costed for NegationPlanner<'_> {
...
     pub(super) fn from_builder(
         builder: DisjunctionPlanBuilder<'a>,
-        variable_index: &HashMap<Variable, VariableVertexId>,
+        parent_variable_index: &HashMap<Variable, VariableVertexId>,
     ) -> Self {
-        let shared_variables: HashSet<_> =
-            builder.branches().iter().flat_map(|pb| pb.shared_variables()).map(|v| variable_index[v]).collect();
-        let input_variables = builder.required_inputs().iter().map(|v| variable_index[v]).collect();
-        Self { input_variables, shared_variables, builder }
+        // note: not every referenced variable is in the parent index & mapped
```
krishnangovindraj, line 654: When is this the case? 
When it's in all branches but not used anywhere else in the parent conjunction?

ir/pattern/mod.rs
```diff
@@ -353,29 +373,33 @@ impl fmt::Display for ValueType {
...
 impl BitAndAssign for BindingMode {
     fn bitand_assign(&mut self, rhs: Self) {
         match (*self, rhs) {
-            (Self::Producing, _) | (_, Self::Producing) => *self = Self::Producing,
-            (Self::Required, _) | (_, Self::Required) => *self = Self::Required,
-            (Self::Referencing, Self::Referencing) => (),
+            (Self::AlwaysBinding, _) | (_, Self::AlwaysBinding) => *self = Self::AlwaysBinding,
+            (Self::RequirePrebound, _) | (_, Self::RequirePrebound) => *self = Self::RequirePrebound,
+            (Self::LocallyBindingInChild, _) | (_, Self::LocallyBindingInChild) => *self = Self::LocallyBindingInChild,
+            (Self::OptionallyBinding, Self::OptionallyBinding) => (),
```
krishnangovindraj, line 391: ~~Can you explain this? Anything alongside an optional should make it RequiredPrebound, shouldn't it?~~

krishnangovindraj (commented):
some more

compiler/executable/match_/planner/vertex/mod.rs
flyingsilverfin: I'm not sure, i think the parent being aware of "optional variables" just as if they're non-optional variables is ok for planning purposes (unless we remove optionals from planning entirely and lump them at the end, but that's a different thing)

compiler/executable/match_/planner/vertex/mod.rs
flyingsilverfin: The inverse: some referenced variables are local to only some branches and therefore never in the parents' index.

ir/pattern/optional.rs
```diff
@@ -38,24 +39,51 @@ impl Optional {
...
+    pub fn named_visible_binding_variables(&self, block_context: &BlockContext) -> impl Iterator<Item = Variable> + '_ {
+        self.visible_binding_variables(block_context).filter(Variable::is_named)
+    }
+
+    fn visible_binding_variables(&self, block_context: &BlockContext) -> impl Iterator<Item = Variable> + '_ {
+        self.variable_binding_modes()
+            .into_iter()
+            .filter_map(|(v, mode)| (mode.is_always_binding() || mode.is_optionally_binding()).then_some(v))
+    }
+}
```
krishnangovindraj, line 59: Slightly sus name given it has optionally_binding in it too.

compiler/executable/match_/planner/mod.rs
```diff
@@ -484,14 +512,13 @@ impl MatchExecutableBuilder {
...
             .collect();
-        let steps = self
-            .steps
-            .into_iter()
-            .map(|builder| builder.finish(&self.index, &named_variables, variable_registry))
-            .collect();
+        let mut finished_steps = Vec::with_capacity(self.steps.len() + 1);
+        finished_steps.extend(
+            self.steps.into_iter().map(|builder| builder.finish(&self.index, &named_variables, variable_registry)),
+        );
```
krishnangovindraj: I guess this can be reverted now

executor/read/pattern_executor.rs
```diff
@@ -383,6 +407,12 @@ fn restore_suspension(
                     control_stack
                         .push(ExecuteDisjunctionBranch { index, branch_index, input: input_row.into_owned() }.into())
                 }
+                StepExecutors::Optional(optional) => {
+                    optional.inner.prepare_to_restore_from_suspension(nested_pattern_depth);
+                    // TODO: verify correct? ***How do we know if any where found??***
+                    //  ---> Can we only be resuming an optional if we already found an answer before?
+                    control_stack.push(ExecuteOptional { index, input: input_row.into_owned(), any_found: true }.into())
+                }
```
krishnangovindraj, line 415: Oh boy... can we have cycles through optionals? I think we might just want to ban that.

ir/translation/match_.rs
```diff
@@ -35,14 +35,26 @@ pub(crate) fn add_patterns(
...
+            _ => None,
+        })
+        .for_each(|optional| {
+            for var in optional.conjunction().referenced_variables() {
+                // if the variable is available in the parent scope, it's bound externally and passed in so not optional
+                if !conjunction.context.is_variable_available_in(conjunction.conjunction.scope_id(), var) {
+                    conjunction.context.set_variable_optionality(var, true)
+                }
+            }
+        });
```
krishnangovindraj, line 57: Does this work for all cases?
We do it after all the constraints for this conjunction, but can we guarantee that a parent conjunction won't be adding this later?

Something like:
```
match 
{ $y isa person;  try { friend ($x, $z); friend($z, $y);  } or { $y isa page;  friend($x, $z); follows($z, $y); }
$x isa person, has id 5;
```

ir/translation/mod.rs
```diff
@@ -106,3 +106,5 @@ macro_rules! verify_variable_available {
     };
 }
 pub(super) use verify_variable_available;
+
+use crate::pattern::variable_category::VariableOptionality;
```
krishnangovindraj: odd place for this

executor/read/nested_pattern_executor.rs
krishnangovindraj: As long as the parent & optional use the same variable positions, sure 
Though it's unclear if the input row will be wide enough to contain all those positions.

compiler/executable/match_/planner/plan.rs
```diff
@@ -726,7 +701,7 @@ impl<'a> ConjunctionPlanBuilder<'a> {
         }
 
         let best_plan =
-            best_partial_plans.into_iter().min().ok_or(QueryPlanningError::ExpectedPlannableConjunction {})?;
+            best_partial_plans.into_iter().min().ok_or_else(|| QueryPlanningError::ExpectedPlannableConjunction {})?;
```
krishnangovindraj: Not much of a difference here. Isn't it just a simple struct?

ir/pattern/negation.rs
```diff
@@ -41,30 +41,39 @@ impl Negation {
...
             .filter_map(|(var, mut mode)| {
-                let status = block_context.variable_status_in_scope(var, self.scope_id());
-                if status == VariableLocality::Parent || mode.is_required() {
-                    mode.set_required();
-                    Some((var, mode))
-                } else {
-                    None
+                if mode.is_always_binding() {
+                    // if it is binding, we demote it to only locally binding (only relevant in the negation)
+                    mode.set_locally_binding_in_child();
```
krishnangovindraj, line 70: ~~This makes me a little uncomfortable, but I don't know why.~~
Ok, I just need to expand what it means for it to be locally binding in child.

krishnangovindraj (commented):
Some more comments, but I can work on them after the merge.

ir/pattern/mod.rs
krishnangovindraj: No, AlwaysBinding & OptionallyBinding just means it's an input to the optional.
I thought it was re-use of an optional variable in a sibling.
So optional & optional should indeed be illegal

ir/translation/match_.rs
flyingsilverfin: In this case we can, because it's add_patterns and we can expect to add all the patterns from TypeQL at once yes - but really the architecture here isn't perfect in that we should return an immutable, finished IR at the end instead of continuing to have the builder to prevent this.

executor/read/pattern_executor.rs
flyingsilverfin: Yeah i'd be ok with that

PR #7509, opened 2025-07-02
Type inference of negations & optionals is no longer interleaved with that of the conjunction

krishnangovindraj:
Since negated & optional patterns cannot influence the type-annotations of the containing conjunction, we need not interleave their type-inference  with that of the parent conjunction. Hence, We refactor the type-inference procedure so that type-inference on a conjunction & its disjunctions is completed before we begin the process on negated & optional subpatterns. 

## Implementation
* Rip out `nested_negations` & `nested_optionals` from `TypeInferenceGraph`, the type-seeding process, and the pruning process.
* For brevity,  disjunctions means a root conjunction and all its nested disjunctions (recursively). negations means any negation (or optional) in the root conjunction or any of its nested disjunctions.  negations may contain more disjunctions.
  - infer_types used to do seeding, pruning & validation of disjunctions & negations interleaved.
  - Now infer_types is first done on all the disjunctions, followed by the negations.
  - This makes the logic easier to follow & makes it obvious that negations don't influence disjunctions.

### DIFF
--- a/compiler/annotation/match_inference.rs
+++ b/compiler/annotation/match_inference.rs
@@ -22,7 +22,7 @@ use ir::{
         VariableRegistry,
     },
 };
-use itertools::{chain, Itertools};
+use itertools::Itertools;
 use storage::snapshot::ReadableSnapshot;
 
 use crate::annotation::{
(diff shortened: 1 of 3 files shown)


krishnangovindraj:
Close in favour of #7512

PR #7510, opened 2025-07-03
Handle trailing partial writes in WAL

dmitrii-ubskii:

We implement the ability to discard partially written trailing WAL records.


compiler/query_structure.rs
dmitrii-ubskii, line 1: fmt

ir/pattern/disjunction.rs
dmitrii-ubskii, line 1: fmt

tests/behaviour/query/language/mod.rs
dmitrii-ubskii, line 1: fmt

database/database.rs
```diff
@@ -204,26 +200,20 @@ impl<D> Database<D> {
...
-                        // fulfill exactly 1 awaiting schema request
-                        *has_schema_transaction = true;
-                        break;
-                    }
-                    None => {}
+                if notifier.send(()).is_ok() {
+                    // fulfill exactly 1 awaiting schema request
+                    *has_schema_transaction = true;
+                    break;
                 }
```
dmitrii-ubskii, line 207: Minor cleanup here, no behaviour changes.

storage/recovery/checkpoint.rs
```diff
@@ -150,7 +150,13 @@ impl Checkpoint {
...
 
         trace!("Finished recovering keyspaces, recovering missing commits");
-        let recovery_start = self.read_sequence_number()? + 1;
+
+        let checkpoint_sequence_number = self.read_sequence_number()?;
+        if checkpoint_sequence_number > durability_client.previous() {
+            panic!("The checkpoint is ahead of the durability service! The durability logs may have been corrupted. Aborting.");
+        }
+
+        let recovery_start = checkpoint_sequence_number + 1;
```
dmitrii-ubskii, line 159: Detect if the checkpoint is ahead of the WAL. This should never happen and may indicate an out-of-date or corrupted log, and we can't really know if the db itself is in a consistent state either.

durability/tests/basic.rs
```diff
@@ -19,11 +26,79 @@ fn basic() {
...
     drop(wal);
 
     let wal = load_wal(&directory);
     let raw_record = wal.iter_any_from(written_entry_id).unwrap().next().unwrap().unwrap();
     let read_record = TestRecord::new(Vec::from(raw_record.bytes));
     assert_eq!(read_record, message);
 }
+
+#[test]
+fn added_zeros() {
```
dmitrii-ubskii, line 38: This can happen if the WAL file is resized and padded with zeros before a record is written.

durability/tests/basic.rs
```diff
@@ -19,11 +26,79 @@ fn basic() {
...
+    let mut wal_iterator = wal.iter_any_from(written_entry_id).unwrap();
+    let raw_record = wal_iterator.next().unwrap().unwrap();
+    let read_record = TestRecord::new(Vec::from(raw_record.bytes));
+    assert_eq!(read_record, message);
+    assert!(wal_iterator.next().is_none());
+    assert_eq!(fs::metadata(wal_file).unwrap().len(), len)
+}
+
+#[test]
+fn added_junk() {
```
dmitrii-ubskii, line 66: This can happen if the WAL file is resized and the tail of it contains some filesystem junk before a record is written.

durability/tests/basic.rs
```diff
@@ -19,11 +26,79 @@ fn basic() {
...
+    let mut wal_iterator = wal.iter_any_from(written_entry_id).unwrap();
+    let raw_record = wal_iterator.next().unwrap().unwrap();
+    let read_record = TestRecord::new(Vec::from(raw_record.bytes));
+    assert_eq!(read_record, message);
+    assert!(wal_iterator.next().is_none());
+    assert_eq!(fs::metadata(wal_file).unwrap().len(), len)
+}
+
+#[test]
+fn corrupted_record() {
```
dmitrii-ubskii: What it says on the tin.

durability/tests/basic.rs
```diff
@@ -19,11 +26,79 @@ fn basic() {
...
+    let message = TestRecord { bytes: b"hello world".to_vec() };
+
+    let wal = create_wal(&directory);
+    let written_entry_id = wal.sequenced_write(TestRecord::RECORD_TYPE, message.bytes()).unwrap();
+    drop(wal);
+
+    let wal_file = &read_dir(directory.path().join("wal")).unwrap().exactly_one().unwrap().unwrap().path();
+    let len = fs::metadata(wal_file).unwrap().len();
+    let mut file = BufWriter::new(OpenOptions::new().read(true).append(true).open(wal_file).unwrap());
+    file.write_all(&thread_rng().gen::<[u8; 32]>()).unwrap();
```
dmitrii-ubskii: Maximum number of bytes `rand` will generate in one go.

durability/wal.rs
```diff
@@ -54,11 +54,10 @@ impl WAL {
             fs::create_dir_all(wal_dir.clone()).map_err(|err| WALError::CreateError { source: Arc::new(err) })?;
         }
 
-        let files = Files::open(wal_dir.clone()).map_err(|err| WALError::CreateError { source: Arc::new(err) })?;
+        let files = Files::open(wal_dir.clone())?;
```
dmitrii-ubskii, line 57: Most of `.map_err`s go away since we may generate `DurabilityServiceError`s (rather than `io::Error`s or `WALError`s) a lot earlier now.

durability/wal.rs
```diff
@@ -341,6 +357,25 @@ impl File {
         Ok(Self { start: DurabilitySequenceNumber::from(num), len, path })
     }
 
+    fn trim_corrupted_tail(&mut self) -> Result<(), DurabilityServiceError> {
```
dmitrii-ubskii, line 360: Called once during startup, and only on the last file in the WAL. The name could probably be improved.

storage/durability_client.rs
```diff
@@ -113,30 +113,13 @@ impl WALClient {
 
     fn serialise_record(record: &impl DurabilityRecord) -> Result<Vec<u8>, DurabilityClientError> {
         let mut buf = Vec::new();
-        let mut encoder = lz4::EncoderBuilder::new()
-            .build(&mut buf)
-            .map_err(|err| DurabilityClientError::CompressionError { source: Arc::new(err) })?;
-        record.serialise_into(&mut encoder)?;
-        encoder.finish().1.map_err(|err| DurabilityClientError::CompressionError { source: Arc::new(err) })?;
+        record.serialise_into(&mut buf)?;
```
dmitrii-ubskii, line 116: All compression handling has been moved to the WAL service.

durability/tests/basic.rs
```diff
@@ -7,8 +7,15 @@
...
 
+use std::{
+    fs::{self, read_dir, OpenOptions},
+    io::{BufWriter, Seek, Write},
+};
+
 use durability::DurabilityService;
 use durability_test_common::{create_wal, load_wal, TestRecord};
+use itertools::Itertools;
+use rand::prelude::*;
```
krishnangovindraj, line 18: We used to have a java rule against `*` imports. I didn't even know rust had those.

durability/tests/basic.rs
```diff
@@ -19,11 +26,79 @@ fn basic() {
...
+    let message = TestRecord { bytes: b"hello world".to_vec() };
+
+    let wal = create_wal(&directory);
+    let written_entry_id = wal.sequenced_write(TestRecord::RECORD_TYPE, message.bytes()).unwrap();
+    drop(wal);
+
+    let wal_file = &read_dir(directory.path().join("wal")).unwrap().exactly_one().unwrap().unwrap().path();
+    let len = fs::metadata(wal_file).unwrap().len();
+    let mut file = BufWriter::new(OpenOptions::new().read(true).append(true).open(wal_file).unwrap());
+    file.write_all(&[0; 120]).unwrap();
```
krishnangovindraj: Is this guaranteed to have been flushed?
If it's not, I guess the len check at the end would still pass

durability/tests/basic.rs
krishnangovindraj: Nice. I didn't think of that case.

durability/tests/basic.rs
```diff
@@ -19,11 +26,79 @@ fn basic() {
...
+    let message = TestRecord { bytes: b"hello world".to_vec() };
+
+    let wal = create_wal(&directory);
+    let written_entry_id = wal.sequenced_write(TestRecord::RECORD_TYPE, message.bytes()).unwrap();
+    drop(wal);
+
+    let wal_file = &read_dir(directory.path().join("wal")).unwrap().exactly_one().unwrap().unwrap().path();
+    let mut file = OpenOptions::new().read(true).write(true).open(wal_file).unwrap();
+    file.seek(std::io::SeekFrom::End(-16)).unwrap();
+    file.write_all(&[0; 16]).unwrap();
```
krishnangovindraj: Is 16 a random choice? Is 0000 a likely corruption mode or is random garbage more likely? I'd like to see random garbage because 00 can be harmless for fields like length, whereas random bytes could be a very big number and cause some trouble.

durability/wal.rs
krishnangovindraj: Do we have any sort of guarantees around acknowledged commits being in the database? 
I guess it's implicit if the only case where a corruption can happen is a crash during write or sync. Is that the current (and reasonable) assumption?
Could you add that context into the PR description please?

krishnangovindraj (approved):
Approved. 
But can we have a short description about the need for this change/the failure modes (corruptions caused by crashes during writes, etc)?

durability/tests/basic.rs
dmitrii-ubskii: Good catch, this used to be in a block which would drop the `file` and flush.

durability/tests/basic.rs
dmitrii-ubskii: `prelude`s are made to be star-imported. I could expand this, but I'm less worried in a test.

durability/tests/basic.rs
dmitrii-ubskii: Somewhat arbitrary, this is around the minimum size of a non-empty lz4 frame (mainly to make sure seek succeeds). Replaced with a for loop and added junk test.

durability/wal.rs
dmitrii-ubskii: The guarantee is that if a commit is acknowledged, it has been persisted to the WAL and the DB. "Persisted to DB" != "checkpointed" though.

If the WAL is corrupted but the last good record is not behind the checkpoint, this is at worst equivalent to a rollback of at most `CHECKPOINT_INTERVAL` (1 minute) in case of corruption. Not the worst thing in case of a catastrophic outage like that.

If the WAL is corrupted and behind the checkpoint, we currently give up on recovery so as not to nuke legitimate data.

storage/recovery/checkpoint.rs
flyingsilverfin: Ok yes i'm happy for this to crash the server and they have to come to us. Thanks!

flyingsilverfin (approved):

PR #7511, opened 2025-07-03
Add HTTP authentication action diagnostics counts

farost:
Add diagnostics reporting for authentication actions performed through the HTTP endpoint.

Previously, due to its excessive usage for every request and low value, we ignored this method. However, it's the most efficient metric to understand the usage of the HTTP endpoint of the server, so it is a valuable piece of usage metrics for us. 

## Implementation
We additionally cleanup some unused code and comments.

### DIFF
--- a/resource/constants.rs
+++ b/resource/constants.rs
@@ -75,8 +75,6 @@ pub mod server {
 
     pub const DATABASE_METRICS_UPDATE_INTERVAL: Duration = Duration::from_secs(10 * SECONDS_IN_MINUTE);
 
-    pub const DEFAULT_ADDRESS: &str = "0.0.0.0:1729";
-    pub const DEFAULT_HTTP_ADDRESS: &str = "0.0.0.0:8000";
     pub const DEFAULT_USER_NAME: &str = "admin";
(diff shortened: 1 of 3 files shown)


server/service/http/transaction_service.rs
```diff
@@ -355,50 +355,15 @@ impl TransactionService {
                 match request {
                     TransactionRequest::Query(query_options, query) => {
                         self.handle_query(query_options, query, response_sender).await
-                        // run_with_diagnostics_async(
```
farost, line 358: We cover it in `http/typedb_service`, outside of this service. Idk how these comments were left here, probably rebases =/

resource/constants.rs
```diff
@@ -75,8 +75,6 @@ pub mod server {
 
     pub const DATABASE_METRICS_UPDATE_INTERVAL: Duration = Duration::from_secs(10 * SECONDS_IN_MINUTE);
 
-    pub const DEFAULT_ADDRESS: &str = "0.0.0.0:1729";
```
farost, line 78: We put it into config now, these lines are not used.

PR #7518, opened 2025-07-08
Rename config and CLI fields to align with the naming conventions

farost:
Perform the following renames in config files:
* `server.http-enabled` -> `server.http.enabled`
* `server.http-address` -> `server.http.address`
* `server.encryption.certificate_key` -> `server.encryption.certificate-key`

Perform the following renames in CLI arguments:
* `logging.logdir` -> `logging.directory`

## Implementation

### DIFF
--- a/server/config.yml
+++ b/server/config.yml
@@ -4,8 +4,9 @@
 
 server:
     address: 0.0.0.0:1729
-    http-enabled: true
-    http-address: 0.0.0.0:8000
+    http:
+        enabled: true
+        address: 0.0.0.0:8000
 
     authentication:
         token-expiration-seconds: 5000
@@ -13,7 +14,7 @@ server:
     encryption:
         enabled: false
         certificate:
-        certificate_key:
+        certificate-key:
         ca-certificate:
 
 storage:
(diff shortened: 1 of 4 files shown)


server/config.yml
```diff
@@ -29,6 +30,3 @@ diagnostics:
     reporting:
         metrics: true
         errors: true
-
-development-mode:
```
farost, line 33: I tested it locally, the default values work just fine. No need to excessively highlight this flag

krishnangovindraj: This is for @flyingsilverfin 
I kept it exhaustive so we don't need a separate documentation of all supported options.

krishnangovindraj (approved):
Deferring one comment to Joshua, else it looks good.

server/config.yml
farost: I don't think we need it documented. It's intentionally not listed in the `help` menu.

flyingsilverfin: I'd probably prefer to leave it out of the user's view -- it's really meant for us as developers, not the users "while developing" their application which they might assume it is and enable... which means we dont get reporting!

PR #7522, opened 2025-07-09
Add server status retrieval

farost:
Add server status APIs to `ServerState` to retrieve server addresses and replica statuses. The general goal is to be able to share these states across replicas and return them to the clients requesting `servers_all`.

## Implementation
Rename `ServerInfo` to `DistributionInfo`.
Introduce `ServerStatus` trait with a `LocalServerStatus` implementation, containing server addresses as strings (unparsed -- we will lose the parsed variant across multiple protocol layers anyway). This trait will be implemented in Cluster as `ClusteredServerStatus`, an extended version with replication information.
Extend `ServerState` to return `server_status` for a single server and `servers_statuses` for multiple servers as a response to `servers_all` grpc request.


server/lib.rs
```diff
@@ -9,44 +9,47 @@ pub mod error;
 pub mod parameters;
 pub mod service;
 pub mod state;
+pub mod status;
```
farost, line 12: Mostly rustfmt

server/lib.rs
```diff
@@ -138,50 +148,42 @@ pub struct Server {
...
-        let grpc_address = Self::resolve_address(self.config.server.address).await;
-        let http_address_opt = if self.config.server.http.enabled {
-            Some(
-                Self::validate_and_resolve_http_address(self.config.server.http.address.clone(), grpc_address.clone())
-                    .await?,
-            )
-        } else {
-            None
-        };
+        let server_state = self.server_state;
```
farost, line 163: We get the addresses from the server state now. Explanation in a separate comment.

server/state.rs
```diff
@@ -162,16 +172,26 @@ impl LocalServerState {
...
-                server_info,
+                distribution_info,
                 &config.diagnostics,
                 config.storage.data_directory.clone(),
                 config.development_mode.enabled,
             )
             .await,
         );
 
+        let grpc_address = Self::resolve_address(&config.server.address).await;
```
farost, line 185: As we construct the ServerState differently in Cluster, we need to make sure that the addresses are resolved in a single place. This place is currently here. If we used to resolve addresses as before, we'd need to copypaste this logic in Cluster.

server/lib.rs
farost: See https://github.com/typedb/typedb/pull/7522/files#diff-cb5b51501b6674285b3f8468601a08cd482914daec21ee49d1105407669e99aaR183

server/state.rs
```diff
@@ -50,13 +42,25 @@ use crate::{
...
 
 pub type BoxServerState = Box<dyn ServerState + Send + Sync>;
+pub type BoxServerStatus = Box<dyn ServerStatus + Send + Sync>;
 
 #[async_trait]
 pub trait ServerState: Debug {
-    async fn servers_register(&self, clustering_id: u64, address: String) -> Result<(), ServerStateError>;
+    async fn distribution_info(&self) -> DistributionInfo;
+
+    async fn grpc_address(&self) -> SocketAddr;
```
farost, line 56: Might be too much, but I found it useful.

lolski: Hmm do we really need `grpc_address` and `http_address` in this class?

farost: I initially put `SocketAddr`s (parsed copyable types) in the `ServerStatus`. Then, you could retrieve this server status from the server state and get the address. 
However, in the protocol, we encode these addresses as strings, just like they are configured in the config files. Thus, for every `servers_all` call, and for every part of sending addresses of servers to the requesting replica, you need to send strings. Thus, we either store strings or always call `SocketAddr::to_string` and then `parse...unwrap()`, which is... ugghh.
Currently, I've done this:
* `ServerStatus` stores the description of the server. It can be retrieved and read + converted into a protocol message. It stores the string representation of addresses for minimal conversions.
* `ServerState` stores already parsed addresses. You don't need to call for `state.server_status().grpc_address().parse().unwrap()`, you just call `state.grpc_address()`. It is safe and predictable: we expect to initialize these addresses on boot up.  

This way, I can use `ServerStatus` to send the description of the server between replicas and to the user, if needed, and `ServerState` for actual connection operations. The domain is not smooth and is for further refactoring, but that's the point where I stopped considering that the whole design of `ServerState`s can be changed.

PR #7526, opened 2025-07-11
Only seek on iterators that are behind in merge

dmitrii-ubskii:

Fix a crash when during a seek on a `KMergeBy`, which is merging multiple sorted tuple iterators, `seek` is called on iterators that are already ahead of the seek target, an invalid operation. It is valid to `seek` the iterator as a whole, as the merge may still be behind the seek point even if some constituent iterators are ahead.

### DIFF
--- a/common/lending_iterator/kmerge.rs
+++ b/common/lending_iterator/kmerge.rs
@@ -92,13 +92,24 @@ where
     fn seek(&mut self, key: &K) {
         self.iterators = mem::take(&mut self.iterators)
             .drain()
-            .map(|mut it| {
-                it.iter.seek(key);
-                it
+            .filter_map(|mut it| {
+                it.iter.peek();
+                if let Some(item) = it.iter.get_peeked() {
+                    if it.iter.compare_key(item, key) == Ordering::Less {
+                        it.iter.seek(key);
+                    }
+                }
+                it.iter.peek().is_some().then_some(it)
             })
             .collect();
-        if let Some(next_iterator) = self.next_iterator.as_mut() {
-            next_iterator.iter.seek(key);
+        if let Some(mut next_iterator) = self.next_iterator.take() {
+            next_iterator.iter.peek();
+            if let Some(item) = next_iterator.iter.get_peeked() {
+                if next_iterator.iter.compare_key(item, key) == Ordering::Less {
+                    next_iterator.iter.seek(key);
+                }
+            }
+            self.next_iterator = Some(next_iterator);
         }
         // force recomputation of heap element
         self.state = State::Used;
@@ -108,7 +119,7 @@ where
         if let Some(inner) = self.next_iterator.as_ref().or(self.iterators.peek()) {
             inner.iter.compare_key(item, key)
         } else {
-            Ordering::Less // let the enclosing iterator exhaust itself?
+            unreachable!("Called `Seekable::compare_key` on an empty KMergeBy") // no inner iterators
         }
     }
 }
--- a/common/lending_iterator/kmerge.rs
+++ b/common/lending_iterator/kmerge.rs
@@ -15,7 +15,7 @@ pub struct KMergeBy<I: LendingIterator, F> {
     phantom_compare: PhantomData<F>,
 }
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, PartialEq, Eq)]
 pub enum State {
     Init,
     Used,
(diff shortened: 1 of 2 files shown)


executor/instruction/iterator.rs
```diff
@@ -125,10 +115,25 @@ impl<I: for<'a> LendingIterator<Item<'a> = TupleResult<'static>> + TupleSeekable
...
+            .filter_map(|mut it| {
+                if let Some(Ok(item)) = it.iter.peek() {
+                    if item < target {
+                        if let Err(err) = it.iter.seek(target) {
+                            return Some(Err(err));
+                        }
+                    }
+                }
+                it.iter.peek().is_some().then_some(Ok::<_, Box<ConceptReadError>>(it))
+            })
```
dmitrii-ubskii, line 134: Check if the merge targets are actually behind the seek target. It's valid for some to be ahead.

common/lending_iterator/kmerge.rs
```diff
@@ -108,7 +119,7 @@ where
         if let Some(inner) = self.next_iterator.as_ref().or(self.iterators.peek()) {
             inner.iter.compare_key(item, key)
         } else {
-            Ordering::Less // let the enclosing iterator exhaust itself?
+            unreachable!("Called `Seekable::compare_key` on an empty KMergeBy") // no inner iterators
```
dmitrii-ubskii, line 124: We no longer use this in `seek` above.

common/lending_iterator/kmerge.rs
```diff
@@ -92,13 +92,24 @@ where
...
-        if let Some(next_iterator) = self.next_iterator.as_mut() {
-            next_iterator.iter.seek(key);
+        if let Some(mut next_iterator) = self.next_iterator.take() {
+            next_iterator.iter.peek();
+            if let Some(item) = next_iterator.iter.get_peeked() {
+                if next_iterator.iter.compare_key(item, key) == Ordering::Less {
+                    next_iterator.iter.seek(key);
+                }
+            }
+            self.next_iterator = Some(next_iterator);
```
dmitrii-ubskii, line 112: The `take` and outlined `peek` is needed so we can use the iterator's own `compare_key`.

common/lending_iterator/kmerge.rs
```diff
@@ -92,13 +92,24 @@ where
...
-                it
+            .filter_map(|mut it| {
+                it.iter.peek();
+                if let Some(item) = it.iter.get_peeked() {
+                    if it.iter.compare_key(item, key) == Ordering::Less {
+                        it.iter.seek(key);
+                    }
+                }
+                it.iter.peek().is_some().then_some(it)
             })
```
dmitrii-ubskii, line 114: Filtering out empty iterators in the merge (otherwise the comparison panics during heap sifts), and only seeking where it's valid to do so.

common/lending_iterator/kmerge.rs
```diff
@@ -90,16 +90,21 @@ where
...
-            .map(|mut it| {
-                it.iter.seek(key);
-                it
+            .filter_map(|mut it| {
+                it.iter.peek();
+                if let Some(item) = it.iter.get_peeked() {
+                    if it.iter.compare_key(item, key) == Ordering::Less {
+                        it.iter.seek(key);
+                    }
+                }
```
dmitrii-ubskii, line 112: The outlined `peek` is needed so we can use the iterator's own `compare_key`.

common/lending_iterator/kmerge.rs
```diff
@@ -90,16 +94,25 @@ where
     F: for<'a, 'b> FnHktHelper<(&'a I::Item<'a>, &'b I::Item<'b>), Ordering> + Copy + 'static,
 {
     fn seek(&mut self, key: &K) {
+        if self.state == State::Used {
+            self.return_last_to_heap();
+            self.find_next_state();
+        }
+        if let Some(next_iterator) = &mut self.next_iterator {
+            next_iterator.iter.seek(key);
+        }
```
dmitrii-ubskii, line 103: This will fail (correctly this time) if the merge as a whole is ahead of the seek target.

common/lending_iterator/kmerge.rs
```diff
@@ -43,7 +43,7 @@ where
         Self { iterators: queue, next_iterator: None, state: State::Init, phantom_compare: PhantomData }
     }
 
-    fn find_next_state(&mut self) {
+    pub fn find_next_state(&mut self) {
```
dmitrii-ubskii, line 46: Had to make this `pub` so that the trait impl in `executor::instruction::iterator` is possible!

PR #7535, opened 2025-07-21
Implement missing reducers: min/max, decimal

dmitrii-ubskii:

We implement min/max reducers for every totally ordered value type (NOTE: does not include `Duration` as it's only partially ordered). We also implement statistical reducers (mean/std/median) for `Decimal`, decaying to `Double` in `std`.


encoding/value/decimal_value.rs
```diff
@@ -352,6 +417,12 @@ mod tests {
 
                 assert_eq!(as_i128(lhs * rhs), bigint_mul_result, "{:?} * {:?} != {:?}", lhs, rhs, lhs * rhs);
             }
+
+            // can only divide by u64
+            let rhs = rng.gen_range(1..=FRACTIONAL_PART_DENOMINATOR);
+            if as_i128(lhs).checked_div(rhs as i128).is_some_and(|res| range.contains(&res)) {
+                assert_eq!(as_i128(lhs / rhs), as_i128(lhs) / rhs as i128, "{:?} / {:?} != {:?}", lhs, rhs, lhs / rhs);
+            }
```
dmitrii-ubskii, line 425: The naïve implementation is identical to the one we're checking here, this test only exists to protect against regressions.

executor/reduce_executor.rs
```diff
@@ -693,3 +738,46 @@ impl ReducerAPI for StdDoubleExecutor {
...
+    }
+
+    fn finalise(self) -> Option<VariableValue<'static>> {
+        if self.count > 1 {
+            let sum = self.sum;
+            let sum_squares = self.sum_squares;
+            let n = self.count;
+            let mean = sum / n;
+            let sample_variance: Decimal = (sum_squares + n * mean * mean - 2 * sum * mean) / (n - 1);
+            Some(VariableValue::Value(Value::Double(sample_variance.to_f64().sqrt())))
```
dmitrii-ubskii, line 778: `sqrt` is the only reason `StdDecimal` returns `Double` instead of `Decimal`. The question remains whether it's even sensible to have `sqrt(Decimal) -> Decimal` in the first place. I'm inclined to say that once we're going irrational, we should probably not overpromise the precision.

executor/reduce_executor.rs
```diff
@@ -260,268 +232,301 @@ impl ReducerAPI for CountVarExecutor {
...
+    Date::date(NaiveDate),
+    DateTime::date_time(NaiveDateTime),
+    DateTimeTZ::date_time_tz(DateTime<TimeZone>),
 }
 
 #[derive(Debug, Clone)]
-struct MaxDoubleExecutor {
-    max: Option<f64>,
+struct MinStringExecutor {
+    min: Option<String>,
```
dmitrii-ubskii, line 363: Diff is very confused from here on.

executor/reduce_executor.rs
```diff
@@ -260,268 +232,301 @@ impl ReducerAPI for CountVarExecutor {
...
-        self.max.map(|v| VariableValue::Value(Value::Integer(v)))
-    }
+minmax_reducer_executors! {
+    Integer::integer(i64),
+    Double::double(f64),
+    Decimal::decimal(Decimal),
+    Date::date(NaiveDate),
+    DateTime::date_time(NaiveDateTime),
+    DateTimeTZ::date_time_tz(DateTime<TimeZone>),
 }
```
dmitrii-ubskii, line 359: `Cow<'a, str>` is the only reason `String` is implemented separately and not in the macro. It's just enough extra handling to be more trouble than it's worth in the macro.

compiler/executable/reduce.rs
```diff
@@ -101,17 +131,31 @@ impl<ID: IrID> ReduceInstruction<ID> {
...
+            ReduceInstruction::MedianDecimal(id) => ReduceInstruction::MedianDecimal(mapping[&id]),
+            ReduceInstruction::StdDecimal(id) => ReduceInstruction::StdDecimal(mapping[&id]),
+            ReduceInstruction::MaxString(id) => ReduceInstruction::MaxString(mapping[&id]),
+            ReduceInstruction::MinString(id) => ReduceInstruction::MinString(mapping[&id]),
+            ReduceInstruction::MaxDate(id) => ReduceInstruction::MaxDate(mapping[&id]),
+            ReduceInstruction::MinDate(id) => ReduceInstruction::MinDate(mapping[&id]),
+            ReduceInstruction::MaxDateTime(id) => ReduceInstruction::MaxDateTime(mapping[&id]),
+            ReduceInstruction::MinDateTime(id) => ReduceInstruction::MinDateTime(mapping[&id]),
+            ReduceInstruction::MaxDateTimeTZ(id) => ReduceInstruction::MaxDateTimeTZ(mapping[&id]),
+            ReduceInstruction::MinDateTimeTZ(id) => ReduceInstruction::MinDateTimeTZ(mapping[&id]),
```
krishnangovindraj, line 158: Would be great if we could macro this away, or make it type-checked somehow.

compiler/executable/reduce.rs
```diff
@@ -38,61 +38,91 @@ pub struct ReduceRowsExecutable {
...
+            Self::MeanDouble(_) => ValueType::Double,
+            Self::MedianDouble(_) => ValueType::Double,
             Self::StdDouble(_) => ValueType::Double,
+
+            Self::SumDecimal(_) => ValueType::Decimal,
+            Self::MaxDecimal(_) => ValueType::Decimal,
+            Self::MinDecimal(_) => ValueType::Decimal,
+            Self::MeanDecimal(_) => ValueType::Decimal,
+            Self::MedianDecimal(_) => ValueType::Decimal,
+            Self::StdDecimal(_) => ValueType::Double,
```
krishnangovindraj, line 113: Nice, I missed this the first time I scrolled over.
I'm wondering whether this can be an associated constant if we were to replace the macros wit parameterised types.

executor/reduce_executor.rs
```diff
@@ -260,268 +232,301 @@ impl ReducerAPI for CountVarExecutor {
...
 
-#[derive(Debug, Clone)]
-struct SumDoubleExecutor {
-    sum: f64,
-    target: VariablePosition,
+sum_reducer_executors! {
+    Integer(i64) (|x| x as i64),
+    Double(f64) (|x| x as f64),
+    Decimal(Decimal) (|x| x),
```
krishnangovindraj, line 273: Do you think we could do this using parametrised types instead of macros?

encoding/value/decimal_value.rs
```diff
@@ -183,7 +195,7 @@ macro_rules! impl_integer_ops {
 impl_integer_ops! {
     Add::add, Sub::sub, Mul::mul
     for
-    { u8, u16, u32, u64, i8, i16, i32, i64 }
+    { u8, u16, u32, u64, usize, i8, i16, i32, i64, isize }
```
krishnangovindraj, line 198: Where do we need `usize` and `isize` for this?

encoding/value/decimal_value.rs
```diff
@@ -214,6 +226,53 @@ macro_rules! impl_ref_ops {
...
+impl Div<u64> for Decimal {
+    type Output = Self;
+
+    fn div(self, rhs: u64) -> Self::Output {
+        let i128 = (self.integer as i128 * FRACTIONAL_PART_DENOMINATOR as i128 + self.fractional as i128) / rhs as i128;
+        let int = i128.div_euclid(FRACTIONAL_PART_DENOMINATOR as i128) as i64;
+        let frac = i128.rem_euclid(FRACTIONAL_PART_DENOMINATOR as i128) as u64;
+        Decimal::new(int, frac)
+    }
+}
```
krishnangovindraj, line 238: Just checking my understanding here in contrast to the Std operator:

i128 `/` does integer division and throws off the decimal. So we'd have 
` (int * DENOM + frac )/DENOM + (truncated_decimal /  DENOM)`
Since there's no loss of precision beyond the digits we wouldn't be able to represent anyway, we can keep it in a `Decimal`

But with sqrt, we pass through f64 so we have a loss of precision that's much worse than what `Decimal` otherwise guarantees?

krishnangovindraj (approved):
Looks good.
I would like to see if we can replace some of the macros introduced with parametrised types & generic impls, but that might be a bigger refactor

executor/reduce_executor.rs
dmitrii-ubskii: Yes, this could be a trait. I don't think it's worth it though for something this small.

encoding/value/decimal_value.rs
dmitrii-ubskii: I added these because I thought `count` would be `usize`, but there's really no reason not to have them.

encoding/value/decimal_value.rs
dmitrii-ubskii: Right, we go from at most 38 digits of precision in `Decimal` to ~16 after the conversion to `f64`. Theoretical best precision of a square root in `Decimal` would be 27 digits.

encoding/value/decimal_value.rs
krishnangovindraj: Fair enough

flyingsilverfin (approved):

kenho811:
@dmitrii-ubskii 

Thanks!

Does it mean after this code is deployed to the next release, I can do stuff like min() and max() on datetime and date as well?

dmitrii-ubskii:
> Does it mean after this code is deployed to the next release, I can do stuff like min() and max() on datetime and date as well?

That's correct.

kenho811:
> > Does it mean after this code is deployed to the next release, I can do stuff like min() and max() on datetime and date as well?
> 
> That's correct.

Thanks. Looking forward to that.

PR #7536, opened 2025-07-24
Publicly expose GRPC error traits to reuse in cluster

farost:
Allow external crate's users to use `IntoGrpcStatus` and `IntoProtocolErrorMessage` traits to convert `TypeDBError`s into typedb protocol errors and, then, GRPC statuses.

## Implementation
Instead of exposing the whole `service/grpc` mod or its `error` submod with service-specific errors, we safely expose only the traits required.

Additionally, for convenience, add version BDD step implementation to handle new `typedb-driver`'s public API tests, affecting the HTTP endpoint.

### DIFF
--- a/tests/behaviour/service/http/http_steps/connection/mod.rs
+++ b/tests/behaviour/service/http/http_steps/connection/mod.rs
@@ -18,7 +18,7 @@ use test_utils::create_tmp_dir;
 
 use crate::{
     generic_step,
-    message::{authenticate, authenticate_default, check_health, databases, send_get_request, users},
+    message::{authenticate, authenticate_default, check_health, databases, send_get_request, version,users},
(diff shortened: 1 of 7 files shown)


tests/behaviour/service/http/http_steps/connection/mod.rs
```diff
@@ -158,6 +158,22 @@ async fn connection_opens_with_a_wrong_port(context: &mut Context, may_error: pa
     );
 }
 
+#[apply(generic_step)]
+#[step(expr = r"connection contains distribution{may_error}")]
```
farost, line 162: I accidentally pushed it to this branch, but it's a safe thing, and I'd need a small PR for these BDD steps anyway. I think it's simpler to merge it now.

lolski (approved):

PR #7537, opened 2025-07-24
Analyse endpoint

krishnangovindraj:

### DIFF
--- a/query/query_manager.rs
+++ b/query/query_manager.rs
@@ -17,7 +17,7 @@ use executor::pipeline::{
     pipeline::Pipeline,
     stage::{ReadPipelineStage, WritePipelineStage},
 };
-use function::function_manager::{validate_no_cycles, FunctionManager, ReadThroughFunctionSignatureIndex};
+use function::function_manager::{FunctionManager, ReadThroughFunctionSignatureIndex, validate_no_cycles};
 use ir::{
     pipeline::function_signature::{FunctionID, HashMapFunctionSignatureIndex},
     translation::pipeline::{translate_pipeline, TranslatedPipeline},
@@ -29,6 +29,10 @@ use resource::{
 use storage::snapshot::{ReadableSnapshot, WritableSnapshot};
 use tracing::{event, Level};
 use typeql::query::SchemaQuery;
+use compiler::annotation::fetch::AnnotatedFetchObject;
+use compiler::query_structure::{ParametrisedQueryStructure, QueryStructure};
+use executor::document::ConceptDocument;
+use ir::pipeline::VariableRegistry;
 
 use crate::{define, error::QueryError, query_cache::QueryCache, redefine, undefine};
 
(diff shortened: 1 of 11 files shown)


compiler/query_structure.rs
```diff
@@ -256,20 +294,22 @@ impl<'a> ParametrisedQueryStructureBuilder<'a> {
...
+            debug_assert!((id as usize) < self.pipeline_structure.blocks.len());
+            self.pipeline_structure.blocks[id as usize] = QueryStructureBlock { constraints: Vec::from(constraints) };
             QueryStructureBlockID(id)
         } else {
-            self.query_structure.blocks.push(QueryStructureBlock { constraints: Vec::from(constraints) });
-            QueryStructureBlockID(self.query_structure.blocks.len() as u16 - 1)
+            self.pipeline_structure.blocks.push(QueryStructureBlock { constraints: Vec::from(constraints) });
+            QueryStructureBlockID(self.pipeline_structure.blocks.len() as u16 - 1)
         };
+        self.pipeline_structure.scope_to_block.insert(scope_id, block_id.clone());
```
krishnangovindraj, line 414: This was the only practical change in this file when I wrote this comment

server/service/http/message/query/mod.rs
```diff
@@ -132,3 +151,175 @@ impl IntoResponse for QueryAnswer {
...
+            };
+            Ok((key, encoded))
+        })
+        .collect::<Result<HashMap<_, _>, _>>()
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct AnalysedQueryResponse {
+    // structure: QueryStructureResponse, // TODO:
```
krishnangovindraj: Harder than expected given we have nested pipelines in fetch & preamble.

compiler/query_structure.rs
```diff
@@ -21,65 +29,132 @@ use serde::{Deserialize, Serialize, Serializer};
...
     },
     VariablePosition,
 };
 
 #[derive(Debug, Clone)]
 pub struct QueryStructure {
-    pub parametrised_structure: Arc<ParametrisedQueryStructure>,
+    pub preamble: Vec<FunctionStructure>,
+    pub pipeline: Option<PipelineStructure>,
+    // fetch: Option<FetchStructure>, // TODO?
```
krishnangovindraj: Unclear how useful the fetch structure is or how to represent it without having extremely elaborate structures.

server/service/http/message/query/query_structure.rs
```diff
@@ -64,11 +66,25 @@ impl<'a, Snapshot: ReadableSnapshot> QueryStructureContext<'a, Snapshot> {
...
+    pipeline: Option<PipelineStructureResponse>,
+    preamble: Vec<FunctionStructureResponse>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct PipelineStructureResponse {
     blocks: Vec<StructureBlock>,
-    pipeline: Vec<QueryStructureStage>,
+    stages: Vec<QueryStructureStage>,
```
krishnangovindraj: Breaks the API, but avoids an awkward pipeline inside pipeline nesting.

krishnangovindraj (commented):
General question : 
```
{
    "structure": {
        "entry": {
              "pipeline" : { ... }
          },
                  "pipeline": {
              "stages" : { ... }
          },
          "preamble":  [
                 { "pipeline": ... },
                  ...
           ]
    }
} 
```

or 

```
{
    "structure": {
        "pipeline": {
              "stages" : { ... }
          },
          "preamble":  [
                 { "stages": ... },
                  ...
           ]
    }
} 
```

compiler/query_structure.rs
flyingsilverfin: Let's leave this one for later IMO --  structure is really the 'graph structure' which doesn't apply to fetch. I'd just delete the TODO and commented line

compiler/query_structure.rs
```diff
@@ -21,65 +29,132 @@ use serde::{Deserialize, Serialize, Serializer};
...
-    pub parametrised_structure: Arc<ParametrisedQueryStructure>,
+    pub preamble: Vec<FunctionStructure>,
+    pub query: Option<PipelineStructure>,
+    // fetch: Option<FetchStructure>, // TODO?
+}
+
+#[derive(Debug, Clone)]
+pub struct FunctionStructure {
+    pub pipeline: Option<PipelineStructure>,
+    // TODO: arguments, returned
```
flyingsilverfin: can be impl'ed i think?

query/query_manager.rs
```diff
@@ -399,4 +428,352 @@ impl QueryManager {
             Box::new(QueryError::Representation { source_query: source_query.to_string(), typedb_source: err })
         })
     }
+
+    pub fn analyse_query<Snapshot: ReadableSnapshot + 'static>(
+        &self,
+        snapshot: Arc<Snapshot>,
+        type_manager: &TypeManager,
+        _thing_manager: Arc<ThingManager>,
```
flyingsilverfin, line 306: you can just remove this shouldn't need it for compile/analyze

query/query_manager.rs
```diff
@@ -399,4 +428,352 @@ impl QueryManager {
...
+        function_manager: &FunctionManager,
+        query: &typeql::query::Pipeline,
+        source_query: &str,
+    ) -> Result<AnalysedQuery, Box<QueryError>> {
+        event!(Level::TRACE, "Running analyse query:\n{}", query);
+        let mut query_profile = QueryProfile::new(tracing::enabled!(Level::TRACE));
+        let compile_profile = query_profile.compilation_profile();
+        compile_profile.start();
+        // 1: Translate
```
flyingsilverfin, line 315: this is the same as executing queries right? can we re-use & modularize the code at all instead of duplicating?

query/query_manager.rs
```diff
@@ -399,4 +428,352 @@ impl QueryManager {
...
+}
+
+#[derive(Debug)]
+pub struct QueryStructureAnnotations {
+    pub preamble: Vec<FunctionStructureAnnotations>,
+    pub query: Option<PipelineStructureAnnotations>,
+    pub fetch: Option<FetchStructureAnnotations>,
+}
+
+impl QueryStructureAnnotations {
```
flyingsilverfin: Feels like this downwards is too much to squeeze into query_manager no?

query/query_manager.rs
```diff
@@ -399,4 +428,352 @@ impl QueryManager {
...
+    attribute_types: impl Iterator<Item = AttributeType>,
+) -> Result<BTreeSet<ValueType>, Box<ConceptReadError>> {
+    attribute_types
+        .filter_map(|attribute_type| {
+            attribute_type
+                .get_value_type(snapshot, type_manager)
+                .map(|ok| ok.map(|(value_type, _)| value_type))
+                .transpose()
+        })
+        .collect::<Result<BTreeSet<_>, _>>()
```
flyingsilverfin: maybe goes into an `analze.rs` module?

server/service/http/typedb_service.rs
```diff
@@ -546,6 +562,31 @@ impl TypeDBService {
         .await
     }
 
+    async fn transactions_analyse_query(
```
flyingsilverfin, line 565: eyy i think this should be singular no? is there a reason they all use `transactions` plural in this file?

flyingsilverfin: interesting the only singular is `transcation_query` ? @farost maybe one for you?

query/query_manager.rs
krishnangovindraj: We'll need the statistics if we want to output the plan

server/service/http/message/query/query_structure.rs
krishnangovindraj: This has been reverted, so no breakage here.

query/query_manager.rs
krishnangovindraj: We can for the read & write pipelines.
For analyse it doesn't seem worth it since we stop at annotations.
There isn't much happening here in common beyond translation & annotation. We can remove the "extract_query_structure" if we're going to make studio call analyse first and then run the query.

query/query_manager.rs
krishnangovindraj: Moved to analyse.rs as suggested.

flyingsilverfin (approved):

flyingsilverfin:
```
{
  "query": {
    "functions": [
      {
        "identifier": ...
        "pipeline": {
          ...
        },
        "return": ...
      }
    ],
    "pipeline": {
      "blocks": [
        {
          "annotations": {
            "variables": {
              "0": {
                "tag": "concept",
                "annotations": [
                  {
                    "kind": "entityType",
                    "label": "hardback"
                  },
                  {
                    "kind": "entityType",
                    "label": "paperback"
                  },
                  {
                    "kind": "entityType", ??? TODO
                    "label": "ebook"
                  }
                ]
              }
            }
          },
          "constraints": [
            {
              "textSpan": {
                "begin": 131,
                "end": 139
              },
              "tag": "isa",
              "instance": {
                "tag": "variable",
                "id": "0"
              },
              "type": {
                "tag": "label",
                "type": {
                  "kind": "entityType",
                  "label": "book"
                }
              }
            }
          ]
        }
      ],
      "stages": [
        {
          "tag": "match",
          "block": {
            "conjunction": [
              {
                "tag": "block",
                "index": 0
              }
            ]
          }
        }
      ],
      "fetch": {
        "annotations": {
          "title": [
            "string"
          ]
        }
      },
      "variables": {
        "0": {
          "name": "book"
        }
      },
      "outputs": [
        "0"
      ]
    },
  },
}

```

PR #7544, opened 2025-08-07
Update protocol to introduce replication ping

farost:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5244,7 +5244,7 @@ dependencies = [
 [[package]]
 name = "typedb-protocol"
 version = "0.0.0"
-source = "git+https://github.com/typedb/typedb-protocol?rev=d5b41223533feb485be7723ab961bb8c89b2fe9c#d5b41223533feb485be7723ab961bb8c89b2fe9c"
+source = "git+https://github.com/typedb/typedb-protocol?rev=39e22dcac2f7d73a25faea6e80105f20551d6f63#39e22dcac2f7d73a25faea6e80105f20551d6f63"
 dependencies = [
  "prost",
  "tonic",
--- a/dependencies/typedb/repositories.bzl
+++ b/dependencies/typedb/repositories.bzl
@@ -29,8 +29,8 @@ def typedb_protocol():
     # TODO: Return ref after merge to master
     git_repository(
         name = "typedb_protocol",
-        remote = "https://github.com/typedb/typedb-protocol",
-        commit = "d5b41223533feb485be7723ab961bb8c89b2fe9c",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_protocol
+        remote = "https://github.com/farost/typedb-protocol",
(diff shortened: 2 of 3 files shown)


database/database.rs
```diff
@@ -39,27 +38,32 @@ use encoding::{
 use error::typedb_error;
 use function::{function_cache::FunctionCache, FunctionError};
 use query::query_cache::QueryCache;
-use resource::constants::database::{CHECKPOINT_INTERVAL, STATISTICS_UPDATE_INTERVAL};
+use resource::{
```
farost, line 41: Many rusfmted files

server/lib.rs
```diff
@@ -26,29 +26,37 @@ use tokio::sync::watch::{channel, Receiver, Sender};
...
-    state::{BoxServerState, BoxServerStatus, LocalServerState},
+    state::{ArcServerState, BoxServerStatus, LocalServerState},
 };
 
-#[derive(Default)]
+#[derive(Debug)]
 pub struct ServerBuilder {
     distribution_info: Option<DistributionInfo>,
-    server_state_builder: Option<Box<dyn FnOnce(String) -> Pin<Box<dyn Future<Output = BoxServerState>>>>>,
+    server_state: Option<ArcServerState>,
```
farost, line 37: No builder, just server_state. Optional.

server/lib.rs
```diff
@@ -26,29 +26,37 @@ use tokio::sync::watch::{channel, Receiver, Sender};
...
+}
+
+impl Default for ServerBuilder {
+    fn default() -> Self {
+        Self { distribution_info: None, server_state: None, shutdown_channel: None, storage_server_id: None }
+    }
 }
 
 impl ServerBuilder {
+    pub fn new() -> Self {
```
farost, line 49: It feels like a natural way of constructing new builders

server/lib.rs
```diff
@@ -57,29 +65,36 @@ impl ServerBuilder {
...
-                Box::new(server_state)
+                Arc::new(server_state)
             }
         };
 
-        Ok(Server::new(distribution_info, config, Arc::new(server_state), shutdown_sender, shutdown_receiver))
+        Ok(Server::new(distribution_info, config, server_state, shutdown_sender, shutdown_receiver))
+    }
+
+    pub fn initialise_storage(&mut self, storage_config: &StorageConfig) -> Result<&str, ServerOpenError> {
```
farost, line 92: You can call this method separately to initialize storage before building the server for server_id retrieval and stuff. It's required for cluster to pre-init its server state.

server/state.rs
```diff
@@ -49,7 +50,9 @@ use crate::{
     status::{LocalServerStatus, ServerStatus},
 };
 
-pub type BoxServerState = Box<dyn ServerState + Send + Sync>;
```
farost, line 52: We never use separate BoxServerStates. It's all arced.


Instead of using `Arc<Box<dyn ServerState>>` everywhere, just use `Arc<dyn ServerState>`:
* It was redundant
* It was preventing us from casting dynamic objects to `ClusteredServerState`s and vice versa because of the two conflicting memory allocation abstractions.

lolski (approved):

PR #7549, opened 2025-08-15
Add server state's associated error type

farost:
Add an associated type `Error` to `ServerState` trait to extend error types in different trait implementations.


server/service/http/message/error.rs
```diff
@@ -33,30 +30,19 @@ impl IntoResponse for HttpServiceError {
...
-                ServerStateError::NotInitialised { .. } => StatusCode::INTERNAL_SERVER_ERROR,
-                ServerStateError::AuthenticationError { .. } => StatusCode::BAD_REQUEST,
-                ServerStateError::DatabaseExport { .. } => StatusCode::BAD_REQUEST,
-            },
+            HttpServiceError::State { typedb_source } => {
+                debug_assert!(
+                    typedb_source.code_prefix() == "LSS",
+                    "Expected only local server state errors. Override for "
+                );
+                match typedb_source.code_number() {
```
farost, line 38: Example of code-based match

PR #7553, opened 2025-08-18
Store answers in a hash table as well to avoid slow inserts in large table

krishnangovindraj:
Answer tables additionally store answers in a HashSet for constant time lookup, making constructing large tables linear instead of quadratic.

## Implementation
Store answers in a HashSet as well as the vector. This isn't great for memory usage but is a quick fix and far better than not having it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -333,9 +333,9 @@ dependencies = [
 
 [[package]]
 name = "async-std"
-version = "1.13.1"
+version = "1.13.2"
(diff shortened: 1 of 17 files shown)


flyingsilverfin:
A slightly more memory-efficient implementation (this doubles the memory right?) could be to use a HashMap of hash -> Vec<index>. To check equality is a bit of a two-step thing but saves storing the entire row twice. Just check if the hash exists in the map, then deref the TableIndex in the map to check if the row is equal to the actual row in the table

PR #7560, opened 2025-08-26
Add WAL truncation methods 

farost:
Expose public interfaces to request WAL's truncation starting from a sequential number. This is a required feature to use WAL as a serial storage for TypeDB Cluster's Raft log, which allows and expects record overriding.

### DIFF
--- a/durability/wal.rs
+++ b/durability/wal.rs
@@ -122,14 +122,6 @@ impl DurabilityService for WAL {
         self.registered_types.insert(durability_record_type, record_name.to_string());
     }
 
-    fn unsequenced_write(&self, record_type: DurabilityRecordType, bytes: &[u8]) -> Result<(), DurabilityServiceError> {
-        debug_assert!(self.registered_types.contains_key(&record_type));
-        let mut files = self.files.write().unwrap();
-        let raw_record = RawRecord { sequence_number: self.previous(), record_type, bytes: Cow::Borrowed(bytes) };
-        files.write_record(raw_record)?;
-        Ok(())
-    }
-
     fn sequenced_write(
         &self,
         record_type: DurabilityRecordType,
@@ -141,6 +133,14 @@ impl DurabilityService for WAL {
         let raw_record = RawRecord { sequence_number: seq, record_type, bytes: Cow::Borrowed(bytes) };
         files.write_record(raw_record)?;
         Ok(seq)
+    }
+
+    fn unsequenced_write(&self, record_type: DurabilityRecordType, bytes: &[u8]) -> Result<(), DurabilityServiceError> {
+        debug_assert!(self.registered_types.contains_key(&record_type));
+        let mut files = self.files.write().unwrap();
+        let raw_record = RawRecord { sequence_number: self.previous(), record_type, bytes: Cow::Borrowed(bytes) };
+        files.write_record(raw_record)?;
+        Ok(())
     }
 
     fn iter_any_from(
@@ -187,17 +187,26 @@ impl DurabilityService for WAL {
         Ok(None)
     }
 
+    fn truncate_from(&self, sequence_number: DurabilitySequenceNumber) -> Result<(), DurabilityServiceError> {
+        let mut files = self.files.write().unwrap();
+        files.truncate_from(sequence_number)?;
+        // TODO: Should we not fsync the while directory instead since it got deleted files?
+        files.sync_all();
+        self.next_sequence_number.store(sequence_number.number(), Ordering::SeqCst);
+        Ok(())
+    }
+
     fn delete_durability(self) -> Result<(), DurabilityServiceError> {
         drop(self.fsync_thread);
         let files = Arc::into_inner(self.files)
             .expect("cannot get exclusive ownership of WAL's Arc<Files>")
             .into_inner()
             .unwrap();
-        files.delete().map_err(|err| DurabilityServiceError::DeleteFailed { source: Arc::new(err) })
+        files.delete()
     }
 
     fn reset(&mut self) -> Result<(), DurabilityServiceError> {
-        self.next_sequence_number.store(DurabilitySequenceNumber::MIN.next().number, Ordering::SeqCst);
+        self.next_sequence_number.store(DurabilitySequenceNumber::MIN.next().number(), Ordering::SeqCst);
         self.files.write().unwrap().reset()
     }
 }
@@ -257,7 +266,7 @@ impl Files {
 
         let last = files.last_mut();
         let writer = if let Some(last) = last {
-            last.trim_corrupted_tail()?;
+            last.trim_corrupted_tail_if_needed()?;
             Some(File::writer(last)?)
         } else {
             None
@@ -309,9 +318,32 @@ impl Files {
         self.files.iter()
     }
 
-    fn delete(self) -> Result<(), io::Error> {
+    fn file_index_containing(&self, sequence_number: DurabilitySequenceNumber) -> Option<usize> {
+        self.files.iter().rposition(|f| f.start.number() <= sequence_number.number())
+    }
+
+    fn truncate_from(&mut self, sequence_number: DurabilitySequenceNumber) -> Result<(), DurabilityServiceError> {
+        let Some(file_index) = self.file_index_containing(sequence_number) else {
+            return Ok(());
+        };
+
+        let truncate_position = self.files[file_index]
+            .find_last_position_before(sequence_number)?
+            .expect("Expected existing position for the truncated sequence number. Do not do unnecessary truncations.");
+
+        while self.files.len() > file_index + 1 {
+            std::fs::remove_file(&self.files.pop().unwrap().path)?;
+        }
+
+        let last = &mut self.files[file_index];
+        last.truncate_from_position(truncate_position)?;
+        self.writer = Some(last.writer()?);
+        Ok(())
+    }
+
+    fn delete(self) -> Result<(), DurabilityServiceError> {
         drop(self.files);
-        std::fs::remove_dir_all(&self.directory)
+        std::fs::remove_dir_all(&self.directory).map_err(|source| source.into())
     }
 
     fn reset(&mut self) -> Result<(), DurabilityServiceError> {
@@ -357,12 +389,12 @@ impl File {
         Ok(Self { start: DurabilitySequenceNumber::from(num), len, path })
     }
 
-    fn trim_corrupted_tail(&mut self) -> Result<(), DurabilityServiceError> {
+    fn trim_corrupted_tail_if_needed(&mut self) -> Result<(), DurabilityServiceError> {
         let mut reader = FileReader::new(self.clone())?;
-        let mut last_successful_read_pos = 0;
+        let mut last_good_position = 0;
         while let Some(record) = reader.read_one_record().transpose() {
             if record.as_ref().is_ok_and(|record| !record.bytes.is_empty()) {
-                last_successful_read_pos = reader.reader.stream_position()?;
+                last_good_position = reader.reader.stream_position()?;
             } else {
                 match record {
                     Ok(_record) => warn!(
@@ -373,11 +405,34 @@ impl File {
                         err,
                     ),
                 }
-                OpenOptions::new().write(true).open(&self.path)?.set_len(last_successful_read_pos)?;
-                self.len = last_successful_read_pos;
+                self.truncate_from_position(last_good_position)?;
                 break;
             }
         }
+
+        Ok(())
+    }
+
+    fn find_last_position_before(
+        &self,
+        sequence_number: DurabilitySequenceNumber,
+    ) -> Result<Option<u64>, DurabilityServiceError> {
+        let mut reader = FileReader::new(self.clone())?;
+        let mut last_good_position = 0;
+
+        while let Some(record) = reader.read_one_record()? {
+            if record.sequence_number.number() == sequence_number.number() {
+                return Ok(Some(last_good_position));
+            }
+            last_good_position = reader.reader.stream_position()?;
+        }
+
+        Ok(None)
+    }
+
+    fn truncate_from_position(&mut self, position: u64) -> Result<(), DurabilityServiceError> {
+        OpenOptions::new().write(true).open(&self.path)?.set_len(position)?;
+        self.len = position;
         Ok(())
     }
 
--- a/durability/durability.rs
+++ b/durability/durability.rs
@@ -48,6 +48,8 @@ pub trait DurabilityService {
         record_type: DurabilityRecordType,
     ) -> Result<Option<RawRecord<'static>>, DurabilityServiceError>;
 
+    fn truncate_from(&self, sequence_number: DurabilitySequenceNumber) -> Result<(), DurabilityServiceError>;
+
     fn delete_durability(self) -> Result<(), DurabilityServiceError>;
 
     fn reset(&mut self) -> Result<(), DurabilityServiceError>;
--- a/storage/durability_client.rs
+++ b/storage/durability_client.rs
@@ -5,7 +5,6 @@
  */
 
 use std::{
-    borrow::Cow,
     io::{self, Read, Write},
     sync::{mpsc, Arc},
 };
@@ -95,6 +94,8 @@ pub trait DurabilityClient {
     fn find_last_unsequenced_type<Record: UnsequencedDurabilityRecord>(
         &self,
     ) -> Result<Option<Record>, DurabilityClientError>;
+
+    fn truncate_from(&self, sequence_number: SequenceNumber) -> Result<(), DurabilityClientError>;
 
     fn delete_durability(self) -> Result<(), DurabilityClientError>;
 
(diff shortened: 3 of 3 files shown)


compiler/annotation/mod.rs
```diff
@@ -234,7 +234,7 @@ pub mod tests {
         EncodingKeyspace,
     };
     use storage::{durability_client::WALClient, MVCCStorage};
-    use test_utils::{create_tmp_dir, init_logging, TempDir};
+    use test_utils::{create_tmp_storage_dir, init_logging, TempDir};
```
farost, line 237: I reuse `create_tmp_dir` in `cluster` to avoid duplications.

durability/BUILD
```diff
@@ -4,20 +4,7 @@
...
-    "//durability/benches:__subpackages__",
-    "//encoding:__subpackages__",
-    "//executor:__subpackages__",
-    "//function:__subpackages__",
-    "//ir:__subpackages__",
-    "//query:__subpackages__",
-    "//storage:__subpackages__",
-    "//tests:__subpackages__",
-])
+package(default_visibility = ["//visibility:public"])
```
farost, line 7: I need WAL and WALClient in cluster.

durability/wal.rs
```diff
@@ -187,17 +187,26 @@ impl DurabilityService for WAL {
         Ok(None)
     }
 
+    fn truncate_from(&self, sequence_number: DurabilitySequenceNumber) -> Result<(), DurabilityServiceError> {
+        let mut files = self.files.write().unwrap();
+        files.truncate_from(sequence_number)?;
+        // TODO: Should we not fsync the while directory instead since it got deleted files?
```
farost: We probably should...

durability/wal.rs
```diff
@@ -309,9 +318,32 @@ impl Files {
...
+        self.files.iter().rposition(|f| f.start.number() <= sequence_number.number())
+    }
+
+    fn truncate_from(&mut self, sequence_number: DurabilitySequenceNumber) -> Result<(), DurabilityServiceError> {
+        let Some(file_index) = self.file_index_containing(sequence_number) else {
+            return Ok(());
+        };
+
+        let truncate_position = self.files[file_index]
+            .find_last_position_before(sequence_number)?
```
farost: To exist with an error before deleting the files

durability/wal.rs
```diff
@@ -357,12 +389,12 @@ impl File {
         Ok(Self { start: DurabilitySequenceNumber::from(num), len, path })
     }
 
-    fn trim_corrupted_tail(&mut self) -> Result<(), DurabilityServiceError> {
+    fn trim_corrupted_tail_if_needed(&mut self) -> Result<(), DurabilityServiceError> {
         let mut reader = FileReader::new(self.clone())?;
-        let mut last_successful_read_pos = 0;
```
farost, line 362: I don't think that the rename is necessary, but I thought that the new name is better at the moment lol

durability/wal.rs
```diff
@@ -373,11 +405,34 @@ impl File {
...
+    }
+
+    fn find_last_position_before(
+        &self,
+        sequence_number: DurabilitySequenceNumber,
+    ) -> Result<Option<u64>, DurabilityServiceError> {
+        let mut reader = FileReader::new(self.clone())?;
+        let mut last_good_position = 0;
+
+        while let Some(record) = reader.read_one_record()? {
```
farost, line 452: Can we do anything smarter? I don't think so

storage/recovery/checkpoint.rs
```diff
@@ -206,13 +206,48 @@ fn find_latest_checkpoint(checkpoint_dir: &Path) -> Result<Option<PathBuf>, Chec
     }
 
     fs::read_dir(checkpoint_dir)
-        .and_then(|mut entries| entries.try_fold(None, |cur, entry| Ok(cur.max(Some(entry?.path())))))
+        .and_then(|mut entries| {
+            entries.try_fold(None, |cur, entry| {
+                let entry_path = entry?.path();
+                let is_complete =
+                    entry_path.is_dir() && entry_path.join(Checkpoint::STORAGE_METADATA_FILE_NAME).exists();
```
farost, line 215: We need the final file to exist

durability/wal.rs
farost: I haven't found how we currently do the directories fsyncs. We do call
```
            context.files.write().unwrap().sync_all();
```
which is
```
    pub(crate) fn sync_all(&mut self) {
        self.files.last_mut().unwrap().writer().unwrap().get_mut().sync_all().unwrap()
    }
```
But it does not fsync the directories metadata, which is probably required... It might be another set of changes.

lolski (approved):
Approved from my side. I have less context of the WAL and checkpoint mechanism, so Joshua's input will be appreciated.

durability/BUILD
flyingsilverfin: sounds right to me!

durability/wal.rs
flyingsilverfin: Good use case for gemini - did you try that? looks like it's pretty easy - you open the directory as a file and fsync that:
```
You can fsync a directory in Rust by opening it as a std::fs::File and then calling the .sync_all() method on it. This is the idiomatic and platform-agnostic way to request that the operating system write the directory's metadata changes to durable storage.

On POSIX-compliant systems (like Linux and macOS), this translates to opening the directory and calling the fsync() syscall on its file descriptor.
```

example:
```
use std::fs::{File, self};
use std::io::{self, Write};
use std::path::Path;

/// Opens a directory and calls fsync on it to make its metadata durable.
///
/// # Arguments
///
/// * `path` - The path to the directory to be synced.
fn fsync_dir(path: &Path) -> io::Result<()> {
    // Open the directory itself as a file.
    let dir = File::open(path)?;
    // sync_all() requests that all buffered data and metadata be written to the underlying storage.
    // On POSIX, this corresponds to the fsync() syscall.
    dir.sync_all()?;
    Ok(())
}

fn main() -> io::Result<()> {
    let dir_path = Path::new("temp_dir");
    let file_path = dir_path.join("my_new_file.txt");

    // 1. Create the directory if it doesn't exist.
    fs::create_dir_all(dir_path)?;

    // 2. Create and write data to a new file inside the directory.
    let mut file = File::create(&file_path)?;
    file.write_all(b"Important data that must not be lost!")?;
    println!("Created and wrote to file: {:?}", file_path);

    // 3. (CRITICAL) fsync the file itself.
    // This ensures the *content* of the file is written to the disk.
    file.sync_all()?;
    println!("fsync'd the file's data.");

    // 4. (CRITICAL) fsync the parent directory.
    // This ensures the *metadata* change (the creation of the new file entry)
    // in the directory is written to the disk.
    fsync_dir(dir_path)?;
    println!("fsync'd the parent directory's metadata.");

    println!("\n✅ Durability sequence complete. The file and its entry in the directory are now on disk.");

    // Clean up
    fs::remove_file(file_path)?;
    fs::remove_dir(dir_path)?;

    Ok(())
}
```

durability/wal.rs
flyingsilverfin: i dont mind, let's just keep your version

durability/wal.rs
farost: I read somewhere that it may not be true for Windows, so I avoided it because idk how to test it and I'm afraid of Windows magic. I'll think about it more on Monday

durability/wal.rs
```diff
@@ -309,14 +318,38 @@ impl Files {
...
+        self.files.iter().rposition(|f| f.start.number() <= sequence_number.number())
+    }
+
+    fn truncate_from(&mut self, sequence_number: DurabilitySequenceNumber) -> Result<(), DurabilityServiceError> {
+        let Some(file_index) = self.file_index_containing(sequence_number) else {
+            return Ok(());
+        };
+
+        let Some(truncate_position) = self.files[file_index].find_last_position_before(sequence_number)? else {
+            // Already does not have anything from this sequence number. Can be changed to an error.
```
flyingsilverfin, line 362: Is this right?

Simplified WAL file:
```
File 1 - Seq 0..=Seq10
[index] [seq nr at index]
---
[0] [0]
[1] [1]
[2] [2]
[3] [3]
[4] [4]
...
```
Then call: truncate at Seq 3. Basically, we want Seq 3 and onwards to be gone.

This does:
```
0) find the file that contains Seq 3, throw the later files out
1) On File 1, call find_last_position_before(Seq 3), which would return index 2
2) Then we call File 1.truncate_from_position(2)
```

truncate_from_position sets the file length to 2... wouldn't that mean just keeping index 0, and index 1? And deleting index 2 onwards as well?

durability/wal.rs
```diff
@@ -309,14 +318,38 @@ impl Files {
...
+            return Ok(());
+        };
+
+        while self.files.len() > file_index + 1 {
+            fs::remove_file(&self.files.pop().unwrap().path)?;
+        }
```
flyingsilverfin, line 368: let's do this first as soon as we get the file index, then the second half of the method will be dealing with that one interesting file.

durability/wal.rs
farost: Hmmm I'll write a test, thanks

durability/wal.rs
```diff
@@ -848,4 +904,109 @@ mod test {
...
+
+        wal.truncate_from(s1).unwrap();
+
+        let read_records = wal.iter_any_from(DurabilitySequenceNumber::MIN).unwrap().collect::<Vec<_>>();
+        assert!(read_records.is_empty(), "expected no records after truncate_from(first)");
+        assert_eq!(wal.current(), s1);
+    }
+
+    #[test]
+    fn test_wal_truncate_from_is_idempotent_for_same_cut() {
```
flyingsilverfin, line 1059: So since the WAL can also contain unsequenced writes - can we test that it works with that? 

Basically, write sequenced (seq 1), write unsequenced (X), unsequenced (Y), write sequenced (2). Then trim 2 onwards.

We should still read Seq 1, X, and Y

farost: Ohhh, good point. I forgot about it

durability/wal.rs
farost: I think the names of the methods are a little confusing here. It's not an array, but a file, and one record takes many bytes, each with its own position. So "before" is actually "first to cut", because the cursor moves to the next byte after the last (successful) read. I've adjusted the naming.

I've extended the first test to be a little more complicated with more records, just in case.

durability/wal.rs
farost: What's preceding:
```
let Some(truncate_position) = self.files[file_index].offset_of(sequence_number)? else {
            // Already does not have anything from this sequence number. Can be changed to an error.
            return Ok(());
        };
```
This is the reason why I say "can be changed to an error".
 
We consider that the storage is continuous. So if `file_index_containing` returns a file index, it should probably contain the record:
```
    fn file_index_containing(&self, sequence_number: DurabilitySequenceNumber) -> Option<usize> {
        self.files.iter().rposition(|f| f.start.number() <= sequence_number.number())
    }
```
and if it does not, we probably don't want to touch the files. 

Additionally, the `offset_of` call can return errors, so, if there are issues, we don't want to delete the files yet, the operation should just fail (although we can have the same issue with partial deletion of files, when we get an error in the middle of the process... Not sure what to do here).

So I acted "safely" here. What do you think?

durability/wal.rs
farost: I added `truncate_from_keeps_prior_unsequenced_records_with_same_seq`. It also writes `Z` after the sequenced `2` to check that it's cleaned with it.

Do you think we need to test multiple different types of unsequenced writes? It's not currently tested at all.

durability/wal.rs
flyingsilverfin: RIght i see that - might be worth leaving a comment in the code as to why it's got this specific ordering to be safe. It's a bit nuanced and easy to miss

durability/wal.rs
```diff
@@ -357,12 +390,12 @@ impl File {
...
 
-    fn trim_corrupted_tail(&mut self) -> Result<(), DurabilityServiceError> {
+    fn trim_corrupted_tail_if_needed(&mut self) -> Result<(), DurabilityServiceError> {
         let mut reader = FileReader::new(self.clone())?;
-        let mut last_successful_read_pos = 0;
+        let mut last_good_position = 0;
         while let Some(record) = reader.read_one_record().transpose() {
             if record.as_ref().is_ok_and(|record| !record.bytes.is_empty()) {
-                last_successful_read_pos = reader.reader.stream_position()?;
+                last_good_position = reader.reader.stream_position()?;
```
flyingsilverfin: Let's call this `last_good_position_end`

flyingsilverfin: Then 'truncate_from' makes sense!

durability/wal.rs
```diff
@@ -848,4 +915,159 @@ mod test {
...
+            .try_collect()
+            .unwrap();
+
+        let reads_before_cut = read_all_records_tupled(&wal);
+        assert_eq!(reads_before_cut.len(), 5);
+
+        let cut = seqs[2];
+        wal.truncate_from(cut).expect("Expected to truncate everything starting from seqs[2] (including itself)");
+
+        assert_eq!(wal.current(), seqs[2], "Expected to have the current seq equal to the cut seq");
```
flyingsilverfin, line 968: nice, thanks!

durability/wal.rs
flyingsilverfin: No just test 1 unsequenced write type. thanks!

durability/wal.rs
```diff
@@ -848,4 +915,159 @@ mod test {
...
+        wal.truncate_from(s2).expect("Expected to cut at s2 to remove S222 and UZZZ");
+        assert_eq!(wal.current(), s2, "Expected current to be equal to the cut seq");
+
+        let reads_after_cut = read_all_records_tupled(&wal);
+        assert_eq!(
+            reads_after_cut,
+            vec![
+                (s1, TestRecord::RECORD_TYPE, b"S111".to_vec()),
+                (s1, UnsequencedTestRecord::RECORD_TYPE, b"UXXX".to_vec()),
+                (s1, UnsequencedTestRecord::RECORD_TYPE, b"UYYY".to_vec()),
```
flyingsilverfin, line 1097: good stuff, much  more confident now!

durability/wal.rs
farost: Alright, I think you're right

durability/wal.rs
farost: Oh, forgot about it, sorry. Will revisit tomorrow

storage/recovery/checkpoint.rs
```diff
@@ -206,13 +206,48 @@ fn find_latest_checkpoint(checkpoint_dir: &Path) -> Result<Option<PathBuf>, Chec
...
+        let mut tmp_file = File::create(&tmp)?;
+        let mut source_file = File::open(source)?;
+        std::io::copy(&mut source_file, &mut tmp_file)?;
+        tmp_file.sync_all()?;
+    }
+    // The directory is not fsynced, so there is a chance that the copied file won't be persisted.
+    // However, if the file is found, it's correct and complete.
+    fs::rename(&tmp, destination)
+}
+
```
flyingsilverfin, line 252: do we have a routine on bootup where we delete any files ending in `.tmp`? Maybe we should also introduce a shared constant for TEMP_FILE_EXTENSION. THis can then be used on bootup to look for incomplete states & cleanup.

flyingsilverfin (approved):
Approved with 1 or 2 comments to address

storage/recovery/checkpoint.rs
farost: Good to have. I considered this to be extremely rare, with the same chance as for our current system crashing with checkpoint files left corrupted. Since it's rare, I didn't mind leaving them just like checkpoints. But generalising the approach is a good thing, good call.

durability/wal.rs
farost: After some time of reading, I can say that Windows does behave differently, and this approach is not enough to make sure that the files are persisted with full accuracy. Some of these operations may even fail.

The code generated by AI to try fsyncing the dir on Windows looks weird, and I decided not to introduce it into the codebase. I could not find how it is usually done (and am surprised that it's still not solved in the language). Some say that "modern OSs handle it automatically and the chances of a loss are miserable" (the documentation does not share it, which makes sense), some say that there is no 100% way (weird).
TIKV (another key-value storage written in Rust) do not care, they just fsync the dir: https://github.com/tikv/tikv/blob/master/components/external_storage/src/local.rs#L144
https://github.com/tikv/tikv/blob/master/components/encryption/src/file_dict_file.rs#L153

Since ignoring errors on fsyncs in WAL is risky (we want to make sure that the files are removed), I've decided to introduce a unix-specific checked action and a windows-specific unchecked action. The windows-specific action can be replaced with another code in the future when we decide to test it thoroughly. For now, I want to cover 99.99% cases (considering the probability of an OS crash when we do these truncations and fsync on initialization).

storage/recovery/checkpoint.rs
farost: I had another look. I don't think it's needed since we remove old checkpoints in `finish`, and all temporary files (I expect at most one temporary file, but in 99.9999% cases there will be 0) will be eventually deleted, when a newer correct checkpoint is created. Startup cleaning sounds a little complicated, so I'll leave it as is.

Issue #7550, opened 2025-08-15
Recursive function with 'first' causes panic


flyingsilverfin:
## Description

A recursive function with a `first` return statement causes an unexpected error.

## Environment

1. TypeDB distribution: CE
2. TypeDB version: 3.4.3

## Reproducible Steps

1. Set up

```
define
fun friendship_hops($start: user, $end: user) -> integer:
match
  {
    friendship (friend: $start, friend: $end);
    let $hops = 1;
  } or {
    friendship (friend: $start, friend: $mid);
    let $hops = 1 + friendship_hops($mid, $end);
  };
  return first $hops;
```

```
insert $u1 isa user, has name "A"; $u2 isa user, has name "B"; $u3 isa user, has name "C"; $u4 isa user, has name "D";
friendship (friend: $u1, friend: $u2);
friendship (friend: $u2, friend: $u3);
friendship (friend: $u3, friend: $u4);
```

2. Execute

```
match $u1 isa user, has name "A"; $end isa user, has name "D"; let $hops = friendship_hops($u1, $end);
```

3. Unexpected result
```
thread 'tokio-runtime-worker' panicked at executor/read/pattern_executor.rs:80:9:
assertion failed: self.control_stack.is_empty()
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:75:14
   2: core::panicking::panic
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:145:5
   3: executor::read::pattern_executor::PatternExecutor::prepare_to_restore_from_suspension
             at ./executor/read/pattern_executor.rs:80:9
```

## Expected result
Return '3' only.

## Additional information
Note that it works if we remove 'first':

```
redefine
fun friendship_hops($start: user, $dest: user) -> { integer }:
match
  {
    friendship (friend: $start, friend: $dest);
    let $hops = 1;
  } or {
    friendship (friend: $start, friend: $mid);
    let $hops = 1 + friendship_hops($mid, $dest);
  };
  return { $hops };
```
```
> match $u1 isa user, has name "A"; $end isa user, has name "D"; let $hops in friendship_hops_dfs($u1, $end); limit 1;

[...]

PR #7561, opened 2025-08-26
Enforce stratification through single return functions

krishnangovindraj:
Single return functions may not call themselves recursively.  Addresses #7550 

This will now fail:
```typeql
  define
    fun last_number() -> integer:
    match
      let $number = last_number() + 1;
    return last $number;
```


flyingsilverfin:
I think this is the cleanest. Recursive functions can't inside the recursion do anything that requires "consuming" the stream IMO. We can consider anything that isn't returning the stream unmodified as illegal?

flyingsilverfin (approved):

PR #7562, opened 2025-08-27
Allow empty define

flyingsilverfin:

Since https://github.com/typedb/typeql/pull/412 is implemented, empty define queries are allowed. This means that schema export that used to return an empty string for empty schemas can now return an empty `define` query, which is much easier to work with on the client side.

This addresses https://github.com/typedb/typedb/issues/7531

## Implementation

- update dependencies with latest TypeQL and Behaviour changes
- empty schema export uses an empty `define` to indicate an empty schema, instead of an empty string

### DIFF
--- a/server/service/export_service.rs
+++ b/server/service/export_service.rs
@@ -14,14 +14,14 @@ pub(crate) fn get_transaction_schema<D: DurabilityClient>(
 ) -> Result<String, DatabaseExportError> {
     let types_syntax = get_types_syntax(transaction)?;
     let functions_syntax = get_functions_syntax(transaction)?;
-    Ok(format!("{}\n{}{}\n", typeql::token::Clause::Define, types_syntax, functions_syntax))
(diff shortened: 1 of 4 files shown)


server/service/export_service.rs
```diff
@@ -14,24 +14,14 @@ pub(crate) fn get_transaction_schema<D: DurabilityClient>(
...
     transaction: &TransactionRead<D>,
 ) -> Result<String, DatabaseExportError> {
     let types_syntax = get_types_syntax(transaction)?;
-
-    let type_schema = match types_syntax.is_empty() {
-        true => String::new(),
-        false => format!("{}\n{}\n", typeql::token::Clause::Define, types_syntax),
-    };
-    Ok(type_schema)
+    Ok(format!("{}\n{}\n", typeql::token::Clause::Define, types_syntax))
```
farost: Are we sure? I'd expect an empty string for an empty schema, not `\n\n`...

flyingsilverfin: TypeQL now allows `define ` as a "valid" query, so we just keep `define` (with some whitespace, doesn't matter much) as the output!

PR #7564, opened 2025-08-28
Work around brew install python breaking our CircleCI pipeline for mac

krishnangovindraj:
The `brew install python@3.9` step fails at the postinstall step and breaks our mac CircleCI pipeline on mac. We work around  this by manually running the postinstall step.


krishnangovindraj:
Oh the other one is merged. That's why I didn't see it

PR #7566, opened 2025-09-02
Lock attributes on updating connections

flyingsilverfin:

We fix an Isolation bug that is exposed under concurrent update (adding an ownership) + concurrent delete transactions. This is a relatively uncommon conflict, and is fixed  by locking attributes that are edited in order to conflict with concurrent deletes of the same attribute. Under highly concurrent operations that include `delete`s, this change might manifest itself as more transaction conflicts `STC2` (Storage Commit 2) errors, which can be resolved with a retry.

## Implementation

We mark attributes that are modified (eg. given new ownerships) as required by locking them, rather than re-putting them into the snapshot.

We also fix a similar issue with links edges, by marking the player and the relation as locked so a concurrent delete of either will not lead a dangling edge.

## Details

Scenario: attribute A exists beforehand with 1 owner X

Tx1: opens at SeqNr 2
  --> deletes last ownership it sees from X to A
  --> commits
  --> auto-delete dependent attribute A
  --> commits, gets sequence number 8

Tx2: opens at SeqNr 4
  --> adds ownership of A from owner Y, and re-puts A
  --> commits
  --> commits at seq nr 6
  --> NO ERROR, and commits first !

Tx1: NO ERROR. Even though we delete attribute A, we only conflict "locks" and "deletes". We didn't set attributes as "locked" when we add attribute ownerships to them, but we do re-insert them - but this isn't enough!

### DIFF
--- a/concept/thing/attribute.rs
+++ b/concept/thing/attribute.rs
@@ -24,7 +24,7 @@ use encoding::{
         Typed,
     },
     layout::prefix::Prefix,
-    value::{value::Value, value_type::ValueType},
+    value::{value::Value},
     AsBytes, Keyable,
 };
 use iterator::State;
@@ -143,23 +143,12 @@ impl ThingAPI for Attribute {
         snapshot: &mut impl WritableSnapshot,
         thing_manager: &ThingManager,
         _storage_counters: StorageCounters,
-    ) -> Result<(), Box<ConceptReadError>> {
-        match self.type_().get_value_type_without_source(snapshot, thing_manager.type_manager())? {
-            Some(value_type) => match value_type {
-                | ValueType::Boolean
-                | ValueType::Integer
-                | ValueType::Double
-                | ValueType::Decimal
-                | ValueType::Date
-                | ValueType::DateTime
-                | ValueType::DateTimeTZ
-                | ValueType::Duration => snapshot.put(self.vertex().into_storage_key().into_owned_array()),
-                // ValueTypes with expensive writes
-                | ValueType::String | ValueType::Struct(_) => thing_manager.lock_existing_attribute(snapshot, self),
-            },
-            None => panic!("Attribute instances must have a value type"),
-        }
-        Ok(())
+    ) {
+        // always lock the attribute, even if put in this snapshot since we don't know
+        // if it pre-existed or concurrently created/deleted, or even put again later in this transaction
+        // contrast this with entities/relations:
+        // those cannot be shared or concurrently created by other transactions unless pre-existing as Persisted
+        thing_manager.lock_existing_attribute(snapshot, self);
     }
 
     fn get_status(
@@ -254,7 +243,7 @@ where
         }
     }
 
-    pub fn seek(&mut self, target: &Attribute) {
+    pub fn seek(&mut self, _target: &Attribute) {
         todo!()
     }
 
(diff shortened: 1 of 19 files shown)


concept/thing/attribute.rs
```diff
@@ -143,23 +143,12 @@ impl ThingAPI for Attribute {
...
-            },
-            None => panic!("Attribute instances must have a value type"),
-        }
-        Ok(())
+    ) {
+        // always lock the attribute, even if put in this snapshot since we don't know
+        // if it pre-existed or concurrently created/deleted, or even put again later in this transaction
+        // contrast this with entities/relations:
+        // those cannot be shared or concurrently created by other transactions unless pre-existing as Persisted
+        thing_manager.lock_existing_attribute(snapshot, self);
```
krishnangovindraj, line 151: What was the reason to only lock strings & structs initially? Is it to avoid the unlikely hash disambiguation collision? 

Can we record this so we don't forget to add it back in case we decide we don't need to lock attributes? (seems unlikely we do, but still)

krishnangovindraj (approved):
This change now escalates most writes to locks. It is a major change from what we used to do, but seems inevitable given:
1. We've seen bugs resulting from inserting edges involviung vertices that were concurrently deleted.
2. The storage layer is unaware of the relationship between edges & vertices, thus requiring some operation on a the vertex to be recorded.

Isolation checking should remain linear in the number of writes in the transaction (we were already linear for dependent-write checking). It may even be possible to [optimise isolation checking further](https://github.com/typedb/typedb/issues/7567).

I'd like @dmitrii-ubskii to do a post-review anyway. I'm a little surprised we didn't catch these cases earlier.

concept/thing/attribute.rs
flyingsilverfin: I think ihad done that because we can't do a simple re-insertion of the strings and attributes because that might be much more expensive than just writing a lock

dmitrii-ubskii:
This breaks a major benefit of always `put`ting an attribute. To slightly reorder the events in the PR description:

- Tx1 opens at SeqNr 2
- Tx1 `insert $y has $a;` (unmodifiable lock on `$a`)
- Tx2 opens at SeqNr 4
- Tx2 `delete has $a from $x;`
- Tx2 commits, sees that it's deleted the last owner of `$a` as far as it's aware and deletes the attribute as well
- Tx1 commits
- Tx1 conflicts: `"Transaction uses data a concurrent commit deletes"` (`$a`)

Except of course it doesn't. Tx1 should simply reinsert the attribute and proceed as normal.

---

There's also an unergonomic side effect to this change. Running through the example as in the PR:

Tx1 opens at SeqNr 2
Tx1 `delete has $a from $x;`
Tx2 opens at SeqNr 4
Tx2 `insert $y has $a;` (unmodifiable lock on `$a`)
Tx2 commits
Tx1 commits, sees that it's deleted the last owner of `$a` and deletes it
Tx1 conflicts: `"Transaction deletes data a concurrent commit requires"` (`$a`)

The thing is: no it doesn't, at least not as far as the user is concerned. The user did not delete anything, and if `$a` happened to have another owner `$z` this would succeed just fine. We're the ones causing the transaction to fail.

krishnangovindraj:
^This is fair, but we'd need a solution that also handles the ordering in the OP. i.e., the deleter committing second and leaving dangling edges.

PR #7568, opened 2025-09-08
Fix missing word in isolation error message

dmitrii-ubskii:

The isolation conflict message when deleting a required key reads "Transaction data a concurrent commit requires."

## Implementation

Add the missing word ("deletes").


krishnangovindraj (approved):
trivial

PR #7574, opened 2025-09-10
Fix development mode configuration in config file

krishnangovindraj:
Fixes development mode configuration in config file

### DIFF
--- a/server/parameters/config.rs
+++ b/server/parameters/config.rs
@@ -25,7 +25,7 @@ pub struct Config {
     #[serde(default)]
     pub diagnostics: DiagnosticsConfig,
     pub logging: LoggingConfig,
-    #[serde(rename = "development-mode.enabled", default)]
+    #[serde(rename = "development-mode", default)]
     pub development_mode: DevelopmentModeConfig,
 }
 


server/parameters/config.rs
```diff
@@ -25,7 +25,7 @@ pub struct Config {
     #[serde(default)]
     pub diagnostics: DiagnosticsConfig,
     pub logging: LoggingConfig,
-    #[serde(rename = "development-mode.enabled", default)]
+    #[serde(rename = "development-mode", default)]
```
farost: I don't think it's even needed as
`#[serde(rename_all = "kebab-case")]`

krishnangovindraj: Good point

server/parameters/config.rs
```diff
@@ -25,7 +25,7 @@ pub struct Config {
     #[serde(default)]
     pub diagnostics: DiagnosticsConfig,
     pub logging: LoggingConfig,
-    #[serde(rename = "development-mode.enabled", default)]
+    #[serde(default)]
```
farost, line 28: If it's forced (unpublished), then it will be overridden. If it's not forced, then, by default, it's false. If it's set, it's set however the user wants.
Sounds good.

PR #7575, opened 2025-09-11
Fix crash when functions sorts empty stream

### LINKED ISSUE #7573: Assertion error in sorting
## Description

Assertion error causes crash in sorting.

## Environment

TypeDB CE 3.5.0

## Reproducible Steps

1. Set up

Boot TypeDB 3.5.0, and load the bookstore example:

2. Execute
```
with fun first_contributor($book: book) -> { contributor }:
  match
  $contributor isa contributor, has name $name;
  authoring ($contributor, $book);
  sort $name;
  limit 1;
  return { $contributor };
match
$book isa book, has title $title;
let $first-contributor = first_contributor($book);
$first-contributor has name $name;
sort $title, $name;
fetch {
  "title": $title,
  "contributor": $name,
};
```

3. Unexpected result
```

Ready!
thread 'tokio-runtime-worker' panicked at executor/read/collecting_stage_executor.rs:217:34:

```

krishnangovindraj:
Fixes a crash when a functions tried to sort an empty stream.

## Implementation
Does some refactoring so the `output_width` is known when the collector is created, rather than having to wait for the first row (which may not come, and caused the collecting batch to be uninitialised).

- Fixes #7573

### DIFF
--- a/executor/read/stream_modifier.rs
+++ b/executor/read/stream_modifier.rs
@@ -274,7 +274,9 @@ impl LastMapper {
 impl StreamModifierResultMapperTrait for LastMapper {
     fn map_output(&mut self, subquery_result: Option<FixedBatch>) -> Option<FixedBatch> {
         if let Some(input_batch) = subquery_result {
-            self.last_row = Some(input_batch.get_row(input_batch.len() - 1).into_owned());
+            self.last_row = (!input_batch.is_empty()).then(|| {
+                input_batch.get_row(input_batch.len() - 1).into_owned()
+            });
             Some(FixedBatch::EMPTY) // Retry this instruction without returning any rows
         } else {
             self.last_row.take().map(FixedBatch::from)


compiler/executable/pipeline.rs
```diff
@@ -397,14 +397,6 @@ fn compile_stage(
             ExecutableStage::Offset(Arc::new(OffsetExecutable::new(offset.offset(), stage_input_positions.clone()))),
             BTreeSet::new(),
         )),
-        AnnotatedStage::Sort(sort) => Ok((
```
krishnangovindraj, line 400: Strangely duplicated.

executor/batch.rs
```diff
@@ -120,6 +120,8 @@ impl<'a> From<MaybeOwnedRow<'a>> for FixedBatch {
 }
 
 impl IntoIterator for FixedBatch {
+    type Item = MaybeOwnedRow<'static>;
+
```
krishnangovindraj, line 124: linter wanted to reorder.

executor/read/collecting_stage_executor.rs
```diff
@@ -126,7 +137,6 @@ pub(super) trait CollectedStageIteratorTrait {
 // Reduce
 pub(super) struct ReduceCollector {
     active_reducer: GroupedReducer,
-    output: Option<BatchRowIterator>,
```
krishnangovindraj, line 129: Unused, and probably forgotten after a previous refactor

executor/read/stream_modifier.rs
```diff
@@ -22,7 +22,7 @@ pub(crate) enum StreamModifierExecutor {
     Select { inner: PatternExecutor, removed_positions: Vec<VariablePosition> },
     Offset { inner: PatternExecutor, offset: u64 },
     Limit { inner: PatternExecutor, limit: u64 },
-    Distinct { inner: PatternExecutor, output_width: u32 },
```
krishnangovindraj, line 25: Current implementation doesn't need to know.

executor/read/stream_modifier.rs
```diff
@@ -266,7 +274,9 @@ impl LastMapper {
 impl StreamModifierResultMapperTrait for LastMapper {
     fn map_output(&mut self, subquery_result: Option<FixedBatch>) -> Option<FixedBatch> {
         if let Some(input_batch) = subquery_result {
-            self.last_row = Some(input_batch.get_row(input_batch.len() - 1).into_owned());
+            self.last_row = (!input_batch.is_empty()).then(|| {
+                input_batch.get_row(input_batch.len() - 1).into_owned()
+            });
```
krishnangovindraj, line 279: Possible edge case in the case of an empty batch

compiler/executable/pipeline.rs
flyingsilverfin: How is that possible when it i was a `match` clause??

krishnangovindraj: Turned yellow as "unreachable" but didn't break.

PR #7584, opened 2025-09-23
Add BDD testing for analyzing queries

krishnangovindraj:
Tests the query structure returned by the analyze endpoint (and used by studios' graph visualizer)

## Implementation
* Implements steps & functor encoding used in typedb/typedb-behaviour#381
* Small fixes around.

### DIFF
--- a/dependencies/typedb/repositories.bzl
+++ b/dependencies/typedb/repositories.bzl
@@ -35,6 +35,6 @@ def typedb_protocol():
 def typedb_behaviour():
     git_repository(
         name = "typedb_behaviour",
-        remote = "https://github.com/krishnangovindraj/typedb-behaviour",
-        commit = "638ab2ea88efd0d3ad205ce5f2e9ca4d9371c722",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        remote = "https://github.com/typedb/typedb-behaviour",
+        commit = "9376d6983dc71e3bae9caaaa0bb17fc063e11c20",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
     )


server/service/http/message/query/concept.rs
```diff
@@ -121,6 +121,13 @@ serializable_response! {
     }
 }
 
+serializable_response! {
+    pub struct TypeFailedInferenceResponse {
+        kind = "failedInference",
+        pub label: String => "label",
+    }
+}
+
```
krishnangovindraj, line 88: This was a bad idea, and has to be absorbed into StructureVertex.

krishnangovindraj:
A fairly obvious but difficult to achieve UX improvement here is to turn the preamble into a map from function name to structure / annotations.

server/service/http/message/query/query_structure.rs
```diff
@@ -237,6 +236,7 @@ enum StructureVertex {
     Variable { id: StructureVariableId },
     Label { r#type: serde_json::Value },
     Value(ValueResponse),
+    FailedTypeInference { label: String },
```
flyingsilverfin: Wouldn't this just be returned as an error? Why do we need this?

server/service/http/message/query/query_structure.rs
```diff
@@ -584,3 +579,247 @@ fn encode_role_type_as_vertex(
...
+    ]);
+
+    pub fn encode_query_structure_as_functor(structure: &QueryStructureResponse) -> (String, Vec<String>) {
+        let pipeline = &structure.query;
+        let query = pipeline.encode_as_functor(pipeline);
+        let preamble = structure.preamble.iter().map(|func| func.encode_as_functor(&func.body)).collect();
+        (query, preamble)
+    }
+
+    #[cfg(test)]
```
flyingsilverfin: Is this the start of BDD implementation in the server for the query structure?

tests/behaviour/steps/query/typeql.rs
```diff
@@ -27,13 +27,17 @@ use itertools::{Either, Itertools};
 use lending_iterator::LendingIterator;
 use macro_rules_attribute::apply;
 use params;
-use query::error::QueryError;
+use query::{analyse::AnalysedQuery, error::QueryError};
```
flyingsilverfin, line 30: Do we want to do a switch to analyze since it's kinda awkward now eh

tests/behaviour/steps/query/typeql.rs
```diff
@@ -200,6 +204,25 @@ fn execute_write_query(
...
 }
 
+fn execute_analyze_query(
+    context: &mut Context,
+    query: typeql::Query,
+    source_query: &str,
+) -> Result<AnalysedQuery, BehaviourTestExecutionError> {
+    with_read_tx!(context, |tx| {
+        tx.query_manager
+            .analyse_query(
```
flyingsilverfin: i think we can just name the method `.analyse`, it's not exactly a Query

server/service/http/message/query/query_structure.rs
krishnangovindraj: Thanks, I think I might be able to remove it indeed.

tests/behaviour/steps/query/typeql.rs
krishnangovindraj: Can be a follow up PR to keep the diff small.

server/service/http/message/query/query_structure.rs
krishnangovindraj: No, pub mod bdd is, much higher. This is a unit test for the functor encoding.
I could rearrange a lot of this.

tests/behaviour/steps/query/typeql.rs
flyingsilverfin: sounds good

compiler/annotation/type_annotations.rs
```diff
@@ -47,14 +59,15 @@ impl BlockAnnotations {
 pub struct TypeAnnotations {
     vertex: BTreeMap<Vertex<Variable>, Arc<BTreeSet<Type>>>,
     constraints: HashMap<Constraint<Variable>, ConstraintTypeAnnotations>,
+    value_type_annotations: Option<BTreeMap<Vertex<Variable>, ExpressionValueType>>,
```
flyingsilverfin, line 62: Yeah this makes a lotta sense IMO! good add!

server/service/http/message/query/annotations.rs
```diff
@@ -290,3 +268,169 @@ fn encode_fetch_object_structure_annotations(
     };
     Ok(encoded)
 }
+
+#[cfg(debug_assertions)]
+pub mod bdd {
```
flyingsilverfin, line 286: i'm a bit confused here: there's no actual tests right?

tests/behaviour/steps/query/typeql.rs
flyingsilverfin: Note: still identical here

tests/behaviour/steps/query/typeql.rs
```diff
@@ -601,3 +626,96 @@ async fn verify_answer_set(context: &mut Context, step: &Step) {
...
+    let analyzed_unencoded = execute_analyze(context, query, query_str).unwrap();
+    let analyzed = with_read_tx!(context, |tx| {
+        encode_analyzed_query(&(*tx.snapshot), &tx.type_manager, analyzed_unencoded).unwrap()
+    });
+    context.analyzed_query = Some(analyzed);
+}
+
+#[apply(generic_step)]
+#[step(expr = r"typeql analyze query{typeql_may_error}")]
+async fn typeql_analyze_query_may_error(context: &mut Context, may_error: params::TypeQLMayError, step: &Step) {
```
flyingsilverfin, line 644: So this means by the way that the analyze executions can't just be `driver` tests right? We actually don't have have an overlap of tests between server and driver right now i think! We might need to define a third set of tests in BDD somewhere ?

krishnangovindraj: I don't understand. I have a small set of analyze tests in the `driver/query.feature`.

server/service/http/message/query/annotations.rs
krishnangovindraj: No, but this is only used by the bdd.

PR #7587, opened 2025-09-24
Make state initialisation and loading extensible

lolski:

The state initialisation and loading have been made extensible as needed by TypeDB Cluster.

## Implementation

- Refactor state initialise / load methods in order to make them extensible
- Make user creation method in user management return a commit record for extensibility. This way, the commit record can be handled either by local or cluster-wide database commit procedure
- Do database management operations through the "unrestricted" method counterparts, so that they can be used for performing system database initialisation
- Fix a bug where the 'load' procedure is never triggered. Instead what had always happened was that the initialise procedure is always called. It happened to work since the initialise method is idempotent.
- Fix a bug where the system database isn't loaded in the load procedure


system/util.rs
```diff
@@ -34,6 +35,36 @@ pub mod transaction_util {
...
+                Arc::new(snapshot),
+                type_manager,
+                thing_manager,
+                function_manager,
+                query_manager,
+                database,
+                transaction_options,
+                profile,
+            );
+            let (profile, result) = tx.finalise();
```
farost, line 61: Can just return `tx.finalise()`

system/util.rs
```diff
@@ -78,6 +109,51 @@ pub mod transaction_util {
...
+                snapshot,
+                type_manager,
+                thing_manager,
+                function_manager,
+                query_manager,
+                database,
+                TransactionOptions::default(),
+                profile,
+            );
+            let (profile, result) = tx.finalise();
```
farost, line 142: Can just return `tx.finalise()`

server/system_init.rs
```diff
@@ -0,0 +1,80 @@
...
+    let commit_record_opt = snapshot.finalise(&mut commit_profile)
+        .map_err(|error| LocalServerStateError::DatabaseSchemaCommitFailed { typedb_source: SchemaCommitError::SnapshotError { typedb_source: error }})?;
+    return Ok((transaction_profile, commit_record_opt));
+}
+
+pub async fn initialise_default_user(user_manager: &user::user_manager::UserManager, server_state: &dyn ServerState) -> Result<(), ArcServerStateError> {
+    if let (mut transaction_profile, Some(commit_record)) = get_default_user_commit_record(user_manager).await? {
+        server_state.database_data_commit(SYSTEM_DB, commit_record, &mut transaction_profile.commit_profile()).await?;
+    }
+    Ok(())
```
farost, line 79: Is the `(_, None)` case acceptable, when there is no `CommitRecord`?

server/service/grpc/typedb_service.rs
```diff
@@ -529,11 +534,43 @@ impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
...
-                    .users_create(&user, &credential, accessor)
-                    .await
-                    .map(|_| Response::new(user_create_res()))
-                    .map_err(|err| err.into_error_message().into_status())
+
+                if !PermissionManager::exec_user_create_permitted(accessor.0.as_str()) {
+                    return Err(LocalServerStateError::OperationNotPermitted {}.into_error_message().into_status());
+                }
+
+                let (mut transaction_profile, create_result) = match self.server_state.user_manager().await {
```
farost, line 542: So `ServerState` does not "generalise" multiple services anymore, and these services should implement the same logic, sometimes through copypaste, sometimes creating `LocalServerStateError` themselves.
Yeah, the design is not very good, but that's what we have now. Refactoring will be expensive.

The fact that you already started missing `PermissionManager` calls shows that it adds complexities and is quite dangerous.

PR #7589, opened 2025-09-25
Fix variable name unwrap when creating error message in insert executable

### LINKED ISSUE #7588: Failed to fetch" when running queries in TypeDB Studio, even after upgrading to 3.5.1
## Description
I encountered a persistent issue when running queries in TypeDB Studio.
The error shown is:

Error: Failed to fetch

At first, the system prompted me to upgrade the cluster (from 3.5.0 to 3.5.1).
After upgrading successfully to 3.5.1, the problem still occurs — the queries cannot be executed, and the same error keeps appearing in the bottom right corner of Studio.

## Environment

TypeDB distribution: Cloud
TypeDB version: Initially 3.5.0, then upgraded to 3.5.1 (issue persists)
Environment: TypeDB Cloud (AWS, US East region)
Client and version: TypeDB Studio (web GUI in Cloud)
Other details: Queries are simple match + insert statements based on the schema. The failure happens immediately after clicking Run.

## Reproducible Steps

Create a schema (entities, attributes, relations).
Open TypeDB Studio in Cloud.
Run a query such as:
match
    $source isa knowledge-source, has source-identifier "psu-bulletin-ug-2025-cs";
    $cs_major isa major, has major-name "Computer Science";
    $req(major: $cs_major, course: $course) isa major-requirement;
    not { $req(sourced-by: $any_source) isa major-requirement; };
insert
    (major: $cs_major, course: $course, sourced-by: $source) isa major-requirement;
Observe that Studio shows Error: Failed to fetch instead of a proper query error or execution result.

Image
Image

## Expected result
Query execution fails with Error: Failed to fetch in the Studio UI.
No detailed error message is returned.
The issue persists both before and after upgrading from 3.5.0 → 3.5.1.

krishnangovindraj:
Fixes an unwrap when reading the variable name for returning an error message

## Implementation
Also replaces other usages of get_variable_name / or replaces unwrap with expect justifying safety.

- fixes #7588

### DIFF
--- a/compiler/annotation/fetch.rs
+++ b/compiler/annotation/fetch.rs
@@ -167,7 +167,7 @@ fn annotate_some(
         FetchSome::SingleVar(var) => Ok(AnnotatedFetchSome::SingleVar(var)),
         FetchSome::SingleAttribute(FetchSingleAttribute { variable, attribute }) => {
             let variable_name = variable_registry.get_variable_name(variable)
-                .expect("Fetched variable names were validated during translation");
+                .expect("Expected fetched variable names to be validated during translation");
             let attribute_type = type_manager
                 .get_attribute_type(snapshot, &attribute)
                 .map_err(|err| AnnotationError::ConceptRead { typedb_source: err })?
@@ -240,7 +240,7 @@ fn annotate_some(
         }
         FetchSome::ListAttributesAsList(FetchListAttributeAsList { variable, attribute }) => {
             let variable_name = variable_registry.get_variable_name(variable)
-                .expect("Fetched variable names were validated during translation");
+                .expect("Expected fetched variable names to be validated during translation");
             let attribute_type = type_manager
                 .get_attribute_type(snapshot, &attribute)
                 .map_err(|err| AnnotationError::ConceptRead { typedb_source: err })?


compiler/annotation/fetch.rs
```diff
@@ -166,7 +166,8 @@ fn annotate_some(
     match some {
         FetchSome::SingleVar(var) => Ok(AnnotatedFetchSome::SingleVar(var)),
         FetchSome::SingleAttribute(FetchSingleAttribute { variable, attribute }) => {
-            let variable_name = variable_registry.get_variable_name(variable).unwrap();
+            let variable_name = variable_registry.get_variable_name(variable)
+                .expect("Fetched variable names were validated during translation");
```
farost: If this fails, the user will see this message: "Fetched variable names were validated during translation"
The canonical form for expects is "Expected to have fetched variable names validated during translation", it makes more sense as an error message. As they actually are not.

ir/translation/function.rs
```diff
@@ -117,7 +117,8 @@ pub fn translate_function_from(
                 false
             }
         }) {
-            let argument_variable = context.variable_registry.get_variable_name(arg).unwrap();
+            let argument_variable = context.variable_registry.get_variable_name(arg)
+                .expect("Argument names were validated earlier");
```
farost, line 121: This one can be updated, too

PR #7590, opened 2025-09-26
Reenable Sentry crash reporting

farost:
Fix a bug when TypeDB Server did not send crash reports for diagnostics even when `--diagnostics.reporting.errors` was enabled.

## Implementation
Fixes the bug introduced in https://github.com/typedb/typedb/pull/7391.

Preserve the ownership of the Sentry's `Guard` until the `main` function is exited. To avoid the same mistake, we encapsulate all the guards and other things required to remain alive while the App is running inside a new struct called `ServerApplication`.

### DIFF
--- a/executor/read/stream_modifier.rs
+++ b/executor/read/stream_modifier.rs
@@ -274,9 +274,7 @@ impl LastMapper {
 impl StreamModifierResultMapperTrait for LastMapper {
     fn map_output(&mut self, subquery_result: Option<FixedBatch>) -> Option<FixedBatch> {
         if let Some(input_batch) = subquery_result {
-            self.last_row = (!input_batch.is_empty()).then(|| {
-                input_batch.get_row(input_batch.len() - 1).into_owned()
-            });
+            self.last_row = (!input_batch.is_empty()).then(|| input_batch.get_row(input_batch.len() - 1).into_owned());
             Some(FixedBatch::EMPTY) // Retry this instruction without returning any rows
         } else {
             self.last_row.take().map(FixedBatch::from)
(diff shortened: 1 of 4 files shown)


executor/read/stream_modifier.rs
```diff
@@ -274,9 +274,7 @@ impl LastMapper {
 impl StreamModifierResultMapperTrait for LastMapper {
     fn map_output(&mut self, subquery_result: Option<FixedBatch>) -> Option<FixedBatch> {
         if let Some(input_batch) = subquery_result {
-            self.last_row = (!input_batch.is_empty()).then(|| {
-                input_batch.get_row(input_batch.len() - 1).into_owned()
-            });
+            self.last_row = (!input_batch.is_empty()).then(|| input_batch.get_row(input_batch.len() - 1).into_owned());
```
farost, line 277: Rustfmt

lolski (approved):

main.rs
```diff
@@ -32,14 +33,32 @@ fn main() {
     config_builder.override_with_cliargs(cli_args);
     let config = config_builder.build().expect("Error validating config file overridden with cli args");
     initialise_logging_global(&config.logging.directory);
-    may_initialise_error_reporting(&config);
```
farost, line 35: Maybe it's better to init everything in main (including the guard) and then just pass it inside the `ServerApplication`... Otherwise, it's a little weird that we init logging and stuff outside of the app, and then the `guard` inside of it.

lolski (approved):

main.rs
lolski: much better! now the guard's lifecycle is bound to the Server's lifecycle

main.rs
```diff
@@ -32,14 +33,32 @@ fn main() {
...
-        }
-    });
+
+    ServerApplication::new(config).run()
+}
+
+struct ServerApplication {
+    runtime: Runtime,
+    config: Config,
+    // This guard sends Sentry reports when it's dropped. It has to be alive for the lifetime of the app.
```
flyingsilverfin, line 43: looks good, thanks!

PR #7591, opened 2025-09-29
Add analyze to GRPC

krishnangovindraj:
Add analyze to GRPC

## Implementation
Adds the analyze RPC endpoint, and associated data structures.

Notably adds queueing for analyze queries in the HTTP API `query_queue`

### DIFF
--- a/dependencies/typedb/repositories.bzl
+++ b/dependencies/typedb/repositories.bzl
@@ -36,5 +36,5 @@ def typedb_behaviour():
     git_repository(
         name = "typedb_behaviour",
         remote = "https://github.com/krishnangovindraj/typedb-behaviour",
-        commit = "3e444bf9b3d7668006fe877dc320439c976cf888",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        commit = "e5372c5f7fca44c4c6b677581bcff63fca321887",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
     )
--- a/server/service/http/message/query/query_structure.rs
+++ b/server/service/http/message/query/query_structure.rs
@@ -4,7 +4,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::{collections::HashMap, str::FromStr};
+use std::collections::HashMap;
 
 use answer::variable::Variable;
 use bytes::util::HexBytesFormatter;
@@ -21,7 +21,8 @@ use ir::pattern::{
     constraint::{Constraint, IsaKind, SubKind},
     ParameterID, Vertex,
 };
-use serde::{Deserialize, Serialize, Serializer};
+use serde::{Deserialize, Serialize};
+use serde_with::{serde_as, FromInto};
 use storage::snapshot::ReadableSnapshot;
 
 use crate::service::http::message::query::concept::{
@@ -123,6 +124,7 @@ pub struct StructureVariableInfo {
     name: Option<String>,
 }
 
+#[serde_as]
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase", tag = "tag")]
 pub(super) enum StructureConstraint {
@@ -175,7 +177,8 @@ pub(super) enum StructureConstraint {
     },
     Expression {
         text: String,
-        assigned: Vec<StructureVertex>, // TODO: Change to just StructureVertex
+        #[serde_as(as="FromInto<Singleton<StructureVertex>>")]
+        assigned: StructureVertex, // TODO: Serialize as is. Breaking change.
         arguments: Vec<StructureVertex>,
     },
     Is {
@@ -435,10 +438,7 @@ fn encode_structure_constraint(
         Constraint::ExpressionBinding(expr) => {
             let text =
                 context.get_call_syntax(constraint).map_or_else(|| format!("Expression#{index}"), |text| text.clone());
-            let assigned = expr
-                .ids_assigned()
-                .map(|variable| encode_structure_vertex(context, &Vertex::Variable(variable)))
-                .collect::<Result<Vec<_>, _>>()?;
+            let assigned = encode_structure_vertex(context, expr.left())?;
             let arguments = expr
                 .expression_ids()
                 .map(|variable| encode_structure_vertex(context, &Vertex::Variable(variable)))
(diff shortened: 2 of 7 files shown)


common/error/error.rs
```diff
@@ -372,6 +372,15 @@ macro_rules! unimplemented_feature {
...
 
+#[cfg(debug_assertions)]
+#[macro_export]
+macro_rules! todo_must_implement {
+    ($msg:literal) => {
+        todo!(concat!("TODO: Must implement: ", $msg)) // Ensure this is enabled when checking in.
+    };
+}
+
+#[cfg(not(debug_assertions))]
```
krishnangovindraj, line 386: Easier to work with, avoids the possibility of someone using `todo_must_implement` but disabling the compile_error so they can continue with the rest of development. The worst that will happen is the snapshot jobs fail after merge.

common/options/options.rs
```diff
@@ -27,6 +27,13 @@ impl Default for TransactionOptions {
     }
 }
 
+#[derive(Debug, Clone, Eq, PartialEq)]
+
+pub enum QueryQueueOptions {
+    Query(QueryOptions),
+    Analyze,
+}
+
```
krishnangovindraj: Wrong place, should be in the file where it's used.

dependencies/typedb/repositories.bzl
```diff
@@ -28,13 +28,13 @@ def typeql():
...
+        commit = "9fba4c5c8105594053ab6b5fea01117e144b7de6",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_protocol
     )
 
 def typedb_behaviour():
     git_repository(
         name = "typedb_behaviour",
-        remote = "https://github.com/typedb/typedb-behaviour",
-        commit = "f74ee6066ccddbbf94618b1c10273c8e5efcfe09",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        remote = "https://github.com/krishnangovindraj/typedb-behaviour",
+        commit = "3614fd94988805660fa014ee0394d2e8cbaf8b63",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
```
krishnangovindraj: upstreams

ir/pattern/constraint.rs
```diff
@@ -1165,7 +1165,7 @@ impl<ID: IrID> fmt::Display for RoleName<ID> {
 }
 #[derive(Clone, Eq, PartialEq, Hash)]
 pub struct Kind<ID> {
-    kind: typeql::token::Kind,
+    kind: encoding::graph::type_::Kind,
```
krishnangovindraj, line 1168: I don't see why we don't just resolve it immediately.

server/service/grpc/transaction_service.rs
```diff
@@ -871,6 +955,39 @@ impl TransactionService {
...
+            let resp = match result {
+                Ok(analyzed) => {
+                    match encode_analyzed_query(&*transaction.snapshot, &*transaction.type_manager, &analyzed) {
+                        Ok(encoded) => ImmediateAnalyzeResponse::analyzed_query(encoded),
+                        Err(err) => ImmediateAnalyzeResponse::non_fatal_err(err),
+                    }
+                }
+                Err(err) => ImmediateAnalyzeResponse::non_fatal_err(err),
+            };
+            let _ = Self::respond_analyze_response(&self.response_sender, req_id, resp).await;
```
krishnangovindraj, line 983: When is it right to ignore a returned ControlFlow?

ir/pattern/constraint.rs
flyingsilverfin: yeah makes sense

server/service/grpc/transaction_service.rs
```diff
@@ -451,6 +476,20 @@ impl TransactionService {
         }
     }
 
+    async fn respond_analyze_response(
```
flyingsilverfin, line 479: i mean if you're going to use analyse with british spelling here (eg handle_analyse_query)  let's do the renaming in a separate PR.

server/service/grpc/transaction_service.rs
flyingsilverfin: `run_write_query` also has this pattern.

In effect the control flow signal is meant to terminate the listen loop on transaction-fatal error.

server/service/grpc/transaction_service.rs
```diff
@@ -1585,3 +1698,9 @@ impl QueryStreamTransmitter {
         Continue(())
     }
 }
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum QueryQueueOptions {
```
flyingsilverfin: I think this is just `QueueOptions`, and then you have Query/Analyze within that. Analyze isn't a query itself

krishnangovindraj: It's inside the `query_queue` though. Maybe that has to be renamed to `request_queue`?

query/analyse.rs
```diff
@@ -48,7 +49,7 @@ pub struct AnalysedQuery {
 pub struct QueryStructureAnnotations {
     pub preamble: Vec<FunctionStructureAnnotations>,
     pub query: PipelineStructureAnnotations,
-    pub fetch: Option<FetchStructureAnnotations>,
+    pub fetch: Option<FetchStructureAnnotationsFields>,
```
krishnangovindraj, line 52: TODO: Go all the way with the renaming so we're at FetchAnnotationsFields

server/service/http/message/query/query_structure.rs
```diff
@@ -581,6 +577,7 @@ fn encode_role_type_as_vertex(
     role_type: &Vertex<Variable>,
 ) -> Result<StructureVertex, Box<ConceptReadError>> {
     if let Some(label) = context.get_role_type(&role_type.as_variable().unwrap()) {
+        // TODO: Make consistent with GRPC API by introducing StructureVertex::NamedRole
```
krishnangovindraj, line 580: This and other pending breaking changes.

krishnangovindraj (commented):
> you can always run analyze for queries - in read/write/schema transactions right?

Yes, I don't think I've added any checks / disabled any of them

> you can analyze a write query, even in a read transaction?

Sure, why not.

query/analyse.rs
krishnangovindraj: Different PR maybe when I update the HTTP api as well

server/service/grpc/transaction_service.rs
flyingsilverfin: good for me!

krishnangovindraj: ~~There's something funky about how I handle it in the HTTP driver. I assume the queue is empty~~ 
Fixed

PR #7593, opened 2025-09-30
Replace serializable_response macro with available serde annotations

krishnangovindraj:
Replace custom implementation of Serialize for concepts in the HTTP API with the derived serialize and serde annotations.

### DIFF
--- a/server/service/http/message/query/concept.rs
+++ b/server/service/http/message/query/concept.rs
@@ -32,93 +32,58 @@ macro_rules! count_fields {
     ($head:ident $(, $tail:ident)*) => { 1 + count_fields!($($tail),*) };
 }
 
-macro_rules! serializable_response {
-    (
-        $vis:vis struct $name:ident {
-            kind = $kind:expr,
-            $( $field_vis:vis $field:ident : $ty:ty => $json_key:literal ),* $(,)?
-        }
-    ) => {
-        #[derive(Debug, Clone, serde::Deserialize)]
-        #[serde(rename_all = "camelCase")]
-        $vis struct $name {
-            $( $field_vis $field : $ty ),*
-        }
-
-        impl serde::Serialize for $name {
-            fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-            where
-                S: serde::Serializer,
-            {
-                let mut state = serializer.serialize_struct(stringify!($name), 1 + count_fields!($($field),*))?;
-                state.serialize_field("kind", $kind)?;
-                $( state.serialize_field($json_key, &self.$field)?; )*
-                state.end()
-            }
-        }
-    };
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", tag = "kind", rename = "entity")]
+pub struct EntityResponse {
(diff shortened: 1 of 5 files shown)


server/service/http/message/query/concept.rs
```diff
@@ -32,93 +32,58 @@ macro_rules! count_fields {
...
-            where
-                S: serde::Serializer,
-            {
-                let mut state = serializer.serialize_struct(stringify!($name), 1 + count_fields!($($field),*))?;
-                state.serialize_field("kind", $kind)?;
-                $( state.serialize_field($json_key, &self.$field)?; )*
-                state.end()
-            }
-        }
-    };
```
krishnangovindraj, line 59: I think this existed for the "kind" (which is a constant) to be serialized? I realised the `tag` can be used for structs, jus as we do for enums.

server/service/http/message/query/concept.rs
```diff
@@ -32,93 +32,58 @@ macro_rules! count_fields {
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", tag = "kind", rename = "entity")]
+pub struct EntityResponse {
+    pub iid: String,
+    pub r#type: Option<EntityTypeResponse>,
```
farost, line 39: What does `r#` do?

server/service/http/message/query/concept.rs
```diff
@@ -32,93 +32,58 @@ macro_rules! count_fields {
...
 }
 
-serializable_response! {
-    pub struct EntityResponse {
-        kind = "entity",
```
farost, line 64: We used to encode the kind of the concept this way.
1. Does it mean that the structure of the response is changed? Are our clients ready? Will our users be ready? Will we update the API version to do so?
2. Do we not need to encode the kind anymore? Are we sure that this "kind" is encoded differently through serde for all the responses, and, thus, the concept can be decoded easily?

server/service/http/message/query/concept.rs
farost: Okay, I see the comment now. 
So we encode the concepts the same way as we did before, and, besides `iid` and `type` keys, we also include a key named `kind`, which takes the name of the Rust struct, which is renamed to the `rename` value. Thus, an Entity document will look like:
```
{
"iid": ....,
"type": ...,
"kind": "entity"
}
```
Correct?

server/service/http/message/query/concept.rs
krishnangovindraj: Escapes keywords, basically.

server/service/http/message/query/concept.rs
krishnangovindraj: Yep.

server/service/http/message/query/concept.rs
krishnangovindraj: 1. No changes.
2. Since you wrote the HTTP API, it's well covered with tests, and [this one](https://github.com/typedb/typedb-behaviour/blob/85079e8179d8ec9bd189281159ec0661dc6422c5/driver/http/http.feature#L250) should test all my changes here.

PR #7594, opened 2025-09-30
Fix relation indices recalculation on schema changes

farost:
Correctly regenerate and delete relation role player indices on schema changes. This fixes a bug when some query results could be ignored because of an incorrectly configured index used for optimization purposes.

## Implementation
**Write transactions** always generate indices after each operation, so the query plan can assume that they exist where they are expected (edge case: temporarily large relations beyond cardinality limits == creating huge indices before throwing them out - acceptable). We preserve the old behavior as is.

**Schema transactions** are complicated. Previously, we used to generate indices after each data operation, but we did nothing on schema changes. However, schema changes require index regeneration, too, as there might be incorrect, excessive, or missing indices. That's what we've been lacking. We cannot do it at operation time (the same reason as for cardinality checks: too expensive), so we need to perform this regeneration on commits.

Now, they work this way:
* no index usage in schema transactions for read operations
* no index update on each operation related to indices, or when cardinality changes
* index updates at commit time based on modified cardinalities, for each relation that has been modified

We simplify the approach by not optimizing the read path. We could potentially check that there are no schema card changes, so the indices can be used, but we don't have an easy way to do so now.

The implementation is a little different from the theory. Instead of sharing the transaction type with `thing_manager`, we use a more general "static schema guarantee" flag in the snapshot. For more information, see PR comments.

P.S. Uncovered one bug, I'm mostly convinced it's unrelated, and I will spend too much time exploring it now:  https://github.com/typedb/typedb/issues/7597


compiler/tests/transformation.rs
```diff
@@ -92,15 +92,40 @@ fn setup_database(storage: &mut Arc<MVCCStorage<WALClient>>) {
     snapshot.commit(&mut CommitProfile::DISABLED).unwrap();
 }
 
-#[test]
-fn test_relation_index_transformation_single() {
+fn translate_and_annotate(
```
farost, line 95: Moved from the bottom

compiler/tests/transformation.rs
```diff
@@ -42,7 +42,7 @@ const START_TIME_LABEL: Label = Label::new_static("start-time");
 fn setup_database(storage: &mut Arc<MVCCStorage<WALClient>>) {
     setup_concept_storage(storage);
 
-    let (type_manager, thing_manager) = load_managers(storage.clone(), None);
+    let (type_manager, thing_manager) = load_managers(storage.clone(), None, false);
```
farost: I added a flag everywhere. By default, it's set to "false" (so no static schema guarantees) to catch potentially dangerous tests. Then, I adjusted and enriched the tests that failed.

compiler/tests/transformation.rs
```diff
@@ -92,15 +92,40 @@ fn setup_database(storage: &mut Arc<MVCCStorage<WALClient>>) {
...
+fn run_test_relation_index_transformation_single<const STATIC_SCHEMA_GUARANTEE: bool>() {
     let (_tmp_dir, mut storage) = create_core_storage();
     setup_database(&mut storage);
-    let (type_manager, _thing_manager) = load_managers(storage.clone(), None);
+    let (type_manager, thing_manager) = load_managers(storage.clone(), None, STATIC_SCHEMA_GUARANTEE);
     let snapshot = storage.clone().open_snapshot_read();
 
     let dog_ownership = type_manager.get_relation_type(&snapshot, &DOG_OWNERSHIP_LABEL).unwrap().unwrap();
-    assert!(type_manager.relation_index_available(&snapshot, dog_ownership).unwrap());
+    assert!(dog_ownership.qualifies_for_relation_index(&snapshot, &type_manager).unwrap());
```
farost: `relation_index_available` is now a `thing_manager` method, which also checks if there are static schema guarantees (the index is produced by the thing manager, so it should be worried about the flag). Then, to avoid mixing type_manager's cardinality check with it, I encapsulated it inside the type and called "qualification for relation index".

compiler/tests/transformation.rs
```diff
@@ -92,15 +92,40 @@ fn setup_database(storage: &mut Arc<MVCCStorage<WALClient>>) {
...
+        &BTreeMap::new(),
+        &EmptyAnnotatedFunctionSignatures,
+        false,
+    )
+    .unwrap();
+    let conjunction = block.into_conjunction();
+    (conjunction, type_annotations)
+}
+
+fn run_test_relation_index_transformation_single<const STATIC_SCHEMA_GUARANTEE: bool>() {
```
farost: Tests in this file are written in two ways based on the static schema guarantees

concept/tests/test_statistics.rs
```diff
@@ -167,14 +167,16 @@ fn read_statistics(storage: Arc<MVCCStorage<WALClient>>, thing_manager: &ThingMa
                 .or_default() += count;
             *this_relation_players.entry(player.type_()).or_default() += 1;
         }
-        for (player_1, count_1) in &this_relation_players {
-            for (player_2, count_2) in &this_relation_players {
-                let link_count = if player_1 == player_2 { count_1 * (count_2 - 1) } else { count_1 * count_2 };
-                if link_count == 0 {
-                    continue;
+        if thing_manager.relation_index_available(&snapshot, relation.type_()).unwrap() {
```
farost: No index available -> no counts, makes sense

concept/tests/test_thing.rs
```diff
@@ -648,9 +648,10 @@ fn role_player_distinct() {
     let person_label = Label::build("person", None);
     let company_label = Label::build("company", None);
 
+    // Schema transaction
```
farost, line 651: Instead of comparing write + read, we compare schema + write + read now. In schema, there are no indices. In write and read, there are indices.

concept/thing/thing_manager.rs
```diff
@@ -107,15 +107,18 @@ pub struct ThingManager {
     vertex_generator: Arc<ThingVertexGenerator>,
     type_manager: Arc<TypeManager>,
     statistics: Arc<Statistics>,
+    // TODO: Ideally, ask the incoming snapshot if there are schema changes instead.
```
farost: Currently, it feels like the best option. I don't want to introduce `TransactionType` in `concept` or `database`. However, I don't like the boolean flag. However x2, I don't want to check the snapshot and check if the schema-based keys are present. 
You may have better ideas!

concept/thing/thing_manager.rs
```diff
@@ -1511,7 +1514,7 @@ impl ThingManager {
         end_role: RoleType,
         storage_counters: StorageCounters,
     ) -> Result<bool, Box<ConceptReadError>> {
-        if !self.type_manager.relation_index_available(snapshot, relation.type_())? {
+        if !self.relation_index_available(snapshot, relation.type_())? {
```
farost: All checks now depend on our local method: the schema should be static + the relation should be eligible (should "qualify")

concept/thing/thing_manager.rs
```diff
@@ -1787,7 +1790,22 @@ impl ThingManager {
         snapshot: &mut impl WritableSnapshot,
         storage_counters: StorageCounters,
     ) -> Result<(), Vec<ConceptWriteError>> {
-        self.validate(snapshot, storage_counters.clone())?;
```
farost, line 1790: Previously, `validate` collected and validated cardinalities.
I split it into two separate stages:
* collect cardinality changes
* validate cardinalities based on these changes

Additionally, I introduced:
* based on these changes, also update relation indices if needed

Splitting validation / updates by separate reasons seems to be a good approach. It won't scale in the current implementation if we'll need to collect other data from the same snapshot regions (e.g., collect entity types with specific flags by iterating through new entity type writes -> two iterations over entity type writes), but we can solve it later, as the collection code seems to be quite local.

Plus, I am happy that I made `thing_manager.rs` a little lighter.

concept/thing/thing_manager.rs
```diff
@@ -2142,115 +2160,44 @@ impl ThingManager {
...
     }
 
-    fn validate(
+    fn validate_cardinalities(
         &self,
         snapshot: &mut impl WritableSnapshot,
+        change_tracker: &CardinalityChangeTracker,
         storage_counters: StorageCounters,
     ) -> Result<(), Vec<ConceptWriteError>> {
-        /*
```
farost, line 2150: All the deleted code just got moved to another file

concept/thing/thing_manager.rs
```diff
@@ -2267,426 +2214,61 @@ impl ThingManager {
...
+        affected_role_types: &HashSet<RoleType>,
+        qualifies_for_relation_index: bool,
+        storage_counters: StorageCounters,
+    ) -> Result<(), Box<ConceptWriteError>> {
+        for role_type in affected_role_types {
+            let mut it = relation.get_players_by_role(snapshot, &self, *role_type, storage_counters.clone());
+            while let Some(player) = it.next() {
+                let (player, count) =
+                    player.map_err(|typedb_source| Box::new(ConceptWriteError::ConceptRead { typedb_source }))?;
+                if qualifies_for_relation_index {
```
farost, line 2248: If qualifies, regenerate the index. If not, remove (clear existing `put`s and record `delete`s if needed). I don't see anything smarter without super complications.

concept/thing/thing_manager.rs
```diff
@@ -3290,69 +2871,6 @@ impl ThingManager {
         self.set_links_count(snapshot, relation, player, role_type, count.unwrap() - decrement_count, storage_counters)
     }
 
-    // TODO:
-    // * Call index regenerations when cardinality changes in schema
-    //   (create role type, set cardinality annotation, unset cardinality annotation, ...)
-    // * Clean up all parts of a relation index to do with a specific role player
-    //   after the player has been deleted.
-    pub(crate) fn relation_index_player_links_unset(
```
farost, line 3298: Moved below "regenerate". The old ordering of "first unset method, then set method" was weird. Also renamed the method, because it's quite good for all operations, not only link unsetting.

concept/thing/thing_manager/validation/commit_time_validation.rs
farost, line 1: It's only about cardinality validations, so I decided that I'd better name it explicitly `cardinality`.

dependencies/typedb/repositories.bzl
```diff
@@ -38,3 +38,7 @@ def typedb_behaviour():
         remote = "https://github.com/typedb/typedb-behaviour",
         commit = "f74ee6066ccddbbf94618b1c10273c8e5efcfe09",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
     )
+#    native.local_repository(
```
farost: Will be removed

encoding/graph/thing/edge.rs
```diff
@@ -805,6 +805,17 @@ impl ThingEdgeIndexedRelation {
         Self { player_from, player_to, relation_type_id, relation_id, role_id_from, role_id_to }
     }
 
+    pub fn into_reverse(self) -> Self {
```
farost, line 808: Just feels nicer

storage/snapshot/buffer.rs
```diff
@@ -177,10 +177,6 @@ impl WriteBuffer {
         self.writes.contains_key(key)
     }
 
-    pub(crate) fn get(&self, key: &[u8]) -> Option<&Write> {
```
farost, line 180: Duplicate

storage/snapshot/buffer.rs
```diff
@@ -256,7 +252,7 @@ impl WriteBuffer {
         &mut self.writes
     }
 
-    pub fn get_write(&self, key: &[u8]) -> Option<&Write> {
+    pub fn writes_get(&self, key: &[u8]) -> Option<&Write> {
```
farost, line 255: Better name (similar to what's around this method)

compiler/transformation/relation_index.rs
```diff
@@ -79,29 +79,29 @@ pub fn relation_index_transformation(
...
-    type_manager: &TypeManager,
+    thing_manager: &ThingManager,
     snapshot: &impl ReadableSnapshot,
     relation: &Vertex<Variable>,
     type_annotations: &TypeAnnotations,
 ) -> Result<bool, StaticOptimiserError> {
     let relation_types = type_annotations.vertex_annotations_of(relation).unwrap();
     for type_ in relation_types.iter() {
-        let index_available = type_manager
+        let index_available = thing_manager
```
farost: The most important part of the execution fix. This will block indices in schema transactions.

concept/thing/thing_manager.rs
krishnangovindraj: Why not add fn `static_schema_guarantee()` to the `ReadableSnapshot` trait?

concept/thing/thing_manager/validation/cardinality_validation.rs
```diff
@@ -0,0 +1,776 @@
...
+    modified_objects_attribute_types: HashMap<Object, HashSet<AttributeType>>,
+    has_modified_owns: bool,
+    modified_objects_role_types: HashMap<Object, HashSet<RoleType>>,
+    has_modified_plays: bool,
+    modified_relations_role_types: HashMap<Relation, HashSet<RoleType>>,
+    has_modified_relates: bool,
+}
+
+impl CardinalityChangeTracker {
+    pub(crate) fn new(
```
krishnangovindraj: nit, but I think we tend to call it `build` if there's work involved and new only if the work is trivial

concept/thing/thing_manager.rs
```diff
@@ -1787,7 +1790,22 @@ impl ThingManager {
         snapshot: &mut impl WritableSnapshot,
         storage_counters: StorageCounters,
     ) -> Result<(), Vec<ConceptWriteError>> {
-        self.validate(snapshot, storage_counters.clone())?;
+        let cardinality_change_tracker =
+            match CardinalityChangeTracker::new(snapshot, self.type_manager(), &self, storage_counters.clone()) {
+                Ok(tracker) => tracker,
+                Err(typedb_source) => return Err(vec![ConceptWriteError::ConceptRead { typedb_source }]),
+            };
```
krishnangovindraj:
```suggestion
            CardinalityChangeTracker::new(snapshot, self.type_manager(), &self, storage_counters.clone())
                .map_err(|err| vec![ConceptWriteError::ConceptRead { typedb_source }])?;
```

concept/thing/thing_manager.rs
```diff
@@ -2267,426 +2214,61 @@ impl ThingManager {
...
-            let (min, max) = minmax_or!(
-                TypeAPI::chain_types(
-                    object_type,
-                    object_type.get_subtypes_transitive(snapshot, self.type_manager(),)?.into_iter().cloned()
-                ),
-                unreachable!("Expected at least one object type")
-            );
-            let mut it = self.get_objects_in_range(
+    ) -> Result<(), Box<ConceptWriteError>> {
+        for (relation, modified_relates) in change_tracker.modified_relations_role_types() {
```
krishnangovindraj, line 2219: Your tests look like they cover every case, So I'm trying to figure out how this regenerates indices for relation instances which weren't touched in the transaction, but whose types had their cardinality modified.

concept/thing/thing_manager.rs
krishnangovindraj: Maybe this is better off as two different functions - one to add and one to remove?

krishnangovindraj (commented):
questions

concept/thing/thing_manager.rs
farost: `modified_relations_role_types` might be a confusing name. It's a map of `modified relations` to `role types` of links that were modified. Either instances or schema. I don't know how to express it "logically". In reality, it collects all schema changes (subs, relates, card updates) and data movements.

farost:
@krishnangovindraj  cannot reply to

> Maybe this is better off as two different functions - one to add and one to remove?

I don't see a reason to iterate twice

concept/thing/thing_manager/validation/cardinality_validation.rs
farost: Sounds good, I'll do it.

farost:
@krishnangovindraj Same, could not reply directly

> Why not add fn static_schema_guarantee() to the ReadableSnapshot trait?

I was not sure if it's a good place for it, but after working with tests, I clearly see that it's the right place to put it. I've created an associated `const bool`.

compiler/tests/transformation.rs
farost: Not anymore

compiler/tests/transformation.rs
farost: Not anymore

compiler/tests/transformation.rs
farost: Not anymore

concept/type_/relation_type.rs
```diff
@@ -626,6 +626,22 @@ impl RelationType {
         }
         Ok(())
     }
+
+    pub fn relation_index_available<Snapshot: ReadableSnapshot>(
```
farost, line 630: Not sure if the two methods are clear enough. Initially, I moved `relation_index_available` into `thing_manager`, but it does not really make sense. It's really only about `type_manager` and `snapshot`.

concept/thing/thing_manager.rs
flyingsilverfin: Yeah i think overall this is sound! Good idea!

flyingsilverfin: Is this "equally light" as it used to be in the previous iteration for just Write transactions? I want to ensure that the typical operation (write tx) is as efficient as can be

concept/thing/thing_manager/validation/cardinality_validation.rs
flyingsilverfin, line 1: Imma take your word for it that this is the right/best way to do this! We still have the old problem of things not fitting in memory, but i guess that's for later

concept/thing/thing_manager.rs
flyingsilverfin: so then should it be called `(modified_relations, relates)`? 

You also have this naming elsewhere in ThingManager:2190

concept/thing/thing_manager.rs
farost: Collection is the same. Validation is the same. 
However, I should not enter the recalculation for static schemas. I thought I added it, my bad.
Once I add it tomorrow, it will be equally light as it used to be.

concept/type_/relation_type.rs
flyingsilverfin: works great here

concept/type_/relation_type.rs
```diff
@@ -626,6 +626,22 @@ impl RelationType {
...
+        &self,
+        snapshot: &Snapshot,
+        type_manager: &TypeManager,
+    ) -> Result<bool, Box<ConceptReadError>> {
+        Ok(Snapshot::STATIC_SCHEMA_GUARANTEE && self.schema_qualifies_for_relation_index(snapshot, type_manager)?)
```
flyingsilverfin: Err let's call it `IMMUTABLE_SCHEMA`

flyingsilverfin: if you're introducing a "new" term ('static') always have a think if we don't already have one!

concept/thing/thing_manager.rs
krishnangovindraj: How would it be schema if its signature is `modified_relations_role_types: HashMap<Relation, HashSet<RoleType>>,` ?
So specifically, how do we track that a certain relation instance has to have its indices regenerated if the cardinality on its relation type has changed?

concept/thing/thing_manager/validation/cardinality_validation.rs
krishnangovindraj: Is it really "for later"? To me it looks like we will now collect ALL instances of modified types in memory, and this could make all schema migrations infeasible.

concept/thing/thing_manager.rs
krishnangovindraj: Is it `collect_modified_schema_capability_cardinalities` ?
Can you point me to the commit/commit range where I can (or what i do to) see the modifications made to cardinality

farost: @krishnangovindraj I don't get the question, and what you mean by "the modifications made to cardinality". Can you elaborate?

The HashMap is collected based on both thing and type changes, it's inside the change tracker, the code is quite small and split by thing / type. Yes, `collect_modified_schema_capability_cardinalities` is one of the places.

If a relation or its links is changed, we add it and the links' role type to the map. If the whole relates' cardinality is changed, we add it, too. Etc, etc.

concept/thing/thing_manager/validation/cardinality_validation.rs
farost: We've been doing it since the first release. I mentioned here that the collection / validation stages are not changed **at all**, I just moved them to a different file. Also, we have this: https://github.com/typedb/typedb/issues/7138 (I updated this issue a little). I agree that we need to prioritize this.

PR #7595, opened 2025-09-30
Allow extending the data commit functionality of user update and delete functionality

lolski:
Make user update and delete functionality use the supplied `ServerState` object to perform data commit. This architecture change would allow TypeDB Cluster to supply its own `ServerState` implementation which will contain cluster-wide data commit functionality

## Implementation
- Make the `update` and `delete` methods in `UserManager` to return commit record
- Remove user update and delete methods in server state as the data commit will make use of the `database_data_commit` functionality
- Update user update and delete functionality in GRPC and HTTP services to include data commit step

### DIFF
--- a/server/service/grpc/typedb_service.rs
+++ b/server/service/grpc/typedb_service.rs
@@ -4,8 +4,6 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use user::permission_manager::PermissionManager;
-use storage::snapshot::CommittableSnapshot;
 use std::{net::SocketAddr, pin::Pin, time::Instant};
 use std::sync::Arc;
 use diagnostics::metrics::ActionKind;
@@ -21,9 +19,6 @@ use typedb_protocol::{
     transaction::{Client as TransactionClientProto, Server as TransactionServerProto},
 };
 use uuid::Uuid;
-use database::transaction::DataCommitError;
-use resource::profile::CommitProfile;
-use user::errors::{UserCreateError, UserDeleteError, UserUpdateError};
 use crate::{
     authentication::{Accessor, AuthenticationError},
     error::LocalServerStateError,
@@ -52,18 +47,18 @@ use crate::{
             ConnectionID,
         },
         transaction_service::TRANSACTION_REQUEST_BUFFER_SIZE,
+        typedb_service::TypeDBService,
     },
     state::ArcServerState,
 };
-use crate::system_init::SYSTEM_DB;
 
 #[derive(Debug)]
-pub(crate) struct TypeDBService {
+pub(crate) struct GRPCTypeDBService {
     address: SocketAddr,
     server_state: ArcServerState,
 }
 
-impl TypeDBService {
+impl GRPCTypeDBService {
     pub(crate) fn new(address: SocketAddr, server_state: ArcServerState) -> Self {
         Self { address, server_state }
     }
@@ -76,7 +71,7 @@ impl TypeDBService {
 }
 
 #[tonic::async_trait]
-impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
+impl typedb_protocol::type_db_server::TypeDb for GRPCTypeDBService {
     // Update AUTHENTICATION_FREE_METHODS if this method is renamed
     async fn connection_open(
         &self,
@@ -535,42 +530,10 @@ impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
                 let (user, credential) =
                     users_create_req(request).map_err(|err| err.into_error_message().into_status())?;
 
-                if !PermissionManager::exec_user_create_permitted(accessor.0.as_str()) {
-                    return Err(LocalServerStateError::OperationNotPermitted {}.into_error_message().into_status());
-                }
-
-                let (mut transaction_profile, create_result) = match self.server_state.user_manager().await {
-                    Some(user_manager) => {
-                        match user_manager.create(&user, &credential) {
-                            (mut transaction_profile, Ok(commit_intent)) => {
-                                let commit_profile = transaction_profile.commit_profile();
-                                let into_commit_record_result = commit_intent.write_snapshot
-                                    .finalise(commit_profile)
-                                    .map_err(|error|
-                                        LocalServerStateError::UserCannotBeCreated { typedb_source: UserCreateError::Unexpected { } }
-                                    );
-                                (transaction_profile, into_commit_record_result
-                                    .map(|commit_record| (commit_intent.database_drop_guard, commit_record)))
-                            }
-                            (transaction_profile, Err(error)) =>
-                                return Err(LocalServerStateError::UserCannotBeCreated { typedb_source: error }.into_error_message().into_status())
-                        }
-                    },
-                    None => return Err(LocalServerStateError::NotInitialised { }.into_error_message().into_status())
-                };
-
-                let create_result = match create_result {
-                    Ok((_, commit_record_opt)) => {
-                        let commit_profile = transaction_profile.commit_profile();
-                        if let Some(commit_record) = commit_record_opt {
-                            self.server_state.database_data_commit(SYSTEM_DB, commit_record, commit_profile).await.unwrap();
-                        }
-                        Ok(Response::new(user_create_res()))
-                    }
-                    Err(err) => return Err(err.into_error_message().into_status())
-                };
-
-                create_result
+                TypeDBService::create_user(&self.server_state, accessor, user, credential)
+                    .await
+                    .map(|_| Response::new(user_create_res()))
+                    .map_err(|err| err.into_error_message().into_status())
             },
         )
         .await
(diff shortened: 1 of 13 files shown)


user/user_manager.rs
```diff
@@ -66,10 +66,10 @@ impl UserManager {
         username: &str,
         user: &Option<User>,
         credential: &Option<Credential>,
-    ) -> Result<(), UserUpdateError> {
-        let update_result = self
+    ) -> (TransactionProfile, Result<(DatabaseDropGuard<WALClient>, WriteSnapshot<WALClient>), UserUpdateError>) {
+        let (transaction_profile, update_result) = self
             .transaction_util
-            .write_transaction_commit(|snapshot, type_mgr, thing_mgr, fn_mgr, query_mgr, _db, _tx_opts| {
+            .write_transaction(|snapshot, type_mgr, thing_mgr, fn_mgr, query_mgr, _dbb, _tx_opts| {
```
lolski: Fix typo

server/state.rs
```diff
@@ -217,7 +207,6 @@ impl LocalServerState {
 
     pub async fn load(&mut self) {
         let system_database = self.database_manager().await.database_unrestricted(SYSTEM_DB).unwrap();
-        println!("load system db: {:?}", system_database);
```
lolski, line 220: Remove unneeded debug statement

server/state.rs
```diff
@@ -104,16 +104,6 @@ pub trait ServerState: Debug {
 
     async fn users_contains(&self, name: &str) -> Result<bool, ArcServerStateError>;
     
-    async fn users_update(
```
lolski, line 107: Remove user update and delete

server/service/grpc/typedb_service.rs
```diff
@@ -589,12 +589,43 @@ impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
                     .map_err(|err| err.into_error_message().into_status())?;
                 let (username, user_update, credential_update) =
                     users_update_req(request).map_err(|err| err.into_error_message().into_status())?;
-                let username = username.as_str();
-                self.server_state
-                    .users_update(username, user_update, credential_update, accessor)
-                    .await
-                    .map(|_| Response::new(user_update_res()))
-                    .map_err(|err| err.into_error_message().into_status())
+
```
lolski, line 555: Update user update and delete functionality in GRPC service to include data commit

server/service/http/typedb_service.rs
```diff
@@ -513,11 +514,41 @@ impl TypeDBService {
                 let user_update = None; // updating username is not supported now
                 let credential_update = Some(Credential::new_password(&payload.password));
                 let username = user_path.username.as_str();
-                service
-                    .server_state
-                    .users_update(username, user_update, credential_update, accessor)
-                    .await
-                    .map_err(|typedb_source| HttpServiceError::State { typedb_source })
+
```
lolski, line 483: Update user update and delete functionality in HTTP service to include data commit

server/service/grpc/typedb_service.rs
```diff
@@ -589,12 +589,43 @@ impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
...
+                if !PermissionManager::exec_user_update_permitted(accessor.0.as_str(), &username) {
```
farost: Can we add an `as_str()` method to `accessor` instead of calling `accessor.0.as_str()` everywhere? Not crucial

farost: Btw, where did we check the permissions before? I don't see it in the deleted code, but I remember that we used to do it

UPD. Saw the refactoring now, left a comment in the HTTP service

server/service/grpc/typedb_service.rs
```diff
@@ -589,12 +589,43 @@ impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
...
+                    return Err(LocalServerStateError::OperationNotPermitted {}.into_error_message().into_status());
+                }
+
+                let (mut transaction_profile, update_result) = match self.server_state.user_manager().await {
```
farost: I'd really prefer to write
```
let user_manager = self.server_state.user_manager().await.ok_or(....)?;
let (mut transaction_profile, update_result) = ....
```

I believe that additional layering adds complexities. Plus, with our current code base (which prefers `?` in many cases, especially in these services), it creates an impression that this "None" branch contains something smarter, but it does not. The code will become much more compact

server/service/http/typedb_service.rs
```diff
@@ -535,11 +566,42 @@ impl TypeDBService {
...
+                                let commit_profile = transaction_profile.commit_profile();
+                                let into_commit_record_result = snapshot
+                                    .finalise(commit_profile)
+                                    .map_err(|error|
+                                        LocalServerStateError::UserCannotBeDeleted { typedb_source: UserDeleteError::Unexpected { } }
+                                    );
+                                (transaction_profile, into_commit_record_result
+                                    .map(|commit_record| (database, commit_record)))
+                            }
+                            (None, Err(error)) =>
```
farost: The line below will become shorter and probably fit better if you call this `error` `typedb_source` (because it will become `UserCannotBeDeleted  { typedb_source }` instead of `UserCannotBeDeleted { typedb_source: error }`

server/state.rs
farost, line 220: You can leave a trace log if you want to

user/user_manager.rs
```diff
@@ -66,10 +66,10 @@ impl UserManager {
         username: &str,
         user: &Option<User>,
         credential: &Option<Credential>,
-    ) -> Result<(), UserUpdateError> {
-        let update_result = self
+    ) -> (TransactionProfile, Result<(DatabaseDropGuard<WALClient>, WriteSnapshot<WALClient>), UserUpdateError>) {
```
farost: Tbh, these results are crazy haha. A tuple of a `TransactionProfile` with `Database....Guard`, etc is not what I expect from a user update

I haven't seen your original PR, but I want to have a look at the new architecture now

user/user_manager.rs
```diff
@@ -80,42 +80,45 @@ impl UserManager {
...
+                    UserGetError::Unexpected { .. } => (None, Err(UserDeleteError::Unexpected {})),
                 }
             }
         }
-        let delete_result = self
+
+        let (transaction_profile, delete_result) = self
             .transaction_util
-            .write_transaction_commit(|snapshot, type_mgr, thing_mgr, fn_mgr, query_mgr, _db, _tx_opts| {
+            .write_transaction(|snapshot, type_mgr, thing_mgr, fn_mgr, query_mgr, _dbb, _tx_opts| {
```
farost: What is `_dbb`

farost: Databasebasebase

user/user_manager.rs
```diff
@@ -80,42 +80,45 @@ impl UserManager {
...
                 user_repository::delete(snapshot, &type_mgr, thing_mgr.clone(), &fn_mgr, &query_mgr, username)
-            })
-            .1;
-        match delete_result {
-            Ok(Ok(())) => Ok(()),
-            Ok(Err(_query_error)) => Err(UserDeleteError::IllegalUsername {}),
-            Err(_commit_error) => Err(UserDeleteError::Unexpected {}),
-        }
+            });
+        let delete_result = match delete_result {
```
farost: Sounds like a `.map(....).map_err(...)` to m, it's quite short as well

server/service/grpc/typedb_service.rs
farost: And update everywhere, if possible

server/service/grpc/typedb_service.rs
```diff
@@ -589,12 +589,43 @@ impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
...
+                        }
+                    },
+                    None => return Err(LocalServerStateError::NotInitialised { }.into_error_message().into_status())
+                };
+
+                let update_result = match update_result {
+                    Ok((_, commit_record_opt)) => {
+                        let commit_profile = transaction_profile.commit_profile();
+                        if let Some(commit_record) = commit_record_opt {
+                            self.server_state.database_data_commit(SYSTEM_DB, commit_record, commit_profile).await.unwrap();
```
farost: I feel like we expose the inner structure of the user management too much, but I guess it's required for scalability, and you know what you're doing.

user/user_manager.rs
farost: Okay, I saw it now, I guess it is what it is.

server/service/http/typedb_service.rs
farost, line 483: Let's make sure we call `PermissionManager` in both `grpc` / `http` for all the needed methods since we dissolved the common `ServerState`'s methods.

server/service/http/typedb_service.rs
```diff
@@ -513,11 +514,41 @@ impl TypeDBService {
                 let user_update = None; // updating username is not supported now
                 let credential_update = Some(Credential::new_password(&payload.password));
                 let username = user_path.username.as_str();
-                service
-                    .server_state
-                    .users_update(username, user_update, credential_update, accessor)
-                    .await
-                    .map_err(|typedb_source| HttpServiceError::State { typedb_source })
+
+                let (mut transaction_profile, update_result) = match service.server_state.user_manager().await {
```
farost: We can make a general `server/service/typedb_service.rs` where we put all this general logic without API-specific calls. It was unnecessary since we started having states, but it got another reason to exist now -- just to avoid copypaste.

user/user_manager.rs
```diff
@@ -80,42 +80,45 @@ impl UserManager {
...
+            Err(_query_error) => Err(UserUpdateError::IllegalUsername {}),
+        };
+        (transaction_profile, update_result)
     }
 
-    pub fn delete(&self, username: &str) -> Result<(), UserDeleteError> {
+    pub fn delete(
+        &self,
+        username: &str
+    ) -> (Option<TransactionProfile>, Result<(DatabaseDropGuard<WALClient>, WriteSnapshot<WALClient>), UserDeleteError>) {
```
farost: If `TransactionProfile` exists only when the `Result` is `Ok` and it does not exist if there is an error, I'd prefer to call `result.map(|....| (transaction_profile, ...))` instead, so we unwrap only once. Then, we have less state to worry about, less room for bugs.

We could also think about explicit `struct`s for operation results if we get bored of these tuples, although they are fine in Rust due to good static analysis.

server/service/grpc/typedb_service.rs
lolski: It's really good that your radar caught it. Unfortunately it's the best we could do given our current architecture based on `ServerState`. You'll see the same issue in the transaction handling section of the code as well.

When we refine that architecture we'll be able to achieve something a lot more elegant.

server/state.rs
lolski: I think this one can go away completely

user/user_manager.rs
lolski: Lol, fixed in 1189a0ce1d349408a8791acde5fe98867ff4584a

user/user_manager.rs
lolski: I've made the signature a lot easier to understand now.

user/user_manager.rs
lolski: > If TransactionProfile exists only when the Result is Ok and it does not exist if there is an error,

`TransactionProfile` may also exist when there's also an error actually.

server/service/http/typedb_service.rs
lolski: Fixed

user/user_manager.rs
farost: Nice

server/service/grpc/typedb_service.rs
lolski: > Can we add an as_str() method to accessor instead of calling accessor.0.as_str() everywhere? Not crucial

Fixed in `94ba07f301f1601d30099e94d70955b47ad1e426`

server/service/grpc/typedb_service.rs
lolski: Fixed in `74e9318981eca09defab244ea896abc68265d526`

server/service/http/typedb_service.rs
lolski: Fixed in `74e9318981eca09defab244ea896abc68265d526`

server/service/http/typedb_service.rs
lolski: Fixed in `74e9318981eca09defab244ea896abc68265d526`

farost (approved):

PR #7596, opened 2025-10-01
Optional writes

dmitrii-ubskii:


executor/write/write_instruction.rs
```diff
@@ -259,7 +259,8 @@ impl AsWriteInstruction for compiler::executable::delete::instructions::ThingIns
         row: &mut Row<'_>,
         storage_counters: StorageCounters,
     ) -> Result<(), Box<WriteError>> {
-        let thing = get_thing(row, &self.thing).clone();
+        let ThingPosition(position) = self.thing;
+        let Some(thing) = row.get(position).get_thing().cloned() else { return Ok(()) };
```
dmitrii-ubskii, line 263: This turns the delete of an unbound variable to a no-op.

diagnostics/metrics.rs
```diff
@@ -427,7 +427,7 @@ impl ActionMetrics {
     }
 
     fn get_attempted(&self, action_kind: &ActionKind) -> u64 {
-        self.get_attempted(action_kind)
+        self.get_action(action_kind).get_attempted()
     }
```
dmitrii-ubskii, line 431: Fix unconditional recursion

concept/thing/thing_manager.rs
```diff
@@ -1624,19 +1624,22 @@ impl ThingManager {
...
+                match snapshot
                     .get_last_existing::<BUFFER_VALUE_INLINE>(key.as_reference(), storage_counters)
-                    .is_ok_and(|option| option.is_some()));
-                ConceptStatus::Persisted
+                    .map_err(|error| Box::new(ConceptReadError::SnapshotGet { source: error }))?
+                {
+                    Some(_) => Ok(ConceptStatus::Persisted),
+                    None => Ok(ConceptStatus::Deleted),
+                }
             })
```
dmitrii-ubskii, line 1641: We now _always check the storage_ rather than assuming that a concept with no associated write in the buffer has been persisted to disk. This can happen if a concept inserted in the current transaction is deleted more than once (matched by multiple rows in a delete pipeline e.g.).

dmitrii-ubskii: This change also makes a bunch of accessor methods on things fallible.

ir/translation/pipeline.rs
```diff
@@ -163,48 +162,44 @@ impl StructuralEquality for TranslatedStage {
 }
 
 pub fn translate_pipeline(
-    snapshot: &impl ReadableSnapshot,
     all_function_signatures: &impl FunctionSignatureIndex,
     query: &typeql::query::Pipeline,
 ) -> Result<TranslatedPipeline, Box<RepresentationError>> {
```
dmitrii-ubskii, line 167: `snapshot` was only ever used as an argument for other translation functions but never actually used here.

dependencies/typedb/repositories.bzl
```diff
@@ -35,6 +35,6 @@ def typedb_protocol():
 def typedb_behaviour():
     git_repository(
         name = "typedb_behaviour",
-        remote = "https://github.com/typedb/typedb-behaviour",
-        commit = "095a58cb823dd8436600526ceb3ff2c0bdf377f9",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        remote = "https://github.com/dmitrii-ubskii/typedb-behaviour",
+        commit = "6967d55606938e885191e8d4ce73191da3240468",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
     )
```
dmitrii-ubskii, line 40: https://github.com/typedb/typedb-behaviour/pull/385

dependencies/typedb/repositories.bzl
```diff
@@ -21,8 +21,8 @@ def typedb_dependencies():
 def typeql():
     git_repository(
         name = "typeql",
-        remote = "https://github.com/typedb/typeql",
-        tag = "3.5.0",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_driver
+        remote = "https://github.com/dmitrii-ubskii/typeql",
+        commit = "b220659baa0a3ec70b5298bd71a7fe66fec9348d",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_driver
     )
```
dmitrii-ubskii, line 26: https://github.com/typedb/typeql/pull/417

compiler/executable/update/executable.rs
```diff
@@ -41,42 +45,122 @@ impl UpdateExecutable {
...
         &variable_positions,
         variable_registry,
-        &mut connection_inserts,
+        input_variables,
     )?;
 
+    let mut optional_updates = Vec::with_capacity(block.conjunction().nested_patterns().len());
+    for nested_pattern in block.conjunction().nested_patterns() {
+        let NestedPattern::Optional(optional) = nested_pattern else {
+            unreachable!("Only optionals are allowed as nested patterns in update")
```
flyingsilverfin: Or rather - 
Non-optional update patterns are illegal and should have been rejected earlier

?

concept/thing/thing_manager.rs
flyingsilverfin: I get why we have this, but it feels a bit of a heavy check. This will be invoked for every time we get the status of a concept that is not explicitly inserted or deleted? Which might be frequently?


Is there theoretically a way to avoid searching storage? Eg. if we make a new kind of `Write`, like `Uninserted`, which basically means it's deleted...? are there other cases we need to protect against?

dmitrii-ubskii: It is a heavy check, we discussed it when I brought this issue up originally. We left it for a separate PR.

And yeah we would need to keep track of keys that were ever created during transaction even if they have then been removed.

PR #7610, opened 2025-10-17
Print ready message only for the same distribution

farost:
Restrict the "Ready!" message printed on the bootup to only CE distributions of TypeDB. This allows other distributions with potentially more complicated initialization to handle the readiness notifications themselves.

## Implementation
Comparing `DISTRIBUTION` directly is the easiest path. Introducing another configurable boolean flag is too much for this job. Requiring changing the distribution name for such controls must be a decent tradeoff for such a minor piece, although it's not very clean.

### DIFF
--- a/server/lib.rs
+++ b/server/lib.rs
@@ -19,7 +19,8 @@ use database::database_manager::DatabaseManager;
 use rand::prelude::SliceRandom;
 use resource::{
     constants::server::{
-        DISTRIBUTION_INFO, GRPC_CONNECTION_KEEPALIVE, SERVER_ID_ALPHABET, SERVER_ID_FILE_NAME, SERVER_ID_LENGTH,
+        DISTRIBUTION, DISTRIBUTION_INFO, GRPC_CONNECTION_KEEPALIVE, SERVER_ID_ALPHABET, SERVER_ID_FILE_NAME,
+        SERVER_ID_LENGTH,
     },
     distribution_info::DistributionInfo,
 };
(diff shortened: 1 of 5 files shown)


server/service/typedb_service.rs
```diff
@@ -34,7 +34,8 @@ impl TypeDBService {
 
         let user_manager = server_state.user_manager().await.ok_or(LocalServerStateError::NotInitialised {})?;
 
-        let (mut transaction_profile, commit_intent) = user_manager.create(&user, &credential)
+        let (mut transaction_profile, commit_intent) = user_manager
+            .create(&user, &credential)
```
farost, line 38: Rusfmt

server/lib.rs
```diff
@@ -303,7 +309,11 @@ impl Server {
             println!("WARNING: TLS NOT ENABLED. This means connections are insecure and transmit username/password credentials unencrypted over the network.");
             println!("**To allow driver connections, drivers must also be configured to *not* use TLS**")
         }
-        println!("\nReady!");
+
+        if distribution_info.distribution == DISTRIBUTION {
+            // Same distribution -> the initialization is finished.
+            println!("\nReady!");
```
farost, line 315: Actually, it's quite hard to handle the readiness status in Cluster. Maybe we should just change this print to "Server ready", and then additionally send "Entered cluster" and "Left cluster" or something.

farost: I'll leave it like this for now

PR #7621, opened 2025-11-03
Only return involved blocks if they're complete

krishnangovindraj:
TODO: GRPC is missing the always involved conjunctions in the pipeline structure.

Returning involved blocks for queries with more than 64 "branches" is unsupported. The GRPC API will not return any blocks rather than returning an incomplete list of blocks.


krishnangovindraj:
Closing this since it's a huge refactor I don't need yet.

PR #7622, opened 2025-11-04
Fix missing involved conjunctions in GRPC ConceptRow response

krishnangovindraj:
Bump protocol for updated version, so drivers not supporting analyze can break.

## Implementation
We weren't including the "always involved blocks" as they weren't captured in provenance.


krishnangovindraj:
Absorb into #7625

PR #7623, opened 2025-11-04
Ensure graceful shutdown for interval tasks

farost:
Add a tracker for `tokio` tasks spawned in the server's components that require graceful shutdown when the server closes. `TokioIntervalRunner`, previously used for such tasks, used to block the `tokio` runtime, which worked for an isolated version of the CE server, but [led to crashes](https://tokio.rs/tokio/topics/shutdown) in more complex applications reusing TypeDB (e.g., TypeDB Cluster).

## Implementation

* Remove `TokioIntervalRunner`;
* Add `TokioTaskTracker` used in the server to track and join these background tasks;
* Add `TokioTaskSpawner` as the clonable part of the `TokioTaskTracker`, which is sent to `TokenManager`, diagnostics `Reporter`, and HTTP `TransactionService`;
* Cover these new components with unit tests;
* Instead of saving the task handles in the components, just spawn the tasks from `TokioTaskSpawner`s and rely on the `TokioTaskTracker` to manage them;
* Await all tracked tasks after TypeDB services are closed.

We currently track only interval tasks because they're the only tasks that require a graceful shutdown. Other tokio tasks can be spawned untracked, as usual.

### DIFF
--- a/server/lib.rs
+++ b/server/lib.rs
@@ -76,9 +76,6 @@ impl ServerBuilder {
     }
 
     pub fn background_tasks_tracker(mut self, background_tasks_tracker: TokioTaskTracker) -> Self {
-        if self.background_tasks_tracker.is_some() {
-            panic!("Tried to override an already existing background tasks tracker which could already have running not awaited tasks.");
-        }
         self.background_tasks_tracker = Some(background_tasks_tracker);
         self
     }


common/concurrency/tokio_task.rs
```diff
@@ -0,0 +1,399 @@
...
+        let (done_sender, done_receiver) = oneshot::channel();
+
+        let mut shutdown_receiver_clone = shutdown_receiver.clone();
+        let tracker_clone = tracker.clone();
+        let shutdown_task = tokio::spawn(async move {
+            let _ = shutdown_receiver_clone.changed().await;
+
+            tracker_clone.close();
+            tracker_clone.wait().await;
+            let _ = done_sender.send(());
```
farost, line 55: When the shutdown signal is received, we already signal all interval jobs to finish. `join` ensures the shutdown is completed gracefully.

common/concurrency/tokio_task.rs
```diff
@@ -0,0 +1,399 @@
...
+    pub fn spawn_interval<F>(
+        &self,
+        mut action: impl 'static + Send + FnMut() -> F,
+        parameters: IntervalTaskParameters,
+    ) -> JoinHandle<F::Output>
+    where
+        F: Future<Output = ()> + Send + 'static,
+    {
+        let mut shutdown_receiver = self.shutdown_receiver.clone();
+        let IntervalTaskParameters { interval, initial_delay, act_on_shutdown } = parameters;
```
farost: Use a struct instead of a line of args with overloads.

server/lib.rs
```diff
@@ -67,10 +75,17 @@ impl ServerBuilder {
         self
     }
 
+    pub fn background_tasks_tracker(mut self, background_tasks_tracker: TokioTaskTracker) -> Self {
```
farost, line 78: Let setting it in the builder similarly to the shutdown_receiver. Currently, we always build our own `LocalServerState` in Cluster, and `LocalServerState` requires task spawners, so we have to prepare our own external Tracker, too.

server/lib.rs
```diff
@@ -218,6 +244,8 @@ impl Server {
         } else {
             grpc_server.await?;
         }
+
+        self.background_tasks_tracker.join().await;
```
farost: Ensure the background tasks are completed. If not, we always have a forced shutdown on another Ctrl+C.

common/concurrency/BUILD
```diff
@@ -15,9 +15,17 @@ rust_library(
     ]),
     deps = [
         "@crates//:tokio",
+        "@crates//:tokio-test",
+        "@crates//:tokio-util",
     ],
 )
 
+rust_test(
+    name = "test_crate_concurrency",
```
flyingsilverfin, line 24: Check other rust_test naming conventions but i think it can just be `test_concurrency`

farost: This is how we name unit tests:

```
command: |
        bazel test //common/concurrency:test_crate_concurrency --test_output=streamed
        bazel test //compiler:test_crate_compiler --test_output=streamed
        bazel test //concept:test_crate_concept --test_output=streamed
        bazel test //database:test_crate_database --test_output=streamed
        bazel test //durability:test_crate_durability  --test_output=streamed
        bazel test //encoding:test_crate_encoding --test_output=streamed
        bazel test //executor:test_crate_executor --test_output=streamed
        bazel test //function:test_crate_function --test_output=streamed
        bazel test //ir:test_crate_ir --test_output=streamed
        bazel test //query:test_crate_query --test_output=streamed
        bazel test //server:test_crate_server --test_output=streamed
        bazel test //storage:test_crate_storage --test_output=streamed
```

flyingsilverfin: ok thanks

common/concurrency/tokio_task.rs
```diff
@@ -0,0 +1,399 @@
...
+///   before letting the runtime drop.
+///
+/// This avoids blocking in `Drop` (which can deadlock a Tokio worker) while still
+/// giving you a single place to wait for all background work to finish.
+/// For more information, see https://tokio.rs/tokio/topics/shutdown.
+#[derive(Debug)]
+pub struct TokioTaskTracker {
+    tracker: TaskTracker,
+    shutdown_receiver: Receiver<()>,
+    done_receiver: oneshot::Receiver<()>,
```
dmitrii-ubskii: I'd prefer if the Receivers were disambiguated here
```suggestion
    shutdown_receiver: watch::Receiver<()>,
    done_receiver: oneshot::Receiver<()>,
```

common/concurrency/tokio_task.rs
dmitrii-ubskii: You can even unpack them in the signature if you want, though it might be too long in this instance.
```rs
IntervalTaskParameters { interval, initial_delay, act_on_shutdown }: IntervalTaskParameters,
```

server/lib.rs
dmitrii-ubskii: There are `?`s in this function above that can early return and prevent this line from running. Is that not an issue?
If it is, I'd suggest extracting the rest of `serve` into a separate function that this wraps so you don't have to worry about this kind of thing.

dmitrii-ubskii (approved):
Looks good, I believe this is _the_ way to implement shutdown according to Tokio. Left a couple style suggestions and a question.

common/concurrency/tokio_task.rs
farost: Right, I keep forgetting about this. It looks better here.

server/lib.rs
farost: Good point. I will abstract the serving logic away and put the shutdown code (an additional shutdown signal + background tasks joining) before returning the result.

server/lib.rs
flyingsilverfin: To be safe, you should probably shut down the old background task tracker if there is one right? However, how can you do that if you don't have the shutdown signal?

server/lib.rs
```diff
@@ -171,30 +195,51 @@ impl Server {
...
+        let serve_result = Self::serve_all(
+            self.distribution_info,
+            self.config.server.encryption,
+            self.server_state,
+            self.shutdown_sender.clone(),
+            self.shutdown_receiver,
+            self.background_tasks_tracker.get_spawner(),
+        )
+        .await;
+        let _ = self.shutdown_sender.send(());
```
flyingsilverfin, line 214: I wonder if there's a way to clean this up?

1) we currently have "spawned" a shutdown handler for ctrl+c (maybe it should be called shutdown_signal_handler?). That handler does `shutdown_sender.send()`. 
2) we also have this `shutdown_sender.send()`. It's probably fine, but this means we can send 2 shutdown signals right?

common/concurrency/tokio_task.rs
```diff
@@ -0,0 +1,398 @@
...
+#[derive(Debug, Clone)]
+pub struct TokioTaskSpawner {
+    tracker: TaskTracker,
+    shutdown_receiver: watch::Receiver<()>,
+}
+
+impl TokioTaskSpawner {
+    /// Spawn an arbitrary background task and track it. The return handle can be used to await this
+    /// single task's completion, but also can be ignored.
+    pub fn spawn<F>(&self, task: F) -> JoinHandle<F::Output>
```
flyingsilverfin, line 87: tasks spawned here from outside still have to manage their own shutdown if they are long-running right?

server/lib.rs
farost: And it's not async. I will put a panic, then.

farost: I generally believe it should not really be a problem since the `serve` consumes itself, but it will be safer.

server/lib.rs
farost: It's intentional. It's okay to send two signals, and we will do it in most cases. But we decided to await the jobs in case of errors, so we send the shutdown signal to start the **graceful** shutdown process even if the server failed. A best effort move.

common/concurrency/tokio_task.rs
farost: Yes, there is no very smart protection atm.

server/lib.rs
flyingsilverfin: ok sounds good

common/concurrency/tokio_task.rs
flyingsilverfin: ok good to know

server/lib.rs
flyingsilverfin: Ah my bad this is in a builder - I wouldn't worry about it here -- match the methods of the other builders you've got above! my bad

PR #7624, opened 2025-11-04
Remove misleading toolchain argument for cloudsmith uploader dependencies

krishnangovindraj:
## Implementation
The uploader dependencies would indeed use that toolchain, but the uploader itself uses the default toolchain. These happen to be the same because it's the first(? or atleast, only) toolchain registered. 
We remove the argument so they both use the default toolchain, and leave a comment where we register the toolchain.


krishnangovindraj:
Linux deployment isn't working. no clue why. 
Will revisit if this ever breaks again.

PR #7625, opened 2025-11-05
Align HTTP analyze response with GRPC

krishnangovindraj:
Align HTTP analyze response with GRPC
## Implementation

Include architectural decisions, new assumptions, unimplemented paths/technical debt and tests written.

### DIFF
--- a/resource/constants.rs
+++ b/resource/constants.rs
@@ -53,7 +53,7 @@ pub mod server {
     pub const DEFAULT_INCLUDE_INSTANCE_TYPES_FETCH: bool = false;
     pub const DEFAULT_ANSWER_COUNT_LIMIT_GRPC: Option<usize> = None;
     pub const DEFAULT_ANSWER_COUNT_LIMIT_HTTP: Option<usize> = Some(10_000);
-    pub const DEFAULT_INCLUDE_STRUCTURE_HTTP: bool = true;
+    pub const DEFAULT_INCLUDE_STRUCTURE_HTTP: bool = true; // True for studio backwards compatibility
     pub const DEFAULT_INCLUDE_STRUCTURE_GRPC: bool = false;
 
     pub const PERF_COUNTERS_ENABLED: bool = true;


server/service/http/message/analyze/structure.rs
```diff
@@ -178,7 +195,7 @@ pub(super) enum StructureConstraint {
     Expression {
         text: String,
         #[serde_as(as = "FromInto<Singleton<StructureVertex>>")]
-        assigned: StructureVertex, // TODO: Serialize as is. Breaking change.
+        assigned: StructureVertex, // TODO: Serialize as is. Breaking change
```
krishnangovindraj: Open question: This is encoded as a list of variables in HTTP, but a single variable in GRPC.
Changing this would mean I have to break studio backwards compatibility. Do we want to do this or let this be the inconsistency?

compiler/query_structure.rs
```diff
@@ -474,7 +474,7 @@ impl From<Variable> for StructureVariableId {
 #[serde(rename_all = "camelCase")]
 pub struct StructureSortVariable {
     pub variable: StructureVariableId,
-    pub ascending: bool,
+    pub ascending: bool, // TODO: Do we want to encode this as tag: Ascending | Descending?
```
krishnangovindraj: Do we want to?

krishnangovindraj: I think we should for consistency.

dependencies/typedb/repositories.bzl
```diff
@@ -28,8 +28,8 @@ def typeql():
 def typedb_protocol():
     git_repository(
         name = "typedb_protocol",
-        remote = "https://github.com/typedb/typedb-protocol",
-        commit = "9d446848f69867b5d921793e6a7c77f8e035b150",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_protocol
+        remote = "https://github.com/krishnangovindraj/typedb-protocol",
```
krishnangovindraj: reminder for self

server/service/http/message/analyze/annotations.rs
```diff
@@ -289,45 +245,48 @@ pub mod bdd {
     use compiler::query_structure::{QueryStructureConjunctionID, QueryStructureStage};
     use itertools::Itertools;
 
-    use super::{
-        FetchStructureAnnotationsResponse, FunctionReturnAnnotationsResponse, FunctionStructureAnnotationsResponse,
-        PipelineStructureAnnotationsResponse, SingleTypeAnnotationResponse, TypeAnnotationResponse,
-        VariableAnnotationsByConjunctionResponse, VariableAnnotationsResponse,
-    };
-    use crate::service::http::message::query::{
+    use super::{FetchStructureAnnotationsResponse, FunctionReturnAnnotationsResponse, SingleTypeAnnotationResponse, TypeAnnotationResponse, ConjunctionAnnotationsResponse, VariableAnnotationsResponse};
```
krishnangovindraj, line 249: Below this line is BDD.
A cleanup thanks to slightly more permissive macros.

server/service/http/message/query/mod.rs
```diff
@@ -153,175 +134,3 @@ impl IntoResponse for QueryAnswer {
         (code, body).into_response()
     }
 }
-
```
krishnangovindraj, line 156: Got it's own module at `analyze/mod.rs`

tests/behaviour/service/http/http_steps/query.rs
krishnangovindraj, line 1: No functional changes. just renames

server/service/http/message/analyze/annotations.rs
krishnangovindraj, line 1: There's a case to refactor the this and structure.rs since they're no longer distinct.

server/service/http/message/analyze/studio.rs
```diff
@@ -0,0 +1,123 @@
...
+use crate::service::http::message::{
+    analyze::structure::{
+        encode_analyzed_pipeline, AnalyzedPipelineResponse, StructureConstraint, StructureConstraintSpan,
+        StructureConstraintWithSpan, StructureVariableInfo, StructureVertex,
+    },
+    query::concept::RoleTypeResponse,
+};
+
+// Backwards compatibility
+pub fn encode_analyzed_pipeline_for_studio(
```
flyingsilverfin: Studio is the consumer, but it's really v0 of structure that goes into query -- perhaps `analyze/compat_v0` instead of `studio` throughout?

server/service/http/message/query/row.rs
```diff
@@ -7,21 +7,25 @@
...
+    http::message::query::concept::{encode_thing_concept, encode_type_concept, encode_value},
+    IncludeInvolvedBlocks,
+};
 
 #[derive(Debug, Serialize)]
 #[serde(rename_all = "camelCase")]
 struct EncodedRow<'a> {
     data: HashMap<&'a str, serde_json::Value>,
-    involved_blocks: Vec<u16>, // TODO: Rename if we break studio compatibility
+    involved_blocks: Option<Vec<u16>>, // TODO: Rename if we break studio compatibility
```
flyingsilverfin, line 28: Is this comment still accurate with the new structure?

server/service/http/message/analyze/annotations.rs
flyingsilverfin: This is what we have in the output now right? we merge structure and analyzed annotations into one interleaved thing? What do we gain currently by having them separate vs merged? keeping in mind we still only include structure for Query structure output right?

krishnangovindraj: They're merged in this PR.
We gain nothing keeping them separate except reusability / that we could return one without the other.
At some point I thought it would be cleaner to have structure, annotations and plan as top-level siblings. 
Now I think plan can just be an optional in a Conjunction.
Doesn't sound very interesting.

server/service/http/message/query/row.rs
krishnangovindraj: Yeah. It would ideally be called involved_conjunctions, but renaming it now would break studio

compiler/query_structure.rs
flyingsilverfin: yeah agree

server/service/http/message/analyze/studio.rs
krishnangovindraj: Well it's an undocumented feature that's only used in studio. 
We're plain breaking the analyze endpoint

resource/constants.rs
```diff
@@ -53,6 +53,8 @@ pub mod server {
     pub const DEFAULT_INCLUDE_INSTANCE_TYPES_FETCH: bool = false;
     pub const DEFAULT_ANSWER_COUNT_LIMIT_GRPC: Option<usize> = None;
     pub const DEFAULT_ANSWER_COUNT_LIMIT_HTTP: Option<usize> = Some(10_000);
+    pub const DEFAULT_INCLUDE_STRUCTURE_HTTP: bool = true;
```
flyingsilverfin: leave a comment: `true for backwards compatibility with Studio`

PR #7627, opened 2025-11-10
Ignore already processed commit records in storage

farost:
Introduce a new `CommitRecord` field called `global_causality_number` to provide a way for the libraries built on top of TypeDB to hint that two separately submitted commits are in fact identical.

A simple usage example is using TypeDB in a distributed environment. Consensus algorithms like Raft consider requests to be idempotent for the state machine and can require multiple resubmissions of the same command, e.g., as a part of recovery. Now, such systems can build their own unique growing identifiers (causality numbers) and assign them to their commit-related commands. In the context of Raft, a great causality number is Raft log entry's index.

**Important**: once you start using causality numbers, you must continue to do so, and each number **must** be bigger than the previous one. Otherwise, excessive records will be ignored without records, and the data will be lost.

This change **does not** break backward compatibility. Databases created using prior versions of TypeDB will successfully be loaded. However, you will not be able to use a database modified in this version by an older release of TypeDB. 

## Implementation
Add `global_causality_number` to `CommitRecord`, extending the structure by 8 bytes (`Option<NonZeroU64>`). Additionally, save the same value in `SchemaSnapshot` and `WriteSnapshot`.

In `storage`, add a method to check if the causality number of an incoming record is below the last preserved record and, thus, its record must be ignored. This approach uses the heuristics mentioned in "Product change and motivation".

**IMPORTANT** 
I'm happy to think about the name and the docs for the field more.
I'm not entirely happy with the risks of a user-side error, but it's the tradeoff for simplicity...

### DIFF
--- a/storage/storage.rs
+++ b/storage/storage.rs
@@ -636,15 +636,18 @@ impl StorageOperation {
 
 #[cfg(test)]
 mod tests {
+    use std::num::NonZeroU64;
+
     use bytes::byte_array::ByteArray;
     use durability::wal::WAL;
-    use resource::profile::StorageCounters;
+    use resource::profile::{CommitProfile, StorageCounters};
     use test_utils::{create_tmp_storage_dir, init_logging};
 
     use crate::{
         durability_client::{DurabilityClient, WALClient},
         key_value::StorageKeyArray,
         keyspace::{IteratorPool, KeyspaceId, KeyspaceSet, Keyspaces},
+        number::SequenceNumber,
         record::{CommitRecord, CommitType},
         snapshot::buffer::OperationsBuffer,
         write_batches::WriteBatches,
@@ -718,5 +721,98 @@ mod tests {
             storage.get::<0>(&IteratorPool::new(), &key_2, seq, StorageCounters::DISABLED).unwrap().unwrap(),
             ByteArray::empty()
         );
+    }
+
+    fn run_casuality_test(init_casuality: Option<u64>, new_casuality: Option<u64>, expected_new_write: bool) {
+        init_logging();
+        let storage_path = create_tmp_storage_dir();
+
+        test_keyspace_set! {
+            PersistedKeyspace => 0: "write",
+        }
+
+        fn commit_record_with_sequence_and_casuality<const A: usize>(
+            open_sequence_number: SequenceNumber,
+            casuality: Option<u64>,
+            data: &[u8; A],
+        ) -> CommitRecord {
+            let key_1 = StorageKeyArray::from((TestKeyspaceSet::PersistedKeyspace, data));
+            let mut operations = OperationsBuffer::new();
+            operations.writes_in_mut(key_1.keyspace_id()).insert(key_1.byte_array().clone(), ByteArray::empty());
+
+            let mut record = CommitRecord::new(operations, open_sequence_number, CommitType::Data);
+            if let Some(value) = casuality {
+                record.global_causality_number = NonZeroU64::new(value);
+            }
+            record
+        }
+
+        fn storage_with_existing_commit(
+            storage_path: std::path::PathBuf,
+            existing_causality: Option<u64>,
+        ) -> MVCCStorage<WALClient> {
+            let mut durability_client = WALClient::new(WAL::create(storage_path.join(WAL::WAL_DIR_NAME)).unwrap());
+            durability_client.register_record_type::<CommitRecord>();
+            let storage =
+                MVCCStorage::<WALClient>::create::<TestKeyspaceSet>("storage", &storage_path, durability_client)
+                    .unwrap();
+
+            let existing =
+                commit_record_with_sequence_and_casuality(SequenceNumber::from(0), existing_causality, b"hello");
+            let mut profile = CommitProfile::DISABLED;
+            let seqnum = storage.commit(existing, &mut profile).unwrap();
+            debug_assert_eq!(seqnum, Some(SequenceNumber::from(1)), "Expected to be the first record in the WAL");
+            storage
+        }
+
+        let storage = storage_with_existing_commit(storage_path.to_path_buf(), init_casuality);
+        // TODO: If we use SequenceNumber::from(1) instead, it crashes
+        let new_record = commit_record_with_sequence_and_casuality(SequenceNumber::from(0), new_casuality, b"world");
+        let mut profile = CommitProfile::DISABLED;
+        let seqnum = storage.commit(new_record, &mut profile).unwrap();
+        match expected_new_write {
+            true => {
+                assert!(
+                    seqnum.is_some(),
+                    "Expected a new write, but the record was ignored due to a big causality number"
+                );
+            }
+            false => {
+                assert!(
+                    seqnum.is_none(),
+                    "Expected no new writes, but the record was applied due to a small causality number"
+                );
+            }
+        }
+    }
+
+    #[test]
+    fn test_causality_bigger_than_previous_is_not_ignored() {
+        run_casuality_test(Some(10), Some(11), true)
+    }
+
+    #[test]
+    fn test_causality_smaller_than_previous_is_ignored() {
+        run_casuality_test(Some(10), Some(9), false)
+    }
+
+    #[test]
+    fn test_causality_same_as_previous_is_ignored() {
+        run_casuality_test(Some(10), Some(10), false)
+    }
+
+    #[test]
+    fn test_causality_number_after_none_is_not_ignored() {
+        run_casuality_test(None, Some(9), true)
+    }
+
+    #[test]
+    fn test_causality_none_after_some_is_not_ignored() {
+        run_casuality_test(Some(11), None, true)
+    }
+
+    #[test]
+    fn test_causality_none_after_none_is_not_ignored() {
+        run_casuality_test(None, None, true)
     }
 }


concept/tests/test_statistics.rs
```diff
@@ -21,7 +21,7 @@ use encoding::value::{label::Label, value::Value, value_type::ValueType};
 use resource::profile::{CommitProfile, StorageCounters};
 use storage::{
     durability_client::WALClient,
-    sequence_number::SequenceNumber,
+    number::SequenceNumber,
```
farost, line 24: Places both `SequenceNumber` type and the new number type definitions into the same small file.

storage/isolation_manager.rs
```diff
@@ -650,196 +644,6 @@ impl SlotMarker {
     }
 }
 
-// TODO: move out of isolation manager
```
farost: Moved to a new file called "record"

storage/number.rs
```diff
@@ -3,9 +3,12 @@
...
  */
+use std::num::NonZeroU64;
 
 use durability::DurabilitySequenceNumber;
 
 // TODO: when we introduce partitioning/remote WAL, we need create a separate type of the sequence number, instead of
 //       passing through to the Durability sequence number
 pub type SequenceNumber = DurabilitySequenceNumber;
+
+pub type CausalityNumber = Option<NonZeroU64>;
```
farost, line 14: `Option<NonZeroU64>` takes 8 bytes, while `Option<u64>` takes 9 bytes (1 for the option flag + the value). 

Pros:
1. We use less space in a critical region
2. Cluster should not use index 0 for commits anyway

Cons:
1. Someone could want to pass 0. It's type safe, so they'd need to either deliberately ignore the check or modify the number. 

I prefer the pros.

storage/record.rs
```diff
@@ -0,0 +1,245 @@
...
+};
+
+#[derive(Serialize, Deserialize)]
+pub struct CommitRecord {
+    // TODO: this could read-through to the WAL if we have to save memory?
+    operations: OperationsBuffer,
+    open_sequence_number: SequenceNumber,
+    commit_type: CommitType,
+
+    /// An optional field used by extensions of TypeDB to provide external causality numbers.
```
farost, line 27: Tried to write up some docs

storage/record.rs
```diff
@@ -0,0 +1,245 @@
...
+    /// Records with the same causality numbers are considered identical, and only the first record
+    /// is processed, ensuring idempotency.
+    /// Each new record expected to be applied must have a new causality number bigger than all the
+    /// causality numbers already saved in the storage. Otherwise, it will be ignored.
+    /// Defaults to 0 (meaning that every record is unique).
+    #[serde(default)]
```
farost, line 33: Not really needed as we use `bincode` (so we need to deserialize carefully anyway), but might play as a good marker that it's an "additional" field. And, of course, can be useful for JSON, etc.

storage/record.rs
```diff
@@ -0,0 +1,245 @@
...
+    pub global_causality_number: CausalityNumber,
```
farost, line 34: We really struggle to come up with a good name. At this point, I got used to `global causality number`:
* `global` -- a part of a bigger domain, managed globally. Might be `external`, but it feels even more confusing
* `causality` -- provides specific ordering, does not guarantee gapless sequences. `global causality` highlights that it's different from `seqnum`
* number -- number

Happy to think more.

farost, line 34: We allow external users to modify this number. It can either be `pub` or have setters. I personally prefer setters, but this is shorter. What do you think?

storage/record.rs
```diff
@@ -0,0 +1,245 @@
...
+}
+
+#[derive(Serialize, Deserialize)]
+struct LegacyCommitRecordV1 {
```
farost, line 38: As we use `bincode` for serialization and deserialization, we need to explicitly support different encoding schemas. Since I moved records to a separate file, I just inserted a special `Legacy` record struct for the old version of records. We can use the same approach for any other automatic migration.

storage/record.rs
```diff
@@ -0,0 +1,245 @@
...
+    pub fn open_sequence_number(&self) -> SequenceNumber {
+        self.open_sequence_number
+    }
+
+    fn deserialise_from(record_type: DurabilityRecordType, reader: impl Read)
+    where
+        Self: Sized,
+    {
+        assert_eq!(Self::RECORD_TYPE, record_type);
+        // TODO: handle error with a better message
```
farost, line 112: Old todos, just copied the code from the original place

storage/record.rs
```diff
@@ -0,0 +1,245 @@
...
+            bincode::serialize(self).unwrap()
+        );
+        bincode::serialize_into(writer, &self)
+    }
+
+    fn deserialise_from(reader: &mut impl Read) -> bincode::Result<Self> {
+        // https://github.com/bincode-org/bincode/issues/633
+        let mut buf = Vec::new();
+        reader.read_to_end(&mut buf).map_err(|e| bincode::ErrorKind::Io(e))?;
+        match bincode::deserialize::<CommitRecord>(&buf) {
```
farost, line 196: Try deserializing as the new type, fallback to the legacy type otherwise

storage/snapshot/snapshot.rs
```diff
@@ -349,40 +350,49 @@ impl<D> ReadableSnapshot for ReadSnapshot<D> {
...
+            .field("global_causality_number", &self.global_causality_number)
+            .finish()
     }
 }
 
 impl<D> WriteSnapshot<D> {
     pub(crate) fn new_with_open_sequence_number(
         storage: Arc<MVCCStorage<D>>,
         open_sequence_number: SequenceNumber,
+        global_causality_number: CausalityNumber,
```
farost, line 371: We have to keep these causality numbers inside `Schema`/`Write` snapshots to generate commit records. We don't modify it, we just pass it from the external caller to the record.

storage/snapshot/snapshot.rs
```diff
@@ -349,40 +350,49 @@ impl<D> ReadableSnapshot for ReadSnapshot<D> {
...
     }
 
     pub fn new_with_commit_record(storage: Arc<MVCCStorage<D>>, commit_record: CommitRecord) -> Self {
         let open_sequence_number = commit_record.open_sequence_number();
-        Self::new(storage, commit_record.into_operations(), open_sequence_number)
+        let global_causality_number = commit_record.global_causality_number;
+        Self::new(storage, commit_record.into_operations(), open_sequence_number, global_causality_number)
     }
 
-    pub fn new_with_operations(
```
farost: Unused

storage/snapshot/snapshot.rs
```diff
@@ -676,7 +700,7 @@ impl<D: DurabilityClient> CommittableSnapshot<D> for SchemaSnapshot<D> {
         let storage = self.storage.clone();
         match self.finalise(commit_profile)? {
             Some(commit_record) => match storage.commit(commit_record, commit_profile) {
-                Ok(sequence_number) => Ok(Some(sequence_number)),
```
farost: This is important. See comment regarding optionality in `storage`.

storage/storage.rs
```diff
@@ -177,13 +179,13 @@ impl<Durability> MVCCStorage<Durability> {
         // guarantee external consistency: we always await the latest snapshots to finish
         let possible_sequence_number = self.isolation_manager.highest_validated_sequence_number();
         let open_sequence_number = self.wait_for_watermark(possible_sequence_number);
-        WriteSnapshot::new_with_open_sequence_number(self, open_sequence_number)
+        WriteSnapshot::new_with_open_sequence_number(self, open_sequence_number, CausalityNumber::None)
```
farost, line 182: By default, open without the causality number.
I am a little worried that we might open this code more, so we leave a way to accidentally break causality, but I'm not sure how to protect from it without a big refactor.

storage/storage.rs
```diff
@@ -220,12 +222,20 @@ impl<Durability> MVCCStorage<Durability> {
         &self,
         commit_record: CommitRecord,
         commit_profile: &mut CommitProfile,
-    ) -> Result<SequenceNumber, StorageCommitError>
+    ) -> Result<Option<SequenceNumber>, StorageCommitError>
```
farost, line 225: In all the users of this method, we used to write `Some(seqnum)`. There, `None` meant "nothing was done".
I make this an "official" method's signature. 
Cannot commit -> `Err(_)`.
Can commit, but no need -> `Ok(None)`
Committed -> `Ok(Some(_))`

Why? Because I don't want to search for the specific seqnum of an already applied record. `self.record_exists` below uses heuristics mentioned in the field's docs for optimization. 
Is it bad? Currently, I don't think so. It may become a problem if we ever want to really use the output `seqnum`. As far as I know, its always like
`match commit_result { Ok(_) => Ok(()), ... }`

storage/storage.rs
```diff
@@ -281,6 +291,23 @@ impl<Durability> MVCCStorage<Durability> {
...
+    fn record_exists(&self, commit_record: &CommitRecord) -> Result<bool, DurabilityClientError>
+    where
+        Durability: DurabilityClient,
+    {
+        let new_causality_number = commit_record.global_causality_number;
+        if new_causality_number.is_none() {
+            return Ok(false);
+        }
+        self.durability_client
+            .iter_sequenced_type_from::<CommitRecord>(self.durability_client.previous())?
```
farost, line 303: Just check the previous record and only the previous record. 

WARNING: If the external user sends unordered records or decides to insert a record with `None`, everything will break. We have to ensure we're transparent with these rules.

storage/storage.rs
```diff
@@ -692,4 +722,97 @@ mod tests {
...
+            let existing =
+                commit_record_with_sequence_and_casuality(SequenceNumber::from(0), existing_causality, b"hello");
+            let mut profile = CommitProfile::DISABLED;
+            let seqnum = storage.commit(existing, &mut profile).unwrap();
+            debug_assert_eq!(seqnum, Some(SequenceNumber::from(1)), "Expected to be the first record in the WAL");
+            storage
+        }
+
+        let storage = storage_with_existing_commit(storage_path.to_path_buf(), init_casuality);
+        // TODO: If we use SequenceNumber::from(1) instead, it crashes
```
farost, line 769: I expect it to be `true` as we open after seqnum 1 is committed. 
The failure is:
```
thread 'tests::test_causality_number_after_none_is_not_ignoredtest isolation_manager::tests::test_highly_concurrent_correctness ... ' panicked at storage/isolation_manager.rs:602:9:
attempt to subtract with overflow
thread 'tests::test_causality_bigger_than_previous_is_not_ignored' panicked at storage/isolation_manager.rs:602:9:
attempt to subtract with overflow
```

So either I use it incorrectly or there is a bug. Maybe I don't understand the meaning of open seqnums, will explore it additionally.

diagnostics/metrics.rs
flyingsilverfin, line 861: externally we're using american spelling: 'analyze'

diagnostics/reporter.rs
flyingsilverfin, line 86: Is this where we're changing to always have an initial delay of 1 hour? We had floated this recently to avoid reporting testing instances of TypeDB

flyingsilverfin: Or had we decided not to do this change since it would change the stats we see and means we don't have continuity with the data collection policy

durability/wal.rs
flyingsilverfin, line 52: Good change!

flyingsilverfin, line 341: can this error?

storage/snapshot/snapshot.rs
flyingsilverfin, line 214: don't forget these - they're in a couple places

storage/number.rs
flyingsilverfin: That's cool - i like it! 

i'd make a note of why this is chosen in the code directly. What's the minimum durability sequence number?

storage/record.rs
flyingsilverfin: So it basically: bincode will error if it can't deserialize the record as the new type, which has more fields, since even that optional number will take up 8 bytes. If this error occurs, we can use the old record type - right?

flyingsilverfin: If we implemented our own serialize/deserialize traits for Serde (instead of deriving it), we could do this transparently? 

eg. if we try to deserialize the fields, and we run out of bytes to deserialize, we can just return the new format with the Option set to `None`?

flyingsilverfin: That being said, i'm not entirely sure that's clearer. It should be equivalent and shorter, but this might be clearer. Not sure! Let's think about what would happen if we wanted to add another field in the future for example...

storage/storage.rs
flyingsilverfin: You resolved this right?

user/user_manager.rs
flyingsilverfin, line 47: This API is now strange IMO - we say "create user", but we don't really do it, we get a "transaction intent" back... this feels like an abstraction mismatch. User management operations should a status indicating a completed user action, or a "PendingOperation" struct or something... 


but in general, I'm not sure without digging in further why this can't just commit fully?

.factory/automation.yml
```diff
@@ -67,6 +68,7 @@ build:
       dependencies: [build]
       command: |
         bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
+        bazel test //common/concurrency:test_crate_concurrency --test_output=streamed
```
flyingsilverfin, line 71: wasn't this in a separate PR already? do you need to rebase this?

server/lib.rs
```diff
@@ -57,94 +75,202 @@ impl ServerBuilder {
...
+            distribution_info,
+            config,
+            server_state,
+            shutdown_sender,
+            shutdown_receiver,
+            background_tasks_tracker,
+        ))
+    }
+
+    pub fn initialise_storage(&mut self, storage_config: &StorageConfig) -> Result<&str, ServerOpenError> {
```
flyingsilverfin, line 120: This shouldn't be pub

server/state.rs
```diff
@@ -4,236 +4,235 @@
...
+
+    async fn databases_get(&self, name: &str) -> Result<Option<Arc<Database<WALClient>>>, ArcServerStateError>;
+
+    async fn databases_get_unrestricted(
+        &self,
+        name: &str,
+    ) -> Result<Option<Arc<Database<WALClient>>>, ArcServerStateError>;
 
-    fn users_all(&self, accessor: Accessor) -> Result<Vec<User>, ServerStateError>;
+    async fn databases_contains(&self, name: &str) -> Result<bool, ArcServerStateError>;
```
flyingsilverfin, line 67: I think this might be something I need to go through with @lolski  and @farost  - I'm struggling a bit with the conceptual model.

We have `ServerState`, which encompasses all the state of the system. We have methods on it directly which can mutate all parts of the state directly (eg. `databases_create` or `token_create`. However, you can _also_ get the `DatabaseManager` from the state, and then do a `put_database` on there, which I presume is unsafe in some form when used in other contexts?

My instinct is that the DatabaseManager, UserManager, etc. should either be hidden entirely, or should be actually implemented by injecting the ServerState into them, and then we treat them as part of `ServerState` somehow (eg. this would eliminate `ServerState::databases_create()`, you would have to do `ServerState::database_manager::databases_create()` or something)

Not sure how related to this PR, but I'm picking it up perhaps more strongly because we're exposing "partial" operations from the UserManager now and that feels odd.
Also we now have an asymmetry: to mutate users, you do `ServerState.user_manager().create_user()`, and then do something with the `Intent` - but for databases, you can directly do `ServerState::databases_create()`...

flyingsilverfin: I suppose one conceptual starting question: is it safe to from within a method somewhere, such as I dunno... `import_servce` to create a database, since it wouldn't go through the ServerState?

server/service/grpc/transaction_service.rs
```diff
@@ -534,19 +466,44 @@ impl TransactionService {
...
+                let (profile, commit_result) = match into_commit_record_result {
+                    Ok((database, Some(commit_record))) => {
+                        let commit_result = server_state
+                            .database_data_commit(database.name(), commit_record, profile.commit_profile())
+                            .await;
+                        (profile, commit_result)
+                    }
+                    Ok((_, None)) => (profile, Ok(())),
+                    Err(error) => {
+                        (profile, Err(LocalServerStateError::DatabaseDataCommitFailed { typedb_source: error }.into()))
```
flyingsilverfin, line 503: Why is this so long in the GRPC transaction service, while the HTTP service continues to use the old `transaction.commit()`? That method by the way seems to do a lot of this right? 

`database::transaction::commit()` does:
1. `self.finalize()
2. `database.data_commit_with_snapshot()`

same as here, but with less error handling?

flyingsilverfin: Ah I see that you want to eliminate `Transaction::commit` entirely! Interesting!

Have we considered pushing the `ServerState` trait wayyyy down to the bottom layer, and then parametrizing our transactions, database manager, user manager etc. with that trait? Then we can keep `Transaction::commit` as the normal method, but it directs its operation into the `ServerState`? Just a thought... cc @lolski

flyingsilverfin: Or, instead of using a trait parameter, we could use `Arc<dyn ServerState>` as well, since it's not exactly on the hot path (hottest is 1 commit per transaction, but not on every read/write from RocksDB)

server/state.rs
```diff
@@ -262,169 +262,279 @@ impl LocalServerState {
...
+        self.server_status().await.map(|status| vec![status])
+    }
+
+    async fn servers_register(
+        &self,
+        _clustering_id: u64,
+        _clustering_address: String,
+    ) -> Result<(), ArcServerStateError> {
+        // todo: error message
+        Err(Arc::new(LocalServerStateError::Unimplemented {
```
flyingsilverfin, line 369: This should be something specific, not `Unimplemented`, such as `NotAvailableInCE` with message `This feature is not available in TypeDB CE, consider using TypeDB Cloud or Enterprise instead.`

server/service/grpc/typedb_service.rs
```diff
@@ -146,72 +159,183 @@ impl typedb_protocol::type_db_server::TypeDb for TypeDBService {
...
+        request: Request<typedb_protocol::server_manager::register::Req>,
+    ) -> Result<Response<typedb_protocol::server_manager::register::Res>, Status> {
+        run_with_diagnostics_async(
+            self.server_state.diagnostics_manager().await,
+            None::<&str>,
+            ActionKind::ServersRegister,
+            || async {
+                let request = request.into_inner();
+
+                let typedb_protocol::server_manager::register::Req { address, replica_id } = request else {
```
flyingsilverfin, line 222: IntelliJ is giving me a `redundant else` warning! Can be simplified therefore

server/service/grpc/transaction_service.rs
flyingsilverfin: Ya ok i suppose pushing the trait way down wouldn't work since the signatures depend on APIs in the middle packages eg `Database` hmmm

system/util.rs
flyingsilverfin, line 53: unused result

flyingsilverfin, line 41: let's get on a call tmrw for sure - curious especially about this and the changes to UserManager

storage/record.rs
farost: Yeah, it depends on the difference between the old and the new data type. If the number of bytes is the same, but the type is different, it may lead to incorrect data. We could handle it separately. Do we expect to do so, though? I don't think we're interested in changing the data types of the storage records just because we want to, it's still a risky migration process.

storage/storage.rs
farost: Not yet, I was working on other cluster things. This specific error should not be introduced by my changes and is a storage thing... I will give it another go today.

diagnostics/metrics.rs
farost: Should've been caught here https://github.com/typedb/typedb/commit/2ba025197dd4c69b5e1cc5f61d5ada7320013191#diff-1f1630a834c2ff9ca2c7cb6a5316ea7f70f4872a6a42ffa3ad56df362c251028

durability/wal.rs
farost: Probably. That's why it's `_best_effort`

storage/snapshot/snapshot.rs
farost: It's Ganesh's comments, we already discussed it.

user/user_manager.rs
farost: It's done for replication and is an old change by Ganesh. There are trade-offs. I thought this moment was discussed by you internally. I think you want to talk with @lolski about this `user_manager` design.

.factory/automation.yml
farost: Yes =(

server/state.rs
farost: No, it's not safe, and we are not fully satisfied with this design. Good questions raised, that's what I'm iterating through in my "cleanup before alpha work" since I haven't touched these areas myself.

system/util.rs
farost: It's not related to this PR, a rebase is required. Sorry for the confusion caused.

farost:
This PR is polluted by conflict-related diffs after the latest target branch's rebase. I moved all the needed commits and comments to https://github.com/typedb/typedb/pull/7638

server/state.rs
farost: Resolved separately, not new

server/service/grpc/transaction_service.rs
farost: @lolski for you, too

PR #7628, opened 2025-11-13
Add in panic logging via the global panic_hook

flyingsilverfin:

We ensure that panics are written to the configured `tracing` log file, by intercepting the panic event and routing it via the logger. This code is borrowed from https://github.com/LukeMathWalker/tracing-panic ! 

## Implementation

We decided not to use the published crate for such a small function to minimize dependencies and future security validation work.

### DIFF
--- a/server/lib.rs
+++ b/server/lib.rs
@@ -21,6 +21,7 @@ use tokio::{
     sync::watch::{channel, Receiver, Sender},
 };
 use tracing::info;
+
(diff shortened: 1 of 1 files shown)


common/logger/log_panic.rs
krishnangovindraj, line 1: Shouldn't this have either of the original licenses  in the header or something?

PR #7632, opened 2025-11-17
Consider include_query_structure option

krishnangovindraj:
We only include the pipeline structure in concept row responses if the include_query_structure flag in query options is set.

### DIFF
--- a/server/service/grpc/transaction_service.rs
+++ b/server/service/grpc/transaction_service.rs
@@ -79,6 +79,7 @@ use crate::service::{
         },
         row::encode_row,
     },
+    may_encode_pipeline_structure,
     transaction_service::{
         init_transaction_timeout, is_write_pipeline, with_readable_transaction, Transaction, TransactionServiceError,
     },
@@ -1143,10 +1144,9 @@ impl TransactionService {
         mut interrupt: ExecutionInterrupt,
         storage_counters: StorageCounters,
     ) {
-        let may_encode_result =
-            IncludeInvolvedBlocks::may_encode_pipeline_structure(&query_options, pipeline_structure, |structure| {
-                encode_analyzed_pipeline_for_query(snapshot.as_ref(), &type_manager, structure)
-            });
+        let may_encode_result = may_encode_pipeline_structure(&query_options, pipeline_structure, |structure| {
+            encode_analyzed_pipeline_for_query(snapshot.as_ref(), &type_manager, structure)
+        });
         let (encoded_structure, include_involved_blocks) = unwrap_or_execute_and_return!(may_encode_result, |err| {
             Self::submit_response_sync(
                 &sender,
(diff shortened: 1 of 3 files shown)


concept/thing/relation.rs
```diff
@@ -444,10 +444,12 @@ impl ThingAPI for Relation {
                 .get_indexed_relations(snapshot, thing_manager, self.type_(), storage_counters.clone())
                 .is_ok_and(|mut iterator| iterator.any(|result| {
                     match result {
-                        Ok(((start, _, relation_type, relation_id, start_role, _), _)) => start == player
-                            && start_role == role
-                            && relation_type == self.type_().vertex().type_id_()
-                            && relation_id == self.vertex().object_id(),
+                        Ok(((start, _, relation_type, relation_id, start_role, _), _)) => {
```
krishnangovindraj, line 447: rustfmt

server/service/grpc/transaction_service.rs
krishnangovindraj, line 1: All the changes are basically an if-else around the existing code.

server/service/http/transaction_service.rs
```diff
@@ -1070,22 +1075,26 @@ impl TransactionService {
...
+                        respond_error_and_return_break!(
+                            responder,
+                            TransactionServiceError::PipelineExecution {
+                                typedb_source: PipelineExecutionError::ConceptRead { typedb_source }
+                            }
+                        );
+                    }
+                };
+                let include_involved_blocks = IncludeInvolvedBlocks::build(pipeline.pipeline_structure());
+                (pipeline_structure_response, include_involved_blocks)
```
dmitrii-ubskii: Can the body here be a function? Or even the entire `if`. It seems to repeat three times across two modules.

dmitrii-ubskii (approved):
Approved with a suggestion.

server/service/mod.rs
```diff
@@ -42,12 +44,18 @@ pub(crate) enum IncludeInvolvedBlocks {
...
+    ) -> Result<(Option<T>, Self), Box<ConceptReadError>> {
+        match (&options.include_query_structure, pipeline) {
+            (false, _) | (true, None) => Ok((None, IncludeInvolvedBlocks::False)),
+            (true, Some(structure)) => {
+                let include_involved_blocks = IncludeInvolvedBlocks::True {
+                    always_involved: structure.parametrised_structure.always_involved_blocks(),
+                };
+                encoder(structure).map(|encoded| (Some(encoded), include_involved_blocks))
             }
         }
```
dmitrii-ubskii: You could use `Option::filter` here:
```rust
if let Some(structure) = pipeline.filter(|_| options.include_query_structure) {
    // ...
} else {
    Ok((None, IncludeInvolvedBlocks::False))
}
```
Or, if we update to 1.88+:
```rust
if let Some(structure) = pipeline && options.include_query_structure {
    // ...
} else {
    Ok((None, IncludeInvolvedBlocks::False))
}
```

server/service/mod.rs
```diff
@@ -42,12 +44,18 @@ pub(crate) enum IncludeInvolvedBlocks {
...
-    pub(crate) fn build(pipeline_structure: Option<&PipelineStructure>) -> Self {
-        match pipeline_structure {
-            None => IncludeInvolvedBlocks::False,
-            Some(structure) => {
-                let always_involved = structure.parametrised_structure.always_involved_blocks();
-                IncludeInvolvedBlocks::True { always_involved }
+    pub(crate) fn may_encode_pipeline_structure<T>(
+        options: &QueryOptions,
+        pipeline: Option<&PipelineStructure>,
+        encoder: impl Fn(&PipelineStructure) -> Result<T, Box<ConceptReadError>>,
```
dmitrii-ubskii: This seems unnecessary.

server/service/mod.rs
```diff
@@ -42,12 +44,18 @@ pub(crate) enum IncludeInvolvedBlocks {
 }
 
 impl IncludeInvolvedBlocks {
-    pub(crate) fn build(pipeline_structure: Option<&PipelineStructure>) -> Self {
-        match pipeline_structure {
-            None => IncludeInvolvedBlocks::False,
-            Some(structure) => {
-                let always_involved = structure.parametrised_structure.always_involved_blocks();
-                IncludeInvolvedBlocks::True { always_involved }
+    pub(crate) fn may_encode_pipeline_structure<T>(
```
dmitrii-ubskii: Why is this on `IncludeInvolvedBlocks`? It doesn't look like a constructor to me.

krishnangovindraj: Can be a free function if you think that's better.

PR #7637, opened 2025-11-20
Complete LocalServerState methods: optional server statuses and user management

farost:

Update protocol (https://github.com/typedb/typedb-protocol/pull/240).
Replicate user management commands to ensure that every replica correctly invalidates tokens when needed (thus, it's different from a simple database commit operation).
Fix unsynced `ServerState` interfaces to use `&str` instead of owned types when needed.
Cleanup and rustfmt.

## Implementation
Move the content of `service/typedb_service.rs` to `state.rs`, similarly to all the other calls.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5220,7 +5220,7 @@ dependencies = [
 [[package]]
 name = "typedb-protocol"
 version = "0.0.0"
-source = "git+https://github.com/typedb/typedb-protocol?rev=6ff962e6ca719d76a99433c6c20fcfe4bbbe094c#6ff962e6ca719d76a99433c6c20fcfe4bbbe094c"
+source = "git+https://github.com/typedb/typedb-protocol?rev=1e5f10e9ee60b467190ac3a9d657616f6d22098c#1e5f10e9ee60b467190ac3a9d657616f6d22098c"
 dependencies = [
  "prost",
  "tonic",
--- a/dependencies/typedb/repositories.bzl
+++ b/dependencies/typedb/repositories.bzl
@@ -31,7 +31,7 @@ def typedb_protocol():
     git_repository(
         name = "typedb_protocol",
         remote = "https://github.com/typedb/typedb-protocol",
-        commit = "6ff962e6ca719d76a99433c6c20fcfe4bbbe094c",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_protocol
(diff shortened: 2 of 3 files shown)


resource/profile.rs
```diff
@@ -290,7 +295,7 @@ impl CommitProfile {
 struct CommitProfileData {
     counters: StorageCounters,
     commit_size: usize,
-    stage_start: Instant,
+    stage_start: Option<Instant>,
```
farost, line 298: Agreed that this should be an optional

server/service/export_service.rs
```diff
@@ -39,7 +39,7 @@ fn get_functions_syntax<D: DurabilityClient>(transaction: &TransactionRead<D>) -
 }
 
 typedb_error! {
-    pub(crate) DatabaseExportError(component = "Database export", prefix = "DBE") {
```
farost, line 42: Cleanup warnings

server/service/grpc/typedb_service.rs
```diff
@@ -167,19 +162,6 @@ impl typedb_protocol::type_db_server::TypeDb for GRPCTypeDBService {
         .await
     }
 
-    async fn server_version(
```
farost, line 170: Reorder

server/service/grpc/typedb_service.rs
```diff
@@ -258,19 +234,149 @@ impl typedb_protocol::type_db_server::TypeDb for GRPCTypeDBService {
...
+                    .users_contains(accessor, name.as_str())
+                    .await
+                    .map(|contains| Response::new(users_contains_res(contains)))
+                    .map_err(|err| err.into_error_message().into_status())
+            },
+        )
+        .await
+    }
+
+    async fn users_create(
```
farost, line 315: Move from `service/typedb_service.rs`

database/transaction.rs
```diff
@@ -163,6 +163,7 @@ impl<D: DurabilityClient> TransactionWrite<D> {
     pub fn finalise(self) -> (TransactionProfile, Result<DataCommitIntent<D>, DataCommitError>) {
         let mut profile = self.profile;
         let commit_profile = profile.commit_profile();
+        commit_profile.start();
```
farost, line 166: We have to **always** start these profiles, otherwise, commits fail with tracing enabled.

lolski (approved):

PR #7638, opened 2025-11-26
Ignore already processed commit records in storage

farost:
### Snapshot ids
Add locally generated `snapshot_id`s for snapshots and `CommitRecord`s. These ids are used in combination with `open_sequence_number` to uniquely identify currently opened snapshots and transactions, as well as committed commit records. 

A new method `commit_record_exists` is added. It can be used to check if a `CommitRecord` with the provided `open_sequence_number` and `snapshot_id` exists locally. This method can be useful for an extension of TypeDB. E.g., TypeDB Cluster uses this method to avoid replaying already existing `CommitRecord`s on recovery, protecting idempotency. 

**Important**: Currently, snapshot ids are generated locally, which can cause collisions if you depend on them without enough awareness. **However**, they are totally safe for a single instance of TypeDB.

**Also important**: `commit_record_exists` can be quite expensive if the record is old, and can be **wrong** if the record is **very** old (thus, removed from the WAL). It's recommended to use it in a limited way.

This change **does not** break backward compatibility. Databases created using prior versions of TypeDB will successfully be loaded. However, you will not be able to use a database modified in this version by an older release of TypeDB. 

### Extensions for TypeDB Cluster
Add new methods to serialize transaction openings and manually close specific transactions through TypeDB Cluster. This helps TypeDB Cluster make its logic safe and friendly to the user, only providing up-to-date data, and protects the uniqueness of snapshot ids in a distributed environment. 

## Implementation
Add `snapshot_id` to `CommitRecord`, extending the structure by 9 bytes (`u64`). Additionally, save the same value in Snapshots.

In `storage`, add a method to check if a commit record with the same seqnum and id exists.

To be able to close transactions without a user command, we add a new channel between transaction services and `ServerState`. The `ServerState` is able to close transactions based on their type on request. This mechanism will also allow us to implement transaction invalidation on the owner's credential updates.

For other implementation details, check out the PR comments.

### DIFF
--- a/server/state.rs
+++ b/server/state.rs
@@ -21,8 +21,8 @@ use database::{
     Database,
 };
 use diagnostics::{diagnostics_manager::DiagnosticsManager, Diagnostics};
-use itertools::Itertools;
 use durability::DurabilitySequenceNumber;
+use itertools::Itertools;
 use options::TransactionOptions;
 use resource::{
     constants::{common::SECONDS_IN_MINUTE, server::DATABASE_METRICS_UPDATE_INTERVAL},
--- a/server/service/http/typedb_service.rs
+++ b/server/service/http/typedb_service.rs
@@ -4,7 +4,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::{collections::HashMap, net::SocketAddr, sync::Arc, time::Duration};
+use std::{collections::HashMap, sync::Arc, time::Duration};
 
 use axum::{
     extract::State,


concept/tests/test_statistics.rs
```diff
@@ -21,7 +21,7 @@ use encoding::value::{label::Label, value::Value, value_type::ValueType};
 use resource::profile::{CommitProfile, StorageCounters};
 use storage::{
     durability_client::WALClient,
-    sequence_number::SequenceNumber,
+    number::SequenceNumber,
```
farost: Places both SequenceNumber type and the new number type definitions into the same small file.

storage/isolation_manager.rs
```diff
@@ -652,196 +646,6 @@ impl SlotMarker {
     }
 }
 
-// TODO: move out of isolation manager
```
farost, line 655: Moved to a new file called "record"

storage/number.rs
```diff
@@ -3,9 +3,12 @@
...
  */
+use std::num::NonZeroU64;
 
 use durability::DurabilitySequenceNumber;
 
 // TODO: when we introduce partitioning/remote WAL, we need create a separate type of the sequence number, instead of
 //       passing through to the Durability sequence number
 pub type SequenceNumber = DurabilitySequenceNumber;
+
+pub type CausalityNumber = Option<NonZeroU64>;
```
farost: `Option<NonZeroU64>` takes 8 bytes, while `Option<u64>` takes 9 bytes (1 for the option flag + the value). 

Pros:
1. We use less space in a critical region
2. Cluster should not use index 0 for commits anyway

Cons:
1. Someone could want to pass 0. It's type safe, so they'd need to either deliberately ignore the check or modify the number. 

I prefer the pros.

farost: Answering @flyingsilverfin 's comment from the previous PR: the minimal seqnum is
`pub const MIN: Self = Self { number: u64::MIN };`
The using library must ensure that it does not pass 0s, it's validated by the type system.

storage/record.rs
```diff
@@ -0,0 +1,245 @@
...
+    open_sequence_number: SequenceNumber,
+    commit_type: CommitType,
+
+    /// An optional field used by extensions of TypeDB to provide external causality numbers.
+    /// Records with the same causality numbers are considered identical, and only the first record
+    /// is processed, ensuring idempotency.
+    /// WARNING: each new record expected to be applied MUST have a new causality number bigger than all the
+    /// causality numbers already saved in the storage. Otherwise, it will be ignored.
+    /// Defaults to None (meaning that every record is unique).
+    #[serde(default)]
```
farost, line 30: Not really needed as we use bincode (so we need to deserialize carefully anyway), but might play as a good marker that it's an "additional" field. And, of course, can be useful for JSON, etc.

storage/record.rs
```diff
@@ -0,0 +1,245 @@
...
+    pub global_causality_number: CausalityNumber,
```
farost: We really struggle to come up with a good name. At this point, I got used to global causality number:

* global -- a part of a bigger domain, managed globally. Might be external, but it feels even more confusing
* causality -- provides specific ordering, does not guarantee gapless sequences. global causality highlights that it's different from seqnum
* number -- number

Happy to think more.

farost: We allow external users to modify this number. It can either be pub or have setters. I personally prefer setters, but this is shorter. What do you think?

storage/record.rs
```diff
@@ -0,0 +1,245 @@
...
+}
+
+#[derive(Serialize, Deserialize)]
+struct LegacyCommitRecordV1 {
```
farost, line 35: As we use bincode for serialization and deserialization, we need to explicitly support different encoding schemas. Since I moved records to a separate file, I just inserted a special Legacy record struct for the old version of records. We can use the same approach for any other automatic migration.

storage/record.rs
```diff
@@ -0,0 +1,245 @@
...
+    pub fn open_sequence_number(&self) -> SequenceNumber {
+        self.open_sequence_number
+    }
+
+    fn deserialise_from(record_type: DurabilityRecordType, reader: impl Read)
+    where
+        Self: Sized,
+    {
+        assert_eq!(Self::RECORD_TYPE, record_type);
+        // TODO: handle error with a better message
```
farost, line 110: Old todos, just copied the code from the original place

storage/record.rs
```diff
@@ -0,0 +1,245 @@
...
+            bincode::serialize(self).unwrap()
+        );
+        bincode::serialize_into(writer, &self)
+    }
+
+    fn deserialise_from(reader: &mut impl Read) -> bincode::Result<Self> {
+        // https://github.com/bincode-org/bincode/issues/633
+        let mut buf = Vec::new();
+        reader.read_to_end(&mut buf).map_err(|e| bincode::ErrorKind::Io(e))?;
+        match bincode::deserialize::<CommitRecord>(&buf) {
```
farost, line 194: Try deserializing as the new type, fallback to the legacy type otherwise

farost: Comments from @flyingsilverfin from the previous PR:

> 
> So it basically: bincode will error if it can't deserialize the record as the new type, which has more fields, since even that optional number will take up 8 bytes. If this error occurs, we can use the old record type - right?
> 
> If we implemented our own serialize/deserialize traits for Serde (instead of deriving it), we could do this transparently?
> 
> eg. if we try to deserialize the fields, and we run out of bytes to deserialize, we can just return the new format with the Option set to None?
> 
> That being said, i'm not entirely sure that's clearer. It should be equivalent and shorter, but this might be clearer. Not sure! Let's think about what would happen if we wanted to add another field in the future for example...
> 


My response:

> Yeah, it depends on the difference between the old and the new data type. If the number of bytes is the same, but the type is different, it may lead to incorrect data. We could handle it separately. Do we expect to do so, though? I don't think we're interested in changing the data types of the storage records just because we want to, it's still a risky migration process.

storage/snapshot/snapshot.rs
```diff
@@ -353,40 +354,49 @@ impl<D> ReadableSnapshot for ReadSnapshot<D> {
 pub struct WriteSnapshot<D> {
     operations: OperationsBuffer,
     open_sequence_number: SequenceNumber,
+    global_causality_number: CausalityNumber,
```
farost: We have to keep these causality numbers inside Schema/Write snapshots to generate commit records. We don't modify it, we just pass it from the external caller to the record.

storage/snapshot/snapshot.rs
```diff
@@ -504,7 +514,7 @@ impl<D: DurabilityClient> CommittableSnapshot<D> for WriteSnapshot<D> {
         let storage = self.storage.clone();
         match self.finalise(commit_profile)? {
             Some(commit_record) => match storage.clone().commit(commit_record, commit_profile) {
-                Ok(sequence_number) => Ok(Some(sequence_number)),
```
farost, line 507: This is important. See comment regarding optionality in storage.

storage/storage.rs
```diff
@@ -177,13 +179,13 @@ impl<Durability> MVCCStorage<Durability> {
         // guarantee external consistency: we always await the latest snapshots to finish
         let possible_sequence_number = self.isolation_manager.highest_validated_sequence_number();
         let open_sequence_number = self.wait_for_watermark(possible_sequence_number);
-        WriteSnapshot::new_with_open_sequence_number(self, open_sequence_number)
+        WriteSnapshot::new_with_open_sequence_number(self, open_sequence_number, CausalityNumber::None)
```
farost: By default, open without the causality number.
I am a little worried that we might open this code more, so we leave a way to accidentally break causality, but I'm not sure how to protect from it without a big refactor.

storage/storage.rs
```diff
@@ -220,14 +222,21 @@ impl<Durability> MVCCStorage<Durability> {
         &self,
         commit_record: CommitRecord,
         commit_profile: &mut CommitProfile,
-    ) -> Result<SequenceNumber, StorageCommitError>
+    ) -> Result<Option<SequenceNumber>, StorageCommitError>
```
farost: In all the users of this method, we used to write `Some(seqnum)`. There, `None` meant "nothing was done".
I make this an "official" method's signature. 
Cannot commit -> `Err(_)`.
Can commit, but no need -> `Ok(None)`
Committed -> `Ok(Some(_))`

Why? Because I don't want to search for the specific seqnum of an already applied record. `self.record_exists` below uses heuristics mentioned in the field's docs for optimization. 
Is it bad? Currently, I don't think so. It may become a problem if we ever want to really use the output `seqnum`. As far as I know, its always like
`match commit_result { Ok(_) => Ok(()), ... }`

storage/storage.rs
```diff
@@ -281,6 +290,23 @@ impl<Durability> MVCCStorage<Durability> {
...
+    fn record_exists(&self, commit_record: &CommitRecord) -> Result<bool, DurabilityClientError>
+    where
+        Durability: DurabilityClient,
+    {
+        let new_causality_number = commit_record.global_causality_number;
+        if new_causality_number.is_none() {
+            return Ok(false);
+        }
+        self.durability_client
+            .iter_sequenced_type_from::<CommitRecord>(self.durability_client.previous())?
```
farost: Just check the previous record and only the previous record.

WARNING: If the external user sends unordered records or decides to insert a record with None, everything will break. We have to ensure we're transparent with these rules.

storage/storage.rs
```diff
@@ -692,4 +721,97 @@ mod tests {
...
+            let existing =
+                commit_record_with_sequence_and_casuality(SequenceNumber::from(0), existing_causality, b"hello");
+            let mut profile = CommitProfile::DISABLED;
+            let seqnum = storage.commit(existing, &mut profile).unwrap();
+            debug_assert_eq!(seqnum, Some(SequenceNumber::from(1)), "Expected to be the first record in the WAL");
+            storage
+        }
+
+        let storage = storage_with_existing_commit(storage_path.to_path_buf(), init_casuality);
+        // TODO: If we use SequenceNumber::from(1) instead, it crashes
```
farost: I expect it to be `true` as we open after seqnum 1 is committed. 
The failure is:
```
thread 'tests::test_causality_number_after_none_is_not_ignoredtest isolation_manager::tests::test_highly_concurrent_correctness ... ' panicked at storage/isolation_manager.rs:602:9:
attempt to subtract with overflow
thread 'tests::test_causality_bigger_than_previous_is_not_ignored' panicked at storage/isolation_manager.rs:602:9:
attempt to subtract with overflow
```

So either I use it incorrectly or there is a bug. Maybe I don't understand the meaning of open seqnums, will explore it additionally.

This is something that I couldn't affect, so I bet that I just don't understand why it's wrong. But there might be another bug.

storage/snapshot/snapshot.rs
```diff
@@ -528,40 +543,49 @@ impl<D: DurabilityClient> CommittableSnapshot<D> for WriteSnapshot<D> {
...
+    fn new(
         storage: Arc<MVCCStorage<D>>,
         operations: OperationsBuffer,
         open_sequence_number: SequenceNumber,
-    ) -> impl ReadableSnapshot {
-        Self::new(storage, operations, open_sequence_number)
-    }
-
-    fn new(storage: Arc<MVCCStorage<D>>, operations: OperationsBuffer, open_sequence_number: SequenceNumber) -> Self {
+        global_causality_number: CausalityNumber,
```
flyingsilverfin: So I assume _not using_ causality numbers i also possible, but you can do something like `let number: CausalityNumber = Option::None` right? 

I think that indicates to me we sohuld have a constant `number::CAUSALITY_DISABLED` or `AUTO` - figuring out what this term is most effectively will help refine the naming of the whole concept IMO

storage/record.rs
flyingsilverfin: Continuing - surely we can't tell if it's an old versus a new type in any other way besides the length? it's impossible to have a different type that's the same length in the way we're building this right?

My thinking is that using a single struct that we manually implement serialize/deserialize makes it clear that we can always add new optional fields at the end and we can always interpet smaller records from old versions by setting those optional fields to None - will always work!

flyingsilverfin: I hope it makes sense - both work and I don't really mind - on disk they're identical.
Base assumption:
- we can add new fields to old records as add-only, and use `None` if the record is shorter. This will allow us to be backwards compatible

i'm not sure which one is less error prone, since the current way means maintaining a copy of the struct for every new extension, and we have to ensure that the new version is exactly [ previous version's fields + new field which must also be an Option anyway ].

storage/snapshot/snapshot.rs
flyingsilverfin, line 1: fyi "not_commitable()" is not a good name - it's more like "is_empty()"

In general name thing by describing what they _do_, not what they're intended to represent/be used as/

flyingsilverfin:
Continuing separately here in case my internet is flaky.

## Problem
Revisiting the main problem: we need to be able to tell if a commit is in the WAL/has been applied in Cluster. We want to do this by making writing to the WAL idempotent. 

In Core, this is a non-issue, since recovery will guarantee any commits written the WAL but not confirmed will simply be recovered, anything not in the WAL/corrupt will be truncated. 

In Cluster, we have an extended problem, since Core recovery will kick in, then we have to apply any commits in the raft log that aren't in the Core WAL. This requires us to check if a commit record we receive in raft is committed into Core or not. **Interesting note: this doesn't have to be super performant since it's going to happen as frequently as a crash, right? --- self edit: i guess this could happen in bulk when a node joins a cluster?**

Thinking sideways a bit - an equivalent operation in Core would be if _the user wants to check after a crash if their transaction committed!_ It could actually be a useful thing to be able to query IMO... So, how would we do that? 

## Did a transaction commit?

The most intuitive thing would be to generate some kind of a transaction ID, and then have a way to check if that transaction ID is committed. However, this feels "expensive" since we technically have to scan the entire WAL to search for a given transaction ID in our initial discussion.

BUT - what if we could make this quite performant? Soooooo here's my new idea - what if the transaction ID is exactly the `concat(OpenSequenceNumber, monotonically increasing number or even a UUID)`? Now, the operation to "check if a transaction committed" is a totally viable to do quite optimally - we have the methods to start iterating durability records from a particular sequence id (and this is quite optimal since we can skip to the relevant file in the WAL and start from there). If we build our transaction ID as I suggested, we can extract the starting sequence number from the transaction ID - making this easy! We just have to check the commit records from then on for a matching ID, if we find nothing - it doesn't exist!

The only way this is unperformant is if a ton of transactions open on the same sequence number and then all commit - you'd have to search all of the commit records from that start time, but that's ok since it's likely quite rare anyhow. Plus, if my assumption above is correct, for Cluster this doesn't even have to be a super optimal operation.

This is also unperformant if you want to check if an ancient Transaction ID  exists, since we have to go back quite far in the WAL. We'd also need to consider the case where we have deleted those records from the WAL since it's so ancient and we just have to error saying "Transaction too old, records not available".

## Current thoughts

Clearly, the current approach of making the WAL support idempotency and stuff with monotonically incrementing number is hella messy and complicated. 3 of us went around with so many approaches and we still ended up in a situation where it's not a safe API. Even worse, I don't see any obvious ways to make it safe without implementing extremely specific core code paths and extremely specific cluster code paths (feels ugly too).

I like the new idea since it basically makes it a problem we can actually push as a feature into Core entirely, and is a fully sound API. We can add the field to the CommitRecord the same way to be backwards compatible, but always set it (no need to disable it in Core). 

If possible, we should optimize this idea though: we don't need to check if the commit record exists in the WAL in the standard commit path, in core or cluster: that's an expensive operation since you're accessing the WAL, which takes out a lock, copies memory, and potentially reads files on disk. Maybe you can introduce a separate path for "idempotent_commit" which is used by cluster on replaying the raft logs?

Lemme know thoughts :D if this is completely impossible/stupid it's possible, lemme know since i came up with kinda tired on the plane :D

If it is sound and there are no obvious reasons not to do it this way -- my intuition tells me these hold, but needs validation -- then I think this is a more general, superior approach. Feel free to design it out between yourselves if you also believe that is true and we can look at it next week or something.

storage/record.rs
farost: Yeah, I see it. I like separating structs as it provides us with an explicit history, and you will not eventually replace it with something else. Writing manual serialization/deserialization methods, replacing the old ones, is a little scary. But is doable with testing. So I can agree with both approaches.

storage/snapshot/snapshot.rs
farost: Sounds good!

farost:
@flyingsilverfin I cannot reply to your comment directly, so I'm gonna share my thoughts here.

Yes, the thing I implemented definitely shines with holes and risks, and I'm not a fan of it.

Your idea sounds interesting and beneficial for both Core and Cluster. I agree. 
The main issue here is that splitting "loading" (recovery and joining to a cluster) and "normal" commit in Cluster is quite complicated, so optimizing this path is not a piece of cake. But, with effort, can be beneficial. 
Another part I worry about is that it's two separate API calls: `check` + `commit`. Cluster execution is an async operation. Without proper protection, it's theoretically possible to introduce a race condition, when we try to apply a commit record once, hang after the `check` stage, and then another commit record application attempt kicks in, goes through the whole `check` + `commit` pipeline, and then the first attempt resumes, submitting a duplicate, thinking that its still absent.

So... Currently, I like the design for Core, it does make sense. But utilizing this will require much effort on the Cluster side. Which is probably okay, but... Yeah, we need to work on it. @lolski

storage/uniqueness.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// Sequence number at which the transaction was opened.
+    /// This lets us efficiently seek into the WAL when checking idempotency.
+    open_sequence_number: SequenceNumber,
+
+    /// Per-open-sequence unique value (can be an external monotonic counter).
+    local_id: u64,
+}
+
+impl TransactionId {
+    pub fn new(open_sequence_number: SequenceNumber) -> Self {
```
farost: I introduced a new `TransactionId` struct. It is based on the open sequence number and a uniquely generated number, not required from the outside world.

The uniqueness is only guaranteed for a single process. But we don't generally open snapshots against old sequence numbers, right? Or are there cases when this is not right? 
But, if the process-scoped uniqueness is okay, we just need a static atomic counter is enough.

storage/storage.rs
farost: This is not needed anymore, but the usages are not really affected. I'll revisit this later, but maybe just leave as is.

storage/snapshot/snapshot.rs
```diff
@@ -352,14 +352,17 @@ impl<D> ReadableSnapshot for ReadSnapshot<D> {
 
 pub struct WriteSnapshot<D> {
     operations: OperationsBuffer,
-    open_sequence_number: SequenceNumber,
+    transaction_id: TransactionId,
```
farost: `TransactionId` consists of two fields:
* `open_sequence_number`
* `local_id`
I thought that storing a copy of the open seqnum does not really make sense.

Additionally, with this notion... I feel like it's actually a `SnapshotId`! Or something closer to snapshots, not transactions. Because we only check if a commit record is stored, and this commit record is based on snapshots, not transactions. Transactions just own the snapshots, but we can work even without transactions. I think it makes more sense. @flyingsilverfin ?

storage/storage.rs
```diff
@@ -56,9 +57,10 @@ pub mod iterator;
 pub mod key_range;
 pub mod key_value;
 pub mod keyspace;
+pub mod record;
 pub mod recovery;
-pub mod sequence_number;
 pub mod snapshot;
+pub mod uniqueness;
```
farost: Renamed `sequence_number` to `uniqueness` to store all id-related names in a single place.

storage/record.rs
farost: We'll need to return to this. Additionally, currently we have a duplication: `transaction_id` contains `open_sequence_number` and another counter. Question: "Do we want to always link transaction ids to open sequence numbers? Is it safe to just store only the transaction id instead? So `CommitRecord` will contain a `TransactionId` **instead of** the open seqnum, with the open seqnum inside of the Id"

What do you think? @flyingsilverfin

storage/snapshot/snapshot.rs
flyingsilverfin: Yup it's definitely a `snapshot_id`, we don't have `transaction` as a term in the storage layer at all :) good notice

flyingsilverfin: Then you'll notice the file should live under `snapshot/shapshot_id.rs`

storage/snapshot/snapshot.rs
```diff
@@ -376,25 +379,22 @@ impl<D> WriteSnapshot<D> {
...
+        }
     }
 }
 
 impl<D> ReadableSnapshot for WriteSnapshot<D> {
     const IMMUTABLE_SCHEMA: bool = true;
 
     fn open_sequence_number(&self) -> SequenceNumber {
-        self.open_sequence_number
+        self.transaction_id.open_sequence_number()
```
flyingsilverfin: I'd actually prefer if we kept these entirely separate. The fact that a Snapshot ID contains the sequence number is pure implementation detail/requirement, and if we don't have a strong reason to couple these together, we should not

storage/record.rs
```diff
@@ -0,0 +1,237 @@
...
+};
+
+#[derive(Serialize, Deserialize)]
+pub struct CommitRecord {
+    // TODO: this could read-through to the WAL if we have to save memory?
+    operations: OperationsBuffer,
+    open_sequence_number: SequenceNumber,
+    commit_type: CommitType,
+
+    /// Optional transaction identifier used for efficient referencing to commit records in the WAL,
```
flyingsilverfin: Not optional in the current version right? Or do we have to have it optional to make sure serialize/deserialize works?

storage/record.rs
flyingsilverfin: As I said before, I think we sohuld keep sequence number and snapshot id separate throughout the entire system. There's a world in which it makes sense for example to use a timestamp instead, and dont feel the need to couple them logically.

storage/uniqueness.rs
flyingsilverfin: we may in the future open transactions against old snapshots, but i dont think we need to optimise for _writes_ against ancient versions. These will be quite likely to fail validation against all the subsequent commits anyhow! But yeah, if that happens, the validation will be slow. 

However, I'm not sure it's avoidable anyhow. Happy to proceed.

my API would be:
```
SnapshotID {

  fn build(sequence_number: SequenceNumber) -> Self 
  }
}
```

storage/storage.rs
flyingsilverfin: Definitely undo - signatures are a really important part of what we signal is possible for this function to do! In fact, while reviewing I looked at the Option and spent 2 mins searching the code for where the option None is used... and it's not. Delete!

flyingsilverfin:
Summary:
- keep split between open seq nr and snapshot id
- push down `transaction_id.rs -> snapshot/snapshot_id.rs`
- restore `sequence_number.rs` module

plus other bits

storage/uniqueness.rs
```diff
@@ -0,0 +1,40 @@
...
+use std::sync::atomic::{AtomicU64, Ordering};
+
+use durability::DurabilitySequenceNumber;
+use serde::{Deserialize, Serialize};
+
+// TODO: when we introduce partitioning/remote WAL, we need create a separate type of the sequence number, instead of
+//       passing through to the Durability sequence number
+pub type SequenceNumber = DurabilitySequenceNumber;
+
+static TRANSACTION_LOCAL_ID_GEN: AtomicU64 = AtomicU64::new(1);
```
flyingsilverfin: I  don't really like local-id... these are meant to be basically globally unique IDs right? and the sequence number is just a thing to help link things.

flyingsilverfin:
Follow up idea:
in the Storage layer, we can have a separate `Snapshot ID` and a `Open Sequence Number`, right? And SnapshotID can just be a globally unique increment atomic u64 (for now, i dunno maybe in future we do something to allow generating it in a distributed way or something?)

Anyhow: in Storage we have both of those IDs. However, we can consider a `TransactionID` to be ` { sequence number, snapshot id }`.

This feels a bit cleaner/less redundant in Storage, but does it make the idempotent commit harder to do? I'm not sure it does - we just have to pass in both the open sequence number and the snapshot id from outside right?

Not sure if this is clear or works - lemme know

storage/uniqueness.rs
farost: I'm not worried about slow validation here. I'm worried about the fact that we may introduce collisions in the generated ids. 

Say, we had a process running and opening transactions against seqnum 5. 
* Seqnum 5 + id 1
* Seqnum 5 + id 2
* Seqnum 5 + id 3. 

Then, we commit and restart. If we had 3 commits, the next open seqnum is 8, and the new generated ids will start with 1 again, but it's not a problem:
* Seqnum 8 + Id 1 (different from seqnum 5 + id 1)
* ...

However, if we're able to open snapshots based on seqnums in the past, we can open a new snapshot with the same `Seqnum 5 + id 1`, introducing this collision.

Do we want to ensure that the generated id is always new? Or do we want to compare not only the id (while searching in the WAL to check if there was such a record), but also the content of the snapshot? That's what I'm worried about.

storage/uniqueness.rs
farost: changed to `UNIQUE_ID_GEN`

storage/record.rs
farost: Yes. I should've marked this in the description.

database/transaction.rs
```diff
@@ -89,6 +90,10 @@ impl<D: DurabilityClient> TransactionRead<D> {
     pub fn close(self) {
         drop(self)
     }
+
+    pub fn id(&self) -> TransactionId {
```
farost, line 94: `TransactionId`s are not needed for the storage. However, they are used as unique keys in the in-memory storage of `TransactionInfo` in `ServerState` to be able to close specific transactions on demand (e.g., when cluster leadership changes).

server/service/grpc/transaction_service.rs
```diff
@@ -137,6 +137,9 @@ pub(crate) struct TransactionService {
     query_queue: VecDeque<(Uuid, QueueOptions, typeql::query::Pipeline, String)>,
     query_responders: HashMap<Uuid, (JoinHandle<()>, QueryStreamTransmitter)>,
     running_write_query: Option<(Uuid, JoinHandle<(Transaction, WriteQueryResult)>)>,
+
+    close_sender: Sender<()>,
```
farost, line 140: Now we can send separate signals for closing a single transaction service. I haven't abstracted it away with anything smart, just an additional channel we track.

farost: In fact, we already can use it to close transactions when user credentials change. I'll add a todo for the future.

server/service/grpc/transaction_service.rs
```diff
@@ -164,19 +168,34 @@ impl TransactionService {
...
                         event!(Level::TRACE, "Shutdown signal received, closing transaction service.");
                         self.do_close().await;
                         return;
                     }
+                    recv_message = self.close_receiver.recv() => {
+                        match recv_message {
+                            Some(()) => {
+                                event!(Level::TRACE, "Transaction close signal received, closing transaction service.");
+                            }
+                            None => {
```
farost: `None` means the senders are dropped, which will not happen since we hold one instance inside `self`. However, this will be fair if we remove it from `self`, so let it be.

server/service/grpc/transaction_service.rs
```diff
@@ -408,13 +427,15 @@ impl TransactionService {
         let transaction_options = transaction_options_from_proto(open_req.options);
         let transaction_timeout_millis = transaction_options.transaction_timeout_millis;
 
-        let transaction_type = typedb_protocol::transaction::Type::try_from(open_req.r#type)
-            .map_err(|_| ProtocolError::UnrecognisedTransactionType { enum_variant: open_req.r#type }.into_status())?;
+        let transaction_type =
+            decode_transaction_type(typedb_protocol::transaction::Type::try_from(open_req.r#type).map_err(|_| {
```
farost, line 434: Just decode to the server-side type, which already existed, but was not needed for GRPC. Now we pass these types around the system and into Cluster.

server/service/grpc/transaction_service.rs
```diff
@@ -408,13 +427,15 @@ impl TransactionService {
...
+                ProtocolError::UnrecognisedTransactionType { enum_variant: open_req.r#type }.into_status()
+            })?);
 
         let database_name = open_req.database;
         let database = self
             .server_state
-            .databases_get_unrestricted(database_name.as_ref())
+            .databases_get_for_transaction(database_name.as_ref(), transaction_type)
```
farost, line 441: We need to serialize transaction opening, so the logic awaits all the other Raft log entries before returning the result to the user. We either implement `transaction_open` or `snapshot_open` in `ServerState` or have this method. 

I had a choice between:
* this (we just ask for the database, this method on the Cluster side checks if we are the primary replica for a write/schema tx or not, and errors if we're not allowed to open transactions)
* `transactions_open` / `snapshot_open`. However, this would make **all** replicas to execute this operation, and all replicas will open unused transactions. Pros: `schema` and `write` locks will be acquired everywhere, adding "extra" protection. Cons: it's really excessive.
* leave `database_get` but add another `transaction_open` which does nothing, but it will do two replication roundtrips, which is **too much**

Not sure how to design it differently now @lolski

server/service/http/transaction_service.rs
```diff
@@ -156,6 +163,9 @@ pub(crate) struct TransactionService {
     transaction: Option<Transaction>,
     query_queue: VecDeque<(TransactionResponder, QueueOptions, typeql::query::Pipeline, String)>,
     running_write_query: Option<(TransactionResponder, JoinHandle<(Transaction, WriteQueryResult)>)>,
+
+    close_sender: Sender<()>,
```
farost, line 164: Same as in `grpc`

server/state.rs
```diff
@@ -74,6 +94,12 @@ pub trait ServerState: Debug {
         name: &str,
     ) -> Result<Option<Arc<Database<WALClient>>>, ArcServerStateError>;
 
+    async fn databases_get_for_transaction(
```
farost, line 100: See `grpc/transaction_service.rs`

server/state.rs
```diff
@@ -98,6 +124,13 @@ pub trait ServerState: Debug {
         commit_profile: &mut CommitProfile,
     ) -> Result<(), ArcServerStateError>;
 
+    async fn database_commit_record_exists(
```
farost, line 132: Allows checking if the commit record already exists. Quite intrusive. When we refactor `ServerState`, we'll need to come up with a better solution... Maybe

server/state.rs
```diff
@@ -129,6 +162,15 @@ pub trait ServerState: Debug {
 
     async fn token_get_owner(&self, token: &str) -> Option<String>;
 
+    async fn transactions_add(
```
farost, line 170: We need to additionally validate that our Raft role hasn't changed while we were opening snapshots, and we can still add this transaction to the pool. So this will be a little overridden in Cluster with the needed checks

server/state.rs
```diff
@@ -129,6 +162,15 @@ pub trait ServerState: Debug {
...
+        &self,
+        transaction_id: TransactionId,
+        transaction_type: TransactionType,
+        close_sender: Sender<()>,
+    ) -> Result<(), ArcServerStateError>;
+
+    async fn transactions_close_types(&self, types: HashSet<TransactionType>) -> Result<(), ArcServerStateError>;
```
farost, line 177: This allows closing schema and write transactions from Cluster when its role changes

server/state.rs
```diff
@@ -193,6 +238,22 @@ impl LocalServerState {
             None
         };
 
+        let transactions = Arc::new(RwLock::new(HashMap::new()));
+        let controlled_transactions = transactions.clone();
+        background_task_spawner.spawn_interval(
```
farost, line 255: A cleanup job, similar to the `http/typedb_service.rs`'s one

server/service/grpc/transaction_service.rs
farost: P.S. `databases_get_unrestricted` was a bug, it should've been `databases_get`.

server/state.rs
```diff
@@ -658,3 +777,9 @@ impl ServerState for LocalServerState {
         self.background_task_spawner.clone()
     }
 }
+
+#[derive(Debug)]
+struct TransactionInfo {
```
farost, line 803: We can extend this with any info we have. E.g., the owner/accessor for this transaction, validaation other user operations within the same transaction (same as in HTTP, just in case?) and also closing the transactions when the user's credentials are updated.

storage/record.rs
farost: Sounds good, done that. But I still try deserializing the new format and then fallback to the old format for simplicity and "safety".

storage/snapshot/snapshot.rs
```diff
@@ -258,6 +261,7 @@ where
 
 pub struct ReadSnapshot<D> {
     open_sequence_number: SequenceNumber,
+    id: SnapshotId,
```
farost, line 264: Added so all snapshots have ids (then, we can create `TransactionId`s for any transaction type

storage/snapshot/snapshot.rs
```diff
@@ -351,15 +359,19 @@ impl<D> ReadableSnapshot for ReadSnapshot<D> {
 }
 
 pub struct WriteSnapshot<D> {
-    operations: OperationsBuffer,
     open_sequence_number: SequenceNumber,
+    id: SnapshotId,
```
farost, line 363: Snapshots are now "labeled" with the combination of `open_seq_num` + `id`, so I moved them to the top.

storage/snapshot/snapshot.rs
```diff
@@ -368,25 +380,29 @@ impl<D> WriteSnapshot<D> {
...
-    }
-
-    fn new(storage: Arc<MVCCStorage<D>>, operations: OperationsBuffer, open_sequence_number: SequenceNumber) -> Self {
         storage.isolation_manager.opened_for_read(open_sequence_number);
-        WriteSnapshot { storage, operations, open_sequence_number, iterator_pool: IteratorPool::new() }
+        WriteSnapshot {
+            storage,
+            operations,
+            open_sequence_number,
+            id: id.unwrap_or_else(|| SnapshotId::new()),
```
farost, line 403: New snapshots **always** generate new snapshot ids.

storage/record.rs
```diff
@@ -0,0 +1,243 @@
...
+    commit_type: CommitType,
+}
+
+impl From<LegacyCommitRecordV1> for CommitRecord {
+    fn from(legacy: LegacyCommitRecordV1) -> Self {
+        CommitRecord {
+            operations: legacy.operations,
+            open_sequence_number: legacy.open_sequence_number,
+            commit_type: legacy.commit_type,
+            snapshot_id: None,
```
farost, line 47: Important: it's not `CommitRecord::new()`, it's a direct struct creation, and `snapshot_id` is left `None`.

storage/snapshot/snapshot_id.rs
```diff
@@ -0,0 +1,26 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use std::sync::atomic::{AtomicU64, Ordering};
+
+use serde::{Deserialize, Serialize};
+
+static UNIQUE_ID_GEN: AtomicU64 = AtomicU64::new(1);
```
farost, line 13: We were forced to make all write transactions limited to **primary** replicas. So two replicas cannot commit at the same time == two replicas cannot commit two transactions opened at the same `sequence_number` with colliding ids. Why?

Case: two parallel commits
1. Primary replica 1 opens a transaction at seqnum S and id X (local value).
2. Another replica 2 becomes a primary and opens a transaction at seqnum S and id X (same value, used its own local generator)
3. If replica 1 (still thinking that it's primary) tries to commit, this commit will not get accepted by the majority since it's a replica of an older term. Then, any commit from replica 2 will not find collisions. Moreover, replica 1 closes its transaction as soon as it gets notified that its role has changed. We are safe.
4. If it doesn't commit, nothing is bad. Transactions are local, there are no collisions.

Case: sequential, unsynced id gens
1. Primary replica 1 opens a transaction at seqnum S and id X (local value).
2. Primary replica 1 commits the transaction.
3. Replica 2 becomes primary and opens a transaction with its local id X. Since transaction opening is ordered through the Raft log (now =) ), TypeDB waits until the commit from step 2 is applied. Then, this transaction's open seqnum will be higher than S. We are safe.
P.S. Same works if the commit fails -- open seqnum also counts failed commits (at least I tested and saw this once -- please tell me if there are caveats @flyingsilverfin )

Case: restarts of the same replica
1. Primary replica 1 opens a transaction at seqnum S and id X (local value).
2. Primary replica 1 restarts and becomes primary again, opening the same transaction (its ID is X or lower again). It's safe, because there are no commit records.
3. Primary replica 1 commits the transaction. 
4. Primary replica 1 restarts and becomes primary again. It opens the same transaction. Its ID is X or lower again. However, since we committed the transaction, it can only open at `open_seqnum` higher than the `open_seqnum` of the previous transaction. We are safe.


**This is safe as long as we allow only primary-bounded write/schema transactions and we don't allow opening transactions based on non-last seqnums**. 
**This is OK for alpha-1. This might NOT be OK for the official release, but we're ready to break storage between alpha versions. Since, we are ready to proceed if this covers 99.95% cases**

storage/record.rs
```diff
@@ -0,0 +1,243 @@
...
+    // TODO: this could read-through to the WAL if we have to save memory?
+    operations: OperationsBuffer,
+    open_sequence_number: SequenceNumber,
+    commit_type: CommitType,
+
+    /// Transaction identifier used for efficient referencing to commit records in the WAL,
+    /// avoiding excessive lookups through the whole filesystem.
+    /// Should be set for all new commit records, but is set to None for old loaded records.
+    #[serde(default)]
+    snapshot_id: Option<SnapshotId>,
```
farost, line 31: Left it at the bottom to highlight that it's an additional field introduced "later". Might be excessive, and we could reorder all the fields following the example of Snapshots: `open_seq_num`, `snapshot_id`, `operations`, `commit_type`... or maybe even `commit_type` before `operations`

farost, line 31: In the previous version of this PR, I used a field with `Option<NonZeroU64>`, which was optimized to 8 bytes. However, this optimization works only directly on `Option<NonZeroU64>` and will not work on `Option<struct { number: NonZeroU64 }>`.

I could introduce a conversion between a `SnapshotId` and `Option<NonZeroU64>` for this optimization. It could be worth it, but might be excessive considering the size of `operations` that we store.

storage/storage.rs
```diff
@@ -692,4 +721,109 @@ mod tests {
...
+            SnapshotId::new(),
+        );
+        let snapshot_id3 = commit_record3.snapshot_id().unwrap();
+        let seqnum3 = commit_record3.open_sequence_number();
+
+        assert!(storage.commit_record_exists(seqnum1, snapshot_id1).unwrap());
+        assert!(storage.commit_record_exists(seqnum2, snapshot_id2).unwrap());
+        assert!(!storage.commit_record_exists(seqnum3, snapshot_id3).unwrap());
+
+        // TODO: This fails because isolation_manager tries to decrement readers while it never
```
farost, line 798: I've put a todo for myself to revisit this, so it won't be lost. It's not something that I introduced, but it just works this way... Current tests are enough for the alpha.

database/database.rs
```diff
@@ -340,6 +339,16 @@ impl<D: DurabilityClient> Database<D> {
             Ok(())
         }
     }
+
+    pub fn commit_record_exists(
+        &self,
+        open_sequence_number: DurabilitySequenceNumber,
+        snapshot_id: SnapshotId,
+    ) -> Result<bool, DatabaseOpenError> {
```
flyingsilverfin, line 347: sensible API to have in Core, tbh!

flyingsilverfin: Or here you can have TransactionID actually?

flyingsilverfin: Then for `self.stoarge.commit_record_exists` you unwrap it into the two components

server/service/grpc/transaction_service.rs
flyingsilverfin: I'm going to leave this one to Ganesh!

flyingsilverfin: ideally not 2 round trips and no locks tho :D

server/state.rs
```diff
@@ -98,6 +123,13 @@ pub trait ServerState: Debug {
         commit_profile: &mut CommitProfile,
     ) -> Result<(), ArcServerStateError>;
 
+    async fn database_commit_record_exists(
+        &self,
+        name: &str,
+        open_sequence_number: DurabilitySequenceNumber,
+        snapshot_id: SnapshotId,
```
flyingsilverfin, line 136: I think packages that can be aware of TransactionID, should use TransactionID. The two part definition is a database --> storage interface.

server/state.rs
flyingsilverfin: interesting - what does this do in core? I'm ok w it - but @lolski should review here

server/state.rs
```diff
@@ -589,6 +678,8 @@ impl ServerState for LocalServerState {
             let commit_profile = transaction_profile.commit_profile();
             self.database_data_commit(SYSTEM_DB, commit_record, commit_profile).await?;
             self.token_manager.invalidate_user(username).await;
+            // TODO: Store users as owners of transactions in TransactionInfo and close transactions
```
flyingsilverfin: make sure we create this issue in Cluster or Here!

server/state.rs
flyingsilverfin: sounds good!

storage/record.rs
flyingsilverfin: bottom is good IMO

flyingsilverfin: simplicity wins if it's annoying at all in my book!

storage/snapshot/snapshot.rs
flyingsilverfin: technically it would be better if this is reflected in the signature, if you split it out into `new` and `new_with`, then you can have the `new` not even take a snapshot id - and it will always generate. Then the `new_with` can always accept a non-otpional one. Maybe `new_with` == `new_with_operations` or some variant, not sure. Just main idea applies

server/state.rs
farost: Should exist, I'll add the issue number.

database/database.rs
farost: We actually have the `CommitRecord`, where we retrieve the two fields in Cluster. We'd just pack it into `TransactionID` and then unpack here for the reasons of the API. We already expose `CommitRecord`s in the interfaces above (for database schema and data commits), so I thought it does not make sense to introduce these wraps.

storage/record.rs
farost: Then, are we good to go with this?

server/service/grpc/transaction_service.rs
farost: By "locks" I mean the schema database-wide lock and the write transaction counters.

storage/snapshot/snapshot_id.rs
flyingsilverfin: What do you mean "open seqnum also counts failed commits"? If a commit fails isolation checks (eg. after the initial local checks are done - after the record goes to the WAL), then a sequence number is used up - yep

flyingsilverfin (approved):
Approved w comments

server/state.rs
farost: I'll leave a TODO for the moment when we refactor `ServerState`, but in the current design, it's just a wrap + unwrap.

server/state.rs
farost: It saves the transaction info, so it's not redundant! It's actually useful in Core, too.
```
let mut transactions_lock = self.transactions.write().await;
transactions_lock.insert(transaction_id, TransactionInfo { transaction_type, close_sender });
```

storage/record.rs
farost: @lolski wdyt?

storage/snapshot/snapshot.rs
farost: So we currently have `new_with_commit_record` and `new_with_open_sequence_number` for the public interfaces. I had a natural desire to actually convert them to
`new` -> creates a new snapshot opened on a seqnum
`from_commit_record` -> just opens a previously existing snapshot again based on a commit record

Or something in this direction. @lolski what do you think?

storage/snapshot/snapshot_id.rs
farost: Yes, that's what I meant. Basically, I meant that if the record is in the WAL, regardless of its status, the next transaction will be opened at `open_seq_num` >= to this record's seqnum

database/database.rs
flyingsilverfin: ouuu wait wait what about if we call this `transaction_committed`?? Is that too loose? That's an excellent API to have in core. We can have commit record exists too tho to reach through.

storage/record.rs
flyingsilverfin: yes good for me

database/database.rs
farost: Sure, I can do this. I see what you want. I will just now wrap the fields from `CommitRecord` into `TransactionId` on the cluster side.

farost: Hm.... I don't actually like `transaction_committed` as this interface also recognises failed commits (commit_records with the rejected status). So I'll stick to the commit record call for now. Need to have a look while refactoring.

PR #7640, opened 2025-12-01
Generalize database import and clean up transactions

farost:
Move database import to `ServerState` to override/disable it in TypeDB Cluster. Clean up transaction-related utils and TODOs.

## Implementation
We want to temporarily disable database import in TypeDB Cluster because of the complications of all its stages' replication (e.g., the cleanup after an unsuccessful import attempt). To do so, we put the main task of the importer into a server state's method.

### DIFF
--- a/server/service/grpc/typedb_service.rs
+++ b/server/service/grpc/typedb_service.rs
@@ -470,7 +470,6 @@ impl typedb_protocol::type_db_server::TypeDb for GRPCTypeDBService {
         &self,
         request: Request<Streaming<typedb_protocol::database_manager::import::Client>>,
     ) -> Result<Response<Self::databases_importStream>, Status> {
-        println!("IMPORT!");
         let request_stream = request.into_inner();
         let (response_sender, response_receiver) = channel(IMPORT_RESPONSE_BUFFER_SIZE);
         let service = DatabaseImportService::new(
--- a/server/service/mod.rs
+++ b/server/service/mod.rs
@@ -5,13 +5,13 @@
  */
 
 use compiler::query_structure::{PipelineStructure, QueryStructureConjunctionID};
-pub use grpc::{IntoGrpcStatus, IntoProtocolErrorMessage};
+pub use grpc::{migration::import_service::DatabaseImportService, IntoGrpcStatus, IntoProtocolErrorMessage};
 use serde::{Deserialize, Serialize};
 
 pub(crate) mod export_service;
 pub(crate) mod grpc;
 pub mod http;
-pub(crate) mod import_service;
+pub mod import_service;
 mod transaction_service;
 
 #[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialOrd, PartialEq, Eq, Hash)]


server/service/grpc/typedb_service.rs
```diff
@@ -481,7 +480,17 @@ impl typedb_protocol::type_db_server::TypeDb for GRPCTypeDBService {
...
             self.server_state.shutdown_receiver().await,
         );
-        tokio::spawn(async move { service.listen().await });
+
+        run_with_diagnostics_async(
+            self.server_state.diagnostics_manager().await,
+            None::<&str>,
+            ActionKind::DatabasesImport,
+            || async {
+                self.server_state.databases_import(service).await.map_err(|err| err.into_error_message().into_status())
```
farost, line 488: We move `databases_import` to `ServerState` to simply return an error from TypeDB Cluster instead of a service job.

We also cover it with Diagnostics. Now, diagnostics for `DatabasesImport` operation will return:
* unsupported error from Cluster (this line)
* inner errors while processing the import (inside the service, as before).

server/service/http/transaction_service.rs
```diff
@@ -405,7 +405,7 @@ impl TransactionService {
                     LoadKind::WriteTransactions,
                 );
                 unwrap_or_execute_else_respond_error_and_return_break!(
-                    transaction.commit().1,
+                    transaction.commit().1, // TODO: Adapt to cluster
```
farost, line 408: What are we going to do with HTTP, anyway? Letting users to use the HTTP endpoint without replication is risky. I feel like it's not hard to actually implement everything (maybe without auto-redirection, but it also doesn't sound complicated). Wdyt?

server/state.rs
```diff
@@ -75,6 +78,8 @@ pub trait ServerState: Debug {
 
     async fn databases_create_unrestricted(&self, name: &str) -> Result<(), ArcServerStateError>;
 
+    async fn databases_import(&self, service: DatabaseImportService) -> Result<JoinHandle<()>, ArcServerStateError>;
```
farost, line 81: Receive the GRPC-based import service directly. I tried coming up with a more general trait processing, but depending on an async / boxed / arced dyn trait impl makes the whole `ServerState` trait less secure, and many errors around the codebase arise. This is just for simplicity of banning this operation now.

server/state.rs
```diff
@@ -429,6 +434,10 @@ impl ServerState for LocalServerState {
             .map_err(|err| arc_server_state_err(LocalServerStateError::DatabaseCannotBeCreated { typedb_source: err }))
     }
 
+    async fn databases_import(&self, service: DatabaseImportService) -> Result<JoinHandle<()>, ArcServerStateError> {
+        Ok(self.background_task_spawner.spawn(async move { service.listen().await }))
```
farost, line 438: Instead of just spawning a task, track it with the `background_task_spawner`. Now, Ctrl+C will cancel this job a little more gracefully.

system/util.rs
```diff
@@ -65,36 +65,6 @@ pub mod transaction_util {
             (profile, result)
         }
 
-        pub fn schema_transaction_commit<T>(
```
farost, line 68: Not used anymore. Safer to remove not to depend on the old `Transaction::commit` method.

PR #7645, opened 2025-12-10
Improve docker setup to use /var/lib/typedb/data for data dir

flyingsilverfin:

We improve the TypeDB Docker setup by using `/var/lib/typedb/data` for the architecture-agnostic data directory for TypeDB. This path is hardcoded into the built-in docker command for starting the command.

This means, we now simplify the docker external volume mount to be:

`docker volume create typedb-data` 
and
`docker create --name typedb -v typedb-data:/var/lib/typedb/data -p 1729:1729 -p 8000:8000 typedb/typedb:latest`

Which works for either ARM or x86 builds.

## Implementation

Use a different storage volume path for TypeDB Docker, and configure docker images that we build to use `/var/lib/typedb/data` as the storage directory.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -290,7 +290,7 @@ docker_container_image(
     operating_system = "linux",
     architecture = "amd64",
     base = "@typedb-ubuntu-x86_64//image",
-    cmd = ["/opt/typedb-server-linux-x86_64/typedb", "server"],
+    cmd = ["/opt/typedb-server-linux-x86_64/typedb", "server", "--storage.data-directory=/var/lib/typedb/data"],
     directory = "opt",
     env = {
         "LANG": "C.UTF-8",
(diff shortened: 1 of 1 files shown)


BUILD
```diff
@@ -290,7 +290,7 @@ docker_container_image(
     operating_system = "linux",
     architecture = "amd64",
     base = "@typedb-ubuntu-x86_64//image",
-    cmd = ["/opt/typedb-server-linux-x86_64/typedb", "server"],
+    cmd = ["/opt/typedb-server-linux-x86_64/typedb", "server", "--storage.data-directory=/var/lib/typedb/data"],
```
krishnangovindraj, line 293: Can you check if we can pass arguments to typedb? I think we may have to use 'entrypoint' as well:
```
entrypoint = ["/opt/typedb-server-linux-x86_64/typedb", "server"],
cmd = ["--storage.data-directory=/var/lib/typedb/data"],
```

flyingsilverfin:
Looks like we have a choice between entrypoint and cmd. The final executed thing when doing `docker run` or invoking it via `docker compose up` is "entrypoint + cmd". 

The `cmd` part can be overridden when invoking `run` - so if you put the binary and args in there, you can use the container and run something else.

For debugging I think this is useful for us (eg. run bash to poke around inside our prepared container) - so i'll leave it all in `cmd` for now!

PR #7647, opened 2025-12-11
Implement min and max expressions

flyingsilverfin:

Implement min and max for expressions, usable for numerical types (`integer`, `double`, and `decimal`). Works for exactly 2 arguments:

```
match 
let $x = min(10, 12);
let $y = max(10, 12);
```

## Implementation

Add new op codes and compilation for min and max in expressions, and new behaviour tests via git dependency.

Depends on https://github.com/typedb/typedb-behaviour/pull/394

### DIFF
--- a/compiler/annotation/expression/mod.rs
+++ b/compiler/annotation/expression/mod.rs
@@ -84,7 +84,7 @@ typedb_error! {
         ),
         UnsupportedDifferentArgumentForBuiltin(
             13,
-            "Built-in function '{function}' received one argument of type '{arg_1_category}' but another of type '{arg_2_category}'.",
+            "Built-in function '{function}' expects matching argument types, but received one argument of type '{arg_1_category}' but another of type '{arg_2_category}'.",
             function: BuiltInFunctionID,
             arg_1_category: ValueTypeCategory,
             arg_2_category: ValueTypeCategory,
--- a/tests/behaviour/debug/debug.feature
+++ b/tests/behaviour/debug/debug.feature
@@ -2,7 +2,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-Feature: Debugging Space
+#noinspection CucumberUndefinedStep
+Feature: TypeQL Query with Expressions
 
   Background: Open connection and create a simple extensible schema
     Given typedb starts
--- a/dependencies/typedb/repositories.bzl
+++ b/dependencies/typedb/repositories.bzl
@@ -35,6 +35,6 @@ def typedb_protocol():
 def typedb_behaviour():
     git_repository(
         name = "typedb_behaviour",
-        remote = "https://github.com/flyingsilverfin/typedb-behaviour",
-        commit = "b7fe1aa9d9e44201cc38129d0a696f422a18461b",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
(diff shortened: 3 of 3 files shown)


tests/behaviour/debug/debug.feature
krishnangovindraj, line 1: This has to be reverted

compiler/annotation/expression/expression_compiler.rs
```diff
@@ -535,6 +539,66 @@ impl<'this> ExpressionCompilationContext<'this> {
...
+                // Push arg2 back
+                self.type_stack.push(arg2_type);
+                // Both arguments must have the same type category
+                if arg1_category != arg2_category {
+                    return Err(Box::new(ExpressionCompileError::UnsupportedArgumentsForBuiltin {
+                        function: builtin.builtin_id(),
+                        category: arg1_category,
+                        source_span: builtin.source_span(),
+                    }));
+                }
```
krishnangovindraj: Misleading message

compiler/annotation/expression/expression_compiler.rs
```diff
@@ -535,6 +539,66 @@ impl<'this> ExpressionCompilationContext<'this> {
...
+                // Compile second argument first (it's on the stack first)
+                self.compile_recursive(self.expression_tree.get(builtin.argument_expression_ids()[1]))?;
+                // Compile first argument
+                self.compile_recursive(self.expression_tree.get(builtin.argument_expression_ids()[0]))?;
+                // Peek at both arguments without popping (arg2 is on top, arg1 is below)
+                let arg2_category = self.peek_type_single()?.category();
+                let arg2_type = self.pop_type()?;
+                let arg1_category = self.peek_type_single()?.category();
+                // Push arg2 back
+                self.type_stack.push(arg2_type);
```
krishnangovindraj: I don't think we need to pop, and peek, and push back.
We can just peek between the two compile_recursive calls and we'll get arg2_category

compiler/annotation/expression/expression_compiler.rs
```diff
@@ -535,6 +539,66 @@ impl<'this> ExpressionCompilationContext<'this> {
                     })?,
                 }
             }
+            BuiltInFunctionID::Min => {
+                // Compile second argument first (it's on the stack first)
```
krishnangovindraj: I don't know where this convention is from. 
The operators seem to compile the left expression and then the right.
And when they pop they always pop the rhs first. So this feels inconsistent with the others.

tests/behaviour/debug/debug.feature
flyingsilverfin: Keeping a bit now so it's easier to write tests into :D

compiler/annotation/expression/expression_compiler.rs
flyingsilverfin: yeah the stack is constructed and ends up in reverse, not in reverse in the expression tree

compiler/annotation/expression/mod.rs
```diff
@@ -82,13 +82,21 @@ typedb_error! {
             category: ValueTypeCategory,
             source_span: Option<Span>,
         ),
-        ListIndexMustBeInteger(
+        UnsupportedDifferentArgumentForBuiltin(
             13,
+            "Built-in function '{function}' received one argument of type '{arg_1_category}' but another of type '{arg_2_category}'.",
```
krishnangovindraj: Wording doesn't explicitly say why this is a problem.

flyingsilverfin: fixed

PR #7651, opened 2025-12-15
Reduce diagnostics noise from CI of TypeDB users

farost:

Add a requirement for `initial_delay` of diagnostics reports, which, if not met, forces the reporter to skip the first report cycle. This allows silencing CI jobs built by the users of TypeDB not using the `diagnostics.reporting.metrics` flag without significant harm to the data (the data will be added to the next report unless the server stops early).

Additionally, clean up the diagnostics logic from the outdated code from TypeDB 2.x not used in TypeDB 3.x.

## Implementation

* Update `reporter.rs`: add `REPORT_INTERVAL` to the calculated initial delay if it's too small
* Remove `is_owned` flag of the database metrics: all databases are "owned" by each replica of Cluster. In 3.x, the primary replica is selected for the whole TypeDB infrastructure, not per database. Thus, this flag is not needed.

### DIFF
--- a/diagnostics/metrics.rs
+++ b/diagnostics/metrics.rs
@@ -27,8 +27,6 @@ use crate::{
     DatabaseHash, DatabaseHashOpt,
 };
 
-const MONITORING_VERSION: usize = 1;
-
 #[derive(Serialize, Deserialize, Debug, Hash, Copy, Clone, PartialEq, Eq)]
 #[serde(rename_all = "lowercase")]
 pub enum ClientEndpoint {
--- a/diagnostics/reporter.rs
+++ b/diagnostics/reporter.rs
@@ -26,7 +26,7 @@ use hyper::{
 use hyper_rustls::{HttpsConnector, HttpsConnectorBuilder};
 use logger::{debug, trace};
 use resource::constants::{
-    common::{SECONDS_IN_HOUR, SECONDS_IN_MINUTE},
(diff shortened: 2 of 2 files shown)


diagnostics/metrics.rs
```diff
@@ -27,8 +27,6 @@ use crate::{
     DatabaseHash, DatabaseHashOpt,
 };
 
-const MONITORING_VERSION: usize = 1;
```
farost, line 30: `MONITORING_API_VERSION` is used for each separate type of report instead.

diagnostics/reporter.rs
```diff
@@ -226,11 +228,16 @@ impl Reporter {
...
-        let delay_secs = if current_minute > scheduled_minute {
+        let mut delay_secs = if current_minute > scheduled_minute {
             report_interval_secs - current_minute * SECONDS_IN_MINUTE + scheduled_minute * SECONDS_IN_MINUTE
         } else {
             (scheduled_minute - current_minute) * SECONDS_IN_MINUTE
         };
+
+        if delay_secs < REPORT_INTERVAL_MIN_DELAY.as_secs() {
+            // Skip this interval's report if the delay is too short to avoid CI reports
+            delay_secs += REPORT_INTERVAL.as_secs();
```
krishnangovindraj, line 239: Ah nice. Adding REPORT_INTERVAL keeps them all in sync for the next report.

PR #7654, opened 2025-12-16
Improve start up resilience

dmitrii-ubskii:

We reduce the possibility of crashes during startup after an operating system failure by doing the following:

- detect and discard corrupted database checkpoints on startup,
- ignore extraneous (non-database) directories in the data directory,
- separate checkpoints in progress from completed checkpoints.

### DIFF
--- a/database/database_manager.rs
+++ b/database/database_manager.rs
@@ -20,7 +20,6 @@ use crate::{database::DatabaseCreateError, Database, DatabaseDeleteError, Databa
 
 type DatabasesMap = HashMap<String, Arc<Database<WALClient>>>;
 type Databases = RwLock<DatabasesMap>;
-type DatabasesReadLock<'a> = RwLockReadGuard<'a, DatabasesMap>;
 type DatabasesWriteLock<'a> = RwLockWriteGuard<'a, DatabasesMap>;
 
 #[derive(Debug)]
@@ -275,7 +274,7 @@ impl DatabaseManager {
     }
 
     pub fn database_names(&self) -> Vec<String> {
-        self.databases.read().unwrap().keys().cloned().filter(|db| Self::is_user_database(db)).collect()
+        self.databases.read().unwrap().keys().filter(|&db| Self::is_user_database(db)).cloned().collect()
     }
 
     pub fn databases(&self) -> RwLockReadGuard<'_, HashMap<String, Arc<Database<WALClient>>>> {
@@ -332,7 +331,7 @@ impl DatabaseManager {
             .map_err(|source| DatabaseCreateError::DirectoryWrite { name: name.to_string(), source: Arc::new(source) })
     }
 
-    fn file_name_lossy(path: &PathBuf) -> String {
+    fn file_name_lossy(path: &Path) -> String {
         path.file_name().unwrap_or("".as_ref()).to_string_lossy().to_string()
     }
 
--- a/database/database_manager.rs
+++ b/database/database_manager.rs
@@ -73,7 +73,7 @@ impl DatabaseManager {
             }
 
             let database_name = entry_path.file_name().unwrap().to_string_lossy();
-            if Self::is_internal_database(&database_name) {
+            if Self::validate_user_database_name(&database_name).is_err() {
                 continue;
             }
 
@@ -133,7 +133,7 @@ impl DatabaseManager {
     }
 
     pub fn put_database(&self, name: impl AsRef<str>) -> Result<(), DatabaseCreateError> {
-        Self::validate_database_name(name.as_ref())?;
+        Self::validate_user_database_name(name.as_ref())?;
         self.put_database_unrestricted(name)
     }
 
@@ -184,7 +184,7 @@ impl DatabaseManager {
             })?;
         }
 
-        Self::validate_database_name(&name)?;
+        Self::validate_user_database_name(&name)?;
 
         let databases = self.databases.write().map_err(|_| DatabaseCreateError::WriteAccessDenied {})?;
         if self.exists_public(&databases, &name) {
@@ -336,7 +336,7 @@ impl DatabaseManager {
         path.file_name().unwrap_or("".as_ref()).to_string_lossy().to_string()
     }
 
-    fn validate_database_name(name: &str) -> Result<(), DatabaseCreateError> {
+    fn validate_user_database_name(name: &str) -> Result<(), DatabaseCreateError> {
         if Self::is_internal_database(name) {
             return Err(DatabaseCreateError::InternalDatabaseCreationProhibited {});
         }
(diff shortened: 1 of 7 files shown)


database/database_manager.rs
```diff
@@ -332,11 +338,11 @@ impl DatabaseManager {
             .map_err(|source| DatabaseCreateError::DirectoryWrite { name: name.to_string(), source: Arc::new(source) })
     }
 
-    fn file_name_lossy(path: &PathBuf) -> String {
+    fn file_name_lossy(path: &Path) -> String {
         path.file_name().unwrap_or("".as_ref()).to_string_lossy().to_string()
     }
 
-    fn validate_database_name(name: &str) -> Result<(), DatabaseCreateError> {
+    fn validate_user_database_name(name: &str) -> Result<(), DatabaseCreateError> {
```
dmitrii-ubskii, line 345: Clarifying; `_system` is a valid database name (indeed, we're using it), but not for the user.

storage/storage.rs
```diff
@@ -127,25 +127,24 @@ impl<Durability> MVCCStorage<Durability> {
         let storage_dir = path.join(Self::STORAGE_DIR_NAME);
 
         Self::register_durability_record_types(&mut durability_client);
-        let (keyspaces, next_sequence_number) = match checkpoint {
```
dmitrii-ubskii, line 130: `match` replaced with `if let`

.factory/automation.yml
```diff
@@ -170,6 +170,17 @@ build:
...
+    test-crash-resilience:
+      image: typedb-ubuntu-22.04
+      command: |
+        sudo apt update
+        sudo apt install -y libclang-dev
+        bazel run @typedb_dependencies//tool/bazelinstall:remote_cache_setup.sh
+        bazel build //:assemble-docker-x86_64.tar //:assemble-all-linux-x86_64-targz
+        tar -xf bazel-bin/typedb-all-linux-x86_64.tar.gz
+        docker load -i bazel-bin/assemble-docker-x86_64.tar
+        tool/test/simulate-crash.sh
```
flyingsilverfin: won't this still block release if its in the normal corectness workflow? i dunno if alex/ganesh remembers how to put it in a different one

storage/recovery/checkpoint.rs
```diff
@@ -169,13 +124,30 @@ impl Checkpoint {
...
-        let metadata_file_path = self.directory.join(Self::STORAGE_METADATA_FILE_NAME);
+        let metadata_file_path = self.directory.join(STORAGE_METADATA_FILE_NAME);
         let metadata = fs::read_to_string(metadata_file_path)
             .map_err(|error| MetadataRead { dir: self.directory.clone(), source: Arc::new(error) })?;
         Ok(SequenceNumber::new(
             metadata.parse().expect("Could not read METADATA file (could try to restore from previous checkpoint)"),
         ))
     }
+
+    fn is_consistent<KS: KeyspaceSet>(&self) -> io::Result<bool> {
```
flyingsilverfin: `is_valid`?

flyingsilverfin: or maybe is `is_complete`? i dont mind consistent tho

storage/recovery/checkpoint.rs
```diff
@@ -193,24 +165,114 @@ fn restore_storage_from_checkpoint(keyspace_dir: PathBuf, keyspace_checkpoint_di
...
     Ok(())
 }
 
-fn find_latest_checkpoint(checkpoint_dir: &Path) -> Result<Option<PathBuf>, CheckpointLoadError> {
-    if !checkpoint_dir.exists() {
-        return Ok(None);
+/// A checkpoint is a directory, which contains at least the storage checkpointing data: keyspaces + the watermark.
+/// The watermark represents a sequence number that is guaranteed to be in all the keyspaces, and after which we may
+/// have to reapply commits to the keyspaces from the WAL.
+pub struct CheckpointWriter {
```
flyingsilverfin, line 215: u wanna rename the other one to CheckpointReader?

tool/test/simulate-crash.sh
```diff
@@ -0,0 +1,20 @@
...
+START=$(date +%s)
+
+while true; do
+    END=$(date +%s)
+    ELAPSED=$((END-START))
+    if [[ $ELAPSED -ge $TIMEOUT ]]; then exit 0; fi
+
+    cid=$(docker run --detach bazel:assemble-linux-x86_64)
+    sleep 10 # give time to start up
+    typedb-all-linux-x86_64/typedb console --username=admin --password=password --address=localhost:1729 --tls-disabled --script=tests/assembly/script.tql
```
flyingsilverfin: can we hit the DB way harder? the likelihood of a write happening while killing it is kinda low no? Eg. a tql script that contains hundreds of sequential commits? With a variety of data, not just 1 put, which does nothing since in the second iteration it'll already contain the data and we write nothing (or maybe we're not being so clever, i forgot lol)

flyingsilverfin (changes requested):
Left a few comments

.factory/automation.yml
dmitrii-ubskii: Yep. Unfortunately, Factory hardcodes that the only two pipelines we can have are `build` and `release`, and as soon as any workflow in build fails, the whole pipeline fails.

storage/recovery/checkpoint.rs
dmitrii-ubskii: I like `is_complete`

.factory/automation.yml
flyingsilverfin: Well, more pointers to start moving off factory if poss

storage/recovery/checkpoint.rs
```diff
@@ -26,93 +26,48 @@ use crate::{
...
-            return Err(MissingStorageData { dir: self.directory.clone() });
-        }
-
-        let previous_checkpoints: Vec<_> = fs::read_dir(self.directory.parent().unwrap())
-            .and_then(|entries| {
-                entries
-                    .map_ok(|entry| entry.path())
-                    .filter(|path| path.is_ok() && path.as_ref().unwrap() != &self.directory)
-                    .try_collect()
+        fs::read_dir(&checkpoint_dir)
```
flyingsilverfin, line 49: shouldn't we sort by the sequence number or something? Imagine we have 5 checkpoints for any reason (idk, crashes during bootup?) or we just change the impl -- this will find the first valid one right? We have max 2 at a time rn?

dmitrii-ubskii: This finds the latest `complete` checkpoint dir. They are named with UNIX timestamps, and this `fold` will find the last one alphabetically.

We can have multiple checkpoints if the server crashes after a checkpoint is finalized but before the old ones are cleaned up, in which case the recovery will start from the newest checkpoint.

flyingsilverfin (approved):

PR #7663, opened 2025-12-24
Fix cases where snapshot would not close resources to prevent isolation manager memory leaks

krishnangovindraj:
Handles cases where a transaction would be closed without decrementing the reader count of the snapshot it was reading from. This caused memory leaks in the isolation manager.

## Implementation
Closes resources wherever I could find a leak without closing.

### DIFF
--- a/resource/constants.rs
+++ b/resource/constants.rs
@@ -127,7 +127,7 @@ pub mod snapshot {
 }
 
 pub mod storage {
-    pub const TIMELINE_WINDOW_SIZE: usize = 100;
+    pub const TIMELINE_WINDOW_SIZE: usize = 16;
     pub const WAL_SYNC_INTERVAL_MICROSECONDS: u64 = 1000;
     pub const WATERMARK_WAIT_INTERVAL_MICROSECONDS: u64 = 50;
     pub const COMMIT_WAIT_FOR_FSYNC: bool = true;


resource/constants.rs
```diff
@@ -127,7 +127,7 @@ pub mod snapshot {
 }
 
 pub mod storage {
-    pub const TIMELINE_WINDOW_SIZE: usize = 100;
+    pub const TIMELINE_WINDOW_SIZE: usize = 16;
```
krishnangovindraj, line 130: Higher contention on the lock might be bad for write performance. Can revert, but changed it to ask the question.

storage/isolation_manager.rs
```diff
@@ -104,8 +104,8 @@ impl IsolationManager {
         let window = self.timeline.get_or_create_window(sequence_number);
         window.insert_pending(sequence_number, commit_record);
         let CommitStatus::Pending(commit_record) = window.get_status(sequence_number) else { unreachable!() };
-        let isolation_conflict = self.validate_all_concurrent(sequence_number, &commit_record, durability_client)?;
-        if isolation_conflict.is_none() {
+        let validation_result = self.validate_all_concurrent(sequence_number, &commit_record, durability_client);
```
flyingsilverfin, line 107: Oof this is suuper subtle. That's actually why I went towards https://github.com/typedb/typedb/pull/7664/files - would you have a look at that idea?

PR #7664, opened 2025-12-26
Cleaner snapshot drop

flyingsilverfin:

We eliminate `SnapshotDropGuard`, which was only protecting us half the time, in favor of a direct `drop` implementation on Snapshot. Misuse of the drop guard could lead to difficult to diagnose and sporadic memory leaks.

## Implementation

Snapshots contain a ReadGuard that is provided by the Isolation Manager when opening a snapshot. Dropping the read guard automatically decrements a shared counter in the isolation manager, which in turn can then periodically clean up the historical commit data.

### DIFF
--- a/storage/isolation_manager.rs
+++ b/storage/isolation_manager.rs
@@ -473,7 +473,11 @@ impl Timeline {
         if sequence_number >= end {
             self.create_windows_to(sequence_number);
         }
-        self.try_get_window(sequence_number).unwrap()
+        if let Some(window) = self.try_get_window(sequence_number) {
+            window
+        } else {
+            panic!();
+        }
     }
 
     fn create_windows_to(&self, sequence_number: SequenceNumber) {
@@ -1050,7 +1054,7 @@ mod tests {
         let timeline = create_timeline();
         // windows are verified as unused and dropped when the watermark moves onwards far enough
 
-        for i in 0..TIMELINE_WINDOW_SIZE * 10 {
+        for i in 1..TIMELINE_WINDOW_SIZE * 10 {
             let tx = MockTransaction::new(&timeline, _seq(i as u64));
             tx_start_commit(&timeline, &tx);
             tx_finalise_commit_status(&timeline, tx, true);


storage/snapshot/snapshot.rs
```diff
@@ -588,7 +638,10 @@ impl<D> ReadableSnapshot for SchemaSnapshot<D> {
     }
 
     fn close_resources(&self) {
-        self.storage.closed_snapshot_write(self.open_sequence_number());
+        // swap ensures close_resources is only called once even if called explicitly and then via Drop
+        if self.cleanup_on_drop.swap(false, Ordering::SeqCst) {
+            self.storage.closed_snapshot_write(self.open_sequence_number());
+        }
```
krishnangovindraj: The function is private so it's easy to check if it's called elsewhere.

storage/snapshot/snapshot.rs
```diff
@@ -490,7 +525,22 @@ impl<D> SchemaSnapshot<D> {
...
+    }
+}
+
+impl<D> Drop for SchemaSnapshot<D> {
+    fn drop(&mut self) {
+        // swap ensures close_resources is only called once even if Drop runs after explicit close_resources call
+        if self.cleanup_on_drop.swap(false, Ordering::SeqCst) {
+            self.storage.closed_snapshot_write(self.open_sequence_number);
+        }
```
krishnangovindraj: Can we not call close_resources from drop?
Should we even have a close_resources, or should we just leave it to drop? Current usage seems to be only in cases where Snapshot drop would be called (the one manual use in master would be covered by snapshot drop)

storage/snapshot/snapshot.rs
```diff
@@ -615,75 +668,11 @@ impl<D: DurabilityClient> CommittableSnapshot<D> for SchemaSnapshot<D> {
...
-impl<S: ReadableSnapshot> Drop for SnapshotDropGuard<S> {
-    fn drop(&mut self) {
-        if let Some(inner) = self.inner.take() {
-            Self::unwrap_arc(inner).close_resources()
-        }
+    fn into_commit_record(mut self) -> CommitRecord {
+        // Disable cleanup on drop - the commit path handles cleanup manually
+        self.cleanup_on_drop.store(false, Ordering::SeqCst);
+        let operations = mem::take(&mut self.operations);
+        CommitRecord::new(operations, self.open_sequence_number, CommitType::Schema)
```
krishnangovindraj: We still can have the problem where the CommitRecord doesn't free the reader.
I think the real solution is to move the guard onto the snapshot number.

I think it is the SequenceNumber that should have a drop guard. e.g.:
`isolation_manager.record_reader() -> ReadSequenceDropGuard`
`isolation_manager.remove_reader(ReadSequenceDropGuard)`

storage/isolation_manager.rs
```diff
@@ -38,7 +38,7 @@ use crate::{
 #[derive(Debug)]
 pub(crate) struct IsolationManager {
     initial_sequence_number: SequenceNumber,
-    timeline: Timeline,
+    timeline: Arc<Timeline>,
```
flyingsilverfin: Required change to make the reader drop work - is conceptually ok

storage/isolation_manager.rs
```diff
@@ -440,9 +440,12 @@ impl Timeline {
         SequenceNumber::from(self.watermark.load(Ordering::SeqCst))
     }
 
-    fn record_reader(&self, sequence_number: SequenceNumber) {
+    fn record_reader(self: Arc<Self>, sequence_number: SequenceNumber) -> ReaderDropGuard {
         if let Some(window) = self.try_get_window(sequence_number) {
             window.increment_readers();
+            ReaderDropGuard { timeline: self.clone(), sequence_number }
+        } else {
+            unimplemented_feature!(ReopenOldSnapshotMVCC)
```
flyingsilverfin: We now just return a fully self contained drop guard.

Interestingly, this highlighted what would be a bug - opening a txn on an old MVCC snapshot that's not got a window in memory should be an error/needs to be dealt with.

storage/recovery/commit_recovery.rs
```diff
@@ -103,7 +103,7 @@ pub(crate) fn apply_recovered(
             }
             RecoveryCommitStatus::Rejected => isolation_manager.load_aborted(commit_sequence_number),
             RecoveryCommitStatus::Pending(commit_record) => {
-                isolation_manager.opened_for_read(commit_record.open_sequence_number());
+                let _ = isolation_manager.opened_for_read(commit_record.open_sequence_number());
```
flyingsilverfin: this would have been a memory leak too I think!

storage/snapshot/snapshot.rs
```diff
@@ -332,10 +345,27 @@ impl<D> WriteSnapshot<D> {
...
+            storage,
+            operations,
+            open_sequence_number,
+            iterator_pool: IteratorPool::new(),
+            reader_guard
+        }
     }
 }
 
+// impl<D> Drop for WriteSnapshot<D> {
```
flyingsilverfin: Not needed in this version, we just drop the guard and it handles the cleanup

storage/snapshot/snapshot.rs
```diff
@@ -456,8 +489,9 @@ impl<D: DurabilityClient> CommittableSnapshot<D> for WriteSnapshot<D> {
         }
     }
 
-    fn into_commit_record(self) -> CommitRecord {
-        CommitRecord::new(self.operations, self.open_sequence_number, CommitType::Data)
+    fn into_commit_record(mut self) -> (ReaderDropGuard, CommitRecord) {
```
flyingsilverfin: return the guard and the record so the caller can decide when to drop the reader handle (not the cleanest, but works)

system/util.rs
```diff
@@ -90,7 +91,7 @@ pub mod transaction_util {
                 profile,
             } = TransactionWrite::open(self.database.clone(), TransactionOptions::default()).unwrap();
             let (rows, snapshot) = fn_(
-                snapshot.into_inner(),
+                Arc::try_unwrap(snapshot).unwrap_or_else(|_| panic!("Expected unique ownership of snapshot")),
```
flyingsilverfin, line 94: the into_inner was a nice shorthand before

tests/behaviour/steps/concept/thing/attribute.rs
```diff
@@ -30,7 +32,7 @@ pub fn attribute_put_instance_with_value_impl(
         let value = value.into_typedb(
             attribute_type.get_value_type_without_source(tx.snapshot.as_ref(), &tx.type_manager).unwrap().unwrap(),
         );
-        tx.thing_manager.create_attribute(tx.snapshot.as_mut().unwrap(), attribute_type, value)
+        tx.thing_manager.create_attribute(Arc::get_mut(&mut tx.snapshot).unwrap(), attribute_type, value)
```
krishnangovindraj, line 35: Can we have  `Transaction::unwrap_snapshot_mut` ?

storage/snapshot/snapshot.rs
krishnangovindraj: I would have liked for CommitRecord to hold the `reader_guard`, but I've realised that can't be deserialized properly.

storage/recovery/commit_recovery.rs
krishnangovindraj: There's a `remove_reader` in `validate_commit`. Though in the rare case where `validate_commit` returns an error (instead of an `Ok(IsolationConflict)`), we do escape from it (using `?`)  before the call. 
That's the main reason I wanted to move towards having a drop guard on the reader.

On that note, 
1. I think `validate_commit` should take the ReadGuard in.
2. We might have a double decrement problem because of the  `remove_reader` in `validate_commit`

storage/isolation_manager.rs
```diff
@@ -518,6 +521,18 @@ impl Timeline {
...
+    timeline: Arc<Timeline>,
+    sequence_number: SequenceNumber,
+}
+
+impl Drop for ReaderDropGuard {
+    fn drop(&mut self) {
+        self.timeline.remove_reader(self.sequence_number)
+    }
+}
+
```
krishnangovindraj, line 526: I'm wondering whether to complicate our code like this.
I'd consider:
1. remove the `Arc<Timeline>` altogether, 
2. Make drop log an error +  `debug_assert(false)`.
3. Make `remove_reader()` accept the `ReadGuard`,
4. resort to `mem::forget` in `remove_reader()` to prevent the logging.

krishnangovindraj: Looks like this is dangerous for cases where we panic, but aren't those fatal anyway?

flyingsilverfin: wait is this in the wrong place i dont follow the comments. Complicate what?

flyingsilverfin: Here's an idea - we can probably just hold an Arc to the Window, which is already Arc'ed. Then we can check the strong count of the Arc == 1 to free the windows. I think this will work - other readers will also hold Arcs, which must be dropped in order for the window to be freed.

Also, we always free windows from the beginning. It's possible a concurrent validation happens which picks up the relevant window for validation - but that means we can't delete it anyway since we only picked up the window since it's required to validate from the start of the transaction to the commit.

Basically I'm thinking we're re-implementing Arc counters to drop stuff - let's just use that mechanism, and it keeps Timeline's ownership clean.

krishnangovindraj: Sorry, that was indeed in the wrong place. Or I deleted the relevant part of the comment.

The complicated macro was for a compile-time detection of `drop` being called for a type. It relied on optimisations inlining stuff to the point where it breaks.

krishnangovindraj: Just using Arcs instead of doing our own reference count does makes sense.

storage/isolation_manager.rs
```diff
@@ -518,6 +508,18 @@ impl Timeline {
...
 }
 
+pub struct ReaderDropGuard {
+    window: Option<Arc<TimelineWindow<TIMELINE_WINDOW_SIZE>>>,
+}
+
+impl Drop for ReaderDropGuard {
+    fn drop(&mut self) {
+        if let Some(window) = self.window.as_ref() {
+            window.decrement_readers();
```
flyingsilverfin, line 522: will get cleaned up later if no more readers

storage/isolation_manager.rs
```diff
@@ -887,32 +890,28 @@ mod tests {
     struct MockTransaction {
         read_sequence_number: SequenceNumber,
         commit_sequence_number: SequenceNumber,
+        reader_drop_guard: ReaderDropGuard,
     }
 
     impl MockTransaction {
-        fn new(read_sequence_number: SequenceNumber, commit_sequence_number: SequenceNumber) -> MockTransaction {
-            MockTransaction { read_sequence_number, commit_sequence_number }
+        fn new(timeline: &Timeline, commit_sequence_number: SequenceNumber) -> MockTransaction {
```
flyingsilverfin, line 901: slight refactor of tests

storage/isolation_manager.rs
```diff
@@ -1053,6 +1041,21 @@ mod tests {
         let (timeline, _) = &*timeline_and_counter;
         assert_eq!(expected_watermark, timeline.watermark());
         let some_index_in_penultimate_window = expected_watermark - TIMELINE_WINDOW_SIZE - 1;
+        timeline.may_free_windows();
         assert_true!(timeline.try_get_window(some_index_in_penultimate_window).is_none());
     }
+
+    #[test]
+    fn windows_are_dropped() {
```
flyingsilverfin, line 1053: new test

storage/recovery/commit_recovery.rs
```diff
@@ -103,10 +103,11 @@ pub(crate) fn apply_recovered(
             }
             RecoveryCommitStatus::Rejected => isolation_manager.load_aborted(commit_sequence_number),
             RecoveryCommitStatus::Pending(commit_record) => {
-                isolation_manager.opened_for_read(commit_record.open_sequence_number());
+                let read_guard = isolation_manager.opened_for_read(commit_record.open_sequence_number());
                 let validated_commit = isolation_manager
                     .validate_commit(commit_sequence_number, commit_record, durability_client)
                     .map_err(|error| DurabilityClientRead { typedb_source: error })?;
+                drop(read_guard);
```
flyingsilverfin, line 110: not sure if the drop has to be after, but leaving it for now

resource/constants.rs
```diff
@@ -127,7 +127,7 @@ pub mod snapshot {
 }
 
 pub mod storage {
-    pub const TIMELINE_WINDOW_SIZE: usize = 100;
+    pub const TIMELINE_WINDOW_SIZE: usize = 32;
```
flyingsilverfin, line 130: Reduced window size to avoid memory presure

storage/isolation_manager.rs
krishnangovindraj: For when I review this again, "later" is when we increment the watermark in `may_increment_watermark`

tests/behaviour/steps/concept/thing/attribute.rs
```diff
@@ -4,6 +4,8 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+use std::sync::Arc;
```
krishnangovindraj, line 7: I think all the `Arc` imports in this directory are unused?

storage/recovery/commit_recovery.rs
krishnangovindraj: I think that's the right place. Once we've validated, we don't need it pinned anymore

tests/behaviour/steps/concept/thing/attribute.rs
flyingsilverfin: i had to undo this change so they're now used

tests/behaviour/steps/concept/thing/attribute.rs
```diff
@@ -148,7 +150,7 @@ async fn delete_attribute(context: &mut Context, var: params::Var) {
         context.attributes[&var.name]
             .clone()
             .unwrap()
-            .delete(tx.snapshot.as_mut().unwrap(), &tx.thing_manager, StorageCounters::DISABLED)
+            .delete(Arc::get_mut(&mut tx.snapshot).unwrap(), &tx.thing_manager, StorageCounters::DISABLED)
```
flyingsilverfin, line 153: Basically i can't put a helper on `tx` in these types of places where we both need to get a mutable `tx.snapshot`, and a reference to something else in `tx` - rust can't figure out that if you do `tx.get_snapshot_mut()` which takes `&mut self` and also use `&tx.thing_manager` that this is two different parts of memory. sad! Let's just keep this as is for now @krishnangovindraj

PR #7668, opened 2026-01-06
Annotations can use losslessy castable arguments

flyingsilverfin:

We fix two panics that arose when using castable values (eg. integers range annotations for decimal value types) in parameters to annotations, such as a `decimal` that used `integer` arguments: `person owns salary @values(10)` or `person owns salary @range(0..1000000);`.

We make consistent the behaviour that:

- Integers are losslessly castable to Decimal and Double value types
- Date is lossless castable to DateTime

We have also updated BDD tests to reflect that doubles are no longer losslessly castable to decimals, since they have different precision.

## Implementation

* AnnotationRange relies on existing Value::partial_cmp instead of having a custom unwrap() mechanism.

### DIFF
--- a/dependencies/typedb/repositories.bzl
+++ b/dependencies/typedb/repositories.bzl
@@ -35,6 +35,6 @@ def typedb_protocol():
 def typedb_behaviour():
     git_repository(
         name = "typedb_behaviour",
-        remote = "https://github.com/flyingsilverfin/typedb-behaviour",
-        commit = "5eb88a9a2124e2fea33785a0e97bc6c59fdc875a",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        remote = "https://github.com/typedb/typedb-behaviour",
+        commit = "7ced0696295bcd83c8b59f6c6b0e0ec62cc13989",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
     )


concept/type_/annotation.rs
```diff
@@ -313,41 +313,52 @@ impl AnnotationRange {
...
-                        Value::Integer(start_inclusive) => start_inclusive < &end_inclusive.clone().unwrap_integer(),
-                        Value::Double(start_inclusive) => start_inclusive < &end_inclusive.clone().unwrap_double(),
-                        Value::Decimal(start_inclusive) => start_inclusive < &end_inclusive.clone().unwrap_decimal(),
-                        Value::Date(start_inclusive) => start_inclusive < &end_inclusive.clone().unwrap_date(),
-                        Value::DateTime(start_inclusive) => start_inclusive < &end_inclusive.clone().unwrap_date_time(),
-                        Value::DateTimeTZ(start_inclusive) => {
-                            start_inclusive < &end_inclusive.clone().unwrap_date_time_tz()
-                        }
-                        Value::String(start_inclusive) => start_inclusive < &end_inclusive.clone().unwrap_string(),
-                        Value::Duration(_) => unreachable!("Cannot use duration for AnnotationRange"),
```
flyingsilverfin, line 348: If this has been caught before, we can just rely on partial_cmp to do the required casting rather than having `unwrap_string` fail now that we can use `@range(0.0dec...10)` using decimal and integer mixed value types

tests/behaviour/steps/params/lib.rs
```diff
@@ -565,69 +566,81 @@ impl Value {
     }
 
     pub fn into_typedb(self, value_type: TypeDBValueType) -> TypeDBValue<'static> {
-        match value_type {
```
flyingsilverfin, line 568: Don't handroll parsing, using TypeQL

tests/behaviour/service/http/driver/concept.rs
krishnangovindraj, line 1: lol, is this you or the AI?

tests/behaviour/steps/params/lib.rs
```diff
@@ -565,68 +563,20 @@ impl Value {
...
-                let value = if self.raw_value.starts_with('"') && self.raw_value.ends_with('"') {
-                    &self.raw_value[1..&self.raw_value.len() - 1]
-                } else {
-                    self.raw_value.as_str()
-                };
-                TypeDBValue::String(Cow::Owned(value.to_string()))
-            }
-            // TODO: Could compare string representations like in driver
-            TypeDBValueType::Struct(_) => todo!(),
+        if value_type == TypeDBValueType::String {
```
krishnangovindraj: Why does string have to be handled separately?

tests/behaviour/service/http/driver/concept.rs
flyingsilverfin: my hacks lol

tests/behaviour/steps/params/lib.rs
flyingsilverfin: It's basically because we parse our BDD rows into `serde::JSON` values, which removes quoting - so basically i just bypass all quoting and throw it straight into our String.
That being said it might have an issue with escape sequences if we don't go through our parser! So ill do the opposite and add back quotes, then throw it through the parser.

tests/behaviour/steps/params/lib.rs
```diff
@@ -565,91 +563,23 @@ impl Value {
...
-        assert!(Self::FRACTIONAL_ZEROES >= value.len());
-        10_u64.pow((Self::FRACTIONAL_ZEROES - value.len()) as u32) * value.trim().parse::<u64>().unwrap()
-    }
-
-    fn parse_date_time_and_remainder(value: &str) -> (NaiveDateTime, &str) {
-        for format in Self::DATETIME_FORMATS {
-            if let Ok((datetime, remainder)) = NaiveDateTime::parse_and_remainder(value, format) {
-                return (datetime, remainder.trim());
+        let value = if value_type == TypeDBValueType::String {
+            // we are either given an in-memory string that is already unquoted and un-escaped
```
flyingsilverfin, line 567: We have this because we use the same code to go from the HTTP JSON response from TypeDB for strings, such as `John "Baba Yaga" Wick`, as well as converting straight values from the BDD steps such as `"John \"Baba Yaga \" Wick"`

concept/type_/annotation.rs
```diff
@@ -313,41 +313,37 @@ impl AnnotationRange {
...
-                        Value::Struct(_) => unreachable!("Cannot use structs for AnnotationRange"),
-                    }
+                    start_inclusive.partial_cmp(end_inclusive) == Some(Ordering::Less)
```
krishnangovindraj, line 346: Does this rule out `@range(2..2)`? (i.e. start == end)
Or is that not allowed/

flyingsilverfin: Yeah it's not legal - same as before. I also thought it was a bit odd but I think it's also ok

PR #7670, opened 2026-01-07
Convert CacheError to use typedb_error macro

flyingsilverfin:

Replace manual CacheError enum with typedb_error! macro invocation, cleaning up old, un-macroed errors.

## Implementation

- Add //common/error dependency to BUILD file
- Update DatabaseImportError to use typedb_source instead of source for CacheError since it's now a TypeDB error type


flyingsilverfin:
Will resume this in the future, was a little test!

PR #7678, opened 2026-01-19
Implement built-in functions length(), iid(), and label()

dmitrii-ubskii:

We implement the 

- `len()`: string length in codepoints / unicode scalar values (USVs): `len("TypeDB") == 6`, `len("こんにちは") == 5`, `len("⭐") == 1`, `len("❤️‍🔥") == 4`;
- `iid()`: an instance's internal ID as a string, to be used in an `iid` constraint in following queries;
- `label()`: a type's label as a string, for convenience in fetch queries.

## Implementation

`length()` is implemented alongside the existing arithmetic built-ins. We also update `NativeValueConvertible` to have a lifetime so that we can extract the `Cow<'_, str>` from the string value without a forced `.into_owned()`. Although currently expression evaluation is performed over `Value<'static>`, so we don't benefit from that at the moment.

Because the current implementation of expression evaluation assumes that every variable and intermediate result is (or can be cast to) a value, and because `iid()` and `label()` operate over arbitrary concepts, they are handled in the same code path as user-defined functions in IR and in the compiler. During execution, they are evaluated as immeediate executors to keep their impact low, as they do not require tabling.

### DIFF
--- a/dependencies/typedb/repositories.bzl
+++ b/dependencies/typedb/repositories.bzl
@@ -22,7 +22,7 @@ def typeql():
     git_repository(
         name = "typeql",
         remote = "https://github.com/typedb/typeql",
-        commit = "9f00091ebd4d925dbc01d9a182a001099f31230b",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_driver
+        commit = "ff46a9b50b021cb8c144998ef2c2e0ee84794363",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_driver
     )
 
 def typedb_protocol():
@@ -36,5 +36,5 @@ def typedb_behaviour():
     git_repository(
         name = "typedb_behaviour",
         remote = "https://github.com/typedb/typedb-behaviour",
-        commit = "bb0fabacf22256c070d070a8a7515177847e3c4d",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        commit = "cf13e741844025a8a0210f25dc295ce966e26d16",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
     )


compiler/annotation/expression/instructions/mod.rs
```diff
@@ -48,5 +48,6 @@ typedb_error! {
         ListIndexNegative(5, "List index is negative: {index}", index: i64),
         ListIndexOutOfRange(6, "List index out of range {index}, list length: {length}", index: i64, length: usize),
         ListRangeOutOfRange(7, "List range out of range {from_index}..{to_index}, list length: {length}", from_index: i64, to_index: i64, length: usize),
+        OverlongString(8, "Found string with length {len} which is too long to fit in a 64-bit signed integer", len: usize),
```
dmitrii-ubskii, line 51: Good luck triggering this.

encoding/value/value.rs
```diff
@@ -483,18 +482,18 @@ impl fmt::Display for Value<'_> {
     }
 }
 
-pub trait NativeValueConvertible: Sized {
+pub trait NativeValueConvertible<'a>: Sized {
     const VALUE_TYPE_CATEGORY: ValueTypeCategory;
```
dmitrii-ubskii, line 486: Needed to allow us to operate over `Cow<'_, str>` in expressions. Currently all values in expression evaluation are `Value<'static>`, so this change doesn't affect much except lay the groundwork for reducing copies in the future.

encoding/graph/thing/vertex_attribute.rs
```diff
@@ -50,7 +50,8 @@ pub struct AttributeVertex {
 
 impl AttributeVertex {
     pub const PREFIX: Prefix = Prefix::VertexAttribute;
-    pub const MAX_LENGTH: usize = PrefixID::LENGTH + TypeID::LENGTH + ValueEncodingLength::LONG_LENGTH;
+    pub const MAX_LENGTH: usize =
+        PrefixID::LENGTH + TypeID::LENGTH + ValueTypeBytes::CATEGORY_LENGTH + ValueEncodingLength::LONG_LENGTH;
```
dmitrii-ubskii, line 54: `ValueTypeBytes` was unaccounted for!

compiler/annotation/expression/compiled_expression.rs
```diff
@@ -58,7 +59,6 @@ impl<ID: IrID> ExecutableExpression<ID> {
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
 pub enum ExpressionValueType {
     // TODO: we haven't implemented ConceptList, only ValueList right now.
-    // TODO: this should hold an actual ValueType, not a Category!
```
dmitrii-ubskii, line 61: Outdated?

common/bytes/bytes.rs
```diff
@@ -116,6 +116,16 @@ impl<const ARRAY_INLINE_SIZE: usize> fmt::Display for Bytes<'_, ARRAY_INLINE_SIZ
     }
 }
 
+impl<const ARRAY_INLINE_SIZE: usize> fmt::LowerHex for Bytes<'_, ARRAY_INLINE_SIZE> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("0x")?;
```
dmitrii-ubskii, line 121: This could be pushed into an `if f.alternate()` branch as for `u64` e.g., if need be.

compiler/annotation/function.rs
```diff
@@ -263,6 +277,30 @@ pub(super) fn annotate_named_function(
...
+                        source_span: label.span(),
+                        typedb_source: TypeInferenceError::ConceptRead { typedb_source },
+                    }
+                })?;
+
+                let concept_types = Iterator::chain(
+                    object_types.into_iter().map(Type::from),
+                    attribute_types.into_iter().map(Type::from),
+                )
+                .collect();
```
dmitrii-ubskii, line 300: Annoyingly, we can't push that to `TypeManager` because `answer` (which houses `answer::Type`) depends on `concept` and not the other way around.

dependencies/typedb/repositories.bzl
```diff
@@ -36,5 +36,5 @@ def typedb_behaviour():
     git_repository(
         name = "typedb_behaviour",
         remote = "https://github.com/typedb/typedb-behaviour",
-        commit = "7ced0696295bcd83c8b59f6c6b0e0ec62cc13989",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        commit = "cf5c677951ecb23d6826e7ed23f93fbd87a017f2",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
```
dmitrii-ubskii: https://github.com/typedb/typedb-behaviour/pull/398

dependencies/typedb/repositories.bzl
```diff
@@ -22,7 +22,7 @@ def typeql():
     git_repository(
         name = "typeql",
         remote = "https://github.com/typedb/typeql",
-        commit = "3011a31ae1fd16f4d387bb801dec8b385cebd954",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_driver
+        commit = "d33dc42c37b49bb763463e69c642b37bb20d8e46",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_driver
```
dmitrii-ubskii: https://github.com/typedb/typeql/pull/427

ir/pattern/expression.rs
```diff
@@ -107,9 +108,9 @@ impl<ID: StructuralEquality> StructuralEquality for ExpressionTree<ID> {
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub enum Expression<ID> {
     Constant(ParameterID),
-    Variable(ID),
+    Variable(ID), // User-defined functions are re-written as an anonymous assignment.
     Operation(Operation),
-    BuiltInCall(BuiltInCall), // Other functions must be re-written as an anonymous assignment.
+    BuiltinValueFunctionCall(BuiltinValueFunctionCall),
```
dmitrii-ubskii, line 113: Renamed: `BuiltinValueFoo` vs `BuiltinConceptFoo`. The former are handled with expressions, the latter — with user-defined functions.

ir/pattern/expression.rs
```diff
@@ -248,46 +252,103 @@ impl BuiltInCall {
...
+            BuiltinValueFunctionID::Abs => fmt::Display::fmt(&typeql::token::Function::Abs, f),
+            BuiltinValueFunctionID::Ceil => fmt::Display::fmt(&typeql::token::Function::Ceil, f),
+            BuiltinValueFunctionID::Floor => fmt::Display::fmt(&typeql::token::Function::Floor, f),
+            BuiltinValueFunctionID::Round => fmt::Display::fmt(&typeql::token::Function::Round, f),
+            BuiltinValueFunctionID::Max => fmt::Display::fmt(&typeql::token::Function::Max, f),
+            BuiltinValueFunctionID::Min => fmt::Display::fmt(&typeql::token::Function::Min, f),
+            BuiltinValueFunctionID::Length => fmt::Display::fmt(&typeql::token::Function::Length, f),
+        }
+    }
+}
```
dmitrii-ubskii, line 317: Git is confused by the rename, this is not new.

executor/read/control_instruction.rs
```diff
@@ -26,6 +26,8 @@ pub(super) enum ControlInstruction {
 
     ExecuteImmediate(ExecuteImmediate),
 
+    ExecuteBuiltinCall(ExecuteBuiltinCall),
```
dmitrii-ubskii: This may be included with `Immediate`, TBD.

compiler/annotation/expression/instructions/unary.rs
```diff
@@ -91,4 +91,9 @@ unary_instruction! {
     MathRoundDecimal = MathRoundDecimalImpl(a1: Decimal) -> i64 { Ok(Decimal::round(a1)) }
     MathCeilDecimal = MathCeilDecimalImpl(a1: Decimal) -> i64 { Ok(Decimal::ceil(a1)) }
     MathFloorDecimal = MathFloorDecimalImpl(a1: Decimal) -> i64 { Ok(Decimal::floor(a1)) }
+
+    LengthString = LengthStringImpl(a1: Cow<'a, str>) -> i64 {
```
flyingsilverfin: not u64?

compiler/annotation/function.rs
flyingsilverfin: can we push anything into the function package?

executor/read/control_instruction.rs
flyingsilverfin: Yeah i think itll go in there

compiler/annotation/expression/instructions/unary.rs
dmitrii-ubskii: `i64` is our TypeDB integer.

executor/read/immediate_executor.rs
```diff
@@ -929,3 +936,102 @@ impl CheckExecutor {
...
+
+        row[res.as_usize()] = match self.builtin_id {
+            BuiltinConceptFunctionID::Iid => {
+                let iid = row[self.argument_positions[0].as_usize()].as_thing().iid();
+                VariableValue::Value(Value::String(Cow::Owned(format!("{iid:x}"))))
+            }
+            BuiltinConceptFunctionID::Label => {
+                let ty = row[self.argument_positions[0].as_usize()].as_type();
+                let label = ty.get_label(&**context.snapshot(), context.type_manager())?;
+                VariableValue::Value(Value::String(Cow::Owned(label.to_string())))
```
flyingsilverfin, line 1027: double check that for roles this is `relation:role`

ir/pattern/expression.rs
```diff
@@ -248,46 +252,103 @@ impl BuiltInCall {
...
+        match self {
+            Self::Iid => FunctionSignature::new(
+                FunctionID::Builtin(self),
+                vec![VariableCategory::Thing],
+                vec![(VariableCategory::Value, VariableOptionality::Required)],
+                false,
+            ),
+            Self::Label => FunctionSignature::new(
+                FunctionID::Builtin(self),
+                vec![VariableCategory::Type],
```
flyingsilverfin, line 343: Here we use `thing` and `type`, but elsewhere we used:

```

fn get_builtin_function_annotated_signature(builtin_id: BuiltinConceptFunctionID) -> AnnotatedFunctionSignature {
    match builtin_id {
        BuiltinConceptFunctionID::Iid => AnnotatedFunctionSignature {
            is_stream: false,
            arguments: vec![FunctionParameterAnnotation::AnyConcept],
            returns: vec![FunctionParameterAnnotation::Value(ValueType::String)],
        },
        BuiltinConceptFunctionID::Label => AnnotatedFunctionSignature {
            is_stream: false,
            arguments: vec![FunctionParameterAnnotation::AnyConcept],
            returns: vec![FunctionParameterAnnotation::Value(ValueType::String)],
        },
    }
}
```

I haven't looked further but that seems a bit at odds - shouldn't they match?

ir/translation/expression.rs
```diff
@@ -210,39 +238,72 @@ fn check_builtin_arg_count(
     }
 }
 
-fn to_builtin_id(
+fn is_builtin_value_function(typeql_id: &BuiltinFunctionName) -> bool {
```
flyingsilverfin, line 241: what's the failure mode, ie. what happens if we forget to update this

ir/pattern/expression.rs
dmitrii-ubskii: Function signature types are more about drawing a line between concept types and value types. Variable categories are mostly for correctness checking I think? In particular, type inference doesn't actually care if it assigns the `person` type annotation to a type variable or a thing variable, they are represented the same way throughout.
`FunctionParameterAnnotation::Concept(..)` can be things or types. `AnyConcept` just piggy-backs off of that. I would like to see that separation when we revisit type inference though.

ir/translation/expression.rs
dmitrii-ubskii: Either `to_builtin_concept_function_id` will fail with an internal error (`RepresentationError::InternalNotAConceptBuiltin`). 
If we somehow miss this check entirely and implement a value builtin as a concept builtin (which involves ignoring `ConceptBuiltin` in multiple places), it's just going to execute out of line of expressions which is ultimately fine.
Keeping in mind that this PR is a halfway change just to get `iid()` and `label()` in.

PR #7681, opened 2026-01-23
Implement date/time and string binary operators

dmitrii-ubskii:

We implement addition and subtraction operators for `date`, `datetime`, `datetime_tz`, and `duration`.

We also implement string concatenation, also performedd with the `+` operator.

## Implementation

Drive-by: datetime-tz parsing from `TypeQLLiteral` is fixed. Previously it would panic on ambiguous or unrepresentable datetimes.

### DIFF
--- a/dependencies/typedb/repositories.bzl
+++ b/dependencies/typedb/repositories.bzl
@@ -36,5 +36,5 @@ def typedb_behaviour():
     git_repository(
         name = "typedb_behaviour",
         remote = "https://github.com/typedb/typedb-behaviour",
-        commit = "374be23eeb01127aab9ed8f1dd5dc47eab6a0460",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        commit = "0fc98ce347a58db78f2516cbb49c38a0155a7885",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
     )


compiler/annotation/expression/instructions/mod.rs
```diff
@@ -49,5 +49,6 @@ typedb_error! {
         ListIndexOutOfRange(6, "List index out of range {index}, list length: {length}", index: i64, length: usize),
         ListRangeOutOfRange(7, "List range out of range {from_index}..{to_index}, list length: {length}", from_index: i64, to_index: i64, length: usize),
         OverlongString(8, "Found string with length {len} which is too long to fit in a 64-bit signed integer", len: usize),
+        NegativeDurationSub(9, "Attempting to subtract future from past: {lhs} - {rhs}", lhs: String, rhs: String),
```
flyingsilverfin: Maybe "Cannot subtract larger date from smaller date"? "Future date from past" is a bit odd

flyingsilverfin:
Looks good - needs BDD tests only and i think we can proceed

compiler/annotation/expression/instructions/mod.rs
dmitrii-ubskii: "Larger date" is also odd. "Later date from earlier" maybe?

ir/translation/literal.rs
```diff
@@ -253,7 +253,14 @@ impl FromTypeQLLiteral for chrono::DateTime<TimeZone> {
...
         let tz = TimeZone::from_typeql_literal(&literal.timezone, source_span)?;
-        Ok(NaiveDateTime::new(date, time).and_local_timezone(tz).unwrap())
+        let date_time = NaiveDateTime::new(date, time);
+        match date_time.and_local_timezone(tz) {
+            MappedLocalTime::Single(dt) => Ok(dt),
+            MappedLocalTime::Ambiguous(earliest, latest) => {
+                Err(LiteralParseError::AmbiguousLocalTime { date_time, tz, earliest, latest, source_span })
+            }
+            MappedLocalTime::None => Err(LiteralParseError::NoSuchLocalTime { date_time, tz, source_span }),
+        }
```
dmitrii-ubskii, line 276: This would crash the server previously.

dependencies/typedb/repositories.bzl
```diff
@@ -36,5 +36,5 @@ def typedb_behaviour():
     git_repository(
         name = "typedb_behaviour",
         remote = "https://github.com/typedb/typedb-behaviour",
-        commit = "cf13e741844025a8a0210f25dc295ce966e26d16",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
+        commit = "7f29029a2d92a5892d3623a93fb43d37d49b013e",  # sync-marker: do not remove this comment, this is used for sync-dependencies by @typedb_behaviour
```
dmitrii-ubskii: https://github.com/typedb/typedb-behaviour/pull/400

compiler/annotation/expression/instructions/operators.rs
```diff
@@ -31,11 +38,65 @@ binary_instruction! {
...
+    }
+
+    OpStringAddString = OpStringAddStringImpl(a1: Cow<'a, str>, a2: Cow<'a, str>) -> Cow<'a, str> { Ok(Cow::Owned(format!("{a1}{a2}"))) }
 }
 
 fn checked_div(a1: f64, a2: f64) -> Result<f64, ExpressionEvaluationError> {
     let res = a1 / a2;
-    if res.is_normal() {
+    if res.is_finite() {
         Ok(res)
```
dmitrii-ubskii, line 100: Did this before I noticed https://github.com/typedb/typedb/pull/7680, should disappear in the rebase

encoding/value/duration_value.rs
```diff
@@ -72,6 +75,65 @@ impl Duration {
...
+            let mut days = 0;
+            loop {
+                let duration = Self::between_dates(earlier.date_naive(), later.date_naive() - Days::new(days));
+                if &earlier + duration <= later {
+                    break duration;
+                } else {
+                    days += 1;
+                }
+            }
+        };
```
dmitrii-ubskii, line 116: `date_duration` is such that `earlier + date_duration` is as close as possible to `later` while still being before it or equal.
This will almost always terminate with `days` either `0` or `1`. In extreme cases where there are gaps in date/time in the given time zone, this is guaranteed to end at the latest when `earlier.date_naive() == later.date_naive() - Days::new(days)`, in which case the `date_duration` would be `0`. This would require a gap starting at least a day ahead of the `earlier` datetime and lasting until at most a day before `later`.
The biggest such gap to date, to my knowledge, is Pacific/Apia when Samoa switched their time zone by 24 hours. The gap is 1 day, so this loop should _at most_ execute three times.

compiler/annotation/expression/instructions/mod.rs
flyingsilverfin: yes!

Issue #7671, opened 2026-01-09
Incorrect read query results when data inserted through schema transaction


sam-butcher:
## Description

Query results of read queries may differ when being run in different transaction types when data has been inserted through a schema transaction.

Specifically, the example I have produces no answers in read/write transactions, but produces answers in a schema transaction.

Schema:
```
define

relation rel, relates x, relates y;

entity x, plays rel:x;
entity y, plays rel:y;
```

Data:
```
insert

$x isa x;
$y isa y;
rel (x: $x, y: $y);
```

Query:
```
match rel (x: $x, y: $y);
```

## Environment

1. TypeDB distribution: Cloud
2. TypeDB version: 3.7.0
3. Environment: Mac
4. Client and version: Studio 3.7.0/Console 3.7.0/HTTP API
5. Other details:

## Reproducible Steps

1. Set up

- Launch TypeDB 3.7.0
- Run the schema definition query
- Run the data insertion query **in a schema transaction**

2. Execute

3. Test/Query

Run the query in a schema transaction, run the query in a read transaction

4. Unexpected result

The read transaction fails to produce any answers

## Expected result

Answers are produced in both cases

## Additional information

Might be to do with relation indices:

```
match 
$rel (x: $x);

match
$rel (y: $y);
```

works even from the bugged state

krishnangovindraj:
This does look like it's caused by the `relation_index_available` check in the `put_links_unordered` method.
If so, the fix is to generate relation indices at commit time, as outlined in #7594

krishnangovindraj:
Fixed in #7683

PR #7683, opened 2026-01-27
Always regenerate indices for modified relations on schema commit

krishnangovindraj:
#7594  changed relation-index behaviour so schema transactions avoid reading or writing to relation-indices - and generate indices at commit time for newly inserted relations, or relations of types that became eligible for indices as part of schema modifications. 

## Implementation
 `modified_relations_role_types` in `CardinalityChangeTracker` already tracks relations modified in a schema transaction - unless these relations were deleted. We add `players_in_deleted_relations` to `CardinalityChangeTracker` for that case. If the cardinality of some role on some relation-type changes in a way that makes the relation-type eligible for relation indices,  the cardinality change tracker adds all instances of that type to `modified_relation_role_types` as well.
We use these two sets to regenerate indices.

Addresses: #7671

### DIFF
--- a/concept/thing/thing_manager.rs
+++ b/concept/thing/thing_manager.rs
@@ -2252,9 +2252,7 @@ impl ThingManager {
                         .iterate_range(&prefix_range, storage_counters.clone())
                         .collect_cloned_vec(|k, _| StorageKeyArray::from(k))
                         .map_err(|source| Box::new(ConceptWriteError::SnapshotIterate { source }))?;
-                    collected.into_iter().for_each(|edge| {
-                        snapshot.delete(edge);
-                    });
+                    collected.into_iter().for_each(|edge| snapshot.delete(edge));
                 }
             }
         }


concept/thing/thing_manager.rs
```diff
@@ -1795,15 +1795,16 @@ impl ThingManager {
         self.validate_cardinalities(snapshot, &cardinality_change_tracker, storage_counters.clone())?;
 
         // For immutable schema, the indices are updated at operation time
-        if !Snapshot::IMMUTABLE_SCHEMA && cardinality_change_tracker.has_modified_relates() {
-            self.update_relation_indices_on_cardinality_changes(
+        if !Snapshot::IMMUTABLE_SCHEMA && (cardinality_change_tracker.has_modified_relates() || true) {
```
krishnangovindraj: I can just remove the `true` if we're convinced of the comment.

concept/thing/thing_manager.rs
```diff
@@ -1795,15 +1795,17 @@ impl ThingManager {
         self.validate_cardinalities(snapshot, &cardinality_change_tracker, storage_counters.clone())?;
 
         // For immutable schema, the indices are updated at operation time
-        if !Snapshot::IMMUTABLE_SCHEMA && cardinality_change_tracker.has_modified_relates() {
```
krishnangovindraj, line 1798: Check should be implicit in the contents of `modified_relations_role_types`

concept/thing/thing_manager.rs
```diff
@@ -2211,29 +2213,55 @@ impl ThingManager {
         }
     }
 
-    fn update_relation_indices_on_cardinality_changes(
+    fn update_relation_indices_on_schema_commit(
```
krishnangovindraj, line 2216: Rename to reflect generalisation

concept/tests/test_thing.rs
```diff
@@ -1649,6 +1649,285 @@ fn role_player_duplicates_ordered_small_card() {
     }
 }
 
+#[test]
+fn schema_transactions_generate_role_player_indices() {
```
krishnangovindraj, line 1653: Defines schema;
Adds relations in schema transaction;
Verifies index is available after commit;
Deletes a relation in a schema transaction;
Verifies indices are deleted after commit.

concept/thing/thing_manager.rs
dmitrii-ubskii: Where does this handle newly inserted relations that didn't have their type modified in the schema transaction? I can see in the test that you've covered that case, but I don't see a change here that corresponds to that.

krishnangovindraj: It's already inserted into `modified_relations_role_types` by `CardinalityChangeTracker::collect_modified_links`.

concept/thing/thing_manager.rs
```diff
@@ -2211,29 +2213,55 @@ impl ThingManager {
...
+                    let prefix_range =
+                        KeyRange::new_within(player_pair_for_relation, ThingEdgeIndexedRelation::FIXED_WIDTH_ENCODING);
+                    // Nuke all
+                    let collected = snapshot
+                        .iterate_range(&prefix_range, storage_counters.clone())
+                        .collect_cloned_vec(|k, _| StorageKeyArray::from(k))
+                        .map_err(|source| Box::new(ConceptWriteError::SnapshotIterate { source }))?;
+                    collected.into_iter().for_each(|edge| {
+                        snapshot.delete(edge);
+                    });
```
dmitrii-ubskii:
```suggestion
                    collected.into_iter().for_each(|edge| snapshot.delete(edge));
```
or
```suggestion
                    for edge in collected {
                        snapshot.delete(edge);
                    }
```

.bazelrc
```diff
@@ -4,6 +4,6 @@
 
 try-import ./.bazel-remote-cache.rc
 
-build --incompatible_strict_action_env --@io_bazel_rules_docker//transitions:enable=false --remote_download_toplevel  --disk_cache=~/.bazel_cache
+build --incompatible_strict_action_env --@io_bazel_rules_docker//transitions:enable=false --remote_download_toplevel
```
flyingsilverfin, line 7: Ok - and ill add it back as a bazel profile

PR #7686, opened 2026-01-29
Introduce bazel ci config setting

flyingsilverfin:

We add a bazel config profile `ci`, which does not use a local disk cache. Local disk caches can help speed up builds, especially when switching branches or creating new worktrees. However, they can be very large and cause CI systems to run out of disk space, while also not providing any benefits (roughly building once per CI job, and already using the global networked cache).

## Implementation

* Create new `ci` config, which disables the disk-cache
* Apply the `ci` config to all bazel jobs in CircleCI and in Factory

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -4,6 +4,9 @@
 
 try-import ./.bazel-remote-cache.rc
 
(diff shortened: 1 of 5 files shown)


krishnangovindraj (approved):
CI passed. Should work fine.

PR #7687, opened 2026-01-30
Tag parameters with value type

dmitrii-ubskii:

We tag extracted parameters in queries with their value type. Previously the plan cache would not be able to distinguish queries with integer literals from ones with datetime literals, e.g., which would cause it to retrieve a wrong compiled plan from the cache, causing a server crash.

### DIFF
--- a/ir/pattern/mod.rs
+++ b/ir/pattern/mod.rs
@@ -123,8 +123,8 @@ impl<ID: IrID> Vertex<ID> {
         }
     }
 
-    pub fn as_parameter(&self) -> Option<ParameterID> {
-        if let &Self::Parameter(v) = self {
+    pub fn as_parameter(&self) -> Option<&ParameterID> {
+        if let Self::Parameter(v) = self {
             Some(v)
         } else {
             None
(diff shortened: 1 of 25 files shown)


ir/pattern/mod.rs
```diff
@@ -208,17 +208,17 @@ impl<ID: fmt::Debug> fmt::Display for Vertex<ID> {
     }
 }
 
-#[derive(Clone, Copy)]
+#[derive(Clone)]
```
dmitrii-ubskii, line 211: No longer `Copy` because `ValueType` is not `Copy` on account of `Struct`s. Those contain either a `DefinitionKey` or a `String` depending on where the value type comes from, neither of which are `Copy`.
Cloning a `ParameterID` is free unless it is a `Struct`. Even so, I didn't end up having to clone in that many places, and most uses simply switched to using references.

flyingsilverfin (approved):
Looks good to me!

krishnangovindraj (approved):
Looks good to me.
I do have a question as to whether we need to do the same with IIDs, since different IIDs can be inferred to have different types? I forget whether this was a problem.

dmitrii-ubskii:
> Looks good to me. I do have a question as to whether we need to do the same with IIDs, since different IIDs can be inferred to have different types? I forget whether this was a problem.

It's not a problem because we don't use IIDs in type inference or anywher else during compilation.

PR #7692, opened 2026-02-11
Extensible KV abstraction layer

flyingsilverfin:

We create a new `kv` package, with the intent of allowing alternative backends to be provided besides just RocksDB. To avoid breaking backwards compatibility, we avoid a larger refactor of hiding `Keyspaces` as an implementation detail of RocksDB, since this has already leaked into and affected WAL structure.

Now, we use `enum` dispatch in `kv` to provide RocksDB (or in the future: in-memory, WASM-compatible, etc backends). This lends itself nicely to using feature flags to additively compile in alternative backends.

In this approach, we could even allow users to configure a different storage backend per-database, or provide multiple alternative RocksDB implementations with different configurations!

## Implementation
                                                                                                                                                                                                                                                     
  - Extract RocksDB into a new kv crate, hiding the concrete storage backend behind enum dispatch (KVStore, KVRangeIterator, KVWriteBatch enums)                                                                                                     
  - Remove the KV: KVStore generic parameter from storage, encoding, concept, database, function, ir, and all higher-level crates                                                                                                                    
  - Move key_range module from storage to common/primitive since it is not storage-specific                                                                                                                                                          
  - Move keyspace management (Keyspaces, KeyspaceSet, KeyspaceId) into the kv crate                                                                                                                                                                  
                                                                                                                                                        

## Design Decisions

We explored different variations:

1) full trait-based isolation (`trait KVStore`)

This is the "conceptually best" solution, but has massive consequences: this propagates absolutely everywhere as a generic. We have a huge surface area of functions and structs in `//concept` and `//traversal`, all of which would acquire that generic. Our existing generic, for durability, doesn't pollute these packages

2) Trait object implementation

Trying to keep strongest typing, and hiding the trait in the `//storage` layer seems feasible, but many of the traits we'd need in `//kv` such as `KVStore` are not object safe. As such, we need many workarounds that induce their own high levels of complexity, such as downcasting and use of `Any`, which we've managed to avoid so far

3) Feature gate

One of our users implemented a webassembly backend just with a simple feature gate that switches the compiled `storage` modules. However, this doesn't scale beyond exactly 2 backend without a mess of feature flags. In fact, features don't really support "either A or B or C or D else E" nicely at all, with macros to help with this feeling like hacks saying `A and not B and not C and not D...` etc.

### DIFF
--- a/kv/rocks/iterator.rs
+++ b/kv/rocks/iterator.rs
@@ -58,21 +58,15 @@ impl RocksRangeIterator {
 
     fn accept_value(condition: &ContinueCondition, value: &<Self as LendingIterator>::Item<'_>) -> bool {
         match value {
-            Ok((key, _)) => {
+            &Ok((key, _)) => {
                 match condition {
                     ContinueCondition::ExactPrefix(prefix) => key.starts_with(prefix),
                     ContinueCondition::EndPrefixInclusive(end_inclusive) => {
-                        // if the key is shorter than the end, and the end starts with the key, then it must be OK
-                        //  example: A will be included when searching up to and including AA
-                        // otherwise, the key is longer and we check the corresponding ranges
-                        end_inclusive.starts_with(key) || &key[0..end_inclusive.len()] <= end_inclusive
+                        // Either the key is before the end prefix in the dictionary order, OR starts with the end prefix.
+                        // E.g. 'ABC' is sorted after 'AB', but will be included in the iterator output because it starts with AB.
+                        key <= end_inclusive || key.starts_with(end_inclusive)
                     }
-                    ContinueCondition::EndPrefixExclusive(end_exclusive) => {
-                        // if the key is shorter than the end, and the end starts with the key, then it must be OK
-                        //  example: A will be included when searching up to but not including AA
-                        // otherwise, the key is longer and we check the corresponding ranges
-                        end_exclusive.starts_with(key) || &key[0..end_exclusive.len()] < end_exclusive
-                    }
+                    ContinueCondition::EndPrefixExclusive(end_exclusive) => key < end_exclusive,
                     ContinueCondition::Always => true,
                 }
             }


kv/rocks/iterpool.rs
```diff
@@ -0,0 +1,55 @@
...
+    unprefixed_iterators: LIFOPool<DBRawIterator<'static>>,
+    prefixed_iterator: LIFOPool<DBRawIterator<'static>>,
+}
+
+impl RocksRawIteratorPool {
+    pub fn new() -> Self {
+        Self {
+            // force never pooling by capping at size 0
+            unprefixed_iterators: LIFOPool::new_capped(0),
+            prefixed_iterator: LIFOPool::new_capped(0),
```
flyingsilverfin, line 24: Right now the pooling doesn't do anything - need to contribute iterator refresh to Rust Rocks bindings!

kv/BUILD
flyingsilverfin, line 1: New crate!

kv/iterator.rs
```diff
@@ -0,0 +1,52 @@
...
+use std::cmp::Ordering;
+
+use bytes::byte_array::ByteArray;
+use lending_iterator::{LendingIterator, Seekable};
+
+use crate::{rocks::iterator::RocksRangeIterator, KVStoreError};
+
+pub type KVIteratorItem<'a> = Result<(&'a [u8], &'a [u8]), Box<dyn KVStoreError>>;
+
+pub enum KVRangeIterator {
```
flyingsilverfin, line 18: One of the new enum wrappers!

kv/keyspaces.rs
flyingsilverfin, line 1: This is just straight moved over

kv/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+use resource::profile::StorageCounters;
+
+use crate::{
+    iterator::KVRangeIterator,
+    keyspaces::{KeyspaceSet, Keyspaces, KeyspacesError},
+    rocks::RocksKVStore,
+};
+
+#[derive(Debug)]
+pub enum KVStore {
```
flyingsilverfin, line 25: New KVStore "trait"-like enum wraper that jut dispatches

kv/write_batches.rs
```diff
@@ -0,0 +1,65 @@
...
+
+impl KVWriteBatch {
+    pub fn put(&mut self, key: impl AsRef<[u8]>, value: impl AsRef<[u8]>) {
+        match self {
+            Self::RocksDB(b) => rocksdb::WriteBatch::put(b, key, value),
+        }
+    }
+}
+
+impl Default for KVWriteBatch {
```
flyingsilverfin, line 26: We'll hide this to avoid accidentally creating the wrong write batch type

storage/snapshot/snapshot.rs
```diff
@@ -222,7 +218,7 @@ impl<D: fmt::Debug> fmt::Debug for ReadSnapshot<D> {
 impl<D> ReadSnapshot<D> {
     pub(crate) fn new(storage: Arc<MVCCStorage<D>>, open_sequence_number: SequenceNumber) -> Self {
         // Note: for serialisability, we would need to register the open transaction to the IsolationManager
-        ReadSnapshot { storage, open_sequence_number, iterator_pool: IteratorPool::new() }
+        ReadSnapshot { storage, open_sequence_number }
```
flyingsilverfin, line 221: Pool is conceptually moved down to the KVstore layer instead of the snapshot layer

storage/storage.rs
```diff
@@ -728,12 +703,46 @@ mod tests {
...
+            &storage_path,
+            durability_client,
+            &None,
+        )
+        .unwrap();
+        assert_eq!(storage.get::<0>(&key_2, seq, StorageCounters::DISABLED).unwrap().unwrap(), ByteArray::empty());
+    }
+}
+
+pub(crate) trait FromOperationsBuffer {
```
flyingsilverfin, line 714: New trait we need to bridge the WriteBatches, which exists in `//kv`, with the `OperationsBuffer`, which is used in Snapshot layer

storage/error.rs
```diff
@@ -4,45 +4,31 @@
...
-}
-
-impl Error for MVCCStorageError {
-    fn source(&self) -> Option<&(dyn Error + 'static)> {
-        match &self.kind {
-            MVCCStorageErrorKind::FailedToDeleteStorage { source, .. } => Some(source),
-            MVCCStorageErrorKind::KeyspaceError { source, .. } => Some(source),
-            MVCCStorageErrorKind::KeyspaceDeleteError { source } => Some(source),
-        }
+typedb_error!(
```
dmitrii-ubskii, line 12: We should standardize on using `{}` for item macros like this

storage/recovery/commit_recovery.rs
```diff
@@ -110,12 +112,12 @@ pub(crate) fn apply_recovered(
...
                 match validated_commit {
                     ValidatedCommit::Write(write_batches) => {
-                        MVCCStorage::persist_commit_status(true, commit_sequence_number, durability_client)
+                        MVCCStorage::<_>::persist_commit_status(true, commit_sequence_number, durability_client)
                             .map_err(|error| DurabilityClientWrite { typedb_source: error })?;
                         pending_writes.push(write_batches);
                     }
                     ValidatedCommit::Conflict(_) => {
-                        MVCCStorage::persist_commit_status(false, commit_sequence_number, durability_client)
+                        MVCCStorage::<_>::persist_commit_status(false, commit_sequence_number, durability_client)
```
dmitrii-ubskii: Why?

storage/storage.rs
```diff
@@ -143,9 +143,12 @@ impl<Durability> MVCCStorage<Durability> {
...
             }
-            Some(checkpoint) => checkpoint
-                .recover_storage::<KS, _>(&storage_dir, &durability_client)
-                .map_err(|error| RecoverFromCheckpoint { name: name.to_owned(), typedb_source: error })?,
+            Some(checkpoint) => {
+                let (kv_store, next_sequence_number) = checkpoint
+                    .recover_storage::<KS, _>(&storage_dir, &durability_client)
+                    .map_err(|error| RecoverFromCheckpoint { name: name.to_owned(), typedb_source: error })?;
+                (kv_store, next_sequence_number)
+            }
```
dmitrii-ubskii: Is this not identical to the original?

storage/storage.rs
```diff
@@ -435,53 +418,46 @@ impl<Durability> MVCCStorage<Durability> {
...
+            .map_err(|e| MVCCStorageError::KeyspaceError {
+                storage_name: (*self.name()).clone(),
+                keyspace_name: self.keyspaces.get(key.keyspace_id()).name(),
+                typedb_source: e.into(),
             })
             .unwrap_or_log()
     }
 
-    pub fn get_raw_mapped<M, V>(&self, key: StorageKeyReference<'_>, mut mapper: M) -> Option<V>
+    pub fn get_raw_mapped<M, V>(&self, key: StorageKeyReference<'_>, mapper: &mut M) -> Option<V>
```
dmitrii-ubskii: This change (and similar below) strictly makes the API less flexible for no apparent reason

storage/storage.rs
```diff
@@ -435,53 +418,46 @@ impl<Durability> MVCCStorage<Durability> {
...
         self.keyspaces
             .get(key.keyspace_id())
-            .get(key.bytes(), |value| mapper(value))
-            .map_err(|e| MVCCStorageError {
-                storage_name: self.name(),
-                kind: MVCCStorageErrorKind::KeyspaceError {
-                    source: Arc::new(e),
-                    keyspace_name: self.keyspaces.get(key.keyspace_id()).name(),
-                },
+            .get(key.bytes(), &mut |value| mapper(value))
```
dmitrii-ubskii:
```suggestion
            .get(key.bytes(), mapper)
```

storage/recovery/checkpoint.rs
```diff
@@ -16,12 +14,15 @@ use std::{
 use chrono::Utc;
 use error::typedb_error;
 use itertools::Itertools;
+use kv::{
+    keyspaces::{KeyspaceSet, Keyspaces, KeyspacesError},
```
dmitrii-ubskii, line 18: Does storage even need to be aware of keyspaces anymore?

encoding/graph/thing/vertex_generator.rs
```diff
@@ -10,9 +10,9 @@ use std::sync::{
 };
 
 use bytes::{byte_array::ByteArray, Bytes};
+use primitive::key_range::KeyRange;
 use resource::profile::StorageCounters;
 use storage::{
-    key_range::KeyRange,
```
dmitrii-ubskii, line 15: This should be the only diff in this file.

kv/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+    RocksDB(RocksKVStore),
+}
+
+impl KVStore {
+    pub fn open_keyspaces<KS: KeyspaceSet>(storage_dir: &Path) -> Result<Keyspaces, KeyspacesError> {
+        // TODO: here we have to pick the storage type? Or is it above?
+        RocksKVStore::open_keyspaces::<KS>(storage_dir)
```
dmitrii-ubskii, line 32: The storage type should be picked by the caller by calling e.g. `open_rocks` or `open_in_memory`.

kv/rocks/pool.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn default() -> Self {
+        Self { pool: Default::default(), size_cap: None }
+    }
+}
+
+impl<T> Clone for LIFOPool<T> {
+    fn clone(&self) -> Self {
+        Self { pool: self.pool.clone(), size_cap: self.size_cap.clone() }
+    }
+}
```
dmitrii-ubskii, line 28: Should have been derived

kv/rocks/pool.rs
```diff
@@ -0,0 +1,80 @@
...
+    ops::{Deref, DerefMut},
+    sync::{Arc, Mutex},
+};
+
+pub trait Poolable {}
+
+#[derive(Debug)]
+pub struct LIFOPool<T> {
+    pool: Arc<Mutex<Vec<T>>>,
+    size_cap: Option<usize>,
```
dmitrii-ubskii, line 15: Will we ever use an uncapped pool?

kv/rocks/pool.rs
```diff
@@ -0,0 +1,80 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use std::{
+    ops::{Deref, DerefMut},
+    sync::{Arc, Mutex},
+};
+
+pub trait Poolable {}
```
dmitrii-ubskii: This seems unnecessary

kv/iterator.rs
```diff
@@ -0,0 +1,52 @@
...
+    RocksDB(RocksRangeIterator),
+}
+
+impl LendingIterator for KVRangeIterator {
+    type Item<'a>
+        = KVIteratorItem<'a>
+    where
+        Self: 'a;
```
dmitrii-ubskii: The bound is unnecessary

kv/iterator.rs
```diff
@@ -0,0 +1,52 @@
...
+        match self {
+            Self::RocksDB(iter) => iter.compare_key(item, key),
+        }
+    }
+}
+
+pub(crate) enum ContinueCondition {
+    ExactPrefix(ByteArray<48>),
+    EndPrefixInclusive(ByteArray<48>),
+    EndPrefixExclusive(ByteArray<48>),
```
dmitrii-ubskii: `48` should be a constant in `resource`

kv/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
+}
+
+pub trait KVStoreError: TypeDBError + Send + Sync + std::fmt::Debug {}
+
+impl<T: KVStoreError + 'static> From<T> for Box<dyn KVStoreError> {
+    fn from(value: T) -> Self {
+        Box::new(value)
+    }
+}
```
dmitrii-ubskii: Why?

kv/rocks/iterator.rs
```diff
@@ -0,0 +1,302 @@
...
+                        //  example: A will be included when searching up to and including AA
+                        // otherwise, the key is longer and we check the corresponding ranges
+                        end_inclusive.starts_with(key) || &key[0..end_inclusive.len()] <= end_inclusive
+                    }
+                    ContinueCondition::EndPrefixExclusive(end_exclusive) => {
+                        // if the key is shorter than the end, and the end starts with the key, then it must be OK
+                        //  example: A will be included when searching up to but not including AA
+                        // otherwise, the key is longer and we check the corresponding ranges
+                        end_exclusive.starts_with(key) || &key[0..end_exclusive.len()] < end_exclusive
+                    }
```
dmitrii-ubskii, line 68: How can a key ever be shorter than the prefix?

kv/rocks/iterator.rs
```diff
@@ -0,0 +1,302 @@
...
+impl LendingIterator for DBIterator {
+    type Item<'a>
+        = Result<(&'a [u8], &'a [u8]), rocksdb::Error>
+    where
+        Self: 'a;
+
+    fn next(&mut self) -> Option<Self::Item<'_>> {
+        let next_state = self.next_internal();
+        match next_state {
+            IteratorItemState::None => unreachable!("State after internal check should be error, Some, or Finished"),
```
dmitrii-ubskii, line 244: `next_internal` calls `take_value_else_retain`, which transitions `Some => None`. This looks very reachable.

kv/rocks/iterator.rs
```diff
@@ -0,0 +1,302 @@
...
+            IteratorItemState::None => {
+                // unchanged
+                Self::None
+            }
+            IteratorItemState::Some(_) => mem::replace(self, IteratorItemState::None),
+            IteratorItemState::Finished => {
+                // unchanged, keep finished
+                Self::Finished
+            }
+            IteratorItemState::Err(err) => Self::Err(err.clone()),
```
dmitrii-ubskii: The formatting is all over the place in this whole file
```suggestion
            IteratorItemState::None => Self::None, // unchanged
            IteratorItemState::Some(_) => mem::replace(self, IteratorItemState::None),
            IteratorItemState::Finished => Self::Finished, // unchanged, keep finished
            IteratorItemState::Err(err) => Self::Err(err.clone()),
```

dmitrii-ubskii (changes requested):

storage/storage.rs
flyingsilverfin: yeah i think i had some debug logging in here i can undo this

storage/recovery/checkpoint.rs
flyingsilverfin: Yeah - keyspaces are being crystallized as a concept, but different engines can choose to implement the different keyspaces separately. It's kinda a meh tradeoff but I think it's the right compromise on generality/minimal changes

kv/rocks/pool.rs
flyingsilverfin: Actually we can't cuz of the generic!

kv/rocks/pool.rs
flyingsilverfin: im not sure yet!

kv/lib.rs
flyingsilverfin: There's a bunch of places throughout where we do this:
```
    pub fn checkpoint(&self, checkpoint_dir: &Path) -> Result<(), Box<dyn KVStoreError>> {
        use RocksKVError::{CheckpointExists, CreateRocksDBCheckpoint};

        let checkpoint_dir = checkpoint_dir.join(self.name);
        if checkpoint_dir.exists() {
            return Err(CheckpointExists { name: self.name, dir: checkpoint_dir.display().to_string() }.into());
        }
        ...
  ```

WIthout that `.into()`, it becomes way more verbose in about 20 places (* the number of kv stores we will support)

File: storage/error.rs (hunk 7 above)
Comment (flyingsilverfin) [reply storage/error.rs:12]: I'm not going to make this change here, since we have about 100 uses with `()` instead of `{}`, but this is a good note we could do in a different PR

File: kv/lib.rs (hunk 20 above)
Comment (flyingsilverfin) [reply kv/lib.rs]: Unless thereos a better trick than `Box::new(...) as Box<dyn KvStoreError>` as the alternative implementation

File: kv/rocks/iterator.rs (hunk 21 above)
Comment (flyingsilverfin) [reply kv/rocks/iterator.rs:68]: Huh i dunno, this is just old code that moved to a new location lemme have a think

Comment (flyingsilverfin) [reply kv/rocks/iterator.rs:68]: So cases:
1) end prefix inclusive
  - if key (say 'AA') is shorter than end inclusive (say 'AAA'), we keep
  - if key (say 'AAAB') is longer than end inclusive (say 'AAA'), we keep if prefix is <=
2) end prefix exclusive
  - same first condition
  - essentially same second condition, except we keep if prefix is < instead of <=

looks right to me!

Comment (flyingsilverfin) [reply kv/rocks/iterator.rs:68]: In general, that's all this function is meant to execute - not enforce how it's meant to be used. As always!

However, I think this case can actually arise. Say we iterate from start prefix `A` to end prefix inclusive `DD`. We can in storage have keys `A, B, C, DDF, E, EEE` - we keep the first 4, and stop on the 5th. The 6th also would not qualify.

File: kv/rocks/iterator.rs (hunk 22 above)
Comment (flyingsilverfin) [reply kv/rocks/iterator.rs:244]: it does transition Some to None, but in doing some it uses `mem::replace`, which returns the old value that was not None so it should be fine.

Kinda weird to read tho i agree not sure how i ever got it right eh

File: kv/lib.rs (hunk 14 above)
Comment (flyingsilverfin) [reply kv/lib.rs:32]: Will evolve, could be that or by config or something. for now will be all

File: kv/rocks/iterator.rs (hunk 21 above)
Comment (flyingsilverfin) [reply kv/rocks/iterator.rs:68]: Ok but i think there's a different bug here: if the Key is _longer_ than the end target but not in the range. Claude found that bug while I was passing my in-memory iteration to the reviewer subagent!!

Solution: we can just defer to Rusts's byte slice lexicographical ordering right?


```
            ContinueCondition::EndPrefixInclusive(end_inclusive) => {
                // pass to Rust's lexicographical byte comparison
                *key <= &**end_inclusive
            }
            ContinueCondition::EndPrefixExclusive(end_exclusive) => {
                // pass to Rust's lexicographical byte comparison
                *key < &**end_exclusive
            }

```
no?

Comment (flyingsilverfin) [reply kv/rocks/iterator.rs:68]: Actually, that would be wrong in the first case. We need this:
```
            ContinueCondition::EndPrefixInclusive(end_inclusive) => {
                // pass to Rust's lexicographical byte comparison
                key.starts_with(end_inclusive) || *key < &**end_inclusive            
            }
            ContinueCondition::EndPrefixExclusive(end_exclusive) => {
                // pass to Rust's lexicographical byte comparison
                *key < &**end_exclusive
            }
```

Basically, the inclusive case is that we want to accept all `AA*` when the condition is `EndPrefixInclusive` - so we should accept `A`, `AA`, `AAABBB` and `AAZ` but as soon as it bcomes `AB` or `B` we exclude.

The `EndPrefixExclusive` is normal lexicographical ordering with `<`

PR #7700, opened 2026-02-20
Optimize transaction open - +240% read-write throughput

flyingsilverfin:

TypeDB's history versioning architecture, combined with the desire to keep full external causality, means that opening a transaction may have to wait for transactions that committed to be 'caught up to' by the visible version number (`watermark`). 

We optimize this by changing to a the earliest version number that retains causality without needing client coorperation (most recent fully committed snapshot) instead of the most recently validated (but not yet KV-persisted, or ack'ed) version number. This leads to shorter wait times while the watermark catches up, speeding up in particular highly parallelized data write transactions by about 5%.

```
  300k ops, 3 runs per branch (bench-opt-pureinsert / bench-master-pureinsert)
  Threads   Batch    opt mean    ±sd     master mean    ±sd      Delta      p
        2      10      7,759    420         7,695      416      +0.8%   0.85
        4      10     14,429    286        14,173    1,194      +1.8%   0.72
        8      10     24,343    557        22,851      415      +6.5%   <0.001  ***
       16      10     34,663  1,861        29,342    1,006     +18.1%   <0.001  ***
        2     100     18,026    337        17,821      523      +1.1%   0.57
        4     100     33,414    673        31,281    1,208      +6.8%   0.008   **
        8     100     55,590  2,131        52,981    2,002      +4.9%   0.12
       16     100     61,629  2,666        58,704    7,699      +5.0%   0.53
```

### Secondary throughput improvements

The secondary consequence of this change are highly interesting. In a mixed 50/50 read and write workload, we increase **total system query throughput by 2.4x**. Note that this and all tests were on small scale data, so mostly were in-memory (larger scale test that hit disk would also be interesting to verify)

```
(batch=1, 100k ops, 8 thread total)
opt 4W+4R    master 4W+4R    Delta
  Write ops/s          1,056          1,119     -5.6%
  Read ops/s           6,881          2,117   +225.1%
  Total ops/s          7,937          3,236   +145.3%
```

This occurs because the reader threads are now spending much less time blocking & waiting for the watermark, and much more time executing.

If we increase thread counts further, we find that reader threads end up starving the writer threads, as the 10 core machine is fully subscribed:
```
opt 8W+8R    master 8W+8R    Delta
  Write ops/s            707          1,892    -62.6%
  Read ops/s          11,191          4,073   +174.7%
  Total ops/s         11,898          5,965    +99.5%
```

To confirm the hypothesis that the write performance reduction is due to total machine oversubscription, we test that pure reader threads are able to fully saturate the CPU. We did an experiment showing ops performed in a fixed time period (10 seconds) with an increasing number of threads:

```
Reader CPU Saturation Test (10 CPU cores, 10s per config)

  Threads   Total reads/s   Per-thread reads/s   Scaling
        1          1,720              1,720        1.0x
        2          3,755              1,877        2.2x
        4          7,391              1,848        4.3
        8         14,475              1,809        8.4x
       10         16,361              1,636        9.5x  (peak)
       12         15,879              1,323        9.2x  (declining)
       16         12,073                755        7.0x  (worse than 8!)
```

So we can confirm that pure reader threads can fully saturate the CPU, and over-parallelizing penalizes and drops throughput! We conclude that write throughput drop in the previous test with 8 read and 8 write threads is because the 8 reader threads are essentially fully saturating 8 cores on their own, and the 8 writer threads are competing for resources. We should theoretically see similar behaviour if we had a 2 core machine with 8 writer threads.

### Test for commit validation cost:

Since we hypothesize that commit validation could take slightly longer, but we spend less time opening transactions, there is likely a point where larger commits can slow down commit validation which can slow down throughput overall. In this test, we write larger and larger batches of straight inserts, which still must be serialized and validated against each other to look for conflicts. (Note: without conflicts these validations can run and finish, without needing to depend on each others' validation results. However - testing with a commit that forces commit waiting doesn't show any different results).

```
PureInsert COMPLETE — all batch sizes, 5 runs, 300k ops

   Batch  Thr   opt mean      ±sd   CV%   mst mean      ±sd   CV%       Δ%        p
  ==========================================================================================
       1    1        332        9  2.6%        339        2  0.6%    -2.1%   0.072
       1    4      1,198       48  4.0%      1,017      122 12.0%   +17.8%   0.002   **
       1    8      2,207      110  5.0%      1,822      230 12.6%   +21.1%   <0.001  ***

     100    1      9,342      399  4.3%      9,878      588  6.0%    -5.4%   0.091
     100    4     31,241    1,185  3.8%     31,931      566  1.8%    -2.2%   0.240
     100    8     47,889    6,881 14.4%     53,577    2,334  4.4%   -10.6%   0.080

    1000    1     10,851       63  0.6%     11,141      102  0.9%    -2.6%   <0.001  ***
    1000    4     37,543      516  1.4%     39,668      557  1.4%    -5.4%   <0.001  ***
    1000    8     57,808    5,816 10.1%     64,092    1,434  2.2%    -9.8%   0.019    *

    2000    1     13,450      177  1.3%     13,542       28  0.2%    -0.7%   0.253
    2000    4     45,045    1,581  3.5%     43,756      445  1.0%    +2.9%   0.079
    2000    8     75,975      934  1.2%     76,528    1,932  2.5%    -0.7%   0.564

    4000    1     13,563      118  0.9%     13,564       55  0.4%    -0.0%   0.978
    4000    4     37,011    1,111  3.0%     37,269      960  2.6%    -0.7%   0.694
    4000    8     65,881    4,023  6.1%     60,821    9,051 14.9%    +8.3%   0.253
```

We see that for tiny commits (size 1), we see significant gains, but at around 1000 we see some reduction in throughput. After that, we go back to being within the noise.

## Implementation

Instead of relying on the IsolationManager to provide the "most recently validated sequence number", we use the `MVCCStorage` itself to provide the "most recently committed sequence number". This is significantly later in the commit path, which means that at a particular time `T`, the assigned number is likely to be lower.

Another way to look at it:

Old approach: at time T, we open a new transaction at `10` because another transactoin just bumped the sequence number from `9` to `10` right when passing isolation validation. Watermark is at `5`. We have to wait for watermark to rise to `10` before the transaction can finish opening.

New approach: at time T, we open a transaction at `9`, because another transaction is still waiting for WAL sync and RocksDB writes before updating the sequence number from `9` to `10`. Watermark is at `5`. We only have to wait for watermark to rise to `9`, which is like is quicker than `10`.

### Benchmark

We also commit a benchmark suite that can be used to profile changes made to the system in terms of end-to-end throughput.

### DIFF
--- a/storage/isolation_manager.rs
+++ b/storage/isolation_manager.rs
@@ -565,17 +565,13 @@ impl<const SIZE: usize> TimelineWindow<SIZE> {
     fn get_status(&self, sequence_number: SequenceNumber) -> CommitStatus<'_> {
         let index = sequence_number - self.start;
         let status = SlotMarker::from(self.slot_status[index].load(Ordering::SeqCst));
-        if let SlotMarker::Empty = status {
-            CommitStatus::Empty
-        } else {
-            let record = self.commit_records[index].get().unwrap();
-            match status {
-                SlotMarker::Empty => unreachable!(),
-                SlotMarker::Pending => CommitStatus::Pending(MaybeOwns::Borrowed(record)),
-                SlotMarker::Validated => CommitStatus::Validated(MaybeOwns::Borrowed(record)),
-                SlotMarker::Applied => CommitStatus::Applied(MaybeOwns::Borrowed(record)),
-                SlotMarker::Aborted => CommitStatus::Aborted,
-            }
+        let lazy_record = || self.commit_records[index].get().unwrap();
+        match status {
+            SlotMarker::Empty => CommitStatus::Empty,
+            SlotMarker::Aborted => CommitStatus::Aborted,
+            SlotMarker::Pending => CommitStatus::Pending(MaybeOwns::Borrowed(lazy_record())),
+            SlotMarker::Validated => CommitStatus::Validated(MaybeOwns::Borrowed(lazy_record())),
+            SlotMarker::Applied => CommitStatus::Applied(MaybeOwns::Borrowed(lazy_record())),
         }
     }
 


krishnangovindraj (changes requested):
If I understand correctly, we have lower transaction open times because our read watermark is lower. This would also mean we end up with a bigger range of commits to validate against.
The current benchmark has a fixed number of threads, so this is artificially limited to the this number of threads. I don't think the numbers are representative. If we did want numbers, I'd prefer a benchmark which sends a fixed number of requests per second, and ensure this approach doesn't diverge sooner than the existing approach( Where divergence means the isolation manager fills up faster than it can validate)

flyingsilverfin:
My instinct is that on the whole, this is conceptually better still, due to where the waiting periods are.

In the existing implementation, we wait for, in my example, watermark to bump from 9 to 10 on open. The open sequence number we receive is the highest validated isolation manager sequence number. This is increased on conflict check, so the remaining work to be done on the commit path is:
1) wait for WAL sync
2) write to KV store

Then we notify that the commit was applied, and the watermark can go up. This path can be quite long.

In the proposed implementation, we immediately see and open on 9. Krishnan correctly points out that we will now incur a larger conflict detection range: we have to validate our commit against commit at 10, as well. _However_, at commit time, this is relatively cheap and fast:
1) we are extremely unlikely to be queued up waiting for commit 10 to finish validating itself, since it's the oldest possible commit we have to validate against
2) we assume conflicts are rare, and commits are generally small - so this is fast for each additional commit to validate against

Primarily, we'll never have to wait for a fsync or a KV store write, which are relatively slow operations.

krishnangovindraj:
Can we commit the benchmark that's been shown, and write one with larger, possibly contending commits to check the performance in that case too?

flyingsilverfin:
Interestingly, I ran a big set of benchmarks that I will aim to check in overnight and got: significant reduction in throughput with higher thread counts on write-heavy workloads, and in mixed read/write workloads we get much higher read throughput and much lower write throughput. Definitely worth looking into if we want that/why it's arising

storage/storage.rs
```diff
@@ -283,7 +295,10 @@ impl<Durability> MVCCStorage<Durability> {
                 commit_profile.snapshot_durable_write_data_confirmed();
                 Err(Durability { name: self.name.clone(), typedb_source: error })
             }
-        }
+        };
+
+        self.update_highest_committed_snapshot(commit_sequence_number);
```
krishnangovindraj, line 318: (Anchor for discord)

PR #7703, opened 2026-02-25
Fix lowering so equality checks against variables can be inlined to iterators

### LINKED ISSUE #7701: Planner can generate attributes across `==` (Optimisation)
## Description
The planner may currently "generates & tests" queries that look up an attribute by value.
```
match
  $e1 isa e1, has a1 $a;
  $e2 isa e2, has a2 == $a; 
```

## Environment

1. TypeDB distribution: Core
2. TypeDB version: 3.8.0
3. Environment: any
4. Client and version: any compatible
5. Other details:

## Reproducible Steps

1. Set up
```
define 
attribute a1, value string;
attribute a2, value string;
entity e1 owns a1;
entity e2 owns a2;
```

2. Execute
Attached python script:
```python
import random
from typedb.driver import TypeDB
from typedb.api.connection.credentials import Credentials
from typedb.api.connection.driver_options import DriverOptions

from typedb.api.connection.transaction import TransactionType

DATABASE = "issue-temp"

x = TransactionType.READ 
driver = TypeDB.driver("localhost:1729", Credentials("admin", "password"), DriverOptions(is_tls_enabled=False))

try:
    tx = driver.transaction(DATABASE, TransactionType.WRITE)
    for i in range(10_000):
        tx.query(f'insert $e1 isa e1, has a1 "{i}";').resolve()
    tx.commit()

    tx = driver.transaction(DATABASE, TransactionType.WRITE)
    for i in range(10_000):
        tx.query(f'insert $e2 isa e2, has a2 "{i}";').resolve()
    tx.commit()
finally:
    driver.close()
```
3. Test/Query
```
match
  $e1 isa e1, has a1 $a;
  $e2 isa e2, has a2 == $a; 
reduce $c = count($e1); # So studio doesn't cheat.
```

4. Unexpected result
Takes much much longer than it should (~quadratic in number of rows inserted)

## Expected result
Faster execution (~linear in number of rows inserted)

### Workaround
Forcing the join instead of the predicate:
```
match $e1 isa e1, has a1 $a1;
insert $a2 isa a2 == $a1;
match $e2 isa e2, has $a2; 
reduce $c = count($e1);
```

krishnangovindraj:
Fixes a bug which prevented statements of the form `$a isa attribute-type == $value;` from efficiently retrieving attribute instances. It worked fine when a literal was used as value.

## Implementation
Fixes a bug in  lowering during `inline_as_optimisation` which didn't consider variables which were already available. We now track input variables and so ones that aren't produced by the `has` but are already available are also considered (fixes #7701). 
Inline comparison checks also treat `lhs` and `rhs` symmetrically, allowing variables to be used on the RHS. Variable comparisons were never inlined till the changes in this PR.

### DIFF
--- a/executor/instruction/checker.rs
+++ b/executor/instruction/checker.rs
@@ -206,6 +206,22 @@ impl<T> Checker<T> {
         }
     }
 
+    fn make_extractor(
+        &self,
+        vertex: &CheckVertex<ExecutorVariable>,
+        row: &MaybeOwnedRow<'_>,
+        context: &ExecutionContext<impl ReadableSnapshot + 'static>,
+    ) -> Box<dyn for<'a> Fn(&'a T) -> VariableValue<'a>> {
+        match vertex.as_variable().and_then(|v| self.extractors.get(&v)) {
+            None => {
+                let value = get_vertex_value(vertex, Some(row), &context.parameters);
+                let owned_value = value.into_owned();
+                Box::new(move |_| owned_value.clone())
+            }
+            Some(&tuple_extractor) => Box::new(tuple_extractor),
+        }
+    }
+
     pub(crate) fn filter(
         &self,
         context: &ExecutionContext<impl ReadableSnapshot + 'static>,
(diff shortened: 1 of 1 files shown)


compiler/executable/match_/planner/mod.rs
```diff
@@ -417,15 +417,18 @@ impl ConjunctionExecutableBuilder {
         let mut inlined = false;
         let mut added_to_current = false;
         let steps_count = self.steps.len();
+        let mut available_variables = Vec::with_capacity(self.steps.last().map_or(0, |s| s.selected_variables.len()));
+        available_variables.extend(self.input_variables.clone());
```
dmitrii-ubskii:
```suggestion
        available_variables.extend(self.input_variables.iter().cloned());
```
Also I'm pretty sure this can be `copied()` rather than `cloned()`

Issue #7702, opened 2026-02-23
Query cache also cache preamble function parameters


krishnangovindraj:
## Description
The parameters for the pipeline in the function are stored in the function itself, meaning they are cached in the query cache. This leads to wrong results if a literal in the function is changed.

## Environment

1. TypeDB distribution: Core
2. TypeDB version: 3.8
3. Environment: n/a
4. Client and version: n/a
5. Other details: n/a

## Reproducible Steps

1. Set up
None.

2. Execute
```
with 
fun magic_number() -> integer:
match let $x = 5;
return first $x;

match
  let $y in magic_number();
```

3. Test/Query
```
with 
fun magic_number() -> integer:
match let $x = 12; # note the change here
return first $x;

match
  let $y in magic_number();
```

4. Unexpected result
5

## Expected result
12

## Additional information
The query can still cache the functions but the parameters must be stored externally and accepted from each translated query separately.

PR #7706, opened 2026-03-02
QueryCache replaces cached preamble function parameters on get

krishnangovindraj:
Fixes a bug where cached queries would use the cached parameters in preamble pipelines, rather than those from the actual query - leading to wrong results when only the parameters in the preamble had been updated (Fixes: #7702)

## Implementation
QueryCache replaces preamble function parameters on `get`

### DIFF
--- a/compiler/executable/function/mod.rs
+++ b/compiler/executable/function/mod.rs
@@ -64,9 +64,14 @@ impl ExecutableFunctionRegistry {
         self.schema_functions.clone()
     }
 
-    pub fn replace_preamble_parameters(&mut self, preamble_parameters: impl Iterator<Item = Arc<ParameterRegistry>>) {
-        self.preamble_functions.iter_mut().zip(preamble_parameters.into_iter()).for_each(|(mut func, params)| {
-            (&mut func.1).parameter_registry = params;
+    pub fn replace_preamble_parameters(
+        &mut self,
+        preamble_parameters: impl Iterator<Item = (usize, Arc<ParameterRegistry>)>,
+    ) {
+        preamble_parameters.for_each(|(index, params)| {
+            if let Some(func) = self.preamble_functions.get_mut(&index) {
+                func.parameter_registry = params;
+            }
         })
     }
 }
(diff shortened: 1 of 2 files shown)


compiler/executable/function/mod.rs
```diff
@@ -63,6 +63,12 @@ impl ExecutableFunctionRegistry {
     pub(crate) fn schema_functions(&self) -> Arc<HashMap<DefinitionKey, ExecutableFunction>> {
         self.schema_functions.clone()
     }
+
+    pub fn replace_preamble_parameters(&mut self, preamble_parameters: impl Iterator<Item = Arc<ParameterRegistry>>) {
```
flyingsilverfin: I'm confused why are we mutating the cache in its entirety? Is this registry per-query?

If yes, shouldn't we construct it with the parameters replaced in the constructor?

compiler/executable/function/mod.rs
```diff
@@ -63,6 +63,12 @@ impl ExecutableFunctionRegistry {
...
+        self.preamble_functions.iter_mut().zip(preamble_parameters.into_iter()).for_each(|(mut func, params)| {
```
flyingsilverfin: Also this is leaking implementation. The order of preamble functions is undefined - it's a hashmap. Iterating over it is unpredictable no? Should we be basically doing - for each preamble parameters _by function id_, get the function from the registry by that ID and replace the parameters right?

krishnangovindraj: Eep. I should indeed get. Updated.

compiler/executable/function/mod.rs
krishnangovindraj: `QueryCache::get` returns a clone.
We're updating the the Registry in its entirety.
Yes, we could do it that way by splitting executable function and Cachable function into two different types and propagating the changes. I would do the same for the whole pipeline in that case - ExecutablePipeline and CachablePipeline. 
At that point, I think I can also try to pull out the ParameterRegistry out of ExecutionContext and put it back into the execution pipeline. Should I try that big change?

flyingsilverfin: let's keep it minimal for now, i think there's room to update this when we go back to perf/query planning or even parametrized queries in the near future

Issue #7695, opened 2026-02-12
Value of newly inserted attribute "is not available" in other statements in same insert


krishnangovindraj:
## Description
We encounter the following error:
```
[WCP5] Could not determine the value of the insert attribute 'n'.
```

in the insert query:
```
insert
  $n isa name "alice";
  $p isa person, has name == $n; 
```

## Environment

1. TypeDB distribution: Core/Enterprise/Cloud
2. TypeDB version:
3. Environment: Linux/Mac/Windows/TypeDB Cloud/Google Cloud/AWS/Azure
4. Client and version:
5. Other details:

## Reproducible Steps

1. Set up
```
define
attribute name, value string;
entity person, owns name;
```
2. Execute
-

3. Test/Query
```
insert
  $n isa name "alice";
  $p isa person, has name == $n; 
```

4. Unexpected result
`[WCP5] Could not determine the value of the insert attribute 'n'.`

## Expected result
The value is available to extract.

## Additional information

This is interestingly also triggered by:
```
insert
  $p isa name "alice";
  try { $x isa person, has name $p;  };
```
But not by:
```
insert
  $p isa name "alice";
 $x isa person, has name $p; 
```
In the former, I would expect $p to be treated as an attribute.

PR #7708, opened 2026-03-03
Optional inserts receive concepts created in parent as "input"

krishnangovindraj:
Reflecting how optional blocks in match statements see shared variables as "inputs", we pass any concept variables inserted in the root as input variables to optional insert blocks. 

## Implementation
From #7695:

This query used to error because it couldn't determine the value for $p.
Now it errors because `$p` being an input variable yet having a `isa` constraint.
```
insert
  $p isa name "alice";
  try { $x isa person, has name $p;  };
```

### DIFF
--- a/compiler/executable/insert/executable.rs
+++ b/compiler/executable/insert/executable.rs
@@ -102,8 +102,8 @@ pub fn compile(
 
     let input_variables_for_nested = input_variables
         .iter()
-        .map(|(v, p)| (v.clone(), p.clone()))
-        .chain(concept_instructions_map.iter().map(|(var, instr)| (var.clone(), instr.inserted_position().0)))
+        .map(|(&v, &p)| (v, p))
+        .chain(concept_instructions_map.iter().map(|(&var, instr)| (var, instr.inserted_position().0)))
         .collect::<HashMap<Variable, VariablePosition>>();
     let mut optional_inserts = Vec::with_capacity(block.conjunction().nested_patterns().len());
     for nested_pattern in block.conjunction().nested_patterns() {


compiler/executable/insert/executable.rs
```diff
@@ -84,6 +84,11 @@ pub fn compile(
         input_variables,
     )?;
 
+    let input_variables_for_nested = input_variables
+        .iter()
+        .map(|(v, p)| (v.clone(), p.clone()))
+        .chain(concept_instructions_map.iter().map(|(var, instr)| (var.clone(), instr.inserted_position().0)))
```
dmitrii-ubskii: Variables and positions are `Copy`. We should keep `.clone()` meaning there's something more expensive going on in general.
```suggestion
        .iter()
        .map(|(&v, &p)| (v, p))
        .chain(concept_instructions_map.iter().map(|(&var, instr)| (var, instr.inserted_position().0)))
```

compiler/executable/update/executable.rs
```diff
@@ -52,7 +55,7 @@ pub fn compile(
     source_span: Option<Span>,
 ) -> Result<UpdateExecutable, Box<WriteCompilationError>> {
     let mut variable_positions = input_variables.clone();
-    let attributes_inserts = add_inserted_concepts(
+    let attributes_inserts_map = add_inserted_concepts(
```
dmitrii-ubskii, line 58: Shouldn't we use these as inputs for optional updates too?

dmitrii-ubskii (changes requested):
Does `put` use the `insert` executor, or should it also be updated?

compiler/executable/update/executable.rs
krishnangovindraj: Good question. I don't think it's needed because we can't create new attribute bindings in an update.

PR #7709, opened 2026-03-03
count($var) must count unique instances of $var

krishnangovindraj:
We fix `count($var)` to count the unique instances of `$var` - the documented behaviour.

## Implementation
Use a `HashSet<VariableValue>` to collect, instead of a single integer to count.


krishnangovindraj:
Actually, the spec requires you to count non-empty rows, not unique values.

PR #7714, opened 2026-03-11
Mark commits applied during recovery

dmitrii-ubskii:

Fix crash during recovery when the commit validation result was not persisted in the WAL.

The isolation manager expects that when a commit is opened reading a sequence number, all the commits up to and including the opening number must have been persisted to storage.

Previously the commits were only marked as validated in the isolation manager, and only persisted at the very end of recovery. This meant that if the commit corresponding to the open sequence number of the unvalidated commit was not present in the checkpointed storage, the isolation manager would consider opening the pending commit for validation a violation of the watermark invariant, a fatal error.

## Implementation

Apply commits eagerly (when validated) and commmunicate that to isolation.

### DIFF
--- a/storage/recovery/commit_recovery.rs
+++ b/storage/recovery/commit_recovery.rs
@@ -107,19 +107,15 @@ pub(crate) fn apply_recovered(
             }
             RecoveryCommitStatus::Rejected => isolation_manager.load_aborted(commit_sequence_number),
             RecoveryCommitStatus::Pending(commit_record) => {
-                while pending_writes
-                    .first_key_value()
-                    .is_some_and(|(&seq, _)| seq <= commit_record.open_sequence_number())
-                {
-                    // must have been applied for a transaction to have been opened on this open_sequence_number
-                    let Some((sequence_number, _)) = pending_writes.pop_first() else {
-                        unreachable!("just checked first is Some(_)")
-                    };
+                let tail = pending_writes.split_off(&commit_record.open_sequence_number().next());
+                for (sequence_number, write_batches) in pending_writes {
+                    keyspaces.write(write_batches).map_err(|error| KeyspaceWrite { source: error })?;
                     isolation_manager.applied(sequence_number).map_err(|error| Internal {
                         name: Arc::new(database_name.to_owned()),
                         source: Arc::new(error),
                     })?;
                 }
+                pending_writes = tail;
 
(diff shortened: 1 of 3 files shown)


storage/recovery/commit_recovery.rs
```diff
@@ -81,28 +81,42 @@ pub fn load_commit_data_from(
...
+                    commit_sequence_number,
+                    WriteBatches::from_operations(commit_sequence_number, commit_record.operations()),
+                );
                 isolation_manager.load_validated(commit_sequence_number, commit_record);
             }
             RecoveryCommitStatus::Rejected => isolation_manager.load_aborted(commit_sequence_number),
             RecoveryCommitStatus::Pending(commit_record) => {
+                let tail = pending_writes.split_off(&commit_record.open_sequence_number().next());
+                for (sequence_number, write_batches) in pending_writes {
+                    keyspaces.write(write_batches).map_err(|error| KeyspaceWrite { source: error })?;
```
flyingsilverfin: hang on --- normally, we write to keyspaces after isolation manager stuff? i haven't read in detail just wanted to make sure we're validating first. pending commits may still be rejected by isolation manager

dmitrii-ubskii: Pending writes, not pending commits. These are commits we've read from the WAL, that are validated by isolation, and were created after the checkpoint sequence number so still need to be written to KV.

storage/recovery/commit_recovery.rs
```diff
@@ -112,7 +126,7 @@ pub(crate) fn apply_recovered(
                     ValidatedCommit::Write(write_batches) => {
                         MVCCStorage::persist_commit_status(true, commit_sequence_number, durability_client)
                             .map_err(|error| DurabilityClientWrite { typedb_source: error })?;
-                        pending_writes.push(write_batches);
+                        pending_writes.insert(commit_sequence_number, write_batches);
```
flyingsilverfin: Ok so how come we're saving the pending writes anyhow - why can't we just eagerly write the pending writes and update isolation manager? Then 
1) we don't need to save the pending writes at all
2) we don't need the new extra loop to split off and write the preceding pending writes .

Or what do you think?

dmitrii-ubskii: You know, I'm not sure! I think the idea was to keep validation separate from writes, but I don't think that's necessary or helpful. Pushed a change with eeager writes.

flyingsilverfin (approved):

PR #7726, opened 2026-03-19
Optimize compilation and memory usage

flyingsilverfin:

Summarize new feature, bugs fixed, or internal change. Include: relevant examples and related issues. 

## Implementation

Include architectural decisions, new assumptions, unimplemented paths/technical debt and tests written.

## Data

Three targeted fixes for query pipelines with many stages (e.g. 1000+ chained inserts):

1. Filter infer_types input to only block-referenced variables Previously, infer_types cloned ALL upstream variable annotations (N entries) for each stage, even though a simple `insert $x isa person;` only references 1-2 variables. Now we collect referenced_vars first and filter.

2. Skip full annotation pass-through for write stages After type inference, all previous-stage annotations were copied into each BlockAnnotations. For write stages, only block-local annotations are needed by downstream compilation, so we skip the O(N) pass-through.

3. Cache output_row_mapping between compile iterations output_row_mapping() was recomputed (allocating a fresh HashMap) on every loop iteration. Now we cache the result and reuse it.

Measured on `insert $x0 isa person; insert $x1 isa person; ...` pipelines:
- Annotation phase for 3000 stages: 1.92s → 0.11s (17.6x faster)
- End-to-end for 3000 stages: 4.94s → 1.30s (3.8x faster)

### DIFF
--- a/compiler/annotation/match_inference.rs
+++ b/compiler/annotation/match_inference.rs
@@ -114,8 +114,12 @@ pub fn infer_types(
     is_write_stage: bool,
 ) -> Result<BlockAnnotations, TypeInferenceError> {
     let mut type_annotations_by_scope = HashMap::new();
-    let input_annotations = previous_stage_variable_annotations
+    // Optimization: only clone annotations for variables actually referenced by this block,
+    // rather than cloning all upstream annotations (which grows with pipeline length).
+    let referenced_vars: HashSet<Variable> = block.conjunction().referenced_variables().collect();
+    let input_annotations: BTreeMap<_, _> = previous_stage_variable_annotations
         .iter()
+        .filter(|(var, _)| referenced_vars.contains(var))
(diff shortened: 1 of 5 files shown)


compiler/annotation/match_inference.rs
```diff
@@ -114,8 +114,12 @@ pub fn infer_types(
     is_write_stage: bool,
 ) -> Result<BlockAnnotations, TypeInferenceError> {
     let mut type_annotations_by_scope = HashMap::new();
-    let input_annotations = previous_stage_variable_annotations
+    // Optimization: only clone annotations for variables actually referenced by this block,
+    // rather than cloning all upstream annotations (which grows with pipeline length).
+    let referenced_vars: HashSet<Variable> = block.conjunction().referenced_variables().collect();
```
krishnangovindraj, line 119: This is actually done in the binding_modes refactor PR.

flyingsilverfin:
Superceded by #7810

PR #7729, opened 2026-03-20
Check each keyspace is consistent when determining if a checkpoint is complete

dmitrii-ubskii:

Summarize new feature, bugs fixed, or internal change. Include: relevant examples and related issues. 

## Implementation

Currently this should only be able to occur if the server was killed during the deletion of a database.


flyingsilverfin:
I think this should now be redundant. We atomically copy the folder over with some of the changes you and @farost  did to make sure if the folder is there, it's complete - right? I'm going to close this and we can re-open if we must

PR #7730, opened 2026-03-20
Optional function returns

krishnangovindraj:
Handles optional function returns properly. This includes the following errors **which may result in schema functions breaking**:
* A function which returns an optional variable is not declared as doing so
* A variable being assigned an optional return value is not marked with `?`
* A variable which was assigned an optional value is re-used in the same stage (subject to the same rules as regular optional variables).

If you do have stored functions, version `3.10.4` will print warnings to the logs and allow you to update your schema functions to conform to `3.11.0` validation. Going straight to `3.11` with invalid functions in the schema can result in a crash-loop.

## Implementation
BDD: https://github.com/typedb/typedb-behaviour/pull/417

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -166,5 +166,5 @@ bazel_dep(name = "typedb_behaviour", version = "0.0.0")
 git_override(
     module_name = "typedb_behaviour",
     remote = "https://github.com/krishnangovindraj/typedb-behaviour",
-    commit = "ee9ab6f7c7997d24d6cc86ae439f24afdde808e7",
+    commit = "0148d83a3a0b06ee21c39f20f0d1772a46d67325",
 )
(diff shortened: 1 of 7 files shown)


ir/lib.rs
```diff
@@ -217,6 +210,18 @@ typedb_error! {
             "Specifying a scoped label constraint on a label is not allowed.",
             source_span: Option<Span>,
         ),
+        UnmarkedOptionalAssignment(
+            32,
+            "The variable '{variable}' is assigned an optional value but not marked with a '?'",
+            variable: String,
+            source_span: Option<Span>,
+        ),
```
krishnangovindraj, line 218: This isn't the behaviour in the spec. Also, I need to check for undeclared optional returns if I don't already.

ir/translation/match_.rs
```diff
@@ -4,6 +4,8 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+use error::todo_must_implement;
```
flyingsilverfin, line 7: what for?

flyingsilverfin (approved):
Approved, but overall question:
- what happens if the user has a fn committed that now fails validation? how do they unblock themselves?

PR #7751, opened 2026-04-13
Reject query pipelines exceeding 1000 stages

flyingsilverfin:

Add MAX_PIPELINE_STAGES limit enforced in translate_pipeline, covering read, write, and analyse paths. New QueryError::PipelineStagesLimitExceeded variant reports the offending stage count.

## Implementation

- Limit pipeline stages
- add tests for errors too large error pipelines (as an integration test, not BDD, to avoid massive BDD files!)

### DIFF
--- a/query/tests/pipeline_stages_limit.rs
+++ b/query/tests/pipeline_stages_limit.rs
@@ -8,11 +8,8 @@ use std::sync::Arc;
 
 use encoding::graph::definition::definition_key_generator::DefinitionKeyGenerator;
 use function::function_manager::FunctionManager;
-use query::{
-    error::QueryError,
-    query_manager::{QueryManager, MAX_PIPELINE_STAGES},
-};
-use resource::profile::CommitProfile;
+use query::{error::QueryError, query_manager::QueryManager};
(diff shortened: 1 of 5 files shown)


query/tests/pipeline_stages_limit.rs
flyingsilverfin, line 1: This isn't in BDD because it would be too long

query/tests/pipeline_stages_limit.rs
```diff
@@ -0,0 +1,108 @@
...
+        &query_str,
+    );
+
+    if let Err(err) = &result {
+        assert!(
+            !matches!(err.as_ref(), QueryError::PipelineStagesLimitExceeded { .. }),
+            "expected no stages-limit error at the limit, got: {:?}",
+            err
+        );
+    }
```
dmitrii-ubskii: Why would there be another error? Is it expected?

dmitrii-ubskii (approved):

query/tests/pipeline_stages_limit.rs
flyingsilverfin: good pt

PR #7753, opened 2026-04-16
Error on assigning to existing variable in a reduce stage

krishnangovindraj:
Error on reduce assigning to existing variable. Source errors when building functions in a Fetch clause are now propagated.


dmitrii-ubskii (approved):

flyingsilverfin:
Do we have a test for this in BDD now?

PR #7764, opened 2026-04-21
Add 'is' checks when executors produce the same variable

krishnangovindraj:
Constraint executors which must assign the retrieved concepts to the same variable now check to ensure the concepts are equal.

## Implementation
BDD: https://github.com/typedb/typedb-behaviour/pull/418/changes#r3131589259
We also implement a fix for a panic in the relation index executor when variables are repeated.


executor/instruction/indexed_relation_executor.rs
```diff
@@ -148,41 +148,37 @@ impl IndexedRelationExecutor {
...
         }
+        // Swap so index 0 is always the sort variable. Sort the rest by mode.
+        let sort_variable_position = output_tuple_positions.iter().position(|x| *x == Some(sort_variable)).unwrap();
+        output_tuple_positions.swap(0, sort_variable_position);
+        output_tuple_positions[1..5].sort_by_key(|x| variable_modes.get(x.unwrap()).unwrap());
         debug_assert!(output_tuple_positions.iter().all(|option| option.is_some()));
 
         let output_tuple_positions = TuplePositions::Quintuple(output_tuple_positions);
-
+        compile_error!(r#"This won't work when variables are duplicated. See Claude's comment:
```
krishnangovindraj: Discussion

executor/instruction/indexed_relation_executor.rs
```diff
@@ -148,52 +152,44 @@ impl IndexedRelationExecutor {
...
-            }
-            debug_assert!(output_tuple_positions[output_index].is_some());
+        let mut output_tuple_positions: [Option<ExecutorVariable>; 5] = [None; 5];
+        for i in 0..5 {
+            output_tuple_positions[i] = Some(variable_component_ordering[i]);
         }
+        // Swap so index 0 is always the sort variable. Sort the rest by mode.
+        let sort_variable_position = output_tuple_positions.iter().position(|x| *x == Some(sort_variable)).unwrap();
+        output_tuple_positions.swap(0, sort_variable_position);
+        output_tuple_positions[1..5].sort_by_key(|x| variable_modes.get(x.unwrap()).unwrap());
```
krishnangovindraj, line 171: I believe this produces the same positions as the earlier for loop, provided `sort_by_key` uses a stable sort. I don't understand why it was that way in the first place.

executor/read/nested_pattern_executor.rs
```diff
@@ -156,6 +156,13 @@ impl InlinedCallExecutor {
 
     pub(crate) fn map_output(&self, input: MaybeOwnedRow<'_>, batch: FixedBatch) -> FixedBatch {
         let mut output_batch = FixedBatch::new(self.output_width);
+        debug_assert!(
```
krishnangovindraj, line 160: Existing behaviour is that assigning to the same variable fails.
https://github.com/typedb/typedb-behaviour/pull/418/changes#r3131589259

executor/tests/execute_links.rs
krishnangovindraj, line 1: I got Claude to write these, though it's also in BDD now. revert?

krishnangovindraj: reverted

executor/read/tabled_call_executor.rs
```diff
@@ -87,6 +87,13 @@ impl TabledCallExecutor {
     pub(crate) fn map_output(&self, returned_batch: FixedBatch) -> FixedBatch {
         let input = &self.active_executor.as_ref().unwrap().input;
         let mut output_batch = FixedBatch::new(self.output_width);
+        debug_assert!(
+            {
+                use std::collections::HashSet;
+                self.assignment_positions.iter().collect::<HashSet<_>>().len() == self.assignment_positions.len()
```
dmitrii-ubskii: With itertools:
```suggestion
                self.assignment_positions.iter().unique().count() == self.assignment_positions.len()
```

executor/instruction/mod.rs
```diff
@@ -354,3 +353,31 @@ type FilterFn<T> = dyn Fn(&Result<T, Box<ConceptReadError>>) -> Result<bool, Box
...
     minmax_or!(types.into_iter(), unreachable!("Empty type iterator"))
 }
+
+fn next_free_internal_variable<T>(start_at: u16, in_map: &HashMap<ExecutorVariable, T>) -> ExecutorVariable {
+    for i in start_at..u16::MAX {
+        let v = ExecutorVariable::Internal(Variable::new_anonymous(0xffffu16 - i));
+        if !in_map.contains_key(&v) {
+            return v;
+        }
+    }
```
dmitrii-ubskii, line 363: Oof.

executor/instruction/mod.rs
```diff
@@ -354,3 +353,31 @@ type FilterFn<T> = dyn Fn(&Result<T, Box<ConceptReadError>>) -> Result<bool, Box
...
+}
+
+pub(super) fn check_producing_same_variable<F: Copy>(
+    candidates: &[(ExecutorVariable, F)],
+    checks: &mut Vec<CheckInstruction<ExecutorVariable>>,
+    extractors: &mut HashMap<ExecutorVariable, F>,
+) {
+    for i in 0..candidates.len() {
+        for j in i..candidates.len() {
+            if candidates[i].0 == candidates[j].0 {
```
dmitrii-ubskii: This could benefit from destructuring:
```suggestion
    for i in 0..candidates.len() {
        let (var_i, extractor_i) = &candidates[i];
        for (var_j, extractor_j) in &candidates[i + 1..] {
            if var_i == var_j {
```

executor/instruction/mod.rs
```diff
@@ -354,3 +353,31 @@ type FilterFn<T> = dyn Fn(&Result<T, Box<ConceptReadError>>) -> Result<bool, Box
...
+    unreachable!("in_map is never so big.")
+}
+
+pub(super) fn check_producing_same_variable<F: Copy>(
+    candidates: &[(ExecutorVariable, F)],
+    checks: &mut Vec<CheckInstruction<ExecutorVariable>>,
+    extractors: &mut HashMap<ExecutorVariable, F>,
+) {
+    for i in 0..candidates.len() {
+        for j in i..candidates.len() {
```
dmitrii-ubskii: Doesn't this mean we always start with `i == j`? So we always add anonymous variables and an `is`?

dmitrii-ubskii (changes requested):

executor/instruction/mod.rs
krishnangovindraj: Eeep.  Good catch here. I had a (+1) in most cases

executor/instruction/mod.rs
krishnangovindraj: Realistically, i can't go above 5.

krishnangovindraj:
Close in favour of #7782

PR #7767, opened 2026-04-22
feat: add read-only PostgreSQL interoperability for BI tools, dbt, and DuckDB

npiesco:
## Product change and motivation

This PR introduces a new read-only interoperability feature for TypeDB CE: TypeDB projection catalogs can now be used from PostgreSQL-compatible BI tools, dbt workflows, and DuckDB through the experimental pgwire surface.

That means TypeDB can participate in read-only analytics and schema-discovery workflows from:
- PostgreSQL-compatible BI tools such as Looker, Metabase, and Power BI
- standard PostgreSQL clients such as `psql`
- dbt through a real packaged `type: typedb` adapter for workflows such as `dbt debug`, `dbt parse`, `dbt compile`, `dbt show`, `dbt test`, and `dbt docs generate`
- DuckDB through its `postgres` extension using `ATTACH ... (TYPE postgres)` against the live pgwire endpoint

## Implementation

- Added connection-scoped pgwire session handling so session expressions such as `current_user`, `current_database()`, `current_schema()`, `pg_backend_pid()`, and `inet_server_port()` reflect the active connection instead of placeholder values.
- Updated the shipped CE config and docs so pgwire is presented as an experimental read-only compat layer and is disabled by default in `server/config.yml`.
- Aligned the built-in `typedb psql` helpers and BI presets with the default development credentials `admin` / `password`.
- Packaged `tools/dbt-typedb` as a real installable adapter on top of `dbt-postgres`, including package metadata, profile template, include assets, tests, and read-only SQL guards that reject write-oriented dbt workflows.
- Added real DuckDB interoperability coverage through a shared live pgwire fixture, including attach-time introspection capture, projection scans, ordering, filtering, `COUNT(*)`, quoted identifiers, alias ordering, `LIMIT 0`, and `pg_catalog` session metadata queries.
- Taught the pgwire facade the DuckDB query shapes it emits, including binary `COPY (SELECT ...) TO STDOUT (FORMAT binary)`, named-column projection over virtual catalog tables, and session-aware `pg_database`, `pg_settings`, and `pg_stat_activity` handling.
- Verified DuckDB against a temporary file-backed `.duckdb` database and assert that the fixture cleans up its temp directory after each run.
- Documented the DuckDB connection flow in the README, including required `INSTALL postgres; LOAD postgres; ATTACH ...` steps and current read-only limitations.

## Validation

- `cargo test -p server pgwire_config_parsed_from_yaml -- --nocapture`
- `cargo test -p server`
[...]

PR #7769, opened 2026-04-22
feat: package dbt-typedb as a real adapter

npiesco:

The repository documents dbt interoperability for the pgwire layer, but it does not currently ship a real installable adapter package. That overstates the integration story and leaves users to assemble it manually.

This change turns `tools/dbt-typedb` into a real package that registers `type: typedb`, reuses `dbt-postgres` connection and catalog behavior, and rejects write-oriented SQL that the read-only pgwire compatibility layer does not support.

This PR is intended to follow typedb/typedb#7767.

## Implementation

Added package metadata and source layout for an installable adapter:
- `tools/dbt-typedb/pyproject.toml`
- `tools/dbt-typedb/src/dbt/...`
- `tools/dbt-typedb/uv.lock`

Implemented:
- adapter plugin registration
- `TypedbCredentials`
- `TypedbConnectionManager`
- read-only SQL gating before unsupported DDL and DML reaches the server
- adapter constraint capability declarations and no incremental strategy support
- packaged dbt include assets and profile template

Updated documentation and example profiles:
- `tools/dbt-typedb/README.md`
- `tools/dbt-typedb/profiles.yml`

Tests written and run:
- `Push-Location tools/dbt-typedb; uv run python -m unittest discover -s tests; Pop-Location`


npiesco:
Closing this split PR in favor of #7767, which now carries the full combined change set.

PR #7770, opened 2026-04-23
[DRAFT/REJECT] Rewrite reused produced variables

krishnangovindraj:
The ick here is that it breaks studio for these cases, since anonymous variables are not returned.

While building the IR, constraints may be rewritten so the variables bound by it are unique.
E.g.: `$r links (some-role: $r);` is rewritten to `$_i links (some-role: $r); $r is $_i`

## Implementation
* ConstraintBuilder `add_<constraint>` methods do this transformation. The constraints themselves remain unaware
* Write stages are updated to handle `is` constraints.
* The relation index transformation must manually handle this.


compiler/transformation/relation_index.rs
```diff
@@ -202,3 +233,29 @@ fn index_decrement_if_removing(index: usize, removed_1: usize, removed_2: usize)
...
+
+        let vv_annotations: BTreeMap<_, _> =
+            vertex_annotations.iter().copied().map(|type_| (type_, vec![type_])).collect();
+        let lr_annotations = LeftRightAnnotations::new(vv_annotations.clone(), vv_annotations);
+        annotations
+            .constraint_annotations_mut()
+            .insert(constraint_is, ConstraintTypeAnnotations::LeftRight(lr_annotations));
+    }
+    Ok(())
+}
```
krishnangovindraj, line 261: If we want to transform all constraints here, these are the things we have to keep in mind + updating the ConstraintAnnotations since we've changed the constraints.
It's not as bad as I thought it was so I might make a branch where I do this.

ir/pattern/constraint.rs
```diff
@@ -287,11 +292,18 @@ impl<'cx, 'reg> ConstraintsBuilder<'cx, 'reg> {
 
     pub(crate) fn add_builtin_function_binding(
         &mut self,
-        assigned: Vec<Variable>,
+        mut assigned: Vec<Variable>,
         builtin_id: super::expression::BuiltinConceptFunctionID,
         arguments: Vec<Variable>,
         source_span: Option<Span>,
     ) -> Result<&FunctionCallBinding<Variable>, Box<RepresentationError>> {
+        for i in 0..assigned.len() {
```
krishnangovindraj, line 301: There's a non built-in one for sure as well where I need to do this.

krishnangovindraj:
#7782  wins out

PR #7772, opened 2026-04-23
Optional function returns are checked and print WARN log lines

krishnangovindraj:
Optional function returns are checked and any errors result in warnings being printed to the log. This is **temporary** to provide a path to non-breaking update and **will be escalated to query-compilation errors in a future release**.

The following cases are checked:
* A function which returns an optional variable is not declared as doing so. 
* A variable being assigned an optional return value is not marked with `?`
* A variable which was assigned an optional value is re-used in the same stage (subject to the same rules as regular optional variables).

## Implementation
```
with fun max_of() -> integer:
match let $x = 5; $x < 3;
return max($x);

match  let $m = max_of();

# prints
2026-04-23T15:50:53.906974Z  WARN ThreadId(22) ir/translation/function.rs:312: Function has inconsistent return optionality. This will fail in the next version:
The optionality of the value returned by the function at index '0' did not match that declared in the signature. 
Signature: max_of() -> integer
Definition: return max($x);
```

```
with fun max_of() -> integer?:
match let $x = 5; $x < 3;
return max($x);

match  let $m = max_of();

# prints 

2026-04-23T15:54:05.685366Z  WARN ThreadId(21) ir/pattern/constraint.rs:374: Function call assigns to unmarked optional variable. This will fail in the next version:
The variable 'm' is assigned an optional value but not marked with a '?'
```

The optional variable reuse case is likely to be erroring already in the pipeline already.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2577,6 +2577,7 @@ dependencies = [
  "structural_equality",
  "test_utils",
  "test_utils_storage",
+ "tracing",
  "typeql",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -616,7 +616,7 @@ features = {}
 
 		[workspace.dependencies.chrono]
 			features = ["alloc", "clock", "default", "iana-time-zone", "js-sys", "now", "oldtime", "serde", "std", "wasm-bindgen", "wasmbind", "winapi", "windows-link"]
-			version = "0.4.41"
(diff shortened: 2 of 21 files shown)


flyingsilverfin (approved):
Looks good! Let's merge tmrw (ensure after my patch release)

flyingsilverfin:
Lots of red

PR #7773, opened 2026-04-23
Fix distinct stage stack overflow

flyingsilverfin:

Distinct stages relied on recursion instead of a loop, which could rapidly lead to stack overflows. We now use a simple loop instead.

## Implementation

Loop instead of recursion in both the DistinctIterator and the rarely used Cartesian inner iterator.

### DIFF
--- a/executor/pipeline/modifiers.rs
+++ b/executor/pipeline/modifiers.rs
@@ -436,14 +436,15 @@ where
     type Item<'a> = Result<MaybeOwnedRow<'a>, Box<PipelineExecutionError>>;
 
     fn next(&mut self) -> Option<Self::Item<'_>> {
-        match self.input.next() {
-            None => None,
-            Some(Err(err)) => return Some(Err(err)),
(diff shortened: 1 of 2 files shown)


executor/pipeline/modifiers.rs
```diff
@@ -436,14 +436,15 @@ where
...
 
     fn next(&mut self) -> Option<Self::Item<'_>> {
-        match self.input.next() {
-            None => None,
-            Some(Err(err)) => return Some(Err(err)),
-            Some(Ok(row)) => {
-                if self.seen.insert(row.clone().into_owned()) {
-                    Some(Ok(row.clone().into_owned()))
-                } else {
-                    self.next()
```
krishnangovindraj, line 446: Fun. You'd expect the compiler to tail-call optimise this?

flyingsilverfin: you would, but apparently Rust doesn't guarantee it

PR #7768, opened 2026-04-22
fix: default pgwire off and align local defaults

npiesco:
## Product change and motivation

TypeDB CE’s pgwire endpoint is experimental and intentionally read-only, but the shipped configuration and surrounding docs and presets make it too easy to interpret it as a generally enabled PostgreSQL-compatible surface.

This change makes the default CE rollout safer and clearer by:
- disabling pgwire in shipped config by default
- describing the feature as an experimental read-only compatibility layer for projection catalogs
- aligning local examples and BI presets with the actual default development credentials

This PR is intended to follow typedb/typedb#7767.

## Implementation

Updated the shipped config and config expectations so pgwire is opt-in:
- `server/config.yml`
- `server/parameters/config.rs`

Updated user-facing docs and helper scripts:
- `README.md`
- `binary/typedb`
- `binary/typedb.bat`

Aligned BI presets and examples around `admin` / `password` and explicit read-only pgwire usage:
- `tools/bi-presets/README.md`
- `tools/bi-presets/looker_typedb_connection.lkml`
- `tools/bi-presets/metabase_typedb.json`
- `tools/bi-presets/powerbi_typedb.json`

Tests run:
- `cargo test -p server pgwire_config_parsed_from_yaml -- --nocapture`
- `cargo test -p server`

PR #7775, opened 2026-04-24
feat: add real DuckDB pgwire interoperability coverage

npiesco:
## Summary
- add real DuckDB and dbt pgwire interoperability coverage using a shared live listener fixture
- teach the pgwire facade the DuckDB query shapes it emits, including binary COPY and session-aware pg_catalog metadata
- verify DuckDB can run against a temp file-backed database and that the fixture cleans up the temp directory afterward

## Dependency
This branch depends on the existing pgwire stack and is intended to land after:
- #7767
- #7768
- #7769

GitHub cannot open an upstream PR on top of 
piesco:pgwire-session-context because that base branch only exists in the fork, so this PR targets master and depends on the stack above.

## Validation
- cargo test --color never -p projection --test duckdb_real_pgwire duckdb_can_attach_and_discover_projection_tables -- --exact --nocapture
- cargo test --color never -p projection --test client_matrix_real_pgwire duckdb_attach_scan_filter_and_detach_succeeds -- --exact --nocapture
- cargo test --color never -p projection --test duckdb_real_pgwire -- --nocapture


npiesco:
Closing in favor of updating #7767 directly so there is a single upstream PR.

PR #7779, opened 2026-04-25
Fix statistics counting

flyingsilverfin:

We fix one tracking error that means that statistics can still on rare occasions skip records that are being counted.

## Implementation

Next sequence number only jumps to the last counted one, not the last retrieved one.


dmitrii-ubskii (commented):
That should fix it, yes.

PR #7780, opened 2026-04-25
Optimize CI build job by reducing cached artifacts download

flyingsilverfin:

We use the `--remote_download_minimal ` in the `ci` config in Bazel, which reduces the number of artifacts downloaded during a build job in FactoryCI.

In particular, this speeds up jobs where little/nothing has changed by hugely reducing the amount of data downloaded. It could also speed up test jobs by reducing number of downloaded artifacts, but the effects seem to be less dramatic.

Have observed a build job of around 8-12 minutes now, instead of up to 20 or 30.

## Implementation

Updates to the bazelrc:
- Add `--remote_download_minimal` to reduce number of downloaded remote artifacts when unnecessary
- Add `--experimental_remote_cache_eviction_retries=5` to prevent build failures when the remote cache has a race condition due to evictions
- Add `--experimental_repository_downloader_retries=5 --http_timeout_scaling=2.0` -- without these, the jobs were generally flaky down to download failures. This fixes download issues and lets us take advantage of the main minimial downloads change without introducing flakiness.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -14,7 +14,7 @@ build --macos_minimum_os=11.0
 try-import ./.bazel-remote-cache.rc
 
 # Allow empty globs (many checkstyle_test patterns may not match)
-common --noincompatible_disallow_empty_glob
+common --noincompatible_disallow_empty_glob --experimental_repository_downloader_retries=5 --http_timeout_scaling=2.0
 
 # Fix Java runfiles path resolution in Bzlmod mode (required by release note validation)
 build --legacy_external_runfiles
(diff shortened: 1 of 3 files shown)


.bazelrc
```diff
@@ -14,7 +14,7 @@ build --macos_minimum_os=11.0
 try-import ./.bazel-remote-cache.rc
 
 # Allow empty globs (many checkstyle_test patterns may not match)
-common --noincompatible_disallow_empty_glob
+common --noincompatible_disallow_empty_glob --experimental_repository_downloader_retries=5 --http_timeout_scaling=2.0
```
flyingsilverfin, line 17: Without the retries, this remote minimal download seems to make the jobs way more unstable.  Not sure why - claude things it's that our jobs get rate limited because we're downloading way more stuff way more frequently. Anyhow - this appears to make it stable & fix it!

RELEASE_NOTES_LATEST.md
```diff
@@ -8,15 +8,28 @@
 
 
 ## New Features
-
+- **Optional function returns are checked and print WARN log lines**
```
flyingsilverfin, line 11: Ran regen since it was missing before

PR #7781, opened 2026-04-26
Add tests for statistics synchronization

flyingsilverfin:

Add more statistics integration tests, as well as a new "load" test, which does a small concurrent, batched load and does a reboot, verifying that all statistics are counted correctly.

## Implementation

- Add new integration tests for statistics updating
- Add new integration test for statistics synchronization on reboot (ensures we include records correctly)
- Add new `stress` config in bazel that we use to run this test 50 times (can be applied to other tests we worry about regressions on in the future)

### DIFF
--- a/concept/tests/test_statistics.rs
+++ b/concept/tests/test_statistics.rs
@@ -6,7 +6,11 @@
 
 #![deny(unused_must_use)]
 
-use std::{collections::BTreeMap, sync::Arc};
+use std::{
+    collections::{BTreeMap, HashMap},
+    hash::Hash,
+    sync::Arc,
+};
 
 use concept::{
     thing::{ThingAPI, object::ObjectAPI, statistics::Statistics, thing_manager::ThingManager},
@@ -28,6 +32,22 @@ use storage::{
 use test_utils_concept::{load_managers, setup_concept_storage};
 use test_utils_encoding::create_core_storage;
 
+fn retain_nonzero_2d<K1: Eq + Hash, K2: Eq + Hash>(map: &mut HashMap<K1, HashMap<K2, u64>>) {
+    for inner in map.values_mut() {
+        inner.retain(|_, v| *v > 0);
+    }
+    map.retain(|_, inner| !inner.is_empty());
+}
+
+fn retain_nonzero_3d<K1: Eq + Hash, K2: Eq + Hash, K3: Eq + Hash>(
+    map: &mut HashMap<K1, HashMap<K2, HashMap<K3, u64>>>,
+) {
+    for middle in map.values_mut() {
+        retain_nonzero_2d(middle);
+    }
+    map.retain(|_, middle| !middle.is_empty());
+}
+
 macro_rules! assert_statistics_eq {
     ($lhs:expr, $rhs:expr) => {
         let Statistics {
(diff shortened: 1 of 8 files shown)


concept/tests/test_statistics.rs
```diff
@@ -362,3 +398,253 @@ fn put_plays() {
...
+    snapshot.commit(&mut CommitProfile::DISABLED).unwrap().unwrap();
+
+    let mut synchronised = Statistics::new(SequenceNumber::MIN);
+    synchronised.may_synchronise(&storage).unwrap();
+
+    assert_statistics_eq!(synchronised, read_statistics(storage, &thing_manager));
+}
+
+#[test]
+fn relation_index_counts() {
```
krishnangovindraj, line 583: Tangential questions:
1. Do these remain up to date if schema modifications cause a relation to go from unindexed to indexed?
2. Do we need to be tracking these separately? (I guess this was discussed long ago, and the answer is that it makes things easier given the encoding is different)

flyingsilverfin: It's a good question... we probably should be able to approximate these values from the other statistics possible? But im quite sure this makes life easier for sure.

Re updating that's a great question! lemme forward this to georgii if he has any idea :|

flyingsilverfin: Yess i found it it's in `update_relation_index_on_schema_commit` ! it is just tracked as any other edge

database/tests/statistics_synchronization.rs
```diff
@@ -0,0 +1,109 @@
...
+                    loop {
+                        let batch_id = next_batch.fetch_add(1, Ordering::Relaxed);
+                        if batch_id >= total_batches {
+                            break;
+                        }
+                        run_insert_batch(&database, batch_id);
+                    }
+                })
+            })
+            .collect();
```
dmitrii-ubskii: Overcomplicated and turns `BATCHES_PER_THREAD` into a lie:
```rs
        let mut handles = Vec::with_capacity(NUM_THREADS);
        for thread_id in 0..NUM_THREADS {
            let database = database.clone();
            let handle = thread::spawn(move || {
                for local_batch_id in 0..BATCHES_PER_THREAD {
                    let batch_id = thread_id * BATCHES_PER_THREAD + local_batch_id;
                    run_insert_batch(&database, batch_id);
                }
            });
            handles.push(handle);
        }
```

database/tests/statistics_synchronization.rs
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+
+    // dbm and database dropped here; IntervalRunner threads shut down synchronously on drop.
+
+    let dbm = DatabaseManager::new(&tmp_dir).unwrap();
+    let database = dbm.database(DB_NAME).unwrap();
+    let metrics = database.get_metrics();
+
+    assert_eq!(metrics.data.entity_count, total_persons as u64, "entity_count after reboot");
```
dmitrii-ubskii, line 81: On failure prints:
```
assertion `left == right` failed: entity_count after reboot
  left: 18
 right: 12
```

That's pretty slick, we should probably adopt this as a convention.

concept/tests/test_statistics.rs
```diff
@@ -66,19 +94,27 @@ macro_rules! assert_statistics_eq {
...
+            relation_role_counts: mut rhs_relation_role_counts,
+            relation_role_player_counts: mut rhs_relation_role_player_counts,
+            player_role_relation_counts: mut rhs_player_role_relation_counts,
+            links_index_counts: mut rhs_player_index_counts,
             ..
         } = $rhs;
         rhs_entity_counts.retain(|_, v| *v > 0);
         rhs_relation_counts.retain(|_, v| *v > 0);
         rhs_attribute_counts.retain(|_, v| *v > 0);
+        rhs_role_counts.retain(|_, v| *v > 0);
```
dmitrii-ubskii: Worth making a `retain_nonzero{_1d}` too then?

concept/tests/test_statistics.rs
```diff
@@ -362,3 +398,253 @@ fn put_plays() {
...
+    name_type.set_value_type(&mut snapshot, &type_manager, &thing_manager, ValueType::String).unwrap();
+    name_type
+        .set_annotation(
+            &mut snapshot,
+            &type_manager,
+            &thing_manager,
+            AttributeTypeAnnotation::Independent(AnnotationIndependent),
+            StorageCounters::DISABLED,
+        )
+        .unwrap();
```
dmitrii-ubskii: I'd rather we don't set `@independent` unless needed, these tests are long enough as it is.

dmitrii-ubskii (approved):

concept/tests/test_statistics.rs
flyingsilverfin: great pt

PR #7782, opened 2026-04-27
Rewrite reused producible variables in transform phase

krishnangovindraj:
Introduces a transformation step which rewrites constraints so the variables bound by it are unique.
E.g.: `$r links (some-role: $r);` is rewritten to `$_i links (some-role: $r); $r is $_i;`

## Implementation
Introduces a transformation phase step that modifies all `conjunction.constraints()` in place if any producible variable is repeated. Then updates:
* Vertex annotations for the newly introduced variables
* ConstraintAnnotations for the newly introduced constraints

### DIFF
--- a/compiler/transformation/unique_constraint_variables.rs
+++ b/compiler/transformation/unique_constraint_variables.rs
@@ -4,10 +4,10 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::BTreeMap;
 
 use ir::{
-    pattern::{Vertex, conjunction::Conjunction, constraint::Constraint, nested_pattern::NestedPattern},
+    pattern::{conjunction::Conjunction, constraint::Constraint, nested_pattern::NestedPattern},
     pipeline::VariableRegistry,
 };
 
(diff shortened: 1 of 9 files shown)


ir/pattern/constraint.rs
```diff
@@ -556,6 +556,137 @@ impl<'cx, 'reg> ConstraintsBuilder<'cx, 'reg> {
...
+            let old_constraint = $constraint.clone().into();
+            if let (Vertex::Variable(kept), Vertex::Variable(changed)) = ($kept, $changed) {
+                let kept = *kept;
+                let new_var = $reg.create_anonymous_variable_copying(kept)?;
+                *changed = new_var.clone();
+
+                // Keep old on LHS
+                $is_collector.push(Is::new(kept, new_var, None));
+                $var_mapping.insert(kept, new_var);
+                $constraint_mapping.insert(old_constraint, $constraint.clone().into());
```
krishnangovindraj: Bug, lhs can't be the old_variable because it clobbers.

krishnangovindraj: Second bug: Calling this multiple times means constraint mapping isn't correct.

compiler/transformation/unique_constraint_variables.rs
```diff
@@ -0,0 +1,79 @@
...
+    annotation::type_annotations::{BlockAnnotations, ConstraintTypeAnnotations, LeftRightAnnotations},
+    transformation::StaticOptimiserError,
+};
+
+pub fn make_constraint_variables_unique(
+    conjunction: &mut Conjunction,
+    variable_registry: &mut VariableRegistry,
+    block_annotations: &mut BlockAnnotations,
+) -> Result<(), StaticOptimiserError> {
+    let (new_checks, constraint_mapping) = conjunction
```
flyingsilverfin: Can you put `make_variables_unique` on `Conjunction` instead? That way you can absorb the `conjunction.register_variable_copy` into that function after it calls `constraints.make_variables_unique`. Right now you're mutating the constraints in one step and then updating the conjunction in another, which is a bit more brittle. 

Then the follow up step before is just managing the annotations to keep them in sync

compiler/transformation/unique_constraint_variables.rs
```diff
@@ -0,0 +1,79 @@
...
+        .constraints_mut()
+        .make_variables_unique(variable_registry)
+        .map_err(|typedb_source| StaticOptimiserError::Representation { typedb_source })?;
+    let annotations = block_annotations.type_annotations_mut_of(conjunction).expect("Expected annotated conjunction");
+
+    for check in new_checks {
+        match check {
+            Constraint::Comparison(cmp) => {
+                conjunction.register_variable_copy(cmp.lhs().as_variable().unwrap(), cmp.rhs().as_variable().unwrap());
```
flyingsilverfin: do we have any guarantee or data structure that tells us the left side is the original one? Can we get that as a return info somehow instead of assuming it here (what if the constraints use rewrite the left one in the future by accident)

flyingsilverfin (commented):
Getting there!

ir/pattern/constraint.rs
```diff
@@ -76,6 +76,110 @@ impl Constraints {
...
+                        old_indexed.get_or_insert_with(|| indexed.clone());
+                        replace_with_anonymous_var(&mut indexed.role_type_2)?;
+                    }
+                    if let Some(old_indexed) = old_indexed {
+                        constraint_mapping.insert(indexed.clone().into(), old_indexed.into());
+                    }
+                }
+                Constraint::FunctionCallBinding(func_call) => {
+                    let mut old_call: Option<FunctionCallBinding<Variable>> = None;
+                    for i in 0..func_call.assigned.len() {
```
flyingsilverfin, line 148: ouuu good catch!

flyingsilverfin (approved):
Looks great! Don't forget about BDD

ir/pattern/constraint.rs
flyingsilverfin: But wait didn't you just ban this in BDD? We can keep this here in case, just want to verify

PR #7783, opened 2026-04-27
Introduce a new RECORD_TYPE for the new type of commit records

farost:

Replace the dynamic deserialization of multiple types of `CommitRecord`s with a separate `RECORD_TYPE` used for the new version of the struct. 

With this approach, we can introduce new versions of `CommitRecord`s with a bit of manual work without worrying about graceful deserialisation of records with essentially different structures but same `RECORD_TYPE`s.

## Implementation

Register `CommitRecord` and `LegacyCommitRecordV1` as separate WAL entry types. Assign a new `RECORD_TYPE` index to `CommitRecord`. Handle all the existing `RECORD_TYPE`s gracefully in the codebase.

### DIFF
--- a/database/tools/read_wal.rs
+++ b/database/tools/read_wal.rs
@@ -11,7 +11,7 @@ use concept::thing::statistics::Statistics;
 use durability::{wal::WAL, DurabilitySequenceNumber, RawRecord};
 use storage::{
     durability_client::{DurabilityClient, DurabilityRecord, WALClient},
-    record::{CommitRecord, StatusRecord},
+    record::{CommitRecord, LegacyCommitRecordV1, StatusRecord},
 };
 
 #[derive(Parser)]
@@ -63,6 +63,10 @@ fn deserialise_record<Record: DurabilityRecord>(raw_bytes: &[u8]) -> Record {
 
 fn print_record(record: RawRecord<'_>) {
     match record.record_type {
+        LegacyCommitRecordV1::RECORD_TYPE => {
+            let legacy: LegacyCommitRecordV1 = deserialise_record(&record.bytes);
+            print_commit(record.sequence_number, CommitRecord::from(legacy));
+        }
         CommitRecord::RECORD_TYPE => print_commit(record.sequence_number, deserialise_record(&record.bytes)),
         StatusRecord::RECORD_TYPE => print_status(record.sequence_number, deserialise_record(&record.bytes)),
         Statistics::RECORD_TYPE => print_statistics(record.sequence_number, deserialise_record(&record.bytes)),
@@ -107,6 +111,7 @@ fn print_range(path: PathBuf, from: u64, to: Option<u64>) {
 
     let mut wal = WALClient::new(wal);
     wal.register_record_type::<Statistics>();
+    wal.register_record_type::<LegacyCommitRecordV1>();
     wal.register_record_type::<CommitRecord>();
     wal.register_record_type::<StatusRecord>();
 
@@ -126,6 +131,7 @@ fn print_at(path: PathBuf, sequence_number: Option<u64>) {
 
     let mut wal = WALClient::new(wal);
     wal.register_record_type::<Statistics>();
+    wal.register_record_type::<LegacyCommitRecordV1>();
     wal.register_record_type::<CommitRecord>();
     wal.register_record_type::<StatusRecord>();
 
--- a/database/tools/replay_wal.rs
+++ b/database/tools/replay_wal.rs
@@ -11,7 +11,7 @@ use concept::thing::statistics::Statistics;
 use durability::{wal::WAL, DurabilitySequenceNumber};
 use storage::{
     durability_client::{DurabilityClient, DurabilityRecord, WALClient},
-    record::{CommitRecord, StatusRecord},
+    record::{CommitRecord, LegacyCommitRecordV1, StatusRecord},
 };
 
 #[derive(Parser)]
@@ -52,6 +52,7 @@ fn main() {
 
     let mut source_wal = WALClient::new(source_wal);
     source_wal.register_record_type::<Statistics>();
+    source_wal.register_record_type::<LegacyCommitRecordV1>();
     source_wal.register_record_type::<CommitRecord>();
     source_wal.register_record_type::<StatusRecord>();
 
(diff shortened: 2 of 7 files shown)


database/tools/read_wal.rs
```diff
@@ -63,6 +63,10 @@ fn deserialise_record<Record: DurabilityRecord>(raw_bytes: &[u8]) -> Record {
 
 fn print_record(record: RawRecord<'_>) {
     match record.record_type {
+        LegacyCommitRecordV1::RECORD_TYPE => {
+            let legacy: LegacyCommitRecordV1 = deserialise_record(&record.bytes);
+            print_commit(record.sequence_number, CommitRecord::from(legacy));
```
farost, line 68: We'll understand that it's a legacy commit record based on its `SnapshotId::UNSET` (0).

database/tools/replay_wal.rs
```diff
@@ -74,6 +76,11 @@ fn main() {
             break;
         }
         match record.record_type {
+            LegacyCommitRecordV1::RECORD_TYPE
+                if cli.kind.is_empty() || cli.kind.contains(&RecordKind::CommitRecord) =>
+            {
+                _ = target_wal.sequenced_write::<LegacyCommitRecordV1>(&deserialise_record(&record.bytes)).unwrap()
```
farost, line 82: Replay the same way it was initially inserted (no conversion)

storage/recovery/commit_recovery.rs
```diff
@@ -60,6 +60,20 @@ pub fn load_commit_data_from_with_context(
         }
 
         match record_type {
+            LegacyCommitRecordV1::RECORD_TYPE => {
+                let legacy = LegacyCommitRecordV1::deserialise_from(&mut &*bytes)
```
farost, line 64: Convert and do the same thing as we do for commit records

storage/record.rs
```diff
@@ -17,25 +17,22 @@ use crate::{
     snapshot::{buffer::OperationsBuffer, lock::LockType, snapshot_id::SnapshotId, write::Write},
 };
 
+/// Legacy records are kept for reading old WAL entries written before TypeDB version upgrade.
 #[derive(Serialize, Deserialize)]
-pub struct CommitRecord {
-    // TODO: this could read-through to the WAL if we have to save memory?
```
farost, line 22: As I understand, it's outdated

storage/record.rs
```diff
@@ -44,11 +41,36 @@ impl From<LegacyCommitRecordV1> for CommitRecord {
...
             commit_type: legacy.commit_type,
-            snapshot_id: None,
+            snapshot_id: SnapshotId::UNSET,
         }
     }
 }
 
+impl DurabilityRecord for LegacyCommitRecordV1 {
+    const RECORD_TYPE: DurabilityRecordType = 0;
+    const RECORD_NAME: &'static str = "legacy_commit_record_v1";
```
farost, line 51: We must be ok with changing the `RECORD_NAME`s now, but this might be a bad pattern for the future if we start saving them on disk.

storage/record.rs
```diff
@@ -44,11 +41,36 @@ impl From<LegacyCommitRecordV1> for CommitRecord {
...
+}
+
+impl SequencedDurabilityRecord for LegacyCommitRecordV1 {}
+
+#[derive(Serialize, Deserialize)]
+pub struct CommitRecord {
+    operations: OperationsBuffer,
+    open_sequence_number: SequenceNumber,
+    commit_type: CommitType,
+    snapshot_id: SnapshotId,
```
farost, line 71: Since the deserialization is no longer a problem, made them mandatory for the new commit records

storage/record.rs
```diff
@@ -17,25 +17,22 @@ use crate::{
...
-    snapshot_id: Option<SnapshotId>,
 }
 
-#[derive(Serialize, Deserialize)]
-struct LegacyCommitRecordV1 {
-    operations: OperationsBuffer,
-    open_sequence_number: SequenceNumber,
-    commit_type: CommitType,
+impl LegacyCommitRecordV1 {
+    pub(crate) fn new(
```
farost, line 29: I basically need this for the tests. I think it's okay to make the legacy records constructors `pub(crate)`, but we may protect it more if you think it's needed

storage/snapshot/snapshot_id.rs
```diff
@@ -23,6 +23,8 @@ pub struct SnapshotId {
 }
 
 impl SnapshotId {
+    pub const UNSET: Self = Self { number: 0 };
```
dmitrii-ubskii, line 26: Is `UNSET` treated any differently from any other `SnapshotId`? If yes, I'd prefer `None` to sentinels. If no, this is fine.

dmitrii-ubskii (approved):

storage/snapshot/snapshot_id.rs
farost: No, it's only needed for the convenience of conversions. We don't expect anyone to depend on SnapshotIds for old records.

database/tools/replay_wal.rs
```diff
@@ -52,6 +52,7 @@ fn main() {
 
     let mut source_wal = WALClient::new(source_wal);
     source_wal.register_record_type::<Statistics>();
+    source_wal.register_record_type::<LegacyCommitRecordV1>();
     source_wal.register_record_type::<CommitRecord>();
```
flyingsilverfin, line 56: do we get errors/warning if there is an unrecognized record type?

storage/storage.rs
```diff
@@ -887,4 +890,65 @@ mod tests {
         assert!(storage.commit_record_exists(seqnum3, snapshot_id3).unwrap());
         assert!(storage.commit_record_exists(seqnum4, snapshot_id4).unwrap());
     }
+
+    #[test]
+    fn test_mixed_legacy_and_current_commit_records() {
```
flyingsilverfin, line 895: good test

database/tools/replay_wal.rs
farost: At which point?

flyingsilverfin: runtime when we load records from log if they're unrecognized at any point?

flyingsilverfin: dont think our arch supports this atm

farost: We do skip them in `commit_recovery` with `_not_storage_record => (), // skip, not storage record`, but we crash in the `replay_wal.rs` tool.

PR #7784, opened 2026-04-28
Fix flaky statistics underflow

flyingsilverfin:

We fix a flaky (occasional `concept` bdd failures) test due to panics in statistics counting. In particular, it occurred in schema transactions that wrote both deleted types and their instances. By doing this in the wrong order, we got an underflow panic.

We also make the server resilient to underflow panics/crashes by logging an error to the server log and sentry (if enabled) with a new macro. We consider this a 'tolerable' error which shouldn't get you into a crash situation.

## Implementation

- Fix ordering issue with updating statistics in a schema transaction 
- Allow underflow to reset a counter to 0, with a logged error
- Drive by warning reduction


diagnostics/lib.rs
```diff
@@ -35,6 +35,14 @@ mod monitoring_server;
 mod reporter;
 mod reports;
 
+#[macro_export]
+macro_rules! error_with_report {
```
flyingsilverfin, line 39: New macro for us to use for things we want to error and know about but not crash for

concept/thing/statistics.rs
```diff
@@ -337,34 +351,49 @@ impl Statistics {
...
+    fn saturating_add(count: &mut u64, delta: i64, label: &str) {
+        match count.checked_add_signed(delta) {
+            Some(value) => *count = value,
+            None => {
+                diagnostics::error_with_report!(
+                    "Unexpected underflow in statistics {} count: {} + {}",
+                    label,
+                    *count,
+                    delta
+                );
```
dmitrii-ubskii, line 363: Would this flood the sentry with ~20 near identical messages on underflow? One of each kind. I'm assuming multiple identical messages from the same endpoint get rolled into one.

dmitrii-ubskii (approved):

concept/thing/statistics.rs
flyingsilverfin: i think they do with a counter. If it becomes too much we can try to simplify it later

flyingsilverfin: ALthough from what i've seen you mostly get 1-2 of these per failure! Eg. has and attribute underflow

PR #7788, opened 2026-04-30
Refactor annotation phase to group annotation context & running variable annotations

krishnangovindraj:
We reduce the number of arguments in functions in the annotation phase, by grouping related arguments together in structures.

## Implementation
* Group `Snapshot, TypeManager, AnnotatedFunctionSignatures` into `AnnotationContext`
  - If we have `VariableRegistry` and `ParameterRegistry`, it's a `PipelineAnnotationContext` 
* Group running concept & value variable annotations into`RunningVariableAnnotations`


compiler/annotation/utils.rs
flyingsilverfin, line 1: We've done a really good job in 3.0 not introducing generic `utils` in many places - do you think this would fit elsewhere well enough? Rust as a `lib.rs` or `mod.rs` for common entry points that i think is a better miscellaneous category if we need than having a `utils.rs`

flyingsilverfin (approved):
Approved w 1 comment

PR #7791, opened 2026-05-01
@doc and @meta annotations

dmitrii-ubskii:

We implement `@doc("docstring")` and `@meta("key", "value")` schema annotations. These annotations can be attached to types and capabilities i.e. `owns`, `plays`, `relates`, and `sub`:

```php
define
entity person @doc("this represents an individual client") @meta("icon", ":silhouette.png"),
  owns name @doc("full name including title");
```
```php
match
  $x isa client;
  let $x_icon = get_meta("icon", $t); # get each client with a UI display icon hint
```

These annotations can also be attached to function definitions. They go at the end of the signature, before the colon introducing the body of the function:
```php
define
fun get_random_number() -> integer
    @doc("chosen by a fair dice roll"):
match
  let $rand = 4;
return first $rand;
```

## Implementation

Other related and drive-by changes:
- added label and scoped label expressions to allow the user to call e.g. `get_doc(person)` directly without having to introduce a type variable;
  - consequently, allowed bare type labels in fetch as they are allowable expressions and it did not make sense to explicitly ban them;
- refactored annotation categories, introduced type and capability specific annotation categories (autogenerated with macros);
- implemented non-grammar builtin typeql functions for metadata (doc and meta) retrieval;
- fixed BDD test paths for cargo.


concept/type_/type_manager/type_reader.rs
```diff
@@ -833,6 +849,7 @@ impl TypeReader {
                     }
                 }
             }
+            CapabilityKind::Sub => todo!(),
```
flyingsilverfin: Here^^

concept/type_/attribute_type.rs
```diff
@@ -472,24 +488,26 @@ impl AttributeType {
     }
 }
 
-#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[derive(Debug, Clone, Eq, PartialEq, Hash, FromAnnotation!, has_annotation_category!)]
```
flyingsilverfin: Deriving a macro ? with ! ?

concept/type_/constraint.rs
```diff
@@ -78,6 +81,8 @@ impl fmt::Debug for ConstraintCategory {
             Self::Regex => write!(f, "{}", AnnotationCategory::Regex),
             Self::Range => write!(f, "{}", AnnotationCategory::Range),
             Self::Values => write!(f, "{}", AnnotationCategory::Values),
+            Self::Doc => write!(f, "{}", AnnotationCategory::Doc),
+            Self::Meta => write!(f, "{}", AnnotationCategory::Meta(String::from("?"))), // FIXME
```
flyingsilverfin: todo

concept/type_/attribute_type.rs
dmitrii-ubskii: Pretty cool huh? This is `macro_rules_attribute::derive`.

concept/type_/annotation.rs
```diff
@@ -836,18 +1024,86 @@ impl TypeEdgePropertyEncoding for AnnotationValues {
...
+        UnsupportedRelationTypeAnnotation(2, "Annotation '{category}' is not supported for relation types.", category: AnnotationCategory),
+        UnsupportedAttributeTypeAnnotation(3, "Annotation '{category}' is not supported for attribute types.", category: AnnotationCategory),
+        UnsupportedRoleTypeAnnotation(4, "Annotation '{category}' is not supported for role types.", category: AnnotationCategory),
+        UnsupportedRelatesAnnotation(5, "Annotation '{category}' is not supported for relates.", category: AnnotationCategory),
+        UnsupportedPlaysAnnotation(6, "Annotation '{category}' is not supported for plays.", category: AnnotationCategory),
+        UnsupportedOwnsAnnotation(7, "Annotation '{category}' is not supported for owns.", category: AnnotationCategory),
+        UnsupportedAliasAnnotation(8, "Annotation '{category}' is not supported for alias.", category: AnnotationCategory),
+        UnsupportedSubAnnotation(9, "Annotation '{category}' is not supported for sub.", category: AnnotationCategory),
+        UnsupportedValueTypeAnnotation(10, "Annotation '{category}' is not supported for value types.", category: AnnotationCategory),
```
dmitrii-ubskii, line 1106: Renamed so that the derive macro is easier to implement. Now it's just `FooAnnotation` => `paste!([<Unsupported FooAnnotation>])`

query/analyse.rs
dmitrii-ubskii, line 1: This file used to be unformatted because it crashed rustfmt. It couldn't handle comments between alternatives in a match pattern.

query/analyse.rs
```diff
@@ -281,83 +281,104 @@ fn build_fetch_entries_annotations<Snapshot: ReadableSnapshot>(
...
+                AnnotatedFetchSome::ListAttributesAsList(_var, attribute_type)
+                | AnnotatedFetchSome::ListAttributesFromList(_var, attribute_type)
+                | AnnotatedFetchSome::SingleAttribute(_var, attribute_type) => {
+                    // TODO: Verify lists can use the same code as SingleAttribute
+                    // TODO: Refine based on owner?
+                    let subtypes = attribute_type.get_subtypes(snapshot, type_manager)?;
+                    let attribute_types = chain!([*attribute_type].into_iter(), subtypes.iter().copied());
+                    FetchStructureAnnotations::Leaf(build_leaf_annotations(snapshot, type_manager, attribute_types)?)
+                }
+                AnnotatedFetchSome::Label(_ty) => FetchStructureAnnotations::Leaf(BTreeSet::default()),
```
dmitrii-ubskii, line 346: New addition. Empty type list because `fetch` analysis only supports value types.

executor/tests/compile_execute.rs
```diff
@@ -159,7 +152,7 @@ fn test_has_planning_traversal() {
     )
     .unwrap();
 
-    let context = ExecutionContext::new(snapshot, thing_manager, Arc::default());
+    let context = ExecutionContext::new(snapshot, thing_manager, Arc::default(), Arc::default());
```
krishnangovindraj, line 163: I know the default already exists, but I'm a bit iffy about defaults when they're not what you use in the normal flow. Here, (during bootup) the `FunctionManager` has to be initialised with a `DefinitionKeyGenerator` that is aware of the latest entry.
And it's only safe to use `default` in tests?

executor/tests/execute_function.rs
```diff
@@ -95,7 +95,7 @@ fn setup_common(schema: &str) -> Context {
     setup_concept_storage(&mut storage);
 
     let (type_manager, thing_manager) = load_managers(storage.clone(), None);
-    let function_manager = FunctionManager::new(Arc::new(DefinitionKeyGenerator::new()), None);
+    let function_manager = Arc::new(FunctionManager::new(Arc::new(DefinitionKeyGenerator::new()), None));
```
krishnangovindraj: Can default this too for consistency.

ir/pipeline/function.rs
```diff
@@ -44,8 +45,18 @@ impl Function {
         argument_labels: Option<Vec<NamedTypeAny>>,
         output: Option<Output>,
         function_body: FunctionBody,
+        annotations: Vec<FunctionAnnotation>,
```
krishnangovindraj, line 48: ~I assume this is safe to cache and just never implemented because it predates the query-cache.~
Ooooh `@annotation`, not `TypeAnnotations`

ir/pipeline/mod.rs
```diff
@@ -130,6 +131,35 @@ typedb_error! {
             return_: ReturnStatement,
             mismatch_index: usize,
         ),
+        IllegalAnnotation(
+            14,
+            "TODO",
```
krishnangovindraj: TODO

concept/type_/type_manager/type_cache/kind_cache.rs
```diff
@@ -314,22 +316,119 @@ impl RelatesCache {
     }
 }
 
-impl<T: KindAPI> CommonTypeCache<T> {
-    fn create<Snapshot>(snapshot: &Snapshot, type_: T) -> CommonTypeCache<T>
+impl CommonTypeCache<EntityType> {
```
krishnangovindraj: Aww is there no way to keep this generic? :(

compiler/annotation/pipeline.rs
```diff
@@ -336,12 +336,35 @@ fn complete_block_annotations_with_value_types(
...
+                        actual: actual.clone(),
+                        source_span: function_call_binding.source_span(),
+                    });
+                }
+            }
+        }
+    }
+
+    block_annotations.set_value_types_of(conjunction, value_types_in_conjunction);
+
```
krishnangovindraj, line 410: I guess this is also in the other PR?

compiler/annotation/fetch.rs
```diff
@@ -40,6 +40,8 @@ pub enum AnnotatedFetchSome {
     SingleAttribute(Variable, AttributeType),
     SingleFunction(AnnotatedFunction),
 
+    Label(Type),
```
krishnangovindraj, line 43: Can this be compiled down to a constant rather than needing a runtime call? Or is it parametrised away? How is this handled for a type variable?

concept/type_/type_manager/type_reader.rs
```diff
@@ -833,6 +849,7 @@ impl TypeReader {
                     }
                 }
             }
+            CapabilityKind::Sub => (),
```
flyingsilverfin, line 852: We're skipping right? Is fine for me

concept/type_/annotation.rs
```diff
@@ -511,22 +526,68 @@ impl fmt::Display for AnnotationValues {
...
-            Self::Cardinality(_) => AnnotationCategory::Cardinality,
-            Self::Regex(_) => AnnotationCategory::Regex,
-            Self::Cascade(_) => AnnotationCategory::Cascade,
-            Self::Range(_) => AnnotationCategory::Range,
-            Self::Values(_) => AnnotationCategory::Values,
-        }
+#[derive(Debug, Default, Clone, Eq, PartialEq, Hash)]
+pub struct AnnotationDoc {
+    // ##########################################################################
+    // ###### WARNING: any changes here may break backwards compatibility! ######
```
flyingsilverfin, line 532: How's that? I don't see serialize derived

encoding/graph/type_/property.rs
```diff
@@ -138,39 +142,34 @@ pub trait TypeVertexPropertyEncoding {
 pub struct TypeEdgeProperty {
     edge: TypeEdge,
     infix: Infix,
-    suffix: Option<ByteArray<0>>,
+    suffix: ByteArray<0>,
```
flyingsilverfin, line 145: Yeah these work good generalization!

concept/type_/role_type.rs
```diff
@@ -312,32 +311,16 @@ impl fmt::Display for RoleType {
...
+        match *self {}
     }
-}
 
-impl From<RoleTypeAnnotation> for Annotation {
-    fn from(_anno: RoleTypeAnnotation) -> Self {
-        unreachable!("RoleTypes do not have annotations!")
+    fn category(&self) -> AnnotationCategory {
+        match *self {}
     }
```
krishnangovindraj, line 324: How does this compile?

ir/translation/constraints.rs
```diff
@@ -653,10 +641,21 @@ pub(super) fn split_out_inline_expressions(
...
+                let as_label = register_type_label(constraints, label)?;
+                constraints.add_label(type_variable, as_label)?;
+                Ok(type_variable)
+            }
+            typeql::Expression::ScopedLabel(scoped_label) => {
+                let type_variable = constraints.create_anonymous_variable(scoped_label.span())?;
+                let as_label = register_type_scoped_label(constraints, scoped_label)?;
+                constraints.add_label(type_variable, as_label)?;
+                Ok(type_variable)
+            }
```
krishnangovindraj, line 636: An inline label gets outlined to a variable with a label constraint?
Nice.

ir/translation/function.rs
```diff
@@ -323,3 +329,73 @@ fn check_consistent_return<T>(
...
+        }
+    }
+
+    fn category(&self) -> AnnotationCategory {
+        match self {
+            FunctionAnnotation::Doc(annotation_doc) => AnnotationCategory::Doc,
+            FunctionAnnotation::Meta(annotation_meta) => AnnotationCategory::Meta(annotation_meta.key().to_owned()),
+        }
+    }
+}
```
krishnangovindraj, line 397: Should the `has_category` function be `is_category`?
Is the implementation of `has_category` not always `self.category == category`?

executor/read/immediate_executor.rs
krishnangovindraj, line 1: This feels like a massive leak of implementation details into what is otherwise a very abstract executor. 
Most logic like this lives in the thing/type managers. 
I'd say the executor package is the wrong place for it, though I could also live with it just being in its own file.

concept/type_/sub.rs
```diff
@@ -24,6 +47,112 @@ impl<T: TypeAPI> Sub<T> {
     }
 }
 
+macro_rules! impl_capability_for_sub {
```
krishnangovindraj, line 51: General question, do you think there's a way to do this with generics rather than macros?

concept/type_/sub.rs
```diff
@@ -24,6 +47,112 @@ impl<T: TypeAPI> Sub<T> {
...
+                Self { subtype, supertype }
+            }
+
+            fn object(&self) -> Self::ObjectType {
+                self.subtype()
+            }
+
+            fn interface(&self) -> Self::InterfaceType {
+                self.supertype()
+            }
```
krishnangovindraj, line 102: Do we see sub as an interface we implement? I can see where it's coming from but it feels a bit strange.

concept/type_/sub.rs
```diff
@@ -24,6 +47,112 @@ impl<T: TypeAPI> Sub<T> {
...
+                let sub_annotation = SubAnnotationCategory::try_from(annotation_category)
+                    .map_err(|typedb_source| ConceptWriteError::Annotation { typedb_source })?;
+                match sub_annotation {
+                    SubAnnotationCategory::Doc => paste::paste!(type_manager.[<unset_sub_ $t:snake _annotation_doc>](snapshot, *self)),
+                    SubAnnotationCategory::Meta(meta) => paste::paste!(type_manager.[<unset_sub_ $t:snake _annotation_meta>](snapshot, *self, meta)),
+                }
+            }
+        }
+
+        impl Capability for Sub<$t> {
```
krishnangovindraj, line 83: Actually, do we really need sub to be a capability? All the noops point to it not being one. Maybe we just need a different trait called `Documentable`?

concept/type_/type_manager.rs
```diff
@@ -3447,6 +3467,345 @@ impl TypeManager {
         self.unset_capability_annotation(snapshot, owns, annotation_category)
     }
 
+    pub(crate) fn set_entity_type_annotation_doc(
```
krishnangovindraj, line 3472: I thought the TypeManager used generics rather than macros, but I guess I was wrong. 
Would you not want to macro these for consistency?

query/definable_status.rs
```diff
@@ -68,30 +67,6 @@ macro_rules! return_exists_different_if_some {
...
-        )*
-    }
-}
-
-get_type_status! {
-    fn get_entity_type_status() -> EntityType = get_entity_type;
-    fn get_relation_type_status() -> RelationType = get_relation_type;
-    fn get_attribute_type_status() -> AttributeType = get_attribute_type;
-}
-
```
krishnangovindraj, line 94: Unused?

query/undefine.rs
```diff
@@ -267,18 +270,59 @@ fn undefine_capability_annotation(
...
+                    match (subtype, supertype) {
+                        $((TypeEnum::$T(subtype), TypeEnum::$T(supertype)) => {
+                            let declared_super = subtype
+                                .get_supertype(snapshot, type_manager)
+                                .map_err(|typedb_source| UndefineError::UnexpectedConceptRead { typedb_source })?;
+                            if declared_super != Some(supertype) {
+                                return Err(resolution_error);
+                            }
+                            let sub = Sub::<paste::paste!([<$T Type>])>::new(subtype, supertype);
+                            check_can_and_need_undefine_capability_annotation(
```
krishnangovindraj, line 307: It's a little weird that this one returns `Result<bool,_>` but it's always true and never used.

krishnangovindraj (commented):
Left suggestions that I don't feel strongly enough about to request changes.
Mainly:
1. Whether meta should be it's own class of keys separate from the existing `Property`s (from Discord)
2. Moving the builtin specific logic out from the executor package so we can keep the executor as abstract as possible.

concept/type_/role_type.rs
dmitrii-ubskii: `RoleTypeAnnotation` is an empty enum. There is no value of type `Self`, so `.category()` could never have been called in the first place.

This `match` creates an `AnnotationCategory` for every variant of `RoleTypeAnnotation` (all zero of them) and that's perfectly acceptable.

query/undefine.rs
dmitrii-ubskii: It sure is. Fixed.

query/definable_status.rs
dmitrii-ubskii: Unused.

compiler/annotation/fetch.rs
dmitrii-ubskii: It already doesn't need a runtime call, this type is resolved during type inference.

ir/translation/function.rs
dmitrii-ubskii: It isn't, `self.category()` would have to allocate for `Meta` to keep the `AnnotationCategory: 'static`. But it should produce the same result.

As for the name, the annotation isn't a category, it has a category imo.

concept/type_/sub.rs
dmitrii-ubskii: Not without rewriting all the way down to type manager and type cache.

concept/type_/sub.rs
dmitrii-ubskii: I'm treating `Capability` as ~`Edge` here, mostly because of all the existing machinery around the capability annotations.

concept/type_/type_manager.rs
dmitrii-ubskii: Maybe in a separate PR, this follows the existing structure.

concept/type_/type_manager/type_cache/kind_cache.rs
dmitrii-ubskii: Only if we pretend `Sub<RoleType>` can have annotations. Which, might as well. Added.

ir/pattern/expression.rs
```diff
@@ -320,30 +320,176 @@ impl fmt::Display for BuiltinValueFunctionID {
...
-                vec![(VariableCategory::Value, VariableOptionality::Required)],
-                false,
-            ),
-            Self::Label => FunctionSignature::new(
-                FunctionID::Builtin(self),
-                vec![VariableCategory::Type],
-                vec![(VariableCategory::Value, VariableOptionality::Required)],
-                false,
-            ),
+            Self::Iid => function_signature!((Thing) -> Value),
```
flyingsilverfin, line 459: Will this support optionality as well?

dmitrii-ubskii: It can with a tweak to the macro.

flyingsilverfin (commented):
Worth outlining in the PR description the new type of 'builtin' function architecture we have that is not a TypeQL defined function and not a user defined function and not a value expression function

tests/behaviour/concept/thing/entity.rs
```diff
@@ -17,7 +17,7 @@ async fn test_entity() {
     let path = "../typedb_behaviour+/concept/thing/entity.feature";
 
     #[cfg(not(feature = "bazel"))]
-    let path = "bazel-typedb/external/typedb_behaviour/concept/thing/entity.feature";
+    let path = "bazel-typedb/external/typedb_behaviour+/concept/thing/entity.feature";
```
flyingsilverfin, line 20: What does this do?

function/function_manager.rs
```diff
@@ -226,10 +271,71 @@ impl FunctionManager {
     pub fn get_functions_syntax(&self, snapshot: &impl ReadableSnapshot) -> Result<String, FunctionReadError> {
         let mut syntax = String::new();
         for function in FunctionReader::get_functions_all(snapshot)? {
-            write!(&mut syntax, "\n{}", function.parsed.unparsed).map_err(|err| err.into())?;
+            write!(&mut syntax, "\n{}", function.parsed.unparsed)?;
         }
         Ok(syntax)
     }
+
+    pub fn get_function_annotations(
```
flyingsilverfin, line 279: Ok just checking but i believe this does match the flow of TypeManager right? eg. TypeManager delegates to either the TypeCache or TypeReader which gets annotations for a type via a function that looks a lot like this. Since we don't have separate cache or reader here, this looks right to me!

flyingsilverfin (approved):
Approved once you finish krishnan's more thorough comments!

tests/behaviour/concept/thing/entity.rs
dmitrii-ubskii: Upgraded Bazel puts a `+` in the repo names in `external/`. Only paths used by tests run through Bazel were updated with the upgrade, not cargo.

function/function_manager.rs
dmitrii-ubskii: Yep, that was the idea.

concept/type_/type_manager/type_reader.rs
dmitrii-ubskii: Yes, `sub` doesn't imply any constraints.

concept/type_/annotation.rs
dmitrii-ubskii: We put this warning on every annotation that carries a value. It's less likely when the value is `String`, granted.

krishnangovindraj (approved):
nice

PR #7792, opened 2026-05-01
Improved query profiler

flyingsilverfin:

We implement a real nested query profiling structure, which correctly captures sub-pattern timings and executions and eliminates redundant printing.

## Implementation

- Rearchitect QueryProfile to allow nesting a StageProfile
- A stage profile contains a PatternProfile
- A pattern profile is made up of Substep profiles
- A substep is either a Step, a Query or another PatternProfile


query/tests/BUILD
```diff
@@ -58,13 +58,21 @@ rust_test(
     deps = deps,
 )
 
+rust_test(
+    name = "test_query_profile",
```
flyingsilverfin, line 62: Verify that logging a complex query profile doesn't panic or anything

flyingsilverfin:
Note: not meant to be perfect, just better - we still can't profile modifier stages properly (don't time their operations i believe), or support multiple reduce stages (they over write each other since only 1 can assign '0' as the ID, which we hardcode

krishnangovindraj (changes requested):
Could we move `extend_or_get` to consistently be at Executor creation? 
It's in the hot loop for some stages like insert

resource/profile.rs
```diff
@@ -693,18 +853,38 @@ impl StepProfile {
...
-impl fmt::Display for StepProfileData {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+impl IndentDisplay for StepProfileData {
+    fn indent_fmt(&self, f: &mut Formatter<'_>, indent: usize) -> fmt::Result {
         let rows = self.rows.load(Ordering::Relaxed);
         let micros = Duration::from_nanos(self.nanos.load(Ordering::Relaxed)).as_micros();
         let micros_per_row: f64 = micros as f64 / rows as f64;
         // TODO: print storage ops
-        write!(
+        // write_indent(f, indent + 1)?;
```
krishnangovindraj: ded

resource/profile.rs
```diff
@@ -610,42 +630,182 @@ impl Display for EncodingProfile {
...
+
+    fn total_nanos(&self) -> u64 {
+        self.substeps.read().unwrap().iter().map(SubstepProfile::total_nanos).sum()
+    }
+}
+
+impl IndentDisplay for PatternProfile {
+    fn indent_fmt(&self, f: &mut Formatter<'_>, indent: usize) -> fmt::Result {
+        if let Some(description) = self.description.as_deref() {
+            // write_indent(f, indent)?;
```
krishnangovindraj, line 813: Does this need to be cleaned up?

query/tests/query_profile.rs
```diff
@@ -0,0 +1,141 @@
...
+    let (iterator, context) = pipeline.into_rows_iterator(ExecutionInterrupt::new_uninterruptible()).unwrap();
+    // Drain the iterator so the executor populates step counters before we render the profile.
+    let _: Vec<_> =
+        iterator.map_static(|row| row.map(|row| row.into_owned()).map_err(|err| err.clone())).into_iter().collect();
+
+    assert!(context.profile.is_enabled(), "expected profile to be enabled under TRACE-level tracing");
+    // The actual test: format the whole profile tree via Display (which delegates to
+    // IndentDisplay::indent_fmt) and ensure it neither panics nor produces empty output.
+    let rendered = format!("{}", context.profile);
+    println!("{}", rendered);
```
krishnangovindraj: Maybe you can strengthen the test by taking this output from one run,  and doing some sort of check on the structure based on it?

flyingsilverfin: yeah i suppose i was mainly wanting to protect us from crashing the users's server in the printing of it if they enable it

flyingsilverfin: but i can at least check the length of the profile's stages stages or something

Issue #7790, opened 2026-05-01
Potential for stale query plans persisting due to race condition in query cache


dmitrii-ubskii:
## Description

Sequence of events:
- read transaction R opens, holds a copy of statistics at time of opening,
- database statistics updates,
- the transaction R receives a query, compiles it, and stores it into the cache.

If the plan would be evicted by the statistics update, this means a query plan using stale statistics erroneously remains in the cache.

The stale plan will persist until one of the following:
- the next statistics update (~ the next database write),
- or the plan is evicted from the cache for being unused.

krishnangovindraj:
This is different from #7787 because it's the case where statistics go out of date due to a write, rather than the cache invalidated because of a schema transaction?

krishnangovindraj:
#7794 says this is fixed.

PR #7793, opened 2026-05-02
Swap out QueryCache on schema commit

### LINKED ISSUE #7787: Likely race between QueryCache reset & write
## Description
Schema commits reset the query cache.
A concurrent read query (writes are locked out) opened may experience a cache-miss , and write an entry to the query cache based on an outdated schema. This can lead to wrong results.
## Environment

1. TypeDB distribution: Core/Enterprise/Cloud
2. TypeDB version: 3.10.3
3. Environment: All
4. Client and version:
5. Other details:

## Reproducible Steps
(Hard)

## Proposed solution
- Swap out the QueryCache altogether on schema commit.
- Store statistics in the QueryCache and call may_evict on insert.

## Additional information

krishnangovindraj:
Prefer #7794  as it fixes more issues at once.

Swaps out QueryCache on schema commit, preventing read pipelines based on outdated schema from being cached (See #7787)

## Implementation
Swaps out `database.query_cache` instead of calling `query_cache.invalidate_all`.
Fixes #7787 but not  #7790

### DIFF
--- a/database/database.rs
+++ b/database/database.rs
@@ -41,10 +41,7 @@ use error::typedb_error;
 use fail_point::{UNFINISHED_CHECKPOINT, fail_point};
 use function::{FunctionError, function_cache::FunctionCache};
 use query::query_cache::QueryCache;
-use resource::{
-    constants::database::{CHECKPOINT_INTERVAL, STATISTICS_UPDATE_INTERVAL},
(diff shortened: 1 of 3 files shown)


krishnangovindraj:
Prefer #7794

PR #7794, opened 2026-05-02
QueryCache tracks statistics & schema commit to avoid caching outdated plans

### LINKED ISSUE #7787: Likely race between QueryCache reset & write

krishnangovindraj:
Updates the QueryCache to explicitly track validity requirements and validate insert calls. By tracking the sequence_number of the latest schema commit, we avoid #7787. By tracking statistics, we avoid #7790 

## Implementation
Fixes #7787, #7790 
Statistics track the latest synchronised statistics. `may_insert` checks the candidate against (1) The sequence number to ensure it is based on the current schema (2) the latest statistics to ensure it was not built on outdated statistics.

### DIFF
--- a/query/query_cache.rs
+++ b/query/query_cache.rs
@@ -70,7 +70,7 @@ impl QueryCache {
         let read_lock = self.validity_requirements.read().unwrap();
         let ValidityRequirements { latest_schema_commit, latest_statistics } = &*read_lock;
         let may_insert = latest_schema_commit
-            .map_or(true, |min_sequence_number| statistics_sequence_number >= min_sequence_number)
+            .map_or(true, |latest_schema_commit_number| statistics_sequence_number >= latest_schema_commit_number)
             && latest_statistics
                 .as_ref()
                 .map_or(true, |stats| !is_pipeline_type_populations_outdated(&stats, &pipeline));
--- a/query/query_cache.rs
+++ b/query/query_cache.rs
@@ -6,7 +6,7 @@
 
 use std::{
     hash::{DefaultHasher, Hash, Hasher},
-    sync::Arc,
+    sync::{Arc, RwLock},
 };
 
 use answer::Type;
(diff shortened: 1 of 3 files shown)


query/query_cache.rs
```diff
@@ -48,44 +58,42 @@ impl QueryCache {
...
+        let ValidityRequirements { latest_schema_commit, latest_statistics } = &*read_lock;
+        let may_insert = latest_schema_commit
+            .map_or(true, |min_sequence_number| read_sequence_number >= min_sequence_number)
+            && latest_statistics
+                .as_ref()
+                .map_or(true, |stats| !is_pipeline_type_populations_outdated(&stats, &pipeline));
+        if may_insert {
+            self.cache.insert(key, pipeline);
+        }
+        drop(read_lock);
```
krishnangovindraj, line 80: We must hold the lock till after the insert to avoid validating `ValidityRequirements` changing between the time we read it and insert.

query/query_cache.rs
```diff
@@ -48,44 +58,42 @@ impl QueryCache {
...
     }
 
-    pub fn may_evict(&self, new_statistics: &Statistics) {
-        let new_statistics = new_statistics.clone(); // it's either this or clone the entire cache
+    pub fn set_statistics_and_invalidate_outdated(&self, new_statistics: Arc<Statistics>) {
+        let mut write_lock = self.validity_requirements.write().unwrap();
+        (*write_lock).latest_statistics = Some(new_statistics.clone());
+        drop(write_lock);
```
krishnangovindraj, line 86: We can drop it here, since ValidityRequirements is for incoming inserts.

query/query_cache.rs
```diff
@@ -97,6 +105,26 @@ impl Default for QueryCache {
     }
 }
 
+fn is_pipeline_type_populations_outdated(statistics: &Statistics, pipeline: &ExecutablePipeline) -> bool {
```
krishnangovindraj, line 108: Move from inline closure so we can reuse.

query/query_cache.rs
```diff
@@ -48,44 +58,42 @@ impl QueryCache {
...
         stages: Arc<Vec<TranslatedStage>>,
         fetch: Arc<Option<FetchObject>>,
         pipeline: ExecutablePipeline,
     ) {
         let key = IRQuery::new(preamble, stages, fetch);
-        self.cache.insert(key, pipeline);
+        let read_lock = self.validity_requirements.read().unwrap();
+        let ValidityRequirements { latest_schema_commit, latest_statistics } = &*read_lock;
+        let may_insert = latest_schema_commit
+            .map_or(true, |min_sequence_number| statistics_sequence_number >= min_sequence_number)
```
flyingsilverfin: naming: `min_sequence_number` is actually `latest_schema_commit_number` right? Should reference `schema` in some form - makes it easier to understand what `map_or` does if i dont have it off the top of my head too

flyingsilverfin (approved):
Approved w 1 comment

PR #7797, opened 2026-05-06
Expose rocksdb memory configuration

flyingsilverfin:

We expose a pair of new memory tuning parameters: RocksDB cache size (`storage.rocksdb.cache-size`) and RocksDB write buffers limit (`storage.rocksdb.write-buffers-limit`). Tuning these gives a more predictable memory limit, and the limits are **shared across open databases**:

```
storage:
    data-directory: "data"
    rocksdb:
        cache-size: 1gb
        write-buffers-limit: 512mb
```

Configurable with integer `kb`, `mb`, or `gb` units.

By default, we will set 1gb of cache and 512mb of write buffers. As these limits are approached:
1) the cache will evict least recently used data blocks
2) the write buffers will more eagerly flush to disk, even if not quite full.

If the write buffers limit can't be maintained, then writes will be throttled.

We expect that currently, with additional overheads, TypeDB will use around 2-2.5gb of RAM at the steady state.

Note that the cache size limit is a _soft_ limit: TypeDB always caches RocksDB indexes and bloom filters, so eg. a 1GB cache limit maybe breached if more than 1GB of indexes and blooms exist.

## Implementation

- Introduce new user-configurable `RocksDBConfig` option in the config file
- Leverage RocksDB `WriteBufferManager` to make a predictable memory profile for write buffers
- Share the write buffer manager and the Cache across all open databases on the server


encoding/encoding.rs
```diff
@@ -97,11 +99,17 @@ impl KeyspaceSet for EncodingKeyspace {
 
         options.create_if_missing(true);
         options.create_missing_column_families(true);
-        options.set_max_background_jobs(10);
+        if let Ok(parallelism) = std::thread::available_parallelism() {
```
flyingsilverfin, line 102: New: dynamically set number of possible flush + compaction jobs == cpu count

common/options/byte_size.rs
krishnangovindraj, line 1: overkill?
We could just have `rocksdb-cache-size-mb`

server/config.yml
```diff
@@ -24,6 +24,9 @@ server:
 
 storage:
     data-directory: "data"
+    rocksdb:
+        cache-size: 1gb
```
flyingsilverfin, line 28: Option: `global-cache-size` to make it clear it's for all databases. Also `global-write-buffers-limit`

storage/keyspace/rocks_resources.rs
```diff
@@ -0,0 +1,55 @@
...
+        self.write_buffers_limit_bytes
+    }
+}
+
+impl Debug for RocksResources {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Rocks LRU Cache size (shared by all databases): {}. \
+             Rocks write buffers size limit (shared by all databases): {}",
```
dmitrii-ubskii: Is this meant to be on one line? That's how it'll print.
Also missing period on second line.

storage/keyspace/rocks_resources.rs
dmitrii-ubskii, line 1: Should be `rocks_resources`

database/database_manager.rs
```diff
@@ -37,21 +39,31 @@ impl DatabaseManager {
...
+        rocksdb_cache_size: ByteSize,
+        rocksdb_write_buffers_limit: ByteSize,
     ) -> Result<Arc<Self>, DatabaseOpenError> {
         let data_directory = data_directory.as_ref().to_owned();
         let import_directory = data_directory.join(Self::IMPORT_DIRECTORY_NAME);
 
-        let databases =
-            RwLock::new(Self::initialise_databases(&data_directory, &import_directory, &diagnostics_manager)?);
+        let rocks_resources =
+            Arc::new(RocksResources::new(rocksdb_cache_size.as_usize(), rocksdb_write_buffers_limit.as_usize()));
```
dmitrii-ubskii: Any reason `RocksResources::new` can't or shouldn't accept `ByteSize` directly?

storage/keyspace/storage_resources.rs
```diff
@@ -0,0 +1,55 @@
...
+        self.cache_limit_bytes
+    }
+
+    pub fn write_buffers_limit_bytes(&self) -> usize {
+        self.write_buffers_limit_bytes
+    }
+}
+
+impl Debug for RocksResources {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
```
dmitrii-ubskii: Our convention is to only `use std::fmt;`:
```suggestion
impl fmt::Debug for RocksResources {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```

dmitrii-ubskii (approved):
Approved with notes.

PR #7798, opened 2026-05-06
Print TypeDB Studio link and Console command on server startup

flyingsilverfin:

Adds clickable/runnable connect hints to the server startup output:
```
  To connect with TypeDB Studio, open: https://studio.typedb.com/connect?address=http://localhost:8000&username=admin
  To connect with TypeDB Console, run: typedb console --address localhost:1729 --tls-disabled --username admin`
```

Both substitute `localhost` when the listener binds to an unspecified address (0.0.0.0 / [::]). The Studio scheme and Console invocation follow the encryption setting; the Studio link is omitted if the HTTP endpoint is disabled.

## Implementation

- Add new printout on bootup to connect with studio or console

### DIFF
--- a/server/lib.rs
+++ b/server/lib.rs
@@ -463,11 +463,52 @@ impl Server {
             println!("  Drivers must be configured to connect *without TLS*.");
         }
 
+        let grpc_connect_address = server_status.grpc_advertise_address();
+        let http_connect_address = server_status.http_advertise_address();
+        if grpc_connect_address.is_some() || http_connect_address.is_some() {
+            println!("To connect:");
+            if let Some(http_connect_address) = http_connect_address {
+                println!("  Studio:  {}", Self::studio_connect_link(http_connect_address, encryption_config));
+            }
+            if let Some(grpc_connect_address) = grpc_connect_address {
+                println!("  Console: {}", Self::console_connect_command(grpc_connect_address, encryption_config));
+            }
+        }
+
         println!();
     }
 
     pub fn print_ready() {
         info!("\nReady!");
+    }
+
+    fn studio_connect_link(http_advertise_address: &str, encryption_config: &EncryptionConfig) -> String {
+        let scheme = if encryption_config.enabled { "https" } else { "http" };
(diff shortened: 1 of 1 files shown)


server/lib.rs
```diff
@@ -463,13 +463,54 @@ impl Server {
...
+        let scheme = if encryption_config.enabled { "https" } else { "http" };
+        let address = format!("{scheme}://{}", Self::reachable_address(http_advertise_address));
+        format!("https://studio.typedb.com/connect?address={address}&username=admin")
+    }
+
+    fn reachable_address(advertise_address: &str) -> String {
+        if let Ok(socket_addr) = advertise_address.parse::<SocketAddr>() {
+            if socket_addr.ip().is_unspecified() {
+                return format!("localhost:{}", socket_addr.port());
+            }
```
krishnangovindraj: I guess that's always right for server output.

flyingsilverfin: should be i think, although we also have the advertise address now so need to think about it some more

server/lib.rs
```diff
@@ -463,13 +463,45 @@ impl Server {
...
         println!();
     }
 
     pub fn print_ready() {
         info!("\nReady!");
     }
 
+    fn studio_connect_link(http_advertise_address: &str, encryption_config: &EncryptionConfig) -> String {
+        let scheme = if encryption_config.enabled { "https" } else { "http" };
+        format!("https://studio.typedb.com/connect?address={scheme}://{http_advertise_address}&username=admin")
```
farost, line 487: I feel like `https://studio.typedb.com/` might be a constant itself

farost (commented):
Love it. One stylistic comment + a question: the PR description mentions transforming `0.0.0.0` into `localhost`. Is it still the case? I think the advertise address must return `127.0.0.1`

farost (approved):

server/lib.rs
flyingsilverfin: will do rolled into later

PR #7800, opened 2026-05-06
Escalate concurrent key and unique writes to transaction failure

flyingsilverfin:

TypeDB failed to validate race conditions between transactions writing attributes covered by `@key` or `@unique` ownerships. Now, a concurrent transaction creating an ownership of a keyed or uniquely-owned attribute will cause one of them to fail with an isolation conflict error, like:

```
Commit in database 'test' failed with isolation conflict: Transaction uses a lock held by a concurrent commit
```

## Implementation

- Update the byte array created in the storage-layer locks that cause subsequent transactions to fail
- Implement a series of isolation tests for transactions, including update-delete, key races, unique races, cardinality conflicts
- Add tests for attribute concept generation based on hashed values to ensure that concurrently committed attribute values that hash to the same prefix + hash also race and cause one transaction to fail (this was succeeding before)

### DIFF
--- a/concept/thing/thing_manager.rs
+++ b/concept/thing/thing_manager.rs
@@ -1809,6 +1809,7 @@ impl ThingManager {
         self.cleanup_relations(snapshot, storage_counters.clone()).map_err(|err| vec![*err])?;
         self.cleanup_attributes(snapshot, storage_counters.clone()).map_err(|err| vec![*err])?;
 
+        println!("Creating commit locks next");
         match self.create_commit_locks(snapshot, storage_counters) {
             Ok(_) => Ok(()),
             Err(error) => Err(vec![ConceptWriteError::ConceptRead { typedb_source: error }]),
@@ -1822,6 +1823,7 @@ impl ThingManager {
     ) -> Result<(), Box<ConceptReadError>> {
         // TODO: Should not collect here (iterate_writes() already copies)
         for (key, _write) in snapshot.iterate_writes().collect_vec() {
+            println!("Creating locks");
             if ThingEdgeHas::is_has(&key) {
                 let has = ThingEdgeHas::decode(Bytes::Reference(key.bytes()));
                 let object = Object::new(has.from());
@@ -1849,30 +1851,31 @@ impl ThingManager {
         snapshot: &mut impl WritableSnapshot,
         owner: &Object,
         attribute: Attribute,
-        storage_counters: StorageCounters,
+        _storage_counters: StorageCounters,
     ) -> Result<(), Box<ConceptReadError>> {
         let unique_constraint_opt = owner.type_().get_owned_attribute_type_constraint_unique(
             snapshot,
             self.type_manager(),
             attribute.type_(),
         )?;
+        dbg!("Unique constraint", &unique_constraint_opt);
         if let Some(unique_constraint) = unique_constraint_opt {
             let attribute_key = attribute.vertex();
-            let attribute_value = snapshot
-                .get_last_existing::<BUFFER_VALUE_INLINE>(
-                    attribute_key.into_storage_key().as_reference(),
-                    storage_counters,
-                )
-                .map_err(|error| Box::new(ConceptReadError::SnapshotGet { source: error }))?
-                .ok_or(ConceptReadError::InternalMissingAttributeValue {})?;
+            // let attribute_value = snapshot
+            //     .get_last_existing::<BUFFER_VALUE_INLINE>(
+            //         attribute_key.into_storage_key().as_reference(),
+            //         storage_counters,
+            //     )
+            //     .map_err(|error| Box::new(ConceptReadError::SnapshotGet { source: error }))?
+            //     .ok_or(ConceptReadError::InternalMissingAttributeValue {})?;
 
+            // Note: trade much smaller locks (eg. not entire string) for rare concurrent lock conflicts when
+            //       two different values has to the same hash
             let lock_key = create_custom_lock_key(
                 [
                     &Infix::PropertyAnnotationUnique.infix_id().bytes(),
                     &*unique_constraint.source().attribute().vertex().to_bytes(),
-                    attribute_key.attribute_id().bytes(),
-                    &attribute_value,
-                    &*owner.vertex().to_bytes(),
+                    attribute_key.attribute_id().deterministic_bytes(),
                     &*unique_constraint.source().owner().vertex().to_bytes(),
                 ]
                 .into_iter(),
--- a/concept/thing/thing_manager.rs
+++ b/concept/thing/thing_manager.rs
@@ -1809,7 +1809,6 @@ impl ThingManager {
         self.cleanup_relations(snapshot, storage_counters.clone()).map_err(|err| vec![*err])?;
         self.cleanup_attributes(snapshot, storage_counters.clone()).map_err(|err| vec![*err])?;
 
-        println!("Creating commit locks next");
         match self.create_commit_locks(snapshot, storage_counters) {
             Ok(_) => Ok(()),
             Err(error) => Err(vec![ConceptWriteError::ConceptRead { typedb_source: error }]),
@@ -1823,7 +1822,6 @@ impl ThingManager {
     ) -> Result<(), Box<ConceptReadError>> {
         // TODO: Should not collect here (iterate_writes() already copies)
         for (key, _write) in snapshot.iterate_writes().collect_vec() {
-            println!("Creating locks");
             if ThingEdgeHas::is_has(&key) {
                 let has = ThingEdgeHas::decode(Bytes::Reference(key.bytes()));
                 let object = Object::new(has.from());
@@ -1858,7 +1856,6 @@ impl ThingManager {
             self.type_manager(),
             attribute.type_(),
         )?;
-        dbg!("Unique constraint", &unique_constraint_opt);
         if let Some(unique_constraint) = unique_constraint_opt {
             let attribute_key = attribute.vertex();
             // let attribute_value = snapshot
--- a/concept/thing/thing_manager.rs
+++ b/concept/thing/thing_manager.rs
@@ -12,7 +12,11 @@ use std::{
     sync::Arc,
 };
 
-use bytes::{Bytes, byte_array::ByteArray, util::increment};
+use bytes::{
+    Bytes,
+    byte_array::ByteArray,
+    util::{concat_bytes, increment},
+};
 use encoding::{
     AsBytes, EncodingKeyspace, Keyable, Prefixed,
     graph::{
@@ -62,7 +66,7 @@ use resource::{
 use storage::{
     key_range::{KeyRange, RangeEnd, RangeStart},
     key_value::{StorageKey, StorageKeyArray, StorageKeyReference},
-    snapshot::{ReadableSnapshot, WritableSnapshot, lock::create_custom_lock_key, write::Write},
+    snapshot::{ReadableSnapshot, WritableSnapshot, write::Write},
 };
 
 use crate::{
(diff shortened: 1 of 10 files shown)


database/tests/test_transaction_isolation.rs
```diff
@@ -0,0 +1,602 @@
...
+
+/// Assert exactly one of two commit outcomes succeeded and the other failed. Returns the failure
+/// for the caller to inspect.
+fn assert_exactly_one_failed(a: Result<(), DataCommitError>, b: Result<(), DataCommitError>) -> DataCommitError {
+    match (a, b) {
+        (Ok(()), Err(err)) | (Err(err), Ok(())) => err,
+        (Ok(()), Ok(())) => panic!("expected one commit to fail with isolation conflict; both succeeded"),
+        (Err(a), Err(b)) => panic!("expected exactly one commit to fail; both failed: a={a:?} b={b:?}"),
+    }
+}
```
krishnangovindraj: This is silly. The first has to succeed and the second has to fail.

database/tests/test_transaction_isolation.rs
```diff
@@ -0,0 +1,602 @@
...
+fn assert_isolation_conflict(err: &DataCommitError) -> &IsolationConflict {
+    match err {
+        DataCommitError::SnapshotError {
+            typedb_source: SnapshotError::Commit { typedb_source: StorageCommitError::Isolation { conflict, .. } },
+        } => conflict,
+        other => panic!(
+            "expected DataCommitError::SnapshotError(SnapshotError::Commit(StorageCommitError::Isolation)), got: {other:?}"
+        ),
+    }
+}
```
krishnangovindraj: Can be stronger as a macro that actually matches the variant of the conflict.
(That was good advice from Dmitrii on my last PR)

database/tests/test_transaction_isolation.rs
```diff
@@ -0,0 +1,602 @@
...
+    );
+
+    let mut tx1 = open_write(database.clone());
+    let mut tx2 = open_write(database.clone());
+    tx1 = run_write(tx1, "insert $p isa person; $f isa friendship;");
+    tx2 = run_write(tx2, "insert $p isa person; $f isa friendship;");
+
+    try_commit(tx1).expect("tx1 commit");
+    try_commit(tx2).expect("tx2 commit");
+}
```
krishnangovindraj, line 114: Trivial case, but ok if we're being complete.

database/tests/test_transaction_isolation.rs
```diff
@@ -0,0 +1,602 @@
...
+    // Owner cardinality is 0..2: combined writes would stay within the bound, so the constraint
+    // is not actually violated. But the cardinality commit lock is intentionally coarse - it is
+    // keyed by (owner, owns(attribute_type)) without inspecting current counts, because each
+    // transaction validates against its own snapshot and would otherwise both pass through and
+    // commit a sum that violates the bound. So both writes contend on the same lock and one is
+    // rejected as an isolation conflict, even though the resulting state would be valid.
+    //
+    // This test pins that behaviour: if it ever changes (e.g., to a finer-grained scheme that
+    // permits non-violating concurrent writes), this test will fail and prompt review of the
+    // cardinality lock-key composition.
```
krishnangovindraj: nice

database/tests/test_transaction_isolation.rs
```diff
@@ -0,0 +1,602 @@
...
+
+/////////////////////////////////
+// Concurrent owns cardinality //
+/////////////////////////////////
+
+#[test]
+fn concurrent_has_writes_violating_owns_cardinality_one_fails() {
+    // Owner cardinality is 0..1 for `name`. Two concurrent has-writes from different snapshots,
+    // each adding a name to the same person, would push the owner to 2 names if both committed -
+    // exceeding the upper bound. The cardinality commit lock must serialise them and reject one.
```
krishnangovindraj: TODO for myself: Check there's one for concurrent deletes and lower bounds.

concept/thing/thing_manager.rs
```diff
@@ -1858,21 +1858,21 @@ impl ThingManager {
...
+            // let attribute_value = snapshot
+            //     .get_last_existing::<BUFFER_VALUE_INLINE>(
+            //         attribute_key.into_storage_key().as_reference(),
+            //         storage_counters,
+            //     )
+            //     .map_err(|error| Box::new(ConceptReadError::SnapshotGet { source: error }))?
+            //     .ok_or(ConceptReadError::InternalMissingAttributeValue {})?;
+
+            // Note: trade much smaller locks (eg. not entire string) for rare concurrent lock conflicts when
+            //       two different values has to the same hash
```
krishnangovindraj, line 1861: Cleanup

concept/thing/thing_manager.rs
```diff
@@ -1858,21 +1858,21 @@ impl ThingManager {
...
             let lock_key = create_custom_lock_key(
```
krishnangovindraj: `create_custom_lock_key` could really be renamed `concat_bytes`

database/tests/test_transaction_isolation.rs
```diff
@@ -0,0 +1,602 @@
...
+    commit_schema(
+        database.clone(),
+        r#"define
+            entity person, owns ref @key;
+            attribute ref, value string;
+        "#,
+    );
+
+    let key_value = "a-string-key-long-enough-to-force-hashed-encoding";
+    debug_assert!(key_value.len() > 16, "must exceed StringAttributeID inline threshold to hit the hashed path");
```
krishnangovindraj: Can we replace this with the constant from StringAttributeID

encoding/tests/test_attribute_vertex.rs
```diff
@@ -190,6 +190,147 @@ fn generate_struct_attribute_vertex() {
...
+                "inline value {:?} must have deterministic_bytes == bytes",
+                value
+            );
+        }
+        // The short-string case must hit the inline branch of StringAttributeID.
+        let short_id = AttributeID::build_inline(Value::String(std::borrow::Cow::Borrowed("hi")));
+        assert!(short_id.unwrap_string().is_inline());
+    }
+
+    // 3. Hashed strings, default hasher: deterministic_bytes drops the disambiguator byte.
```
krishnangovindraj: `// 2.` is missing.

encoding/tests/test_attribute_vertex.rs
```diff
@@ -190,6 +190,147 @@ fn generate_struct_attribute_vertex() {
...
+            .unwrap()
+            .attribute_id()
+            .unwrap_string();
+        let id_b = generator
+            .create_attribute_string(type_id, s_b_bytes.as_reference(), &mut snapshot)
+            .unwrap()
+            .attribute_id()
+            .unwrap_string();
+        assert_ne!(id_a.deterministic_bytes_ref(), id_b.deterministic_bytes_ref());
+    }
```
krishnangovindraj: Do we want this level of testing? I can't imagine humans writing this.

encoding/tests/test_attribute_vertex.rs
```diff
@@ -190,6 +190,147 @@ fn generate_struct_attribute_vertex() {
...
+        assert_ne!(id_a.get_hash_disambiguator(), id_b.get_hash_disambiguator());
+        assert_ne!(id_a.bytes_ref(), id_b.bytes_ref());
+        assert_eq!(
+            id_a.deterministic_bytes_ref(),
+            id_b.deterministic_bytes_ref(),
+            "hash-collided structs must produce equal deterministic_bytes"
+        );
+        assert_eq!(id_a.deterministic_bytes_ref().len(), id_a.bytes_ref().len() - 1);
+    }
+}
```
krishnangovindraj, line 308: This feels like the right place to ensure that the same string in two different snapshots get the same `deterministic_bytes`.

concept/thing/thing_manager.rs
```diff
@@ -1858,21 +1858,21 @@ impl ThingManager {
...
                 [
                     &Infix::PropertyAnnotationUnique.infix_id().bytes(),
                     &*unique_constraint.source().attribute().vertex().to_bytes(),
-                    attribute_key.attribute_id().bytes(),
-                    &attribute_value,
-                    &*owner.vertex().to_bytes(),
+                    attribute_key.attribute_id().deterministic_bytes(),
                     &*unique_constraint.source().owner().vertex().to_bytes(),
```
krishnangovindraj, line 1867: Never noticed but the use of source here is really clever.

krishnangovindraj: This is the function for unique, which can't be tightened, but I wonder if `card` acts the same way or writes a key for each constraint source it gets. The code looks like it does, but I'd like a test to make sure we never change it.

Requested here: https://github.com/typedb/typedb/pull/7800/changes#r3200450082

database/tests/test_transaction_isolation.rs
```diff
@@ -0,0 +1,602 @@
...
+    );
+    tx2 = run_write(
+        tx2,
+        r#"match $g isa friendship, has rid 200; $p isa person, has id 1; insert $g links (friend: $p);"#,
+    );
+
+    let err = assert_exactly_one_failed(try_commit(tx1), try_commit(tx2));
+    assert_isolation_conflict(&err);
+}
+
```
krishnangovindraj, line 544: Can we add a test for the case where an interface is subject to an inherited cardinality constraint from a super-interface, and a tightened one from a sub-interface (inherited or declared).

database/tests/test_transaction_isolation.rs
flyingsilverfin: yeah why not

database/tests/test_transaction_isolation.rs
flyingsilverfin: wait why this is fine no?

database/tests/test_transaction_isolation.rs
flyingsilverfin: I think we don't need this per se, but we do need a test that validates card correctly using an inherited card/unique/key. There's no difference for inherited or inherited + tightened since this test isn't actually checking for card violations, only for concurrent modifications.

encoding/tests/test_attribute_vertex.rs
flyingsilverfin: no will delete

database/tests/test_transaction_isolation.rs
krishnangovindraj: It's fine, but it makes it look as if they're concurrent commits, while you know exactly which one should fail.

database/tests/test_transaction_isolation.rs
krishnangovindraj: I do mean for concurrent modifications.
```
define 
entity tbase ,
  plays r:super @card(0..2);
entity tsub sub tbase,
  plays r:sub @card(0..1);
```

We need to be sure that the following insert conflicting locks:
`match $t isa tsub; insert $_ isa r, links (sub: $t);`
`match $t isa tsub; insert $_ isa r, links (super: $t);`

I've seen the loop that handles it in the code, but we should still test it in case we decide to refactor it.

flyingsilverfin: added

encoding/tests/test_attribute_vertex.rs
```diff
@@ -331,6 +302,92 @@ fn attribute_id_deterministic_bytes_strips_disambiguator_for_hashed_ids() {
...
+            .unwrap()
+            .attribute_id()
+            .unwrap_string();
+        let id_b = generator
+            .create_attribute_string(type_id, s_bytes.as_reference(), &mut snapshot_b)
+            .unwrap()
+            .attribute_id()
+            .unwrap_string();
+        assert!(!id_a.is_inline() && !id_b.is_inline());
+        assert_eq!(id_a.deterministic_bytes_ref(), id_b.deterministic_bytes_ref());
```
krishnangovindraj, line 361: Wait no, Can we assert an isolation conflict here please?

krishnangovindraj (changes requested):
Just the one question.
And the PR description

krishnangovindraj (approved):

flyingsilverfin:
Fixes https://github.com/typedb/typedb/issues/7795

PR #7802, opened 2026-05-11
Add default advertise-address retrieved from listen-address

farost:

Remove pre-inserted `advertise-address` values from server endpoints configuration. Instead, auto-generate the advertise addresses based on the listen addresses following these rules:
* if the listen address is an unspecified address (`0.0.0.0`), substitute the IP with loopback `127.0.0.1`, preserving the original port -- this allows the clients to use `127.0.0.1` instead of `0.0.0.0` for connections, which is usually the intention;
* otherwise, use the listen address for advertising as is.

This reduces the amount of fields required for modifications in case the user wants to change the default server ports and helps avoid sudden misconfigurations while operating with local servers.

Nothing changed for default configuration:

```
./typedb server

=====================================================================================
     ________ __      __ _____    _______  _____    _____       _______  _______
    |__    __|\  \  /  /|   _  \ |   _   ||   _  \ |   _  \    |   _   ||   _   |
       |  |    \  \/  / |  | |  ||  | |__||  | |  ||  | |  |   |  | |  ||  | |__|
       |  |     \    /  |  |/  / |  |___  |  | |  ||  |/  /    |  | |__||  |___
       |  |      |  |   |   __/  |   ___| |  | |  ||   _  \    |  |  __ |   ___|
       |  |      |  |   |  |     |  |  __ |  | |  ||  | |  |   |  | |  ||  |  __
       |  |      |  |   |  |     |  |_|  ||  |/  / |  |/  /    |  |_|  ||  |_|  |
       |__|      |__|   |__|     |_______||_____/  |_____/     |_______||_______|

=====================================================================================

Running TypeDB CE 3.10.4 in development mode.
Serving:
  gRPC:  0.0.0.0:1729 (connect via 127.0.0.1:1729)
  HTTP:  0.0.0.0:8000 (connect via 127.0.0.1:8000)
  Admin: 127.0.0.1:1728 (localhost only)
TLS: disabled
  WARNING: TLS NOT ENABLED. Credentials are transmitted unencrypted in plaintext.
  Drivers must be configured to connect *without TLS*.
```

## Implementation

Enhance the `unwrap_or_else` logic when building the server state.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -119,7 +119,7 @@ native_artifact_files.artifact(
     name = "typedb_console_artifact",
     group_name = "typedb-console-{platform}",
     artifact_name = "typedb-console-{platform}-{version}.{ext}",
-    tag = "3.10.0-alpha-1",
+    commit = "e8e5a70125f86250268e30cd22d25312ca89e46b",
 )
 use_repo(
     native_artifact_files,
--- a/server/state/mod.rs
+++ b/server/state/mod.rs
@@ -9,7 +9,12 @@ pub mod server_operator;
 pub mod transaction_operator;
 pub mod user_operator;
 
-use std::{collections::HashSet, net::SocketAddr, path::PathBuf, sync::Arc};
+use std::{
+    collections::HashSet,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
+    path::PathBuf,
+    sync::Arc,
+};
 
 use concurrency::{IntervalRunner, TokioTaskSpawner};
 use database::database_manager::DatabaseManager;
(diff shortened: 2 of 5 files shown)


server/state/mod.rs
```diff
@@ -351,3 +368,32 @@ impl ServerStateBuilder {
         }
     }
 }
+
+#[cfg(test)]
```
farost, line 371: Already runs in CI

MODULE.bazel
```diff
@@ -119,7 +119,7 @@ native_artifact_files.artifact(
     name = "typedb_console_artifact",
     group_name = "typedb-console-{platform}",
     artifact_name = "typedb-console-{platform}-{version}.{ext}",
-    tag = "3.10.0-alpha-1",
+    commit = "e8e5a70125f86250268e30cd22d25312ca89e46b",
```
farost, line 122: Not to depend on an alpha version

flyingsilverfin (approved):

PR #7811, opened 2026-05-19
Add state storage counters

farost:

Introduce a small set of state counters that ride alongside every `CommitRecord` and are preserved in the storage and replicated between nodes of a distributed cluster to ensure the synchronization of these counters between any possible primary node. 

**This change makes the storage not forward-compatible, making it unable to downgrade the server version with new WAL records from this commit.**

### Locally generated counters analysis

**State counters:**
* Entity / Relation IIDs per `type_id`: generated locally, `fetch_add` value is encoded directly into the entity vertex on disk. Require replication so every new leader replica knows about the previous values.
* `SnapshotId`: Part of the `(open_seq, snapshot_id)` commit-idempotency key. Currently, their replication is not required, but it is a nice protective measure in case some of the invariants break. Now, `SnapshotId`s are unique even if you use an `open_seq` from the past.

**Not state counters:**
* Attribute IDs: content-derived based on the value, not a server state.
* TypeIDs and struct/function keys: allocators scan the storage for the first unused ID, so they do not require a separate state.
* WAL seqnum-based numbers: not a separately generated state.

State counters are saved in `CommitRecord`s so they could be shared with other replicas on commit and restored on load. **This leads to the following expectation:** The state must be restored, and since it's in the WAL, the WAL must contain all the recent counter updates. We may need to put these counters into the storage separately if we consider introducing WAL truncation.

**Important**: legacy commit records do not contain entity/relation IID values, so they still must check the storage for their values.

Also, for simplification, counters are updated regardless of the state of concurrent snapshots. E.g.:
1. You open Tx1, Tx2, and Tx3.
2. Tx1 inserts data
3. Tx2 inserts data
4. Tx2 commits data
5. Tx3 inserts data
6. Tx1 commits data
7. Tx3 closes

In the end, the state will have the counters INCLUDING TX3, because Tx1's commit will use Tx3 updates. This is simple and safe, and we must have enough counters to support this mechanism.

### Architecture

- Introduce a new `CommitRecord` of version 3 to store the state counters.
- `resource::state_counter` introduces `CounterId` (the wire-format discriminator, bincode-serialized) and `StateCounters` (the per-database backing store). `StateCounters` is composed of two small inner types: `PerTypeIdCounters` (dense `Box<[AtomicU64]>` indexed by `TypeIDUInt`) for the IID counters, and `GlobalCounter` for single-valued counters like `SnapshotId`.
- `MVCCStorage` owns the `Arc<StateCounters>` and exposes it via `state_counters()`. `Database::apply_counter_advances` is the replica's catch-up hook.
- Snapshots now hold a new `SnapshotCounters` struct consisting of an `Arc<StateCounters>`, a `HashSet<CounterId>` of touched counters, and a preset advances vector (non-empty only when the snapshot was reconstructed from an incoming `CommitRecord`). `WritableSnapshot::allocate_counter` is the single allocation entry point.
- At commit time, `SnapshotCounters::into_advances` reads each touched counter's *live* `next_value` and stamps `Vec<(CounterId, u64)>` onto the `CommitRecord` (a vector for serializability).
- `ThingVertexGenerator` no longer owns atomics: `create_entity` / `create_relation` route through `snapshot.allocate_counter`. `Database::load` respects the new structure as well.

`apply_counter_advances` uses `fetch_max`, so order (tx1 can allocate counters before tx2, but get committed after tx2) doesn't matter.

Used in https://github.com/typedb/typedb-cluster/pull/691

### DIFF
--- a/database/database.rs
+++ b/database/database.rs
@@ -41,7 +41,10 @@ use error::typedb_error;
 use fail_point::{UNFINISHED_CHECKPOINT, fail_point};
 use function::{FunctionError, function_cache::FunctionCache};
 use query::query_cache::QueryCache;
-use resource::constants::database::{CHECKPOINT_INTERVAL, STATISTICS_UPDATE_INTERVAL};
+use resource::{
+    constants::database::{CHECKPOINT_INTERVAL, STATISTICS_UPDATE_INTERVAL},
+    state_counter::CounterId,
+};
 use storage::{
     MVCCStorage, StorageDeleteError, StorageOpenError, StorageResetError,
     durability_client::{DurabilityClient, DurabilityClientError, WALClient},
@@ -100,6 +103,10 @@ impl<D> Database<D> {
 
     pub fn name(&self) -> &str {
         &self.name
+    }
+
+    pub fn apply_counter_advances(&self, advances: &[(CounterId, u64)]) {
+        self.storage.state_counters().apply_advances(advances);
     }
 
     pub(super) fn reserve_write_transaction(&self, timeout_millis: u64) -> Result<(), TransactionError> {
@@ -487,10 +494,6 @@ impl Database<WALClient> {
         match Arc::get_mut(&mut self.type_vertex_generator) {
             None => return Err(CorruptionPartialResetTypeVertexGeneratorInUse {}),
             Some(type_vertex_generator) => type_vertex_generator.reset(),
-        }
-        match Arc::get_mut(&mut self.thing_vertex_generator) {
-            None => return Err(CorruptionPartialResetThingVertexGeneratorInUse {}),
-            Some(thing_vertex_generator) => thing_vertex_generator.reset(),
         }
 
         let thing_statistics = Arc::get_mut(&mut locked_schema.thing_statistics).unwrap();
--- a/encoding/graph/thing/vertex_generator.rs
+++ b/encoding/graph/thing/vertex_generator.rs
@@ -71,6 +71,9 @@ impl ThingVertexGenerator {
         storage: Arc<MVCCStorage<D>>,
         large_value_hasher: fn(&[u8]) -> u64,
     ) -> Result<Self, EncodingError> {
+        // All counters are loaded from WAL commit records. However, legacy commit records do not
+        // contain state counters for this generator, so we must check the storage on load
+        // as far as there are legacy records on disk.
         Self::seed_state_counters_from_storage(storage)?;
         Ok(ThingVertexGenerator { large_value_hasher })
     }
--- a/encoding/graph/thing/vertex_generator.rs
+++ b/encoding/graph/thing/vertex_generator.rs
@@ -4,13 +4,13 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::sync::{
-    Arc,
-    atomic::{AtomicU64, Ordering},
-};
+use std::sync::Arc;
 
 use bytes::{Bytes, byte_array::ByteArray};
-use resource::profile::StorageCounters;
+use resource::{
+    profile::StorageCounters,
+    state_counter::{CounterId, StateCounters},
+};
 use storage::{
     MVCCStorage,
     key_range::KeyRange,
@@ -32,7 +32,7 @@ use crate::{
             vertex_attribute::{AttributeID, AttributeVertex, IntegerAttributeID, StringAttributeID},
             vertex_object::{ObjectID, ObjectVertex},
         },
-        type_::vertex::{TypeID, TypeIDUInt, TypeVertex, build_type_vertex_prefix_key},
+        type_::vertex::{TypeID, TypeVertex, build_type_vertex_prefix_key},
     },
     layout::prefix::Prefix,
     value::{
@@ -45,8 +45,6 @@ use crate::{
 
 #[derive(Debug)]
 pub struct ThingVertexGenerator {
-    entity_ids: Box<[AtomicU64]>,
-    relation_ids: Box<[AtomicU64]>,
     large_value_hasher: fn(&[u8]) -> u64,
 }
 
@@ -58,19 +56,11 @@ impl Default for ThingVertexGenerator {
 
 impl ThingVertexGenerator {
     pub fn new() -> Self {
-        // TODO: we should create a resizable Vector linked to the id of types/highest id of each type
-        //       this will speed up booting time on load (loading this will require MAX types * 3 iterator searches) and reduce memory footprint
         Self::new_with_hasher(seahash::hash)
     }
 
     pub fn new_with_hasher(large_value_hasher: fn(&[u8]) -> u64) -> Self {
-        // TODO: we should create a resizable Vector linked to the id of types/highest id of each type
-        //       this will speed up booting time on load (loading this will require MAX types * 3 iterator searches) and reduce memory footprint
-        ThingVertexGenerator {
-            entity_ids: Self::allocate_empty_ids(),
-            relation_ids: Self::allocate_empty_ids(),
-            large_value_hasher,
-        }
+        ThingVertexGenerator { large_value_hasher }
     }
 
     pub fn load<D>(storage: Arc<MVCCStorage<D>>) -> Result<Self, EncodingError> {
@@ -81,6 +71,12 @@ impl ThingVertexGenerator {
         storage: Arc<MVCCStorage<D>>,
         large_value_hasher: fn(&[u8]) -> u64,
     ) -> Result<Self, EncodingError> {
+        Self::seed_state_counters_from_storage(storage)?;
+        Ok(ThingVertexGenerator { large_value_hasher })
+    }
+
+    fn seed_state_counters_from_storage<D>(storage: Arc<MVCCStorage<D>>) -> Result<(), EncodingError> {
+        let counters = storage.state_counters().clone();
         let read_snapshot = storage.clone().open_snapshot_read();
         let entity_types = read_snapshot
             .iterate_range(
@@ -102,9 +98,6 @@ impl ThingVertexGenerator {
             )
             .collect_cloned_vec(|k, _v| TypeVertex::decode(Bytes::Reference(k.bytes())).type_id_().as_u16())
             .map_err(|err| EncodingError::ExistingTypesRead { source: err })?;
-
-        let entity_ids = Self::allocate_empty_ids();
-        let relation_ids = Self::allocate_empty_ids();
         for type_id in entity_types {
             let mut max_object_id =
                 ObjectVertex::build_entity(TypeID::new(type_id), ObjectID::new(u64::MAX)).to_bytes().into_array();
@@ -117,7 +110,10 @@ impl ThingVertexGenerator {
                 if ObjectVertex::is_entity_vertex(StorageKeyReference::new(ObjectVertex::KEYSPACE, &prev_bytes)) {
                     let object_vertex = ObjectVertex::decode(&prev_bytes);
                     if object_vertex.type_id_() == TypeID::new(type_id) {
-                        entity_ids[type_id as usize].store(object_vertex.object_id().as_u64() + 1, Ordering::Relaxed);
+                        counters.update_to_at_least(
+                            CounterId::EntityVertex { type_id },
+                            object_vertex.object_id().as_u64() + 1,
+                        );
                     }
                 }
             }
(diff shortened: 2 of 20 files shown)


database/database.rs
```diff
@@ -102,6 +105,10 @@ impl<D> Database<D> {
         &self.name
     }
 
+    pub fn apply_counter_advances(&self, advances: &[(CounterId, u64)]) {
```
farost, line 108: Direct accessor for cluster

database/database.rs
```diff
@@ -488,10 +495,6 @@ impl Database<WALClient> {
             None => return Err(CorruptionPartialResetTypeVertexGeneratorInUse {}),
             Some(type_vertex_generator) => type_vertex_generator.reset(),
         }
-        match Arc::get_mut(&mut self.thing_vertex_generator) {
```
farost, line 491: The state counters replace this, they are reset in a different place

encoding/graph/thing/vertex_generator.rs
```diff
@@ -58,19 +56,11 @@ impl Default for ThingVertexGenerator {
 
 impl ThingVertexGenerator {
     pub fn new() -> Self {
-        // TODO: we should create a resizable Vector linked to the id of types/highest id of each type
```
farost, line 61: This TODO moves inside the counters

encoding/graph/type_/vertex.rs
```diff
@@ -102,9 +103,6 @@ impl fmt::Debug for TypeID {
     }
 }
 
-// TODO: this type should move into constants.rs, similarly to DefinitionIDUInt
```
farost, line 105: Done

resource/resource.rs
```diff
@@ -8,3 +8,4 @@ pub mod constants;
 pub mod distribution_info;
 pub mod perf_counters;
 pub mod profile;
+pub mod state_counter;
```
farost, line 11: It's not strictly a storage thing, so I put it in `resource` for general traits. But with the current look of the implementation, I think it better suits somewhere in `storage`. Wdyt? @flyingsilverfin

resource/state_counter.rs
```diff
@@ -0,0 +1,246 @@
...
+
+#[derive(Copy, Clone, Hash, PartialEq, Eq, Debug, Serialize, Deserialize)]
+pub enum CounterId {
+    EntityVertex { type_id: TypeIDUInt },
+    RelationVertex { type_id: TypeIDUInt },
+    SnapshotId,
+}
+
+const COUNTER_ID_MAX_SIZE: usize = 8;
+const _: () = assert!(size_of::<CounterId>() <= COUNTER_ID_MAX_SIZE);
```
farost, line 24: Not to blow up the serializible object

resource/state_counter.rs
```diff
@@ -0,0 +1,246 @@
...
+    iter,
+    sync::atomic::{AtomicU64, Ordering},
+};
+
+use serde::{Deserialize, Serialize};
+
+use crate::constants::encoding::TypeIDUInt;
+
+#[derive(Copy, Clone, Hash, PartialEq, Eq, Debug, Serialize, Deserialize)]
+pub enum CounterId {
```
farost, line 17: Every state counter type

resource/state_counter.rs
```diff
@@ -0,0 +1,246 @@
...
+    EntityVertex { type_id: TypeIDUInt },
+    RelationVertex { type_id: TypeIDUInt },
+    SnapshotId,
+}
+
+const COUNTER_ID_MAX_SIZE: usize = 8;
+const _: () = assert!(size_of::<CounterId>() <= COUNTER_ID_MAX_SIZE);
+
+pub struct StateCounters {
+    entity_ids: PerTypeIdCounters,
```
farost, line 27: Had a choice between a generic set of `TCounters` with more trait members, but I thought that working with the fields directly is simpler and cleaner. The generic implementation blew up quite fast

resource/state_counter.rs
```diff
@@ -0,0 +1,246 @@
...
+
+    pub fn next_value(&self, id: CounterId) -> u64 {
+        match id {
+            CounterId::EntityVertex { type_id } => self.entity_ids.next_value(type_id),
+            CounterId::RelationVertex { type_id } => self.relation_ids.next_value(type_id),
+            CounterId::SnapshotId => self.snapshot_id.next_value(),
+        }
+    }
+
+    pub fn update_to_at_least(&self, id: CounterId, value: u64) {
```
farost, line 69: Incremental only, so `to_at_least`

resource/state_counter.rs
```diff
@@ -0,0 +1,246 @@
...
+    pub fn reset(&self) {
+        self.entity_ids.reset();
+        self.relation_ids.reset();
+        self.snapshot_id.reset();
+    }
+}
+
+/// One [`AtomicU64`] per [`TypeIDUInt`].
+struct PerTypeIdCounters {
+    // TODO: we should create a resizable Vector linked to the id of types/highest id of each type
```
farost, line 92: The moved todo from the past

resource/state_counter.rs
```diff
@@ -0,0 +1,246 @@
...
+    //       this will speed up booting time on load (loading this will require MAX types * 3 iterator searches) and reduce memory footprint
+    atomics: Box<[AtomicU64]>,
+}
+
+impl PerTypeIdCounters {
+    const INITIAL: u64 = 0;
+    const COUNT: usize = TypeIDUInt::MAX as usize + 1;
+
+    fn new() -> Self {
+        Self { atomics: iter::repeat_with(|| AtomicU64::new(Self::INITIAL)).take(Self::COUNT).collect() }
```
farost, line 102: I use the same approach as in Thing IID Generator before: pre-allocate aggressively for each `TypeID`

resource/state_counter.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    fn reset(&self) {
+        for atomic in self.atomics.iter() {
+            atomic.store(Self::INITIAL, Ordering::SeqCst);
+        }
+    }
+}
+
+struct GlobalCounter {
```
farost, line 124: A more general counter for something like SnapshotID

storage/snapshot/snapshot.rs
```diff
@@ -231,7 +234,8 @@ impl<D: fmt::Debug> fmt::Debug for ReadSnapshot<D> {
 impl<D> ReadSnapshot<D> {
     pub(crate) fn new(storage: Arc<MVCCStorage<D>>, open_sequence_number: SequenceNumber) -> Self {
         // Note: for serialisability, we would need to register the open transaction to the IsolationManager
-        ReadSnapshot { open_sequence_number, id: SnapshotId::new(), iterator_pool: IteratorPool::new(), storage }
+        let id = SnapshotId::from_number(storage.state_counters().allocate(CounterId::SnapshotId));
```
farost, line 237: `SnapshotId` is now just an ID, it does not drive the value generation, so it requires a state counter to initialize.

storage/record.rs
```diff
@@ -63,12 +65,61 @@ impl DurabilityRecord for LegacyCommitRecordV1 {
 
 impl SequencedDurabilityRecord for LegacyCommitRecordV1 {}
 
+#[derive(Serialize, Deserialize)]
+pub struct LegacyCommitRecordV2 {
```
farost, line 69: A new legacy commit record, identical to CommitRecord before this change

storage/record.rs
```diff
@@ -63,12 +65,61 @@ impl DurabilityRecord for LegacyCommitRecordV1 {
...
+
+impl SequencedDurabilityRecord for LegacyCommitRecordV2 {}
+
 #[derive(Serialize, Deserialize)]
 pub struct CommitRecord {
     operations: OperationsBuffer,
     open_sequence_number: SequenceNumber,
     commit_type: CommitType,
     snapshot_id: SnapshotId,
+    counter_advances: Vec<(CounterId, u64)>,
```
farost, line 126: Still thinking if I want to see an encapsulating struct for this vec

encoding/graph/thing/vertex_generator.rs
```diff
@@ -81,6 +71,12 @@ impl ThingVertexGenerator {
         storage: Arc<MVCCStorage<D>>,
         large_value_hasher: fn(&[u8]) -> u64,
     ) -> Result<Self, EncodingError> {
+        Self::seed_state_counters_from_storage(storage)?;
+        Ok(ThingVertexGenerator { large_value_hasher })
+    }
+
+    fn seed_state_counters_from_storage<D>(storage: Arc<MVCCStorage<D>>) -> Result<(), EncodingError> {
```
farost, line 81: We still seed the type_id-based counters based on the storage. I think, to complete the `SnapshotId` "protection", I must pick the latest change of SnapshotId, too, and initialize it on cold loading. This will provide us the opportunity to safely open write transactions in the past (snapshot ids are always unique). 

Also, I'm not happy that the CommitRecords are extended only for replication, and the counter updates take additional storage without any usage (we read these values from the storage, not from the WAL). I will think about it once again tomorrow with a fresh head

storage/recovery/commit_recovery.rs
```diff
@@ -162,16 +178,19 @@ pub(crate) fn apply_recovered(
...
             RecoveryCommitStatus::Validated(commit_record) => {
+                let counter_advances = commit_record.counter_advances().to_vec();
                 let write_batches = WriteBatches::from_operations(commit_sequence_number, commit_record.operations());
                 isolation_manager.load_validated(commit_sequence_number, commit_record);
                 keyspaces.write(write_batches).map_err(|error| KeyspaceWrite { source: error })?;
                 fail_point!(RECOVERY_PARTIAL_WRITE);
                 isolation_manager
                     .applied(commit_sequence_number)
                     .map_err(|error| Internal { name: Arc::new(database_name.to_owned()), source: Arc::new(error) })?;
+                state_counters.apply_advances(&counter_advances);
```
farost, line 189: Apply all the counter updates from the WAL records. Vec iterations can slow it down, but for regular databases, the slowdown must be minimal, since the number of keys is limited to the number of types in the database.

encoding/graph/thing/vertex_generator.rs
```diff
@@ -81,6 +71,15 @@ impl ThingVertexGenerator {
         storage: Arc<MVCCStorage<D>>,
         large_value_hasher: fn(&[u8]) -> u64,
     ) -> Result<Self, EncodingError> {
+        // All counters are loaded from WAL commit records. However, legacy commit records do not
```
farost, line 74: We still need to check the storage

storage/snapshot/snapshot_id.rs
```diff
@@ -4,19 +4,8 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::sync::atomic::{AtomicU64, Ordering};
-
 use serde::{Deserialize, Serialize};
 
-// Safety: this process-local generator produces unique SnapshotIds within a single process
```
farost, line 11: We safe and restore SnapshotId numbers from the WAL, so all these complicated limitations are no longer needed. Do whatever you want as far as the value is recovered.

storage/recovery/commit_recovery.rs
```diff
@@ -162,16 +178,19 @@ pub(crate) fn apply_recovered(
     for (commit_sequence_number, commit) in recovered_commits {
         match commit {
             RecoveryCommitStatus::Validated(commit_record) => {
+                let counter_advances = commit_record.counter_advances().to_vec();
```
farost, line 181: We currently save these counters only to the WAL. However, if we introduce WAL truncation, we will need to pack these values into a separate consistent store, because we will not be able to ensure that all the counter values are still in the WAL updates.

storage/snapshot/snapshot.rs
farost, line 237: Currently, every `ReadSnapshot` moves the global `SnapshotID` counter, which can be excessive. Idk if it's correct to "optimize" it away

farost: If we don't commit anything, the restart will restore the counter. However, if we start the server and then do a single write commit, we effectively increase the counter by 4:
1. We currently open 3 read snapshots on boot: ThingVertexGenerator::new, TypeCache::new and FunctionCache::new. 
2. The committed snapshot will do +1 + commit, and the +4 will be saved

If we don't commit anything and just restart, the counter will not be touched. Just an example of how it can be excessive. 

But it should not be a real problem considering the counter = u64.

farost:
Closed in favor of https://github.com/typedb/typedb/pull/7816

PR #7812, opened 2026-05-20
Introduce 'given' stage for running the same query on multiple inputs

krishnangovindraj:
Introduces the TypeQL `given` stage which allows `rows" to be passed to a query as input. By running the same pipeline for multiple rows of input data (~~SIMD~~ SQMD?) in a single query, it avoids the network and query-compilation overhead that running multiple queries  (or just the query-compilation overhead in case of one large query). Since given rows are separate from the TypeQL query string, using to get input parameters into queries is safe from TypeQL-injection (c.f. working them into the TypeQL query string).

**Examples:** 
```
# Insert a new `person`s  with name =`$n` for each row given.
given $n: string;
insert $p isa person, has name == $n;
```

```
# Get the name for each person `$p` in the given rows.
given $p: person;
match $p has name $n;
```

For examples on how to construct the given rows in the GRPC drivers, Please check that driver's README on [github](https://github.com/typedb/typedb-driver)

## Implementation
* Protocol & HTTP query requests have GivenRows.
* These are decoded into a batch by the QueryManager, via a trait `GivenRows`. 
* A `GivenStage` type-checks the input row (and optionality checks too)

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -106,12 +106,12 @@ workspace_refs = use_repo_rule("@typedb_bazel_distribution//common/workspace_ref
 workspace_refs(
     name = "typedb_workspace_refs",
     workspace_commit_dict = {
-#        "typeql+": "a1fae91ebd68ed2d7b2eb5035c1287c3bc66789b",
-        "typedb_protocol+": "9a45ab625534a01afe5ecf89acaa06cecc92cafb",
+        "typeql+": "e16626a2129c554436195384a9fda3a3ea36fa2f",
+        "typedb_protocol+": "810e94e12961e4408ee9433d010fd68691783973",
     },
     workspace_tag_dict = {
-        "typeql+": "3.11.0",
-#        "typedb_protocol+": "3.11.0",
+#        "typeql+": "3.11.0-rc0",
+#        "typedb_protocol+": "3.11.0-rc0",
     },
 )
 
@@ -159,20 +159,20 @@ git_override(
 bazel_dep(name = "typeql", version = "0.0.0")
 git_override(
     module_name = "typeql",
-    remote = "https://github.com/typedb/typeql",
-    tag = "3.11.0",
+    remote = "https://github.com/krishnangovindraj/typeql",
+    commit = "e16626a2129c554436195384a9fda3a3ea36fa2f",
 )
 
 bazel_dep(name = "typedb_protocol", version = "0.0.0")
 git_override(
     module_name = "typedb_protocol",
-    remote = "https://github.com/typedb/typedb-protocol",
-    commit = "9a45ab625534a01afe5ecf89acaa06cecc92cafb",
+    remote = "https://github.com/krishnangovindraj/typedb-protocol",
+    commit = "4e8b1179560de715c903099ff1721839fabe984e",
 )
 
 bazel_dep(name = "typedb_behaviour", version = "0.0.0")
 git_override(
     module_name = "typedb_behaviour",
-    remote = "https://github.com/typedb/typedb-behaviour",
-    commit = "353cc2cb845287c3eee2ef8ef8ba5321bde49a9d",
+    remote = "https://github.com/krishnangovindraj/typedb-behaviour",
+    commit = "d96430563804855ef6cb1049fc660970bc4e2cee",
 )
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -174,5 +174,5 @@ bazel_dep(name = "typedb_behaviour", version = "0.0.0")
 git_override(
     module_name = "typedb_behaviour",
     remote = "https://github.com/krishnangovindraj/typedb-behaviour",
-    commit = "d96430563804855ef6cb1049fc660970bc4e2cee",
+    commit = "858af82700791fbd88b6f3bd0ed648f767b2ef31",
 )
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -107,7 +107,7 @@ workspace_refs(
     name = "typedb_workspace_refs",
     workspace_commit_dict = {
         "typeql+": "e16626a2129c554436195384a9fda3a3ea36fa2f",
-        "typedb_protocol+": "810e94e12961e4408ee9433d010fd68691783973",
+        "typedb_protocol+": "bb9c513d4dcff269feb88015dcf99354ad378657",
     },
     workspace_tag_dict = {
 #        "typeql+": "3.11.0-rc0",
@@ -167,7 +167,7 @@ bazel_dep(name = "typedb_protocol", version = "0.0.0")
 git_override(
     module_name = "typedb_protocol",
     remote = "https://github.com/krishnangovindraj/typedb-protocol",
-    commit = "4e8b1179560de715c903099ff1721839fabe984e",
+    commit = "bb9c513d4dcff269feb88015dcf99354ad378657",
 )
 
 bazel_dep(name = "typedb_behaviour", version = "0.0.0")
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -174,5 +174,5 @@ bazel_dep(name = "typedb_behaviour", version = "0.0.0")
 git_override(
     module_name = "typedb_behaviour",
     remote = "https://github.com/krishnangovindraj/typedb-behaviour",
-    commit = "858af82700791fbd88b6f3bd0ed648f767b2ef31",
+    commit = "76771f3c64e674e437515f96e6f1813524592820",
 )
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -174,5 +174,5 @@ bazel_dep(name = "typedb_behaviour", version = "0.0.0")
 git_override(
     module_name = "typedb_behaviour",
     remote = "https://github.com/krishnangovindraj/typedb-behaviour",
-    commit = "ace384ce53bdca85f8f500eb093c214fdcda3d59",
+    commit = "0cad5eb587bbfe56873a8098a637ad7dc0bf47bf",
 )
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -174,5 +174,5 @@ bazel_dep(name = "typedb_behaviour", version = "0.0.0")
 git_override(
     module_name = "typedb_behaviour",
     remote = "https://github.com/krishnangovindraj/typedb-behaviour",
-    commit = "76771f3c64e674e437515f96e6f1813524592820",
+    commit = "cdac9be68cf5844377509e815e0cdf8286f1c168",
 )
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -106,8 +106,8 @@ workspace_refs = use_repo_rule("@typedb_bazel_distribution//common/workspace_ref
 workspace_refs(
     name = "typedb_workspace_refs",
     workspace_commit_dict = {
-        "typeql+": "e16626a2129c554436195384a9fda3a3ea36fa2f",
-        "typedb_protocol+": "bb9c513d4dcff269feb88015dcf99354ad378657",
+        "typeql+": "405c5383a35277acc40bb0921ffd2e8eb1b1b00d",
+        "typedb_protocol+": "3a39f0b94fd7ea316ba5a4f74d44d5afc454cedf",
     },
     workspace_tag_dict = {
 #        "typeql+": "3.11.0-rc0",
@@ -159,20 +159,20 @@ git_override(
 bazel_dep(name = "typeql", version = "0.0.0")
 git_override(
     module_name = "typeql",
-    remote = "https://github.com/krishnangovindraj/typeql",
-    commit = "e16626a2129c554436195384a9fda3a3ea36fa2f",
+    remote = "https://github.com/typedb/typeql",
+    commit = "405c5383a35277acc40bb0921ffd2e8eb1b1b00d",
 )
 
 bazel_dep(name = "typedb_protocol", version = "0.0.0")
 git_override(
     module_name = "typedb_protocol",
-    remote = "https://github.com/krishnangovindraj/typedb-protocol",
-    commit = "bb9c513d4dcff269feb88015dcf99354ad378657",
+    remote = "https://github.com/typedb/typedb-protocol",
+    commit = "3a39f0b94fd7ea316ba5a4f74d44d5afc454cedf",
 )
 
 bazel_dep(name = "typedb_behaviour", version = "0.0.0")
 git_override(
     module_name = "typedb_behaviour",
-    remote = "https://github.com/krishnangovindraj/typedb-behaviour",
-    commit = "0cad5eb587bbfe56873a8098a637ad7dc0bf47bf",
+    remote = "https://github.com/typedb/typedb-behaviour",
+    commit = "a6cc06acd204c1e000abe3646ca2e115583b5ce4",
 )
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -174,5 +174,5 @@ bazel_dep(name = "typedb_behaviour", version = "0.0.0")
 git_override(
     module_name = "typedb_behaviour",
     remote = "https://github.com/krishnangovindraj/typedb-behaviour",
-    commit = "cdac9be68cf5844377509e815e0cdf8286f1c168",
+    commit = "690e05d5dcd31bae9933ff0cf2349bcf6fab6486",
 )
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -174,5 +174,5 @@ bazel_dep(name = "typedb_behaviour", version = "0.0.0")
 git_override(
     module_name = "typedb_behaviour",
     remote = "https://github.com/krishnangovindraj/typedb-behaviour",
-    commit = "690e05d5dcd31bae9933ff0cf2349bcf6fab6486",
+    commit = "ace384ce53bdca85f8f500eb093c214fdcda3d59",
 )
--- a/compiler/annotation/mod.rs
+++ b/compiler/annotation/mod.rs
@@ -119,6 +119,13 @@ typedb_error!(
             variable: String,
             source_span: Option<Span>,
         ),
+        CouldNotResolveGivenRowDeclaredType(
+            18,
+            "An error occurred when trying to resolve the type of the given rows at index: {index}.",
+            index: usize,
+            source_span: Option<Span>,
+            typedb_source: TypeInferenceError
+        ),
     }
 );
 
--- a/compiler/executable/pipeline.rs
+++ b/compiler/executable/pipeline.rs
@@ -13,16 +13,20 @@ use std::{
 use answer::{Type, variable::Variable};
 use concept::thing::statistics::Statistics;
 use ir::{
-    pattern::{Pattern, Vertex, conjunction::Conjunction, nested_pattern::NestedPattern},
+    pattern::{
+        Pattern, Vertex, conjunction::Conjunction, nested_pattern::NestedPattern,
+        variable_category::VariableOptionality,
+    },
     pipeline::{VariableRegistry, function_signature::FunctionID, reduce::AssignedReduction},
 };
+use itertools::Itertools;
 
 use crate::{
-    VariablePosition,
+    ExecutorVariable, VariablePosition,
     annotation::{
         fetch::{AnnotatedFetch, AnnotatedFetchObject, AnnotatedFetchSome},
-        function::{AnnotatedPreambleFunctions, AnnotatedSchemaFunctions},
-        pipeline::AnnotatedStage,
+        function::{AnnotatedPreambleFunctions, AnnotatedSchemaFunctions, FunctionParameterAnnotation},
+        pipeline::{AnnotatedGiven, AnnotatedStage},
     },
     executable::{
         ExecutableCompilationError,
@@ -30,10 +34,11 @@ use crate::{
         fetch::executable::{ExecutableFetch, compile_fetch},
         function::{ExecutableFunctionRegistry, FunctionCallCostProvider, executable::compile_functions},
         insert::{self, executable::InsertExecutable},
-        match_::planner::conjunction_executable::ConjunctionExecutable,
+        match_::{instructions::CheckInstruction, planner::conjunction_executable::ConjunctionExecutable},
         modifiers::{
             DistinctExecutable, LimitExecutable, OffsetExecutable, RequireExecutable, SelectExecutable, SortExecutable,
         },
+        next_executable_id,
         put::PutExecutable,
         reduce::{ReduceExecutable, ReduceRowsExecutable},
         update::executable::UpdateExecutable,
@@ -76,6 +81,7 @@ impl<'a> IntoIterator for &'a TypePopulations {
 #[derive(Debug, Clone)]
 pub struct ExecutablePipeline {
     pub executable_functions: ExecutableFunctionRegistry,
+    pub executable_given: Option<Arc<GivenExecutable>>,
     pub executable_stages: Vec<ExecutableStage>,
     pub executable_fetch: Option<Arc<ExecutableFetch>>,
     pub pipeline_structure: Arc<ParametrisedPipelineStructure>,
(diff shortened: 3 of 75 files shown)


compiler/executable/pipeline.rs
```diff
@@ -583,3 +601,43 @@ fn find_referenced_functions_in_fetch(
...
+        optionality: Vec<VariableOptionality>,
+    ) -> Self {
+        debug_assert!(variables.len() == expected_types.len() && variables.len() == optionality.len());
+        Self { executable_id, variables, expected_types, optionality }
+    }
+
+    pub(crate) fn empty(executable_id: u64) -> Self {
+        Self::new(executable_id, Vec::new(), Vec::new(), Vec::new())
+    }
+
```
krishnangovindraj: unused

executor/pipeline/initial.rs
krishnangovindraj, line 1: This part is a bit funky. We still have an `InitialStage` which now has to take the GivenRows. But it's a second `GivenStage` that validates these rows. 
Though this feels brittle, it keeps the flow for what's in the (IR->annotate->compile->cache) flow (here, the given stage) and what's injected in for each instantiation of the query (the actual inputs and the initial stage).
I didn't find a clean way to just inject the rows into the GivenStage, but I did not look very long.

executor/pipeline/initial.rs
```diff
@@ -31,22 +31,21 @@ impl InitialStage {
 }
 
 pub struct InitialIterator {
-    iterator: Box<FixedBatchRowIterator>,
+    iterator: Box<BatchRowIterator>,
+    index: u32,
```
krishnangovindraj: unused.  removed, but cross-check

ir/pipeline/mod.rs
```diff
@@ -315,16 +315,17 @@ impl VariableRegistry {
         self.variable_source_spans.get(&variable).cloned()
     }
 
-    pub(crate) fn register_function_argument(
+    pub(crate) fn register_input_variable(
```
krishnangovindraj, line 318: Generalises function argument to input_variable.
As of now, function arguments can't be optional.

ir/translation/constraints.rs
```diff
@@ -431,35 +430,15 @@ fn add_typeql_isa(
     Ok(())
 }
 
-fn parse_iid(mut iid: &str) -> ByteArray<THING_VERTEX_MAX_LENGTH> {
```
krishnangovindraj, line 434: Moved to `translation/mod.rs` so we can use it in the http service to parse concepts in given rows.

ir/translation/pipeline.rs
```diff
@@ -57,11 +59,13 @@ impl TranslatedPipeline {
         translation_context: PipelineTranslationContext,
         value_parameters: ParameterRegistry,
         translated_preamble: Vec<Function>,
+        translated_given: Option<TranslatedGiven>,
```
krishnangovindraj, line 61: Big question here: Should this just be a stage? The argument against it being a stage is that it's only allowed to be in the beginning, whereas the other stages can mostly be (almost) anywhere in a pipeline.

ir/translation/pipeline.rs
```diff
@@ -197,31 +202,44 @@ pub(crate) fn translate_pipeline_stages(
...
     let mut translated_stages: Vec<TranslatedStage> = Vec::with_capacity(stages.len());
+    let mut translated_given = None;
+    let mut translated_fetch = None;
+
     for (i, stage) in stages.iter().enumerate() {
         let translated = translate_stage(translation_context, value_parameters, all_function_signatures, stage)?;
         match translated {
             Either::First(stage) => translated_stages.push(stage),
-            Either::Second(fetch) => {
+            Either::Second(Either::First(given)) => {
```
krishnangovindraj: TODO: I should just introduce a local struct for readability.

ir/translation/pipeline.rs
```diff
@@ -265,3 +284,39 @@ fn translate_stage(
...
+    fn hash(&self) -> u64 {
+        let mut hasher = DefaultHasher::new();
+        self.variables.hash_into(&mut hasher);
+        self.labels.hash_into(&mut hasher);
+        hasher.finish()
+    }
+
+    fn equals(&self, other: &Self) -> bool {
+        self.variables.equals(&other.variables) && self.labels.equals(&other.labels)
+    }
```
krishnangovindraj, line 324: Could have it's own file.

query/error.rs
```diff
@@ -36,5 +36,19 @@ typedb_error! {
...
+            "The width of the given rows provided does not match that declared in the query.",
+            declared_width: u32,
+            actual_width: u32,
+        ),
+        ConceptReadErrorValidatingGivenRows(
+            24,
+            "An error occured while validating whether the concepts in the given rows exist",
+            typedb_source: Box<ConceptReadError>,
+        ),
+        UnimplementedLanguageFeature(255, "Unimplemented '{feature}'.", feature: error::UnimplementedFeature),
```
krishnangovindraj: Two unused.

query/query_cache.rs
```diff
@@ -156,12 +164,23 @@ impl StructuralEquality for IRQuery {
...
     }
 
     fn equals(&self, other: &Self) -> bool {
-        self.preamable.equals(&other.preamable) && self.stages.equals(&other.stages) && self.fetch.equals(&other.fetch)
+        let inputs_equal = match (self.given.as_ref(), other.given.as_ref()) {
+            (Some(x), Some(y)) => x.equals(y),
+            _ => false,
+        };
+
+        inputs_equal // self.given.equals(&other.given)
```
krishnangovindraj: Not sure why this isn't just inlined. Wouldn't `None`, `None` be equal too?
Ah ok, it's because it's Option<Arc> instead of Arc<Option>.
Reverting

query/query_manager.rs
```diff
@@ -46,6 +66,8 @@ use crate::{
     redefine, undefine,
 };
 
+pub type GivenRows = Batch;
```
krishnangovindraj: I did a rename at this level in case we want to change things later, though some of the internals are exposed in the server package.
Now has headers so is it's own type.

query/query_manager.rs
```diff
@@ -491,3 +550,25 @@ fn annotate_and_compile_query(
...
+                Err(Box::new(QueryError::GivenRowsWidthDoesNotMatchDeclared {
+                    declared_width: width,
+                    actual_width: batch.width(),
+                }))
+            } else {
+                Ok(batch)
+            }
+        }
+    }
+}
```
krishnangovindraj: Only the shape & existence. Not the content.

server/service/http/message/query/concept.rs
```diff
@@ -330,3 +335,17 @@ pub fn encode_value_type(
...
+    } else {
+        value.value.to_string()
+    };
+    let parsed = parse_value(to_parse.as_str()).map_err(|typedb_source| {
+        HttpServiceError::ParsingValueFailedForGivenEntry { value: to_parse.clone(), typedb_source }
+    })?;
+    Value::from_typeql_literal(&parsed, None).map_err(|typedb_source| {
+        HttpServiceError::TranslatingValueFailedForGivenEntry { value: to_parse.clone(), typedb_source }
+    })
+}
```
krishnangovindraj: TODO: This hasn't actually been tested for all value types.

query/analyse.rs
```diff
@@ -108,6 +115,25 @@ pub struct FunctionStructureAnnotations {
...
 
+fn build_given_annotations(given: &AnnotatedGiven) -> GivenStructureAnnotations {
+    given
+        .variables
+        .iter()
+        .zip(given.expected_types.iter().zip(given.optionality.iter()))
+        .map(|(variable, (annotation, optionality))| {
+            let is_optional = *optionality == VariableOptionality::Optional;
+            let annotations = match annotation {
+                FunctionParameterAnnotation::AnyConcept => PipelineVariableAnnotation::Instance(vec![]),
```
krishnangovindraj: This should be flagged unreachable.

krishnangovindraj (commented):
It's a lot less scary if you check off the trivial files first. 
The changes in compiler/executor are quite small/trivial. 
It's the server package changes that are bulky.

compiler/annotation/mod.rs
```diff
@@ -119,6 +119,13 @@ typedb_error!(
             variable: String,
             source_span: Option<Span>,
         ),
+        CouldNotResolveGivenRowDeclaredType(
+            18,
+            "An error occurred when trying to resolve the type of the given rows at index: {index}.",
```
flyingsilverfin, line 124: Maybe we can improve this error with a bit more of a 'what went wrong' or 'what is expected' (or whenever the typical errors will crop up from)

query/query_cache.rs
```diff
@@ -47,10 +47,11 @@ impl QueryCache {
     pub(crate) fn get(
         &self,
         preamble: Arc<Vec<Function>>,
+        given_rows: Arc<Option<TranslatedGiven>>,
```
flyingsilverfin: This naming is confusing: I was about to leave a comment that the 'given rows' shouldn't be in the cache key. But this is actually just the `given` spec - so just rename it to `given` here! Check places this comes from and flows to for this naming mismatch as well please

query/query_manager.rs
```diff
@@ -130,21 +153,21 @@ impl QueryManager {
         // 1: Translate
         let TranslatedPipeline {
             translated_preamble,
+            translated_given: translated_given,
```
flyingsilverfin: nb a couple of places you have these redundancies already in this file

query/query_manager.rs
```diff
@@ -491,3 +550,25 @@ fn annotate_and_compile_query(
     compile_profile.compilation_finished();
     Ok(executable_pipeline)
 }
+
+fn validate_given(
```
flyingsilverfin: The primary function is not actually validation, but translating to `batch` right? So naming should reflect that (eg. `convert_to_batch` but it returns a Result type).

Secondly, if it's creating Batches, is it better off in the //executor package?

server/service/http/message/query/concept.rs
```diff
@@ -44,6 +47,7 @@ pub struct RelationResponse {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase", tag = "kind", rename = "attribute")]
 pub struct AttributeResponse {
+    pub iid: String,
```
flyingsilverfin, line 46: Oh are we adding this? I thought we'd already have it! Eg. we have it in grpc right?

Note: i think today we mentioned that attribute IID may not be stable as a way to look up a concept...?

server/service/http/message/query/concept.rs
flyingsilverfin: Good to note - we'll want that!

tests/behaviour/query/language/given.rs
```diff
@@ -0,0 +1,23 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+#![allow(unexpected_cfgs, reason = "features defined in Bazel targets aren't currently communicated to Cargo")]
+
+use steps::Context;
+
+#[tokio::test]
+async fn test_write_given() {
```
flyingsilverfin, line 12: why `write_given`? it's just `given.rs` no?

MODULE.bazel
```diff
@@ -106,12 +106,12 @@ workspace_refs = use_repo_rule("@typedb_bazel_distribution//common/workspace_ref
...
-#        "typeql+": "a1fae91ebd68ed2d7b2eb5035c1287c3bc66789b",
-#        "typedb_protocol+": "25ab3fb02715062bd036f8b83908de92b106a202",
+        "typeql+": "e16626a2129c554436195384a9fda3a3ea36fa2f",
+        "typedb_protocol+": "bb9c513d4dcff269feb88015dcf99354ad378657",
     },
     workspace_tag_dict = {
-        "typeql+": "3.11.0",
-        "typedb_protocol+": "3.11.0",
+#        "typeql+": "3.11.0-rc0",
+#        "typedb_protocol+": "3.11.0-rc0",
```
flyingsilverfin, line 114: Don't forget!

query/analyse.rs
```diff
@@ -109,6 +116,28 @@ pub struct FunctionStructureAnnotations {
     pub body: PipelineStructureAnnotations,
 }
 
+fn build_given_annotations(given: &AnnotatedGiven) -> GivenStructureAnnotations {
```
flyingsilverfin, line 119: Can you pop an example of a Given output (can be from HTTP or gRPC) into the PR comments or here to show what the output of Analyze will be with a 'given' query pls?

ir/translation/pipeline.rs
flyingsilverfin: small enough to have here IMO right now

ir/translation/pipeline.rs
flyingsilverfin: i think what you've got is good - it's a stage that is a type validation/declaration that converts inputs into the expected format. Kinda the inverse of Fetch in some ways - I think that's good for now until we have clear evidence for the other way.

ir/translation/mod.rs
```diff
@@ -115,3 +126,24 @@ macro_rules! verify_variable_available {
     };
 }
 pub(super) use verify_variable_available;
+
+pub fn parse_iid(iid: &str) -> Result<ByteArray<THING_VERTEX_MAX_LENGTH>, ()> {
```
flyingsilverfin, line 130: This the right place for this? IR shouldn't be 'executing' rows I think? Although not too sure about this one without digging deeper

ir/translation/constraints.rs
flyingsilverfin: ah ok that explains it - cheers

ir/pipeline/mod.rs
flyingsilverfin: We will allow function args to be optional though won't we? Or does this change already set that up

executor/pipeline/pipeline.rs
```diff
@@ -234,19 +244,26 @@ impl<Snapshot: WritableSnapshot + 'static> Pipeline<Snapshot, WritePipelineStage
...
-        let initial_iterator =
-            WriteStageIterator::Initial(Box::new(InitialIterator::new(crate::batch::FixedBatch::SINGLE_EMPTY_ROW)));
+        let initial_iterator = WriteStageIterator::Initial(Box::new(InitialIterator::new(given_rows)));
+        let mut stages = if let Some(given) = executable_given {
+            let mut stages = Vec::with_capacity(executable_stages.len() + 1);
+            stages.push(WritePipelineStage::Given(Box::new(GivenStageExecutor::new(given))));
+            stages
+        } else {
+            Vec::with_capacity(executable_stages.len())
+        };
```
flyingsilverfin: Will this change without our conversation this morning about converting the concept values given from the client side into the expected declared type here? (eg. casting or filling in missing optional values?)

not thought deeply about this just noting this feels like it's approximately the area that would adapt

flyingsilverfin (changes requested):

server/service/http/message/query/concept.rs
krishnangovindraj: It's tested now through the integration test on the driver.

query/analyse.rs
krishnangovindraj: https://github.com/typedb/typedb-protocol/pull/251/changes ?

ir/translation/mod.rs
krishnangovindraj: Ah ok, I saw your other comment.
~~It's already been used by the IID constraint here. This is a translation without the encoding so it's not in the execution path.~~

executor/pipeline/pipeline.rs
krishnangovindraj: That'll likely be handled inside the GivenStageExecutor

query/analyse.rs
krishnangovindraj: It's the same type as the conjunction annotations so it's not new. I'll have to figure out the curl scripts to get the HTTP response.

krishnangovindraj: Good catch. It's a little strange for the variables to be inside the query. I guess fetch doesn't have this problem since it has it's own pipelines. We could move `given` inside the query, but that would make it feel like a full stage.
```
{
    "source": "given $x: person, $n: string; match $x has name == $n;",
    "given": {
        "variables": [
            "0",
            "1"
        ],
        "variableAnnotations": {
            "1": {
                "isOptional": false,
                "tag": "value",
                "valueTypes": [
                    "string"
                ]
            },
            "0": {
                "isOptional": false,
                "tag": "instance",
                "annotations": [
                    {
                        "kind": "entityType",
                        "label": "person"
                    }
                ]
            }
        }
    },
    "query": {
        "conjunctions": [
            {
                "constraints": [
                    {
                        "textSpan": {
                            "begin": 48,
                            "end": 53
                        },
                        "tag": "comparison",
                        "lhs": {
                            "tag": "variable",
                            "id": "2"
                        },
                        "rhs": {
                            "tag": "variable",
                            "id": "1"
                        },
                        "comparator": "=="
                    },
                    {
                        "textSpan": {
                            "begin": 39,
                            "end": 53
                        },
                        "tag": "has",
                        "owner": {
                            "tag": "variable",
                            "id": "0"
                        },
                        "attribute": {
                            "tag": "variable",
                            "id": "2"
                        }
                    },
                    {
                        "textSpan": {
                            "begin": 43,
                            "end": 47
                        },
                        "tag": "isa",
                        "instance": {
                            "tag": "variable",
                            "id": "2"
                        },
                        "type": {
                            "tag": "label",
                            "type": {
                                "kind": "attributeType",
                                "label": "name",
                                "valueType": "string"
                            }
                        }
                    }
                ],
                "annotations": {
                    "variableAnnotations": {
                        "1": {
                            "isOptional": false,
                            "tag": "value",
                            "valueTypes": [
                                "string"
                            ]
                        },
                        "0": {
                            "isOptional": false,
                            "tag": "instance",
                            "annotations": [
                                {
                                    "kind": "entityType",
                                    "label": "person"
                                }
                            ]
                        },
                        "2": {
                            "isOptional": false,
                            "tag": "instance",
                            "annotations": [
                                {
                                    "kind": "attributeType",
                                    "label": "name",
                                    "valueType": "string"
                                }
                            ]
                        }
                    }
                }
            }
        ],
        "stages": [
            {
                "tag": "match",
                "block": 0
            }
        ],
        "variables": {
            "1": {
                "name": "n"
            },
            "0": {
                "name": "x"
            },
            "2": {
                "name": null
            }
        },
        "outputs": [
            "0",
            "1"
        ]
    },
    "preamble": [],
    "fetch": null
}
```

server/service/http/message/query/concept.rs
krishnangovindraj: >  i think today we mentioned that attribute IID may not be stable as a way to look up a concept
Eh? Isn't it as stable as other IIDs?

query/given_rows.rs
```diff
@@ -0,0 +1,113 @@
...
+pub trait GivenRows {
+    fn variables(&self) -> &[String];
+    fn into_batch_mapped(
+        self,
+        declared_variable_positions: &HashMap<&str, VariablePosition>,
+    ) -> Result<Batch, GivenRowDecodeError>;
+}
+
+pub trait GivenRowsDecoder<T> {
+    fn decode(what: T) -> Result<GivenRowEntry, GivenRowDecodeError>;
```
flyingsilverfin: normally we would use `item`

server/service/grpc/transaction_service.rs
```diff
@@ -1797,3 +1855,67 @@ impl QueueOptions {
...
+        GivenRowDecodeError::ConceptDecode {
+            typedb_source: Box::new(ConceptDecodeError::$variant {
+                iid: HexBytesFormatter::borrowed($iid.as_ref()).into_owned(),
+            }),
+        }
+    };
+}
+
+pub struct GivenRowsDecoderGrpc;
+impl GivenRowsDecoder<typedb_protocol::query::req::GivenEntry> for GivenRowsDecoderGrpc {
```
flyingsilverfin: Is this because we can't directly impl the trait on the grpc message directly? Ie. the struct is not local and the trait is also not local?

follow up question: is there a way to roll into the decide method into the `impl GivenRows` trait itself to avoid two traits?

server/service/http/message/query/concept.rs
flyingsilverfin: Err hmm i think it probably is actually now that I think about it. Ah wait ok - in TypeDB Cluster it might not be (?) I actually think it is just might be worth having a double think about

query/query_manager.rs
```diff
@@ -491,3 +546,31 @@ fn annotate_and_compile_query(
     compile_profile.compilation_finished();
     Ok(executable_pipeline)
 }
+
+fn validate_and_decode_given(
```
flyingsilverfin, line 550: I feel like this is not actually doing any obvious decoding that is visible here - it's more like `try_convert_to_batch` or something? The validation is in the try? I dont' mind this name tho

flyingsilverfin: So to summarize: we get a GRPC batch. Then we convert it into a Batch via the trait. Where/when do we validate that the conversion from the input given types to the expected given types works? Eg. we can be given a row with an integer but we can use that for where the executable has a 'Decimal' right?

query/given_rows.rs
```diff
@@ -0,0 +1,113 @@
...
+}
```
flyingsilverfin: My instinct is that we might be able to simplify this somehow - you always implement both of these adjacently i think, so maybe that can be merged (or maybe not if we can't it's fine)

query/given_rows.rs
```diff
@@ -0,0 +1,113 @@
...
+        ConceptDecode(1, "An error occurred while decoding the provided concept.", typedb_source: Box<ConceptDecodeError>),
+        InvalidIIDFormatForGivenEntry(2, "The provided iid string '{iid}' was invalid.", iid: String),
+        ParsingValueFailedForGivenEntry(3, "An error occured while parsing the provided value '{value}'.", value: String, typedb_source: typeql::Error),
+        TranslatingValueFailedForGivenEntry(4, "An error occured while translating the provided value '{value}'.", value: String, typedb_source: LiteralParseError),
+        GivenRowsVariableWasNotDeclared(5, "The variable '{variable}' was not declared in the query.", variable: String),
+    }
+}
+
+#[derive(Debug)]
+pub struct GivenRowsSimple {
```
flyingsilverfin, line 111: This one is weird to me - is it just for tests? Or for traits to be passed in? Maybe let's simplify it further and call it `GivenRowsAbsent` or `NoGivenRows` or something? Or maybe it hints that something should be an optional somewhere...

flyingsilverfin (changes requested):

query/given_rows.rs
krishnangovindraj: Yeah, I had to complicate it to support dictionaries over the wire.
Then I simplified the HTTP path so even a dictionary can be converted to the header-rows format.

query/given_rows.rs
krishnangovindraj: It's probably the right type to use for internal queries such as the ones in `system`, since we'll be dealing in concepts. It's also used in tests so the path remains tested. It makes more sense to have just one of these than a separate `test` one and a `GivenRowsAbsent`

server/service/grpc/transaction_service.rs
krishnangovindraj: Good point. Done.

tests/behaviour/query/language/given.rs
krishnangovindraj: Mostly an arbitrary grouping but it's majorly a write stage.
I could rename it to `test_given`. I'd just need an extra line in the factory job.

flyingsilverfin: it's not by definition a write stage, it's a parametrized query - let's run it as such

query/analyse.rs
flyingsilverfin: Given on its own and sibling to `fetch` seems right

compiler/annotation/mod.rs
krishnangovindraj: That's in the source TypeInference error .

query/query_manager.rs
krishnangovindraj: `validate_and_decode_given` is what it is now.
If the compiler/executor doesn't need to know about something it's probably easier if they don't, so I'd rather have it here. Ideally the executor remains a very abstract machine that gives everything in the format it expects.

ir/pipeline/mod.rs
krishnangovindraj: Unclear if we will

flyingsilverfin (approved):

PR #7813, opened 2026-05-20
Optimize iterator KMerge

flyingsilverfin:
##  Product Change & Motivation 

We optimize the read path of our executors and iterators by letting the core KMerge iterator box iterators that are moved in and out of the internal heap frequently, and sometimes bypassing it entirely.

We also introduce new executor microbenchmarks, which are extremely useful for optimizing the overhead of our iterators. See `//executor/benches/BUILD`, which also shows how to produce a flamegraph — confirmed working on Linux and on Mac 26.

```
  >> bazel run --compilation_mode=opt //executor/benches:bench_instructions -- --bench [ "has_reverse_unbound/multi_type" ] # bench specific test

  # To generate a flamegraph instead of timing, pass --profile-time <seconds>.
  # Criterion will run each matched bench under pprof's SIGPROF sampler and write
  # a flamegraph.svg per bench function.
  >> bazel run --compilation_mode=opt //executor/benches:bench_instructions -- --bench --profile-time 30 [ "has_reverse_unbound/multi_type" ] # bench specific test
```
##  Implementation
  
- Box PeekWrapper inside KMergeBy's heap. Sift swaps collapse to a single pointer move; comparator and dispatch unchanged via auto-deref. has_reverse_unbound/multi_type: 246 ms → 44 ms.
  
- Bypass KMerge-of-1. Short-circuit KMergeBy::new(iters) when iters.len() == 1. ~7% win on the common single-attribute-type case
  
- Microbench infrastructure. New executor/benches/bench_instructions.rs (criterion + pprof)


krishnangovindraj (changes requested):
Given the flamegraph shows a bottleneck on memcpy, and we now know there's 1.4k bytes being moved around in the heap by memcpy, we should try to store something smaller (`(Key, IteratorIndex)`?) in the BinaryHeap and see how that affects performance. (And how much of a difference this change makes after that change)

Also, if we're going to be doing a lot of this (as has been the case), it's probably time to create proper micro-benchmarks so we're principled about these changes. We're currently doing lots of work (even if it's Claude) in writing the benchmarks and running the investigations only to throw it off as a one-off investigation

concept/thing/thing_manager.rs
```diff
@@ -395,7 +395,7 @@ impl ThingManager {
             AttributeID::Decimal(id) => Ok(Value::Decimal(id.read().as_decimal())),
             AttributeID::Date(id) => Ok(Value::Date(id.read().as_naive_date())),
             AttributeID::DateTime(id) => Ok(Value::DateTime(id.read().as_naive_date_time())),
-            AttributeID::DateTimeTZ(id) => Ok(Value::DateTimeTZ(id.read().as_date_time())),
+            AttributeID::DateTimeTZ(id) => Ok(Value::DateTimeTZ(Box::new(id.read().as_date_time()))),
```
flyingsilverfin: Boxing large variants has a large impact on read perf

executor/benches/bench_instructions.rs
flyingsilverfin, line 1: Microbenchmark to test the efficiency of our individual executor iterators and iterate quickly on optimizing ovhereads

executor/benches/bench_instructions.rs
```diff
@@ -0,0 +1,307 @@
...
+            let iter = has_reverse_multi_exec
+                .get_iterator(&has_reverse_multi_ctx, MaybeOwnedRow::empty(), StorageCounters::DISABLED)
+                .unwrap();
+            assert_count(iter, count_multi);
+        })
+    });
+    group.finish();
+}
+
+struct FlamegraphProfiler<'a> {
```
flyingsilverfin, line 454: Flamegraphs work on mac and linux - see BUILD for instructions on how to use

executor/instruction/indexed_relation_executor.rs
```diff
@@ -352,13 +352,24 @@ impl IndexedRelationExecutor {
...
                     },
                 );
-                let merged_tuples: KMergeBy<IndexedRelationTupleIterator<IndexedRelationsIterator>, TupleOrderingFn> =
-                    KMergeBy::new(iterators, unsafe_compare_result_tuple);
-                Ok(TupleIterator::IndexedRelationsMerged(SortedTupleIterator::new(
-                    merged_tuples,
-                    self.tuple_positions.clone(),
-                    &self.variable_modes,
-                )))
+                if iterators.len() == 1 {
```
flyingsilverfin, line 355: These bypasses are still a 7% perf win compared to not using the optimized KMerge iterator

encoding/value/value.rs
```diff
@@ -41,10 +41,10 @@ pub enum Value<'a> {
     Decimal(Decimal),
     Date(NaiveDate),
     DateTime(NaiveDateTime),
-    DateTimeTZ(DateTime<TimeZone>),
+    DateTimeTZ(Box<DateTime<TimeZone>>),
     Duration(Duration),
     String(Cow<'a, str>),
-    Struct(Cow<'a, StructValue<'static>>),
+    Struct(Box<Cow<'a, StructValue<'static>>>),
```
krishnangovindraj: `DateTime<TimeZone>` is bloated because it uses the offset inline. Boxing it is overkill and I'd rather use a more efficient inline packing. When we encode it, we encode one i64 and two u32s. So even with all fields aligned, this doesn't NEED to be larger than 30 bytes. It might be worth storing it as something more compact and converting it to a full `DateTime<Timezone` just-in-time need to. 

For structs, the bloat comes from the fact that we store the `ByteArray` directly in the `DefinitionKey`. We should just store the u16.

.bazelrc
```diff
@@ -14,7 +14,7 @@ build --macos_minimum_os=11.0
 try-import ./.bazel-remote-cache.rc
 
 # Allow empty globs (many checkstyle_test patterns may not match)
-common --noincompatible_disallow_empty_glob --experimental_repository_downloader_retries=5 --http_timeout_scaling=2.0
+common --@rules_rust//:rustfmt.toml=//:rustfmt_config --noincompatible_disallow_empty_glob --experimental_repository_downloader_retries=5 --http_timeout_scaling=2.0
```
flyingsilverfin, line 17: This is a drive-by: we don't need to list it for every time we run rustfmt test now!

.factory/automation.yml
```diff
@@ -21,7 +21,7 @@ build:
         bazel build --config=ci //... 
         bazel run --config=ci @typedb_dependencies//tool/checkstyle:test-coverage
         bazel test --config=ci $(bazel query 'kind(checkstyle_test, //...)') 
-        bazel test --config=ci $(bazel query 'kind(rustfmt_test, //...)') --@rules_rust//:rustfmt.toml=//:rustfmt_config
+        bazel test --config=ci $(bazel query 'kind(rustfmt_test, //...)')
```
flyingsilverfin, line 28: Added config to the bazelrc as a drive-by

flyingsilverfin:
We'll wait for a sanity review from @dmitrii-ubskii !

encoding/value/value.rs
```diff
@@ -41,7 +41,7 @@ pub enum Value<'a> {
     Decimal(Decimal),
     Date(NaiveDate),
     DateTime(NaiveDateTime),
-    DateTimeTZ(DateTime<TimeZone>),
+    DateTimeTZ(Box<DateTime<TimeZone>>),
```
dmitrii-ubskii: This only "wins" us 16 bytes at the cost of indirection whenever we want to do any operations with the datetime value.

Despite what it says in the PR description, the size of `DateTime<TimeZone>` is only 48 bytes, with `Value` containing the unboxed value growing to 56 bytes (+8 bytes for discriminant + padding).

The smallest `Value<'_>` can be is 40 bytes since `Cow<'_, str>` alone is 32 bytes.

In particular, cloning `DateTimeTZ` values would now also involve indirection. Currently this only happens for `String` values.

We clone values all the time between executor steps, and when we return results to the user. It's unavoidable for `String`s, but we shouldn't do it for values we don't have to. Notably, the microbenchmark introduced in this PR is completely blind to this issue, because it only looks at a single `Has` executor, and it also does not involve `DateTimeTZ` in the first place.

compiler/executable/match_/instructions/mod.rs
```diff
@@ -64,7 +64,7 @@ pub struct VariableModes {
 }
 
 impl VariableModes {
-    fn new() -> Self {
+    pub fn new() -> Self {
         VariableModes { modes: HashMap::new() }
     }
```
dmitrii-ubskii, line 69: Not super happy about exposing these. I kinda wish there was a way to only export some symbols for tests / benches in Rust but alas.

dmitrii-ubskii (changes requested):

dmitrii-ubskii (approved):

